PR #456, opened 2021-06-10
Use forked api-extractor build - remove ts-ignores throughout codebase

JamesBurnside:
# What
* Use fork of api-extractor built from: https://github.com/microsoft/rushstack/pull/1916
  * Fork located at: `common\config\api-extractor\fork-pr-1916\microsoft-api-extractor-7.13.2-pr1916.1.0.tgz`
* Remove all ts-ignores throughout codebase as fork fixes this bug
* Remove eslint rules temporarily allowing the ts-ignores

# Why
There seems to be no traction in getting https://github.com/microsoft/rushstack/pull/1916 merged in and we have had the tsignore workaround for too long.

# How Tested


packages/communication-react/review/communication-react.api.md
```diff
@@ -35,6 +35,8 @@ import { IContextualMenuItem } from '@fluentui/react';
 import { IStyle } from '@fluentui/react';
 import { MediaStreamType } from '@azure/communication-calling';
 import { MicrosoftTeamsUserKind } from '@azure/communication-common';
+import { OutputParametricSelector } from 'reselect';
```
alcail, line 38: nit: why not the following?

import { OutputParametricSelector, OutputSelector } from 'reselect';

alcail (approved):

packages/communication-react/review/communication-react.api.md
JamesBurnside: I think due to a small bug in the fix for the api-extractor original bug, but not something we have control over

prprabhu-ms (approved):

PR #457, opened 2021-06-10
Add customized render to chat api as proposed

PorterNan:
Add customized render to chat api as proposed
Clean up some hooks code in composite

# What

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -70,12 +72,12 @@ const getDefaultComponents = (
...
         key={`typing indicator display string ${index}`}
       >
-        {index < typingUsers.length - 1 ? typingUser.displayName + ', ' : typingUser.displayName}
+        {index < typingUsersToRender.length - 1 ? typingUser.displayName + ', ' : typingUser.displayName}
       </span>
     );
   });
 
-  const countOfUsersNotMentioned = typingUsers.length - typingUsersMentioned.length;
+  const countOfUsersNotMentioned = typingUsersToRender.length - typingUsersMentioned.length;
```
prprabhu-ms, line 80: Should this continue to use `typingUsers`? It makes sense to not name participants without a `displayName` property but they should be included in the count of number of participants that are typing.

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -88,7 +90,8 @@ const getDefaultComponents = (
 };
 
 const defaultTypingString = (typingUsers: CommunicationParticipant[]): string => {
-  return typingUsers.length > 0 ? (typingUsers.length > 1 ? ' are typing...' : ' is typing...') : '';
+  const typingUsersToRender = typingUsers.filter((typingUser) => typingUser.displayName !== undefined);
```
prprabhu-ms, line 93: Similar to above, should this continue to use `typingUsers` because participants without a `displayName` are still typing ;)

packages/react-composites/src/composites/CallComposite/hooks/usePropsFor.ts
```diff
@@ -1,72 +1,15 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import {
```
prprabhu-ms, line 4: Changes to this file seem unrelated to the rest of this PR.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -47,6 +57,17 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
   const threadStatusProps = useAdaptedSelector(getThreadStatusProps);
 
+  const onRenderMessageAvatar = useMemo(
+    () => onRenderAvatar && ((userId: string) => onRenderAvatar(userId, 'chatThread')),
+    [onRenderAvatar]
+  );
+
+  const onRenderParticipantAvatar = useMemo(
+    () =>
+      onRenderAvatar && ((participant: CommunicationParticipant) => onRenderAvatar(participant.userId, 'chatThread')),
```
prprabhu-ms, line 67: s/chatThread/participantList/

packages/react-composites/src/composites/ChatComposite/hooks/usePropsFor.ts
```diff
@@ -1,37 +1,15 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { ChatClientState } from 'chat-stateful-client';
```
prprabhu-ms, line 4: These changes don't seem functionally related to this PR.

prprabhu-ms (approved):
LGTM, but please remove changes that are unrelated functionally to keep the PR crisp.

prprabhu-ms:
+1 to this not being a `breaking-change`: Adding an optional argument to a function signature is non-breaking in JS. Semantically, the default value for this optional argument is also consistent with previous behaviour.

packages/communication-react/review/communication-react.api.md
```diff
@@ -522,7 +522,9 @@ export type ChatCompositeClientState = {
 export type ChatCompositeProps = {
     adapter: ChatAdapter;
     fluentTheme?: PartialTheme | Theme;
-    onRenderAvatar?: (userId: string) => JSX.Element;
+    onRenderAvatar?: (userId: string, avatarType?: 'chatThread' | 'participantList') => JSX.Element;
```
JamesBurnside, line 525: this list could get huge (which might be totally fine), but are there other options for what this interface might look like?
e.g.
```
customRenderers?: {
  onRenderAvatar: ()=> JSX.Element,
  onRenderAvatar: ()=> JSX.Element etc.
}
```

(not saying this is better! Just wondering if we explored options for what this customization interface would look like)

JamesBurnside: Then would the documentation highlight what the `onRenderAvatar` would change?

JamesBurnside: Maybe the interface is a good thing to bring up to the azure review board and get their opinions

JamesBurnside (approved):
This is awesome! Are we planning to add new custom render functions as we decide/as requested by users? Or is there a plan to cover certain customization scenarios?

packages/react-components/src/components/TypingIndicator.tsx
PorterNan: It was filtered outside of this component previously, I just moved the code into the function to make code more clear, it is still sharing the same logic
In previous design, only people with displayname will be shown in the typing indicator like
A, B are typing, this count is used for determine whether it is 'are typing' or 'is typing'

packages/communication-react/review/communication-react.api.md
PorterNan: Agree, currently this api allows developer to use just one way to render all the avatar(that could be most likely the case), but if they are willing to, they could render different avatar in different places by second parameter

PR #458, opened 2021-06-10
Build time improvements: Eject samples from CRA

JamesBurnside:
# What
Eject samples from CRA
* Use a barebones webpack (and webpack dev server for development)

Also:
* Upgrade to webpack5 to fix dev server hanging issue
* Fix dependencies that broke build when `rush update --full` performed
* Remove use of preval to get build time and use webpack variable injection instead
* Upgrade rush and pnpm to try fix build issue with api-extractor after rush update full
  * Ended up having to extract api-extractor to a project in out repo - rush+pnpm doesn't handle directly using tarballs, making this PR pretty enormous (but is just the tarball expanded into files)
  * Hopefully we can remove all these very soon:
      image

# Why
Perf
* We have large CI build times with easiest large gains coming from samples

Dev inner loop improvements
* No longer rely on rush module linking for samples and directly alias to source code
* This massivly improves local build times, more reliable hot reloading, much faster hot reloading and less confusing output in the terminal

# Perf

CI Runs:
| Before | After | Improvement |
| -- | -- | -- |
| 13-16 minutes | 8-10minutes | -5-6 minutes |
| image | image |

Local Tests for Calling Sample:
| | Before | After | Improvement |
| -- | -- | -- | -- |
| Production Build (dependencies already built) | 63s | 51s | -12s |
| Production Build (dependencies not built) | 5min 4s | 51s | -4min 15s |
| `npm run start` (dependencies already built) (until homepage loaded) | 2min 17s | 27s | -1min 50s |
| `npm run start` (dependencies not built) (until homepage loaded) | 3min 56s | 27s | -2min 29s |
| Hot reload (change made in `react-components` dir) | 20 seconds (+ other packlets still recompiling) | 4s | -16s |

Profiler still crashes 😥

# How Tested
* Samples working locally
* Deployed calling sample to hosted azure web app - tested desktop edge and ios safari


prprabhu-ms (commented):
The build time improvements are amazing. Bravo!

I'm a bit conflicted about sample apps using the uncompiled package sources though.
Is the source mapping also required for build time improvements, or is it only for hot reloading?

The concern would be some build related issues slipping through the cracks / the samples not quite working if they code is copied as-is to an independent project (which customers are expected to do to kickstart their apps).

JamesBurnside:
> The build time improvements are amazing. Bravo!
> 
> I'm a bit conflicted about sample apps using the uncompiled package sources though.
> Is the source mapping also required for build time improvements, or is it only for hot reloading?
> 
> The concern would be some build related issues slipping through the cracks / the samples not quite working if they code is copied as-is to an independent project (which customers are expected to do to kickstart their apps).

Just for hot reloading (though build time will also be affected as first the dependencies will need built then the sample app webpack run).

I agree with the concerns but I think the solution is to move samples to using the public npm package (either latest of alpha, we can decide), then for local dev you run a script that points the sample to the local communication-react folder to support hot reloading.

prprabhu-ms (approved):

prprabhu-ms:
Land this already so I can speed up my edit-compile-test cycle for the day :D

anjulgarg (approved):

JamesBurnside:
> Land this already so I can speed up my edit-compile-test cycle for the day :D

I don't feel hesitant anymore having to run the samples! It's really nice 😅

anjulgarg:
Thanks for landing this @JamesBurnside 
Did you have to remove polyfills and shims in webpack config?
I remember the original ejected webpack being a very large file. 
Are we removing something that might cause our sample to stop working as expected on a specific browser?
If yes, I think we should test the generated sample thoroughly on each major browser before we merge this in.

PR #459, opened 2021-06-11
Replace maps with objects in `CallClientState`

prprabhu-ms:
# Why
Redux recommends using plain objects instead of maps in state. e.g., this is required for time-travel debugging for work.

# How Tested
* `rush build`
* `rush test`
* Storybook `CallComposite` canvases.
* Local sample run.

# Process & policy checklist


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -43,9 +43,9 @@ export class CallContext {
 
   constructor(userId: CommunicationUserKind, maxListeners = 50) {
     this._state = {
-      calls: new Map<string, CallState>(),
+      calls: {},
```
JamesBurnside, line 46: Can we give these a type?
maybe (untested)
```
Record<string, CallState>
```

JamesBurnside (approved):

alcail (approved):

PR #461, opened 2021-06-11
Replace `Map` with `Object` for `ChatClientState.threads`

prprabhu-ms:
# Why
Maps don't play well with Redux state store.

For more context, a github issue with discussions on why maps are a bad idea in Redux.
https://github.com/reduxjs/redux/issues/1499

# How Tested
* `rush build`
* `rush test`
* Local storybook
* Local Chat sample

# Process & policy checklist


JamesBurnside (approved):

packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -77,7 +77,7 @@ export class ChatContext {
   }
 
   public createThreadIfNotExist(threadId: string, properties?: ChatThreadProperties): boolean {
-    const exists = this.getState().threads.has(threadId);
+    const exists = Object.prototype.hasOwnProperty.call(this.getState().threads, threadId);
```
anjulgarg, line 80: Curious, is this not equivalent to
`const exists = this.getState().threads[threadId]`

anjulgarg (approved):

packages/chat-stateful-client/src/ChatContext.ts
prprabhu-ms: The reason is a bit pedantic:

```
const a = {a: 1, b: undefined};

console.log('b', a['b'], Object.prototype.hasOwnProperty.call(a, 'b'));
console.log('c', a['c'], Object.prototype.hasOwnProperty.call(a, 'c'));
```

Gives:

    b undefined true
    c undefined false

PR #467, opened 2021-06-14
CI build time improvement - replace slow github marketplace action

JamesBurnside:
# What
Remove the create-issue github marketplace action we are using with basic direct curl request.

# Why
It takes 30s every CI - even when its not used (it has to pull and build the action at the start of the run regardless of if the step using it is triggered).
Direct step uses 0s when not used.

# How Tested
Tested creating an action with CI_FAILURE label: https://github.com/Azure/communication-ui-library/runs/2821940435?check_suite_focus=true
Tested no secondary action is created if one exists already: https://github.com/Azure/communication-ui-library/runs/2822046030?check_suite_focus=true


prprabhu-ms (approved):
Shut up and take my money.

.github/workflows/ci.yml
```diff
@@ -72,9 +72,14 @@ jobs:
...
-          filename: .github/failed_ci_build_template.md
-          update_existing: true
+        run: |
+          # check for an issue that is already open
+          curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/search/issues?q=org:Azure+repo:communication-ui-library+label:CI_FAILURE+state:open > $HOME/ci_issues.json
+          issue_count=$(jq -r '.total_count' $HOME/ci_issues.json)
+          if [ $issue_count -gt 0 ]; then
+            echo "No need to create new issue, one already exists"
+          else
+            curl -X POST -H 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' https://api.github.com/repos/Azure/communication-ui-library/issues -d '{"title":"[TEST IGNORE] Main branch CI Failed", "body": "Main CI Pipeline has failed. Please investigate.", "labels":["CI_FAILURE"]}'
```
prprabhu-ms, line 82: Remove `[TEST IGNORE]` from subject.

JamesBurnside: nice catch 🚀

alcail (approved):

PR #468, opened 2021-06-14
Localization with provider.

mgamis-msft:
# What

Implemented localization with a simple provider. This is PR is mainly for the LocalizationProvider but I also included some major components like TypingIndicator which shows plural cases and MessageThread which has dates. This also uses two potential testing methods which is jest snapshots and enzyme.

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -0,0 +1,124 @@
...
+  storage?: Storage;
+  children: React.ReactNode;
+};
+
+export const LocalizationProvider = (props: LocalizationProviderProps): JSX.Element => {
+  const { children, initialLocale, locales, storage, localeStringsLoader } = props;
+
+  useEffect(() => {
+    if (storage) {
+      const cachedLocale = storage.getItem(LOCALE_CACHE_KEY);
```
alkwa-msft, line 54: I don't think we want to require the use of local storage at the component layer. 
This should be an implementation detail that contoso needs to figure out

We should say if you give us a big set of strings we'll disperse them across the composite for you

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -66,6 +67,10 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
   const [menuHidden, setMenuHidden] = useState<boolean>(true);
   const containerRef = useRef<HTMLDivElement>(null);
   const theme = useTheme();
+  const { strings } = useLocale();
```
alkwa-msft, line 70: I think we want to take in the "locale strings" as just a string prop. It should be in the composite code to hook in the strings from the localization provider as the correct prop to be used in a component.

JamesBurnside:
PR description still says this is a draft - should this be reviewed as a draft or something to merge?

packages/communication-react/rollup.config.js
```diff
@@ -3,15 +3,27 @@
...
     input: './dist/dist-esm/communication-react/src/index.js',
     output: {
-      file: './dist/dist-cjs/communication-react/index.js',
+      dir: './dist/dist-cjs/communication-react',
       format: 'cjs',
       sourcemap: true
-    }
+    },
+    plugins: [
+      commonjs(),
```
JamesBurnside, line 21: The common rollup config has `[commonjs(), sourcemaps(), svg()]` (`common\config\rollup\rollup.config.js`), can the json one be added to the common config and these removed from this file?

packages/communication-react/rollup.config.js
```diff
@@ -3,15 +3,27 @@
...
+import svg from 'rollup-plugin-svg';
+import json from '@rollup/plugin-json';
 
 export default [
   {
     ...commonConfig(Package),
     input: './dist/dist-esm/communication-react/src/index.js',
     output: {
-      file: './dist/dist-cjs/communication-react/index.js',
+      dir: './dist/dist-cjs/communication-react',
```
JamesBurnside, line 16: what does changing from file to dir do here?

packages/communication-react/review/communication-react.api.md
```diff
@@ -726,6 +726,9 @@ export type DefaultChatHandlers = {
     onLoadPreviousChatMessages: (messagesToLoad: number) => Promise<boolean>;
 };
 
+// @public (undocumented)
```
JamesBurnside, line 729: ensure all new public functionality has documentation

packages/react-components/rollup.config.js
```diff
@@ -2,6 +2,29 @@
...
 import Package from './package.json';
-import commonConfig from '../../common/config/rollup/rollup.config';
+import commonjs from '@rollup/plugin-commonjs';
+import sourcemaps from 'rollup-plugin-sourcemaps';
+import svg from 'rollup-plugin-svg';
+import json from '@rollup/plugin-json';
 
-export default commonConfig(Package);
+export default {
+  context: 'window',
```
JamesBurnside, line 11: can we continue using the commonConfig here?

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -49,13 +50,18 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
...
+
+  const defaultRenderText = useMemo(
+    () => (props?: IButtonProps): JSX.Element => {
+      return (
+        <Label key={'videoLabelKey'} className={mergeStyles(controlButtonLabelStyles, props?.styles?.label)}>
+          {props?.checked ? strings.camera_button_on_text : strings.camera_button_off_text}
+        </Label>
+      );
+    },
+    [strings]
```
JamesBurnside, line 63: Will this still rerender the component when the props have changed? (props.checked and props.styles.label)

packages/react-components/package.json
```diff
@@ -95,7 +95,10 @@
     "ts-node": "^9.1.1",
     "typescript": "4.1.5",
     "uuid": "^8.1.0",
-    "webpack": "5.38.1"
+    "webpack": "5.38.1",
+    "@rollup/plugin-json": "~4.1.0",
```
JamesBurnside, line 99: nit: alphabetize these

packages/react-components/src/components/ParticipantItem.tsx
JamesBurnside: +1 to this, we need to keep components completely pure (they should never rely on being wrapped in a provider).
I think we'll need to create a task to remove the `useTheme` from components also (getTheme should hopefully work instead)

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -443,6 +446,10 @@ export type MessageProps = {
    * Message to render. It can type `ChatMessage` or `SystemMessage` or `CustomMessage`.
    */
   message: ChatMessage | SystemMessage | CustomMessage;
+  /**
+   * Strings from locale.
+   */
+  strings: ILocaleKeys;
```
JamesBurnside, line 452: This type should be the strings needed for just MessageProps (probably a separate property per string needed so developers can use intellisense to know what strings this component needs)

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -56,57 +52,85 @@ const getDefaultComponents = (
...
+        ),
+        numOthers: <>`${countOfUsersNotMentioned}`</>
+      })}
+    </Stack>
+  );
 };
 
-const defaultTypingString = (typingUsers: CommunicationParticipant[]): string => {
-  return typingUsers.length > 0 ? (typingUsers.length > 1 ? ' are typing...' : ' is typing...') : '';
+const formatElements = (str: string, vars: IObjectMap<JSX.Element>): JSX.Element[] => {
```
JamesBurnside, line 94: What is this function doing? This needs renamed as `formatElements` is far too generic a name.

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -56,57 +52,85 @@ const getDefaultComponents = (
...
-
   const countOfUsersNotMentioned = typingUsers.length - typingUsersMentioned.length;
-  if (countOfUsersNotMentioned > 0) {
-    displayComponents.push(
-      <span className={mergeStyles(typingIndicatorVerbStyle, styles?.typingString)} key={'UsersCountKey'}>
-        {` and ${countOfUsersNotMentioned} other${countOfUsersNotMentioned === 1 ? '' : 's'}`}
-      </span>
-    );
+
+  let typingString = '';
```
JamesBurnside, line 67: The `getIndicatorComponents` function is really long, consider pulling this nice if statement out into a unit tested utility function

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -56,57 +52,85 @@ const getDefaultComponents = (
     const additionalCharCount =
```
JamesBurnside, line 52: Do all languages separate lists by commas?

packages/react-components/src/components/tests/MessageThread.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under the MIT license.
+
+import React from 'react';
+import { MessageThread } from '../MessageThread';
+import { ChatMessage } from '../../types';
+import Enzyme from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import { mountWithLocalization } from './enzymeUtils';
+import { act } from 'react-dom/test-utils';
+import germanStrings from '../../localization/translated/de.json';
```
JamesBurnside, line 11: I like the test! But this should probably be a pseudo language that is used exclusively for testing - this gives the flexibility to test unusual things like obtuse unicode characters

packages/react-components/src/components/tests/TypingIndicator.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+Enzyme.configure({ adapter: new Adapter() });
+
+test('TypingIndicator test in english', async () => {
+  const component = mountWithLocalization(
+    <TypingIndicator typingUsers={[{ userId: 'user2', displayName: 'Claire' }]} />,
+    englishStrings
+  );
+  await act(async () => component);
+  component.update();
+  expect(component.html()).toBe(
```
JamesBurnside, line 21: can we use toMatchSnapshot here? (or toMatchInlineSnapshot)? (might need an extra plugin to work with enzyme)

JamesBurnside: https://www.npmjs.com/package/enzyme-to-json

packages/react-components/src/components/tests/MessageThread.test.tsx
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
```
JamesBurnside, line 1: test file should be inline beside component instead of its own tests directory

JamesBurnside: (can check with the team what everyone prefers in case the directory gets really long. I prefer beside the src file as then tests are treated as a primary concern, always easy to find the file, and easy to see when a file has no tests (as there is no .test file beside it)

packages/react-components/src/components/utils/Datetime.ts
```diff
@@ -75,9 +75,31 @@ export const formatTimestampForChatMessage = (messageDate: Date, todayDate: Date
...
+    case 3:
+      return strings.wednesday;
+    case 4:
+      return strings.thursday;
+    case 5:
+      return strings.friday;
+    case 6:
+      return strings.saturday;
+    default:
+      return '';
```
JamesBurnside, line 103: nit: `throw` here if an invalid day is passed

packages/react-components/src/components/utils/Datetime.ts
```diff
@@ -75,9 +75,31 @@ export const formatTimestampForChatMessage = (messageDate: Date, todayDate: Date
...
-    return weekDayToDayName[messageDate.getDay()] + ' ' + formatTimeForChatMessage(messageDate);
+    const day = weekDayToDayName(messageDate.getDay(), strings);
+    return (day !== '' ? day : formatDateForChatMessage(messageDate)) + ' ' + formatTimeForChatMessage(messageDate);
   }
 
   // If message date is in previous or older weeks then timestamp string is like 2021-01-10 1:30 p.m.
   return formatDateForChatMessage(messageDate) + ' ' + formatTimeForChatMessage(messageDate);
 };
+
+const weekDayToDayName = (day: number, strings: ILocaleKeys): string => {
```
JamesBurnside, line 86: nit: just `dayToDayName`, weekDay looks like a typo of weekday implying just Mon-Fri

packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+        if (forceReload) {
+          window.location.reload();
+        } else if (loadLocaleStrings) {
+          const localeData = await loadLocaleStrings(loc.locale);
+          setStrings(localeData);
+          _setLocale(loc);
+
+          document.documentElement.lang = loc.locale;
+          document.documentElement.dir = loc.rtl ? 'rtl' : 'ltr';
```
JamesBurnside, line 90: RTl vs LTR is normally set at the HTML level of a page: https://www.w3schools.com/TAgs/att_dir.asp#:~:text=The%20dir%20attribute%20specifies%20the%20text%20direction%20of,and%20can%20be%20used%20on%20any%20HTML%20element.

Looks like fluent also grabs it from here: https://developer.microsoft.com/en-us/fluentui#/styles/web/localization

packages/react-components/src/components/CameraButton.tsx
mgamis-msft: I don't think it does because I am memoizing a function that depends only on the strings values but does not depend on the values of props.checked and props.styles.label since props is an argument of this memoized function.

packages/react-components/src/components/tests/TypingIndicator.test.tsx
mgamis-msft: Sure I could do that. I just have one question. I have a ParticipantItem test example in this PR that uses jest snapshot just to show another option. Changes not related to localization to a component would cause the need for updating snapshots. Are we ok with that?

packages/react-components/src/localization/LocalizationProvider.tsx
mgamis-msft: Right. I could be wrong but doesn't document.documentElement get the root element of a page which is <html> for an HTML page?

packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -0,0 +1,124 @@
...
+    async (locale: string, forceReload?: boolean) => {
+      const loc = locales[locale];
+
+      if (loc) {
+        if (storage) {
+          storage.setItem(LOCALE_CACHE_KEY, loc.locale);
+        }
+
+        if (forceReload) {
+          window.location.reload();
```
JamesBurnside, line 83: when is force reload used?

packages/react-components/src/localization/LocalizationProvider.tsx
JamesBurnside: Yep you're right, I was thinking the user would set `rtl` and `lang` in their html markup and we would pick it up from there instead of defining it per locale and overwriting whatever was set already

JamesBurnside: (happy if we should be setting it based on chosen locale, but just want to check)

packages/react-components/src/components/tests/TypingIndicator.test.tsx
JamesBurnside: Yep, ideally we should have snapshot tests for every component testing different props. I think a good goal would be unit test snapshots in the react-components folder, then integration tests in storybook instead of having snapshots in storybook

packages/react-components/src/components/tests/MessageThread.test.tsx
prprabhu-ms: +1 to side-by-side.

I've seen no benefit from splitting them. They're supposed to be tightly coupled with the components, so might as well keep together.

prprabhu-ms (commented):
Read through once to wrap my head around the approach. Will review in depth tomorrow :)

packages/communication-react/rollup.config.js
mgamis-msft: Not why I was getting error `[!] Error: You must set either output.file for a single-file build or output.dir when generating multiple chunks.` at first but reverted it back

packages/react-components/src/components/TypingIndicator.tsx
mgamis-msft: Good point. I think Arabic does not for 3 or less. This will take me more time in a separate ticket and instead of strings we may need to keep a callback in the locale file instead.

packages/react-components/src/localization/LocalizationProvider.tsx
mgamis-msft: Removed.

packages/react-components/review/react-components.api.md
```diff
@@ -145,6 +161,106 @@ export interface JumpToNewMessageButtonProps {
...
 
+// @public
+export const LocaleContext: React_2.Context<ILocale>;
+
+// @public (undocumented)
+export const locales: {
+    'en-US': {
+        locale: {
+            locale: string;
+            strings: {
```
alcail, line 172: nit: could we change this 'strings' name to something like 'componentStrings' instead. I'm just afraid of confusion with the type 'string'

mgamis-msft: That's good point.. Will do.

packages/communication-react/review/communication-react.api.md
```diff
@@ -952,7 +931,7 @@ export type MessageContentType = 'text' | 'html' | 'richtext/html' | 'unknown';
 // @public
 export type MessageProps = {
     message: ChatMessage | SystemMessage | CustomMessage;
-    strings: ILocaleKeys;
+    dateStrings: MessageThreadStrings;
```
prprabhu-ms, line 934: One of the conclusions from the meeting was to standardize on a `props.strings` field with the relevant fields here.

packages/react-components/src/components/utils/Datetime.ts
```diff
@@ -52,7 +52,11 @@ export const formatDateForChatMessage = (messageDate: Date): string => {
  * @param messageDate - date of message
  * @param currentDate - date used as offset to create the user friendly timestamp (e.g. to create 'Yesterday' instead of an absolute date)
  */
-export const formatTimestampForChatMessage = (messageDate: Date, todayDate: Date, strings: ILocaleKeys): string => {
+export const formatTimestampForChatMessage = (
```
prprabhu-ms, line 55: Not for this PR: DateTime handling across locales is complicated: https://docs.microsoft.com/en-us/dotnet/standard/globalization-localization/globalization#handle-dates-and-times

Let's file a separate work item for this?

packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -1,124 +1,47 @@
...
-  locale: ILocale;
-  locales: ILocaleCollection;
-  strings: ILocaleKeys;
-  setLocale: (locale: string, forceReload?: boolean) => void;
+/**
+ * Locale information to change strings in components
+ */
+export interface ILocale {
+  /** Set of strings for components */
+  localeStrings: ILocaleKeys;
```
prprabhu-ms, line 17: One of the recommendations from the meeting today was to avoid losing typing information in this interface.

So something like:

    export interface ILocate {
        messageThread: MessageThreadStrings,
        ...
    }

packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -1,124 +1,47 @@
...
 
-export interface ILocaleContext {
-  locale: ILocale;
-  locales: ILocaleCollection;
-  strings: ILocaleKeys;
-  setLocale: (locale: string, forceReload?: boolean) => void;
+/**
+ * Locale information to change strings in components
+ */
+export interface ILocale {
```
prprabhu-ms, line 15: We don't use the `I` prefix for interfaces: simply `Locale`

packages/react-components/src/components/utils/Datetime.ts
mgamis-msft: Sure. I filed this task https://skype.visualstudio.com/SPOOL/_workitems/edit/2484990.

packages/react-components/src/localization/LocalizationProvider.tsx
mgamis-msft: Now I see. Thanks for the example. Done.

packages/communication-react/review/communication-react.api.md
```diff
@@ -891,11 +893,21 @@ export interface JumpToNewMessageButtonProps {
 export const lightTheme: PartialTheme & CallingTheme;
 
 // @public
-export const LocaleContext: React_2.Context<ILocale>;
+export interface Locale {
```
prprabhu-ms, line 896: nitty nit: Do we intend to include stuff other than strings in Locale? If so, might be cleaner to bundle all strings inside a field, with a shorter nam for the fields!

```ts
export interface Locale {
    strings: ComponentStrings;
}

export interface ComponentStrings {
  cameraButton: CameraButtonStrings;
  endCallButton: EndCallButtonStrings;
  // ...
}

// ... usage
 const locale.strings.cameraButton.onText
```

This is a nitty nit, so choose based on taste ;)

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -16,6 +16,20 @@ export interface TypingIndicatorStylesProps extends BaseCustomStylesProps {
...
+ */
+export interface TypingIndicatorStrings {
+  /** String to use when one user is typing */
+  singular: string;
+  /** String to use when multiple users are typing */
+  plural: string;
+  /** String to use when multiple users are typing with one other user abbreviated */
+  shortenedPlural: string;
+  /** String to use when multiple users are typing with multiple other users abbreviated */
+  shortenedPlural2: string;
```
prprabhu-ms, line 30: Can we name this better? How about `singleUser`, `multipleUsers`, `multipleUsersAbbreviateOne`, `multipleUsersAbbreviateMany`?

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -16,6 +16,20 @@ export interface TypingIndicatorStylesProps extends BaseCustomStylesProps {
   typingString?: IStyle;
 }
 
+/**
+ * Strings of TypingIndicator that can be overridden
+ */
+export interface TypingIndicatorStrings {
+  /** String to use when one user is typing */
+  singular: string;
```
prprabhu-ms, line 24: It's not clear from the type definition how these strings will be used.

Could you provide an example string for each (in english) along with users that are typing, and show what the result will be. i.e., do we format the users with the string, concat it... ?

EDIT: I see from the tests that these strings have formatting placeholders. This is great. Please document those placeholders, preferably with example.

prprabhu-ms (approved):
This looks great now.

Thanks a lot for the design discussion and the (minor) iteration on the approach.

packages/react-components/src/components/EndCallButton.tsx
```diff
@@ -1,12 +1,23 @@
...
+import { useLocale } from '../localization';
+
+/**
+ * Strings of EndCallButton that can be overridden
+ */
+export interface EndCallButtonStrings {
+  /**
+   * Label of button
+   */
+  text: string;
```
JamesBurnside, line 19: nit: I think this should be `label` or `labelText` for the buttons. So users can make the connection that this text is related to the showLabel prop

packages/react-components/src/components/EndCallButton.tsx
```diff
@@ -56,13 +74,16 @@ export const EndCallButton = (props: EndCallButtonProps): JSX.Element => {
...
-  };
+  const defaultRenderText = useCallback(
+    (props?: IButtonProps): JSX.Element => {
+      return (
+        <Label key={'callEndLabelKey'} className={mergeStyles(controlButtonLabelStyles, props?.styles?.label)}>
+          {text}
+        </Label>
+      );
+    },
+    [text]
```
JamesBurnside, line 85: Should `props?.styles?.label` also be in this dep array?

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -53,6 +64,10 @@ export interface ParticipantItemProps {
    * ```
    */
   styles?: ParticipantItemStylesProps;
+  /**
+   * Optional strings to override in component
+   */
+  strings?: ParticipantItemStrings;
```
JamesBurnside, line 70: Should this also have `Partial<ParticipantItemStrings>` to match the other interfaces?

packages/react-components/src/components/ParticipantItem.tsx
JamesBurnside: Filed a bug for the useTheme (https://skype.visualstudio.com/SPOOL/_workitems/edit/2495786). Resolving this.

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -18,16 +16,104 @@ export interface TypingIndicatorStylesProps extends BaseCustomStylesProps {
   typingString?: IStyle;
 }
 
+/**
+ * Strings of TypingIndicator that can be overridden
+ */
+export interface TypingIndicatorStrings {
+  /**
+   * String template to use when one user is typing. Placeholders: [user].
+   */
```
JamesBurnside, line 25: Keep each of these as one large comment block and use `@example` https://tsdoc.org/pages/tags/example/

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -56,57 +147,79 @@ const getDefaultComponents = (
     const additionalCharCount =
       2 * (countOfUsersMentioned - 1) + (typingUser.displayName ? typingUser.displayName.length : 0);
     if (totalCharacterCount + additionalCharCount <= MAXIMUM_LENGTH_OF_TYPING_USERS || countOfUsersMentioned === 1) {
-      typingUsersMentioned.push(typingUser);
+      if (typingUser.displayName) {
+        typingUsersMentioned.push(typingUser);
+      }
       totalCharacterCount += additionalCharCount;
```
JamesBurnside, line 153: Should these also be inside the `if (typingUser.displayName) {`
Not really sure if countOfUsersMentioned includes a user with no display name

JamesBurnside: If we should not include users with no display name, might be simpler above this for loop to instead do:
```
const typingUsersWithNames = typingUsers.filter(user => !!user.displayName);
for (const typingUser of typingUsersWithNames) {
```

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -56,57 +147,79 @@ const getDefaultComponents = (
...
+    };
+  }
 
-const defaultTypingString = (typingUsers: CommunicationParticipant[]): string => {
-  return typingUsers.length > 0 ? (typingUsers.length > 1 ? ' are typing...' : ' is typing...') : '';
+  return (
+    <Stack horizontal className={mergeStyles(typingIndicatorStringStyle, styles?.typingString)} key="typingStringKey">
+      {formatElements(typingString, variables)}
+    </Stack>
+  );
```
JamesBurnside, line 203: nit: this function is ~70 lines long, considering breaking up into multiple functions

packages/react-components/src/components/styles/TypingIndicator.styles.ts
```diff
@@ -4,25 +4,16 @@
...
-  textOverflow: 'ellipsis',
-  paddingRight: '0.1875rem'
+  height: '2.125rem'
 });
 
-export const typingIndicatorVerbStyle = mergeStyles({
+export const typingIndicatorStringStyle = mergeStyles({
   fontWeight: 400,
-  whiteSpace: 'nowrap',
+  whiteSpace: 'pre',
```
JamesBurnside, line 12: When does this wrap vs use the ellipses? (or whats the purpose of `pre` here?)

packages/react-components/src/components/utils/testUtils.tsx
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT license.
+
+import React from 'react';
+import { mount, shallow, ShallowWrapper } from 'enzyme';
+import { LocalizationProvider, Locale } from '../../localization/LocalizationProvider';
+import en_US from '../../localization/translated/en-US.json';
+
+export const mountWithLocalization = (node: React.ReactElement, locale: Locale): ShallowWrapper => {
+  return mount(node, {
+    wrappingComponent: LocalizationProvider,
```
JamesBurnside, line 11: this is awesome!

packages/react-components/src/components/EndCallButton.tsx
mgamis-msft: Nope since props is a argument of the callback.
image

common/config/rollup/rollup.config.js
```diff
@@ -17,5 +18,5 @@ export default (packageJson) => ({
     format: 'cjs',
     sourcemap: true
   },
-  plugins: [commonjs(), sourcemaps(), svg()]
+  plugins: [commonjs(), sourcemaps(), svg(), json({ compact: true })]
```
JamesBurnside, line 21: Hmm worried this will bundle all strings for all languages into the js file, not sure how we would avoid this - I'll file a bug

JamesBurnside: https://skype.visualstudio.com/SPOOL/_workitems/edit/2495870

packages/react-components/src/components/TypingIndicator.tsx
mgamis-msft: I had originally tried using `@example` but I kept getting this ugly warnings.
image
Would it be okay if I change the rules in eslintrc.js regarding these violations?

packages/react-components/src/localization/localizationUtils.tsx
```diff
@@ -0,0 +1,66 @@
...
+  if (!str) {
+    return '';
+  }
+  if (!vars) {
+    return str;
+  }
+
+  // regex to search for the pattern "{}"
+  const placeholdersRegex = /{(\w+)}/g;
+  return str.replace(placeholdersRegex, (_: string, k: string) => vars[k] || `{${k}}`);
```
JamesBurnside, line 28: Where is this function being used? I'm worried that if a string has {name} in it already this will unintentionally replace chunks of the string it shouldn't

packages/react-components/src/components/EndCallButton.tsx
JamesBurnside: oh nice!

packages/react-components/src/components/TypingIndicator.tsx
JamesBurnside: Does having:
```
@example example desc
`` `
example code
`` `
```
work? (without the space before the last backtick, just put that in so this comment renders correctly)

packages/react-components/src/localization/localizationUtils.tsx
mgamis-msft: I have removed this util function since it is not used.

packages/react-components/src/components/styles/TypingIndicator.styles.ts
mgamis-msft: Removed the styling regarding wrapping. I changed it to `pre` so that whitespace in the span elements are not ignored.
`whiteSpace: nowrap`
image
vs.
`whiteSpace: pre`
image

packages/react-components/src/localization/localizationUtils.tsx
JamesBurnside: Same concern with `formatElements`, but given this is used sparingly, is an unlikely scenario to run into and the function itself is not publicly exposed (so we can make changes without them being breaking changes) this could be a fix for a later time if someone files a bug on it.

JamesBurnside:
Note: will need a follow up PR to update the documentation in storybook

packages/react-components/src/components/styles/TypingIndicator.styles.ts
JamesBurnside: Ah thanks!

JamesBurnside (approved):

packages/react-components/src/localization/locales.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import en_US from './translated/en-US.json';
```
JamesBurnside, line 4: @prprabhu-ms were there any concerns here about treeshaking? I remember having a disucssion but can't remember the outcome

JamesBurnside: I wonder if dynamic importing would help here https://2ality.com/2017/01/import-operator.html#loading-code-on-demand

JamesBurnside: (@mgamis-msft I wouldn't block the PR on this but we will need to solve this if a consumer can't treeshake the languages they don't want)

mgamis-msft: Thanks for the dynamic import suggestion. I was thinking something similar like async callback prop on `LocalizationProvider` that returns the component strings which can be assigned something like:
`const loadLocaleStrings= async (locale: string): Promise<ComponentStrings> => {
  switch (locale) {
    ...
    case "en-US":
    default:
      return (await import("translated/en-US.json")).default;
  }
};
`
but I will address it in bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2495870. Thanks.

PR #469, opened 2021-06-14
Replace `Map` with `Array` for `unparentedViews`

prprabhu-ms:
# What
Drop last use of `Map` in the stateful clients' state.
Using an `Array` for the `unparentedViews` doesn't have major performance implications because the total number of entries is limited by the number of video devices available locally and because the list is not iterated over unconditionally (there is no base selector on this list).

# Why
Play well with Redux.

# How Tested
* `rush build`
* `rush test`
* Local storybook, toggle preview on configuration screen.

# Process & policy checklist


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -52,7 +52,7 @@ export class CallContext {
         cameras: [],
         microphones: [],
         speakers: [],
-        unparentedViews: new Map<LocalVideoStreamState, LocalVideoStreamState>()
```
JamesBurnside, line 55: this used to be a map of an object.. to itself?

JamesBurnside: (where key and value the same here?)

JamesBurnside: never mind read the comment where this is defined

JamesBurnside (approved):

anjulgarg (approved):

packages/calling-stateful-client/src/CallContext.ts
prprabhu-ms: Yeah, this was a very weird map. I had to double and triple check that we didn't actually depend on the keys being referentially equal across some API calls.

PR #470, opened 2021-06-15
Alcail/remote video stream limitation

alcail:
# What
updated VideoGallery for remote participants to render placeholder instead of empty video when the video stream has no children due to calling SDK limitation and sort the tiles to have the placeholders after the video streams

# Why
task https://skype.visualstudio.com/SPOOL/_workitems/edit/2460609

# How Tested
ran calling sample with at least 8 participants and turned camera on/off randomly to check the behaviour.

Started with all cameras are on, then turned the cameras off/on with ending with camera all cameras on again with camera of user 'Aline2' being the last one.
image

# Process & policy checklist


JamesBurnside (approved):

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -225,12 +248,20 @@ const RemoteVideoTile = React.memo(
       };
     }, [onDisposeRemoteStreamView, userId]);
 
+    const renderVideoStreamElement = useMemo(() => {
```
prprabhu-ms, line 251: For readability: Please add a comment here explaining when `renderElement` is not well defined (i.e., about the resource limit) and also that returning `undefined` results in the avatar being shown.

prprabhu-ms (approved):

packages/react-components/src/components/VideoGallery.tsx
alcail: will do :)

anjulgarg (approved):

anjulgarg:
Very nicely done @alcail <3

PR #474, opened 2021-06-16
Add minimal sample to embed composites on a static html page

DominikMe:
# What
Adding a very basic sample to explain how to embed the CallComposite and ChatComposite on an HTML page with minimal effort.

# Why
The on-ramp for using ACS is quite high, also many developers are not familiar with React and don't want to adopt it. This sample is meant as an easy start to be copied. It solves bundling and abstracts away React. The remaining work to do for the developer is to write a service for managing users and access tokens. Maybe we can simplify this in the future with something pre-built too 🤔.

# How Tested
Built and run locally.

# Process & policy checklist


samples/StaticHtmlComposites/README.md
```diff
@@ -0,0 +1,38 @@
...
+- An existing Communication Services resource, or [create one](https://docs.microsoft.com/en-us/azure/communication-services/quickstarts/create-communication-resource).
+
+## Commands
+
+### Install dependencies
+```
+rush update
+```
+
+> Note: There is currently a bug in the local `@azure/communication-react` package. Run `npm install` instead to install the in-market package from npm.
```
DominikMe, line 22: I'll post a screenshot in Teams

alcail (approved):

samples/OneToOneCall/webpack.config.js
```diff
@@ -1,9 +0,0 @@
-// Copyright (c) Microsoft Corporation.
```
DominikMe, line 1: @prprabhu-ms I deleted this because it seemed like it was forgotten in the OneToOne deletion PR?

samples/StaticHtmlComposites/src/composites.js
```diff
@@ -0,0 +1,31 @@
...
+  const { containerId, userId, token, groupId, displayName } = args;
+  const adapter = await createAzureCommunicationCallAdapter(userId, token, { groupId }, displayName ?? 'anonymous');
+  ReactDOM.render(React.createElement(CallComposite, { adapter }, null), document.getElementById(containerId));
+  return adapter;
+};
+
+export const loadChatComposite = async function (args) {
+  const { containerId, userId, token, endpointUrl, threadId, displayName } = args;
+  const adapter = await createAzureCommunicationChatAdapter(
+    userId,
```
DominikMe, line 23: The order here is easy to get wrong. We should change the API to use objects if there are more than 2 arguments in a function. 

```typescript
createAzureCommunicationChatAdapter({
   userId: ...,
   token: ...,
   ...
 });

File: samples/StaticHtmlComposites/README.md [hunk 4]
```diff
@@ -0,0 +1,46 @@
+# Basic sample for using CallComposite and ChatComposite
```
JamesBurnside, line 1: nit: our samples will be switching to using only composites shortly, update this to specify using Call and Chat Composites in a plain html application

JamesBurnside:
Should this be a storybook example? Another option could be in a blog post/its own repo
Otherwise we'll need to add a pointer to this on the repo readme

samples/StaticHtmlComposites/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This sample is the bare minimum to embed Calling or Chat in your web app without needing to set up React. It consists of the following files:
```
JamesBurnside, line 3: This sample uses react but mentions here 'without needing to setup react'. Is this because our current quickstarts all use the CRA?

JamesBurnside:
This will need a CI build job added for this

samples/StaticHtmlComposites/README.md
JamesBurnside: Given this still uses the server I'm worried this gives of the wrong impression of the 'bare minimum'

DominikMe:
> This will need a CI build job added for this

Added.

samples/StaticHtmlComposites/README.md
DominikMe: Rephrased the intro to make it more clear that it still uses React but hides it and that this is intended for those who don't build React apps (but still want to use composites).

Also added yet another (third?) callout that the server portion is missing/fake.

samples/StaticHtmlComposites/README.md
DominikMe: Reworded.

DominikMe:
> Should this be a storybook example? Another option could be in a blog post/its own repo
I definitely think we should increase visibility to make the on-ramp for non-React devs easier. Are you suggesting to not add this sample and convert it to a storybook example or repo/quickstart instead? Making it easy to copy including the package.json and webpack.config.js is important imo but can be achieved in multiple ways. My preference is to check this in and then somebody else decides :-)

> Otherwise we'll need to add a pointer to this on the repo readme
Can you point me to the exact location where you want this mentioned? The root README and MDs in the docs folders mention samples in general, without explicitly mentioning individual ones. So in that sense, this new sample is covered.

DominikMe:
> > This will need a CI build job added for this
> 
> Added.

@JamesBurnside The build command for the sample succeeds fine locally, but not in CI :-( Help?
https://github.com/Azure/communication-ui-library/pull/474/checks?check_run_id=2859348171

DominikMe:
Partially addresses [User Story 2348666: As a developer, I can use this on my static webpage through using a javascript file](https://skype.visualstudio.com/SPOOL/_workitems/edit/2348666)

JamesBurnside (approved):

JamesBurnside:
> > Should this be a storybook example? Another option could be in a blog post/its own repo
> > I definitely think we should increase visibility to make the on-ramp for non-React devs easier. Are you suggesting to not add this sample and convert it to a storybook example or repo/quickstart instead? Making it easy to copy including the package.json and webpack.config.js is important imo but can be achieved in multiple ways. My preference is to check this in and then somebody else decides :-)
> 
> > Otherwise we'll need to add a pointer to this on the repo readme
> > Can you point me to the exact location where you want this mentioned? The root README and MDs in the docs folders mention samples in general, without explicitly mentioning individual ones. So in that sense, this new sample is covered.

Happy to check this in then have those discussions async-ly in team chats 👍

PR #476, opened 2021-06-16
Cache node_modules on CI builds

JamesBurnside:
# What
Improve build times by caching node_modules folder
* Cache key is a hash of the contents of the lock file
* Caches the pnpm store folder per suggestion on rush issue: https://github.com/microsoft/rushstack/issues/1847
* With GitHub actions we have 5GB of cache allowance (evicts oldest) so no issues there
  * Our cache size is ~177mb per cache. New cache created when lockfile is changed in a PR

# Why
Faster CI times
| before | after |
| -- | -- |
| `rush install` | `restore from cache + rush install` |
| 55s per action | 34s per action |

Only 7s to restore from cache! However, still takes ~26s for the `rush install` - let me know if you spot another cache point improvement from the build logs!

# How Tested
First CI run (no cache exists - creates a new cache): https://github.com/Azure/communication-ui-library/runs/2842984119
Subsequent CI run making use of the cache:
https://github.com/Azure/communication-ui-library/runs/2843025134


prprabhu-ms (approved):

prprabhu-ms:
You can shave off another 37 seconds if you also cache `common/temp/node_modules`.
Should be safe to cache with the same key? After all the node_modules are just decompressed result from the pnpm store.

wdyt?

prprabhu-ms:
37 seconds on my slow laptop ;) You can't shave that off the 34 second CI time. ymmv.

JamesBurnside:
> You can shave off another 37 seconds if you also cache `common/temp/node_modules`.
> Should be safe to cache with the same key? After all the node_modules are just decompressed result from the pnpm store.
> 
> wdyt?

I think we should do this 👍

common/config/rush/pnpm-lock.yaml
```diff
@@ -21201,3 +21201,4 @@ specifiers:
   '@rush-temp/react-composites': file:./projects/react-composites.tgz
   '@rush-temp/server': file:./projects/server.tgz
   '@rush-temp/storybook': file:./projects/storybook.tgz
+
```
prprabhu-ms, line 21204: Intentional? Or whitespace change to test your PR?

prprabhu-ms (approved):

common/config/rush/pnpm-lock.yaml
JamesBurnside: Yeah intentional, I was hoping it would cache bust but I guess a single blank new line in a 21000 line file didn't make enough of a difference 😅 (still got the old cache for some of the jobs)

alcail (approved):

common/config/rush/pnpm-lock.yaml
prprabhu-ms: It should cache bust. Perhaps github's build cache is distributed and the first job actually helped out the ones that started later? Or is it just a local cache on a persistent VM?

JamesBurnside: It only uploads the cache at the end of the job, however some jobs definitely grabbed from a cache when they shouldn't of given the change to this file.. don't have time to look into it right now but I'll leave this PR open until I do

JamesBurnside: Fixed the cache key, had `\`s instead of `/`s in the glob path 🤦‍♂️

JamesBurnside: > > You can shave off another 37 seconds if you also cache `common/temp/node_modules`.
> > Should be safe to cache with the same key? After all the node_modules are just decompressed result from the pnpm store.
> > wdyt?
> 
> I think we should do this 👍

Caching node_modules had no impact on the time for rush install, but did increase the restore from cache by ~6s so removing

PR #479, opened 2021-06-17
Fix type checking in sample apps

JamesBurnside:
# What
Run tsc as part of the build process
* Fix existing build issues

**NOTE** This only adds type checking to the prod build, not to the `npm run start` command. Open to adding it there too (it would probably use the webpack typescript checker then)

# Why
Ensure we have type checking for builds

## Perf impact
Sample CI build: 40s --> 45s (this looks to be because tsc is run in parallel to the webpack task and the webpack task takes slightly longer) 
Overall CI build: no impact - all sample build tasks are dwarfed by storybook build task

# How Tested
Build fails if there is a compiler error:
image


samples/Chat/src/app/ChatArea.tsx
```diff
@@ -31,12 +31,16 @@ export const ChatArea = (props: ChatAreaProps): JSX.Element => {
...
-      <MessageThread {...chatThreadProps} onRenderAvatar={props.onRenderAvatar} numberOfChatMessagesToReload={5} />
+      <MessageThread
+        {...(chatThreadProps as any)}
+        onRenderAvatar={props.onRenderAvatar}
+        numberOfChatMessagesToReload={5}
+      />
       <Stack.Item align="center" className={sendBoxParentStyle}>
         <div style={{ paddingLeft: '0.5rem', paddingRight: '0.5rem' }}>
-          <TypingIndicator {...typingIndicatorProps} />
+          <TypingIndicator {...(typingIndicatorProps as any)} />
```
JamesBurnside, line 41: given Nan has a change for updating the usePropsFor, and we are migrating to using the composites instead of this raw usage, I'm inclined to not debug into this..

JamesBurnside: however it is something that our end users might be hitting

JamesBurnside: I will file a bug instead

prprabhu-ms (approved):

alcail (approved):

PR #480, opened 2021-06-17
Wireframe localization API

prprabhu-ms:
# What
Exploring localization API options.

## Components

Allows increasing levels of localization control:

* Do nothing, get en-US
* Use `LocalizationProvider` with `locale`, get canned translations that we ship and support. But can't modify strings inside.
* Use `LocalizationProvider` with `strings`, provide strings as you see fit in one place for all components.
* Use `strings` props on individual components, full control (and cost).

## Composite

Allows increasing levels of localization control:

* Do nothing, get en-US
* Pass in `localization.locale` prop, get canned translations that we ship and support. But can't modify strings inside.
* Pass in `Localization.strings` prop, provide strings as you see fit in one place for all components.

### DIFF
--- a/packages/react-components/src/localization/localization.ts
+++ b/packages/react-components/src/localization/localization.ts
@@ -1,0 +1,33 @@
+import { CameraButtonStrings } from '../components/CameraButton';
+
+interface LocalizationProps {
+  /**
+   * Specific strings overides for components.
+   *
+   * Values here override those provided via the locale.
+   */
+  strings?: LocalizationStrings;
+
+  /**
+   * Locale for strings used by the components.
+   *
+   * Defaults to en-us (same as the default if no LocalizationProvider is used).
+   *
+   * For the list of supported locales, see ...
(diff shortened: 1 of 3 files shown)


packages/react-components/src/localization/localization.ts
```diff
@@ -0,0 +1,33 @@
+import { CameraButtonStrings } from '../components/CameraButton';
+
+interface LocalizationProps {
+  /**
+   * Specific strings overides for components.
+   *
+   * Values here override those provided via the locale.
+   */
+  strings?: LocalizationStrings;
```
JamesBurnside, line 9: Is this structure able to treeshake unused strings?

prprabhu-ms: Discussed IRL: Plan is to keep all strings / nested objects optional.

prprabhu-ms:
Just intended for demonstration.

PR #482, opened 2021-06-17
Document plans for error reporting

prprabhu-ms:
# Why
Agree on architecture, help plan out detailed work.

Pretty Markdown link
https://github.com/Azure/communication-ui-library/blob/error-handling-exploration/docs/architecture/ErrorReporting.md

# How Tested

Thought. Really. Hard.

I hope.


mgamis-msft (approved):

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,122 @@
...
+
+This document describes how errors are handled within the ACS UI library and how they are reported to the surrounding application.
+
+## Error sources
+
+There are two categories of errors, and they are handled differently.
+
+1. ACS backend errors: Errors returned by the REST calls to ACS backend services.
+
+   1. These can be transient or permanent (special case: permission error).
```
JamesBurnside, line 14: nit: The special case isn't quite returned by a REST call, this is a browser API call (e.g. no mic permission) - and something that can be resolved by the user at run time
Worthy of its own category? (as these will often by the only time the error will be presented to the web app user)

JamesBurnside, line 14: What is the permission error here? (if the likes of video permission is handled in category 2)

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,122 @@
...
+    errors: ErrorState[];
+}
+
+interface ErrorState {
+    /**
+     * Locally generated unique ID for error.
+     * Allows for de-duplicating similar errors from repeated API calls.
+     */
+    id: string;
+    sdkClientMethod: 'CallClient.createCallAgent' | 'CallClient.getDeviceManager'; // ... and many more.
```
alkwa-msft, line 39: do we define all of these in the stateful client projects?

docs/architecture/ErrorReporting.md
JamesBurnside: Ah I see there's an example below, can you add the example here too

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,122 @@
...
+
+The ACS UI library includes an error bar component and corresponding react component bindings to surface some of the errors from the state on the UI. Surfaced errors include:
+
+* For chat:
+  * Failure to connect to ACS backend ("you are offline").
+  * Permission error when connecting to ACS backend.
+  * Permission error for a particular thread (user not in the thread).
+  * Sending messages throttled by ACS backend (temporary error).
+* For calling:
+  * Permission error when connecting to ACS backend.
```
alkwa-msft, line 60: Calling has a huge set of codes/subcodes for why a call has ended. Are we thinking we need to account for all of these as well?

Silly question, do we need to supply "error messages" converting content from code/subcode into a string?

docs/architecture/ErrorReporting.md
prprabhu-ms: The scope here is limited to errors returned from backend REST calls, so video permission errors are covered by (2).

You raise an important point though: How do we report / surface errors in camera permissions? This document currently states that we don't.

* If we don't have permissions: We don't list devices in the device manager / gray out "start video" button etc.
* If we lose those permissions after the call has started (e.g., user went and removed those permissions at the browser level):
  * In the current implementation, the headless SDK handles the actual rendering. It might be that it catches and reports errors in rendering. If that's the case, the error will get handled as in (1) (even though I think that's conceptually wrong). Assuming that the headless SDK doesn't catch the errors...
  * my assumption is that we can't tell the difference between failure modes: "Camera disconnected", "permission revoked", "OS level driver failure" should all look the same at the browser API for security reasons.
  * My recommended approach is to surface the error in the `VideoTile` -- show an error screen for the video that failed to render.
  * Surfacing these errors back to Contoso is challenging: adapter is an API surface for the ACS backend but the video rendering is a purely UI operation.

Let's discuss this problem a bit more. I'd like to add this to the doc in some form.

docs/architecture/ErrorReporting.md
prprabhu-ms: Yes.

Undecided which is worse: Duplicating the list of methods in this literal type, or just allowing `string` `sdkClientMethod`.

Since the errors are reported and used asynchronous to the method that failed, we need _some_ way to know what method failed.

docs/architecture/ErrorReporting.md
prprabhu-ms: This section lists the errors we'll surface in the error bar. Since this is intended for the end-user, we'll want to keep the information limited -- clearly distinguish errors that are actionable (and tell them what they can do) from those that aren't, but don't give them gory details for unactionable errors.

We will need some mapping from error code/subcode to strings to report  to the end-user
* This mapping will be part of the bindings for the error bar component.
* The total number of target strings will be limited (so we may have to pick out specific (sub)codes for some errors, but default to a generic string for most.

Doest that make sense?

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,122 @@
...
+     */
+    id: string;
+    sdkClientMethod: 'CallClient.createCallAgent' | 'CallClient.getDeviceManager'; // ... and many more.
+    error: Error;
+}
+```
+
+Stateful client API method calls continue to return errors (e.g. as a failed promise). So they can be handled directly as well.
+
+The stateful clients stores the most recent errors, evicting older errors if necessary. The number of errors stored can be configured, or set to -1 to disable deletion of older errors. e.g., the composites disable error eviction in the stateful client and delete older errors themselves.
```
anjulgarg, line 46: Let's look at a scenario.
-> Contoso makes a call 
-> SDK throws Error 
-> Contoso is using `on('error', callback)` to handle the error
-> Contoso detects and handles the error fixing the cause of error
-> Contoso makes a call again, and it is successful this time.

Questions:
1. Do we remove the error from error state in this case?
2. If yes, how are you envisioning the mechanism to work?
3. If not, does that mean that Contoso can end up finding that error in the state and getting confused?

prprabhu-ms: Thanks a lot for bringing that up. You're absolutely right, we need to define how errors are cleared.

I think errors should be automatically cleared by the stateful client on success, as you mentioned. Additionally, stateful client should expose a way to clear errors (e.g. when the `ErrorBar` is dimissed).

Updated the doc, please take a look.

docs/architecture/ErrorReporting.md
prprabhu-ms: Update doc uses an object with fields keyed by the method names. Even more reason to use literal types now.

docs/architecture/ErrorReporting.md
JamesBurnside: My thinking was adapter isn't an ACS backend API surface, but just the business logic driving the UI (note: it can be totally detached from ACS and we can plug in any custom adapter).
The adapter currently handles the turning on and off of video (despite that not being a backend service driven feature) so I think its OK to surface this error in the adapter onError event.

My initial feeling is this should be a third category of error but not sure what that is as I don't want to be too specific (browser permissions error?)

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,140 @@
...
+    WARNING: This document describes future plans for error handling.
+    The UI library does not (yet) implement error handling as described here.
+
+This document describes how errors are handled within the ACS UI library and how they are reported to the surrounding application.
+
+## Error sources
+
+There are two categories of errors, and they are handled differently.
+
+1. ACS backend errors: Errors returned by the REST calls to ACS backend services.
```
JamesBurnside, line 12: I think this doc needs to be agnostic of ACS as the adapter interface isn't coupled to ACS

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,140 @@
...
+     * Allows for de-duplicating similar errors from repeated API calls.
+     */
+    id: string;
+    error: Error;
+}
+```
+
+Stateful client API method calls continue to return errors (e.g. as a failed promise). So they can be handled directly as well.
+
+The stateful clients stores the most recent errors per method, evicting older errors if necessary. The number of errors stored can be configured, or set to -1 to disable deletion of older errors. e.g., the composites disable error eviction in the stateful client and delete older errors themselves. By default, only the most recent error is stored.
```
JamesBurnside, line 47: What's an example of when a user uses the last n errors stored in our state instead of just using the onError event?

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,140 @@
...
+
+These errors are _unexpected_ in that the UI library has a bug or is unable to handle an error from the browser. Neither the end-user, nor the surrounding application, is expected to act on the errors. As such, they aren't handled via the mechanism used for ACS backend errors.
+
+The recommended way to handle these errors is to use React `ErrorBoundary`.
+
+    TODO: Add storybook example of using `ErrorBoundary`
+
+### Browser compatibility
+
+A special case of UI errors are errors due to browser incompatibility. The UI library does not determine browser compatibility at runtime. Any errors arising on unsupported browser (versions) are considered UI errors. The surrounding application can decide how to handle browser compatibility - e.g. by checking and warning end-users about browser versions or using polyfills where applicable. e.g. The samples included in this repository include a browser compatibility check.
```
JamesBurnside, line 138: Re the existing sample browser errors - are these the ones that check if mobile and if not firefox? If so I aren't those ACS specific errors so are a category 1 error instead of special case? What are other examples of this special case?

docs/architecture/ErrorReporting.md
prprabhu-ms: When using pure components, there is no `onError` event.

Similarly, the `ErrorBar` component will directly work on the stateful `errors` field and might want to show more than one error.

JamesBurnside: Pure components shouldn't give errors though (except for category 2 unexpected errors - but they're category 2 and "aren't handled via the mechanism used for ACS backend errors." and should use the react error boundary)

So does this state only exist to pipe errors into the ErrorBar component for composites? What if instead of an errorbar relying on state - we have specific state for error bar errors, as I'm sure most errors shouldn't be presented to an end user - only ones they can act upon

JamesBurnside: (My aim here is to avoid us having to expose and maintain an extra error interface in the state - that sounds difficult to use as errors a developer/user might care about might suddenly disappear because new ones evict them)

docs/architecture/ErrorReporting.md
prprabhu-ms: Thanks for bringing up that adapter isn't tied to ACS, and that it holds all state for the UI. This has real technical implications (e.g., the `ErrorBar` component must allow for a way to extend the error -> string mapping to include errors not from stateful).

This document must talk about ACS, because it applies to error reporting throughout the library, including from the stateful layer / components-plus-bindings use-case.

But I've reworked the composite section to include the case when the adapter isn't backed by ACS.

docs/architecture/ErrorReporting.md
prprabhu-ms: The sources of errors are still classifiable into two: from ACS backend / everything else.

I've reworded the composite section to treat this better. See if that helps.

docs/architecture/ErrorReporting.md
prprabhu-ms: Yeah this is about those browser checks.

Category 1 errors aren't ACS specific errors. They're errors returned by the ACS backends.

The errors here really stem from whatever limitations the JS headless SDKs + our UI library has. So these are still category 2 errors - we'll probably get js errors due to this incompatibility.

I reworded this to make it explicit that I'm only talking about the case when you're using the ACS SDKs.

docs/architecture/ErrorReporting.md
prprabhu-ms: By pure components I mean pure components + bindings (i.e., building without the adapters).
And we're still discussing category 1 errors here.

Two explicit use-cases so far:
- `ErrorBar`
- Disable `SendBox` when error indicates we can't send messages (e.g. when backend throttles us).

Think of this as doing for errors exactly what we're doing for state -- converting the response from API calls to data in a way that reactive UI can bind to the data.

As for error eviction, as stated, it is possible to configure the stateful client to never evict any errors and fully control when errors are dropped. But more likely than not, the default of keeping the last 1 error will suffice for most users, especially because this will save the last error per method.

JamesBurnside: Ah I was reading all this through the lens of composite-only. I really like having last n _per method_ - that makes a lot of sense and much easier to use (think I missed that in my last read through and it was just last n errors of any kind)

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,144 @@
...
+* Application clears the errors via a new Stateful client method.
+
+  ```ts
+  interface StatefulChatClient {
+      // ... Other methods
+      clearErrors(method: SdkClientMethod): void;
+  }
+  ```
+
+* If a previously failed method is retried and succeeds, the stateful client clears the errors on that (and sometimes related) method.
```
anjulgarg, line 64: I would recommend coming up with a POC in our existing library for this. 
Preferably using typescript decorators to make it less intimidating to maintain.

I have a gut feeling that this interception of each method for clearing error state based on the success of its response might end up becoming too complex.

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,144 @@
...
+
+    WARNING: This document describes future plans for error handling.
+    The UI library does not (yet) implement error handling as described here.
+
+This document describes how errors are handled within the ACS UI library and how they are reported to the surrounding application.
+
+## Error sources
+
+There are two categories of errors, and they are handled differently.
+
```
JamesBurnside, line 11: What if hypothetically we introduced a way to directly upload files to azure file storage for chat messages, would errors from that be an ACS backend error? (or a new category if we were to implement it?)

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,144 @@
...
+* `SendBox.sendMessageOnClick()` handler runs when the user clicks on a button, outside Reach component lifecycle.
+* The asynchronously returned error is dropped by the handler.
+* But the error is also persisted by the stateful client.
+  * The `ErrorBar` gets updated based on the state update.
+  * The `ChatComposite` also updates its own internal state and triggers an event that the application acts on.
+
+## UI errors
+
+These errors are _unexpected_ in that the UI library has a bug or is unable to handle an error from the browser. Neither the end-user, nor the surrounding application, is expected to act on the errors. As such, they aren't handled via the mechanism used for ACS backend errors.
```
JamesBurnside, line 134: You mention below this that the special case should be handled by the application but here that "Neither the end-user, nor the surrounding application, is expected to act on the errors.". I think if cat2 is a catch-all there's likely going to be some errors an application will want to act on. Especially if we label local device permissions as a category 2

docs/architecture/ErrorReporting.md
JamesBurnside: I'm still confused on the definitions sorry! This error isn't because the browser didn't implement the ACS SDK, the "unsupported browser" error is an error from the ACS SDK (even though under the hood it might be due to browser limitation) - so by definition would fall into category 1 as I get the error when calling the ACS SDK?

JamesBurnside: Then for the composite section it says "All errors are teed to an `errors` field". When referring to these errors, are they cat1 errors? (though they might not be from an ACS SDK?)

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,144 @@
...
+
+```ts
+interface CallClientState {
+    // ... other fields ...
+    errors: { [key: SdkClientMethod]: ErrorState[];
+}
+
+type SdkClientMethod = 'CallClient.createCallAgent' | 'CallClient.getDeviceManager'; // ... and many more
+
+interface ErrorState {
```
DominikMe, line 35: @prprabhu-ms Have we looked at using a type similar to the standard error object defined for Microsoft REST APIs?
https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#error--object

```typescript
interface Error {
    code: string,
    message: string,
    target?: string,
    innererror?: Error;
    details?: Error[];
}
```

`target` can then be the function name

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,144 @@
...
+
+But in both cases, all ACS backend API calls are mediated by the (chat and calling) stateful clients. Thus, the stateful client tees errors to a special field in the state:
+
+```ts
+interface CallClientState {
+    // ... other fields ...
+    errors: { [key: SdkClientMethod]: ErrorState[];
+}
+
+type SdkClientMethod = 'CallClient.createCallAgent' | 'CallClient.getDeviceManager'; // ... and many more
```
DominikMe, line 33: This type mapping should work. I tested that it omits the constructor and private properties.

```suggestion
type ErrorTarget = keyof CallClient | keyof DeviceManager | keyof ...;
```

[TypeScript playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAE2AYwDYEMrDm9BnPOAYXVVSNQEtgA7eAbwCg4W4wpKA3dGbAfWSlUAQQDmtGAG5mrdlx79BZAMox0yANbTWbDt15w+STpWTAAsuhrpxUbaxksA9ACoXjli+JYFcdMSEKajoAOg84LwABMABXACMqZHCosEx0AFscQKoJAHkwGEoIGjxkpw9kYrwYKBjkGGgACggCopKAfgAuPxoATwBKe2c3ZLgAFQALbBIyMQk4SkIYvGAEOAa4CasEVGwlVDwwnQjxiBwfAxgpgNnxOgAaNnxCfyIIdPSYmlMeNrGIDS0IhYJB0SikOAQOIAKxQ8HYEBMSDWADMoO84MoACIAaSOOi8AFpxtcZuQcnQsjRITRUL1NuhONhith1IUmTcRHd4JQSmoaGY-PB-IV0sB8awiXAAJoQGJU87AXz+GjAADunKC8151Ssgo2yAurLgqo1ZLmoVGxNl8sE1IQi0gK0hKL8mopMAA5IRkDEoFhKWyuNMhBaYI9tlkyOtSdlgl7SscvCxDUreObuY1+nAxVcIGsbOheeszqnlSb1ZywxLPHBovFEqNomlMg1ATRgasJODUHBiZNsG3aIrQYVSCE4ABVZ3CABefuAbw+Xx+hWK-3bndHPbgaIxkXQ86wTkqy++gjXNEJp-SxRLivLqe346bqSgGUhrSqfZJIdueS-Epd2gPwEAdS8IRaS9CCoQFEEdDB6WsMUawicodDLdNQ0zfpuk4CBKAQIY4Fcdwk1-OAsWAEwzEsaxbAWJYVjWDYthoHZsDFB1-GMUxgEIPA6gmPwfQyYB3zwR50lMdEwAmFlJJ6NY8DAJVASgQ5XwbUwyg8cQYComiLCsGxxKzPCCKIjxGAAX0YRgYF6VT3XjcxgDzBBCAAXjgQFeggV0yS1SkAB9fOAfzXXodBuhoGJ0jicS4Bs6RGEqPk4BAbogo9NyPO8uAACJMMXbCJEKyQgA)

docs/architecture/ErrorReporting.md
prprabhu-ms: Separated out components and composites entirely. PTAL.

docs/architecture/ErrorReporting.md
prprabhu-ms: Dropped the confusing classification of errors. Important point is errors that arise from the ACS clients are handled by the stateful client. Similarly, errors that arise behind the adapter API are expected to be handled by the adapters.

PTAL.

docs/architecture/ErrorReporting.md
prprabhu-ms: Dropped the confusing classification of errors. Important point is errors that arise from the ACS clients are handled by the stateful client. Similarly, errors that arise behind the adapter API are expected to be handled by the adapters.

PTAL.

docs/architecture/ErrorReporting.md
```diff
@@ -0,0 +1,146 @@
...
+    WARNING: This document describes future plans for error handling.
+    The UI library does not (yet) implement error handling as described here.
+
+This document describes how errors are handled within the ACS UI library and how they are reported to the surrounding application.
+
+Different error handling strategies are employed when using the composites or the components library. Let's look at each in turn.
+
+## Components library
+
+An application built using the components library uses the stateful clients to make REST calls to the Azure Communication Services backends. It uses the pure UI components and the component bindings package to  connect the UI components to the stateful clients.
```
JamesBurnside, line 12: nit: double space
```suggestion
An application built using the components library uses the stateful clients to make REST calls to the Azure Communication Services backends. It uses the pure UI components and the component bindings package to connect the UI components to the stateful clients.
```

JamesBurnside (approved):
This looks awesome! Thanks for the patience for all my questions!

docs/architecture/ErrorReporting.md
prprabhu-ms: Looked at this a bit and I think we'll have to stick to some very abstract type (perhaps any) for the error. Reasons:

* @azure/{communication-calling, communication-chat, communication-identity} do not return identical error types.
  * Chat returns the REST error (so mostly what you suggested).
  * Calling returns a different error type with code/subcode. Call transfer errors have their own shape.
  * Identity doesn't export any error types at all.
* We mostly get errors via rejected promises and typescript doesn't strongly type promise rejections: https://stackoverflow.com/questions/50071115/typescript-promise-rejection-type

So trying to strongly type the errors requires us to downcast and transform from `any` across a variety of types.

Will iterate on the exact type as I implement this for the stateful clients.

PR #483, opened 2021-06-17
Task/2437467 Mute icon with participant name in VideoTile

anjulgarg:
# What
Displays a mute indicator in `VideoTile`
Can be toggled and hidden through `VideoTile` and `VideoGallery` props
Fixing linting errors in components

# Why

# How Tested

# Process & policy checklist


packages/communication-react/review/communication-react.api.md
```diff
@@ -1127,9 +1127,9 @@ export type ParticipantsRemovedListener = (event: {
 
 // @public (undocumented)
 export interface PlaceholderProps {
-    displayName?: string;
-    noVideoAvailableAriaLabel?: string;
-    userId?: string;
+    displayName?: string | undefined;
```
prprabhu-ms, line 1130: I'm confused. `displayName?` already means it's optional, no?

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -51,6 +51,17 @@ export interface VideoGalleryProps {
...
   /** Callback to render a particpant avatar */
   onRenderAvatar?: (props: PlaceholderProps, defaultOnRender: (props: PlaceholderProps) => JSX.Element) => JSX.Element;
+
+  /**
+   * Whether to display the user's name in video tile.
+   * @defaultValue `true`
+   */
+  showDisplayName?: boolean;
+  /**
+   * Whether to display a mute icon beside the user's display name.
```
prprabhu-ms, line 61: What happens if `showDisplayName` is false but `showMuteIndicator` is true?

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -186,13 +209,13 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
         <Modal isOpen={true} isModeless={true} dragOptions={DRAG_OPTIONS} styles={floatingLocalVideoModalStyle}>
           {localParticipant && defaultOnRenderLocalVideoTile}
         </Modal>
-        <GridLayout styles={styles}>{defaultOnRenderRemoteParticipants}</GridLayout>
+        <GridLayout styles={styles ?? {}}>{defaultOnRenderRemoteParticipants}</GridLayout>
```
prprabhu-ms, line 212: Will this create spurious renders for `GridLayout` when `styles` is `undefined`? This creates a new object each time.

prprabhu-ms (approved):

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -129,18 +139,40 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+            >
+              {displayName}
+            </Text>
           )}
-        >
-          {displayName}
-        </Text>
-      )}
+        </Stack.Item>
+        <Stack.Item style={{ paddingLeft: '4px' }}>
```
alkwa-msft, line 165: do we want to move this style into a style file and also make sure we convert px to rem for font scaling

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -129,18 +139,40 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+          // when video is on, the displayName has a grey-ish background, so no use of theme
+          { color: isVideoReady ? palette.neutralPrimary : theme.palette.neutralPrimary },
+          styles?.displayNameContainer
+        )}
+      >
+        <Stack.Item>
+          {displayName && showDisplayName && (
+            <Text
+              style={{
+                fontSize: '12px',
```
alkwa-msft, line 156: px to rem as well here

packages/react-components/src/components/VideoGallery.tsx
alkwa-msft: Its optional in the SDK but I am starting to question if it should be optional in the UI components.  This would be a good question to  raise if there is a use case for having no name? 

Silly idea. Perhaps we use the acs id if the no display name is set? The only way for someone to go out of their way to have no name is to explicitly apply "" to the displayName prop?

packages/communication-react/review/communication-react.api.md
alkwa-msft: Its optional in the SDK but I am starting to question if it should be optional in the UI components. This would be a good question to raise if there is a use case for having no name?

Silly idea. Perhaps we use the acs id if the no display name is set? The only way for someone to go out of their way to have no name is to explicitly apply "" to the displayName prop?

packages/react-components/src/components/VideoTile.tsx
JamesBurnside: Careful with anything that does paddingLeft or paddingRight - will this work in RTL?

alkwa-msft: Thats a great point to bring up James!

https://rtlstyling.com/posts/rtl-styling/

.media[dir="rtl] img {
    float: right;
    margin-right: 0;
    margin-left: 16px;
}

We may need to address rtl specific areas where we are explicitly doing things on the left or the right.

packages/communication-react/review/communication-react.api.md
anjulgarg: Good point @prprabhu-ms 
Will update the code

@alkwa-msft Contoso might want a minimalistic UI and not want to display the name. I think it's better to have a choice than not have one at all.

anjulgarg: With that being said, I have removed the option from `VideoGallery` as it is already becoming a god object. Will think of a better way to compose a video gallery.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Removed showDisplayName option.
But if no display name is available, only a mute icon will be displayed.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: It should not. But I get nervous around such topics and will change it. :D

packages/react-components/src/components/VideoTile.tsx
anjulgarg: Fixed it through a more sensible styling.

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -129,18 +133,38 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
-      )}
+        </Stack.Item>
+        <Stack.Item>
+          {showMuteIndicator && isMuted !== undefined ? (
+            isMuted ? (
+              <MicOffIcon size="small" />
+            ) : (
+              <MicIcon size="small" />
+            )
+          ) : null}
```
alkwa-msft, line 164: cant we just say { a && b && <Component/> } we dont need to else return null?

alkwa-msft (approved):

packages/communication-react/review/communication-react.api.md
JamesBurnside: This comment is resolved but still shows `?: .. | undefined` - is this change pending a push?

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -186,13 +200,13 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
         <Modal isOpen={true} isModeless={true} dragOptions={DRAG_OPTIONS} styles={floatingLocalVideoModalStyle}>
           {localParticipant && defaultOnRenderLocalVideoTile}
         </Modal>
-        <GridLayout styles={styles}>{defaultOnRenderRemoteParticipants}</GridLayout>
+        <GridLayout styles={styles as any}>{defaultOnRenderRemoteParticipants}</GridLayout>
```
JamesBurnside, line 203: What's the reason as any needed added here? Can this be avoided?

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -257,7 +277,7 @@ const RemoteVideoTile = React.memo(
       }
 
       return <StreamMedia videoStreamElement={renderElement} />;
-    }, [renderElement, renderElement?.childElementCount]);
+    }, [renderElement]);
```
JamesBurnside, line 280: Why does the useMemo no longer need to update on childElementCount?

JamesBurnside (changes requested):
Small changes regarding the `?: .. | undefined` and use of `as any`

packages/react-components/src/components/VideoTile.tsx
anjulgarg: Sure.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: So the reason is that styles takes in a particular type which can not be undefined or null.
I can either assign styles a default `{}` or add an assignment here like
`styles={styles ?? {}}`
Based on a prev comment by @prprabhu-ms showing concern around this approach causing re-renders,  I decided to move away from default assignments and trick the typing system.

Let me know if you think of a better approach.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: We are watching `renderElement` and `renderElement?.childElementCount` becomes redundant

JamesBurnside: The useMemo (I'm assuming) does a shallow check so will only check the renderElement reference not the underlying properties (so childElementCount could change but this won't rerender).

packages/react-components/src/components/VideoGallery.tsx
JamesBurnside: We should always avoid any as we no longer will get compiler checks. Probably easiest option is above the method define `const emptyStyles = {}` then do `styles={styles ?? emptyStyles}`

packages/communication-react/review/communication-react.api.md
anjulgarg: Forgot to rebuild the package.

anjulgarg: Does this resolve your concern @JamesBurnside

JamesBurnside: Not quite, the question here is why have both `?` and ` | undefined`
Is it not sufficient to keep this as `displayName?: string;` and not have the `| undefined`

anjulgarg: @JamesBurnside I think you are referring to the `| undefined` check there.
I was getting type error without using undefined inside VideoTile.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Good advice, thanks

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Although, I tested it and it works. Still reverting it just to be safe.

JamesBurnside (approved):

packages/communication-react/review/communication-react.api.md
anjulgarg: Looks like I was getting weird type errors due to a faulty vs code.
I have reverted unnecessary type modifications.

PR #485, opened 2021-06-18
Update calling sample homepage in preparation for move to use composite

JamesBurnside:
# What
* Calling homepage now takes in the display name and/or teams link
* There is a join page that takes in display name
  * On using this - this isn't ideal and we should make the composite more flexible
* Update the window url to support a copyable teams link
  * Joining a teams meeting will now update the window.location.href to have `teamsLink=...` that is copy-pastable/sharable
* Some cleanup to move many things out of the app.tsx file
  * Use webpack variable to get the dependency version number instead of reaching in at runtime to the package.json
  * Move functions in app.tsx into utils file (reduce initial load on developers understanding the sample)
  * Move tracking the window size out of the app.tsx file (unnecessary in this PR as it will be totally removed once the logic is replaced with composites but didn't manually revert it for the same reason)

## Screenshots

| | Before | After |
| -- | -- | -- |
| Homepage | image | image |
| Join Page | image | image |
| Configuration Page | image | image |

# Why
* Match figma: https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/ACS-UI?node-id=2032%3A773
* Preparation for move to composite

# How Tested
Locally:
* Started a new call
* Joined an existing call
* Left a call and rejoined
* Joined a Teams meeting


common/config/webpack/sampleapp.webpack.config.js
```diff
@@ -51,9 +51,11 @@ const webpackConfig = (sampleAppDir, env) => ({
     new webpack.DefinePlugin({
       'process.env.PRODUCTION': env.production || !env.development,
       'process.env.NAME': JSON.stringify(require(path.resolve(sampleAppDir, 'package.json')).name),
-      'process.env.VERSION': JSON.stringify(require(path.resolve(sampleAppDir, 'package.json')).version)
+      'process.env.VERSION': JSON.stringify(require(path.resolve(sampleAppDir, 'package.json')).version),
+      __CALLINGVERSION__: JSON.stringify(require(path.resolve(sampleAppDir, 'package.json')).dependencies['@azure/communication-calling']),
+      __CHATVERSION__: JSON.stringify(require(path.resolve(sampleAppDir, 'package.json')).dependencies['@azure/communication-chat']),
```
alkwa-msft, line 56: just want to confirm this is a webpack config shared by all of the samples and thats why we have a mention of the chat version even tho we are only changing the calling sample in this change?

alkwa-msft (approved):

common/config/webpack/sampleapp.webpack.config.js
JamesBurnside: Yep its shared, so for calling this will just resolve to undefined (and __CHATVERSION__ isn't used anywhere in calling).
(You're right though its not necessary for this PR - just typed them both quickly)

PorterNan (approved):
Aha, I see the visual difference here, love the new design, makes a lot of sense

PR #486, opened 2021-06-18
Use composite in the calling sample

JamesBurnside:
# What
Replace all the sample functionality that is a duplicate of the composite functionality with the composite itself

# Why
We can't test both the samples and composites - we need one hot path for testing.
This will also make sure our composites are being flexible for real app usage.

# How Tested
* Started new Call
* Shared link and joined in another tab
* Left call - showed call ended screen
* Rejoined call
* Shows error screen if server failed to return user token
* Joined teams meeting
tbd:
* kicked from teams meeting


samples/Calling/src/app/App.tsx
```diff
@@ -22,207 +18,135 @@ import {
...
+    }
+    case 'call': {
+      if (userCredentialFetchError) {
         return (
           <CallError
-            title="Oops! You are no longer a participant of the call."
-            reason="Access to the meeting has been stopped"
-            rejoinHandler={() => setPage('createCallClient')}
+            title="Error getting user credentials from server"
+            reason="Ensure the sample server is running."
```
alkwa-msft, line 127: Does this also handle if you start up the server but you forgot to set your key?

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const CallScreen = (props: CallScreenProps): JSX.Element => {
+  const { token, userId, callLocator, displayName, onCallEnded, onCallError } = props;
+  const [adapter, setAdapter] = useState<CallAdapter>();
+  const { currentTheme } = useSwitchableFluentTheme();
+
+  useEffect(() => {
+    (async () => {
+      console.log('creating adapter');
```
alkwa-msft, line 28: just want to confirm we want to put this creating adapter log line here?

samples/Calling/src/app/App.tsx
JamesBurnside: Yeah - the client error message is always generic but the server will output a good error though when the token is requested

samples/Calling/src/app/views/CallScreen.tsx
JamesBurnside: nope nice catch, will remove

alkwa-msft (approved):
This looks awesome! Can't wait to try it out.

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -0,0 +1,58 @@
...
+      adapter.on('error', (e) => {
+        console.error(e);
+        onCallError(e);
+      });
+      setAdapter(adapter);
+      adapterRef.current = adapter;
+    })();
+
+    return () => {
+      adapterRef?.current?.dispose();
```
PorterNan, line 49: Is there reason to add adapterRef here? If that is just for unmount, add the adapter to dependency array should work

JamesBurnside: It was to ensure the dispose was called (otherwise I got a 'Call Agent already exists' error on rejoining a call)

PorterNan (approved):
Excellent, that solves a lot of current problems we have for now! Do we have 2-3 screenshot to show what will be the visual difference between the non-composite one and composite one?

samples/Calling/src/app/views/CallScreen.tsx
PorterNan: Yea, that is for unmount right? If you put adapter into dependency array and unmount directly using adapter.?dispose() that should work

PorterNan: If no adapter in dependency array, adapter in unmount function will be always 'undefined'

PorterNan: But you might need to add an if-else to adapter creation logic above

JamesBurnside: The adapter is set asyncly in the useEffect, so my thought was the return will capture the original adapter reference (i.e. undefined) in its closure. But let me test that

JamesBurnside: The adapter is set asyncly in the useEffect, so my thought was the return will capture the original adapter reference (i.e. undefined) in its closure.
If we put the adapter in the dep array that useEffect will be called twice - which I guess we can do a `if(!adapter) createAdapter()` but isn't this what useRef is for?

JamesBurnside:
> Excellent, that solves a lot of current problems we have for now! Do we have 2-3 screenshot to show what will be the visual difference between the non-composite one and composite one?

Here's some. The reason I didn't add to the PR description was I don't want these reviewed, whether we like the different looks or not will be filed as composite improvement bugs but not tackled in this PR.

| Before | After |
| -- | -- |
| image | image |

PR #491, opened 2021-06-19
Selected camera bug fix

mgamis-msft:
# What

SDK initializes cameras with one dummy camera:
`{ id: 'camera:id', name: '', deviceType: 'USBCamera' }`
So selectedCamera will have this value { id: 'camera:id', name: '', deviceType: 'USBCamera' } before new cameras are updated. So it makes sense that we only reset selectedCamera to the first camera if we don't see it in the new list of cameras.

Video where I also test adding a new camera and showing that selectedCamera is NOT set again because it is still on the list.
https://user-images.githubusercontent.com/79475487/122629625-49a07100-d073-11eb-928d-4b4778f20584.mp4

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2475044

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/calling-stateful-client/src/CallContext.ts
+++ b/packages/calling-stateful-client/src/CallContext.ts
@@ -570,7 +570,7 @@ export class CallContext {
   public setDeviceManagerCameras(cameras: VideoDeviceInfo[]): void {
     this.setState(
       produce(this._state, (draft: CallClientState) => {
-        if ((!draft.deviceManager.cameras || draft.deviceManager.cameras.length === 0) && cameras.length > 0) {
+        if (cameras.length > 0 && !cameras.some((camera) => camera.id === draft.deviceManager.selectedCamera?.id)) {
           draft.deviceManager.selectedCamera = cameras[0];
         }
         draft.deviceManager.cameras = cameras;
--- a/packages/calling-stateful-client/src/CallContext.ts
+++ b/packages/calling-stateful-client/src/CallContext.ts
@@ -570,7 +570,13 @@ export class CallContext {
   public setDeviceManagerCameras(cameras: VideoDeviceInfo[]): void {
     this.setState(
       produce(this._state, (draft: CallClientState) => {
-        if (cameras.length > 0 && !cameras.some((camera) => camera.id === draft.deviceManager.selectedCamera?.id)) {
+        if (
+          /** SDK initializes cameras with one dummy camera with value { id: 'camera:id', name: '', deviceType: 'USBCamera' } before camera permissions are granted.
+          So selectedCamera will have this value before new cameras are updated. We should reset selectedCamera to the first camera when there are cameras AND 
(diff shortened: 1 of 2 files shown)


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -570,7 +570,7 @@ export class CallContext {
   public setDeviceManagerCameras(cameras: VideoDeviceInfo[]): void {
     this.setState(
       produce(this._state, (draft: CallClientState) => {
-        if ((!draft.deviceManager.cameras || draft.deviceManager.cameras.length === 0) && cameras.length > 0) {
+        if (cameras.length > 0 && !cameras.some((camera) => camera.id === draft.deviceManager.selectedCamera?.id)) {
```
prprabhu-ms, line 573: Please add comment explaining why this is done in the code. Essentially, copy the PR description into the code comment.

The SDK behaviour is highly unusual, and will definitely trip me up the next time I read this :/

prprabhu-ms: Have we asked the SDK team why we get a dummy camera in the first place? That doesn't sound right.

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -570,7 +570,13 @@ export class CallContext {
   public setDeviceManagerCameras(cameras: VideoDeviceInfo[]): void {
     this.setState(
       produce(this._state, (draft: CallClientState) => {
-        if ((!draft.deviceManager.cameras || draft.deviceManager.cameras.length === 0) && cameras.length > 0) {
+        if (
+          /** SDK initializes cameras with one dummy camera with value { id: 'camera:id', name: '', deviceType: 'USBCamera' } before camera permissions are granted.
```
alkwa-msft, line 574: can we keep this comment above the if block so its all together?

alkwa-msft (approved):

packages/calling-stateful-client/src/CallContext.ts
mgamis-msft: Sure.

packages/calling-stateful-client/src/CallContext.ts
mgamis-msft: Sure added a comment and I just let them know about this potential bug here https://teams.microsoft.com/l/message/19:ea8322f638f34acdb74162cacd55c92b@thread.skype/1624308079847?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e9c1fc3-39df-4486-a26a-456d80e80f82&parentMessageId=1624308079847&teamName=Azure%20Communication%20Services&channelName=Calling&createdTime=1624308079847

prprabhu-ms (approved):

JamesBurnside (approved):

PR #492, opened 2021-06-21
POC - moving fluent to peer deps

JamesBurnside:
# What
* Move fluent to peer dependency of npm package
* Update the alpha package release task to publish this PR if a customer wants to try this out
  * NOT FOR CHECKING IN - JUST FOR RUNNING THE RELEASE OF THIS BRANCH

# Why
* Customers who don't want to install fluent but just use the stateful are forced to (and fluent deps are BIG)

# Why not
* Makes the install line for most consumers:
    ```
    npm i @azure/communication-react @fluentui/react @fluentui/react-icons@^0.3.9 @fluentui/react-northstar@^0.51.2 @fluentui/react-icons-northstar^0.51.2
    ```
  Instead of
    ```
    npm i @azure/communication-react
    ```

# How Tested


JamesBurnside:
This can be reopened if the need arises

PR #493, opened 2021-06-22
upgrading to latest calling package

alkwa-msft:
# What
Upgrading to the latest calling sdk version 1.1.0-beta.2

# Why
We should be on the the latest for the latest features. The 1.1.0 update specifically has a fix for chrome deprecation of an important set of features used in calling today.

# How Tested
Ran it locally in both a group call and teams interop scenarios


JamesBurnside (commented):
Does storybook also need this updated?

mgamis-msft (approved):

JamesBurnside (approved):

PR #495, opened 2021-06-23
[POC - DO NOT REVIEW] Chat button in call control (generic function signature)

JamesBurnside:
Do not review - this is being used for prototyping and evaluation

Result:
image

Notes:
* **Cannot enforce consistent styling and behavior**. With this fn signature there is no way to enforce the same styling and behavior if a dev adds a new button and wants it to be consistent with the other buttons.
  The 3rd party dev would have to look at our source code to ensure behavior and styling matches (e.g. we do not show label when the screen is smaller)
* Can only add another button to the center row of buttons - cannot add buttons to a left or right hand section.


JamesBurnside:
Closing this as it doesnt need to be open to use for evaluation  and clogs up the pull request screen

PR #498, opened 2021-06-24
[POC] Chat button in call control (restricted function signature)

JamesBurnside:
Result:
image

Also see previous PoC with open function signature: #495 

Notes
* Developers are forced into rendering a controlbarbutton allowing us to keep styling and behavior of the button consistent with the other buttons
* Override fn takes an array of either defaultButtonNames (for the existing buttons we provider) or customButtons
* override function can just be a prop instead of a function (small implementation detail, ignore reviewing that)

### DIFF
--- a/packages/communication-react/review/communication-react.api.md
+++ b/packages/communication-react/review/communication-react.api.md
@@ -325,7 +325,14 @@ export type CallCompositeProps = {
     fluentTheme?: PartialTheme | Theme;
     callInvitationURL?: string;
     onRenderAvatar?: (props: PlaceholderProps, defaultOnRender: (props: PlaceholderProps) => JSX.Element) => JSX.Element;
+    overrideCallControlButtons?: (defaultButtons: CallControlButtonCollection) => CallControlButtonCollection;
 };
+
+// @public (undocumented)
+export type CallControlButton = DefaultCallControlButton | CustomCallControlButton;
+
+// @public (undocumented)
+export type CallControlButtonCollection = CallControlButton[];
 
 // @public (undocumented)
 export type CallEndedListener = (event: {
@@ -627,6 +634,18 @@ export type CommunicationParticipant = {
 // @public
 export const ControlBar: (props: ControlBarProps) => JSX.Element;
 
+// @public
+export const ControlBarButton: (props: ControlBarButtonProps) => JSX.Element;
+
+// @public
+export interface ControlBarButtonProps extends IButtonProps {
+    icon?: JSX.Element;
+    labelText?: string;
+    showLabel?: boolean;
+    toggledIcon?: JSX.Element;
+    toggledLabelText?: string;
+}
+
 // @public (undocumented)
 export type ControlBarLayoutType = 'horizontal' | 'vertical' | 'dockedTop' | 'dockedBottom' | 'dockedLeft' | 'dockedRight' | 'floatingTop' | 'floatingBottom' | 'floatingLeft' | 'floatingRight';
 
@@ -673,6 +692,12 @@ export const createStatefulCallClient: (args: StatefulCallClientArgs, options?: 
 export const createStatefulChatClient: (args: StatefulChatClientArgs, options?: StatefulChatClientOptions | undefined) => StatefulChatClient;
 
 // @public (undocumented)
+export interface CustomCallControlButton extends ControlBarButtonProps {
+    // (undocumented)
+    newProp: string;
+}
+
+// @public (undocumented)
 export type CustomMessage = Message<'custom'>;
 
 // @public (undocumented)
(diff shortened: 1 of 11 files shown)


packages/react-composites/review/react-composites.api.md
```diff
@@ -244,8 +245,15 @@ export type CallCompositeProps = {
     fluentTheme?: PartialTheme | Theme;
     callInvitationURL?: string;
     onRenderAvatar?: (props: PlaceholderProps, defaultOnRender: (props: PlaceholderProps) => JSX.Element) => JSX.Element;
+    overrideCallControlButtons?: (defaultButtons: CallControlButtonCollection) => CallControlButtonCollection;
 };
 
+// @public (undocumented)
+export type CallControlButton = DefaultCallControlButton | CustomCallControlButton;
```
JamesBurnside, line 252: We could extend this further to allow rendering anything in here with 
```suggestion
export type CallControlButton = DefaultCallControlButton | CustomCallControlButton | JSX.Element
```

packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Button label text
+   */
+  toggledLabelText?: string;
+
+  /**
+   * Button icon in the non-toggled state
+   */
+  icon?: JSX.Element;
```
JamesBurnside, line 31: nit: this should be: `(props: IButtonProps) => JSX.Element`

JamesBurnside: same with labelText

JamesBurnside: label text should probably just be label also, but small nits not for this PR

packages/communication-react/review/communication-react.api.md
```diff
@@ -684,6 +709,9 @@ export type CustomMessagePayload = {
 // @public
 export const darkTheme: PartialTheme & CallingTheme;
 
+// @public (undocumented)
+export type DefaultCallControlButton = 'microphone' | 'camera' | 'screenShare' | 'participants' | 'leave';
```
prprabhu-ms, line 713: nit: Name this `DefaultCallControlButtons` or `DefaultCallControlButtonTypes`, or anything to distinguish from `CustomCallControlButton`. One is a string literal type, while the other is an interface.

packages/communication-react/review/communication-react.api.md
```diff
@@ -325,8 +325,15 @@ export type CallCompositeProps = {
     fluentTheme?: PartialTheme | Theme;
     callInvitationURL?: string;
     onRenderAvatar?: (props: PlaceholderProps, defaultOnRender: (props: PlaceholderProps) => JSX.Element) => JSX.Element;
+    overrideCallControlButtons?: (defaultButtons: CallControlButtonCollection) => CallControlButtonCollection;
```
prprabhu-ms, line 328: Mixing string literal types with actual objects in a single union type makes me squirm a little.

Technically, Contoso could ignore all the `defaultButtons` provided here, and build out all the button themselves from the button components. So we don't gain any safety by hiding the constructed buttons (and only providing names to this callback).

How about making `CallControlButtonCollection` only contain actual buttons. And the callback gets the fully constructed `defaultButtons` that Contoso is expected to pass back to us along with their own buttons?

packages/communication-react/review/communication-react.api.md
```diff
@@ -627,6 +634,18 @@ export type CommunicationParticipant = {
 // @public
 export const ControlBar: (props: ControlBarProps) => JSX.Element;
 
+// @public
+export const ControlBarButton: (props: ControlBarButtonProps) => JSX.Element;
+
+// @public
+export interface ControlBarButtonProps extends IButtonProps {
```
prprabhu-ms, line 641: I like this idea a lot.

packages/communication-react/review/communication-react.api.md
```diff
@@ -714,6 +742,9 @@ export type DefaultChatHandlers = {
     onLoadPreviousChatMessages: (messagesToLoad: number) => Promise<boolean>;
 };
 
+// @public (undocumented)
+export const defaultControlButtonStyle: IButtonStyles;
```
prprabhu-ms, line 746: nit: This should be a function that returns a freshly cloned object, so that Contoso can't (unintentionally) modify the value of the constant for all buttons.

packages/react-composites/review/react-composites.api.md
prprabhu-ms: Naah, I'd recommend against opening that door.

prprabhu-ms:
To a comment in OP: Override function needs to be a function (vs a prop) so that Contoso doesn't have to guess what the default order of buttons is. My suggestion in the PR takes this one step further -- I want us to construct and provide the buttons to the callback as argument that Contoso then passes us back.

JamesBurnside:
Closing this as it won't be the direction we will go in for the meeting composite. We do provide a default call control button now though #585

PR #499, opened 2021-06-24
Alcail/calling composite limitation

anjulgarg:
# What
Limits the number of remote video streams to 1 in Calling Composite

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -39,6 +39,8 @@ export interface VideoGalleryProps {
   localVideoViewOption?: VideoStreamOptions;
   /** Remote videos view options */
   remoteVideoViewOption?: VideoStreamOptions;
+  /** Limitation on number of remote video stream to display */
+  remoteVideoStreamLimitation?: number;
```
prprabhu-ms, line 43: `remoteVideoStreamLimitation` feels overly generic (limitation of what sort? Width? Count?)

How about `maxRemoteVideoStreams` or `maxRemoteVideoStreamCount`?

packages/react-composites/src/composites/CallComposite/MediaGallery.tsx
```diff
@@ -52,12 +52,22 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
+        remoteVideoStreamLimitation={1}
         styles={VideoGalleryStyles}
         layout="floatingLocalVideo"
         onRenderAvatar={props.onRenderAvatar}
       />
     );
-  }, [props.onRenderAvatar, videoGalleryProps]);
+  }, [
+    props.onRenderAvatar,
+    videoGalleryProps.localParticipant,
```
prprabhu-ms, line 63: +1

prprabhu-ms (approved):

JamesBurnside:
Why do we have to limit the remote streams to 1?

prprabhu-ms:
> Why do we have to limit the remote streams to 1?

I think this is temporary.

packages/communication-react/review/communication-react.api.md
```diff
@@ -1342,6 +1342,7 @@ export interface VideoGalleryProps {
     onRenderLocalVideoTile?: (localParticipant: VideoGalleryLocalParticipant) => JSX.Element;
     onRenderRemoteVideoTile?: (remoteParticipant: VideoGalleryRemoteParticipant) => JSX.Element;
     remoteParticipants?: VideoGalleryRemoteParticipant[];
+    remoteVideoStreamLimitation?: number;
```
JamesBurnside, line 1345: Open question, should this be exposed to the composite interface and set by the sample itself?

JamesBurnside: Actually, lets say no and be customer driven. Never mind.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Haha we think alike @prprabhu-ms 
My money is also on `maxRemoteVideoStreams`

Since all this work was done by @alcail I didn't wanna make changes unless anyone suggested them ;)

Thanks for reviewing,

JamesBurnside:
> > Why do we have to limit the remote streams to 1?
> 
> I think this is temporary.

What's the reason though?

anjulgarg:
> > > Why do we have to limit the remote streams to 1?
> > 
> > 
> > I think this is temporary.
> 
> What's the reason though? (just because I've found the stream quality for multiple streams massively improved lately due to acs backend improvements)

CC: @alkwa-msft

alkwa-msft:
> > > > Why do we have to limit the remote streams to 1?
> > > 
> > > 
> > > I think this is temporary.
> > 
> > 
> > What's the reason though? (just because I've found the stream quality for multiple streams massively improved lately due to acs backend improvements)
> 
> CC: @alkwa-msft

This is a limitation on the bandwidth restrictions of webrtc and how we are optimizing the stream. Today we use the desktop browser but consider the mobile browsers as well. In the backlog we have a user story to increase the number of remote streams to 4 for all browsers except safari. I think its important to do that work stream as a separate dev item since it involves adding browser checks as well.

JamesBurnside:
> This is a limitation on the bandwidth restrictions of webrtc and how we are optimizing the stream. Today we use the desktop browser but consider the mobile browsers as well. In the backlog we have a user story to increase the number of remote streams to 4 for all browsers except safari. I think its important to do that work stream as a separate dev item since it involves adding browser checks as well.

This is a really bad experience if we provide a group call that only supports one video stream. On mobile today is it just a slow video stream? Whats the long term plan to be able to support more streams?

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -176,8 +179,10 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
             onCreateRemoteStreamView={onCreateRemoteStreamView}
             onDisposeRemoteStreamView={onDisposeRemoteStreamView}
             isAvailable={remoteVideoStream?.isAvailable}
+            renderElement={
+              !maxRemoteVideoStreams || index < maxRemoteVideoStreams ? remoteVideoStream?.renderElement : undefined
```
JamesBurnside, line 183: From Alex's comment the reason this is being implemented is due to bandwidth and webrtc issues, since this change is purely UI being changing I want to confirm this change does reduce the bandwidth/webrtc problems (do we only consume the bandwidth if we use the render element or is stateful still creating it?)

anjulgarg: Good point. From what I remember, we only consume bandwidth when then renderElement is actually rendered in the DOM. Can someone confirm this?
@alkwa-msft @prprabhu-ms

alkwa-msft: when we startRender that is when we start using more of the bandwidth

JamesBurnside: From your comment alex sounds like we also need to ensure onCreateRemoteStreamView isn't being called (which, with only very lightly skimming this file, looks like its being called
image
)

packages/react-composites/src/composites/CallComposite/MediaGallery.tsx
```diff
@@ -52,12 +52,22 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
         {...videoGalleryProps}
         localVideoViewOption={localVideoViewOption}
         remoteVideoViewOption={remoteVideoViewOption}
+        maxRemoteVideoStreams={1}
```
alkwa-msft, line 55: can we please make this a constant and pump it through?

packages/react-components/src/components/VideoGallery.tsx
alkwa-msft: yes and since we're using the condition of the maxRemoteStreams as part of the renderElement boolean I think we're good here.

alkwa-msft (approved):

PR #501, opened 2021-06-25
Fix undefined selected camera

### LINKED ISSUE #497: Camera preview doesn't work in latest CallComposite
**Describe the bug; what happened?**

Updating @azure/communication-react from `1.0.0-alpha-20210623.9.0` to `1.0.0-alpha-20210624.13.0` broke the camera preview functionality for me, which is perhaps caused by https://github.com/Azure/communication-ui-library/pull/491.

**What are the steps to reproduce the issue?**
1. Use the calling composite sample on main (or use 1.0.0-alpha-20210624.13.0)
2. Start or join a call
3. Check if Camera field is persisted with your first choice
4. Try to press 'Turn on'
5. (won't work)
6.  Select another camera in the list
7. Switch back to your original camera
8. Press turn on (works!)

**What behavior did you expect?**
Camera will load on first try, without having to reselect.

**If applicable, provide screenshots:**

https://user-images.githubusercontent.com/1424596/123241274-4a0c9200-d4e1-11eb-8315-715d5e62563f.mp4

**In what environment did you see the issue?**

- OS & Device: Windows 10, Surface Laptop 3
- Browser: Edge Chromium
- Version: Version 91.0.864.54 (Official build) (64-bit)

**Is there any additional information?**
Happy to share my public repro site via Teams internally if that helps.

mgamis-msft:
# What

Assigning the selected camera on the same draft.deviceManager instead of calling this.setDeviceManagerSelectedCamera seems to fix the issue.

# Why
Fixes #497 

To fix bug where cameras were not being in the OptionsButton during a call.
https://user-images.githubusercontent.com/79475487/123460174-16db0900-d59c-11eb-8fc6-d5db6b35ff85.mp4
because selectedCamera is undefined in this [line of code](https://github.com/Azure/communication-ui-library/blob/main/packages/react-components/src/components/OptionsButton.tsx#L99)
After this one-line fix
https://user-images.githubusercontent.com/79475487/123460248-33774100-d59c-11eb-879a-cd3ab48f8482.mp4

# How Tested

Manually on sample app

# Process & policy checklist

### DIFF
--- a/packages/calling-stateful-client/src/CallContext.ts
+++ b/packages/calling-stateful-client/src/CallContext.ts
@@ -574,7 +574,7 @@ export class CallContext {
          * camera permissions are granted. So selectedCamera will have this value before the actual cameras are obtained. Therefore we should reset
          * selectedCamera to the first camera when there are cameras AND when current selectedCamera does not exist in the new array of cameras **/
         if (cameras.length > 0 && !cameras.some((camera) => camera.id === draft.deviceManager.selectedCamera?.id)) {
-          this.setDeviceManagerSelectedCamera(cameras[0]);
+          draft.deviceManager.selectedCamera = cameras[0];
         }
         draft.deviceManager.cameras = cameras;
       })
--- a/change/calling-stateful-client-95dd015d-8ef1-46c3-b56b-e99bad22cb27.json
+++ b/change/calling-stateful-client-95dd015d-8ef1-46c3-b56b-e99bad22cb27.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Fix stateful client selected camera assignment.",
+  "packageName": "calling-stateful-client",
+  "email": "miguelgamis@microsoft.com",
+  "dependentChangeType": "patch"
+}


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -574,7 +574,7 @@ export class CallContext {
          * camera permissions are granted. So selectedCamera will have this value before the actual cameras are obtained. Therefore we should reset
          * selectedCamera to the first camera when there are cameras AND when current selectedCamera does not exist in the new array of cameras **/
         if (cameras.length > 0 && !cameras.some((camera) => camera.id === draft.deviceManager.selectedCamera?.id)) {
-          this.setDeviceManagerSelectedCamera(cameras[0]);
+          draft.deviceManager.selectedCamera = cameras[0];
```
prprabhu-ms, line 577: Oops. Good catch.

Makes sense that if `immer::produce()` is not reentrant. `this._state` is overwritten with the draft here at the end of the block, so any updates by `setDeviceManagerSelectedCamera` will be lost.

prprabhu-ms (approved):

JamesBurnside (approved):

PR #503, opened 2021-06-25
Jinan/reorganize js bundle

PorterNan:
# What

Update sample to enable user grab the js file directly without any modification
Update readme
# Why

# How Tested

# Process & policy checklist


samples/StaticHtmlComposites/package.json
```diff
@@ -4,9 +4,9 @@
   "description": "Basic bundle for communication react composites and usage on a static HTML page.",
   "private": true,
   "scripts": {
-    "build": "webpack",
+    "build": "webpack --config webpack.build.config.js",
```
JamesBurnside, line 7: Instead of whole file for setting mode production, can just use `--mode=production` CLI arg instead here

JamesBurnside (approved):

samples/StaticHtmlComposites/index.html
```diff
@@ -4,26 +4,31 @@
   <title>Basic example for CallComposite and ChatComposite</title>
 </head>
 <body>
-  <div id="video-call"></div>
-  <div id="chat"></div>
-  <script src="./dist/bundle.js"></script>
+  <!-- height need to be set for composite to fit the layout -->
+  <div id="video-call" style="height: 500px;"></div>
```
alkwa-msft, line 8: prefer to use vh or rem over pixel

samples/StaticHtmlComposites/index.html
```diff
@@ -4,26 +4,31 @@
...
   <script type="module">
-    const { user, token } = await app.createUserWithToken();
+    const { user, token } = await service.createUserWithToken();
+    // Call composite sample code starts here
     const displayName = 'Bob';
 
-    const callAdapter = await app.loadCallComposite({
+    const callAdapter = await callComposite.loadCallComposite({
       containerId: 'video-call',
       groupId: '6e606975-7004-4c03-be37-e0851ed1850b', // any GUID
```
alkwa-msft, line 20: do we think we might want to bring in a dependency which randomly generates a guid? It could be a privacy concern if someone uses this and then sees other random people in the call

alkwa-msft (approved):

samples/StaticHtmlComposites/index.html
PorterNan: Good catch! Haha, I can imagine the scenes and it feels scary

PR #504, opened 2021-06-25
2482193 chat composite puppeteer

anjulgarg:
# What
Setting up the base for writing browser based tests for Chat Composite.

# Why

# How Tested

# Process & policy checklist


samples/Calling/package.json
```diff
@@ -21,7 +21,6 @@
     "@azure/communication-identity": "1.0.0",
     "@azure/communication-calling": "1.1.0-beta.2",
     "@azure/communication-common": "1.0.0",
-    "@azure/communication-react": "1.0.0-beta.1",
```
JamesBurnside, line 24: should this have been removed? (Ideally the sample _only_ references this package)

packages/react-composites/tests/browser/app/public/index.html
```diff
@@ -0,0 +1,16 @@
+<!-- Copyright (c) Microsoft Corporation. -->
+<!-- Licensed under the MIT license. -->
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="description" content="Chat Composite - Test App" />
+    <title>Chat Composite - Test App</title>
```
JamesBurnside, line 10: nit: avoid `chat` references here as this looks generic and can be used for any integration testing

packages/react-composites/tests/browser/app/index.tsx
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import React, { useState, useEffect } from 'react';
+import ReactDOM from 'react-dom';
+
+import { ChatAdapter, createAzureCommunicationChatAdapter, ChatComposite } from '../../../src';
```
JamesBurnside, line 7: something in the path or filename should indicate this is a chat app

samples/Calling/package.json
anjulgarg: It was a duplicate entry.

packages/react-composites/tests/browser/app/index.tsx
anjulgarg: Done

packages/react-composites/tests/browser/app/public/index.html
anjulgarg: Sure.

packages/react-composites/package.json
```diff
@@ -31,6 +33,7 @@
     "@azure/communication-calling": "1.1.0-beta.2",
     "@azure/communication-chat": "1.0.0",
     "@azure/communication-common": "1.0.0",
+    "@azure/communication-identity": "1.0.0",
```
PorterNan, line 36: I guess we can put it in dev-dependencies?

packages/react-composites/tests/browser/chat/app/identity.ts
```diff
@@ -0,0 +1,32 @@
...
+    userAndTokens.push(await tokenClient.createUserAndToken(['chat']));
+  }
+
+  const chatClient = new ChatClient(endpointUrl, new AzureCommunicationTokenCredential(userAndTokens[0].token));
+  const threadId = (await chatClient.createChatThread({ topic: topic })).chatThread?.id ?? '';
+  await chatClient.getChatThreadClient(threadId).addParticipants({
+    participants: displayNames.map((displayName, i) => ({ id: userAndTokens[i].user, displayName: displayName }))
+  });
+
+  return displayNames.map((displayName, i) => ({
```
PorterNan, line 25: How are return an Object to make it type-strong?

Or define the returned array/Tuple like [CommunicationUserIdentifier, string, string, string ,string]

Then you can spread this array like

createAzureCommunicationChatAdapter(...ret, displayName)

packages/react-composites/tests/browser/chat/app/serve.js
```diff
@@ -0,0 +1,4 @@
+const express = require('express');
```
PorterNan, line 1: Maybe a "webpack devserver" will serve all our needs? So that we don't need to build the bundle then serve it in the server

packages/react-composites/tests/browser/chat/app/server.ts
```diff
@@ -0,0 +1,21 @@
+import express from 'express';
```
PorterNan, line 1: Same here, use webpack dev server so that we can save the whole file of code here

packages/react-composites/tests/browser/chat/app/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT license.
+
+import React, { useState, useEffect } from 'react';
+import ReactDOM from 'react-dom';
+
+import { ChatAdapter, createAzureCommunicationChatAdapter, ChatComposite } from '../../../../src';
+
+import { createUserAndThread } from './identity';
+
+const connectionString = process.env.CONNECTION_STRING;
```
PorterNan, line 11: Is this running in browser code?
So dotnet-webpack plugin is parsing the whole env object to the browser stack, interesting learning

packages/react-composites/tests/browser/chat/app/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+          data[0].token,
+          data[0].endpointUrl,
+          data[0].threadId,
+          displayName
+        )
+      );
+    };
+
+    initialize();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
PorterNan, line 32: Let's dispose adapter here when useEffect ends (don't forget to add adapter to dependency array)

packages/react-composites/tests/browser/chat/app/webpack.config.js
```diff
@@ -0,0 +1,52 @@
...
+      },
+      {
+        test: /\.svg/,
+        type: 'asset/inline'
+      }
+    ]
+  },
+  plugins: [
+    new HtmlWebpackPlugin({ template: path.join(__dirname, 'public', 'index.html') }),
+    new Dotenv({ path: path.join(__dirname, '.env') })
```
PorterNan, line 45: I was pretty sure I saw this .env file somewhere, is it get removed?

One more question is there a way to override env value by running some command line? I guess this overriding ability will be needed when we are setting up our CI

PorterNan (approved):
This is a fantastic one, we can follow this pattern to create some more validations for our all the samples, at least do some basic UI validation on them!

packages/react-composites/tests/browser/chat/app/serve.js
anjulgarg: I need to run the server inline in nodejs code that behaves well with jest.
couldn't make webpack devserver to work that way.

packages/react-composites/tests/browser/chat/app/index.tsx
anjulgarg: yes. :D

packages/react-composites/tests/browser/chat/app/webpack.config.js
anjulgarg: I need to figure that out somehow. In any case, we will be able to pipe an env variable to an .env file through CLI to make it work. But I will invest more time when we get to that stage.

packages/react-composites/package.json
anjulgarg: Good catch

packages/react-composites/tests/browser/chat/app/identity.ts
anjulgarg: Adding types. But can't spread it since I am using the 0th index of the result.

packages/react-composites/README.md
```diff
@@ -1,3 +1,36 @@
...
+```sh
+CONNECTION_STRING=<Resource Connection String>
+```
+
+#### Running the tests
+
+We use Jest as our test runner. Browser based tests can also be run using Jest.
+
+```sh
+rushx test
```
prprabhu-ms, line 35: Doesn't this mean that these tests will be run in CI right now, and break?

Exclude these tests from the default `rush test` and add a separate script to pakage.json

anjulgarg: Very good point. Thanks for catching it. I have excluded the tests from this command for now using jest config. :)

packages/react-composites/jest.config.js
```diff
@@ -3,7 +3,13 @@
...
 const commonConfig = require('../../common/config/jest/jest.config');
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const path = require('path');
 
 module.exports = {
-  ...commonConfig
+  ...commonConfig,
+  roots: [
+    path.join(__dirname, 'src')
+    // path.join(__dirname, 'tests')
```
JamesBurnside, line 13: Add comment why this is commented out (or remove commented line)

packages/react-composites/tests/browser/chat/app/identity.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation.
```
alkwa-msft, line 1: Are we keeping these headers or no? I recall this was a bit of a conversation earlier.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -70,7 +70,7 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
   );
 
   return (
-    <Stack className={chatContainer} grow>
+    <Stack id="ChatComposite" className={chatContainer} grow>
```
JamesBurnside, line 73: Is it possible a user may have two chat composites in their app? (should this be a classname instead?)

packages/react-composites/tests/browser/chat/app/serve.js
alkwa-msft: This looks like we're just using the express server as a static content server

We can probably create some routes on the webpackdevserver as well 

Could be a future modification

packages/react-composites/tests/browser/chat/app/server.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Server } from 'http';
+
+let server: Server;
+const app = express();
+app.use(express.static(__dirname + '/dist'));
+
+export const startServer = (): Promise<void> =>
+  new Promise((resolve, reject) => {
+    server = app.listen(3000, () => {
+      console.log('App listening on 3000');
```
JamesBurnside, line 13: remove console log?

packages/react-composites/tests/browser/chat/app/webpack.config.js
```diff
@@ -0,0 +1,52 @@
...
+const webpack = require('webpack');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+const Dotenv = require('dotenv-webpack');
+
+module.exports = {
+  entry: path.join(__dirname, 'index.tsx'),
+  devtool: 'eval-source-map',
+  resolve: {
+    extensions: ['.ts', '.tsx', '.js'],
+    alias: {
```
JamesBurnside, line 14: hmm, note to self there's a of these alias configs spread across tsconfigs and webpacks files, I wonder if there's a global place we can put them to avoid much redundancy and ensure they're kept consistent

JamesBurnside (approved):

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
anjulgarg: Good Point.

packages/react-composites/tests/browser/chat/app/identity.ts
anjulgarg: I get eslint errors if we dont.

packages/react-composites/package.json
```diff
@@ -104,9 +111,14 @@
...
     "ts-jest": "^26.4.4",
     "ts-node": "^9.1.1",
+    "ts-loader": "^8.0.12",
     "typescript": "4.1.5",
     "uuid": "^8.1.0",
-    "webpack": "5.38.1"
+    "webpack": "5.38.1",
+    "webpack-cli": "4.7.2",
+    "webpack-dev-server": "3.11.2",
+    "dotenv-webpack": "~7.0.3",
```
JamesBurnside, line 120: nit: alphabetize

alkwa-msft (approved):

packages/react-composites/package.json
anjulgarg: Oh I didn't add it manually, it was done by npm/rush automatically.

PR #505, opened 2021-06-25
Bump prettifier versions

prprabhu-ms:
# What
Bump `prettefier` version and get us to a clean state so CI can enforce.

# Why
I want to use string template types. Our current version of prettifier can't understand those.

# How Tested
* `rush prettifier:check`
* CI

# Process & policy checklist


.github/workflows/ci.yml
```diff
@@ -41,6 +41,9 @@ jobs:
       # Perform lint check
       - name: Run linter
         run: rush lint
+      # Perform formatting check
+      - name: Run prettifier
+        run: rush prettifier:check
```
JamesBurnside, line 46: The reason we didn't have this extra step in before was that the eslint config has a prettier extension and covered this, just want to check this step is necessary

common/config/package.json
```diff
@@ -7,7 +7,9 @@
     "lint": "",
     "lint:fix": "",
     "build": "",
-    "test": ""
+    "test": "",
+		"prettier": "",
```
JamesBurnside, line 11: nit: tabs -> spaces

.github/workflows/ci.yml
JamesBurnside: (also in htmlstatic example package.json)

JamesBurnside (approved):
LGTM! Just double check that prettier:check is needed and eslint doesn't cover it

common/config/package.json
JamesBurnside: (also in htmlstatic example package.json)

.github/workflows/ci.yml
prprabhu-ms: I saw some minor differences, as in I had to run `prettifier --write` across the projects to make the check happy.

We might want to actually disable the eslint prettier extension. `eslint` is setup per project, while we want prettifier to be applied uniformly across projects (at least that's what our documentation claims)

JamesBurnside: The only issue with disabling prettier in eslint is sometimes their rules conflict, however we'll probably have to pay the prettier cost twice if we have it as both its own check and in eslint

JamesBurnside: (both CI runtime cost and cost that developers have to run and fix both prettier and eslint errors)

prprabhu-ms: Let me pull out this CI change for starters. We can always add that back in if it makes sense.

We already have this conflict in principle: the pre-commit hook forces you to fix prettier errors, while CI uses the eslint-prettier plugin. And I found that the didn't quite agree. My guess is that it's because the per-project eslint rules didn't have the same prettier config as the top-level one.

samples/Calling/package.json
```diff
@@ -10,8 +10,8 @@
     "start:app": "webpack-cli serve --mode=development --env development",
     "start:prod": "node dist/server.js",
     "test": "",
-    "prettier": "prettier --write --config ../../.prettierrc **/*.js **/*.ts",
-    "prettier:check": "prettier --check --config ../../.prettierrc **/*.js **/*.ts",
+    "prettier": "prettier --no-error-on-unmatched-pattern --write --config ../../.prettierrc --ignore-path=../../.prettierignore \"**/*.js\" \"**/*.jsx\" \"**/*.ts\" \"**/*.tsx\"",
```
PorterNan, line 13: Is it possible to add this one into the config file? And I guess this is design for those small repos that doesn't have enough files, should we add this flag to all the repos? (I am a little worried about that we might introduce a wrong pattern in the future to match no files, but this flag just hides all of them)

common/config/rush/command-line.json
```diff
@@ -59,6 +59,17 @@
       "summary": "Trigger incremental build with watch command.",
       "enableParallelism": false,
       "watchForChanges": true
+    },
+    {
+      "commandKind": "bulk",
+      "name": "prettier:check",
```
PorterNan, line 65: Love this command, I was doing a manual check 1 by 1 before

PorterNan (approved):
Awesome, some patterns of new prettier make more sense to me, I don't review all tsx file changes because I assume they are auto fixed by new prettier:D

samples/Calling/package.json
prprabhu-ms: Some packages have only `.ts` files while some have `.tsx` as well. Some have `.js`, some don't.

I think it's better to just add all patterns for all packages right now. Otherwise whenever the first `.tsx` file is added, we'd have to remember to go update this script.

prprabhu-ms: Unfortunately, can't move this CLI arg to config.

PR #506, opened 2021-06-25
Fix: scrollTop does not exist on undefined object

JamesBurnside:
# What
Remove `any` types on message thread and fix null checks

# Why
I recall @ddematheu2 hitting an error when playing around with composites that scrollTop was called on an undefined object after hiding a chat window.
Looks to be because we have an async function performing work on a ref that will have been cleaned up - fixing the `any` types as these would've caught this error.

# How Tested
🤞

### DIFF
--- a/packages/react-components/src/components/MessageThread.tsx
+++ b/packages/react-components/src/components/MessageThread.tsx
@@ -613,7 +613,7 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
   }, [sendMessageStatusIfAtBottom]);
 
   const handleScroll = (): void => {
-    if (chatScrollDivRef.current === undefined) {
+    if (!chatScrollDivRef.current) {
       return;
     }
 
--- a/packages/react-components/src/components/MessageThread.tsx
+++ b/packages/react-components/src/components/MessageThread.tsx
@@ -541,8 +541,8 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
   const isAllChatMessagesLoadedRef = useRef(isAllChatMessagesLoaded);
   isAllChatMessagesLoadedRef.current = isAllChatMessagesLoaded;
 
-  const chatScrollDivRef: any = useRef();
-  const chatThreadRef: any = useRef();
+  const chatScrollDivRef = useRef<HTMLElement>(null);
+  const chatThreadRef = useRef<HTMLElement>(null);
   const isLoadingChatMessagesRef = useRef(false);
 
   const messagesRef = useRef(messages);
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -604,13 +604,19 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
   }, [showMessageStatus, onMessageSeen, chatMessageIdJustSeen]);
 
   const scrollToBottom = useCallback((): void => {
-    chatScrollDivRef.current.scrollTop = chatScrollDivRef.current.scrollHeight;
+    if (chatScrollDivRef.current) {
+      chatScrollDivRef.current.scrollTop = chatScrollDivRef?.current.scrollHeight;
```
mgamis-msft, line 608: super nit: is the `?` necessary?

mgamis-msft (approved):

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -625,11 +631,13 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
     setIsAtTopOfScrollRef(atTop);
 
     // Make sure we do not stuck at the top if more messages are being fetched.
-    if (chatScrollDivRef.current.scrollTop === 0 && !isAllChatMessagesLoadedRef.current)
+    if (chatScrollDivRef.current && chatScrollDivRef.current.scrollTop === 0 && !isAllChatMessagesLoadedRef.current) {
```
PorterNan, line 634: chatScrollDivRef.current?.scrollTop === 0
Will this work?

PorterNan: Ohh, you might wanna pass the type guard in following line

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -712,7 +720,11 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
   useEffect(() => {
     (async () => {
       if (onLoadPreviousChatMessages) {
-        while (chatScrollDivRef.current.scrollTop <= 200 && !isAllChatMessagesLoadedRef.current) {
+        while (
+          chatScrollDivRef.current &&
+          chatScrollDivRef.current.scrollTop <= 200 &&
```
PorterNan, line 725: I remember
(undefined <= 200) === false

So maybe this could be:
chatScrollDivRef.current?.scrollTop <= 200

PorterNan (approved):
A nice bug catch! One more usability bug for this scroller is sometimes it will be too quick to reach the top when scroller becomes too long (according to my personal experience), but not related to this PR

packages/react-components/src/components/MessageThread.tsx
JamesBurnside: nope! will remove

packages/react-components/src/components/MessageThread.tsx
JamesBurnside: Oh I already have a 
```
if (!chatScrollDivRef.current) {
  return;
}
```
I can remove this extra check

packages/react-components/src/components/MessageThread.tsx
JamesBurnside: I thought that also but tsc complains:
image

packages/react-components/src/components/MessageThread.tsx
JamesBurnside: Removed

packages/react-components/src/components/MessageThread.tsx
JamesBurnside: Removed

PR #507, opened 2021-06-25
Set `ensureConsistentVersions` in rush.json to `true`

JamesBurnside:
# What
Set `ensureConsistentVersions` to true in the rush.json and fixup mismatching versions

# Why
This was removed a while back when we weren't ready to fully update headless sdk versions.

# How Tested
Ran chat locally just


packages/api-extractor-fork/bin/api-extractor
```diff
@@ -1,2 +0,0 @@
-#!/usr/bin/env node
```
JamesBurnside, line 1: not sure what happened here, reverting

PorterNan (approved):
This is great, since we have a more stable repo, this should fix some problems I encountered for many times

BTW, do we wanna unlock typescript version from 4.1.5 as well? It was problematic to update typescript version before, but I guess we will need another validation to see if it works well now(since the issue happened 3 months ago)

JamesBurnside:
> This is great, since we have a more stable repo, this should fix some problems I encountered for many times
> 
> BTW, do we wanna unlock typescript version from 4.1.5 as well? It was problematic to update typescript version before, but I guess we will need another validation to see if it works well now(since the issue happened 3 months ago)

Yep I think we should be able to upgrade the typescript version freely! but we should keep the definition in the package.jsons as a `~` or a specific version as typescript doesn't follow semver so having `^` may break someone at any time

alkwa-msft (approved):

PR #510, opened 2021-06-28
Prototype implementation of error handling

prprabhu-ms:
# What
This is a prototype implementation for how errors from the SDK will be plumbed through the layers:

Uses a couple methods on Chat as examples.

* Stateful sets the latest error per method in the state.
* Adapter mirrors the latest error per method in the state.
* Adapter also geneartes on('error') events on errors.

In particular, the following ideas are prototyped:
* Stateful clears error for a method on success of subsequent method calls
* Adapter exposes errors in a form that the API is not tied to the stateful-backed implementation, but the bridge code is trivial.

Still todo:
* Add methods for Contoso to clear errors (didn't have a large bearing on implementation plan)
* Add unittests for adapters (there aren't any at the moment, so need some work to se things up).

# Why
Help set direction for comprehensive error reporting.

# How Tested
Unittests for stateful. Prayers for composite.

# Process & policy checklist


packages/chat-stateful-client/review/chat-stateful-client.api.md
```diff
@@ -21,14 +22,38 @@ export type ChatClientState = {
     threads: {
         [key: string]: ChatThreadClientState;
     };
+    latestErrors: ChatErrors;
 };
 
+// @public
+export class ChatError extends Error {
```
prprabhu-ms, line 29: Somewhat inline with @DominikMe suggestion, decided we do want to include the target that failed in the rethrown exception. But didn't go all the way in defining what the error internals look like, since we don't control that and the internals different greatly in Chat vs Calling.

packages/chat-stateful-client/review/chat-stateful-client.api.md
```diff
@@ -52,6 +77,9 @@ export type ChatThreadProperties = {
 // @public
 export const createStatefulChatClient: (args: StatefulChatClientArgs, options?: StatefulChatClientOptions | undefined) => StatefulChatClient;
 
+// @public
+export const isChatErrorTarget: (target: string) => boolean;
```
prprabhu-ms, line 81: This is unfortunate. If there's an idiomatic way to enumerate the values of a string literal union type at runtime, please let me know. The reason this is exposed is so that the adapter can restrict the operations to only the targets acceptable to the stateful client in `useAdaptedSelector` implementation.

packages/chat-stateful-client/src/ChatClientState.ts
```diff
@@ -14,6 +14,12 @@ export type ChatClientState = {
    * Object with {@Link ChatThreadClientState} fields, keyed by {@Link ChatThreadClientState.threadId}.
    */
   threads: { [key: string]: ChatThreadClientState };
+  /**
+   * Stores the latest error for each API method.
+   *
+   * See documentation of {@Link ChatErrors} for details.
+   */
+  latestErrors: ChatErrors;
```
prprabhu-ms, line 22: As @JamesBurnside had recommended, limiting both the stateful and adapter state to store only the latest errors.

In contrast, the on('error') event from adapter will trigger for each error.

There's no demonstrated use-case for storing more than one error per method and it adds a lot of complexity both for Contoso and the implementation to deal with lists of errors (e.g., when do we clear the list / who does).

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -224,10 +241,25 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
...
+  off(event: 'error', listener: ChatErrorListener): void;
   off(event: ChatEvent, listener: (e: any) => void): void {
+    if (event === 'error') {
+      this.context.off(event, listener);
+      return;
+    }
     this.emitter.off(event, listener);
   }
+
+  private async asyncTeeErrorToEventEmitter<T>(f: () => Promise<T>): Promise<T> {
```
prprabhu-ms, line 253: The on('error') event is *not* generated from the error stored in state.

This is one example of a general idea: Any computation of events by diffing state will be lossy. We must allow state updates to be batched for performance reasons. Users of stateful clients only care about binding to the "current" state of the data. Thus, it is possible that an observer of state changes misses some intermediate transitions.

.github/workflows/ci.yml
```diff
@@ -41,6 +41,9 @@ jobs:
       # Perform lint check
       - name: Run linter
         run: rush lint
+      # Perform formatting check
```
prprabhu-ms, line 44: Needs to be removed. Bad merge from main.

packages/chat-stateful-client/.vscode/launch.json
```diff
@@ -18,7 +18,16 @@
       "type": "node",
       "request": "launch",
       "runtimeExecutable": "${workspaceRoot}/node_modules/.bin/jest",
-      "args": ["test", "--runInBand", "--no-cache", "--watchAll=false", "/${fileBasename}"],
+      "args": [
```
prprabhu-ms, line 21: Leaked from local debugging. Needs to be removed.

packages/chat-stateful-client/review/chat-stateful-client.api.md
```diff
@@ -21,14 +22,38 @@ export type ChatClientState = {
...
     clientMessageId?: string;
     status: MessageStatus;
 };
 
+// @public
+export type ChatMethodName<T, K extends keyof T> = T[K] extends Function ? (K extends string ? K : never) : never;
+
+// @public
+export type ChatObjectMethodNames<TName extends string, T> = {
+    [K in keyof T]: `${TName}.${ChatMethodName<T, K>}`;
```
JamesBurnside, line 54: This is snazzy!

packages/chat-stateful-client/src/ChatClientState.ts
```diff
@@ -45,3 +51,82 @@ export type ChatThreadClientState = {
...
+   * Error thrown by the failed SDK method.
+   */
+  public inner: Error;
+
+  constructor(target: ChatErrorTargets, inner: Error) {
+    super();
+    this.target = target;
+    this.inner = inner;
+    this.name = 'ChatError';
+    this.message = `${this.target}: ${this.inner.message}`;
```
JamesBurnside, line 88: Is this purely informational or do we expect developers to write conditionals of this?

JamesBurnside: Actually better question, which property do we expect developers to write conditionals over?

packages/chat-stateful-client/src/ChatClientState.ts
```diff
@@ -45,3 +51,82 @@ export type ChatThreadClientState = {
...
+export type ChatMethodName<T, K extends keyof T> = T[K] extends Function ? (K extends string ? K : never) : never;
+
+/**
+ * Method to decide at runtime if a string is an error target.
+ */
+export const isChatErrorTarget = (target: string): boolean => {
+  for (const targetPrefix in ChatErrorTargetPrefixes) {
+    const [prefix, obj] = targetPrefix;
+    if (target.startsWith(prefix)) {
+      target = target.substring(prefix.length);
```
JamesBurnside, line 120: nit/question: if target starts with prefix, and we take a substring of length prefix, does target == prefix? (i.e. could be `target = prefix;`)

packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -317,6 +325,54 @@ export class ChatContext {
     }
   }
 
+  /**
+   * Tees any errors encountered in an async function to the state.
```
JamesBurnside, line 329: Is `tee` something easily understandable to non-english first language devs or is it a colloquial verb? (another option could be propagate but I do like tee)

packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -317,6 +325,54 @@ export class ChatContext {
...
+   *
+   * If the function succeeds, clears associated errors from the state.
+   *
+   * @param f Async function to execute.
+   * @param target The error target to tee error to.
+   * @param clearTargets The error targets to clear errors for if the function succeeds. By default, clears errors for `target.
+   * @returns Result of calling `f`. Also re-raises any exceptions thrown from `f`.
+   * @throws ChatError. Exceptions thrown from `f` are tagged with the failed `target.
+   */
+  public async asyncTeeErrorToState<T>(
```
JamesBurnside, line 339: nit: shouldn't need the async if the function itself is marked async

JamesBurnside: oh never mind this is a decorator of kinds, never mind keep the async

JamesBurnside: in fact could we use a javascript decorator here? (I've never actually used one so might not be possible but seems like a good application of one)

packages/communication-react/review/communication-react.api.md
```diff
@@ -829,6 +865,9 @@ export interface IncomingCallState {
     startTime: Date;
 }
 
+// @public
+export const isChatErrorTarget: (target: string) => boolean;
```
JamesBurnside, line 869: Can this be ommitted from the public api or will external devs need to make use of it also?

JamesBurnside:
Looks great to me!

packages/chat-stateful-client/review/chat-stateful-client.api.md
JamesBurnside: Yeah unfortunately literal union types don't exist as an object in the transpiled javascript (purely used for typing), is the downsides of this that we have to maintain an array of the known targets?

packages/communication-react/review/communication-react.api.md
prprabhu-ms: +1 I thought of this later last night. I'll move the implementation to the adapter that uses it.

packages/chat-stateful-client/src/ChatContext.ts
prprabhu-ms: I looked at decorators but can't use one here. They're meant to be used for classes or class methods. We're dispatching a lambda off of a switch statement in an object proxy...

The purpose of decorators is to avoid losing the "class method"-ness of class methods when wrapping with other higher-order functions. Without decorators, the setup would look like:
- You have a class with a method.
- You have a wannaBeDecorator() which accepts a function and returns a modified function.
- If you directly wrap the class method with wannaBeDecorator, you'll lose certain metadata that the function has from being a class method.

The decorator form of wannaBeDecorator() helps preserve that metadata.

In this case, no such need exists -- the functions being modified are already unnamed function objects.

Reference: https://www.simplethread.com/understanding-js-decorators/

packages/chat-stateful-client/src/ChatContext.ts
prprabhu-ms: `tee` is standard for linux developers: https://www.man7.org/linux/man-pages/man1/tee.1.html

Seems to mean the same thing on most platforms: https://en.wikipedia.org/wiki/Tee_(command)

packages/chat-stateful-client/src/ChatClientState.ts
prprabhu-ms: We expect them to use `this.target` and `this.inner`.

That's why I created a custom `Error` subclass over simply throwing a new `Error('${target}: ${inner}')` in the helper function.

packages/chat-stateful-client/review/chat-stateful-client.api.md
prprabhu-ms: Had to bump `prettify` so it'd let me use this :P

packages/chat-stateful-client/src/ChatClientState.ts
prprabhu-ms: `target.substring(prefix.length)` gives back the substring of `target` *starting at* index `prefix.length` to the end of the string: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring

This should be covered by unit tests I added, so I'm confident this works as expected.

packages/chat-stateful-client/review/chat-stateful-client.api.md
prprabhu-ms: Essentially.

https://www.npmjs.com/package/runtypes could be used here. But we don't have that dependency yet, 
Anyway, I'll move this off the public API, then we can improve upon the implementation later on.

packages/communication-react/review/communication-react.api.md
prprabhu-ms: Done.

packages/chat-stateful-client/review/chat-stateful-client.api.md
prprabhu-ms: Moved implementation to the composite where it's used, off public API.

Can consider using a generic library to get runtime type reflection later.

packages/chat-stateful-client/src/ChatClientState.ts
```diff
@@ -45,3 +51,61 @@ export type ChatThreadClientState = {
...
+/**
+ * Errors teed from API calls to the Chat SDK.
+ *
+ * Each property in the object stores the latest error for a particular SDK API method.
+ *
+ * Errors from this object can be cleared by calling the TODO(implement me) {@Link clearError} method.
+ * Additionally, errors are automatically cleared when:
+ * - The state is cleared.
+ * - Subsequent calls to related API methods succeed.
+ *   See documentation of individual stateful client methods for details on when errors may be automatically cleared.
```
JamesBurnside, line 64: super nit: extra spacing in comment
```suggestion
 * See documentation of individual stateful client methods for details on when errors may be automatically cleared.
```

JamesBurnside (approved):

packages/chat-stateful-client/review/chat-stateful-client.api.md
```diff
@@ -21,14 +22,38 @@ export type ChatClientState = {
...
 export type ChatMessageWithStatus = ChatMessage & {
     clientMessageId?: string;
     status: MessageStatus;
 };
 
+// @public
+export type ChatMethodName<T, K extends keyof T> = T[K] extends Function ? (K extends string ? K : never) : never;
+
+// @public
+export type ChatObjectMethodNames<TName extends string, T> = {
```
PorterNan, line 53: This is brilliant! Love this one.
Do we get a chance to validate this one for minimized js bundle(js-uglifier vs js minify)? I wonder if this still works, or it might be working in another unexpected way. Looks like this is generating the object name + function name in run time, curious what it would be like when code gets minified

prprabhu-ms: Right you are.

The type definition here is OK, that's anyway only for type safety, and doesn't survive in the minified code.

But the logic that tried to list the valid operations in `useAdaptedSelectors` was wrong. I dropped it. Please take another look.

alkwa-msft (approved):

packages/chat-stateful-client/review/chat-stateful-client.api.md
```diff
@@ -21,14 +22,38 @@ export type ChatClientState = {
...
+// @public
+export class ChatError extends Error {
+    constructor(target: ChatErrorTargets, inner: Error);
+    inner: Error;
+    target: ChatErrorTargets;
+}
+
+// @public
+export type ChatErrors = {
+    [target in ChatErrorTargets]: Error;
```
mgamis-msft, line 37: Just to clarify this is a more constrained form of Record<string, Error> that ensures the string is an actual method from ChatClient or ChatThreadClient? If yes, that's pretty cool.

packages/chat-stateful-client/src/ChatClientState.ts
mgamis-msft, line 22: Sorry, might be a stupid question but would we have a selector for this in the future to surface errors to the UI? For example, if we failed to change the topic of the chat. The error would be available in latestErrors['ChatThreadClient.updateTopic']? Would the user of the selector would need to know the stateful function name?

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -153,7 +156,9 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
   }
```
mgamis-msft, line 156: How come the functions up here are not teed up to the emitter?

packages/chat-stateful-client/review/chat-stateful-client.api.md
prprabhu-ms: Yep, that's all it is.

And you get auto-complete in vscode when you look for errors in the state :)

packages/chat-stateful-client/src/ChatClientState.ts
prprabhu-ms: Yes and no.

There will be some generic selectors which just show _any_ error from the object (e.g, show the latest 3 errors in an `ErrorBar`). Those will not need to know the exact values.

There will be others that need to pick out a particular error target (e.g., disable the `SendBox` when we failed to send message because the backend throttled us). Those need to know the operation that failed in _some_ way. I think directly using the method names is best. Otherwise, we'd end up creating another enum that somehow maps to the methods. The bindings will also get the type information, so you can auto-complete to find what property names are valid in `ChatErrors`.

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
prprabhu-ms: Oh, this is a prototype PR that sets up all the plubming and a couple of examples that demonstrate the mechanics end-to-end. I like to do this to make it easier for me to develop and for you to review -- less repetitive changes to look at.

Once this PR lands, I'll upload a fairly mechanical PR to go add the same behaviour to all the remaining methods.

mgamis-msft (approved):

packages/chat-stateful-client/src/ChatClientState.ts
```diff
@@ -45,3 +51,61 @@ export type ChatThreadClientState = {
...
+export type ChatErrors = {
+  [target in ChatErrorTargets]: Error;
+};
+
+/**
+ * Error thrown from failed stateful API methods.
+ */
+export class ChatError extends Error {
+  /**
+   * The API method target that failed.
```
PorterNan, line 75: Maybe we could move all these types to somewhere and make this file shorter?

prprabhu-ms: Will look at splitting up this file in one of my follow up PRs.

packages/chat-stateful-client/src/StatefulChatThreadClient.ts
```diff
@@ -120,11 +120,13 @@ class ProxyChatThreadClient implements ProxyHandler<ChatThreadClient> {
       }
       case 'updateTopic': {
         return async (...args: Parameters<ChatThreadClient['updateTopic']>) => {
-          const result = await chatThreadClient.updateTopic(...args);
-          const [topic] = args;
+          return this._context.asyncTeeErrorToState(async () => {
```
PorterNan, line 123: I guess we might need to wrap every function with this in the future? Or is it just for this one?

Since we are already using proxy, maybe we could just add a single wrapper for all the functions using proxy.

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -153,7 +156,9 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
...
 
   async setTopic(topicName: string): Promise<void> {
-    await this.handlers.updateThreadTopicName(topicName);
+    await this.asyncTeeErrorToEventEmitter(async () => {
+      await this.handlers.updateThreadTopicName(topicName);
```
PorterNan, line 160: Will we add error wrapper to updateTopic function in stateful client? If yes, then we might not need to wrap it again in adapter implementation

PR #512, opened 2021-06-29
Localized strings in control buttons

mgamis-msft:
# What

Localized strings in control buttons

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2484983

# How Tested

Added unit tests for button labels. Still trying to figure our how to test menu strings because it is hard to simulate clicks on FluentUI components with enzyme.

# Process & policy checklist


JamesBurnside (approved):

packages/communication-react/review/communication-react.api.md
```diff
@@ -638,7 +638,10 @@ export interface ComponentStrings {
     endCallButton: EndCallButtonStrings;
     messageThread: MessageThreadStrings;
     microphoneButton: MicrophoneButtonStrings;
+    optionsButton: OptionsButtonStrings;
```
PorterNan, line 641: Sry, I am late to the party (missed your first PR of localization Provider, which I just read 10 mins ago, and that is an excellent one)

I have several questions in general for localization, and look forward to some discussions:

1. I love your context/Provider way of providing a big object for localization(that is a smart design), and looks like we are providing 2 different options - one is from provider, one is passing strings from props. Are there any special concern that we are providing 2nd options, is it possible that we allow developers to overrider partial of strings in provider then reach the same purpose?(BTW, adding an interface for each component for string is adding more burden to component development)

2. Naming, according to my experience, the best naming method for each localization string is using its English string, with camel case and no space (unless it's too too long, more than 4-5 words), there are several reasons:
   -Naming is hard, could be time consuming for developers to think of(and code reviewer will be against your idea of naming in the end), so if the string name standard is just the English string itself, it will be time saving and 100% straightforward
   -Reusability, some of strings in the same value could be shared, for example an 'Azure Communication' string could be used either in config screen and calling screen as title, we don't need to create 2 strings here, instead of strings.ConfigScreenTitle and strings.CallScreenTitle, we just need one strings.AzureCommunication
   -Readability, sometimes I need to go to more than 1 file to find out which string the Button label really is, which interrupts my code reading, would be nice that property name itself represents its meaning, and under most cases I don't have to jump to another file to see the real string.

mgamis-msft: Hey Nan, no problem. Thanks for the feedback!

1. Yes it is possible that the Provider takes in a deep partial locale such that developers can provide only part of the json entries. PartialDeep from type-fest is one option and it helps with can help with intellisense since I use it in this [PR's tests](https://github.com/Azure/communication-ui-library/pull/512/files#diff-6f11376f221fa781705f9ca4614df58d1c627f6d6372915b86ece73d702d16d6) to create a test locale.

I agree it is a burden to add an interface for each component but after discussion with Prathmesh we agreed that point of the interfaces is that developers will have a consistent expectation that strings that are localizable/overridable are in the 'strings' property with an interface for the particular component.

2. 
Naming is hard - The naming is generally componentName.{where the string is used}. I agree the second part is very subjective. 
For example, for the camera button label when it camera is off the entry name, I named the entry cameraButton.offLabel. Did you have something like 'cameraButton.turnOn' in mind? Or did you mean flat json key-values like 'cameraButtonTurnOn'?

Reusability - Currently there are no strings that are the same for any two components. That could obviously change as we update our components.
But I agree, that the possibility of redundancy of a string is one of the drawbacks of having a nested entries per component.

Readability - In general, in every component there should initially be a resolution of which strings should be used
`const localeStrings = useLocale().strings.(componentName);
const strings = { ...localeStrings, ...props.strings };`

now this strings json object could potentially be passed around in the component with several helper methods. Is this what you mean by 'looking in several places'?
If yes, I agree but I don't think this can be avoided with any form of strings coming which could come from props. I would love to know another approach that I hadn't considered 
which may simpler.

This is good input and I may need to present during demo because nested json structure for localization is not the most popular approach. But it seems TD Build recognizes as an alternative [here](https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47/oauth2/authorize?response_type=id_token&redirect_uri=https%3A%2F%2Fwww.1eswiki.com%2F.auth%2Flogin%2Faad%2Fcallback&client_id=a8320cd7-e30c-4eb0-ac83-e5c97a1d3601&scope=openid+profile+email&response_mode=form_post&nonce=c2ed4fc360bb4f6bbfabff9b0746d105_20210702102611&state=redir%3D%252Fwiki%252FLocalization_Guideline%253A_JSON&sso_reload=true)

PorterNan: Thanks for walking through these questions 1 by 1, all your answers make sense to me. Now I get that we are using a nested json object in context to control localization globally

Let's come back to question one, there are 2 assumptions I got from your answer: 1. each component localized strings are partial overridable; 2. Props override has higher priority than context override
Imagine I am a developer who wants to override just one component, there are 2 ways I can do for it

1. using provider:
```
<Provider overrideStrings={{optionButtonStrings: {label: 'Options', cameraMenuTitle: 'Camera'}}}>
<OptionalButton />
</Provider>
```
2. using props
<OptionalButton strings = {{label: 'Options', cameraMenuTitle: 'Camera'}} />
 
Do we have a guideline on different scenarioes for which options should I choose? It seems to me that we are providing 2 very similar solutions for our customers to make their own choice, which might causes a little bit confusion(or someone will mix this 2 method up in just one project which is a little messy). IMO, it looks like first option is more for i18n, and 2nd one is more for overriding single string but not for i18n, but overall they are still pretty similar to each other

prprabhu-ms: Note that we extracted per-component-interface for strings (mostly) from existing props per component. So this is just better encapsulation of existing props.

The question around having multiple ways to do the same thing is a very important one. Thanks for raising that.
I find the current setup natural, but I don't have much experience using UI libraries.

* Each component accepts `strings` prop. You never have to guess what strings can be localized for a component -- it always accepts all those strings via props. This doesn't add too much work to component author -- you anyway have to make sure the strings inside the component are localized. This also doesn't add work to the consumer of the component, since they can just skip setting those props. And it serves as compile-time-verified documentation for strings.
* The `LocalizationProvider` is how Contoso would connect their app's localization framework with all our components in a single place.

As you said, the former will likely be used by somebody trying to override some single string, not for i18n purposes; the latter will likely be used for i18n. Still the former serves as good obvious documentation when using the component.

prprabhu-ms: About using the same string for multiple components -- I don't think this will be a very common case. And in those cases, it's better to be DAMP than DRY -- explicitly listing the duplicate string in a very dumb strings struct is better than figuring out the duplicate places where a single string ends up getting used.

mgamis-msft: Thanks guys for the feedback and pointing out the potential confusion of having to ways to provide strings. I will start to work on documentation for storybook for Localization. I think it will help developers if I explain how localization and overriding works and could also help us convey our approach to our architects and get further feedback.

alkwa-msft (approved):

PR #517, opened 2021-06-30
Chat Composite - E2E Tests

anjulgarg:
# What
Playwright based browser automation for running chat composite tests

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/SendBox.tsx
```diff
@@ -13,6 +13,7 @@ import {
 } from './styles/SendBox.styles';
 import { BaseCustomStylesProps } from '../types';
 import { isDarkThemed } from '../theming/themeUtils';
+import { COMPONENT_UI_IDS } from './identifiers';
```
prprabhu-ms, line 16: nit: Use CamelCase for constants.

packages/react-composites/jest.config.js
```diff
@@ -8,10 +8,5 @@ const path = require('path');
 
 module.exports = {
   ...commonConfig,
-  roots: [
-    path.join(__dirname, 'src')
-    // Removing the `tests` dir from jest tests directories to prevent these tests
-    // from running with the command `npm test` | `rushx test`
-    // path.join(__dirname, 'tests')
-  ]
+  roots: [path.join(__dirname, 'src'), path.join(__dirname, 'tests')]
```
prprabhu-ms, line 11: Don't add these to CI just yet. I'm wary of running any tests in CI that depend on a real ACS backend service being available. CI should be hermetic.

prprabhu-ms (commented):
Overall, I'd recommend an ID injection approach.

This PR increases the composite API surface by exporting fixed IDs for internal elements. If an API exists, it will be used ;)
What happens if there are two composite instances in the same app?

Could we invert this: The composite exposes an optional prop that accepts an ID for different elements in the composite (this prop starts looking very much like the locale prop we're already adding). The test then creates the composite with the IDs it wants and goes on to target elements that way.

Contoso could also use this feature now to target internal elements. Either way we're expanding the API for composites, but client-supplied IDs seem like a better choice.

packages/react-composites/jest.config.js
anjulgarg: Yes, it's just for the draft to give everyone the complete picture of how the end product will look like :)

packages/react-components/src/components/SendBox.tsx
anjulgarg: Any particular reason?

prprabhu-ms: Hmm.. Ignore this recommendation. Looks like we don't have a style enforcement for this. Nor does Azure: https://github.com/Azure/azure-sdk-for-js/blob/e823b26f4600a698eb1c06f48da9b780fb2331a7/common/tools/eslint-plugin-azure-sdk/src/configs/azure-sdk-base.ts#L67

In JavaScript, top level definitions are all constants anyway (most functions are also just `const` function objects).
So I thought that there might be a convention to not capitalize string / numerical "constants" either ('cause what's so different about them?)

A quick look around tells me that UPPER_CASE is indeed the convention though, so you're right:

* [Airbnb](https://github.com/airbnb/javascript/blob/master/README.md#naming--uppercase) uses UPPER_CASE, but the style guide explicitly says that it should be used for exported constants that are guaranteed to not change (in case of nested object values). This is what I was afraid of...
* https://google.github.io/styleguide/jsguide.html#naming-constant-names also uses UPPER_CASE
* https://github.com/rwaldron/idiomatic.js/ also uses UPPER_CASE

packages/react-composites/jest.config.js
alkwa-msft: I am curious if its possible to pass in a mock adapter where we can stub out the calls to the backend and have it return what we want to return for the puppeteer tests?

packages/react-composites/tests/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": "./",
+    "outDir": "../dist",
```
JamesBurnside, line 5: nit: does this need an outDir if noEmit is set to true?

packages/communication-react/review/communication-react.api.md
```diff
@@ -426,6 +426,14 @@ export interface CameraButtonStrings {
     onLabel: string;
 }
 
+// @public (undocumented)
+export const CHAT_UI_IDS: {
```
JamesBurnside, line 430: nit (though realize this is draft but leaving this here anyway): these look like they should be excluded from the public api

packages/react-composites/jest.config.js
prprabhu-ms: A 100 times this!

My super-secret evil plan with https://github.com/Azure/communication-ui-library/blob/main/packages/react-composites/src/composites/ChatComposite/adapter/StubChatClient.ts is to implement a full fake for the backend services (i.e., fully functional implementation that's fully local and deterministic).

I think we absolutely need something like that to be able to run integration tests in CI.

packages/react-composites/tests/tsconfig.json
prprabhu-ms: Counter-point: integration tests should be black-box tests -- if they need IDs for components inside the composite, _some way_ to get at those IDs should be part of the public API.

Making integration tests grey-box tests inevitably leads to some business logic in the system under test that is test-specific.

packages/communication-react/review/communication-react.api.md
prprabhu-ms: Counter-point: integration tests should be black-box tests -- if they need IDs for components inside the composite, some way to get at those IDs should be part of the public API.

Making integration tests grey-box tests inevitably leads to some business logic in the system under test that is test-specific.

packages/react-composites/jest.config.js
anjulgarg: Can do but since we are using this for E2E tests, a stub would defeat the purpose. What do you think?

packages/communication-react/review/communication-react.api.md
anjulgarg: +1 to what @prprabhu-ms said. By not exposing these constants in the public API we end up doing absolute imports and break the ability of this test to be run on a black box composite...

packages/react-composites/tests/tsconfig.json
anjulgarg: Nope. Removed it.

packages/communication-react/review/communication-react.api.md
anjulgarg: I have mixed feelings about this. Would love to know what others think

anjulgarg:
> Overall, I'd recommend an ID injection approach.
> 
> This PR increases the composite API surface by exporting fixed IDs for internal elements. If an API exists, it will be used ;)
> What happens if there are two composite instances in the same app?
> 
> Could we invert this: The composite exposes an optional prop that accepts an ID for different elements in the composite (this prop starts looking very much like the locale prop we're already adding). The test then creates the composite with the IDs it wants and goes on to target elements that way.
> 
> Contoso could also use this feature now to target internal elements. Either way we're expanding the API for composites, but client-supplied IDs seem like a better choice.

Thanks for the suggestion. It's a great idea. I am looking forward to more such suggestions.
100% onboard with this one. Not sure exactly how to implement it in a sensible way though.
What are your thoughts about merging this branch and working on making these ID's injectable in another branch?

packages/communication-react/review/communication-react.api.md
JamesBurnside: I think keep them public, I've been swayed 😄 easier to keep them public and have the azure review board remove them than the other way around

packages/react-composites/jest.config.js
```diff
@@ -10,8 +10,7 @@ module.exports = {
   ...commonConfig,
   roots: [
     path.join(__dirname, 'src')
-    // Removing the `tests` dir from jest tests directories to prevent these tests
-    // from running with the command `npm test` | `rushx test`
+    // Uncomment the following line to run E2E browser tests
```
JamesBurnside, line 13: Can we have an e2e.jest.config instead of having to manipulate this file?

packages/react-composites/package.json
```diff
@@ -118,8 +118,11 @@
     "webpack": "5.38.1",
     "webpack-cli": "4.7.2",
     "webpack-dev-server": "3.11.2",
-    "dotenv-webpack": "~7.0.3",
-    "@types/puppeteer": "~5.4.3"
+    "@types/puppeteer": "~5.4.3",
+    "playwright": "~1.12.3",
+    "dotenv": "~10.0.0",
+    "faker": "~5.5.3",
+    "@types/faker": "~5.5.7"
```
JamesBurnside, line 125: nit: alphabetize

packages/react-composites/README.md
```diff
@@ -4,7 +4,7 @@ Todo: _documentation to follow_
 
 ## Testing
 
-### Automated Browser Tests
+### Automated Browser Tests (E2E)
 
 Currently, browser based tests for Chat composite on Chrome using puppeteer are supported.
```
JamesBurnside, line 9: playwright instead of puppeteer here?

packages/react-composites/jest.config.js
JamesBurnside: Then also a test:e2e command in the package.json that targets this maybe?

packages/react-composites/README.md
```diff
@@ -16,12 +16,12 @@ rush update
...
-The webpack bundler for test react app reads environment variables from the `.env` file and assigns them to the `process.env` variable.
-These variables can then be accessed inside the code using `process.env.{VAR_NAME}`
+Create a `.env` file inside `tests/browser` directory.
 
-Create a `.env` file inside `tests/browser/app` directory.
+The E2E browser tests require a valid connection string to run. This connection string is provided to the test runner through environment variables.
+The environment variables inside a `.env` file are loaded at the beginning of E2E tests.
 
-Sample `tests/browser/app/.env` file
+Sample `tests/browser/.env` file
```
JamesBurnside, line 24: Should we add a `.env` file that has CONNECTION_STRING=REPLACE_WITH_CONNECTION_STRING ?

JamesBurnside: So its a similar experience to running the samples

anjulgarg: .env files should never be committed to the source.
That would defeat the purpose of using a .env file.

packages/communication-react/review/communication-react.api.md
alkwa-msft: I kind of like the idea that we have these ids here so if someone else wants to do end to end testing with UI components they have a bunch of hooks to have good end to end tests?

packages/react-composites/jest.config.js
anjulgarg: Done

anjulgarg: Great advice. Thanks

packages/react-composites/package.json
anjulgarg: Done

packages/react-composites/README.md
anjulgarg: Done

packages/communication-react/review/communication-react.api.md
anjulgarg: Is acs the official acronym for our service?

prprabhu-ms: With the ID injection approach (OK to do a follow up PR), ID collision is not an issue.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -1,31 +1,56 @@
...
+import faker from 'faker';
+import { dataUiId, createUserAndThread, encodeQueryData, getNameInitials } from '../utils';
 import { startServer, stopServer } from './app/server';
 import { CHAT_UI_IDS } from '../../../src';
-import { COMPONENT_UI_IDS } from '../../../../react-components';
+import { COMPONENT_UI_IDS } from '@internal/react-components';
+
+dotenv.config({ path: path.join(__dirname, '..', '.env') });
+
+jest.setTimeout(30000);
```
prprabhu-ms, line 14: For folks like me, can you add a commet here saying how many minutes / hours this is?

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -34,18 +59,58 @@ describe('Chat Composite Puppeteer Tests', () => {
...
-    await page.waitForSelector(dataUiId(COMPONENT_UI_IDS.sendboxTextfield), { visible: true });
-    await page.type(dataUiId(COMPONENT_UI_IDS.sendboxTextfield), 'HELLO WORLD');
-    await page.keyboard.press('Enter');
-    const inner_html = await page.$$eval('.ui-chat__message__content', (nodes) => {
+    await pages[0].bringToFront();
+    await pages[0].waitForSelector(dataUiId(COMPONENT_UI_IDS.sendboxTextfield));
+    await pages[0].type(dataUiId(COMPONENT_UI_IDS.sendboxTextfield), 'How the turn tables');
+    await pages[0].keyboard.press('Enter');
+    await pages[0].waitForSelector(`[data-ui-status="delivered"]`);
+    const inner_html = await pages[0].$$eval(`[data-ui-status="delivered"]`, (nodes) => {
```
prprabhu-ms, line 74: I'm a bit wary of adding attributes beyond IDs so that the tests can target these elements. Could you (perhaps IRL) quickly explain why we needed this attribute?

packages/react-composites/jest.config.js
prprabhu-ms: The daydreaming idea isn't to implement stubs, but a fake -- a full implementation of the service backend, but hermetic (e.g, for chat this shouldn't be difficult -- just maintaining a few threads and users and sending notifications etc. in memory).

It makes sense for the service teams to maintain a fully-local fake implementation to be used for other services to run integration tests against. So really, the Calling team should maintain their fake etc.

But all this is highly tangential here.

prprabhu-ms:
I think it's OK to land this and then iterate.

Thanks for writing this. It's tickling exactly the kind of questions we want. One such observation: I clearly see the value in being able to load the composite in a browser and automating interaction with it (like sending a message). I am unsure about the value in the targeted assertions in these tests (e.g. check that one of the participants has the initials of the local participant) because targeting those elements requires pulling out HTML elements by CSS class names from Fluent / NorthStar. Those CSS class names aren't in our control, and at the least, it makes it fairly hard to write and maintain these tests.

How would this approach compare with: Use Playwright to interact with the composite, and get it in and end state we want. Then just do a snapshot test.

-- 

As for ID injection, it's totally fine to do it in a follow up PR, but I'd recommend doing it soon so that we don't end up with a lot of IDs that we need to migrate.

prprabhu-ms (approved):

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
JamesBurnside: What's the purpose of this? (Is it an arbitrary timeout to wait until the server etc has booted up? If so this both adds arbitrary extra time to tests and adds non-determinism where if the server takes longer our tests will fail)

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -1,31 +1,56 @@
...
+
 let browser: Browser;
-let page: Page;
+const pages: Array<Page> = [];
+const participants: Array<string> = [];
 
-describe('Chat Composite Puppeteer Tests', () => {
+describe('Chat Composite E2E Tests', () => {
   beforeAll(async () => {
     await startServer();
```
JamesBurnside, line 34: This file will need to be split up across different files if we are to add lots more tests, will we need to call startServer in each file? If so can this be moved to a global setup (perhaps: https://jestjs.io/docs/configuration#globalsetup-string?)

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -34,18 +59,58 @@ describe('Chat Composite Puppeteer Tests', () => {
...
     });
-    expect(inner_html).toEqual('<div>HELLO WORLD</div>');
+    expect(inner_html).toEqual('How the turn tables');
+  });
+
+  test('can view myself in participant list', async () => {
+    await pages[0].bringToFront();
+    await pages[0].waitForSelector(`${dataUiId(COMPONENT_UI_IDS.participantList)} .ms-Persona`);
+    const initials = await pages[0].$$eval(
+      `${dataUiId(COMPONENT_UI_IDS.participantList)} .ms-Persona .ms-Persona-initials span`,
```
JamesBurnside, line 84: What are the class names needed here for? is the data-ui-id not enough?

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
prprabhu-ms: Heh, I assumed this was an overall timeout for the test suite in case something got stuck.

Yep: https://jestjs.io/docs/jest-object#jestsettimeouttimeout

And the docs say this is in milliseconds. So the timeout here is 30 seconds, up from the default of 5 seconds. My guess is that tstarting the server takes longer than 5 seconds, hence the call.

But all this should be in a comment above the `setTimeout` call.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
prprabhu-ms: I dislike magic.

I like the fact that, right here in the test, it says what happens before the tests run. So my recommendation would be: do nothing right now. When we split the tests into multiple files, extract a function into a helper module, that still gets called in `beforeAll` block in each test suite.

anjulgarg: Thanks for the call out. Let's refactor when we reach that point.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: The actual initials rest in a `span` element which needs to be accessed like so. It's a deeply nested element.

prprabhu-ms: My guess is that this is trying to extract the string content of a component we get from react-northstar package.

I think this makes these tests y harder to write and maintain, and potentially brittle.

anjulgarg:
> I think it's OK to land this and then iterate.
> 
> Thanks for writing this. It's tickling exactly the kind of questions we want. One such observation: I clearly see the value in being able to load the composite in a browser and automating interaction with it (like sending a message). I am unsure about the value in the targeted assertions in these tests (e.g. check that one of the participants has the initials of the local participant) because targeting those elements requires pulling out HTML elements by CSS class names from Fluent / NorthStar. Those CSS class names aren't in our control, and at the least, it makes it fairly hard to write and maintain these tests.
> 
> Also, these assertions leave with somewhat low confidence that the UI looks good. James brought up the point earlier of preferring jest unit-tests where we can over e2e tests for speed etc. In line with that, the real reason we want the e2e tests is to get coverage for the final UI, right?
> 
> How would this approach compare with: Use Playwright to interact with the composite, and get it to an end state we want. Then just do a snapshot test.
> 
> --
> 
> As for ID injection, it's totally fine to do it in a follow up PR, but I'd recommend doing it soon so that we don't end up with a lot of IDs that we need to migrate.

I agree with the low confidence after this tests. Playwright provides visual comparison tests using pixelmatch. Perhaps we should leverage those along with automation to ensure that our UI is working/looking as expected. I will try writing a test like that in another PR.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: The good thing about `data-` attributes in html is that they don't affect how the UI renders and are only used for custom purposes.

I have added a logic in message content which changes its data attribute value to its delivery status allowing us to hook into all the delivered messages in a browser.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
JamesBurnside: It would be great if splitting into separate files could be a fast follow to this PR! Folks writing tests shouldn't have to figure the part of how to split it up as new files and as it stands that file is gonna get real huge real fast (with only 4 tests its already over 100 lines long)

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
JamesBurnside: From @anjulgarg your other comments yeah lets address this in a separate PR

JamesBurnside (approved):

mgamis-msft (approved):

mgamis-msft:
> I think it's OK to land this and then iterate.
> 
> Thanks for writing this. It's tickling exactly the kind of questions we want. One such observation: I clearly see the value in being able to load the composite in a browser and automating interaction with it (like sending a message). I am unsure about the value in the targeted assertions in these tests (e.g. check that one of the participants has the initials of the local participant) because targeting those elements requires pulling out HTML elements by CSS class names from Fluent / NorthStar. Those CSS class names aren't in our control, and at the least, it makes it fairly hard to write and maintain these tests.
> 
> Also, these assertions leave with somewhat low confidence that the UI looks good. James brought up the point earlier of preferring jest unit-tests where we can over e2e tests for speed etc. In line with that, the real reason we want the e2e tests is to get coverage for the final UI, right?
> 
> How would this approach compare with: Use Playwright to interact with the composite, and get it to an end state we want. Then just do a snapshot test.
> 
> --
> 
> As for ID injection, it's totally fine to do it in a follow up PR, but I'd recommend doing it soon so that we don't end up with a lot of IDs that we need to migrate.

I think there is value in E2E tests regarding tests actions that call handlers like remove participant because it is an integrated test with the stateful client and the UI. We would need to check that the participant item is of that user has been removed from the UI after a certain amount of time. But as far UI tests are deeply based on HTML and CSS, we should probably rely more on unit tests for those as those are determined by component props which only the component itself needs to know about.

PR #520, opened 2021-07-02
Localized strings in SendBox and MessageStatusIndicator

mgamis-msft:
# What

Localized strings of SendBox and MessageStatusIndicator components

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2487751
https://skype.visualstudio.com/SPOOL/_workitems/edit/2484999

# How Tested

Unit tests with jest and enzyme. Also tested LocalizationProvider for chat sample.

# Process & policy checklist


packages/communication-react/review/communication-react.api.md
```diff
@@ -1256,6 +1263,7 @@ export interface SendBoxProps {
     onRenderSystemMessage?: (systemMessage: string | undefined) => React_2.ReactElement;
     onSendMessage?: (content: string) => Promise<void>;
     onTyping?: () => Promise<void>;
+    strings?: Partial<SendBoxStrings>;
```
prprabhu-ms, line 1266: Let's be consistent with how we handle optional strings. Ideally:

* Providing `strings` is always optional.
* When providing `strings`, all fields are optional.

As for how particular strings are made optional, there are two approaches:
* Use `Partial<XStrings>`
* All fields in `XStrings` are always optional.

I personally prefer the latter -- we never want the fields to be required, so the type should itself mark them all as optional.

packages/react-components/src/components/MessageStatusIndicator.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+import React from 'react';
+import { MessageStatusIndicator } from './MessageStatusIndicator';
+import Enzyme from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import { mountWithLocalization, createTestLocale } from './utils/testUtils';
+import { TooltipHost } from '@fluentui/react';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('MessageStatusIndicator should work with localization', () => {
```
prprabhu-ms, line 13: Great test.

packages/react-components/src/components/MessageStatusIndicator.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  test('Should localize tooltip text', async () => {
+    const messageStatusIndicatorStrings = {
+      seenTooltipText: Math.random().toString(),
```
prprabhu-ms, line 16: Soft recommendation: I'd recommend against using random strings. Unittests should be as deterministic as possible. The randomness here _should be_ harmless, but still.

Fuzz testing is a useful tool. But they have different use than unit-tests. If we want fuzz testing, we should set it up to run continuously outside of CI. If the fuzz test finds a bug, we should roughly convert the _input that failed_ into a deterministic unit-test so that it _always fails_, then fix that bug to make the unittest pass.

packages/react-components/src/components/MessageStatusIndicator.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+    const component = mountWithLocalization(<MessageStatusIndicator status="sending" />, testLocale);
+    expect(component.find(TooltipHost).props().content).toBe(messageStatusIndicatorStrings.sendingTooltipText);
+    component.setProps({ status: 'delivered' });
+    expect(component.find(TooltipHost).props().content).toBe(messageStatusIndicatorStrings.deliveredTooltipText);
+    component.setProps({ status: 'seen' });
+    expect(component.find(TooltipHost).props().content).toBe(messageStatusIndicatorStrings.seenTooltipText);
+    component.setProps({ status: 'failed' });
+    expect(component.find(TooltipHost).props().content).toBe(messageStatusIndicatorStrings.failedToSendTooltipText);
+  });
+});
```
prprabhu-ms, line 31: If we don't already have these: Add a test to check that strings supplied via props override those from the locale.

This behaviour should also be documented in the props and in Locale types.

packages/react-components/src/localization/translated/ar.json
```diff
@@ -10,6 +10,15 @@
     "multipleUsersAbbreviateMany": "... وآخرون يكتبون {users}",
     "delimiter": "و "
   },
+  "sendBox": {
```
prprabhu-ms, line 13: How are you getting these translations?

mgamis-msft: These are just from Google translate. These are just for testing the look and feel of an RTL language that does not use the Roman alphabet. This helped me for TypingIndicator. The proper translations will be provided by TD build.

packages/communication-react/review/communication-react.api.md
mgamis-msft: Thanks for pointing this out. The reason why I chose the former which is something like:
``` const interface XProps {
    ...
    strings?: Partial<XStrings>
}
```
so that all string fields are require when defining `ComponentStrings` because it is defined as::
```
export interface ComponentStrings {
    x: XStrings,
    y: YStrings
}
```
The above enforces me to write complete locales.
For example, if I delete string for a component from my locale file like this:
image
I will get the error:
image
and if I just delete one field from a component like this one:
image
I will get this error:
image

But maybe we want developers to provide partial overrides and defaulting the rest to en-US strings. In this case, I agree that all fields in `XStrings` should be optional.

packages/react-components/src/components/MessageStatusIndicator.test.tsx
mgamis-msft: My intention was to test that ANY string defined in `XString` can be localized and so much to provide invalid, unexpected, or random data as fuzz testing is defined to be. But I don't mind changing the randomness of the unit tests. DO you prefer hardcoded strings to test with?

packages/communication-react/review/communication-react.api.md
prprabhu-ms: iiuc,
* `useLocale` must always provide all strings
* Composite pros or Component props can provide some strings and skip the rest.

This sounds like the right approach to me, and the way you implement that via `Partial` and required fields also looks great. Let's make this consistent (I saw a mix of approaches in the implementation which led me to the initial comment) and document these requiremets in the storybook documentation for localization.

packages/react-components/src/components/MessageStatusIndicator.test.tsx
prprabhu-ms: It's very easy to ignore unittest failure in CI if it fails only some of the time. So yes, deterministic unittests are preferred.

prprabhu-ms (approved):

alkwa-msft (approved):

PR #521, opened 2021-07-02
When an operation succeeds, clear related errors from state

prprabhu-ms:
# What

# Why
See `ReportingErrors.md` for explanation.

# How Tested
unittests

# Process & policy checklist

### DIFF
--- a/packages/chat-stateful-client/src/StatefulChatClient.ts
+++ b/packages/chat-stateful-client/src/StatefulChatClient.ts
@@ -65,24 +65,34 @@ const proxyChatClient: ProxyHandler<ChatClient> = {
       }
       case 'startRealtimeNotifications': {
         return async function (...args: Parameters<ChatClient['startRealtimeNotifications']>) {
-          return context.asyncTeeErrorToState(async () => {
-            const ret = await chatClient.startRealtimeNotifications(...args);
-            if (!receiver.eventSubscriber) {
-              receiver.eventSubscriber = new EventSubscriber(chatClient, context);
-            }
-            return ret;
-          }, 'ChatClient.startRealtimeNotifications');
+          return context.asyncTeeErrorToState(
+            async () => {
+              const ret = await chatClient.startRealtimeNotifications(...args);
+              if (!receiver.eventSubscriber) {
+                receiver.eventSubscriber = new EventSubscriber(chatClient, context);
+              }
+              return ret;
+            },
+            'ChatClient.startRealtimeNotifications',
+            ['ChatClient.startRealtimeNotifications', 'ChatClient.stopRealtimeNotifications']
+          );
         };
       }
       case 'stopRealtimeNotifications': {
         return async function (...args: Parameters<ChatClient['stopRealtimeNotifications']>) {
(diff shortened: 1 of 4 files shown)


packages/chat-stateful-client/src/StatefulChatClient.ts
```diff
@@ -65,24 +65,34 @@ const proxyChatClient: ProxyHandler<ChatClient> = {
...
+              const ret = await chatClient.stopRealtimeNotifications(...args);
+              if (receiver.eventSubscriber) {
+                receiver.eventSubscriber.unsubscribe();
+                receiver.eventSubscriber = undefined;
+              }
+
+              return ret;
+            },
+            'ChatClient.stopRealtimeNotifications',
+            ['ChatClient.stopRealtimeNotifications', 'ChatClient.startRealtimeNotifications']
```
alkwa-msft, line 94: does this mean for every type of function as we implement it we need to be aware of all of the other functions in that category? i am curious if we could somehow do something by looking at the namespace part of the category.. in this case "ChatClient" and clear out any other errors with that namespace if we are successful?

prprabhu-ms: I don't expect us to clear errors for all methods when a method succeeds.
e.g., when `startRealtimeNotifications` succeeds, it should not clear errors for `listChatThreads` etc.

So, my plan is to just list these out for now. Since this is internal implementation, it's easy to change. If we find that a lot of methods end up with long lists of targets to clear, we can introduce an argument `clearAllMethods` as you suggest.

alkwa-msft (approved):

packages/chat-stateful-client/src/StatefulChatClient.ts
JamesBurnside: If I'm writing a new function, is there a place I can check to see what errors I might need to clear? (or add this function's error to be cleared to others)

JamesBurnside: Maybe a better question: what's the process for writing a new function regarding this error handling?

JamesBurnside (approved):
LGTM! Just concerned folks adding a new function won't have a good grasp on what errors they will need to clear, and if there function error would need cleared by other methods

prprabhu-ms:
> LGTM! Just concerned folks adding a new function won't have a good grasp on what errors they will need to clear, and if there function error would need cleared by other methods

This is a risk, but I don't (currently) see a way to solve the problem - There are sets of related methods where we want to clear errors on success. The assumption is that method relationship is independent of the use-case. So it's useful for the stateful client, rather than the Contoso app, to encode this relationship by clearing errors on related methods. So we keep the challenge / complexity of knowing when to clear errors an implementation detail of the stateful clients. To keep things sane:

* The default is to only clear the errors on the method that succeeded.
* We should continue to be conservative in what methods we consider related -- only clear methods on a small number of methods which are closely related.
* If, in the future, we think we need more complex logic to decide when to clear errors, we can add complexity to the implementation without breaking the API.

PR #522, opened 2021-07-03
Extract test helpers module

prprabhu-ms:
# What
Refactor some test helpers in their own module

# Why
* Before this PR, helpers were imported from one test module to another. This import confuses `jest` and it ends up running the tests from the imported module twice. This makes the tests run longer / adds confusion on failure.

The mocks / helpers here could be cleaned up quite a bit, but restricting myself (mostly) to just moving the helpers for now.

# How Tested
`rush test`

# Process & policy checklist


prprabhu-ms:
Please look at https://github.com/Azure/communication-ui-library/pull/523 instead.

PR #523, opened 2021-07-03
Extract test helpers module

prprabhu-ms:
# What
Refactor some test helpers in their own module

# Why
* Before this PR, helpers were imported from one test module to another. This import confuses jest and it ends up running the tests from the imported module twice. This makes the tests run longer / adds confusion on failure.
* The mocks / helpers here could be cleaned up quite a bit, but restricting myself (mostly) to just moving the helpers for now.

# How Tested
`rush test`

# Process & policy checklist

### DIFF
--- a/change/chat-stateful-client-081fed3a-3527-4bfa-b623-7da85ea465f4.json
+++ b/change/chat-stateful-client-081fed3a-3527-4bfa-b623-7da85ea465f4.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Extract test helpers",
+  "packageName": "chat-stateful-client",
+  "email": "prprabhu@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/chat-stateful-client/src/TestHelpers.ts
+++ b/packages/chat-stateful-client/src/TestHelpers.ts
@@ -1,0 +1,110 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { ChatClient, ChatThreadItem } from '@azure/communication-chat';
+import { ChatClientState } from './ChatClientState';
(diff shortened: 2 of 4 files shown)


packages/chat-stateful-client/src/TestHelpers.ts
```diff
@@ -0,0 +1,110 @@
...
+  return createMockIterator(mockChatThreads);
+};
+
+// [1, 2 ... 5] array
+const seedArray = Array.from(Array(5).keys());
+
+export const mockChatThreads: ChatThreadItem[] = seedArray.map((seed) => ({
+  id: 'chatThreadId' + seed,
+  topic: 'topic' + seed,
+  createdOn: new Date(seed * 10000),
```
alkwa-msft, line 104: is this 10000 just an arbitrary value to make a larger number?  is there any comment we could make here for the 10000?

alkwa-msft (approved):

change/chat-stateful-client-081fed3a-3527-4bfa-b623-7da85ea465f4.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
JamesBurnside, line 2: nit (**that doesn't need fixed**): I'd mark this as a `none` in future, I don't think its necessary for the changelog (always happy to be argued wrong though :D )

JamesBurnside (approved):

packages/chat-stateful-client/src/TestHelpers.ts
prprabhu-ms: Dunno. This PR is a cut-pasta :D

change/chat-stateful-client-081fed3a-3527-4bfa-b623-7da85ea465f4.json
prprabhu-ms: Makes perfect sense. I wanted to skip changelog, but didn't realize I could choose none.

PR #527, opened 2021-07-05
Publish to chromatic from beta candidate branch

prprabhu-ms:
# Why
Storybook deployment for bug bashing.

# How Tested
Prayers.

# Process & policy checklist

### DIFF
--- a/.github/workflows/publish-chromatic.yml
+++ b/.github/workflows/publish-chromatic.yml
@@ -9,7 +9,7 @@ on:
       - packages/react-composites/**
       - packages/communication-react/**
   push:
-    branches: [main]
+    branches: [main, beta-candidate]
   workflow_dispatch:
 
 jobs:


prprabhu-ms:
The idea is to land this, and then cherry-pick the PR to `beta-candidate`. If all goes well, merging the cherry-pick PR into `beta-candidate` will trigger a `publish-chromatic` workflow and give us a storybook to play with.

.github/workflows/publish-chromatic.yml
```diff
@@ -9,7 +9,7 @@ on:
       - packages/react-composites/**
       - packages/communication-react/**
   push:
-    branches: [main]
+    branches: [main, beta-candidate]
```
JamesBurnside, line 12: my vote here would be we just use `release`
So this release branch would be: `release/1.0.0-beta.2`
Thoughts?

This, with small update to set the correct branchname and overwrite an existing branch with that name in the case we want to choose a different alpha candidate, would then be all we need to run to create new beta candidates: https://github.com/Azure/communication-ui-library/blob/main/.github/workflows/npm-release-bump.yml

prprabhu-ms: There are two workflows to consider:
* We chose a new alpha tag to create a new beta candidate off of
* We merge changes onto a beta candidate branch to fix critical bugs.

We want to automate setup for bug-bashing in both these cases.

prprabhu-ms:
Quick experiment I didn't follow up on.

PR #530, opened 2021-07-06
Fix components to respond to RTL

mgamis-msft:
# What

Fix our ACS components to respond to RTL
Components addressed:
- FluentThemeProvider
- MessageThread
- ControlBar
TypingIndicator always ltr because strings in strings files should already be in the right direction

Added rtl toolbar button in storybook.
Video of rtl switching in storybook:
https://user-images.githubusercontent.com/79475487/124593650-bc4f7180-de13-11eb-926e-3cbd1be4c9bd.mp4

Will document how to set RTL as a separate ticket.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2506615

# How Tested

Manual testing

# Process & policy checklist


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -270,6 +270,13 @@ const DefaultChatMessageRenderer: DefaultMessageRendererType = (props: MessagePr
               : formatTimeForChatMessage(payload.createdOn)
             : undefined
         }
+        style={
+          getRTL()
+            ? payload.mine
```
JamesBurnside, line 275: Is this change due to the same northstar issue you mentioned below?
```
 // This is a bug in fluentui react northstar not reversing left and right margins for the author
// in the message bubble
```

If so can you add a similar comment here (as ideally we shouldn't need to have conditional margins styles)

packages/storybook/.storybook/preview.tsx
```diff
@@ -129,5 +134,17 @@ export const globalTypes = {
         { value: 'ar', title: 'Arabic' }
       ],
     },
+  },
+  rtl: {
```
JamesBurnside, line 138: This is awesome thanks for adding this! So great for testing!

JamesBurnside (approved):

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -260,5 +260,9 @@ export const TypingIndicator = (props: TypingIndicatorProps): JSX.Element => {
     styles
   );
 
-  return <Stack className={mergeStyles(typingIndicatorContainerStyle, styles?.root)}>{indicatorComponent}</Stack>;
+  return (
+    <Stack dir="ltr" className={mergeStyles(typingIndicatorContainerStyle, styles?.root)}>
```
JamesBurnside, line 264: How come we need to hard code ltr here?

JamesBurnside: Does it flip the string if we don't? (I'd assume rtl to be the same as ltr for this component)

JamesBurnside:
Note: found some small bugs (not ours or not blocking for this PR though):
1. send button icon should be mirrored:
    image
  Fluent is supposed to automatically flip this so it could be a bug in the react-icons package (https://developer.microsoft.com/en-us/fluentui#/styles/web/localization)
1. Video tile name when too long cuts off the start of the name (in LTR it cuts off the end of the name)
    image
  This should probably be fixed in a totally seperate item where we use ellipses or something for a long name
1. Should punctuation be moved to the start of the text in messages? 
    image
  Update: looks to be a bug in the northstar package, they're very inconsistent on when it does and doesn't (https://fluentsite.z22.web.core.windows.net/0.56.0/components/chat/definition):
    image

packages/react-components/src/components/TypingIndicator.tsx
alkwa-msft: Don't we want the layout of the typing indicator to come in from the right if its rtl?

packages/react-components/src/components/ControlBar.tsx
```diff
@@ -54,23 +54,23 @@ export interface ControlBarProps {
 export const ControlBar = (props: ControlBarProps): JSX.Element => {
   const { styles, layout } = props;
   const theme = useTheme();
-  const controlBarStyle = controlBarStyles[layout ?? 'horizontal'];
+  const rtl = getRTL();
+  const controlBarClassName = useMemo(() => {
```
prprabhu-ms, line 58: `useMemo` is unnecessary here because `mergeStyles` returns a string (the className of registered CSS class).

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -317,7 +326,7 @@ const memoizeAllMessages = memoizeFnAll(
         ) : (
           <Persona text={payload.senderDisplayName} hidePersonaDetails={true} size={PersonaSize.size32} />
         ),
-        contentPosition: payload.mine ? 'end' : 'start',
+        contentPosition: (getRTL() ? !payload.mine : payload.mine) ? 'end' : 'start',
```
prprabhu-ms, line 329: I'd extract this into a private helper function `xor` to increase readability.

prprabhu-ms (approved):

packages/react-components/src/components/ControlBar.tsx
anjulgarg: This is valid use I think because this will prevent mergeStyles from running again on component update unless a dependency changes. `mergeStyles` is an expensive method.

anjulgarg (approved):

PR #531, opened 2021-07-06
Tee errors from `listChatThreads` to state

prprabhu-ms:
API methods that reuturn async iterators are a bit more complex than other methods. We need to handle immediate errors as well errors when iterating, since there could be a REST call at either point.

# How Tested
`rush test`

# Process & policy checklist


packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -357,6 +357,36 @@ export class ChatContext {
...
+   *
+   * If the function succeeds, clears associated errors from the state.
+   *
+   * @param f Function to execute.
+   * @param target The error target to tee error to.
+   * @param clearTargets The error targets to clear errors for if the function succeeds. By default, clears errors for `target`.
+   * @returns Result of calling `f`. Also re-raises any exceptions thrown from `f`.
+   * @throws ChatError. Exceptions thrown from `f` are tagged with the failed `target.
+   */
+  public withErrorTeedToState<T>(f: () => T, target: ChatErrorTargets, clearTargets?: ChatErrorTargets[]): () => T {
```
prprabhu-ms, line 371: * The implementation here is very similar to `asyncTeeErrorToState`. I couldn't figure out a sane way to dedupe. Ideas welcome.
* Unlike `asyncTeeErrorToState`, it was more ergonomic to write this as a higher order function. In a future PR, I'll convert `asyncTeeErrorToState` to also be a higher order function, to make the two consistent.

packages/chat-stateful-client/src/iterators/createDecoratedIterator.ts
```diff
@@ -56,3 +57,41 @@ export const createDecoratedIterator = <ItemType, OptionsType>(
...
+ * @param target See {@Link ChatContext.asyncTeeErrorToState}.
+ * @param clearTargets See {@Link ChatContext.asyncTeeErrorToState}.
+ * @returns A function to create an iterator that handles errors when iterting over the iterator from `iteratorCreator`.
+ */
+export const createErrorHandlingIterator = <ItemType, OptionsType>(
+  iteratorCreator: IteratorCreatorFn<ItemType, OptionsType>,
+  context: ChatContext,
+  target: ChatErrorTargets,
+  clearTargets?: ChatErrorTargets[]
+): IteratorCreatorFn<ItemType, OptionsType> => {
```
PorterNan, line 74: Glad that we have an generic wrapper for all those iterators!

packages/chat-stateful-client/src/iterators/createDecoratedIterator.ts
```diff
@@ -4,6 +4,7 @@
 import { ListPageSettings } from '@azure/communication-chat';
 import { ChatContext } from '../ChatContext';
 import { PagedAsyncIterableIterator } from '@azure/core-paging';
+import { ChatErrorTargets } from '../ChatClientState';
 
 type IteratorCreatorFn<T, OptionsType, PageT = T[]> = (options?: OptionsType) => PagedAsyncIterableIterator<T, PageT>;
```
PorterNan, line 10: Sry, this request is unrelated, but I just noticed in line 24/25, I named the iterator as threadsIterator, which I believe is not generic enough because it refers not only to thread now, could you please help me make a change - either iterator or innerIterator would be good for this case

packages/chat-stateful-client/src/iterators/createDecoratedListThreads.ts
```diff
@@ -18,5 +18,14 @@ export const createDecoratedListThreads = (
       context.updateThread(chatThreadItem.id, properties);
     }
   };
-  return createDecoratedIterator(chatClient.listChatThreads.bind(chatClient), context, setThreadProperties);
+
+  return createDecoratedIterator(
+    createErrorHandlingIterator(
+      context.withErrorTeedToState(chatClient.listChatThreads.bind(chatClient), 'ChatClient.listChatThreads'),
```
PorterNan, line 24: 'ChatClient.listChatThreads' will this string be logged in the Error somewhere, or it is just a type guard?
If that is logged, one concerns for this string is, if someone is using an invasive minifier, it might change the name of class/functions in the runtime, which might mismatch other auto generated strings in last PR

PorterNan (approved):
Love this iterator wrapping function, reusable, and will save a lot of code for other iterator in our repo!

packages/chat-stateful-client/src/iterators/createDecoratedListThreads.ts
prprabhu-ms: The string isn't logged. It's the key used to store the error in the `latestErrors` field in the state.

Minification / uglification  shouldn't be a problem here. What was the runtime mismatch you were thinking of?
* We get type safety because the strings used here are verified against a string literal union type.
* At runtime, we never look at method names, the keys in the `latestErrors` are always literal strings in the code, like here.

JamesBurnside (approved):

packages/chat-stateful-client/src/iterators/createDecoratedIterator.ts
prprabhu-ms: Will do at the top of the stack.

PR #533, opened 2021-07-06
Call composite mobile responsive fixes

JamesBurnside:
# What
A set of call composite responsive fixes (responsive wrt to small vs large screens)

## Does this mean call composites works beautifully on mobile now??
No.. Its better! But still a lot of polish to go.

# Why
Improvements for mobile usage

# How Tested
Locally - using responsive emulator

# Changes:
| Before | After |
| -- | -- |
| image | image
| image | image 
| image | image 
| image | image |


prprabhu-ms (approved):
Less is More.

packages/react-composites/src/composites/CallComposite/Call.tsx
```diff
@@ -101,25 +86,8 @@ export const Call = (props: CallCompositeProps): JSX.Element => {
   return (
     <FluentThemeProvider fluentTheme={fluentTheme}>
       <CallAdapterProvider adapter={adapter}>
-        <Stack className={callContainerStyle(theme)} grow>
```
mgamis-msft, line 104: Nice cleanup here

packages/react-composites/src/composites/CallComposite/ConfigurationScreen.tsx
```diff
@@ -30,19 +31,22 @@ export const ConfigurationScreen = (props: ConfigurationScreenProps): JSX.Elemen
   const { video: cameraPermissionGranted, audio: microphonePermissionGranted } = useSelector(devicePermissionSelector);
 
   return (
-    <CallConfiguration {...props}>
```
mgamis-msft, line 33: I'm glad this "component" I created way back when is gone now

mgamis-msft (approved):

PR #535, opened 2021-07-06
Security fix: Prefix internal packages with `@internal`

JamesBurnside:
# What
Prefix internal packages names with `@internal`

## How
Find and replace where package name did not start with `/`. Builds and runs fine but with so many files easily could have miss named something.

# Why
Security concern. If someone published packages under our packlets names with malicious code then if a user did npm install instead of rush install they would be fed potentially malicious code from the unscrupulous npm package.

AFAIK `@internal` cannot be used as an org name publicly.

# How Tested
Ran sampleslocally and quickly checked storybook url in PR.


prprabhu-ms (approved):
Oh, I totally reviewed each line of this PR.

alkwa-msft (approved):

PR #551, opened 2021-07-08
Jinan/update use selector

PorterNan:
# What

Update useSelector() doc
Add target parameter to useSelector()
# Why

# How Tested

# Process & policy checklist


packages/storybook/stories/Stateful Client/UseSelector.stories.mdx
```diff
@@ -32,6 +31,42 @@ const ChatInfo = ({id: string, userName: string}) => {
 <ChatInfo {...chatInfoProps} />
 ```
 
+### Passing your own props
+
+When creating your own selector, you can pass in your own parameters into selector by using second parameter of selector function:
```
alkwa-msft, line 36: When creating your own selector, you can inject your own custom properties through the second parameter of the selector function.

packages/communication-react/src/mergedHooks.ts
```diff
@@ -26,10 +26,13 @@ export type Selector = (state: ClientState, props: any) => any;
...
   selector: ParamT,
-  selectorProps?: ParamT extends Selector ? Parameters<ParamT>[1] : undefined
+  selectorProps?: ParamT extends Selector ? Parameters<ParamT>[1] : undefined,
+  type?: 'calling' | 'chat'
 ): ParamT extends Selector ? ReturnType<ParamT> : undefined => {
-  const callProps = useCallingSelector(selector as any, selectorProps);
-  const chatProps = useChatSelector(selector as any, selectorProps);
+  const callingMode = !type || type === 'calling';
+  const chatMode = !type || type === 'chat';
+  const callProps = useCallingSelector(callingMode ? (selector as any) : undefined, selectorProps);
```
prprabhu-ms, line 34: Why do we need to cast to `any` here?

prprabhu-ms (approved):

packages/communication-react/src/mergedHooks.ts
PorterNan: Cause selector won't always meet the type of those 2 different calling/chat selector, but we do need to walk through these 2 selectors(and no way to check their type in the runtime)

alkwa-msft (approved):

PR #449, opened 2021-06-09
Hook up error listeners for Chat and Calling adapters

allenplusplus:
# What
Hook up error listeners for Chat and Calling adapters.

# Why
It should address this bug where HTTP error screen is shown when token/thread is incorrect: https://skype.visualstudio.com/SPOOL/_workitems/edit/2457313.

# How Tested
Tested via unit test and manually check composites on storybook.

# Process & policy checklist

- [X] I have updated the project documentation to reflect my changes if necessary.
- [X] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.

**Is this a breaking change?**

- [ ] This change causes current functionality to break.

PR #552, opened 2021-07-08
Fix adapter function always needs a bind

PorterNan:
Update adapter to arrow function
Remove CallAdapter from public api

# What

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/communication-react/review/communication-react.api.md
+++ b/packages/communication-react/review/communication-react.api.md
@@ -38,7 +38,7 @@ import { MicrosoftTeamsUserKind } from '@azure/communication-common';
 import { OutputParametricSelector } from 'reselect';
 import { OutputSelector } from 'reselect';
 import { PartialTheme } from '@fluentui/react';
-import { PermissionConstraints } from '@azure/communication-calling';
+import type { PermissionConstraints } from '@azure/communication-calling';
 import { PersonaPresence } from '@fluentui/react';
 import { PhoneNumberIdentifier } from '@azure/communication-common';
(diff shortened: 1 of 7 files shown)


alkwa-msft:
Can we link the bug/issue that this change is related to in order to build some context?

packages/communication-react/review/communication-react.api.md
```diff
@@ -66,99 +66,6 @@ export type AreParamEqual<A extends (props: any) => JSX.Element | undefined, B e
 // @public (undocumented)
 export type AreTypeEqual<A, B> = A extends B ? (B extends A ? true : false) : false;
 
-// @public (undocumented)
-export class AzureCommunicationCallAdapter implements CallAdapter {
```
JamesBurnside, line 70: unintentionally removed?

JamesBurnside, line 70: nice!

JamesBurnside (approved):

prprabhu-ms (changes requested):
I don't think we should do this.

See https://github.com/Azure/communication-ui-library/pull/449#discussion_r652949367

Arrow functions create a new function object per instance of the class and are an anti-pattern.
Is it surprising to users that they need to bind methods to the object if they're using them as callback. That is standard JavaScript behavior, right?

anjulgarg (approved):

alcail (approved):

alkwa-msft:
@PorterNan  can we close this since we did the binding as part of prathmesh's change?

PorterNan:
Will open another one to remove some apis from our public api files

PR #556, opened 2021-07-08
Indentifier Injection POC for Composites/Components

anjulgarg:
# What
Provides a way for contoso to inject custom identifiers in components.
Enabling anyone to modify those elements or hook into them for things like browser based testing.

# Why
Browser based automated tests need such unique identifiers to hook into the UI elements

# How Tested

# Process & policy checklist


packages/react-components/review/react-components.api.md
```diff
@@ -123,6 +123,9 @@ export type CustomMessagePayload = {
 // @public
 export const darkTheme: PartialTheme & CallingTheme;
 
+// @public (undocumented)
+export const defaultIdentifiers: Identifiers;
```
prprabhu-ms, line 127: Let's not export these. These are only required for tests, and should be defined in the tests directly.

By exporting we're adding them to API contract, for no value.

packages/react-components/review/react-components.api.md
```diff
@@ -165,6 +168,30 @@ export interface GridLayoutProps {
 // @public (undocumented)
 export type GridLayoutType = 'standard';
 
+// @public (undocumented)
+export const IdentifierContext: React_2.Context<Identifiers>;
+
+// @public (undocumented)
+export const IdentifierProvider: (props: IdentifierProviderProps) => JSX.Element;
```
prprabhu-ms, line 175: I'm fully onboard with:
* Component props for injecting IDs.
* Composite prop to inject IDs for all used components (similar to the `strings` prop).

I'm not yet convinced about the need to add a `Provider` for these though. Unlike internationalization, there is no need for Contoso to add identifiers exhaustively for all components, so let's just stick to the props for now?

For implementation in the Composite, you can define the `IdentifierProvider` inside the composite implementation to avoid having to drill through props. Then, at the point where the composite instnatiates the components, you would have to use the provider and set props for the component.

prprabhu-ms: We can always add the `IdentifierProvider` to the API in the future if we see a business need. Revese is not possible.

packages/react-components/review/react-components.api.md
```diff
@@ -500,6 +527,9 @@ export interface TypingIndicatorStylesProps extends BaseCustomStylesProps {
     typingUserDisplayName?: IStyle;
 }
 
+// @public (undocumented)
+export const useIdentifiers: () => Identifiers;
```
prprabhu-ms, line 531: See comment about `IdentifierProvider`. This would also go away if the provider is removed from the API.

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -239,11 +241,11 @@ const generateTextMessageContent = (payload: ChatMessagePayload): JSX.Element =>
 const generateMessageContent = (payload: ChatMessagePayload): JSX.Element => {
```
prprabhu-ms, line 241: Capitalize to make its own component.

I'm wary of non-component functions returning components. It's very easy to (and, in fact, we have this bug at ToT right now in MessageThread already) end up using a function like this inside a `useMemo`. eslint is unable to detect this, but it cause undefined behaviour at runtime.

packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -102,6 +102,7 @@ const onRenderParticipantsDefault = (
  */
 export const ParticipantList = (props: ParticipantListProps): JSX.Element => {
   const { excludeMe = false, myUserId, participants, onParticipantRemove, onRenderAvatar, onRenderParticipant } = props;
+  const ids = useIdentifiers();
```
prprabhu-ms, line 105: Does this mean that all uses of `ParticipantList` *must* be wrapped with an `IdentifierProvider` now?

We want to avoid that. Providers are a nifty tool to avoid drilling through props. But they shouldn't replace props in the component definition.

This is related to my comment about removing the `IdentifierProvider` from the API.

packages/react-composites/tests/browser/config.ts
```diff
@@ -22,3 +22,9 @@ export const PARTICIPANT_NAMES = [
   'Adrian Leffler',
   'Schuyler Yost'
 ];
+
+export const IDS = {
```
prprabhu-ms, line 26: +1

This is the setup I was expecting. Tests define and use the identifiers they want.

<3

packages/react-components/src/components/ParticipantList.tsx
anjulgarg: Nope. Think of providers as a way to set a context. If the context is already there, this hook will return the default values even when the component is not wrapped inside provider.

packages/react-components/src/components/MessageThread.tsx
anjulgarg: I did capitalize the other ones, missed this one. Thanks for pointing out.

packages/react-components/review/react-components.api.md
anjulgarg: That won't work.
Assigning an ID to `<ParticipantList data-ui-id="someid" />` won't actually assign the element an id.
For that you need to assign it to the `<Stack />` being used inside the component. 
Which means that the only way to inject such ids into components through composites is to expose another component prop that does this for us. That increases the API surface even more...

packages/react-components/review/react-components.api.md
anjulgarg: The reason behind this approach is mentioned above.

packages/react-components/review/react-components.api.md
anjulgarg: This approach requires default values to allow components to work without being wrapped inside providers.

packages/react-components/review/react-components.api.md
mgamis-msft: I think Provider should be ok for this case because the context is not expected to change much (or at all) during run time. So rerendering due to IDS context should not be an issue.

mgamis-msft (approved):

prprabhu-ms (approved):
Was waiting for another opinion before approval.

PR #558, opened 2021-07-09
Bind all methods of CallAdapter on construction

prprabhu-ms:
# What
* Bind all `CallAdapter` interface methods in constructor
* Bind all private methods used as callbacks for notification from SDK.

# Why
* Better ergonomics for Contoso
* Correctness for notificatin from SDK.

# How Tested
Local storybook.

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -344,21 +375,21 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
   private unsubscribeCallEvents(): void {
     for (const subscriber of this.participantSubscribers.values()) {
       subscriber.unsubscribeAll();
     }
     this.participantSubscribers.clear();
-    this.call?.off('remoteParticipantsUpdated', this.onRemoteParticipantsUpdated);
-    this.call?.off('isMutedChanged', this.isMyMutedChanged);
-    this.call?.off('isScreenSharingOnChanged', this.isScreenSharingOnChanged);
-    this.call?.off('idChanged', this.callIdChanged);
+    this.call?.off('remoteParticipantsUpdated', this.onRemoteParticipantsUpdated.bind(this));
```
PorterNan, line 389: Looks like these listeners are arrow functions already, do we wanna modify it to normal function(or keep it arrow without a binding?), might be good to check chat side as well

PorterNan (approved):

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
prprabhu-ms: Converted to normal functions.

PR #560, opened 2021-07-09
Add js bundles to release

PorterNan:
# What

Add js bundles to release
It only happens for beta-release
# Why

# How Tested

# Process & policy checklist


.github/workflows/npm-release-publish.yml
```diff
@@ -71,6 +71,25 @@ jobs:
           git tag ${{ steps.version.outputs.version }}
```
JamesBurnside, line 71: Do we still need this step or can it be removed?

JamesBurnside: (oh we do if we use a different name to the version - see other comment)

.github/workflows/npm-release-publish.yml
```diff
@@ -71,6 +71,25 @@ jobs:
...
+        run: |
+          echo "::set-output name=tagname::$(TZ=":America/Vancouver" date +'%Y-%m-%d-%H%M')"
+          echo "::set-output name=releasename::$(TZ=":America/Vancouver" date +'%Y-%m-%d-%H%M')"
+
+      # Upload releases
+      - name: Upload Public Preview Release
+        uses: softprops/action-gh-release@v1
+        with:
+          name: PublicPreview ${{ steps.releasenames.outputs.releasename }}
+          tag_name: Release/${{ steps.releasenames.outputs.tagname }}
```
JamesBurnside, line 86: For the name should we just use the same as the npm version?
```suggestion
          tag_name: ${{ steps.version.outputs.version }}
```

JamesBurnside (approved):
Small comment on version name but looks great!

prprabhu-ms (approved):

alkwa-msft (approved):

.github/workflows/npm-release-publish.yml
PorterNan: Not sure if the tag_name parameter in next step is to create a new tag or just reuse tags we have, since it is working well, will check this in, and let's test it when next release

PR #562, opened 2021-07-09
Provide participants when creating chat thread

prprabhu-ms:
# What
Storybook: Use a different API method to add participants to thread.

# Why
Avoids a bug: The user that creates the thread gets added automatically to the thread, without a display name. Display name can't be changed after that.

# How Tested
Storybook.

# Process & policy checklist


alkwa-msft (approved):

JamesBurnside (approved):
Can we update this in a couple other places too:
image

PorterNan (approved):

prprabhu-ms:
> Can we update this in a couple other places too:
> image

The only other place really is the server in the `ChatComposite` stories. It just so happens, that each of those examples has only two usres -- the storybook user, and a bot. And we add the storybook user first. So the bot doesn't get to see the storybook user's name ;)

Let's fix it, but no need to merge those fixes to the beta release branch.

PR #570, opened 2021-07-12
Re-enable check treeshaking project

JamesBurnside:
# What
Re-enable check treeshaking project by allowing that project webpack version 4

# Why
This was disabled when we upgraded to webpack 5 but it has a blocking issue to upgrading.
Turns out in rush you can allow some mismatching version numbers - so allowing it for this case

# How Tested
Ran `rush build -t @internal/check-treeshaking`

### DIFF
--- a/common/config/rush/common-versions.json
+++ b/common/config/rush/common-versions.json
@@ -51,12 +51,13 @@
    * This design avoids unnecessary churn in this file.
    */
   "allowedAlternativeVersions": {
-    /**
-     * For example, allow some projects to use an older TypeScript compiler
-     * (in addition to whatever "usual" version is being used by other projects in the repo):
-     */
-    // "typescript": [
-    //   "~2.4.0"
-    // ]
+    "webpack": [
+      // All projects should use this version by default
+      "5.38.1",
+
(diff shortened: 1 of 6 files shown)


alkwa-msft (approved):

common/config/rush/common-versions.json
```diff
@@ -51,12 +51,13 @@
...
   "allowedAlternativeVersions": {
-    /**
-     * For example, allow some projects to use an older TypeScript compiler
-     * (in addition to whatever "usual" version is being used by other projects in the repo):
-     */
-    // "typescript": [
-    //   "~2.4.0"
-    // ]
+    "webpack": [
+      // All projects should use this version by default
```
prprabhu-ms, line 55: You shouldn't need this. As the comment above says, only list the "additional versions" that are allowed.

prprabhu-ms (approved):

common/config/rush/common-versions.json
JamesBurnside: It always complains if I don't put both in, I assume its a bug in how they detect the primary version but I can't see how they choose the primary version (assumed it would be the most common)

PR #571, opened 2021-07-13
Perf optimization - remove getCall from selector

PorterNan:
# What
Remove getCall from selector
- Use more specific props in the call to target the right update

_Visiting call object is causing more perf regression_
_Every update within the call will trigger the re-render_
# Why

# How Tested

Tested locally
# Process & policy checklist


packages/react-composites/src/composites/CallComposite/ScreenSharePopup.tsx
```diff
@@ -87,7 +87,9 @@ export const ScreenSharePopup = (props: ScreenSharePopupProps): JSX.Element => {
             disabled={stoppingInProgress}
           />
         </Stack>
-        <FocusTrapCallout focusTrapProps={{ isClickableOutsideFocusTrap: true }}></FocusTrapCallout>
+        <FocusTrapCallout
+          focusTrapProps={{ isClickableOutsideFocusTrap: true, forceFocusInsideTrap: false }}
```
PorterNan, line 91: This workaround is causing problem when we get the right behavior of re-rendering(which is making this workaround behaves correctly), so an additional parameter is needed to solve the bug

packages/react-composites/src/composites/CallComposite/ScreenSharePopup.tsx
```diff
@@ -69,7 +69,7 @@ export const ScreenSharePopup = (props: ScreenSharePopupProps): JSX.Element => {
   }, [onStopScreenShare]);
 
   // Beware: The FocusTrapCallout is a workaround to allow users to tab-navigate out of the Modal. There is an open
-  // issue here: https://github.com/microsoft/fluentui/issues/16179
+  // issue here: https://github.com/microsoft/fluentui/issues/18924
```
PorterNan, line 72: This issue is not resolved but stayed too long in fluentUI repo, reopened a new issue and hopefully we will raise some additional awareness from fluentui team
(There should be some better work around when we wanna make some accessibility features, like adding onBlur behavior of the stop button)

alkwa-msft (approved):

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -21,8 +28,33 @@ export const getIncomingCallsEnded = (state: CallClientState): IncomingCallState
...
-export const getCall = (state: CallClientState, props: CallingBaseSelectorProps): CallState | undefined =>
-  state.calls[props.callId];
+export const getCallExists = (state: CallClientState, props: CallingBaseSelectorProps): boolean =>
+  !!state.calls[props.callId];
+
+export const getRemoteParticipants = (
+  state: CallClientState,
+  props: CallingBaseSelectorProps
+):
+  | undefined
```
prprabhu-ms, line 38: Am I reading this wrong? Why is the type `| undefined | {....}` The first element in the union is nothing?

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -21,8 +28,33 @@ export const getIncomingCallsEnded = (state: CallClientState): IncomingCallState
...
+  | {
+      [keys: string]: RemoteParticipantState;
+    } => state.calls[props.callId]?.remoteParticipants;
+
+export const getIsScreenSharingOn = (state: CallClientState, props: CallingBaseSelectorProps): boolean =>
+  state.calls[props.callId]?.isScreenSharingOn ?? false;
```
prprabhu-ms, line 44: The base selectors should cheaply pull out values from the state. They should not contain any other business logic, in particular default values.

Instead of defaulting to `false`, I'd recommend returngin `boolean | undefined` and dealing with undefined values in the caller.

Same for `getIsMuted`.

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -21,8 +28,33 @@ export const getIncomingCallsEnded = (state: CallClientState): IncomingCallState
...
+
+export const getIsMuted = (state: CallClientState, props: CallingBaseSelectorProps): boolean =>
+  state.calls[props.callId]?.isMuted ?? true; // default isMuted should be true for privacy purpose
+
+export const getLocalVideoStreams = (
+  state: CallClientState,
+  props: CallingBaseSelectorProps
+): LocalVideoStreamState[] | undefined => state.calls[props.callId]?.localVideoStreams ?? undefined;
```
prprabhu-ms, line 52: Just `state.calls[props.callId]?.localVideoStreams` is enough.

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -21,8 +28,33 @@ export const getIncomingCallsEnded = (state: CallClientState): IncomingCallState
...
+
+export const getScreenShareRemoteParticipant = (
+  state: CallClientState,
+  props: CallingBaseSelectorProps
+): string | undefined => state.calls[props.callId]?.screenShareRemoteParticipant ?? undefined;
```
prprabhu-ms, line 57: Just `state.calls[props.callId]?.screenShareRemoteParticipant` is enough.

packages/react-composites/src/composites/CallComposite/selectors/baseSelectors.ts
```diff
@@ -15,6 +16,11 @@ export const getPage = (state: CallAdapterState): CallCompositePage => state.pag
 export const getLocalMicrophoneEnabled = (state: CallAdapterState): boolean => state.isLocalPreviewMicrophoneEnabled;
 export const getDisplayName = (state: CallAdapterState): string | undefined => state.displayName;
 export const getIdentifier = (state: CallAdapterState): string => toFlatCommunicationIdentifier(state.userId);
+export const getLocalVideoStreams = (state: CallAdapterState): LocalVideoStreamState[] | undefined =>
+  state.call?.localVideoStreams ?? undefined;
```
prprabhu-ms, line 20: Don't need to explicitly return `undefined`.

prprabhu-ms (approved):
Some stylistic comments.

Overall, looks great!

packages/calling-component-bindings/src/baseSelectors.ts
PorterNan: It is beautifier forcing me to switch the line and add another | there

(Looks like it stands for switch a line for nothing)

PR #574, opened 2021-07-13
Add an ErrorBar component to show errors

prprabhu-ms:
# What
First PR to add a minimal `ErrorBar` component:

* Added `ErrorBar` component that shows all errors from the Chat state.
* Added selectors and handlers for the new component.
* Added a story for the component.
* Used the component in the `ChatComposite`, adding in the necessary plumbing to `ChatAdapter`.

# Why
A step towards surfacing errors on the UI for ChatComposite.

# How Tested
* `rush test`
* Local storybook: ChatComposite.

# Process & policy checklist


JamesBurnside:
Knob doesn't quite match usage
image
image

^ So no way to see how that send error message was contructed. Could the knob instead take in the error string ?

JamesBurnside (approved):

prprabhu-ms:
> Knob doesn't quite match usage
> image
> image
> 
> ^ So no way to see how that send error message was contructed. Could the knob instead to choose from a selection of ErrorTypes?
> 
> Also if just one story collapse the story folder into a single entity

* The error to show is selected by picking a value from the `ErrorType` enum. Made this more obvious by directly using enum values in the options.
* Keeping the nested story because I want to soon add another canvas that shows how multiple errors might be shown.

packages/chat-stateful-client/src/StatefulChatClient.ts
```diff
@@ -13,6 +13,10 @@ export interface StatefulChatClient extends ChatClient {
   getState(): ChatClientState;
   onStateChange(handler: (state: ChatClientState) => void): void;
   offStateChange(handler: (state: ChatClientState) => void): void;
+  /**
+   * Clear all errors in the state for provided target methods.
+   */
+  clearErrors(targets: ChatErrorTargets[]): void;
```
PorterNan, line 19: Hmm.. We are adding more apis to the stateful client (ideally it should be state related), but it seems to me that it is the only way to clear error. I am thinking about something like this in stateful client:

Do we have to expose the ability for stateful user to clear the lastErrors array? Is it possible that we don't clear the last error, but contoso choose whether to show it or not in their contoso logic/composite logic(Like a boolean value named 'hidden' directly into message bar)?

@DominikMe would be cool if you have a smart idea here

packages/chat-component-bindings/src/errorBarSelector.ts
```diff
@@ -0,0 +1,27 @@
...
+ *   - `ErrorType` is never repeated in the returned errors.
+ *   - Errors are returned in a fixed order by `ErrorType`.
+ */
+export const errorBarSelector = createSelector([getLatestErrors], (latestErrors): { activeErrors: ErrorType[] } => {
+  // The order in which the errors are returned is significant: The `ErrorBar` shows errors on the UI in that order.
+  // There are several options for the ordering:
+  //   - Sorted by when the errors happened (latest first / oldest first).
+  //   - Stable sort by error type.
+  //
+  // We chose to stable sort by error type: We intend to show only a small number of errors on the UI and we do not
```
PorterNan, line 21: The question here is multiple errors vs just single errors for error bar? It looks like we make it 'ErrorsBar' or 'ErrorBars' instead of a single ErrorBar

packages/chat-component-bindings/src/errorBarSelector.ts
```diff
@@ -0,0 +1,27 @@
...
+import { createSelector } from 'reselect';
+
+/**
+ * Select active errors from the state for the `ErrorBar` component.
+ *
+ * Invariants:
+ *   - `ErrorType` is never repeated in the returned errors.
+ *   - Errors are returned in a fixed order by `ErrorType`.
+ */
+export const errorBarSelector = createSelector([getLatestErrors], (latestErrors): { activeErrors: ErrorType[] } => {
```
PorterNan, line 15: So currently errors are all coming from stateful client, any future plan of catching errors from UI side?

PorterNan (approved):
Awesome PR! 

I would like to have another around of discussion on define what a stateful client is (which could be helpful to design stateful client apis in the future), it is going a little bit more than what I thought before

packages/chat-stateful-client/src/StatefulChatClient.ts
prprabhu-ms: We need _some_ way to clear old errors. Unlike chat messages, list of participants etc, the backing services don't have a store of encountered errors, so we aren't caching information from the backing services. So the clearing mechanism can't be -- clear errors in the backing services and the stateful cache gets updated automatically.

If we didn't want to store errors from stateful methods in the stateful client, we'd need a separate store just for errors, but that would be less ergonomic.

packages/chat-component-bindings/src/errorBarSelector.ts
prprabhu-ms: Up for renaming this component on the top of the stack.

packages/chat-component-bindings/src/errorBarSelector.ts
prprabhu-ms: Not in the stateful client. We could save it "somewhere else", and the selector could merge the various sources.

packages/chat-stateful-client/src/StatefulChatClient.ts
DominikMe: Adding `clearErrors` feels wrong to me. Stateful clients should not expand on the base client API besides adding very few state specific functions.

Have we considered an `applyState(newState)` function instead? 

We can separately export helper methods like `clearErrors(state: ChatClientState): ChatClientState` which modify the state with immerJs to make it easier for the developer. That would be a similar pattern to DraftJs' `Modifier` for their Editor `ContentState`.
https://draftjs.org/docs/api-reference-modifier

packages/react-components/src/localization/translated/de.json
```diff
@@ -59,5 +59,8 @@
     "thursday": "Donnerstag",
     "friday": "Freitag",
     "saturday": "Samstag"
+  },
+  "errorBar": {
+    "sendMessageGeneric": "Fehler beim senden einer nachricht"
```
DominikMe, line 64: Upper-case "Senden" and "Nachricht" please :-)

packages/chat-stateful-client/src/StatefulChatClient.ts
prprabhu-ms: This is a great idea. 
Filed https://skype.visualstudio.com/SPOOL/_sprints/taskboard/Web%20UI/SPOOL/CY2021-Q3/Sprint%2041%20(Ends%20July%2025)?workitem=2519191 to iterate on this API on the top of the stack.

PR #578, opened 2021-07-13
Chat automation tests

anjulgarg:
# What
Tests for
- User can receive message
- Sender can view that the message status has turned to seen
- Receiver can view a typing indicator
- Typing indicator disappears after 12 seconds. 

# Why

# How Tested

# Process & policy checklist


JamesBurnside:
Do we need to be concerned with the impact on repo file size footprint?

Looks like each file is ~15kb - so after ~700 images we'll hit 10GB - however if we replace one image with another - does git store the old binary files or can it do a delta like with txt files? (if it does the former we could be storing _a lot_ of images)

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -63,21 +70,56 @@ test.describe('Chat Composite E2E Tests', () => {
...
+    expect(await pages[0].screenshot()).toMatchSnapshot('5-typing-indicator.png');
+  });
+
+  test('typing indicator disappears after 12 seconds', async () => {
+    const page = pages[0];
+    await page.bringToFront();
+    await page.waitForTimeout(12000);
+    const el = await page.$(dataUiId(IDS.typingIndicator));
+    expect(await el?.innerHTML()).toBeFalsy();
+    expect(await page.screenshot()).toMatchSnapshot('6-typing-indicator-disappears.png');
```
JamesBurnside, line 123: If this fails, how do we view the diff you showed before? If its via running locally only, is there a command I can run to get view the diff for a failed/named test?

JamesBurnside: Likewise for reviewers is there a way they could see the diff also ?

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -63,21 +70,56 @@ test.describe('Chat Composite E2E Tests', () => {
...
+    for (const idx in pages) {
+      await compositeLoaded(pages[idx]);
+      expect(await pages[idx].screenshot()).toMatchSnapshot(`1-page-${idx}-chat-screen.png`);
+    }
   });
 
   test('can send message', async () => {
+    const page = pages[0];
+    await page.bringToFront();
+    await page.type(dataUiId(IDS.sendboxTextfield), 'How the turn tables');
```
JamesBurnside, line 83: 😄

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -63,21 +70,56 @@ test.describe('Chat Composite E2E Tests', () => {
...
+    await pages[0].waitForSelector(dataUiId(IDS.typingIndicator));
+    const el = await pages[0].$(dataUiId(IDS.typingIndicator));
+    expect(await el?.innerHTML()).toContain(participants[1]);
+    expect(await pages[0].screenshot()).toMatchSnapshot('5-typing-indicator.png');
+  });
+
+  test('typing indicator disappears after 12 seconds', async () => {
+    const page = pages[0];
+    await page.bringToFront();
+    await page.waitForTimeout(12000);
```
JamesBurnside, line 120: 😢

JamesBurnside: Is there anyway to mock timers and speed this up?

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -63,21 +70,56 @@ test.describe('Chat Composite E2E Tests', () => {
...
+    }
   });
 
   test('can send message', async () => {
+    const page = pages[0];
+    await page.bringToFront();
+    await page.type(dataUiId(IDS.sendboxTextfield), 'How the turn tables');
+    await page.keyboard.press('Enter');
+    await page.waitForSelector(`[data-ui-status="delivered"]`);
+    stubTimestamps(page);
```
JamesBurnside, line 86: what does `stubTimestamps` do?

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: If you run this locally and the outcome doesn't match. A new dirrectory `test-result` will be created inside `react-composite` with the old, new and diff images

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Changes all the timestamps on a page to `timestamps` otherwise visual parity tests will always fail because messages will have different timestamps

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Can't help it unfortunately.

anjulgarg:
> Do we need to be concerned with the impact on repo file size footprint?
> 
> Looks like each file is ~15kb - so after ~700 images we'll hit 10GB - however if we replace one image with another - does git store the old binary files or can it do a delta like with txt files? (if it does the former we could be storing _a lot_ of images)

You mean 10MB?
Git stores the old binary files in complete form. AFAIK.
But this shouldn't be a problem. We squash everytime we merge.
Plus once an image snapshot has been generated, it shouldn't ideally change

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
JamesBurnside: Hmm, I would say if we have to arbitrary wait 12 seconds we need to move this test to a non-playwright test were timers can be mocked

Given that typing indicator checks the Date.now() can we mock that https://stackoverflow.com/a/52876069?

JamesBurnside:
> > Do we need to be concerned with the impact on repo file size footprint?
> > Looks like each file is ~15kb - so after ~700 images we'll hit 10GB - however if we replace one image with another - does git store the old binary files or can it do a delta like with txt files? (if it does the former we could be storing _a lot_ of images)
> 
> You mean 10MB?
> Git stores the old binary files in complete form. AFAIK.
> But this shouldn't be a problem. We squash everytime we merge.
> Plus once an image snapshot has been generated, it shouldn't ideally change

Hahaha yep my bad that's how bytes work 😅 Yah 10MB is totally fine

I disagree that snapshots shouldn't change though, we should expect to have to update these any time and plan ahead that we don't get blocked on that, squash merging will still maintain the old history just the linaege merged into the main branch is squashed, it will also keep the history for each file across muitiple PRs regardless of squash.

That said even if we have 500 tests written that change every single day, it would take 4 years to hit 10GB (someone check my math) so we should be totally fine on file size footprint

packages/communication-react/review/communication-react.api.md
```diff
@@ -889,6 +889,8 @@ export interface Identifiers {
     participantList: string;
     // (undocumented)
     sendboxTextfield: string;
+    // (undocumented)
```
prprabhu-ms, line 892: Please document. All public API must be documented.

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -238,7 +240,12 @@ const getIndicatorComponent = (
   );
 
   return (
-    <Stack horizontal className={mergeStyles(typingIndicatorStringStyle, styles?.typingString)} key="typingStringKey">
+    <Stack
+      horizontal
+      data-ui-id={ids.typingIndicator}
+      className={mergeStyles(typingIndicatorStringStyle, styles?.typingString)}
+      key="typingStringKey"
```
prprabhu-ms, line 247: Tangential to this PR: The `key` is unnecessary. Explicit keys are only really needed in lists (e.g. `Stack.Item`).

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -253,7 +260,7 @@ export const TypingIndicator = (props: TypingIndicatorProps): JSX.Element => {
 
   const typingUsersToRender = typingUsers.filter((typingUser) => typingUser.displayName !== undefined);
 
-  const indicatorComponent = getIndicatorComponent(
+  const indicatorComponent = IndicatorComponent(
```
prprabhu-ms, line 263: Please inline this definition on line 272.

packages/react-components/src/identifiers/IdentifierProvider.tsx
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
```
prprabhu-ms, line 1: This diff looks wrong to me. Perhaps you wanted to create the PR against another branch?

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -63,21 +70,56 @@ test.describe('Chat Composite E2E Tests', () => {
...
-  test('composite loads correctly with participant list', async () => {
-    await pages[0].bringToFront();
-    await pages[0].waitForSelector(dataUiId(IDS.sendboxTextfield));
-    await pages[0].waitForSelector(dataUiId(IDS.participantList));
-    stubTimestamps(pages[0]);
-    expect(await pages[0].screenshot()).toMatchSnapshot('1-chat-screen.png');
+  test('composite pages load completely', async () => {
+    for (const idx in pages) {
+      await compositeLoaded(pages[idx]);
+      expect(await pages[idx].screenshot()).toMatchSnapshot(`1-page-${idx}-chat-screen.png`);
```
prprabhu-ms, line 76: What determines the names of these snapshots?

Let's avoid using numbers in the file names. When we reorder the tests in this file (or add one in the middle), we'll either end up with non-contiguous numbers while reading the file (confusing) or will have to rename all the files (unnecessary).

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
prprabhu-ms: Recommend renaming this function so that it conveys the intent vs. what it does.

How about `stubNonDeterministicElements(page)` or `replaceNonDeterministicElementsWithPlaceholder(page)`?

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -17,6 +17,13 @@ const stubTimestamps = (page: Page): void => {
   }, messageTimestampId);
 };
 
+const compositeLoaded = async (page: Page): Promise<void> => {
+  await page.waitForLoadState('load');
+  await page.waitForSelector(dataUiId(IDS.sendboxTextfield));
+  await page.waitForSelector(dataUiId(IDS.participantList));
+  await page.waitForTimeout(1000);
```
prprabhu-ms, line 24: Does this add a second to each test? Why is this necessary?

This can also introduce flakiness to the test as 1 second may not be enough for things to settle on a loaded CI server.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -63,21 +70,56 @@ test.describe('Chat Composite E2E Tests', () => {
...
+    await page.keyboard.press('Enter');
+    await page.waitForSelector(`[data-ui-status="delivered"]`);
+    stubTimestamps(page);
+    expect(await page.screenshot()).toMatchSnapshot('2-send-message.png');
+  });
+
+  test('can receive message', async () => {
+    const page = pages[1];
+    await page.bringToFront();
+    await page.waitForSelector(`[data-ui-status="delivered"]`);
```
prprabhu-ms, line 93: I'm confused by this test.

* We're waiting for `delivered` status to test if the message is received.
* Who sent the message? I expected us to send a message from one page and then check that it's received on the other.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -63,21 +70,56 @@ test.describe('Chat Composite E2E Tests', () => {
...
+    stubTimestamps(page);
+    expect(await page.screenshot()).toMatchSnapshot('3-receive-message.png');
+  });
+
+  test('read message has a viewed status', async () => {
+    const page = pages[0];
+    await page.bringToFront();
+    await page.waitForSelector(`[data-ui-status="seen"]`);
```
prprabhu-ms, line 101: I'm confused by this test. Who sent the message?

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
prprabhu-ms: Vote to drop this test if there's no way to mock out the timers.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Too verbose.
I think `stubMessageTimestamps` should suffice.

packages/react-components/src/components/TypingIndicator.tsx
anjulgarg: Didn't add it. It was already there.

packages/react-components/src/components/TypingIndicator.tsx
anjulgarg: Any particular reason why?

packages/react-components/src/identifiers/IdentifierProvider.tsx
anjulgarg: Nope its correct. The file is named incorrectly. `IndentifierProvider`
Renaming the file. That's why you see it here like this.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Not to each test. Adds 1 second per browser page to overall test run.
So in this case, a total of 2 seconds.
And yes, it is a flaky test but I couldn't find a better way. 
Improving this is in my todo.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Right. I named them such for local development. Makes sense to remove the numbers. Will do

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: If you notice, the message was sent in the previous test by page[0]
and is being checked here by page[1]

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: page[0] sent the message inside test `can send message`
Any suggestions to make it more clear?

packages/react-components/src/components/TypingIndicator.tsx
anjulgarg: Any particular reason?

packages/react-components/src/identifiers/IdentifierProvider.tsx
anjulgarg: Renamed the file. That's why you see it like this.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Total time increased is 1 second per browser page.
So a total of 2 seconds are added to these tests.
Without doing this, messages sent at the beginning were not getting delivered to the recipient.
Couldn't find a better way to handle this. 
Improving this is in my todo list.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Right. I did this for local debugging. Makes sense to remove numbers.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: page[0] sent the message in a previous test.
page[1] is checking to see if it shows on their screen

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: page[0] sent the message in a previous test.
Any suggestions to make this clear?

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Good idea @JamesBurnside 
Nan and I tried it earlier for mocking message timestamps but it didn't work.
Thankfully it worked for mocking typingindicator behavior.
Updated the code.

Note: we still have to wait a second.

alcail (approved):

packages/react-components/src/components/TypingIndicator.tsx
prprabhu-ms: I'm a bit wary of using Component functions outside of JSX syntax. It's all too easy to end up wrapping this in a `useEffect` or `useMemo` that React then complains about. We had this problem (have it still?) in `main` last week.

packages/react-components/src/identifiers/IdentifierProvider.tsx
prprabhu-ms: Hmm. Github could do a better job with tracking renames then...

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
prprabhu-ms: Oh... these tests aren't independent?

Is there a way to make them isolated? Or is it prohibitive in terms of runtime? Tests that aren't isolated can be flaky / one test can lead to failures for all the tests downstream making debugging much harder.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
prprabhu-ms: Covered in another discussion thread, so closing this one.

I didn't realize that these tests aren't isolated.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
prprabhu-ms: I read playwright docs and now I understand this better.

My recommendation is: use `test.beforeAll` to start the browser and the app server. Then, use `test.beforeEach` to create a new browser context and new users and threads for each test. This way they stay isolated behaviorally.

anjulgarg: Running each test in isolation means spawning a new browser page and initializing it for every test. Increasing runtime.
Although, I myself want to make the tests isolated. Not 100% sure if it's worth it right now. Thoughts?

anjulgarg: Plus deeper scenarios will require a lot of logic duplication that other tests already tested. Any suggestions for avoiding that?

prprabhu-ms (approved):

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
prprabhu-ms: As discussed IRL, let's land this PR with sequential tests, making it more obvious that the tests need to run sequentially. We should take another look at this before we can add the tests to CI.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
prprabhu-ms: As discussed IRL, let's add a comment explaining why we have this wait.

I am confident this is discovering a real bug in our implementation. We lose messages that appear right when the stateful client is being initialized.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
JamesBurnside: jest by default tries to run tests in parallel, is there something in the config (`--runinband`?) to ensure these run sequentially? (I can't remember if its tests suites are run in parallel or includes tests within a suite)

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
JamesBurnside: Unresolving conversation, still shows 12s - is there a local change waiting to be pushed?
If we still need the arbitrary wait for 1 second can you add a comment saying why (assume this is because our typing indicator has a `setTimeout((), 1000)` for checking if it should update the indicator)

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
JamesBurnside: The chat composite in storybook sometimes loses the first bot message, I assumed it was a strange storybook rendering issue where it rerenders somepoint after the first message but could be the same underlying bug

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: @JamesBurnside Probably forgot to push the changes. The code has now been updated.
Although typingindicator selector doesn't have a setTimeout, this wait is to ensure we don't encounter intermittent failures and allow selector to mutate the state based on the events received from underlying ACS SDK.

packages/communication-react/review/communication-react.api.md
anjulgarg: Done

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Looks like it.

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: These tests are not being run by jest. Playwright has its own test runner  and we are using that. It runs test in sequence.

PR #580, opened 2021-07-14
Delete banner showing user not in thread

prprabhu-ms:
# What
Delete the banner that shows when a user is not in thread.

# Why
* This duplicates notification that is in `ErrorBar` now.
* The business logic was buggy: When an incorrect userId is provided but with a valid token for a user in the thread, the banner is shown but the user can still send messages.

# How Tested
ChatComposite in local storybook.

# Process & policy checklist


prprabhu-ms:
Oops. Wrong base branch.

PR #584, opened 2021-07-14
Add example to play with ChatComposite resizing

prprabhu-ms:


prprabhu-ms:
Just an example to play with.

PR #585, opened 2021-07-14
Create default control bar button

JamesBurnside:
# What
Create default control bar button
* Create template button
* Make other buttons consume this button
  * This should be a no-op for these components
  * Though fixed some gaps were props were not overridable by customers and always used the default
* Add storybook story for the default button

# Why
Currently its not possible to create a custom button that also uses the same styling and behavior as our current control bar buttons.
We will need this ourselves for creating chat button in the meeting/call composite so makes sense to expose this

# How Tested
Tested storybook and call composite


packages/react-components/review/react-components.api.md
```diff
@@ -47,7 +48,7 @@ export type CallParticipant = CommunicationParticipant & {
 export const CameraButton: (props: CameraButtonProps) => JSX.Element;
 
 // @public
-export interface CameraButtonProps extends IButtonProps {
+export interface CameraButtonProps extends ControlBarButtonProps {
     localVideoViewOption?: VideoStreamOptions;
     onToggleCamera?: (options?: VideoStreamOptions) => Promise<void>;
     showLabel?: boolean;
```
alcail, line 54: did you forget to remove this showLabel?

alcail (approved):

packages/react-components/review/react-components.api.md
JamesBurnside: Updated 👍

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -94,13 +71,14 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
...
-    <DefaultButton
+    <ControlBarButton
       {...props}
       disabled={props.disabled || waitForCamera}
       onClick={onToggleCamera ? onToggleClick : props.onClick}
-      styles={componentStyles}
-      onRenderIcon={onRenderIcon ?? defaultRenderIcon}
-      onRenderText={showLabel ? onRenderText ?? defaultRenderText : undefined}
+      onRenderOnIcon={props.onRenderOnIcon ?? onRenderCameraOnIcon}
+      onRenderOffIcon={props.onRenderOnIcon ?? onRenderCameraOffIcon}
```
prprabhu-ms, line 79: Did you mean to check for `props.onRenderOffIcon`?

packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const defaultRenderText = useCallback(
+    (props?: IButtonProps): JSX.Element => {
+      const labelText = props?.text ?? strings?.label ?? (props?.checked ? strings?.onLabel : strings?.offLabel);
+      return (
+        <Label key={labelKey} className={mergeStyles(controlButtonLabelStyles, props?.styles?.label)}>
+          {labelText}
+        </Label>
+      );
+    },
+    [labelKey, strings?.label, strings?.offLabel, strings?.onLabel]
```
prprabhu-ms, line 82: Why the difference of treatment between `props.styles.labels` vs `props.strings.label` etc?

Contrast with just a private module level functions that takes `props` as arguments. Or even better, a `DefaultRenderText` component with `props`.

prprabhu-ms: Same comment for `defaultRenderIcon` below.

packages/react-components/src/components/ScreenShareButton.tsx
```diff
@@ -39,46 +32,31 @@ export interface ScreenShareButtonProps extends IButtonProps {
...
   return (
-    <DefaultButton
+    <ControlBarButton
       {...props}
       onClick={props.onToggleScreenShare ?? props.onClick}
-      styles={componentStyles}
-      onRenderIcon={onRenderIcon ?? defaultRenderIcon}
-      onRenderText={showLabel ? onRenderText ?? defaultRenderText : undefined}
+      onRenderOnIcon={props.onRenderOnIcon ?? onRenderScreenShareOnIcon}
+      onRenderOffIcon={props.onRenderOnIcon ?? onRenderScreenShareOffIcon}
```
prprabhu-ms, line 57: Did you mean `props.onRenderOffIcon`?

prprabhu-ms (approved):

packages/react-components/src/components/CameraButton.tsx
JamesBurnside: I did nice catch!

packages/react-components/src/components/ControlBarButton.tsx
JamesBurnside: Do you mean why are some left out of this dependency array? If so - total oversight and I'll add them now, I actually thought the eslint rule for hooks was meant to catch this but it isn't being applied to useCallback...

JamesBurnside: Actually props don't need in there because its a function taking in props (still odd there's no linting to make this decision). Did you mean why are some prefixed with `props` and others extracted in the props above? I have no good answer :D I will update to none are extracted in the props above I think

prprabhu-ms: Yeah, I meant the latter. Some are captured by the closure, while others are used from the `props` arg, and it isn't clear to me why there is that difference.

As I said, this really looks like a candidate for a private `DefaultRenderTest` component to me.

anjulgarg (approved):

packages/react-components/src/components/ControlBarButton.tsx
JamesBurnside: Oh yah that's much cleaner now, nice suggestion

packages/react-components/src/components/ScreenShareButton.tsx
JamesBurnside: updated

packages/react-components/src/components/CameraButton.tsx
JamesBurnside: updated

PR #587, opened 2021-07-14
Make chat composite topic optional to show

JamesBurnside:
# What
Make the header containing the chat topic optional to show

# Why
Meeting composite doesn't want to show this - highly likely a customer won't as well e.g. if they just want to embed a chat

# How Tested
Storybook basic example
| `showTopic: true` | `showTopic: false`
| -- | -- |
| image | image


alcail (approved):

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -34,8 +34,16 @@ export type ChatCompositeProps = {
  * Additional customizations for the chat composite
  */
 export type ChatOptions = {
-  /** Choose to show the participant pane */
```
prprabhu-ms, line 37: There are two consistency arguments here:

* All boolean options have the default value of `false`: Default value of `true` is confusing -- not setting the boolean field is the same as setting it to `true`...
* All feature toggles are feature enablers (i.e., `showParticipantPane` and `showTopic` etc vs `hideParticipantPane` and `hideTopic`).

Trying to decide which default is the "obvious one" is difficult in many cases -- all customers don't want the same thing, and don't want the same thing over time. 

My personal choice here would be to keep the toggles as enablers (i.e. `show*`) and default them all to `false`.

JamesBurnside: Very good point, I agree it should always default to false so we should either have `showTopic` and have it as `false` by default, or `hideTopic` and default to `false`.

My only reason to default to showing everything as true is it makes the Out-of-box experience richer (and those who then don't want it can hide those things)

So I think my preference is to have:
```
{
  hideParticipantPane?: boolean; // defaults to false
  hideTopic?: boolean; // defaults to false
}
```

But would love to hear more opinions on this

JamesBurnside: Hmm but then when we release new features, to not be breaking, they will have to be feature enablers not disablers so it would be better to be off by default:
```
{
  showParticipantPane?: boolean; // defaults to false
  showTopic?: boolean; // defaults to false
}
```

JamesBurnside: (updated to `show*` and defaulting to false, but happy to change this if folks want)

anjulgarg (approved):

prprabhu-ms (approved):

PR #589, opened 2021-07-14
Remove sendBoxMaxLength from chat composite

JamesBurnside:
# What
Remove sendBoxMaxLength option from chat composite.

# Why
This doesn't follow responsive design and assumes that the REM value is 16px which it mightn't be.


JamesBurnside:
I don't have the historical context to why this exists so let me know if this should be kept, but currently its a very odd option for us to be supporting and the 16px math isn't quite right

prprabhu-ms (approved):

alcail (approved):

PR #591, opened 2021-07-15
Refactor to avoid using hooks inside useMemo

anjulgarg:
# What
Bugfix for issue https://skype.visualstudio.com/SPOOL/_workitems/edit/2519307

# Why
To populate custom ids for message and message timestamp, we used IdentityProvider Context Provider and used a hook named `useIdentifiers` inside useMemo causing console errors in Chat Composite.

# How Tested

# Process & policy checklist


mgamis-msft (approved):

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -743,7 +745,7 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
   // To rerender the defaultChatMessageRenderer if app running across days(every new day chat time stamp need to be regenerated)
   const defaultChatMessageRenderer = useCallback(
     (messageProps: MessageProps) => {
-      return DefaultChatMessageRenderer(messageProps);
+      return DefaultChatMessageRenderer(messageProps, ids);
```
prprabhu-ms, line 748: Recommend sending in only what you need:
```suggestion
      return DefaultChatMessageRenderer(messageProps, { messageTimestamp: ids.messageTimestamp });
```

prprabhu-ms (approved):

alcail (approved):

PR #593, opened 2021-07-16
Clean up browser test (docs)

prprabhu-ms:
# What
Some simple clean up in E2E tests and docs.

# Why
Avoid the confusion I had.


packages/react-composites/package.json
```diff
@@ -16,7 +16,7 @@
     "build:e2e:chat": "webpack-cli build --config ./tests/browser/chat/app/webpack.config.js --mode=production --env production",
     "test": "jest",
     "test:coverage": "npm run test -- --coverage",
-    "test:e2e:chat": "playwright test chat",
+    "test:e2e:chat": "playwright test tests/browser/chat",
```
anjulgarg, line 19: This won't work. Could you double check?

anjulgarg:
Thanks a lot for creating this PR.
<3

packages/react-composites/package.json
prprabhu-ms: Work for me. Why should this not work?

The positional argument is treated as regexp, so `chat` is a bit too broad (it could easily pick up jest tests from `src/*`)
image

anjulgarg: Try running the command directly.
`npx playwright test tests/browser/chat`
Does it work?

anjulgarg: Doesn't work for me. Just checked.

anjulgarg: image

anjulgarg: I think this is because you are running it in codespaces on a linux environment.
Windows and Linux have different path resolution

anjulgarg: You can verify by running it on your dev machine using this
`npx playwright test tests\\browser\\chat`

prprabhu-ms: Great catch. Verified that using `.` works for both, as this is a regexp.

mgamis-msft (approved):

packages/react-composites/package.json
```diff
@@ -16,7 +16,7 @@
     "build:e2e:chat": "webpack-cli build --config ./tests/browser/chat/app/webpack.config.js --mode=production --env production",
     "test": "jest",
     "test:coverage": "npm run test -- --coverage",
-    "test:e2e:chat": "playwright test chat",
+    "test:e2e:chat": "playwright test tests.browser.chat",
```
anjulgarg, line 19: Haha Very smart.

anjulgarg (approved):

PR #594, opened 2021-07-16
Chat automation tests

anjulgarg:
# What
The following tests are added
- Participant can rejoin the chat 
- Custom Data model

# Why

# How Tested

# Process & policy checklist


mgamis-msft (approved):

packages/react-composites/tests/browser/chat/app/index.tsx
```diff
@@ -39,11 +40,34 @@ function App(): JSX.Element {
...
 
-  return (
-    <>
-      {chatAdapter && <ChatComposite identifiers={IDS} adapter={chatAdapter} options={{ showParticipantPane: true }} />}
-    </>
-  );
+  if (customDataModel) {
+    return (
+      <>
+        {chatAdapter && (
```
JamesBurnside, line 46: nit: much of this is consistent across the if statement regardless of customDataModel, extract what is applicable to the conditional to better contextualize why the if is needed and reduce redundancy
```
const onRenderTypingIndicator = customDataModel ? () => <text id="custom-data-model-typing-indicator">Someone is typing...</text> | undefined;
```

change/@internal-react-composites-606157b6-7ad0-4634-8c79-294452b5e13a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
prprabhu-ms, line 2: nitty nit: I am beginnging to mark PRs that have no API and functional change as "none".

packages/react-composites/tests/browser/chat/app/identity.ts
```diff
@@ -1,40 +0,0 @@
-// Copyright (c) Microsoft Corporation.
```
prprabhu-ms, line 1: This file is already gone on `main`.

packages/react-composites/tests/browser/chat/app/index.tsx
prprabhu-ms: I'd go a step further: The `customDataModel` URL parameter is problematic in two ways:
* It's not clear in `ChatComposite.test.ts` what setting this URL param does. When reading the test, I have to go see the implementation of `app/index.tsx` to get an idea of what the test does.
* Makes `app/index.tsx` inflexible and tied to that single test case.

Instead, how about adding URL parameters:
* `overrideTypingIndicator=<value>`
* `overrideRenderMessage=<value>`
* `overrideRenderAvatar=<value>`

And that test just sets all three:

    page.goto('....overrideRenderAvatar=Avatar&...')

Now, if we wanted to, we could split that test easily into three different cases -- one for each render override.

This also makes the implementation in `app.index.tsx` more natural:

    onRenderMessage={()=>{params.overrideRenderMessage ? <text id="overide-render-message">{params.overideRenderMessage}</text> : undefined}}

prprabhu-ms:
> Are these running as part of the CI? I'd love to understand the perf more to know if/how many tests becomes too many tests for CI runs. If we're only using these as before-release tests then there's no worry here 

We're not running these tests in CI. They are live tests (so there is concern around flakiness) and we can't parallelize them because of throttling from the ACS service backends.

JamesBurnside:
> We're not running these tests in CI. They are live tests (so there is concern around flakiness) and we can't parallelize them because of throttling from the ACS service backends.

Ah makes sense! What's the impact each knew test roughly adds? (like before this PR how long did tests take, and after this PR how long do tests take)

I'm happy for this PR to go in (seems great!) but I'm very hesistant of adding too many tests if they are flakey and not run periodically - when someone runs it locally they won't know if their changes caused a regression or perhaps it was just because that test hasn't been run in a while

packages/react-composites/tests/browser/chat/app/index.tsx
anjulgarg: The issue with that is that it only allows you to provide text values. We won't be able to test complex components as custom data model.

anjulgarg:
> > We're not running these tests in CI. They are live tests (so there is concern around flakiness) and we can't parallelize them because of throttling from the ACS service backends.
> 
> Ah makes sense! What's the impact each knew test roughly adds? (like before this PR how long did tests take, and after this PR how long do tests take)
> 
> I'm happy for this PR to go in (seems great!) but I'm very hesistant of adding too many tests if they are flakey and not run periodically - when someone runs it locally they won't know if their changes caused a regression or perhaps it was just because that test hasn't been run in a while

Surprisingly, playwright tests are super fast.
The current set of tests complete in under 15 seconds.
That includes playwright test suite compiling ts, spawning pages, waiting for them to load and running the actual tests.
On an average each test would add ~ 1 sec.

Your concern is valid. I think unless we add these tests to the CI pipeline, we won't know how flaky/useful they are.

packages/react-composites/tests/browser/chat/app/index.tsx
anjulgarg: > nit: much of this is consistent across the if statement regardless of customDataModel, extract what is applicable to the conditional to better contextualize why the if is needed and reduce redundancy
> 
> ```
> const onRenderTypingIndicator = customDataModel ? () => <text id="custom-data-model-typing-indicator">Someone is typing...</text> | undefined;
> ```

Can do.

JamesBurnside:
> > > We're not running these tests in CI. They are live tests (so there is concern around flakiness) and we can't parallelize them because of throttling from the ACS service backends.
> > 
> > 
> > Ah makes sense! What's the impact each knew test roughly adds? (like before this PR how long did tests take, and after this PR how long do tests take)
> > I'm happy for this PR to go in (seems great!) but I'm very hesistant of adding too many tests if they are flakey and not run periodically - when someone runs it locally they won't know if their changes caused a regression or perhaps it was just because that test hasn't been run in a while
> 
> Surprisingly, playwright tests are super fast.
> The current set of tests complete in under 15 seconds.
> That includes playwright test suite compiling ts, spawning pages, waiting for them to load and running the actual tests.
> On an average each test would add ~ 1 sec.
> 
> Your concern is valid. I think unless we add these tests to the CI pipeline, we won't know how flaky/useful they are.

Wow nice!!

JamesBurnside (approved):

prprabhu-ms:
> > The current set of tests complete in under 15 seconds.
> > That includes playwright test suite compiling ts, spawning pages, waiting for them to load and running the actual tests.
> > On an average each test would add ~ 1 sec.
> 
> Wow nice!!

Alas, this number will grow quickly.
My `ErrorBar` tests add another 15 seconds (combined).

If only we could run the tests in parallel...

PR #596, opened 2021-07-17
Add first browser test for ErrorBar

prprabhu-ms:
# What
* Add the first browser test for ErrorBar. Takes care of boilerplate.
* To aid code sharing, refactor the browser tests so far into a reusable fixture.

# Why
'cause I don't wanna test manually.

# How Tested
`rushx test:e2e:chat`


prprabhu-ms:
@anjulgarg This PR conflicts in a big way with #594 

Let's land that one first. I'm happy to rebase this on top and refactor the tests added in #594 as well.

packages/react-composites/tests/browser/chat/ErrorBar.test.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { test } from './fixture';
+import { loadPage, stubMessageTimestamps, waitForCompositeToLoad } from '../utils';
+import { expect } from '@playwright/test';
+
```
alcail, line 7: Should we add a test to verify that there's no ErrorBar displayed if there's no error?

packages/react-composites/tests/browser/chat/fixture.ts
```diff
@@ -0,0 +1,92 @@
...
+  pages: [
+    async ({ serverUrl, testBrowser, users }, use) => {
+      const pages = await Promise.all(users.map(async (user) => loadPage(testBrowser, serverUrl, user)));
+      await use(pages);
+    },
+    { scope: 'worker' }
+  ]
+});
+
+const SERVER_URL = 'http://localhost:3000';
```
alcail, line 91: nit: just a personal taste to have those const declared at the beginning not the end

packages/react-composites/tests/browser/utils.ts
```diff
@@ -49,10 +72,35 @@ export const createUserAndThread = async (
...
+
+const encodeQueryData = (data: IdentityType): string => {
+  const qs: Array<string> = [];
+  for (const d in data) {
+    qs.push(encodeURIComponent(d) + '=' + encodeURIComponent(data[d]));
+  }
+  return qs.join('&');
+};
+
+const CONNECTION_STRING = process.env.CONNECTION_STRING ?? '';
```
alcail, line 100: nit: same as before, better to have those constants declared at the beginning

alcail (approved):

packages/react-composites/tests/browser/chat/fixture.ts
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { PARTICIPANT_NAMES } from '../config';
```
anjulgarg, line 4: `PARTICIPANT_NAMES` have been removed from config. How is this working?

packages/react-composites/tests/browser/chat/fixture.ts
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Creates a Chat thread and adds users to the thread.
+   *
+   * @returns the created users' identities.
+   */
+  users: [
+    // playwright forces us to use a destructuring pattern for first argument.
+    /* eslint-disable-next-line no-empty-pattern */
+    async ({}, use) => {
+      const users = await createChatThreadAndUsers(PARTICIPANTS);
```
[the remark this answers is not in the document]
anjulgarg: Also, why are we importing PARTICIPANT_NAMES above if PARTICIPANTS is declared here?

packages/react-composites/tests/browser/utils.ts
anjulgarg: +1

packages/react-composites/tests/browser/chat/fixture.ts
prprabhu-ms: Conflicts with the intent to declare public stuff up top.

A reader only intending to use this package doesn't need to know about `SERVER_URL`, so I don't want to show it to them first.

packages/react-composites/tests/browser/chat/ErrorBar.test.ts
prprabhu-ms: +1, thanks for calling that out.

I will add one to the following PR that introduces the remaining tests. This one mostly uses one test to get the boilerplate down.

packages/react-composites/tests/browser/chat/fixture.ts
prprabhu-ms: Oops. Don't now how this leaked in. That variable is actually deleted from `../config`

packages/react-composites/tests/browser/chat/fixture.ts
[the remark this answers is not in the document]
prprabhu-ms: Moved constant up, removed spurious (...and wrong) import.

packages/react-composites/tests/browser/chat/fixture.ts
prprabhu-ms: Moved it up to make progress. I don't care too much about this. But the intent was to keep clutter off a reader who only cares about public API of the package ( i.e., the important reader ;) )

anjulgarg:
> @anjulgarg This PR conflicts in a big way with #594
> 
> Let's land that one first. I'm happy to rebase this on top and refactor the tests added in #594 as well.

Merged #594

anjulgarg (approved):

packages/react-composites/package.json
```diff
@@ -16,7 +16,7 @@
     "build:e2e:chat": "webpack-cli build --config ./tests/browser/chat/app/webpack.config.js --mode=production --env production",
     "test": "jest",
     "test:coverage": "npm run test -- --coverage",
-    "test:e2e:chat": "playwright test tests.browser.chat",
+    "test:e2e:chat": "playwright test --workers=1 tests.browser.chat",
```
JamesBurnside, line 19: this looks good, but just double checking this wasn't a param added for testing something locally

prprabhu-ms: Unfortunately we can't run these tests in parallel with any amount of robustness. The thread creation will fail intermittently because we're creating multiple threads with the same connection string in quick succession.

PR #601, opened 2021-07-20
Introduce modifiers for user-initiated state changes

prprabhu-ms:
# What
* Introduce a `StatefulChatClient.modifyState()` method for all caller initiated state changes.
* Provide a `newClearErrorsModifier` factory function for the modifier that clears errors.
* Use the new facility for the dismiss handler in `ErrorBar` component.

# Why
Better API for user-initiated state changes than before this PR.

# How Tested
* Storybook ChatComposite - dismissing errors works as expected.
* Ran e2e tests!

# Process & policy checklist

Drops the recently introduced (and not yet released) `StatefulChatClient.clearErrors()` method.


prprabhu-ms:
Especially @DominikMe  and @PorterNan PTAL.

Thanks for the idea of using state transformers.

JamesBurnside:
This looks good to me! but I'll let other's that have done more in this space review first, don't feel I can foresee any perf impacts or missed cases

packages/chat-stateful-client/src/modifiers/ClearErrors.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { ChatClientState, ChatErrorTargets } from '../ChatClientState';
+import { ChatStateModifier } from '../StatefulChatClient';
+
+export const newClearErrorsModifier = (targets: ChatErrorTargets[]): ChatStateModifier => {
```
alkwa-msft, line 7: I am curious if this can just be called clearErrorsModifier?

prprabhu-ms: It is a factory function that returns a modifier. I generally follow the `new*` naming pattern for factory functions.

But don't have very strong opinions if others think that the `new` is unnecessary.

alkwa-msft: ok so this is really creating an instance of a modifier meant for clearing specific errors. ok sure i think this should be fine.

alkwa-msft (approved):

packages/chat-stateful-client/src/StatefulChatClient.ts
```diff
@@ -141,6 +146,20 @@ export const createStatefulChatClient = (
...
 
+/**
+ * A function to modify the state of the StatefulChatClient.
+ *
+ * Provided as a callback to the {@Link StatefulChatClient.modifyState} method.
+ *
+ * The function must modify the provided state in place as much as possible.
+ * Making large modifications can lead to bad performance by causing spurious rerendering of the UI.
+ *
+ * Modifier function must return true if and only if some modification was made to the state.
```
DominikMe, line 157: What's the benefit of returning true/false? Can't the modifier just return the unchanged state instead and we add a ref equality check before setting it?

packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -51,6 +53,16 @@ export class ChatContext {
     return this._state;
   }
 
+  public modifyState(modifier: ChatStateModifier): void {
```
DominikMe, line 56: I'm curious why you chose an API where the developer passes a lambda instead of the modified state directly.

DominikMe (commented):
Added two questions

packages/chat-stateful-client/src/ChatContext.ts
prprabhu-ms: The intent is to make writing a performant modifier easy (conversely, writing a bad one harder). This API allows the developer to "modify state in place". Thus, by default only necessary changes will be made. With an API that simply accepts an updated state, developer has to be more careful about how they modified the state (modify in place? Be careful about what parts are new vs not).

We expect all modifiers to use `immer` like facility. This API forces that choice on the developer - they're less likely to make mistakes and we can reason about how state is updated in a uniform way within the `Context` and via modifiers.

packages/chat-stateful-client/src/StatefulChatClient.ts
prprabhu-ms: This is related to the other discussion about modifier via lambda vs return value. So let's discuss there.

The way `modifyState` works, the modifier is a lambda that changes state in-place. So we don't get a new object as a return value.

DominikMe: Let me ask another way:
Why can't the Context's `setState` function not do a reference equality check internally instead of relying on a Contoso provided bool flag?
The design with the bool return value leaves the pitfall open that Contoso's writes a modifier that actually modifies the state but returns false, or the other way around. I have trouble seeing the value in the bool return value.

packages/chat-stateful-client/src/ChatContext.ts
DominikMe: That's what I suspected, we let the developer write a plain lambda without having to know how to use immer, sounds good!

packages/chat-stateful-client/src/StatefulChatClient.ts
prprabhu-ms: +100

Updated the existing `batch()` API to do this. (In a separate PR), might be worth following this emit-if-changed pattern internally as well. I foudn some spurios events being emitted in calling in my recent unittest refactor.

DominikMe (approved):
Thanks for addressing the bool flag comment! Lgtm ✅

PR #603, opened 2021-07-20
Add system message to component/bindings/composite support

PorterNan:
# What

- Add system message to MessageThread component
- Add system message for selector props
- Link system message to participant received event

image

# Why

# How Tested

# Process & policy checklist


packages/chat-component-bindings/src/utils/constants.ts
```diff
@@ -3,3 +3,12 @@
 
 export const MINIMUM_TYPING_INTERVAL_IN_MILLISECONDS = 8000;
 export const PARTICIPANTS_THRESHOLD = 20;
+
+export const ACSKnownMessageType = {
```
JamesBurnside, line 7: This should be fixed here to not require this enum: https://github.com/Azure/azure-sdk-for-js/issues/15276
(might not be released yet though?)
But doesn't look to suppoer richtext/html https://github.com/Azure/azure-sdk-for-js/blob/12002a4ae839df764ba6415e8ee2d0bc6b9a4858/sdk/communication/communication-chat/src/generated/src/models/index.ts#L301

JamesBurnside (approved):
This is very cool! Looks great to me just one comment if we can avoid creating our own enum that we have to keep in sync with the chat sdk enum

packages/chat-component-bindings/src/chatThreadSelector.ts
```diff
@@ -83,38 +143,41 @@ export const updateMessagesWithAttached = (chatMessagesWithStatus: ChatMessage[]
      * in this case, we only want to show the read statusToRender of last message of the new messages block)
      */
     let attached: boolean | MessageAttachedStatus = false;
+    const previousMessage = messages[index - 1];
```
alcail, line 146: nit: those are constants used only in the 'else' in line 155, why not declare them inside this 'else' instead?
the previousMessage can be anything if index is 0... 
I know it will not be used in case of index being 0, but just don't like the possibility of being used in that case though...
Or can we declare them as undefined if index === 0 for previous and index > messages.length for next?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -178,18 +182,43 @@ const DefaultJumpToNewMessageButton = (props: JumpToNewMessageButtonProps): JSX.
...
+    const payload = props.message.payload;
+    if (payload.type === 'content') {
+      return (
+        <SystemMessageComponent
+          iconName={(payload.iconName ?? '') as SystemMessageIconTypes}
+          content={payload.content ?? ''}
+          containerStyle={props?.messageContainerStyle}
+        />
+      );
+    } else if (payload.type === 'participantAdded') {
```
alcail, line 205: nit: no need for 'else if' here. 'if' can be used only as you already return a value in previous 'if'.

packages/react-components/src/localization/translated/fr_FR.json
```diff
@@ -58,7 +58,9 @@
     "wednesday": "Mercredi",
     "thursday": "Jeudi",
     "friday": "Vendredi",
-    "saturday": "Samedi"
+    "saturday": "Samedi",
+    "participantJoined": "rejoint le chat.",
```
alcail, line 62: 'a rejoint la conversation.'

packages/react-components/src/localization/translated/fr_FR.json
```diff
@@ -58,7 +58,9 @@
     "wednesday": "Mercredi",
     "thursday": "Jeudi",
     "friday": "Vendredi",
-    "saturday": "Samedi"
+    "saturday": "Samedi",
+    "participantJoined": "rejoint le chat.",
+    "participantLeft": "a quitté le chat."
```
alcail, line 63: 'a quitté la conversation.'

packages/react-components/src/localization/translated/fr_FR.json
```diff
@@ -58,7 +58,9 @@
     "wednesday": "Mercredi",
     "thursday": "Jeudi",
     "friday": "Vendredi",
-    "saturday": "Samedi"
+    "saturday": "Samedi",
+    "participantJoined": "rejoint le chat.",
+    "participantLeft": "a quitté le chat."
   },
   "errorBar": {
     "unableToReachChatService": "Tu es hors ligne",
```
alcail, line 66: 'Vous êtes hors ligne.'

alcail (approved):

packages/chat-component-bindings/src/utils/constants.ts
alkwa-msft: the html/richtext is coming from teams

I think we need to support text, html and html/richtext

JamesBurnside: Does an issue need filed on the chat team to add `html/richtext` to this enum?

JamesBurnside: Or perhaps they need a new enum `enum ChatMessageTeamsInteropType = ChatMessageType | 'html/richtext'` or something. I think we can definitely keep this in for now but can you file an issue on the chat team and add a comment here we should remove this type once the chat team has a fix/supported type we can use?

packages/react-components/src/localization/translated/fr_FR.json
JamesBurnside: Who needs a Touchdown service when you have an Aline! 😁

packages/chat-component-bindings/src/utils/constants.ts
PorterNan: Looks like this is not in released version of js sdk, lemme leave a TODO here and we could check it later when it comes

PR #604, opened 2021-07-20
[Meeting composite] Add chat and people pane with basic toggle buttons

JamesBurnside:
# What
* Add chat & people buttons and the content pane to the right hand side of the composite (integrating buttons into the control bar itself will be the next PR)
* Toggle chat and people panes with the buttons

# Why
Adding participant view and chat view into meeting composite

# How Tested
Locally.

| state | screenshot |
| -- | -- |
| neutral call screen | image |
| chat opened | image |
| people opened | image |


anjulgarg:
The position of chat and people button changes?

JamesBurnside:
> The position of chat and people button changes?

Yeah, the next PR will be actually having them in the control bar element the other buttons are in to everything is integrated.

The white chunk at the side the current buttons cause will also disappear

alcail (approved):

packages/react-composites/src/composites/MeetingComposite/SidePane.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const EmbeddedPeoplePane = (props: {
+  participants: CommunicationParticipant[];
+  myUserId: string;
+  inviteLink?: string;
+  onClose: () => void;
+}): JSX.Element => {
+  const { inviteLink } = props;
+  return (
+    <SidePane headingText={'People'} onClose={props.onClose}>
+      <Stack tokens={{ childrenGap: '0.5rem' }}>
```
anjulgarg, line 43: Similar to how `className={mergeStyles()}` is recommended to be provided as a variable to avoid re-rendering. Is that the case with tokens as well?

anjulgarg (approved):

packages/react-composites/src/composites/MeetingComposite/SidePane.tsx
JamesBurnside: Great point, my gut reaction is yes as tokens prop will have a new reference, I'll update these

JamesBurnside: Updated 👍

PR #607, opened 2021-07-20
Center composite storybook pages' contents

JamesBurnside:
# What
Ensure composite storybook content is aligned to the center of the viewport

# Why
Currently things like `Initializing...` were not centered (see before-after below)

# How Tested
Locally tested each composite in storybook to ensure no unexpected layout change regressions

| before | after |
| -- | -- |
| image | image |


packages/storybook/stories/CallComposite/BasicExample.tsx
```diff
@@ -29,7 +30,7 @@ export const BasicExample: () => JSX.Element = () => {
   }, [knobs]);
 
   return (
-    <div style={COMPOSITE_EXPERIENCE_CONTAINER_STYLE}>
+    <Stack horizontalAlign="center" verticalAlign="center" styles={compositeExperienceContainerStyle}>
```
prprabhu-ms, line 33: If you're replacing `div` with `Stack`, shouldn't you also wrap the contents with `Stack.Item`?

prprabhu-ms (approved):

packages/storybook/stories/CallComposite/BasicExample.tsx
JamesBurnside: As far as I understand, each child is essentially a Stack.Item, the reason to use a Stack.Item is if you want to apply certain properties to different children (e.g. if different items grow at different rates relative to each other).

alcail (approved):

anjulgarg (approved):

PR #608, opened 2021-07-20
Disable `ErrorBar` in `ChatComposite` by default

prprabhu-ms:
# What
Disable `ErrorBar` in `ChatComposite` and add a feature flag to enable it.
Also enable it in automated tests, so that `ErrorBar` tests continue to provide coverage.

# Why
Error reporting work is still WIP so we don't want to include it in the next beta release.

# How Tested
Automated tests with/without feature flag to enable the ErrorBar.

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
+++ b/packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
@@ -64,6 +64,7 @@ export const ChatComposite = (props: ChatCompositeProps): JSX.Element => {
         <ChatAdapterProvider adapter={adapter}>
           <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=0" />
           <ChatScreen
+            showErrorBar={options?.showErrorBar}
             showParticipantPane={options?.showParticipantPane}
             showTopic={options?.showTopic}
             onRenderAvatar={onRenderAvatar}
--- a/packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
+++ b/packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
@@ -35,6 +35,14 @@ export type ChatCompositeProps = {
  */
 export type ChatOptions = {
   /**
+   * UNSTABLE: Feature flag to enable ErrorBar.
+   *
+   * This option will be removed once ErrorBar is stable.
+   *
+   * @defaultValue false
+   */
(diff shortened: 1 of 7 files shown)


prprabhu-ms:
Only adds an optional flag, so not a breaking change.

Now when I do remove that flag... that will be breaking ;)

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -34,6 +34,14 @@ export type ChatCompositeProps = {
  * Additional customizations for the chat composite
  */
 export type ChatOptions = {
+  /**
+   * UNSTABLE: Feature flag to enable ErrorBar.
+   *
+   * This option will be removed once ErrorBar is stable.
```
anjulgarg, line 40: IMO even when the error bar becomes stable, it's good to have it as an optional. CONTOSO might not want to show the error bar...

packages/storybook/stories/ErrorBar/ErrorBar.stories.tsx
```diff
@@ -25,6 +25,10 @@ const getDocs: () => JSX.Element = () => {
         `ErrorBar` is a wrapper on fluent UI's `MessageBar` with additional features for surfacing Azure Communication
         Services errors on the UI consistently.
       </Description>
+      <Description>
+        Set the `showErrorBar` option for `ChatComposite` to use an `ErrorBar` to show errors. This feature is still
```
anjulgarg, line 29: I would recommend removing the `This feature...` line from here.
We are still in Beta and this is a given anyway.
Plus we might forget to remove this piece of text once ErrorBar becomes stable.

anjulgarg (approved):

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
JamesBurnside: Could also tag it `@experimental` https://tsdoc.org/pages/tags/experimental/

JamesBurnside (approved):

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
prprabhu-ms: Tagged it @experimental.

Not sure we want to keep this option later on, so playing it safe.

PR #609, opened 2021-07-20
Persisting options button in sample app to avoid hiding

anjulgarg:
# What
Persist the options button to avoid hiding it on re-renders.

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/CallControls.tsx
+++ b/packages/react-composites/src/composites/CallComposite/CallControls.tsx
@@ -47,6 +47,7 @@ export const CallControls = (props: GroupCallControlsProps): JSX.Element => {
           callInvitationURL={callInvitationURL}
         />
       )}
+      {/* By setting `persistMenu` to true, we prevent options menu from getting hidden everytime a participant joins or leaves. */}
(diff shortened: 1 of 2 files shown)


packages/react-composites/src/composites/CallComposite/CallControls.tsx
```diff
@@ -47,7 +47,7 @@ export const CallControls = (props: GroupCallControlsProps): JSX.Element => {
           callInvitationURL={callInvitationURL}
         />
       )}
-      <OptionsButton {...optionsButtonProps} showLabel={!compressedMode} />
+      <OptionsButton persistMenu={true} {...optionsButtonProps} showLabel={!compressedMode} />
```
prprabhu-ms, line 50: Please add a comment in the code explaining the bug and how this fixes the bug.

prprabhu-ms (approved):

alkwa-msft (approved):

PorterNan (approved):

JamesBurnside:
@anjulgarg does this need to be applied to the participant button also?

PR #610, opened 2021-07-20
[Meeting composite] Add meeting call control bar

JamesBurnside:
# What
* Allow hiding the call composite's control bar
* Create a meeting call composite control bar
  * This uses the call composite call control directly
  * Then adds the chat and people buttons to the side

# Why
* Meeting composite design needs the chat and people button in the call control bar
* Using CallCompositeInternal format to avoid exposing showCallControls property on the call composite to the public API until our customization story is complete.

# How Tested
Sssssssstorybook

## Screenshots

| state | screenshot |
| -- | -- |
| config page (unchanged in this PR) | image |
| call neutral | image |
| people pane | image |
| chat pane | image |


packages/react-composites/src/composites/MeetingComposite/styles/SidePane.styles.ts
```diff
@@ -9,7 +9,7 @@ const palette = theme.palette;
 export const sidePaneContainerStyles: IStackItemStyles = {
   root: {
     height: '100%',
-    minWidth: '21.5rem'
+    width: '21.5rem'
```
JamesBurnside, line 12: note: this is not good with responsive design in mind, but that will be a future task

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -57,29 +57,37 @@ export const MeetingComposite = (props: MeetingCompositeProps): JSX.Element => {
...
   }, [showPeople]);
 
+  const endCallClick = (): void => {
+    callAdapter.setPage('configuration');
+  };
+
   return (
-    <Stack styles={{ root: { width: '100%', height: '100%' } }} horizontal>
-      <CallComposite adapter={callAdapter} fluentTheme={fluentTheme} />
+    <Stack styles={{ root: { width: '100%', height: '100%' } }}>
```
JamesBurnside, line 65: todo: check if I can use Stack stretch properties here instead

anjulgarg (approved):

alcail (approved):

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
JamesBurnside: I am updating this in the styling fixes I am working on. Closing this.

PR #615, opened 2021-07-21
Calling automation - config screen

anjulgarg:
# What
Test for local config screen validation

# Why

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/call/fixture.ts
```diff
@@ -52,8 +52,16 @@ export const test = base.extend<unknown, ChatWorkerFixtures>({
...
-        args: ['--start-maximized', '--disable-features=site-per-process'],
-        headless: false
+        executablePath: process.env.CHROME_PATH,
+        args: [
+          '--start-maximized',
+          '--disable-gpu',
+          '--use-fake-ui-for-media-stream',
+          '--use-fake-device-for-media-stream',
+          '--allow-file-access',
+          `--use-file-for-fake-video-capture=${path.join(__dirname, 'puppy.y4m')}`
```
JamesBurnside, line 62: To help with file size, sounds like we can use mpjeg also? (if its more compressed?)
https://source.chromium.org/chromium/chromium/src/+/master:media/capture/video/file_video_capture_device.h;l=33

JamesBurnside: If we use a super small file that is literally one frame, what happens? (does it just loop continuosuly?)

JamesBurnside: From that link, is just JPEG supported?? 
image

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -12,7 +27,19 @@ test.describe('Call Composite E2E Tests', () => {
       await pages[idx].waitForSelector(dataUiId('call-composite-device-settings'));
       await pages[idx].waitForSelector(dataUiId('call-composite-local-preview'));
       await pages[idx].waitForSelector(`${dataUiId('call-composite-start-call-button')}[data-is-focusable="true"]`);
-      expect(await pages[idx].screenshot()).toMatchSnapshot(`page-${idx}-call-screen.png`);
+      await stubLocalCameraName(pages[idx]);
+      expect(await pages[idx].screenshot()).toMatchSnapshot(`page-${idx}-call-screen.png`, { threshold: 0.5 });
```
JamesBurnside, line 31: for threshold, is this something we decide per test or will we apply .5 to all tests? (what does 0.5 really mean?)

JamesBurnside: Ah looks to be a guess and hope value 😅 https://github.com/mapbox/pixelmatch#api
What was venturing beyond the threshold in these tests?

packages/react-composites/tests/browser/call/fixture.ts
anjulgarg: Did try to use MJPEG and JPEG but it didn't work. I will try again today.

alcail (approved):

packages/react-composites/tests/browser/call/fixture.ts
anjulgarg: So JPEG and MJPEG is not working.
But I found a .y4m file in chromium repository that's just 744kb and is small enough to be used in our tests.

JamesBurnside: Sounds good to me!

JamesBurnside: (assuming no licensing/copyright issues from out using theirs)

packages/react-composites/tests/browser/utils.ts
```diff
@@ -11,7 +11,7 @@ export const dataUiId = (v: string): string => `[${DATA_UI_ID}="${v}"]`;
 const DATA_UI_ID = 'data-ui-id';
 const CONNECTION_STRING = process.env.CONNECTION_STRING ?? '';
 const PAGE_VIEWPORT = {
```
JamesBurnside, line 13: can you leave a comment here that this should match the value set in the fixture.ts (if it needs to), or use the same value from a variable from a const file

JamesBurnside (approved):

packages/react-composites/tests/browser/call/fixture.ts
anjulgarg: It's just 4 colored squares :D

packages/react-composites/tests/browser/call/CallComposite.test.ts
anjulgarg: Tweaked the code and it doesn't have a need for threshold. (For now)

PR #616, opened 2021-07-21
Add min height to call composite media gallery

JamesBurnside:
# What
* Ensure media gallery has a minimum height - below which the composite is not usable so this still fits in our responsive design
* Update the composite documentation to ensure it has a set height and width by default so users copying and pasting have a good experience
* Mention this issue in the call composite styling documentation

# Why
If the parent of the call composite doesn't have a height set, the video gallery was not visible.

# How Tested
In storybook
| before (if no height is set) | after (if no height is set) | default experience
| -- | -- | -- |
| image | image | image |

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/styles/CallScreen.styles.ts
+++ b/packages/react-composites/src/composites/CallComposite/styles/CallScreen.styles.ts
@@ -30,9 +30,7 @@ export const callControlsContainer = mergeStyles({
 export const containerStyles: IStackStyles = {
   root: {
     height: '100%',
-    width: '100%',
-    flexDirection: 'column',
-    display: 'flex'
+    width: '100%'
   }
 };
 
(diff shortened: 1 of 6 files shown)


packages/react-composites/src/composites/CallComposite/styles/CallScreen.styles.ts
```diff
@@ -30,27 +30,22 @@ export const callControlsContainer = mergeStyles({
...
   }
 };
 
 export const subContainerStyles: IStackStyles = {
   root: {
     overflow: 'hidden',
     width: '100%',
-    flexDirection: 'column',
-    display: 'flex',
-    flexBasis: '0'
```
PorterNan, line 45: Not sure if I remember this right, I remember set this flexBasis is something related to safari browser(it has the worst flex box support among all the browsers) compatibility, can we double check all the stylings are correct in safari/firefox browser?

PorterNan (approved):
Looks great, we finally have the min-height for the composite!

alcail (approved):

anjulgarg (approved):

packages/react-composites/src/composites/CallComposite/styles/CallScreen.styles.ts
JamesBurnside: Good note, I will double check this

JamesBurnside: Yep looks to be good on safari
Removing fixed height:
image
default experience:
image

JamesBurnside: And firefox: 
image

PR #620, opened 2021-07-23
Fix local preview overlapping the call control bar

alkwa-msft:
# What
We want to have the default location of the local videos stream PIP over the call controls. In the mobile responsive view it is sitting on top of buttons in the call controls.

# Why

 Lets just change the default bottom of the PIP to be 5rem (to be above the call controls)

# How Tested
Ran it locally in responsive view with an additional participant to force the local video stream PIP

# Process & policy checklist


alkwa-msft:
What we had before :
image

With the 1rem -> 5rem change
image

packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -27,7 +27,7 @@ export const floatingLocalVideoModalStyle: IStyleFunctionOrObject<IModalStylePro
     boxShadow: theme.effects.elevation8,
     borderRadius: theme.effects.roundedCorner4,
     position: 'absolute',
-    bottom: '1rem',
+    bottom: '5rem',
```
JamesBurnside, line 30: I see this also in the meeting composite where it overlaps the side pane. Is it possible to position this relative to the parent instead?

normally you can set `position: relative` on the parent and the absolute position will use the parent instead of the page here

JamesBurnside: Having the padding bottom also means if the control bar changes this will be incorrect again and if the videogallery is used outside of a control bar it will randomly be floating:
image

(though that could be fixed by passing in value from the composite instead of hardcoding this at the component layer - but if possible it would be better if we could set this relative to the parent so it should move if the parent changes size)

JamesBurnside: Updated, needed to apply a layout as the models by default are rendered at the end of the document, now its rendered inside the video gallery component

Results (the video tile is positioned relative to the video gallery - even when the text box is dynamically resized at runtime):
| test | result |
| -- | -- |
| content to the right hand side | image |
| content below | image |

JamesBurnside: When testing meeting composite, correctly did not overlap control bar or side pane when opened:
image

JamesBurnside (approved):

alcail (approved):

PR #621, opened 2021-07-23
adding in sendbox padding so we dont overlap with the icon

alkwa-msft:
# What
The sendbox icon is getting overlapped with the text for the new message to send

# Why
We need to fix this as it can be difficult to press on the button. We shouldn't be overlapping the text with the button.

# How Tested
Ran it locally

# Process & policy checklist


alkwa-msft:
Before:
image

After:
image

packages/react-components/src/components/styles/SendBox.styles.ts
```diff
@@ -57,7 +57,8 @@ export const sendBoxStyle = mergeStyles({
   },
   ':-moz-placeholder': {
     fontSize: '0.875rem'
-  }
+  },
+  padding: '0 2rem 0 0.5rem'
```
JamesBurnside, line 61: note: this will need a condition for RTL, I think the fix for this should be in the send box component to use a stackbox, something like:
```
<Stack horizontal>
  <Stack.Item grow>{textArea}</StackItem>
  <Stack.Item>{sendIcon}</Stack.Item>
</Stack> 
```
(^that code is speculative, haven't tried but semantically it should be something like that)

JamesBurnside: Turns out RTL flips padding, that's fun!
image
image

JamesBurnside: And doing stacks won't work because the icon needs to be floating inside the textbox to have it encased inside the textbox (and fluentUI doesn't provide clickable icons within text boxes)

JamesBurnside (approved):

alcail (approved):

PR #622, opened 2021-07-23
Fix parameter changes of composites for js bundle

PorterNan:
# What

There is a change for composite parameters: https://github.com/Azure/communication-ui-library/pull/590

This PR is to adjust js bundle function calls to updated parameters
# Why

# How Tested

# Process & policy checklist


alcail (approved):

JamesBurnside (approved):

samples/StaticHtmlComposites/src/callComposite.js
```diff
@@ -3,11 +3,17 @@
 
 import React from 'react';
 import ReactDOM from 'react-dom';
+import { AzureCommunicationTokenCredential } from '@azure/communication-common';
 import { CallComposite, createAzureCommunicationCallAdapter } from '@azure/communication-react';
 
 export const loadCallComposite = async function (args, htmlElement) {
   const { userId, token, groupId, displayName } = args;
-  const adapter = await createAzureCommunicationCallAdapter(userId, token, { groupId }, displayName ?? 'anonymous');
+  const adapter = await createAzureCommunicationCallAdapter(
```
alkwa-msft, line 11: Was there a good reason for moving around the parameters in this function? I feel this is a breaking change? I am fine with a breaking change but do we have a good reason for this?

JamesBurnside: These params should definitely be a named object too so order doesn't not matter (there's too many of them), maybe too late to do for this beta release though

alkwa-msft (approved):

samples/StaticHtmlComposites/src/callComposite.js
JamesBurnside: PR here if we do, will do chat also next, can then decide to cp to this beta or not https://github.com/Azure/communication-ui-library/pull/625

PR #628, opened 2021-07-26
Call Composite Automation Tests

anjulgarg:
# What
Tests for 
- Control Bar Camera Button
- Participant Button
- Video Gallery

# Why

# How Tested

# Process & policy checklist


mgamis-msft (approved):

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -28,7 +27,59 @@ test.describe('Call Composite E2E Tests', () => {
...
+      expect(await page.screenshot()).toMatchSnapshot(`page-${idx}-video-gallery.png`);
+    }
+  });
+
+  test('participant list loads correctly', async ({ pages }) => {
+    for (const idx in pages) {
+      const page = pages[idx];
+      page.bringToFront();
+      await page.click(dataUiId('call-composite-participants-button'));
+      await page.waitForTimeout(1000);
```
alkwa-msft, line 69: I thought we had to wait 1s so we can compare the snapshot to the expected image but we don't always seem to need to do it for all of the tests?

anjulgarg: This here is a special case because clicking on participants icon displays a dropdown menu that has an animation. So it takes it a while to render completely. Hence, the wait of 1second. We don't need to wait a sec for all tests.

alkwa-msft: ok if its a special case perhaps we should put a comment since this is specific around our UI today.

alkwa-msft (approved):

PR #629, opened 2021-07-27
Video device icon change in OptionsButton

mgamis-msft:
# What

Changed icon of video device icon in OptionsButton
https://user-images.githubusercontent.com/79475487/127219993-d942bae2-64b8-4aac-8cd1-f2c4722e009f.mp4

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2525175

# How Tested

# Process & policy checklist


packages/react-components/src/components/OptionsButton.tsx
```diff
@@ -136,7 +138,9 @@ const generateDefaultMenuProps = (
           key: camera.id,
           text: camera.name,
           title: camera.name,
-          iconProps: { iconName: 'Camera' },
+          onRenderIcon: () => (
+            <Video20Regular style={{ color: theme.palette.themePrimary, display: 'flex', alignItems: 'center' }} />
```
alkwa-msft, line 142: can we put this styling into a class?

alkwa-msft (dismissed):
Lets put the inline style into a class.

mgamis-msft:
> Lets put the inline style into a class.

Sure. I have moved it to a separate file.

JamesBurnside (approved):

packages/react-components/src/components/OptionsButton.tsx
```diff
@@ -136,7 +139,7 @@ const generateDefaultMenuProps = (
           key: camera.id,
           text: camera.name,
           title: camera.name,
-          iconProps: { iconName: 'Camera' },
+          onRenderIcon: () => <Video20Regular style={{ ...iconStyle, ...{ color: theme.palette.themePrimary } }} />,
```
JamesBurnside, line 142: Instead of setting the color manually in the style here, you should be able to use `primaryFill="currentColor"` as a prop on `Video20Regular` to ensure we don't lose any hover/active/disabled coloring (which may not apply but best to avoid overwriting the color when necessary, hopefully we'll be able to remove setting primary fill when https://github.com/microsoft/fluentui/issues/17804 is fixed).

mgamis-msft: Unfortunately, it seems like using prop `primaryFill="currentColor` sets the color to `theme.palette.black`.
image
image

JamesBurnside: Ah right because we want the icon to be different from the text - my bad!

prprabhu-ms (approved):

PR #632, opened 2021-07-29
The LocalStream PIP is a pretty substantial delta from our core experience and it should be optional

alkwa-msft:
# What
I want to make the layout of how we render the local stream rendering to be by default in the grid and changeable to a floating PIP through options in the call composite.

# Why
In our video gallery customers started seeing a super different  behavior. The Local stream was not longer in the grid and was in a small picture in picture floating modal.

# How Tested
Locally

# Process & policy checklist

### DIFF
--- a/packages/communication-react/review/communication-react.api.md
+++ b/packages/communication-react/review/communication-react.api.md
@@ -338,6 +338,7 @@ export type CallCompositeProps = {
     locale?: Locale;
     callInvitationURL?: string;
     onRenderAvatar?: (props: PlaceholderProps, defaultOnRender: (props: PlaceholderProps) => JSX.Element) => JSX.Element;
+    options?: CallOptions;
     identifiers?: Identifiers;
 };
 
@@ -371,6 +372,11 @@ export interface CallingTheme {
         callRedDarker: string;
     };
 }
+
+// @public
+export type CallOptions = {
+    showFloatingVideo?: "floatingLocalVideo" | "default";
+};
 
 // @public
 export type CallParticipant = CommunicationParticipant & {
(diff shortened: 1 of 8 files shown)


packages/communication-react/review/communication-react.api.md
```diff
@@ -372,6 +373,11 @@ export interface CallingTheme {
     };
 }
 
+// @public
+export type CallOptions = {
+    showFloatingVideo?: "floatingLocalVideo" | "default";
```
prprabhu-ms, line 378: I'd expect a field named `showFloatingVideo` to be boolean. For an enum, I'd expect something like `videoMode`... but I can't think of a specific enough name for this enum. What values beyond these two do you anticipate in the future?

Also, if this is made boolean, it should be `enableFloatingLocalVideo`, for consistency with Chat options.

packages/react-composites/src/composites/CallComposite/Call.tsx
```diff
@@ -29,18 +29,38 @@ export type CallCompositeProps = {
...
 };
 
+/**
+ * Additional customizations for the chat composite
+ */
+export type CallOptions = {
+  /**
+   * UNSTABLE: Feature flag for the local stream to be in a PIP (picture in picture) floating on the bottom right corner.
+   *
+   * This option will be removed once VideoGallery is stable.
```
prprabhu-ms, line 43: If we intend to remove this option when we mark the `VideoGallery` stable, I'm assuming that PIP mode will be the only choice? In that case, it'll flip default and surprise customers (closer to stable, so worse).

Should we start by defaulting to local video (and providing the option to go back to embedded local video for those customers who don't want this change suddenly)?

prprabhu-ms (approved):
LGTM, pending inversion of default value.

alcail (approved):

PR #633, opened 2021-07-29
FluentThemeProvider prop 'applyTo' exposed. Fix dark theme on IphoneSE.

mgamis-msft:
# What

FluentThemeProvider prop 'applyTo' exposed to give option on scope of theme application. This allows us to fix bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2525645.

# Why

For bug: https://skype.visualstudio.com/SPOOL/_workitems/edit/2525645

# How Tested

Manual testing.
IPhoneSE view on Chrome.
image

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/theming/FluentThemeProvider.tsx
+++ b/packages/react-components/src/theming/FluentThemeProvider.tsx
@@ -14,6 +14,15 @@ export interface FluentThemeProviderProps {
   children: React.ReactNode;
   /** Theme for components. Defaults to a light theme if not provided. */
   fluentTheme?: PartialTheme | Theme;
+  /**
+   * Defines where body-related theme is applied to.
+   * Setting to 'element' will apply body styles to the root element of ThemeProvider.
+   * Setting to 'body' will apply body styles to document body.
+   * Setting to 'none' will not apply body styles to either element or body.
+   *
+   * @defaultvalue 'element'
+   */
+  applyTo?: 'element' | 'body' | 'none';
 }
 
 const wrapper = mergeStyles({
@@ -43,7 +52,7 @@ const initialFluentNorthstarTheme = mergeNorthstarThemes(teamsTheme, {
  * This provider handles applying any theme provided to both the underlying Fluent UI controls, as well as the Fluent React Northstar controls.
  */
 export const FluentThemeProvider = (props: FluentThemeProviderProps): JSX.Element => {
-  const { fluentTheme, children } = props;
+  const { fluentTheme, applyTo, children } = props;
   // if fluentTheme is not provided, default to light theme
   const fluentUITheme: Theme = mergeThemes(defaultTheme, fluentTheme);
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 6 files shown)


alkwa-msft:
We just had some major issues during lets build around having css on the body. Is there a way we could avoid setting things on the body in this PR?

JamesBurnside:
> We just had some major issues during lets build around having css on the body. Is there a way we could avoid setting things on the body in this PR?

@mgamis-msft correct me if I'm wrong but I believe this just exposes the fluent's theme provider option of applying theme at the body level - but by default it will still apply to `element` that it wraps.

Then this PR updates just the samples to apply theme at the body level (fixing the page level background color)

JamesBurnside (approved):

mgamis-msft:
> > We just had some major issues during lets build around having css on the body. Is there a way we could avoid setting things on the body in this PR?
> 
> @mgamis-msft correct me if I'm wrong but I believe this just exposes the fluent's theme provider option of applying theme at the body level - but by default it will still apply to `element` that it wraps.
> 
> Then this PR updates just the samples to apply theme at the body level (fixing the page level background color)

Yes that is great summary of what this PR is. Thank you!

mgamis-msft:
> > We just had some major issues during lets build around having css on the body. Is there a way we could avoid setting things on the body in this PR?
> 
> @mgamis-msft correct me if I'm wrong but I believe this just exposes the fluent's theme provider option of applying theme at the body level - but by default it will still apply to `element` that it wraps.
> 
> Then this PR updates just the samples to apply theme at the body level (fixing the page level background color)

Yes that is great summary of what this PR is. Thank you!

mgamis-msft:
> > We just had some major issues during lets build around having css on the body. Is there a way we could avoid setting things on the body in this PR?
> 
> @mgamis-msft correct me if I'm wrong but I believe this just exposes the fluent's theme provider option of applying theme at the body level - but by default it will still apply to `element` that it wraps.
> 
> Then this PR updates just the samples to apply theme at the body level (fixing the page level background color)

Yes, that is great summary of what this PR is about. Thank you!

packages/react-components/src/theming/FluentThemeProvider.tsx
```diff
@@ -14,6 +14,15 @@ export interface FluentThemeProviderProps {
...
   fluentTheme?: PartialTheme | Theme;
+  /**
+   * Defines where body-related theme is applied to.
+   * Setting to 'element' will apply body styles to the root element of ThemeProvider.
+   * Setting to 'body' will apply body styles to document body.
+   * Setting to 'none' will not apply body styles to either element or body.
+   *
+   * @defaultvalue 'element'
+   */
+  applyTo?: 'element' | 'body' | 'none';
```
JamesBurnside, line 25: instead of manually defining this, is there a type exposed for this we could use from the ThemeProvider's package? I can't find one so very possible there isn't

mgamis-msft: There is not unfortunately. I basically just copied their comment and typing.
image

JamesBurnside: Thanks for checking this! Food for thought (not in this PR): should we avoid inline string literal definitions and instead export them as types so they can be imported and used by customers?

mgamis-msft: That is a good point but I have now created a new PR for a better fix here https://github.com/Azure/communication-ui-library/pull/641. Sorry for the sudden change but thank you for reviewing. I will close this PR.

mgamis-msft:
There is a better and simpler solution on this PR https://github.com/Azure/communication-ui-library/pull/641.

PR #635, opened 2021-07-30
Alcail/storybook from knobs to controls

alcail:
# What
Replaced Knobs by Controls add-on
I updated Storybook package from 6.1 to 6.2 but not 6.3 yet as the last one asks for typescript 4.3 (not sure if we want to go there yet)

# Why
as Knobs add-on is being deprecated in the new versions of Storybook, controls add-on will take over

# How Tested
ran storybook and tested controls in each story

with knobs:
image

with controls:
image

# Process & policy checklist

updating Storybook package version


packages/storybook/stories/ChatComposite/CustomDataModelExample.stories.tsx
```diff
@@ -17,7 +17,7 @@ import { ConfigHintBanner, addParrotBotToThread } from './snippets/Utils';
 
 const avatars = ['Default', 'Cat', 'Fox', 'Koala'];
 
-const getAvatarSymbol = (AvatarName: string) => {
+const getAvatarSymbol = (AvatarName: string): any => {
```
alkwa-msft, line 20: can we  return a string for this?

I want to avoid "any" if possible

packages/storybook/stories/CallComposite/JoinExistingCall.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+};
+
+export const JoinExistingCall = JoinExistingCallStory.bind({});
+
+export default {
+  id: `${COMPOSITE_FOLDER_PREFIX}-call-joinexistingcall`,
+  title: `${COMPOSITE_FOLDER_PREFIX}/CallComposite/Join Existing Call`,
+  component: CallComposite,
+  argTypes: {
+    callLocator: { control: 'text', defaultValue: '', name: 'Call locator (ACS group ID or Teams meeting link)' },
```
JamesBurnside, line 51: Is it possible to set placeholder text here? Instead of "Adjust string dynamically" 
image

(If its possible, doesn't have to updated in this PR, more just curious but if possible would be nice for us to update it at some point)

packages/storybook/stories/CallComposite/JoinExistingCall.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+    userId: { control: 'text', defaultValue: '', name: 'User identifier for user' },
+    token: { control: 'text', defaultValue: '', name: 'Valid token for user' },
+    displayName: { control: 'text', defaultValue: '', name: 'Display Name' },
+    callInvitationURL: {
+      control: 'text',
+      defaultValue: '',
+      name: 'Optional URL to invite other participants to the call'
+    },
+    // Hiding auto-generated controls
```
JamesBurnside, line 60: Should `locale` also be hidden here?
image
Alternatively it could be a dropdown of selectable locales

JamesBurnside: Is there a feature to auto hide default controls?

JamesBurnside: (Someone might add a prop to the composite and it show as a control)

packages/storybook/stories/CallComposite/JoinExistingCall.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export default {
+  id: `${COMPOSITE_FOLDER_PREFIX}-call-joinexistingcall`,
+  title: `${COMPOSITE_FOLDER_PREFIX}/CallComposite/Join Existing Call`,
+  component: CallComposite,
+  argTypes: {
+    callLocator: { control: 'text', defaultValue: '', name: 'Call locator (ACS group ID or Teams meeting link)' },
+    userId: { control: 'text', defaultValue: '', name: 'User identifier for user' },
+    token: { control: 'text', defaultValue: '', name: 'Valid token for user' },
+    displayName: { control: 'text', defaultValue: '', name: 'Display Name' },
```
JamesBurnside, line 54: A lot of these are duplicated in lots of stories, suggestion would be to have a single place to put them in and reference from, i.e.
```
/// utils/controls.ts

export const displayNameControl = { control: 'text', defaultValue: '', name: 'Display Name' }
```

We could also have the hidden ones in a constant place also:
```
export const defaultCallCompositeControlsToHide = {
    adapter: { control: false, table: { disable: true } },
    fluentTheme: { control: false, table: { disable: true } },
    onRenderAvatar: { control: false, table: { disable: true } },
    identifiers: { control: false, table: { disable: true } }
  },
```

JamesBurnside:
Question - given "Controls" is our only tab section in this area, is there an API to remove that heading?
image
(^i.e. remove the `Controls (3)` bar)

packages/storybook/stories/CallComposite/JoinExistingCall.stories.tsx
JamesBurnside: ok nope there seems to not be a way currently: https://stackoverflow.com/questions/62641599/storybook-addon-controls-howto-not-auto-generate-control-for-a-certain-prop 
Probably worth filing an issue on them for though!

JamesBurnside (approved):
This awesome!! Huge effort updating all these! Left some suggestion comments, but not blocking and happy for them to be ignored if they're more work than worth!

packages/storybook/stories/CallComposite/JoinExistingCall.stories.tsx
alcail: Lol, yep, there's currently no way to automatically hide controls

PR #640, opened 2021-08-03
changelog for beta3

alkwa-msft:
# What
We need to describe the changes in 1.0.0-beta3

# Why
Developers need to understand what has changed. There are two breaking changes in our API around composites and it is important to point these out.

# How Tested
Ran locally

# Process & policy checklist

Yes, there is an API change in the composites. Since there are a number of parameters in the composite that are growing, we decided to put it into an object of properties.

We are also using the `CommunicationTokenCredential` over keep the token as a string

### DIFF
--- a/packages/communication-react/CHANGELOG.md
+++ b/packages/communication-react/CHANGELOG.md
@@ -11,7 +11,20 @@ Thu, 22 Jul 2021 17:42:41 GMT
 
 ### Changes
 
-- Bump @internal/chat-stateful-client to v1.0.0-beta.3 ([PR #554](https://github.com/azure/communication-ui-library/pull/554) by 2684369+JamesBurnside@users.noreply.github.com)
+* [Breaking change] Call Composite and Chat Composite adapters now take in a `CommunicationTokenCredential` in the constructor instead of the token
+
+* [Breaking change] Call Composite and Chat Composite adapters take in an object containing all of the parameters instead of passing in the parameters individually
+
+* Fixing the sendbox component where the padding would overlap with the icon
+
+* Fixing the local preview overlapping the call control bar
+
+* Setting minHeight on the call composite media gallery
+
+* Adding a small fix for supporting typescript < 4.1
+
+### Storybook page changes
+* Fixing link in "Using Composites in a non-react environment"
 
 ## [1.0.0-beta.2](https://github.com/azure/communication-ui-library/tree/@azure/communication-react_v1.0.0-beta.2)
 


packages/communication-react/CHANGELOG.md
```diff
@@ -11,7 +11,20 @@ Thu, 22 Jul 2021 17:42:41 GMT
...
+* [Breaking change] Call Composite and Chat Composite adapters take in an object containing all of the parameters instead of passing in the parameters individually
+
+* Fixing the sendbox component where the padding would overlap with the icon
+
+* Fixing the local preview overlapping the call control bar
+
+* Setting minHeight on the call composite media gallery
+
+* Adding a small fix for supporting typescript < 4.1
+
```
JamesBurnside, line 25: Some extra things we could consider adding:
react-components:
- React hook useTheme added to ACS library.
- Fix truncation from MessageThread system message.
- Add default call control bar button for creating custom call control buttons
- Fix cursor on control button labels.
- Add identifier injection mechanism in components for e2e testing
- Exporting locales as separate entites.
- Localized strings for SendBox and MessageStatusIndicator
- Fix MessageThread hook related console errors.
- Add an ErrorBar component
react-composites:
- Make participant pane in the chat composite optional. Disabled by default.
- Chat composite topic heading is now optional. Disabled by default.
- Add ErrorBar in ChatComposite behind a feature flag
- Bugfix to prevent options menu from getting hidden every time a participant joins or leaves
- Drop duplicate thread status banner

JamesBurnside: Also: Publish `js` files to use composites outside of a react app as part of the GitHub release

packages/communication-react/CHANGELOG.md
```diff
@@ -11,7 +11,20 @@ Thu, 22 Jul 2021 17:42:41 GMT
...
+### Storybook page changes
```
JamesBurnside, line 26: Do we want to include storybook changes going forward? Given this changelog is for the npm package it's not necessary but I do see value still. If we do want to we should require change files for changes in the storybook package also

anjulgarg (approved):

packages/communication-react/CHANGELOG.md
alkwa-msft: Oh right. 

I'll add those. I think for now till we're ready to share more about the ErrorBar and Localization we will exclude them from these notes for now.

packages/communication-react/CHANGELOG.md
alkwa-msft: Thats a good point I think if theres significant changes we'll create a storybook section for now ill just add this one item under the story book changes to be under "Changes" instead

JamesBurnside (approved):

PR #649, opened 2021-08-06
Bugfix for [2537130] and [2475620]

anjulgarg:
# What
Fixes the following bugs
- https://skype.visualstudio.com/SPOOL/_workitems/edit/2537130
- https://skype.visualstudio.com/SPOOL/_workitems/edit/2475620

# Why

# How Tested

# Process & policy checklist


alkwa-msft (approved):

prprabhu-ms (approved):
w00t!

packages/calling-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -220,7 +223,7 @@ export const createDefaultCallingHandlers = memoizeOne(
       }
 
       if (screenShareStream && screenShareStream.isAvailable && !screenShareStream.view) {
-        callClient.createView(call.id, participant.identifier, screenShareStream, options);
+        callClient.createView(call.id, participant.identifier, screenShareStream, { scalingMode: 'Fit' });
```
anjulgarg, line 226: hardcoded this here since it is highly unlikely that CONTOSO would ever want to use a different scaling mode for screenshare.
Using `Crop` would crop the contents of screenshare and `Stretch` would warp it.
`Fit` is the only mode that maintains the integrity of the screen being shared.

prprabhu-ms: Please add this comment to the code for posterity.

PR #651, opened 2021-08-06
Fix issue where screenshare dialogs eats user input across the whole web page

JamesBurnside:
# What
Host the screen share dialog model in a div element inside the composite

# Why
For some users, the screenshare dialog was eating input across the whole screen.
This was because the body html element (that the modal dialog is hosted in) had:
```css
body { display: flex }
```
Which messed up Fluent's modal dialog hosting.

By instead hosting the dialog inside a component in our composite we get two benefits:
* We constrain the dialog to be within our composite
* We control the dialog's host div's css and hence customers won't hit this issue

# How Tested
In a test app. Was able to repro the problem then build a local package with these changes and test that the screenshare dialog no longer eats all input.


packages/react-composites/src/composites/CallComposite/CallScreen.tsx
```diff
@@ -115,6 +115,7 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     }
   }
 
+  const screenShareHostHostId = 'UILibraryMediaGallery';
```
alcail, line 118: nit: why the double 'Host' in the name?

alcail (approved):

packages/react-composites/src/composites/CallComposite/CallScreen.tsx
JamesBurnside: typo! :D I'll update that

JamesBurnside: Updated!

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -201,7 +201,7 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
   ]);
 
   if (shouldFloatLocalVideo()) {
-    const floatingTileHostId = 'UILibaryFloatingTileHost';
+    const floatingTileHostId = 'UILibraryFloatingTileHost';
```
anjulgarg, line 204: what are these ids for?

JamesBurnside: They are referenced by the layerprop hostid: https://developer.microsoft.com/en-us/fluentui#/controls/web/layer#ILayerProps

For fluent modal dialogs they essentially say, render the element in the element with this ID and not in the default location (which is at the end of the DOM directly under body)

anjulgarg (approved):

PR #654, opened 2021-08-07
Fix show moderator

PorterNan:
# What

System message will show moderator join, which is not the right logic
This 'filter' logic shares the logic of what we do in participant list (filter no name participant as moderator)
(moderator logic could be a pattern problem, which we should have a better solution to identify it)
# Why

# How Tested

# Process & policy checklist


alcail (approved):

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -204,22 +204,28 @@ const DefaultSystemMessageRenderer: DefaultMessageRendererType = (props: Message
...
     if (payload.type === 'participantAdded') {
-      return (
-        <SystemMessageComponent
-          iconName={(payload.iconName ?? '') as SystemMessageIconTypes}
-          content={generateParticipantsStr(payload.participants) + strings.messageThread.participantJoined}
-          containerStyle={props?.messageContainerStyle}
-        />
-      );
+      const participantsStr = generateParticipantsStr(payload.participants);
+      if (participantsStr !== '') {
```
JamesBurnside, line 208: nit: can you add a comment here that this is here to filter out _our_ moderator that has no display name

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -204,22 +204,28 @@ const DefaultSystemMessageRenderer: DefaultMessageRendererType = (props: Message
...
+        return (
+          <SystemMessageComponent
+            iconName={(payload.iconName ?? '') as SystemMessageIconTypes}
+            content={participantsStr + strings.messageThread.participantLeft}
+            containerStyle={props?.messageContainerStyle}
+          />
+        );
+      }
```
smsohan, line 228: These two code blocks look somewhat similar. So, we have a possibility to combine these into one. Also notice that we can make payload.type and the I18n key similar. `participantAdded` vs `strings.messageThread.participantJoined` and `participantRemoved` vs `strings.messageThread.participantLeft`

If these names were consistent, we could DRY things up as follows:

```TypeScript
if (payload.type === 'participantJoined' || payload.type === 'participantRemoved') {
       const participantsStr = generateParticipantsStr(payload.participants);
       if (participantsStr !== '') {
         return (
           <SystemMessageComponent
             iconName={(payload.iconName ?? '') as SystemMessageIconTypes}
             content={participantsStr + strings.messageThread[payload.type]}
             containerStyle={props?.messageContainerStyle}
           />
         );
       }
     }
}
```

The benefit of collapsing the two logical blocks into one is the mental model. Looking at this code one can tell it handles `participantJoined` and `participantRemoved` in the same manner.

Thoughts?

PorterNan: Thanks for the comment, I am thinking now moving this to selector code so this component will be ACS unrelated

PR #658, opened 2021-08-10
Refactor to apply rtl only within scope.

mgamis-msft:
# What

rtl prop added to FluentThemeProvider so that we only apply rtl only within scope. Removed usage of `setRTL` from fluentui to avoid applying rtl to the entire html root.

Note: There is a minor bug on FluentUI where some icon margins are not horizontally "flipped" like below:
Icon margins (known)
logged in issue:
https://github.com/microsoft/fluentui/issues/18836

# Why

- To fix docs being affected by rtl in storybook
- To limit rtl when users potentially embed our composites/components into their apps, as discussed in Let's Build session

# How Tested

Tested stories and samples.

# Process & policy checklist


alcail (approved):

packages/react-components/src/theming/FluentThemeProvider.tsx
```diff
@@ -66,11 +71,6 @@ export const FluentThemeProvider = (props: FluentThemeProviderProps): JSX.Elemen
       ChatMessage: {
         timestamp: {
           WebkitTextFillColor: fluentUITheme.palette.neutralSecondary
-        },
```
prprabhu-ms, line 69: Setting `rtl` through the Provider also gets rid of all these northstar bugs?

prprabhu-ms (approved):

JamesBurnside (approved):

packages/react-components/src/theming/FluentThemeProvider.tsx
mgamis-msft: Yes. I should have set 'rtl' property instead of the 'dir' property on line 83.

PR #661, opened 2021-08-10
Make calling stateful unittests damp

prprabhu-ms:
# What
Refactors about half of the unittests in the calling stateful package to:
* Remove use of global state
* Replace the `testData` fixture that was updated opaquely by test helpers with functional object creation helpers. This makes the tests a lot more readable.
* Remove some boilerplate duplication by merging the tests where it made sense.
* Delete a couple god mocks that made it extremely difficult to both understand the tests and to add more tests. Replaced with test-specific fakes created with the functional object creation test helpers.
* Hoist the tests for DeviceManager to test the API rather than the implementation (Context/InnerContext). There are still unittests left that should be hoisted similarly... CallAgent etc.
* Add expectations about the number of events emitted. Found a few places where we're emitting spurious events. This PR doesn't fix those. Added a `FIXME` comment instead.

This PR paves the way to add more unittests around error handling without making the test setup unmanageable.

# How Tested
`rush test`

# Process & policy checklist


packages/calling-stateful-client/src/DeviceManager.test.ts
```diff
@@ -0,0 +1,233 @@
...
+    const deviceAccess = await (await client.getDeviceManager()).askDevicePermission({ audio: false, video: false });
+    expect(deviceAccess.audio).toBe(true);
+    expect(deviceAccess.video).toBe(true);
+    expect(client.getState().deviceManager.deviceAccess).toBeDefined();
+    expect(client.getState().deviceManager.deviceAccess?.audio).toBe(true);
+    expect(client.getState().deviceManager.deviceAccess?.video).toBe(true);
+  });
+});
+
+const createStatefulCallClientWithDeviceManager = (deviceManager: MockDeviceManager): StatefulCallClient => {
```
JamesBurnside, line 164: Does it make sense to move here and below to a helper file? (e.g. `DeviceManagerTestUtils` or something?) to help reduce the size of the file

packages/calling-stateful-client/src/DeviceManager.test.ts
```diff
@@ -0,0 +1,233 @@
...
+      [speakerWithName('secondSpeaker')],
+      createMockDeviceManagerWithMicrophones([microphoneWithName('firstMicrophone')])
+    );
+    const client = createStatefulCallClientWithDeviceManager(manager);
+    // FIXME? Device manager is only created on request, so state is not updated unless device manager is created.
+    await client.getDeviceManager();
+
+    manager.emit('audioDevicesUpdated', {});
+    expect(
+      await waitWithBreakCondition(() => client.getState().deviceManager.microphones[0]?.name === 'firstMicrophone')
```
JamesBurnside, line 137: I forget - what's the reason we need to wait literal milliseconds here? Is there a way to mock timers to avoid it?
(This isn't a change of this PR I know so this is not a blocking comment - just want to get rid of arbitrary waits 😅 )

packages/calling-stateful-client/src/TestUtils.ts
```diff
@@ -5,7 +5,9 @@ import {
   Call,
```
JamesBurnside, line 5: Instead of a single `TestUtils` fie that becomes a dumping ground for all util functions, can we have a `/testutils/` folder that can then have mutiple more-specific test util files

packages/calling-stateful-client/src/StatefulCallClient.ts
```diff
@@ -234,10 +234,23 @@ export const createStatefulCallClient = (
...
-  const internalContext: InternalCallContext = new InternalCallContext();
+  return createStatefulCallClientWithDeps(
+    new CallClient(options?.callClientOptions),
+    new CallContext(args.userId, options?.maxStateChangeListeners),
+    new InternalCallContext()
+  );
+};
 
+/**
+ * Package-internal version of createStatefulCallClient that allows dependency injection.
```
JamesBurnside, line 245: Nice!!

packages/calling-stateful-client/src/DeviceManager.test.ts
prprabhu-ms: Renaming suggestions welcome. This isn't actually a wait. This is more a wait-till-condition-becomes-true-or-timeout.
This is needed so that we can assert conditions that become true once asynchronous operations complete.
So, when the test is passing, the expected wait is ~0 second.

prprabhu-ms: Incidentally, this PR fixes a bug where this helper was always waiting 100 milliseconds before checking for the first time.

packages/calling-stateful-client/src/DeviceManager.test.ts
prprabhu-ms: Don't have strong opinion on this.

Moving functions around is easy :)

packages/calling-stateful-client/src/TestUtils.ts
prprabhu-ms: +1
 
Probably better to do that as follow up PR though.

packages/calling-stateful-client/src/DeviceManager.test.ts
JamesBurnside: I guess amortized 0 seconds is acceptable :P (though my worry is it sets a precedence where the codebase feels its acceptable to have wait-x-milliseconds conditions)

JamesBurnside (approved):

packages/calling-stateful-client/src/DeviceManager.test.ts
prprabhu-ms: Realized after commenting here that the first check is likely going to always fail -- the async operations haven't had a chance to run.
So... we'll still waste 100 milliseconds.

But hey, it's just the test machine's CPU. After some experimentation, 10ms is a better wait time on my workstation. The test runtime goes down from ~20 seconds to ~2 seconds.

So... there.

packages/calling-stateful-client/src/DeviceManager.test.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  test('should detect videoDevicesUpdated and update cameras in state using deviceManager', async () => {
+    const manager = createMockDeviceManagerWithCameras([cameraWithName('camera')]);
+    const client = createStatefulCallClientWithDeviceManager(manager);
+    // FIXME? Device manager is only created on request, so state is not updated unless device manager is created.
+    await client.getDeviceManager();
+
+    manager.emit('videoDevicesUpdated', {});
+    expect(await waitWithBreakCondition(() => client.getState().deviceManager.cameras.length === 1)).toBe(true);
+    expect(client.getState().deviceManager.cameras[0].name).toBe('camera');
```
mgamis-msft, line 47: Thanks for taking the time to write these unit tests. Could you also check that `client.getState().deviceManager.cameras[0].name` to be 'camera'?

prprabhu-ms: That's line 47. Did I miss something?

PR #662, opened 2021-08-10
Fix partially composed chat messages being lost

JamesBurnside:
# What
Only hide the side panes in the meeting composite instead of not rendering them completely

# Why
Partially composed chat messages are lost when the chat pane is opened and closed in meeting composite

# How Tested
In storybook > Meeting Composite > Basic Example > Open Chat > Type Message > Close Chat > Open Chat -- typed message is still there

### DIFF
--- a/packages/react-components/src/components/styles/SendBox.styles.ts
+++ b/packages/react-components/src/components/styles/SendBox.styles.ts
@@ -41,7 +41,7 @@ export const suppressIconStyle = {
 };
 
 export const sendBoxStyle = mergeStyles({
-  minHeight: '0',
+  minHeight: '2.25rem', // prevents the input text box from being sized to 0px when the meeting composite chat pane is closed.
   maxHeight: '8.25rem',
   outline: 'red 5px',
   fontWeight: 400,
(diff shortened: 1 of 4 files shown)


prprabhu-ms (approved):

packages/react-components/src/components/styles/SendBox.styles.ts
```diff
@@ -41,7 +41,7 @@ export const suppressIconStyle = {
 };
 
 export const sendBoxStyle = mergeStyles({
-  minHeight: '0',
+  minHeight: '2.25rem', // prevents the input text box from being sized to 0px when the meeting composite chat pane is closed.
```
JamesBurnside, line 44: Screenshot for context:
image

PorterNan, line 44: So this number won't back to what it is when the display  = show?

PorterNan (approved):

packages/react-components/src/components/styles/SendBox.styles.ts
prprabhu-ms: This is `minHeight`. I think the idea is to avoid going down to 0 for hidden component so that internal text isn't lost.

When it's shown again, the flex container will fill up the given space (so larger than 2.25 rem).

Is my understanding correct, @JamesBurnside ?

JamesBurnside: This is to avoid (what looks like) an edge case fluent bug, when we set display to `none` on the side pane, when fluent does its calculations for the textbox height (because the textbox has `autoSizeHeight = true` to resize when there are multiple lines) it sets the height to 0px explicitly causing this:
image

By setting min-height here we subvert that issue (note the textbox still has a height of 0px being applied to it but the min-height makes that redundant). 

In the collapsed scenario when you start typing in the 0px it springs up the correct size implying its definitely a fluent bug

PR #669, opened 2021-08-11
Start sending canned application ID with chat user agent

prprabhu-ms:
# What
* Adds a canned application ID to the chat user agent.
Still need to figure out the right way to extract our own package version, but for now this will let us validate the pipeline end-to-end.


packages/acs-ui-common/src/telemetry.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// TODO: Find a way to import version from ../package.json
+// This package is rooted at acs-ui-common/src so acs-ui-common/package.json is not accessible.
+const version = '0.0.0-alpha';
```
prprabhu-ms, line 6: Any ideas for reading version off of `package.json`

Doing this:

```
  import { version } from '../package.json`
```

doesn't work because the package is rooted at `acs-ui-common/src`. Importing from `acs-ui-common/package.json` messes with `tsc` (it ends up rooting the package at `acs-ui-common`.

JamesBurnside: Probably will need a prebuild step for this that writes this to a `currentVersion.ts` or something similar that this file can export from, don't think tsc can do this.

I worry about `require()` any file that is not meant to be part of the output as the output code will keep that require line and hence the package.json will need to be in the right place in the packaged up dist folder.

packages/chat-stateful-client/src/StatefulChatClient.ts
```diff
@@ -135,7 +136,7 @@ export const createStatefulChatClient = (
   options?: StatefulChatClientOptions
 ): StatefulChatClient => {
   return createStatefulChatClientWithDeps(
-    new ChatClient(args.endpoint, args.credential, options?.chatClientOptions),
+    new ChatClient(args.endpoint, args.credential, addApplicationId(options?.chatClientOptions)),
```
anjulgarg, line 139: If I wanted to add more options here. I would have to rewrite this function `addApplicationId` or call it inside another.
I am not seeing a lot of value in this mechanism.
Might be better to just explicitly use here.
```
{
    ...options,
    userAgentOptions: { userAgentPrefix: getApplicationId() }
}
```

Am I missing something?

packages/acs-ui-common/src/telemetry.ts
JamesBurnside: Alternatively you could use a package like this: https://github.com/indexzero/node-pkginfo
That crawls up the directory and uses the first package.json it finds - but not ideal

JamesBurnside: Other option:
* hard code into a file
* create/overwrite a `rush bump` command
* have it perform `beachball bump` && `manual command to update hard coded value`
* Add unit test to ensure this hard coded value always matches the package.json

prprabhu-ms: Yeah, seems like must choose the "least bad". I was thinking of the build step approach. Better than any non-conventional setup at runtime as it quickly blows up due to the different ways our library gets used.

JamesBurnside:
Do we currently send telemetry when doing development work? (don't want to impact signals gathered from this telemetry)

packages/chat-stateful-client/src/StatefulChatClient.ts
prprabhu-ms: No reason. This used to be a fatter function but I extracted that into acs-ui-common.

In any case, thanks to your comment I realized that I should be passing in the modified options in both arguments. Updated.

prprabhu-ms:
> Do we currently send telemetry when doing development work? (don't want to impact signals gathered from this telemetry)

Telemetry from development / test work is OK as long as we have a handle on the connection strings being used. Those can then be excluded from analysis.

packages/acs-ui-common/src/telemetry.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const sanitize = (version: string): string => {
```
JamesBurnside, line 8: nit: add a comment what is being sanitized (not obvious that this will strip off everything after `alpha`)

JamesBurnside (approved):
LGTM! Reset my vote if you add the correct version collection in this PR and not a seperate one

prprabhu-ms:
> LGTM! Reset my vote if you add the correct version collection in this PR and not a seperate one

Thanks much. I'm lazy, I'm not going to add correct version collection in this PR. I want _something_ in for end-to-end testing, and that thing you ask for -- it's work, sir, it's work indeed.

anjulgarg (approved):

PR #671, opened 2021-08-12
Add Call Composite Remove Participant

JamesBurnside:
# What
* Ensure the ParticipantList inside the ParticipantButton is getting the necessary handlers. In this case it is `onParticipantRemove`.

# Why
Adds in the option to remove participants in the people list in the participant button

# How Tested
In Calling Sample (ignore the RTL, that is a separate bug that crept into main branch):
image
image

### DIFF
--- a/packages/calling-component-bindings/src/handlers/createHandlers.ts
+++ b/packages/calling-component-bindings/src/handlers/createHandlers.ts
@@ -33,7 +33,7 @@ export type DefaultCallingHandlers = {
   onHangUp: () => Promise<void>;
   onCreateLocalStreamView: (options?: VideoStreamOptions) => Promise<void>;
   onCreateRemoteStreamView: (userId: string, options?: VideoStreamOptions) => Promise<void>;
-  onParticipantRemove: (userId: string) => void;
+  onParticipantRemove: (userId: string) => Promise<void>;
   onDisposeRemoteStreamView: (userId: string) => Promise<void>;
   onDisposeLocalStreamView: () => Promise<void>;
 };
@@ -270,8 +270,8 @@ export const createDefaultCallingHandlers = memoizeOne(
       }
     };
 
-    const onParticipantRemove = (userId: string): void => {
-      call?.removeParticipant(fromFlatCommunicationIdentifier(userId));
+    const onParticipantRemove = async (userId: string): Promise<void> => {
+      await call?.removeParticipant(fromFlatCommunicationIdentifier(userId));
     };
 
     return {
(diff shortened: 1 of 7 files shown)


packages/calling-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -270,8 +270,8 @@ export const createDefaultCallingHandlers = memoizeOne(
       }
     };
 
-    const onParticipantRemove = (userId: string): void => {
-      call?.removeParticipant(fromFlatCommunicationIdentifier(userId));
+    const onParticipantRemove = async (userId: string): Promise<void> => {
```
smsohan, line 273: Not sure how we handle errors in such `async` calls. If this call fails, would the UI show an error?

JamesBurnside: @prprabhu-ms is currently working on the error handling but two things should happen:
1. An onError event should be fired back to the adapter such that Contoso can handle/log/do whatever with that
1. If appropriate show a user facing error message in the banner. These banners make sense for the likes of accessing the camera but will be a judgment call on our part if we want to tell the user certain system errors like these

alcail (approved):

packages/react-composites/src/composites/CallComposite/CallControls.tsx
```diff
@@ -28,6 +29,14 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
   const cameraButtonProps = usePropsFor(CameraButton);
   const screenShareButtonProps = usePropsFor(ScreenShareButton);
   const participantsButtonProps = usePropsFor(ParticipantsButton);
+
+  // TODO: Here we grab the participant list props and add them to the
```
JamesBurnside, line 33: This will be handled in a seperate PR

packages/calling-component-bindings/src/handlers/createHandlers.ts
prprabhu-ms: This will be handled by the in-progress error handling.
* Error will be piped to `CallClientState.latestErrors`
* For errors we want to show the user, there will be bindings such that they show up via `ErrorBar`
* Errors will also be saved in adapter state and thrown by adapter method (by doing nothing -- it'll thrown by the adapter method after this PR, the problem is that handling the errors in adapter methods is hard. Likely the method is called from some click handler or useEffect block).

This setup is already in place for Chat. I'm setting it up for Calling in the next couple days.

prprabhu-ms: This is almost correct: https://github.com/Azure/communication-ui-library/blob/main/docs/architecture/ErrorReporting.md

Will do another pass once implementation is complete to fix up any places where plans didn't survive contact with the reality.

prprabhu-ms (approved):

PR #672, opened 2021-08-12
Update GitHub action to not comment about visual regression tests if build fails

JamesBurnside:
# What
Quick fix to ensure github action only comments about visual regression tests when the tests fail and not the build fails

# Why
When the packages build fails you would get a PR comment saying that you should update the UI tests. This comment should only happen if the _tests_ fail and not if the build step fails

# How Tested
Many CI runs, verified:
* All passes - no comment
* Build failes - no comment
* Build passes but tests fail - adds the comment

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,13 +169,13 @@ jobs:
         env:
           CONNECTION_STRING: ${{ secrets.CONNECTION_STRING }}
       - name: Upload snapshot diff
-        if: ${{ steps.visualregressiontests.outcome != 'success' }}
+        if: ${{ always() && steps.visualregressiontests.outcome != 'success' }}
         uses: actions/upload-artifact@v2
         with:
           name: snapshots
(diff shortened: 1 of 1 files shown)


PorterNan (approved):
Thanks for improving this one!

.github/workflows/ci.yml
```diff
@@ -160,19 +160,20 @@ jobs:
       - name: Build
         run: rush build -t sample-static-html-composites
       - name: Visual Regression Tests
+        id: visualregressiontests
```
PorterNan, line 163: Do we wanna add - between words

JamesBurnside: I just wasn't confident that github actions wouldn't break when doing `steps.visual-regression-tests.outcome` 😅

JamesBurnside: (completing without making this change as keeping this to a v. quick fix and waiting for the ci to run a couple times to test is just too much 😭)

PR #677, opened 2021-08-12
Custom Avatar Data Model Overhaul

anjulgarg:
# What
A draft of a design for custom avatar data injection into chat composite.
If everyone is onboard with this pattern, I will implement this everywhere it can be applied inside composites.

Look at `packages/storybook/stories/ChatComposite/snippets/Container.snippet.tsx`  in this PR for usage example

Additionally
- Moved `AvatarPersona` inside composite.
- Created a type `OnRenderAvatarType` and used it everywhere `onRenderAvatar` function is used to avoid duplication of code.
- onRender functions now also supply the callback with `options` and `defaultOnrender` 
- These options can be used to render a Persona in a default fashion.

Custom Image
image

Custom Initials
image

Custom Avatar BG and Initials Color
image

# Why

# How Tested

# Process & policy checklist


packages/storybook/stories/ChatComposite/snippets/Container.snippet.tsx
```diff
@@ -3,10 +3,31 @@ import {
...
+  ChatComposite,
+  createAzureCommunicationChatAdapter,
+  AvatarPersonaDataProvider,
+  AvatarPersonaData
+} from '@azure/communication-react';
 import { Theme, PartialTheme } from '@fluentui/react';
+import _ from 'lodash';
 import React, { useState, useEffect, useMemo } from 'react';
 
+// Sample usage of Custom Data Provider for AvatarPersona with Memoization.
```
anjulgarg, line 17: This piece of code is here as an example code for PR reviewers.
I will remove this once I move this PR out of a draft.

Question for folks. How should we best showcase this to CONTOSO that they need to take care of memoization themselves (possibly using this mechanism)?

packages/react-components/src/components/AvatarPersona.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * An Avatar component made using the `Persona` component.
+ * It allows you to specify a `userId` and a `dataProvider` to retrieve the `AvatarPersonaData`.
+ * Read more about `Persona` component at https://developer.microsoft.com/en-us/fluentui#/controls/web/persona
+ */
+export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
+  const { userId, dataProvider } = props;
+  const [data, setData] = React.useState<AvatarPersonaData | undefined>();
+
+  useEffect(() => {
+    (async () => {
```
alkwa-msft, line 61: If we're saying the Provider for avatars is a react style provider it will change will force a refresh on anything registered to it. However if it does it also implies that the react provider for avatars will always "synchronously" have all of the avatar information for every user. So we can just directly use props from our avatar provider reference?

anjulgarg: The provider for avatar is a callback function. Not really a react style provider.
I don't think I understand what you are saying

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -466,6 +474,10 @@ export type MessageThreadProps = {
    * @param messageStatusIndicatorProps - props of type MessageStatusIndicatorProps
    */
   onRenderMessageStatus?: (messageStatusIndicatorProps: MessageStatusIndicatorProps) => JSX.Element | null;
+  /**
+   * A callback function that can be used to provide custom data to an Avatar.
+   */
+  avatarPersonaDataProvider?: AvatarPersonaDataProvider;
```
PorterNan, line 480: I 100% agree that this should be exposed in composite level(and that's a smart design! Love the plain obj we return for userId)

But I am not sure whether we should expose this in component level - in component level, we are asking skilled engineers to do their own stuff, and this should scenario could be covered by onRenderAvatar

I guess we could do something like this

1. pass PersonaDataProvider to composite
2. In composite, implement an onRenderAvatar function and call personaDataProvider internally
3. pass onRenderAvatar function down to the component

It might fit our concept of the boundary between component and composite better - 
composite is something we prebuild for developers, just provide us {[key]: value}, composite will render things for you (maybe we can even remove onRenderAvatar function from composite if this design is powerful enough),

but component - if you choose this path, you are a skilled engineer, and wanna to have more control, only override color of persona is not enough(you could do it by onRenderAvatar still though), then we provide you onRenderAvatar api to do anything you want

anjulgarg: Thanks for reviewing the PR and your comments @PorterNan 
If we use an `onRender` function, the issue will arise when Components want to change the size of Personas.
If composites provide the Persona components to the underlying components using `onRender`, any design change in Persona becomes really painful and difficult to test.

By allowing only primitive data modifications, we ensure that our design system is not broken by CONTOSO or `onRender` functions.

packages/react-components/src/components/AvatarPersona.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Image URL to use, should be a square aspect ratio and big enough to fit in the image area.
+   */
+  imageUrl?: string;
+  /**
+   * The user's initials to display in the image area when there is no image.
+   * @defaultvalue Derived from `text`
+   */
+  imageInitials?: string;
+  /**
```
prprabhu-ms, line 24: The last two fields here seem to mix styling with content.

In component props, we should uniformly support the `styles` field for that instead.

This isn't a component prop though, which muddies the water a little bit.

If the intent is to provide a way to set foreground and background color *only* when displaying initials, one way would be to introduce a `PersonaInitials` component, and optionally expose `PersonaInitialsProps` here (which would inlude content, `text`, as well as the usual `styles`.)

packages/react-components/src/components/AvatarPersona.tsx
```diff
@@ -0,0 +1,80 @@
...
+  initialsTextColor?: string;
+};
+
+/**
+ * Callback function used to provide custom data to the AvatarPersona component.
+ */
+export type AvatarPersonaDataProvider = ((userId: string) => Promise<AvatarPersonaData>) | undefined;
+
+export interface AvatarPersonaProps extends IPersonaProps {
+  /**
```
prprabhu-ms, line 41: What is the benefit of accepting two props `userId` and `dataProvider` as opposed to a single callback that gives us the data we need?

    export interface AvatarPersonaProps extends IPersonaProps {
      getPersonaData: () => Promise<AvatarPersonaData>
    }

prprabhu-ms: I also don't quite understand why this is a callback at the component level? Do we expect to use the same `AvatarPersona` component to display personas for different users? Do we expect any of the other persona data to change?

If the intent is simply to support asynchronous loading of persona data, that can be done by the caller by updating the props being passed into the composite.

I understand that we absolutely need a `(userId: string) => PersonaData` callback at higher levels -- e.g. `MessageThread`, `VideoGallery` and the composites, because there are multiple users and we don't a-priori know what users there are.

packages/react-components/src/components/AvatarPersona.tsx
prprabhu-ms: I'd recommend renaming this type to something other than `*Provider` as that suffix has a very specific meaning in React.

packages/react-components/src/components/MessageThread.tsx
prprabhu-ms: I like to think of the components independent of the composites, first.

So, 
- `AvatarPersona` should simply take static props.
- Complex components, like `ParticipantPane` that need to show the avatar for a dynamic number of users, should provide a way to inject `AvatarPersona`, i.e. `(userId: string) => Promise<AvatarPersona>`.
- Composites, should collect together just the data that end-users might want to give us, i.e., `(userId: string) => Promise<AvatarPersonaData>`.

Also, related to one of my comments above, `AvatarPersonaProps` should follow the practices of other props -- arbitrary styling via `styles` prop etc. `AvatarPersonaData` on the other hand bundles together just the things a user of the composite wants (e.g. `initialsTextColor`).

packages/react-components/src/components/AvatarPersona.tsx
anjulgarg: Any suggestions?

packages/react-components/src/components/MessageThread.tsx
anjulgarg: @prprabhu-ms 
1. Any particular reason why `AvatarPersona` should only take static props?
2. This means duplication of code everywhere `AvatarPersona` is used. Specifically, the code for  handling and fetching the data through CONTOSO provided provider function. That also defeats the purpose of wrapping `Persona` component inside a custom `AvatarPersona` components whose sole purpose is to reduce duplication of common code that handles fetching of contoso data.
3. That's partially how it is happening in this design. If composite were to fetch the data, it would prematurely fetch the data even before an avatar is rendered. Plus you would have to track the state of userId in composite before you could call this function. More complexity at the composite layer to maintain. When it could have simply be offloaded to a child component `AvatarPersona` which has a single responsiblity.

Using `AvatarPersona` in the current way also means that we are implementing SOLID principles to a large extent.

packages/react-components/src/components/AvatarPersona.tsx
anjulgarg: The current props exposed here have been borrowed from the actual `Persona` components. 
This allows us to achieve uniformity and makes it easier for people already familiar with Fluent UI to understand this component.

Exposing a `styles` object would defeat the purpose as it would allow CONTOSO to mess up with the styling of the component which is exactly what we are trying to avoid.

packages/react-components/src/components/AvatarPersona.tsx
anjulgarg: How would the Component pass a `userId` to the callback function `dataProvider` if it doesn't get the `userID` as a prop from composite?

anjulgarg: This callback is no different than an `onRender` function we have exposed in numerous components.
Instead of returning a custom components, this callback just returns custom data...

JamesBurnside: My liking behind removing the callback is that the callback is async and makes the component no longer pure and dumb - there is an initial render, then a re-render when the async fetch is complete - meaning UI doesn't map 1:1 with props.

You could make it synchronous but then there's no value in it being a callback and may as well just be the props directly.

(disclaimer: could be wrong about this) My thinking on why `onRender` methods are callbacks and not direct props is that a react functional component is a callback so keeping that function signature presumably allows developers to do `onRenderAvatar={<MyAvatar />}`

anjulgarg:
Things we have consensus on:
- Don't expose `AvatarPerson` as a component
- Don't expose `onRender` function through composites.
- Always use `onRender` at component layer.
- Add optional props to `onRender` functions `(userId: string, { size, color, etc })` to allow style reusability.

TODO:
- Move `AvatarPersona` inside composite.
- Pass it down to dependent components like `MessageThread` using `onRender` props.
- Create a task to add prescrioptive sizes to onRender callback functions.

packages/react-components/src/components/AvatarPersona.tsx
```diff
@@ -0,0 +1,80 @@
...
+    (async () => {
+      if (dataProvider && userId) {
+        const data = await dataProvider(userId);
+        setData(data);
+      }
+    })();
+  }, [dataProvider, userId]);
+
+  return (
+    <Persona
```
alcail, line 70: By default the size will be 48px. Is this what we want? A fixed size?

image

anjulgarg: I don't think so. We would be explicitly providing size prop

mgamis-msft:
Great job on this @anjulgarg. My takeaways from this PR are:
- `AvatarPersonaDataCallback` is exposed on the composite which we use internally for `AvatarPersona` internally in MediaGallery and the ChatScreen. This way Contoso does not need to worry so much about creating the avatar with html, css, sizing, theming, etc. but just worry about giving us a callback to give us the necessary details in `AvatarPersonaData` (particularly imageUrl) to do it for them. But the downside is if Contoso wants more control they would not be able to (not to my knowledge anyway)
- Made the type of onRender props of our components all type `OnRenderAvatarType` for consistency which is good customers using our components to understand
This is a great first step so I will approve

mgamis-msft (approved):

alcail (approved):

packages/communication-react/review/communication-react.api.md
```diff
@@ -1199,6 +1216,18 @@ export const newClearErrorsModifier: (targets: ChatErrorTargets[]) => ChatStateM
 // @public (undocumented)
 export type OmitNever<T> = Pick<T, AllKeys<T>[keyof AllKeys<T>]>;
 
+// @public
+export type OnRenderAvatarType = (userId?: string, options?: {
```
prprabhu-ms, line 1220: Name this `OnRenderAvatarCallback`, for consistency?

packages/communication-react/review/communication-react.api.md
```diff
@@ -1199,6 +1216,18 @@ export const newClearErrorsModifier: (targets: ChatErrorTargets[]) => ChatStateM
...
+    hidePersonaDetails?: boolean;
+    initialsTextColor?: string;
+    presence?: PersonaPresence;
+    size?: PersonaSize;
+    coinSize?: number;
+    text?: string;
+    styles?: IStyleFunctionOrObject<IPersonaStyleProps, IPersonaStyles>;
+},
+defaultOnRender?: (props: PlaceholderProps) => JSX.Element) => JSX.Element;
```
prprabhu-ms, line 1229: I'm probably being naive here: Would it be possible for the `userId` and `options` here to simply be replaced with `PlaceholderProps`? Those are all the arguments that our internal implementation to render the avatar needs, so we give the exact same arguments to Contoso. In the simplest case, they can simply pass those through to `defaultOnRender`. In more complex cases, every business logic that we can implement in our default implementation potentially requires an override in Contoso's implementation, and they get the exact same arguments to act on.

This is a more general question: For each render callback where we have a default implementation, should the arugment to the render callback be identical to the arguments for the default implementation?

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -86,15 +88,17 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
...
-      initialsTextColor="white"
-    />
+  const avatarOptions = {
+    text: displayName,
+    size: PersonaSize.size32,
+    presence: presence,
+    initialsTextColor: 'white'
+  };
+
+  const Avatar = onRenderAvatar ? (
```
prprabhu-ms, line 98: use of Capitalized name here is a bit weird. This is a component instance, not a Component definition.

packages/react-components/src/types/OnRender.ts
```diff
@@ -0,0 +1,40 @@
...
+    presence?: PersonaPresence;
+    /** Preset Persona Size number  */
+    size?: PersonaSize;
+    /** Persona coin size in pixels  */
+    coinSize?: number;
+    /** Display name to be used in Persona  */
+    text?: string;
+    styles?: IStyleFunctionOrObject<IPersonaStyleProps, IPersonaStyles>;
+  },
+  /**
```
prprabhu-ms, line 35: Ah, options is directly passable!

Going in the opposite direction than my earlier comment: Perhaps we should simply extract the options type to a named type and use that instead of `PlaceholderProps` below?

packages/react-composites/src/composites/CallComposite/MediaGallery.tsx
```diff
@@ -55,10 +58,14 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
         remoteVideoViewOption={remoteVideoViewOption}
         styles={VideoGalleryStyles}
         layout="floatingLocalVideo"
-        onRenderAvatar={props.onRenderAvatar}
+        onRenderAvatar={(userId, options) => (
```
prprabhu-ms, line 61: Soft recommendation: Consider memoization.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -76,7 +66,9 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
           {props.showErrorBar ? <ErrorBar {...errorBarProps} /> : <></>}
           <MessageThread
             {...messageThreadProps}
-            onRenderAvatar={onRenderMessageAvatar}
+            onRenderAvatar={(userId, options) => (
```
prprabhu-ms, line 69: Memoization here might be quite important: The message thread props are extensive, so MessageThread might be rerendered often. Also, there are probably many instances of each avatar (for each message), so it's worth avoiding rerendering them.

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -0,0 +1,79 @@
...
+    (async () => {
+      if (dataProvider && userId) {
+        const data = await dataProvider(userId);
+        setData(data);
+      }
+    })();
+  }, [dataProvider, userId]);
+
+  return (
+    <Persona
```
prprabhu-ms, line 70: Does this resize properly on video tile? The `coinSize` prop isn't being respected here, if I read this correctly.

prprabhu-ms (approved):
This is great to see! Thanks for iterating on the approach.

packages/communication-react/review/communication-react.api.md
anjulgarg: PlaceholderProps is only being used in VideoTile currently. 
I added it as a 3rd argument to not have to re-write a whole bunch of code.
userId, and options play nicely when you are not trying to customize a VideoTile component.

anjulgarg: And no to your more general question

packages/react-components/src/types/OnRender.ts
anjulgarg: Ah yes, that was the idea originally. I think I forgot to do it

anjulgarg: Oh right. The signature of options and placeholder props are different. They can't be directly passed

packages/react-composites/src/composites/CallComposite/MediaGallery.tsx
anjulgarg: Not sure how I can do memoization here.
What do you suggest?

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
anjulgarg: Something like this?
image

PR #678, opened 2021-08-13
Add initial meeting adapter interface. Restructure CallAdapter and ChatAdapter interfaces.

JamesBurnside:
# What
Initial meeting adapter interface - this pulls together the pieces of chat and calling adapter it needs.

This led to chat and calling adapters having there interfaces split so ensure meeting could pick the correct parts and ensure consistency across all composites

# Why
Part of expected evolution of meeting composite - Meeting composite currently takes separate chat and calling adapters which doesn't make much sense.


packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
```diff
@@ -0,0 +1,72 @@
...
+  on(event: 'participantsLeft', listener: ParticipantLeftListener): void;
+  on(event: 'meetingEnded', listener: ParticipantLeftListener): void;
+  on(event: 'error', listener: (e: Error) => void): void;
+
+  off(event: 'participantsJoined', listener: ParticipantJoinedListener): void;
+  off(event: 'participantsLeft', listener: ParticipantLeftListener): void;
+  off(event: 'meetingEnded', listener: ParticipantLeftListener): void;
+  off(event: 'error', listener: (e: Error) => void): void;
+
+  // Call interfaces needing re-declared
```
JamesBurnside, line 37: Instead of redefining here, is there a clever way to do `Omit<CallAdapter, on(event: 'EVENT_TO_OMIT')>` instead of doing `Omit<CallAdapter, on(event: 'on')>` which I am doing currently

JamesBurnside:
Here is the corresponding state that relates strongly to this and would follow the same decisions made: #679

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
```diff
@@ -0,0 +1,72 @@
...
+  IsMuteChangedListener,
+  IsScreenSharingOnChangedListener,
+  IsSpeakingChangedListener,
+  ParticipantJoinedListener,
+  ParticipantLeftListener
+} from '../../CallComposite';
+import { ChatAdapter, MessageReadListener, MessageReceivedListener, MessageSentListener } from '../../ChatComposite';
+import { MeetingState } from '../state/MeetingState';
+
+export type ConflictingProps = 'getState' | 'onStateChange' | 'offStateChange' | 'on' | 'off';
```
prprabhu-ms, line 19: Would like to hear more opinions here.

I would argue for an alternative, more verbose, approach: Define two interfaces `ThinCallAdapter` and `ThinChatAdapter` that explicitly redefines all the props we want to pull in, but uses `Pick` to make sure we can't possibly add props that weren't there in the first place. Then, have `MeetingAdapter extends ThinCallAdapter, ThinChatAdapter {...}`.

Reasoning:
* It is easier to understand for the user of `MeetingAdapter`
  * They don't have to go looking for definitions of `CallAdapter` and `ChatAdapter` elsewhere. Everything is listed right here.
  * They don't have to deal with negation logic to figure out what properties exist, e.g. `CallAdapter` has a prop -> It was dropped in the `Omit` -> It was re-added in the `MeetingAdapter`. Switching between additive / subtractive logic is difficult.
* It avoids accidentally expanding the `MeetingComposite` API as the `CallAdapter` and `ChatAdapter` APIs evolve. Conflicting names will get caught by typescript, but that's only one kind of conflict. There might be differently named properties (in the future) that are semantically in conflict. By whitelisting, we must intentionally expand the `MeetingComposite` API each time we want to.

It is more verbose, but we're only redefining the interface. The implementation still gets to share it all.  And the use of `Pick` ensure that we don't accidentally add properties that don't exist in the underlying composites.

wdyt?

JamesBurnside: I like that much more also 👌

alcail (approved):

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
JamesBurnside: Updated; broke down the current Chat and Calling Adapter interfaces into smaller extendable interfaces, then used `Pick` on the methods that Meeting wants to implement.
Can we do this also with the `on` methods? I couldn't find a way given how they're defined

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+export interface MeetingAdapter
+  extends AdapterState<MeetingState>,
+    AdapterDisposal,
+    AdapterPages<MeetingCompositePage>,
+    MeetingAdapterHandlers,
+    Pick<
+      CallAdapterHandlers,
+      | 'joinCall'
```
JamesBurnside, line 74: Remove `joinCall`, `leaveCall`, `startCall` from `Pick<>` and add new `joinMeeting`, `startMeeting` and `leaveMeeting`

JamesBurnside: this is done futher down in PR stack

packages/communication-react/review/communication-react.api.md
```diff
@@ -253,8 +262,30 @@ export interface CallAdapter {
...
+// @public
+export type CallAdapterClientState = {
+    userId: CommunicationUserKind;
+    displayName?: string;
+    call?: CallState;
+    devices: DeviceManagerState;
+};
+
+// @public (undocumented)
+export interface CallAdapterHandlers {
```
prprabhu-ms, line 276: Recommend against the `Handler` and `Selector` terminology in this API. The adapter API is disassociated from the stateful API and those words don't mean anything in this context. Using those words here will make it harder to keep these API layers conceptually separate.

prprabhu-ms: How about `CallAdapterCallManagement` and `CallAdapterDeviceManagement`. The interfaces are then more aligned with `@azure/communication-calling`, but Contoso gets to use all the interfaces from a single exported object. So we get the best of both worlds.

packages/react-composites/review/react-composites.api.md
```diff
@@ -290,16 +311,34 @@ export type CallIdChangedListener = (event: {
...
+export interface ChatAdapter extends ChatAdapterHandlers, AdapterState<ChatState>, AdapterDisposal, AdapterErrorHandlers, ChatAdapterSubscribers {
+}
+
+// @public
+export type ChatAdapterErrors = {
+    [operation: string]: Error;
+};
+
+// @public (undocumented)
+export interface ChatAdapterHandlers {
```
prprabhu-ms, line 323: Similar to above, this could become `ChatAdapterThreadManagement`

prprabhu-ms:
I like this PR *a lot*.

* It'd be good to get this in before the Azure review since it's a such a material change to the API shape.
* Would like more eyes from the team. Might be worth bringing up in parking lot and making sure folks have seen it.

packages/communication-react/review/communication-react.api.md
JamesBurnside: Yeah I agree handlers isn't the best term, I'll update to Call/Device/ChatManagement

JamesBurnside: Do you think the likes of startCamera and setCamera would fall under DeviceManagement or CallManagement?

JamesBurnside: Thinking: 
```
export interface CallAdapterCallManagement {
  joinCall(microphoneOn?: boolean): Call | undefined;
  leaveCall(forEveryone?: boolean): Promise<void>;
  startCamera(): Promise<void>;
  stopCamera(): Promise<void>;
  onToggleCamera(options?: VideoStreamOptions): Promise<void>;
  mute(): Promise<void>;
  unmute(): Promise<void>;
  startCall(participants: string[]): Call | undefined;
  stopScreenShare(): Promise<void>;
  createStreamView(remoteUserId?: string, options?: VideoStreamOptions): Promise<void>;
}

export interface CallAdapterDeviceManagement {
  askDevicePermission(constrain: PermissionConstraints): Promise<void>;
  queryCameras(): Promise<VideoDeviceInfo[]>;
  queryMicrophones(): Promise<AudioDeviceInfo[]>;
  querySpeakers(): Promise<AudioDeviceInfo[]>;
  setCamera(sourceId: VideoDeviceInfo, options?: VideoStreamOptions): Promise<void>;
  setMicrophone(sourceId: AudioDeviceInfo): Promise<void>;
  setSpeaker(sourceId: AudioDeviceInfo): Promise<void>;
}
```

JamesBurnside: Updated to reflect this

packages/react-composites/review/react-composites.api.md
JamesBurnside: updated

prprabhu-ms (approved):
LGTM.

I was a bit squeamish about the huge adapter interface. This is a good split that will allow to evolve the interface sanely.

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
```diff
@@ -0,0 +1,131 @@
...
+      | 'onToggleCamera'
+      | 'mute'
+      | 'unmute'
+      | 'startCall'
+      | 'startScreenShare'
+      | 'stopScreenShare'
+      | 'createStreamView'
+      | 'disposeStreamView'
+    >,
+    Pick<
```
anjulgarg, line 53: Why is Pick used here? Wouldn't this need all the props defined in `CallAdapterDeviceManagement` interface?

anjulgarg (approved):

anjulgarg:
Very nice cleanup @JamesBurnside 
Left a small comment.

packages/communication-react/review/communication-react.api.md
```diff
@@ -1178,17 +1205,88 @@ export interface LocalVideoStreamState {
     view?: VideoStreamRendererViewState;
 }
 
-// @public
+// @alpha
+export interface MeetingAdapter extends AdapterState<MeetingState>, AdapterDisposal, AdapterPages<MeetingCompositePage>, MeetingAdapterSubscriptions {
+}
+
+// @alpha
+export interface MeetingAdapterMeetingManagement extends Pick<CallAdapterCallManagement, 'joinCall' | 'leaveCall' | 'startCamera' | 'stopCamera' | 'onToggleCamera' | 'mute' | 'unmute' | 'startCall' | 'startScreenShare' | 'stopScreenShare' | 'createStreamView' | 'disposeStreamView'>, Pick<CallAdapterDeviceManagement, 'setCamera' | 'setMicrophone' | 'setSpeaker' | 'askDevicePermission' | 'queryCameras' | 'queryMicrophones' | 'querySpeakers'>, Pick<ChatAdapterThreadManagement, 'fetchInitialData' | 'sendMessage' | 'sendReadReceipt' | 'sendTypingIndicator' | 'loadPreviousChatMessages'> {
```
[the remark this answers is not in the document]
prprabhu-ms: See discussion here: https://github.com/Azure/communication-ui-library/pull/678#discussion_r688700680

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
```diff
@@ -0,0 +1,131 @@
...
+      | 'askDevicePermission'
+      | 'queryCameras'
+      | 'queryMicrophones'
+      | 'querySpeakers'
+    >,
+    Pick<
+      ChatAdapterThreadManagement,
+      'fetchInitialData' | 'sendMessage' | 'sendReadReceipt' | 'sendTypingIndicator' | 'loadPreviousChatMessages'
+    > {
+  removeParticipant(userId: string): Promise<void>;
```
PorterNan, line 67: It looks like a common part of each adapter, but not sure which interface this should be staying(maybe just keep it here)

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
```diff
@@ -0,0 +1,131 @@
...
+  on(event: 'participantsLeft', listener: ParticipantLeftListener): void;
+  on(event: 'meetingEnded', listener: ParticipantLeftListener): void;
+  on(event: 'error', listener: (e: Error) => void): void;
+
+  off(event: 'participantsJoined', listener: ParticipantJoinedListener): void;
+  off(event: 'participantsLeft', listener: ParticipantLeftListener): void;
+  off(event: 'meetingEnded', listener: ParticipantLeftListener): void;
+  off(event: 'error', listener: (e: Error) => void): void;
+
+  // Call subscriptions
```
PorterNan, line 86: Is it possible to reuse interface from callSubscriber and chatSubscriber?

PorterNan (approved):
It looks great! Abstract all the common part of each adapter - so we won't be afraid of adding more adapters in the future(hopefully we will only need 3 adapters in the end)

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
JamesBurnside: I would love to be able to but can't see a way from how we do the `on` interface. One option is to change `on('string', ...)` to be `onString(...)`

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -160,6 +141,18 @@ export interface CallAdapter {
   off(event: 'error', listener: (e: Error) => void): void;
 }
 
+/**
+ * Call Composite Adapter interface.
+ */
+export interface CallAdapter
+  extends AdapterState<CallAdapterState>,
```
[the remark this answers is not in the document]
JamesBurnside: In theory it could be a `type` instead of interface to do this but I'd prefer to keep as interface

PR #680, opened 2021-08-13
Tee errors to CallClientState

prprabhu-ms:
# What
Call counterpart of #510 and friends.
Tees errors for a variety of API calls in Calling to the `CallClientState`.

# Why
Will allow us to hook these errors up into the `ErrorBar` and also expose them via the `CallAdapter`.

# How Tested
* `rush test`
* Sanity check call composite via storybook.

# Process & policy checklist


packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -430,4 +430,82 @@ export interface CallClientState {
...
+  /**
+   * Stores the latest error for each API method.
+   *
+   * See documentation of {@Link CallErrors} for details.
+   */
+  latestErrors: CallErrors;
 }
+
+/**
+ * Errors teed from API calls to the Calling SDK.
```
alcail, line 442: you mean tied?

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -430,4 +430,82 @@ export interface CallClientState {
...
+ *
+ * Each property in the object stores the latest error for a particular SDK API method.
+ *
+ * Errors from this object can be cleared using the TODO(implement me) {@Link newClearErrorsModifier}.
+ * Additionally, errors are automatically cleared when:
+ * - The state is cleared.
+ * - Subsequent calls to related API methods succeed.
+ * See documentation of individual stateful client methods for details on when errors may be automatically cleared.
+ */
+export type CallErrors = {
```
alcail, line 452: nit: name might be confusing with CallErrors being an array of CallError

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -430,4 +430,82 @@ export interface CallClientState {
...
+    this.inner = inner;
+    this.name = 'CallError';
+    this.message = `${this.target}: ${this.inner.message}`;
+  }
+}
+
+/**
+ * String literal type for all permissible keys in {@Link CallErrors}.
+ */
+export type CallErrorTargets =
```
alcail, line 481: nit: why plural when there's only one?

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -640,4 +644,95 @@ export class CallContext {
     this._atomicId++;
     return id;
   }
+
+  /**
+   * Tees any errors encountered in an async function to the state.
```
alcail, line 649: nit: ties?

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -640,4 +644,95 @@ export class CallContext {
...
+   *
+   * If the function succeeds, clears associated errors from the state.
+   *
+   * @param f Async function to execute.
+   * @param target The error target to tee error to.
+   * @param clearTargets The error targets to clear errors for if the function succeeds. By default, clears errors for `target`.
+   * @returns Result of calling `f`. Also re-raises any exceptions thrown from `f`.
+   * @throws CallError. Exceptions thrown from `f` are tagged with the failed `target.
+   */
+  public withAsyncErrorTeedToState<Args extends unknown[], R>(
```
alcail, line 659: nit: withAsyncErrorTiedToState?

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -640,4 +644,95 @@ export class CallContext {
...
+    f: (...args: Args) => Promise<R>,
```
alcail, line 660: nit: personally don't like the single letter name for a parameter as it's not explicit at all
asyncAction or something around it might be better

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -640,4 +644,95 @@ export class CallContext {
...
+    target: CallErrorTargets,
+    clearTargets?: CallErrorTargets[]
+  ): (...args: Args) => Promise<R> {
+    return async (...args: Args): Promise<R> => {
+      try {
+        const ret = await f(...args);
```
alcail, line 666: nit: response might be a better name

packages/react-composites/src/composites/CallComposite/hooks/useAdaptedSelector.ts
```diff
@@ -84,7 +84,9 @@ const memoizeState = memoizeOne(
     callsEnded: [],
     deviceManager,
     callAgent: { displayName },
-    calls
+    calls,
+    // TODO: FIXME
```
alcail, line 88: nit: why this comment?

JamesBurnside (approved):

packages/calling-stateful-client/src/CallClientState.ts
prprabhu-ms: This is an object that contains many errors, so the name seems fine.

We shouldn't encode type information in names (so no CallErrorMap or such). Will stick to this one.

packages/calling-stateful-client/src/CallClientState.ts
prprabhu-ms: Good point. Need to go change `ChatErrorTargets` as well, in a separate PR.

packages/calling-stateful-client/src/CallContext.ts
prprabhu-ms: Done here, but: I actually really like single letter names for local variables, and sometimes function parameters -- it means that your functions are so short, and named such that you don't need any context to understand what that variable / parameter means.

In this example `withAsyncErrorTeedToState`, takes exactly one argument, the function to wrap. The type of the function to wrap is important (tells you that it's a function), but the name doesn't help all that much.
Also, the implementation is short enough that `f` suffices (in my opinion).

Also, named it `action`, not `asyncAction`, the `async` part is in the type.

packages/react-composites/src/composites/CallComposite/hooks/useAdaptedSelector.ts
prprabhu-ms: Removed in a PR higher in the stack. I added more context to the comment so it's not as enigmatic :D

alcail (approved):

PR #684, opened 2021-08-14
Do not squelch exception in batch mode

prprabhu-ms:
# What
* Fix a bug introduced in batch mode in #601
* Add unittests that would have caught this bug.

# Why
(Latent) bug fix.

# How Tested
`rush test`

# Process & policy checklist


packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -438,6 +438,7 @@ export class ChatContext {
       }
     } catch (e) {
       this._state = priorState;
+      throw e;
     } finally {
       this._batchMode = false;
```
JamesBurnside, line 443: Question on catch-finally behavior: will this finally block be called if the exception is rethrown?

JamesBurnside (approved):

alcail (approved):

packages/chat-stateful-client/src/ChatContext.ts
prprabhu-ms: Yes, finally always runs.

PR #694, opened 2021-08-18
Generate snapshot in github action for composite

PorterNan:
# What

Generate snapshot in github action for composite, fix the problem of unstable retries

# Why

# How Tested

# Process & policy checklist


JamesBurnside:
Are these camera enabled screenshots correct? (Should they have the `y4m` pic?)

packages/react-composites/tests/browser/utils.ts
```diff
@@ -20,22 +20,16 @@ export const TOPIC_NAME = 'Cowabunga';
...
-  // The participant list can be empty when there are no participants to load.
-  // For example, this happens when incorrect credentials are used in the ErroBar tests.
-  //
-  // Only waiting for the element to be attached to the DOM is enough, especially given
-  // that we wait for 1 second after this step. Reconsider this when we remove that 1 second
-  // wait.
-  await page.waitForSelector(dataUiId(IDS.participantList));
+
   // @TODO
   // We wait 1 sec here to work around a bug.
```
JamesBurnside, line 27: update to 3 seconds

JamesBurnside: (and L37)

JamesBurnside (approved):
LGTM! Only issue is that it looks like the fake video isn't working anymore but otherwise this is awesome!

PorterNan:
> > Are these camera enabled screenshots correct? (Should they have the `y4m` pic? The camera name looks like it isn't constructing the camera correctly?)
> > image
> 
> It is by designed, a pure color image will help us generate stable result - otherwise the camera stream will be different every time because of the compression server side does

Haha, that was me, I just noticed I set auto signed in for this account in my laptop edge

anjulgarg (approved):

PR #695, opened 2021-08-18
Create stub `AzureCommunicationMeetingAdapter`

JamesBurnside:
# What
Create the stub for the `AzureCommunicationMeetingAdapter`

# Why
This is a seperate PR so the changes in subsequent PRs that implement actual functionally can be easily discerned


packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
```diff
@@ -93,7 +90,11 @@ export interface MeetingAdapter
       ChatAdapterHandlers,
       'fetchInitialData' | 'sendMessage' | 'sendReadReceipt' | 'sendTypingIndicator' | 'loadPreviousChatMessages'
     >,
-    MeetingAdapterSubscriptions {}
+    MeetingAdapterSubscriptions {
+  joinMeeting(microphoneOn?: boolean): void;
+  leaveMeeting(forEveryone?: boolean): Promise<void>;
```
JamesBurnside, line 95: remove `forEveryone`, this can work for calls but you can't make everyone leave a chat thread

alcail (approved):

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
prprabhu-ms: `leaveMeeting(forEveryone)` anyway sounds weird to me. You can remove others, you can't really leave for them :)

If we want the functionality to kick everyone, we can add a separate method removeAllParticipants(), although I'm not sure why we'd want that functionality at all.

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
```diff
@@ -93,7 +90,11 @@ export interface MeetingAdapter
...
+  startMeeting(participants: string[]): void;
```
prprabhu-ms, line 96: Please document these public methods.

In particular, document that `participants` is an array of ACS user IDs (right?) Are these flattened IDs? Every time we use `string` for user ID in the API, we must explicitly state how to get this string from an actual ID.

packages/react-composites/src/composites/MeetingComposite/adapter/MeetingAdapter.ts
JamesBurnside: It was purely reflecting the call sdk that has leaveCall(forEveryone) - but yeah removing the forEveryone and just having `leaveMeeting`

prprabhu-ms (approved):

PR #696, opened 2021-08-18
Meeting adapter implementation part I - forward 1-1 methods to call and chat adapters

JamesBurnside:
# What
Forward the basic function calls straight to the appropriate adapter

Doing state and `on` methods will come next but are more involved

# Why
Part of the overall adapter implementation


packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -85,52 +86,70 @@ export class AzureCommunicationMeetingAdapter implements MeetingAdapter {
...
   }
   public dispose(): void {
-    throw new Error('Method not implemented.');
+    this.unsubscribeMeetingEvents();
+    this.callAdapter.dispose();
+    this.chatAdapter.dispose();
   }
   public setPage(page: MeetingCompositePage): void {
-    throw new Error('Method not implemented.');
+    switch (page) {
```
JamesBurnside, line 94: nit: isntead of switch statement this can just be:
```
export function meetingPageToCallPage(page: MeetingCompositePage): CallCompositePage {
  return page === 'meeting' ? 'call' : page;
}
```

alcail (approved):

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -67,13 +67,14 @@ export class AzureCommunicationMeetingAdapter implements MeetingAdapter {
   }
 
   public joinMeeting(microphoneOn?: boolean): void {
-    throw new Error('Method not implemented.');
+    this.callAdapter.joinCall(microphoneOn);
   }
-  public async leaveMeeting(forEveryone?: boolean): Promise<void> {
-    throw new Error('Method not implemented.');
+  public async leaveMeeting(): Promise<void> {
+    await this.chatAdapter.removeParticipant(this.chatAdapter.getState().userId);
```
prprabhu-ms, line 73: Both chat and calling state save userId. Should we hoist a single `userId` to the meeting adapter state directly instead of the duplicate values?

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -67,13 +67,14 @@ export class AzureCommunicationMeetingAdapter implements MeetingAdapter {
...
+    await this.callAdapter.leaveCall();
   }
   public startMeeting(participants: string[]): void {
-    throw new Error('Method not implemented.');
+    this.callAdapter.startCall(participants);
```
prprabhu-ms, line 77: I'm a bit confused by this API. I understand that using the `MeetingComposite`, we can join an existing Teams meeting.

But what does starting a new meeting mean? How would we get an associated chat thread?

prprabhu-ms (approved):

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
JamesBurnside: updated

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
JamesBurnside: We do have one at the MeetingAdapterClientState level (where displayName etc is stored). However here we use the chat one is simply to get around the fact that chat doesn't have an easy leaveChat() fn and instead requires you to remove yourself so was the simplest place to grab that user id.

If you wanted the meetingAdapter to also have this ID I could update the ctor to be:
```
constructor(userId: string, callAdapter: CallAdapter, chatAdapter: ChatAdapter) {
  this.userId = userId;
}
```

But pushing back a bit just to make the code simpler - happy to update though if you feel strongly

PR #699, opened 2021-08-18
ParticipantItem truncation of long names

mgamis-msft:
# What

Refactored ParticipantItem to have two sections
                   Section 1                                                           Section 2
-------------------------------------------------------------------------------------------------
|        (Persona and '(you)' if you)           |      (menu button if hovered otherwise icons)     |
-------------------------------------------------------------------------------------------------
<- calc(100% - iconContainerWidth)--><-------------- iconContainerWidth ------------->
iconContainerWidth = 4rem unless specified differently in styles prop (line 113 in [ParticipantItem diff](https://github.com/Azure/communication-ui-library/compare/mgamis/participant-item-fix?expand=1#diff-7cb1c54833e2004edac5e0854a646aba8bfa5d94b983f3978816922dc0957d46R113))
Assigning a width to Section 1 above will cause Persona to truncate when the width exceeded 

Also added styling to ensure multiple icons are aligned to address this bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2475577

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2540605
https://skype.visualstudio.com/SPOOL/_workitems/edit/2475577

# How Tested

Tested in calling sample:
image
Tested in storybook canvas:
https://user-images.githubusercontent.com/79475487/129973982-eb4ce2bb-a1fc-44da-8af4-575cf11d001f.mp4

Need someone to verify that this bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2475577 is fixed on Mac because I don't have one.

# Process & policy checklist


alcail (approved):

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -97,14 +102,38 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
...
 
-  const meTextStyle = mergeStyles({ color: theme.palette.neutralTertiary }, styles?.me);
-  const contextualMenuStyle = mergeStyles({ background: theme.palette.neutralLighterAlt }, styles?.menu);
-  const menuButtonContainerStyle = mergeStyles({
-    root: {
-      '&:hover': { background: 'none' },
-      '&:active': { background: 'none' }
-    }
-  });
+  const meTextStyle = useMemo(
```
prprabhu-ms, line 105: Here and below: You don't need to memoize the result of mergeStyles as it's a string (CSS class name).

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -97,14 +102,38 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
...
+    () => mergeStyles({ color: theme.palette.neutralTertiary }, styles?.me),
+    [theme.palette.neutralTertiary, styles?.me]
+  );
+  const contextualMenuStyle = useMemo(
+    () => mergeStyles({ background: theme.palette.neutralLighterAlt }, styles?.menu),
+    [theme.palette.neutralLighterAlt, styles?.menu]
+  );
+  const iconsContainerWidth = useMemo(
+    () => (styles?.iconContainer as CSSProperties)?.width ?? '3.5rem',
```
prprabhu-ms, line 114: No need to memoize string (Only object references should be memoized to avoid rerenders)

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -81,12 +86,12 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
   const [menuHidden, setMenuHidden] = useState<boolean>(true);
   const containerRef = useRef<HTMLDivElement>(null);
   const theme = useTheme();
-  const { strings } = useLocale();
+  const localeStrings = useLocale().strings.participantItem;
```
prprabhu-ms, line 89: `ParticipantItem` is becoming too complex. Please split it up into multiple (internal) Components.

OK to do as a follow up refactor PR.

prprabhu-ms (approved):

packages/react-components/src/components/ParticipantItem.tsx
mgamis-msft: Sorry I may have missed something but I thought useMemo is for variables that are computation heavy that only depend on a few props. I'm trying to avoid using mergeStyles if theme and styles have not changed. For example, when a participant unmutes, I dont want to call mergeStyles again.

prprabhu-ms: Oh, OK.

`useMemo` is used more to avoid creating a new result object, to avoid rerendering a Component that uses that object as a prop. It can be used to simply avoid computation as you do here as well, but usually computations like this aren't the costly step.

PR #700, opened 2021-08-18
Use Meeting adapter in the meeting composite

JamesBurnside:
# What
Update meeting composite to be backed by the `AzureCommunicationMeetingAdapter`

This involved creating two façade classes: `MeetingBackedCallAdapter` and `MeetingBackedChatAdapter` that could use the meeting adapter but still satisfy the interfaces needed by the CallComposite and ChatComposite and perform a directional state conversion between MeetingState -> Call/ChatState as well as Call/ChatState -> MeetingState.

# How Tested
Ran meeting composite in storybook locally using the new azure communication adapter, was able to:
* Toggle video and microphone in pre join screen
* Join the meeting with video and microphone enabled and disabled
* Shown lobby screen when joining a teams meeting
  * **Toggling microphone not working here! (existing issue)**
* Can join ACS meeting or teams meeting
* Can toggle video and microphone in meeting
* Can screenshare in meeting
* Can send and receive chat messages in meeting
* Can switch microphone and video source in meeting and pre join screens


alcail (approved):

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -32,16 +33,30 @@ export type MeetingCompositeProps = {
...
+  const { meetingAdapter, fluentTheme } = props;
+
+  if (!meetingAdapter) {
+    throw 'Meeting adapter is undefined';
+  }
+
+  const [callAdapter, setCallAdapter] = useState<CallAdapter>(new MeetingBackedCallAdapter(meetingAdapter));
+  const [chatAdapter, setChatAdapter] = useState<ChatAdapter>(new MeetingBackedChatAdapter(meetingAdapter));
+
+  useEffect(() => {
```
prprabhu-ms, line 45: meetingAdapter is a `prop` that is never changed internally. So, you should be able to simply `useRef` for `callAdapter` and `chatAdapter` and directly create those two (without a `useEffect` block).

`useEffect` is intended to respond to some external change that isn't a directly result of a props change.

prprabhu-ms: Ah, I guess the idea is to not recreate the adapters if props other than meetingAdapter change?

Even in that case, I'd just store them all as ref (because it is never the case that any of them change without a prop change):


    const { meetingAdapter: newMeetingAdapter, fluentTheme } = props;
    const meetingAdapter = useRef(newMeetingAdapter);
    const chatAdapter = useRef(new MeetingBackedCallAdapter(meetingAdapter));
    ...

    // On prop update only.
    if (newMeetingAdapter !== meetingAdapter.current) {
        meetingAdapter.current = newMeetingAdatper;
        chatAdatper.current = new MeetingBackedChatAdapter(newMeetingAdapter);
        ...
    }

JamesBurnside:
> @JamesBurnside James Burnside FTE Let's check this in when you are back to the team before the Azure Review, this is one of the most important piece for that
> 
> Or at least change the api file we are uploading

I wouldn't worry too much for this API review as we won't be releasing meeting composite as part of initial GA (but I would like the feedback on it so I will try and get it in)

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
JamesBurnside: Updated to use `useMemo` instead to make it much simpler:
```
const callAdapter: CallAdapter = useMemo(() => new MeetingBackedCallAdapter(meetingAdapter), [meetingAdapter]);
const chatAdapter: ChatAdapter = useMemo(() => new MeetingBackedChatAdapter(meetingAdapter), [meetingAdapter]);
```

PR #702, opened 2021-08-19
Use `ErrorBar` to show some errors in `CallComposite`

prprabhu-ms:
# What
* Hook up `ErrorBar` with `CallClientState.latestErrors` to show errors when muting/unmuting, starting/stopping video or starting/stopping screenshare.
* FIXME: The errors aren't actually bubbled up yet, because the calling SDK drops these errors on the ground. Working on that separately.

# Why
Hook up error reporting end-to-end for Calling with most obvious examples.

# How Tested
* `rush build`
* Sanity checked in storybook, but can't see actual errors yet.


alcail (approved):

packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { getLatestErrors } from './baseSelectors';
+import { ErrorType } from '@internal/react-components';
+import { createSelector } from 'reselect';
+
+/**
+ * Select the first fiew active errors from the state for the `ErrorBar` component.
```
anjulgarg, line 9: few? five?

packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * In case there are many errors, only a few top errors are returned to avoid
```
anjulgarg, line 11: Can we add absolute numbers here instead of many/few?

packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -0,0 +1,54 @@
...
+ * filling up the UI with too many errors.
+ *
+ * Invariants:
+ *   - `ErrorType` is never repeated in the returned errors.
+ *   - Errors are returned in a fixed order by `ErrorType`.
```
anjulgarg, line 16: Ordered by timestamp or name?

anjulgarg (approved):

packages/calling-component-bindings/src/errorBarSelector.ts
prprabhu-ms: Done

packages/calling-component-bindings/src/errorBarSelector.ts
prprabhu-ms: Done

PR #704, opened 2021-08-19
Enable ErrorBar in ChatComposite

prprabhu-ms:
# What
Drop experimental option to enable ErrorBar. Always enable it.

# Why
Both ChatComposite and CallComposite now (soon?) show errors via ErrorBar, so enable this for next beta release.
We can add back an option to disable it if we find a need. For now, it seems better to just enable it unconditionally.

# How Tested
`rush build`


alcail (approved):

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -52,15 +52,6 @@ export type ChatCompositeProps = {
  * Additional customizations for the chat composite
  */
 export type ChatOptions = {
-  /**
```
PorterNan, line 55: Love the jsdoc comment here!

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -69,7 +68,7 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
       {!!showTopic && <ChatHeader {...headerProps} />}
       <Stack className={chatArea} tokens={participantListContainerPadding} horizontal grow>
         <Stack className={chatWrapper} grow>
-          {props.showErrorBar ? <ErrorBar {...errorBarProps} /> : <></>}
```
PorterNan, line 72: Can we make it true by default but still allow user not to show it(so keep it in options, but true by default)

PorterNan: Contoso might have their own error handling outside of our our composite, I believe there should be cases they don't wanna show it, but totally agree that we show it if no one changes the options

anjulgarg:
So we are completely disallowing Contoso from hiding the Error Bar?

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
anjulgarg: Seems a little extreme to force an Error Bar

prprabhu-ms:
Bad bad dobby, as discussed here: https://teams.microsoft.com/l/message/19:5e150d250f724d1089c949305efdc8cc@thread.skype/1629493799171?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e9c1fc3-39df-4486-a26a-456d80e80f82&parentMessageId=1629493799171&teamName=Azure%20Communication%20Services&channelName=Web%20UI%20-%20Team&createdTime=1629493799171

PR #707, opened 2021-08-20
Custom avatar e2e test fixes

anjulgarg:
# What
Updating e2e tests to run with the new custom avatar model

# Why

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -127,10 +127,11 @@ test.describe('Chat Composite custom data model', () => {
     await page.type(dataUiId(IDS.sendboxTextfield), 'How the turn tables');
     await page.keyboard.press('Enter');
     await page.waitForSelector(`[data-ui-status="delivered"]`);
-
+    // Now that we are using a promise based custom avatar data injection, we
+    // need to wait for the avatar to load before we can take a screenshot.
+    await page.waitForTimeout(1000);
```
prprabhu-ms, line 132: This adds 1 second to the test unconditionally.

Instead of this, can you inject and wait for an HTML attribute here as well? (e.g., inject a more specific test in the `text: 'CustomDataHookTextWithMagicNumber23997'`` and then wait for an HTML element with that text).

packages/react-composites/tests/browser/chat/app/index.tsx
```diff
@@ -64,7 +64,17 @@ function App(): JSX.Element {
                 )
               : undefined
           }
-          onRenderAvatar={customDataModel ? () => <text id="custom-data-model-avatar">Avatar</text> : undefined}
+          onFetchAvatarPersonaData={
+            customDataModel
+              ? (userId) =>
+                  new Promise((resolve) =>
+                    resolve({
+                      imageInitials: 'CI',
```
alcail, line 72: nitty nit: CN for Custom Name?

alcail (approved):

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Can't do that here since we are not injecting an html element, rather injecting data values to the avatar.

packages/react-composites/tests/browser/chat/app/index.tsx
anjulgarg: Intentional to show that imageInitial can be independent of the displayName

alcail: got it ;)

prprabhu-ms:
Merging this PR to unblock other PRs that can't land while the test is broken. Let's continue discussions / iteration on ID injection in a follow up.

PR #708, opened 2021-08-20
Improving E2E tests for chat after custom avatar update

anjulgarg:
# What
Improving E2E tests for chat after custom avatar update

# Why

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -127,9 +127,9 @@ test.describe('Chat Composite custom data model', () => {
     await page.type(dataUiId(IDS.sendboxTextfield), 'How the turn tables');
     await page.keyboard.press('Enter');
     await page.waitForSelector(`[data-ui-status="delivered"]`);
-    // Now that we are using a promise based custom avatar data injection, we
-    // need to wait for the avatar to load before we can take a screenshot.
-    await page.waitForTimeout(1000);
+    await page.waitForFunction(() => {
+      return document.querySelectorAll('[data-ui-id="chat-composite-participant-custom-avatar"]').length >= 2;
```
PorterNan, line 131: Let's make it ===2 just in case it generates more than 2 avatars(if this is the right number) because of some unstable behavior(to prevent unstable snapshot)?

PorterNan (approved):

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -95,7 +95,12 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
                 <ParticipantList
                   {...participantListProps}
                   onRenderAvatar={(userId, options) => (
-                    <AvatarPersona userId={userId} {...options} dataProvider={onFetchAvatarPersonaData} />
+                    <AvatarPersona
+                      data-ui-id="chat-composite-participant-custom-avatar"
```
PorterNan, line 99: BTW, should this come from some indetifierProvider or something?

prprabhu-ms (approved):

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
anjulgarg: Haha I made it >= 2 to make it future proof. Sure, I will change it.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
anjulgarg: So this is a bit tricky here.
`IdentityProvider` is a part of the react-components package and contains ids for pure components.
If we add ids used inside composite to the provider as well, it muddies its scope

anjulgarg: What can be done here is the component based identityprovider can be extended and recreated for composites.

anjulgarg: So there are multiple places in composite where data-ui-ids have been hardcoded.
We can go ahead like this for now and have a separate task for cleaning data-ui-ids in composites ;)

PorterNan: Sounds good, we might want to find a consistent way all across composite and component in the end, but sounds it is more complicated than this PR itself

PR #709, opened 2021-08-20
Jinan/participant list test fix

PorterNan:
# What

Fix unstale participant list test

# Why

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -82,9 +82,10 @@ test.describe('Call Composite E2E CallScreen Tests', () => {
       page.bringToFront();
 
       await page.click(dataUiId('call-composite-participants-button'));
-      // Clicking on participants icon displays a dropdown menu that has an animation.
-      // We wait 1 second for that animation to complete.
-      await page.waitForTimeout(1000);
+      const buttonCallOut = await page.waitForSelector('.ms-Callout');
```
prprabhu-ms, line 85: Won't this target *any* callout on the page?

Can we scope this to `ParticipantList` somehow?

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -82,9 +82,10 @@ test.describe('Call Composite E2E CallScreen Tests', () => {
...
+      // This will ensure no animation is happening for the callout
+      await buttonCallOut.waitForElementState('stable');
```
prprabhu-ms, line 87: Very cool.

prprabhu-ms (approved):

packages/react-composites/tests/browser/call/CallComposite.test.ts
anjulgarg: Indeed very cool

packages/react-composites/tests/browser/call/CallComposite.test.ts
anjulgarg: +1 to what @prprabhu-ms  commented.
This is a hack. I started out with the tests this way but realised that this is not future proof.
What you can do is add a data-ui-id to the container element and use multiple selectors to narrow down the selected component.

`const buttonCallOut = await page.waitForSelector('[dta-ui-id-"something"] .ms-Callout');`

anjulgarg (approved):

packages/react-composites/tests/browser/call/CallComposite.test.ts
PorterNan: I agree with that we should add data-ui-id whenever it is possible, but as @anjulgarg mentioned in the chat thread, it is not that feasible(and this will be only happening for participantListButton)

But good news is Callouts can only be shown one at the same time on the screen, if there is one more Callouts appears, we need to come back and add new tests/modify test anyway

alcail (approved):

packages/react-composites/tests/browser/call/CallComposite.test.ts
PorterNan: Just looked at the possibility on what @anjulgarg  said:
image
Sadly, the CallOut component is not containing by any parent, it is just a global div element floating all around, I will check this one in and we can take another look if there is a use case that shows multiple CallOut in the same screen at the same time

anjulgarg: LOL yes, this is what stopped me from doing it the right way :D

JamesBurnside: We may want to consider parenting the callouts to components we control (likely by setting a hostId prop somewhere in the button's menu props) similar to: https://github.com/Azure/communication-ui-library/pull/651/files

That would allow us to target the callout by parent class/id but also means we control the css that hosts the callout - our customers could have anything in their body css which could screw up the flyout as it did our modal dialogs (see: https://github.com/Azure/communication-ui-library/pull/651)

PR #710, opened 2021-08-20
[DRAFT] Custom Icons

anjulgarg:
# What
Shows a sample implementation of custom icon injection.
Really easy implementation provided by Fluent UI.

Official Docs don't have this information  so had to dig around a bit to find this.
Reference docs can be found here.
https://github.com/microsoft/fluentui/wiki/Using-icons#registering-custom-icons

If we go ahead with this implementation, the work involved is
- Replace direct icon usage like `<Video20icon />` with `<Icon iconName="Video20Icon" />` to make this work.
- Expose `registerIcons` function through the composites to Contoso.
- Adds docs in strorybook

image
image

# Why

# How Tested

# Process & policy checklist


alcail:
wondering if we can have something like following or something similar:

on Contoso side:
  ContosoIcons = {
    VideoOn: <text>Video On</text>
  }

Registration on our side:
  registerIcons({
    icons: {
      VideoOn:  ContosoIcons.VideoOn ?? <Video20Filled key={'videoIconKey'} />
    }
  });

and in component:
  const onRenderCameraOnIcon = (): JSX.Element => <Icon iconName="VideoOn" />;

anjulgarg:
> wondering if we can have something like following or something similar:
> 
> on Contoso side:
> ContosoIcons = {
> VideoOn: Video On
> }
> 
> Registration on our side:
> registerIcons({
> icons: {
> VideoOn: ContosoIcons.VideoOn ?? <Video20Filled key={'videoIconKey'} />
> }
> });
> 
> and in component:
> const onRenderCameraOnIcon = (): JSX.Element => ;

That's the idea @alcail

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -43,10 +44,8 @@ export interface CameraButtonProps extends ControlBarButtonProps {
   strings?: Partial<CameraButtonStrings>;
 }
 
-const onRenderCameraOnIcon = (): JSX.Element => <Video20Filled key={'videoIconKey'} primaryFill="currentColor" />;
-const onRenderCameraOffIcon = (): JSX.Element => (
-  <VideoOff20Filled key={'videoOffIconKey'} primaryFill="currentColor" />
-);
+const onRenderCameraOnIcon = (): JSX.Element => <Icon iconName="Video20Filled" />;
```
prprabhu-ms, line 47: iirc, The `primaryFill="currentColor"` attribute was added to avoid a warning from fluent about incorrect use of the `primaryFill` CSS property. You might want to keep that. (If you don't see any new warnings, you're good to go)

also, nit: Make this a componet? `DefaultCameraOnIcon` etc?

prprabhu-ms (approved):

packages/react-components/src/components/CameraButton.tsx
alcail: James has fixed that in one of his PR and this is not needed anymore: https://github.com/Azure/communication-ui-library/pull/687

anjulgarg: @prprabhu-ms Regarding renaming it, this is throwaway code used only as sample ;)

PR #711, opened 2021-08-23
Remove `useLocale` and friend from API

prprabhu-ms:
# What
* Remove `useLocale` and `LocaleContext` from API
* Split canned locales to own internal folder -- simplifies the export from the localization module.

# Why
These are part of internal implementation of the components.

# How Tested
* `rush test`
* storybook - play around with localization some.

# Process & policy checklist

 Removes unintentional exports


packages/communication-react/review/communication-react.api.md
```diff
@@ -1240,13 +1237,6 @@ export interface MicrophoneButtonStrings {
     onLabel: string;
 }
 
-// @public
-export const namedLocales: Record<string, {
```
prprabhu-ms, line 1244: This was an alternative export for the named exports `en_US` etc.
Removed because:
* Redundant
* Not described in storybook
* Actually works badly with tree-shaking -- customers would get all language bundles by pulling this symbol.

mgamis-msft (approved):

mgamis-msft:
Thanks for this clean up, @prprabhu-ms. Looks good!

packages/communication-react/review/communication-react.api.md
prprabhu-ms: Turns out this is used in the storybook to populate the list of available locales.

TODO: Drop this use in storybook, instead just list the locales there.

packages/react-components/src/locales/index.ts
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT license.
+
+import { Locale } from '../localization/';
+import _en_US from './translated/en_US.json';
+import _de_DE from './translated/de_DE.json';
+import _ar_SA from './translated/ar_SA.json';
+import _fr_FR from './translated/fr_FR.json';
+
+/** Locale for English (US) */
+export const en_US: Locale = { strings: _en_US };
```
alcail, line 11: As said in Miguel PR around this, wondering if it will be tter to have a single object gathering all those locales

prprabhu-ms: Actually, it's better to keep these constants separate. I want to land a test in the treeshaking package that if an app only imports en_US, none of the other locales / JSON data is loaded. This means that Contoso can simply import the handful of locales they want without paying the cost of a larger app bundle from the unused locales.

mgamis-msft: I think Aline means we can simply get rid of the `Locale` object since it only has one prop and simply use `ComponentStrings` as the context for LocalizationProvider. I like that idea too. But we can do that in another PR.

alcail (approved):

packages/communication-react/review/communication-react.api.md
prprabhu-ms: Done

PR #713, opened 2021-08-23
[POC] Set locales for composites, in addition to components

prprabhu-ms:
# What
Add a way to set locale for composites -- for strings used by composites that don't come from a component.

# Why
Need a way to localize these strings.

# How Tested
* TBD: storybook

# TODO [This is still a POC]

- [ ] Rename component exports `Locale`, `en_US` etc to avoid confusion with the newly added composite equivalents.
- [ ] Modify `ChatComposite` API similarly
- [ ] Fix storybook so that preview also sets composites' locale
- [ ] Update storybook documentation to describe the new `CompositeLocale` and friends.


prprabhu-ms:
Ugh. Created PR against the wrong branch

PR #714, opened 2021-08-23
[POC] Set locales for composites, in addition to components

prprabhu-ms:
# What
Add a way to set locale for composites -- for strings used by composites that don't come from a component.

# Why
Need a way to localize these strings.

# How Tested
* TBD: storybook

# TODO [This is still a POC]

- [ ] Rename component exports `Locale`, `en_US` etc to avoid confusion with the newly added composite equivalents.
- [ ] Modify `ChatComposite` API similarly
- [ ] Fix storybook so that preview also sets composites' locale
- [ ] Update storybook documentation to describe the new `CompositeLocale` and friends.


packages/react-composites/src/composites/localization/LocalizationProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+/**
+ * @description Provider to provide localized strings for this library's react components.
+ * @remarks Components will be provided localized strings in English (US) by default if this
+ * provider is not used
+ */
+export const LocalizationProvider = (props: LocalizationProviderProps): JSX.Element => {
+  const { locale, children } = props;
+  return <LocaleContext.Provider value={locale}>{children}</LocaleContext.Provider>;
```
mgamis-msft, line 53: Nice. This should work This needs the LocalizationProvider from `react-components` like so:
`  return (
    <LocaleContext.Provider value={locale}>
      <ComponentLocalizationProvider locale={locale.component}>{children}</ComponentLocalizationProvider>
    </LocaleContext.Provider>
  );`
  where `ComponentLocalizationProvider` is imported:
  `import {  LocalizationProvider as ComponentLocalizationProvider } from '@internal/react-components';`

prprabhu-ms: Right you are. Done.

mgamis-msft (approved):

PR #716, opened 2021-08-25
[DRAFT] [NO-REVIEW] Custom Icons

anjulgarg:
# What
Implementation of custom icons and custom icon injection in composites.

# Why
Allows a user to inject custom JSX elements to override the default icons in composites and components

# How Tested

# Process & policy checklist


packages/communication-react/review/communication-react.api.md
```diff
@@ -205,6 +205,16 @@ export type AzureCommunicationChatAdapterArgs = {
     threadId: string;
 };
 
+// @public (undocumented)
+export interface BaseCompositeProps {
+    fluentTheme?: PartialTheme | Theme;
+    icons?: CompositeIcons;
```
JamesBurnside, line 211: Composite icons will vary per composite - e.g. call composite won't have messageSend icon - I think we should explicitly state icons per composite and not have them inherit all default icons

JamesBurnside: Maybe using the `Pick` syntax to select the icons out of the default icons object for use in the composite to ensure no non-existent icons

JamesBurnside: cc @anjulgarg

PR #718, opened 2021-08-25
[Not to check in] Internal api review

PorterNan:
# What

Let's use this PR for internal discussion for api review
Mainly focus on:
1. Meeting composite api review
2. Localization
3. Error handling
4. New Customization api for composite
5. Html bundle api
6. ClientState modifier
7. More api changes after private preview here

Also, a reminder for some points mentioned from last time azure review (notes taken by Allen):
- Don't pass in CallClient to declaratify and just create it inside declaratify
- tokenCredential variable rename to userCredential in storybook example
- Can we avoid connection strings in storybook
- Pass client to handler to support other languages what don't have closures
- Do we need createStatefulChatClient or could we just call new StatefulChatClient
- What happens if state is desync
- How to clear state and rehydrate
- How to control size of state
- How can customers know when there is a desync to then call resync
- Avoid exporting class and export the interface instead (also reference interface instead of class in parameters)

# Why

# How Tested

# Process & policy checklist


review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+import { UnknownIdentifier } from '@azure/communication-common';
+import { UnknownIdentifierKind } from '@azure/communication-common';
+import { VideoDeviceInfo } from '@azure/communication-calling';
+
+// @public (undocumented)
+export type AllKeys<T> = {
+    [K in keyof T]: T[K] extends never ? never : K;
+};
+
+// @public
```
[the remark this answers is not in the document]
JamesBurnside: I like the idea of having `loadLocale('ar_SA')` and then have `locales = 'ar_SA' | 'en_CA' etc.`.

Doing `import { ar_SA } from '@azure/communication-react';` is kinda odd - like its reads out loud as if I am importing a language moniker but no information that that moniker represents locales for components & composites.

Probably something to ask about in the review though, don't think we need to change now.

My understanding is lazy loading is achieved through a good bundler and not anything we provide. Something like:
```js
/// myLazyLang.ts
import { ar_SA } from "@azure/communication-react";
export default ar_SA;

/// someOtherFile.ts
const lazyLoadedAR_SA = import(/*webpackChunkName: "myLazyLang" */ "./myLazyLang")).default;
```

Not sure if treeshaking would handle that correctly though and create a chunk with just that language

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+}
+
+// @public (undocumented)
+export type ControlBarLayoutType = 'horizontal' | 'vertical' | 'dockedTop' | 'dockedBottom' | 'dockedLeft' | 'dockedRight' | 'floatingTop' | 'floatingBottom' | 'floatingLeft' | 'floatingRight';
+
+// @public
+export interface ControlBarProps {
+    children?: React_2.ReactNode;
+    layout?: ControlBarLayoutType;
+    styles?: BaseCustomStylesProps;
```
JamesBurnside, line 814: `BaseCustomStylesProps` is very generic, should this be whatever the control bar parent actually is? e.g. `IStackStyles`

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+}
+
+// @public (undocumented)
+export const createAzureCommunicationCallAdapter: ({ userId, displayName, credential, locator, callClientOptions }: AzureCommunicationCallAdapterArgs) => Promise<CallAdapter>;
+
+// @public (undocumented)
+export const createAzureCommunicationChatAdapter: ({ endpointUrl, userId, displayName, credential, threadId }: AzureCommunicationChatAdapterArgs) => Promise<ChatAdapter>;
+
+// @public (undocumented)
+export const createDefaultCallingHandlers: (callClient: StatefulCallClient, callAgent: CallAgent | undefined, deviceManager: StatefulDeviceManager | undefined, call: Call | undefined) => {
```
JamesBurnside, line 824: Should this be an interface? e.g. `DefaulCallingHandlers`

JamesBurnside: In fact yep it should, weh ave one for chat below

JamesBurnside: DefaultCallingHandlers already exists below, think this just needs updated to use that as the return type

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+    updateThreadTopicName: (topicName: string) => Promise<void>;
+    onLoadPreviousChatMessages: (messagesToLoad: number) => Promise<boolean>;
+    onDismissErrors: (errorTypes: ErrorType[]) => void;
+};
+
+// @public
+export const defaultIdentifiers: Identifiers;
+
+// @public (undocumented)
+export type DefaultMessageRendererType = (props: MessageProps, ids?: {
```
JamesBurnside, line 917: `Default` and `Type` seem weird here - should this just be `MessageRenderer`? (or at least just `MessageRendererType`)

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+
+// @public
+export interface LocalVideoStreamState {
+    mediaStreamType: MediaStreamType;
+    source: VideoDeviceInfo;
+    view?: VideoStreamRendererViewState;
+}
+
+// @public
+export const MeetingComposite: (props: MeetingCompositeProps) => JSX.Element;
```
JamesBurnside, line 1121: I will move this out of this review (hoping I can tag as @alpha and it will go into its own api file) - the PRs for it won't be ready before wednesday

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+export type Message<T extends MessageTypes> = {
+    type: T;
+    payload: T extends 'chat' ? ChatMessagePayload : T extends 'system' ? SystemMessagePayload<'participantAdded' | 'participantRemoved'> | SystemMessagePayload<'topicUpdated'> | SystemMessagePayload<'content'> : CustomMessagePayload;
+};
+
+// @public (undocumented)
+export type MessageAttachedStatus = 'bottom' | 'top';
+
+// @public (undocumented)
+export type MessageContentType = 'text' | 'html' | 'richtext/html' | 'unknown';
```
JamesBurnside, line 1141: Why do we have `richtext/html` when https://github.com/Azure/azure-sdk-for-js/blob/a79a2444b91e850d3bba4139a635d411a162d1c5/sdk/communication/communication-chat/review/communication-chat.api.md `ChatMessageType` does not?

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+// @public
+export type ChatErrors = {
+    [target in ChatErrorTarget]: Error;
+};
+
+// @public
+export type ChatErrorTarget = 'ChatClient.createChatThread' | 'ChatClient.deleteChatThread' | 'ChatClient.getChatThreadClient' | 'ChatClient.listChatThreads' | 'ChatClient.off' | 'ChatClient.on' | 'ChatClient.startRealtimeNotifications' | 'ChatClient.stopRealtimeNotifications' | 'ChatThreadClient.addParticipants' | 'ChatThreadClient.deleteMessage' | 'ChatThreadClient.getMessage' | 'ChatThreadClient.getProperties' | 'ChatThreadClient.listMessages' | 'ChatThreadClient.listParticipants' | 'ChatThreadClient.listReadReceipts' | 'ChatThreadClient.removeParticipant' | 'ChatThreadClient.sendMessage' | 'ChatThreadClient.sendReadReceipt' | 'ChatThreadClient.sendTypingNotification' | 'ChatThreadClient.updateMessage' | 'ChatThreadClient.updateTopic';
+
+// @public (undocumented)
+export type ChatMessage = Message<'chat'>;
```
JamesBurnside, line 656: This conflicts with the imported type `ChatMessage` - can we avoid it? (why we see ChatMessage_2 throughout this doc)

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+    messageStatusContainer?: (mine: boolean) => IStyle;
+    newMessageButtonContainer?: IStyle;
+    systemMessageContainer?: ComponentSlotStyle;
+}
+
+// @public (undocumented)
+export type MessageTypes = 'chat' | 'system' | 'custom';
+
+// @public
+export const MicrophoneButton: (props: MicrophoneButtonProps) => JSX.Element;
```
JamesBurnside, line 1235: Fluents definitions are little different to this, they extend a react functional component - do we want to align with them?
e.g.
image

https://github.com/microsoft/fluentui/blob/master/packages/react/etc/react.api.md

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+export const useChatThreadClient: () => ChatThreadClient;
+
+// @public (undocumented)
+export const useDeviceManager: () => StatefulDeviceManager | undefined;
+
+// @public (undocumented)
+export const useIdentifiers: () => Identifiers;
+
+// @public (undocumented)
+export const usePropsFor: <Component extends (props: any) => JSX.Element>(component: Component, type?: "chat" | "calling" | undefined) => ChatReturnProps<Component> extends never ? CallingReturnProps<Component> extends never ? undefined : CallingReturnProps<Component> : ChatReturnProps<Component>;
```
JamesBurnside, line 1700: Can we use interface definitions to make these more readable?
e.g.
```
export const usePropsFor: <T extends React.Component>(component: T, componentType?: SDKType) => ComponentProps<T>
```

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+}
+
+// @public
+export interface VideoGalleryRemoteParticipant extends VideoGalleryParticipant {
+    isSpeaking?: boolean;
+    screenShareStream?: VideoGalleryStream;
+}
+
+// @public (undocumented)
+export const videoGallerySelector: OutputParametricSelector<CallClientState, CallingBaseSelectorProps, {
```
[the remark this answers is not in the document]
JamesBurnside: Here is why (nice mini read too): https://github.com/microsoft/rushstack/issues/2799#issuecomment-881804485

review/communication-react.api.md
```diff
@@ -0,0 +1,1828 @@
...
+}
+
+// @public (undocumented)
+export interface VideoTileStylesProps extends BaseCustomStylesProps {
+    displayNameContainer?: IStyle;
+    overlayContainer?: IStyle;
+    videoContainer?: IStyle;
+}
+
+// (No @packageDocumentation comment for this package)
```
JamesBurnside, line 1826: we should add package documentation for this package as its our primary export

review/communication-react.api.md
PorterNan: Looks like this one only contains a root property, which is more accurate for this case:
image

review/communication-react.api.md
PorterNan: Assigned to myself

review/communication-react.api.md
PorterNan: Assigned to myself

review/communication-react.api.md
PorterNan: Their buttons are just parts of our implementation, I believe they keep it in a class style for legacy code support - for us, we can just align functional component all around the repo

review/communication-react.api.md
PorterNan: Since you are the author for this package, do you have idea of package documentation? I will leave this as an open question since it is not necessary for azure review though

review/communication-react.api.md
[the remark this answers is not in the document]
PorterNan: Nice! Thanks for this information!

review/communication-react.api.md
JamesBurnside: Not sure what is expected here but I'll assign this to myself

prprabhu-ms:
I think we handled everything here. Let's close this now?

PorterNan:
Close for now, leave some comments unresolved for further actions

PR #731, opened 2021-08-27
Move check-treeshaking to a new tools dir

prprabhu-ms:
# What
Create a new `tools/` top level directory and move the `check-treeshaking` project there.

# Why
Preparation to add a couple more tools. Limit the `packages/` folder to packlets and packages that are part of our released product. Thus, everything in `packages/` will be soon forced to have the same package version.

# How Tested
`rush build`


.github/workflows/ci.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           fi
       # Tree-shaking check
       - name: Treeshaking check
-        run: rush build --only @internal/check-treeshaking
+        run: rush build --only @internal-tools/check-treeshaking
```
JamesBurnside, line 64: Can this stay as `@internal/check-treeshaking`? The reason for the `@internal` was from a security concern to avoid any scenario where someone creates a malicious package with that name and publishes it on npm - then somewhere down the line this is included as a dev dependency or someone does an npm install on the repo and it pull the malicious package code instead

prprabhu-ms: Done.

@internal doesn't seem to be handled specially by NPM? It's just an unlikely NPM prefix, right? I thought @internal-tools would also meet that criterion, and would let us keep tools and release-targeted packages separate.

But I agree that having multiple such prefixes will add confusion, so reverted to @internal.

JamesBurnside: I remember reading somewhere about it being a reserved organization name... but I can't find any source on that...

If I try to make it at least it fails:
image

JamesBurnside (approved):

alcail (approved):

PR #737, opened 2021-08-30
Locale exports small fix

mgamis-msft:
# What

- Fixed English (US) and English (GB) composite locales mix up
- Fixed touchdown build yml trigger to only run on main
- Refactor locale variables

# Why

- Correct exports for English (US) and English (GB)
- Trigger touchdown build to only run on main

# How Tested

# Process & policy checklist


packages/react-components/src/localization/locales/index.ts
```diff
@@ -3,54 +3,54 @@
...
+import nl_NL from './nl-NL/strings.json';
+import pt_BR from './pt-BR/strings.json';
+import ru_RU from './ru-RU/strings.json';
+import tr_TR from './tr-TR/strings.json';
+import zh_CN from './zh-CN/strings.json';
+import zh_TW from './zh-TW/strings.json';
 
 const createComponentStrings = (localizedStrings: PartialDeep<ComponentStrings>): ComponentStrings => {
-  const strings: ComponentStrings = { ..._en_US };
+  const strings: ComponentStrings = { ...en_US };
```
prprabhu-ms, line 22: Tangential to this PR, but why do we fallback to `en_US` here? It's a good thing if `tsc` complains when a string is missing in some locale -- ensures that missing localization strings are caught as build errors.

prprabhu-ms (approved):

packages/react-components/src/localization/locales/index.ts
mgamis-msft: I wanted it to be easier for someone to create a PR with a localized string such that they only need to add the string to en_US strings otherwise there would be build errors. The touchdown pipeline would then handle the translation of that added string after the PR has been merged to main. The TAT on new strings from Touchdown is 5 days so I figure it be best to separate the process.

alcail (approved):

PR #738, opened 2021-08-30
Active Speaker Indicator

anjulgarg:
# What
Active Speaker Indicator in Video Gallery

# Why
When a remote participant is speaking, you can see a border around their video tile.

# How Tested

# Process & policy checklist


packages/react-components/src/components/styles/VideoTile.styles.ts
```diff
@@ -12,6 +12,10 @@ export const rootStyles: IStyle = {
   borderRadius: theme.effects.roundedCorner4
 };
 
+export const isSpeakingStyles: IStyle = {
+  border: '4px solid #0078D4'
```
PorterNan, line 16: Use rem instead? 0.25rem for this case

PorterNan (approved):

packages/react-components/src/components/styles/VideoTile.styles.ts
prprabhu-ms, line 16: The color for this border should also be picked from the theme palette. This hard-coded color may not look good on inverted backgrounds etc.

prprabhu-ms (approved):

alcail (approved):

packages/react-components/src/components/styles/VideoTile.styles.ts
anjulgarg: Yes. I dreamt of this last night :D

PR #739, opened 2021-08-31
Localized strings in call and chat composites

mgamis-msft:
# What

Localized strings in call composite

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2555551

# How Tested

Checked that calling sample and composite in storybook is working.

# Process & policy checklist


anjulgarg:
Are we planning to add other languages for these strings?

packages/react-composites/src/composites/CallComposite/StartCallButton.tsx
```diff
@@ -5,16 +5,18 @@ import { PrimaryButton } from '@fluentui/react';
...
   isDisabled: boolean;
 }
 
-const buttonText = 'Start call';
-
 export const StartCallButton = (props: StartCallButtonProps): JSX.Element => {
   const { isDisabled, onClickHandler } = props;
+  const locale = useLocale();
+
+  const buttonText = locale.strings.call.startCallButtonText;
```
prprabhu-ms, line 19: nit: Simply inline in line 29. It's used exactly once.

prprabhu-ms (approved):

common/scripts/ensure-consistent-versions.js
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
```
anjulgarg, line 1: Why is this script here?
I thought it was part of your PR @prprabhu-ms  ?
cc @mgamis-msft

mgamis-msft:
> Are we planning to add other languages for these strings?

Yes, but that is handled by a separate CI after this has been merged to main.

common/scripts/ensure-consistent-versions.js
mgamis-msft: Not sure how this happened after I merged main to my branch but I have now pruned these other changes from this PR.

alcail (approved):

PR #740, opened 2021-08-31
Remove redundant `show` prefix from `visualElements` fields

prprabhu-ms:
# What
* Drops the redundant prefix, also removing the confusion around enabled-by-default vs disabled-by-default visual elements.

Thanks to @anjulgarg for a review comment to this effect.

### DIFF
--- a/change/@internal-react-composites-3d05cbe2-320d-4ac9-ac5b-8a35f47c72de.json
+++ b/change/@internal-react-composites-3d05cbe2-320d-4ac9-ac5b-8a35f47c72de.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Drop `show` prefix from `visualElements` fields",
+  "packageName": "@internal/react-composites",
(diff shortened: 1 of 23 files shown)


alcail:
I personnally don't agree with this removing of 'show' prefix.
'errorBar', participantPane' and 'topic' are not good names for booleans IMO

JamesBurnside:
> I personnally don't agree with this removing of 'show' prefix.
> 'errorBar', participantPane' and 'topic' are not good names for booleans IMO, even if they are inside a VisualElements object

I'm happy either way, but I think I lean to having an indicator this is is a boolean - for example having just `errorBar` could be read as being able to pass in your own errorBar component to this prop.

Maybe there is a more idcative name `VisualElements` could be? (but naming is hard and I can't think of one...)

JamesBurnside (approved):

prprabhu-ms:
On balance, I agree with Aline. Closing this @anjulgarg poke me if you _really_ want this change to land.

PR #742, opened 2021-08-31
Dominant Speakers h/s and implementation in video gallery.

anjulgarg:
# What
Adds Dominant Speaker support to Video Gallery

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/calling-component-bindings/review/calling-component-bindings.api.md
+++ b/packages/calling-component-bindings/review/calling-component-bindings.api.md
@@ -18,6 +18,8 @@ import { CameraButton } from '@internal/react-components';
 import { Common } from '@internal/acs-ui-common';
 import { CommunicationUserIdentifier } from '@azure/communication-common';
 import { DeviceManagerState } from '@internal/calling-stateful-client';
+import { DominantSpeakers } from '@internal/react-components';
+import { DominantSpeakersInfo } from '@azure/communication-calling';
 import { EndCallButton } from '@internal/react-components';
 import { ErrorBar } from '@internal/react-components';
 import { ErrorType } from '@internal/react-components';
@@ -195,6 +197,9 @@ export const getDeviceManager: (state: CallClientState) => DeviceManagerState;
 export const getDisplayName: (state: CallClientState) => string | undefined;
 
 // @public (undocumented)
+export const getDominantSpeakers: (state: CallClientState, props: CallingBaseSelectorProps) => undefined | DominantSpeakersInfo;
+
+// @public (undocumented)
 export const getIdentifier: (state: CallClientState) => string;
 
 // @public (undocumented)
@@ -317,9 +322,10 @@ renderElement: HTMLElement | undefined;
 };
 };
 remoteParticipants: VideoGalleryRemoteParticipant[];
+dominantSpeakers: DominantSpeakers | undefined;
 }, (res1: string | undefined, res2: {
 [keys: string]: RemoteParticipantState;
-} | undefined, res3: LocalVideoStreamState[] | undefined, res4: boolean | undefined, res5: boolean | undefined, res6: string | undefined, res7: string) => {
+} | undefined, res3: LocalVideoStreamState[] | undefined, res4: boolean | undefined, res5: boolean | undefined, res6: string | undefined, res7: string, res8: DominantSpeakersInfo | undefined) => {
 screenShareParticipant: VideoGalleryRemoteParticipant | undefined;
 localParticipant: {
 userId: string;
@@ -333,6 +339,7 @@ renderElement: HTMLElement | undefined;
 };
 };
 remoteParticipants: VideoGalleryRemoteParticipant[];
+dominantSpeakers: DominantSpeakers | undefined;
 }>;
 
 // (No @packageDocumentation comment for this package)
--- a/packages/react-components/review/react-components.api.md
+++ b/packages/react-components/review/react-components.api.md
@@ -250,6 +250,12 @@ export type DefaultMessageRendererType = (props: MessageProps, ids?: {
 }) => JSX.Element;
 
 // @public
+export interface DominantSpeakers {
+    speakersList: ReadonlyArray<string>;
+    timestamp: Date;
+}
+
+// @public
 export const EndCallButton: (props: EndCallButtonProps) => JSX.Element;
 
 // @public
@@ -690,6 +696,7 @@ export type VideoGalleryParticipant = {
 
 // @public
 export interface VideoGalleryProps {
+    dominantSpeakers?: DominantSpeakers;
     layout?: 'default' | 'floatingLocalVideo';
     localParticipant: VideoGalleryLocalParticipant;
     localVideoViewOption?: VideoStreamOptions;
--- a/packages/react-components/src/components/VideoGallery.tsx
+++ b/packages/react-components/src/components/VideoGallery.tsx
@@ -6,6 +6,7 @@ import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { useIdentifiers } from '../identifiers/IdentifierProvider';
 import {
   BaseCustomStylesProps,
+  DominantSpeakers,
   OnRenderAvatarCallback,
   VideoGalleryLocalParticipant,
   VideoGalleryRemoteParticipant,
@@ -46,6 +47,8 @@ export interface VideoGalleryProps {
   localVideoViewOption?: VideoStreamOptions;
   /** Remote videos view options */
   remoteVideoViewOption?: VideoStreamOptions;
+  /** A list of speakers ordered by most active to least active in a call. */
+  dominantSpeakers?: DominantSpeakers;
   /** Callback to create the local video stream view */
   onCreateLocalStreamView?: (options?: VideoStreamOptions) => Promise<void>;
   /** Callback to dispose of the local video stream view */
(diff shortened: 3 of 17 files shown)


packages/react-components/review/react-components.api.md
```diff
@@ -249,6 +249,12 @@ export type DefaultMessageRendererType = (props: MessageProps, ids?: {
     messageTimestamp?: string;
 }) => JSX.Element;
 
+// @public
+export interface DominantSpeakers {
+    speakersList: ReadonlyArray<string>;
+    timestamp: Date;
```
JamesBurnside, line 255: Is/will `timestamp` used in the components anywhere? If not can we remove it from the react-components API?

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -119,15 +147,16 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
     onRenderAvatar,
-    showMuteIndicator
+    showMuteIndicator,
+    dominantSpeakers
   } = props;
   const [sortedRemoteParticipants, setSortedRemoteParticipants] = useState<VideoGalleryRemoteParticipant[]>([]);
 
   const ids = useIdentifiers();
 
   useEffect(() => {
```
JamesBurnside, line 157: Is there a reason this is inside a `useEffect`? using useEffect with `set` will mean two separate renders and looks like this is just to avoid recreating the sort every render.
Could this instead use a useRef/useMemo/useCallback (not sure which is most applicable) and be something like:
```
const sortedRemoteParticipants = useMemo(() => sortParticipants(remoteParticipants), [remoteParticipants]);
```

JamesBurnside (approved):

packages/react-components/review/react-components.api.md
anjulgarg: Not yet, but probably will once we tweak the implementation.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Good point. I overlooked the fact this this is inside useEffect. Let me see if I can memoize it.

packages/react-components/review/react-components.api.md
JamesBurnside: Can it be removed until it is needed? UI components don't need to be driven by the same props as state so potentially won't be required (I'm thinking timestamp won't be needed if we want to keep the components using it pure and instead the speakersList would dictate order)

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -75,26 +78,51 @@ const DRAG_OPTIONS: IDragOptions = {
...
-    }
-    if (!p2?.videoStream?.renderElement?.childElementCount) {
-      return -1;
-    }
+  const participantsWithVideo: VideoGalleryRemoteParticipant[] = [];
+  const participantsWithoutVideo: VideoGalleryRemoteParticipant[] = [];
 
-    return 0;
+  participants.forEach((p) => {
+    if (p.videoStream?.renderElement?.childElementCount) participantsWithVideo.push(p);
```
prprabhu-ms, line 94: Please don't use the short syntax for `if` that skips `{}`. It has caused severe security issues in the past.

If you don't mind walking the list twice, you can use a filter:

    const participantsWithVideo = participants.filter((p) => p.videoStream?.renderElement?.childElementCount > 0);
    const participantsWithoutVideo = participants.filter((p) => p.videoStream?.renderElement?.childElementCount === 0);

I do prefer your current use of `forEach`, but recommend adding curly braces.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -75,26 +78,51 @@ const DRAG_OPTIONS: IDragOptions = {
...
+    else participantsWithoutVideo.push(p);
   });
+
+  // If dominantSpeakers are available, we sort the video tiles basis on dominant speakers.
+  if (dominantSpeakers) {
+    participantsWithVideo.sort((a, b) => {
+      const idxA = dominantSpeakers.speakersList.indexOf(a.userId);
```
prprabhu-ms, line 101: This potentially has very bad performance because it'll walk `speakersList` once for each `participantWithVideo`.

Instead, convert `speakersList` to a map `userId: index` first.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -75,26 +78,51 @@ const DRAG_OPTIONS: IDragOptions = {
...
 
-    return 0;
+  participants.forEach((p) => {
+    if (p.videoStream?.renderElement?.childElementCount) participantsWithVideo.push(p);
+    else participantsWithoutVideo.push(p);
   });
+
+  // If dominantSpeakers are available, we sort the video tiles basis on dominant speakers.
+  if (dominantSpeakers) {
+    participantsWithVideo.sort((a, b) => {
```
prprabhu-ms, line 100: Unfortunately, older browsers do not guarantee stable sorting of arrays: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description

So this might lead to tiles getting reordered on rerender for no reason.

A different approach would give stable sorting and will perform better:

    participantsWithVideoById = new Map();
    participantsWithVideo.forEach((p) => {
        participantsWithVideoById[p.userId] = p;
    })

    sortedParticipantsWithVideoById = new Map();
    dominantSpeakers.speakersList.forEach((userId) => {
        if (participantsWithVideoById[userId] !== undefined) {
            sortedParticipantsWithVideo[userId] = participantsWithVideoById[userId];
        }
    })
    participantsWithVideo.forEach((p) => {
        if(sortedParticipantsWithId[p.userId] === undefined) {
            sortedParticipantsWithId[p.userId] = p;
        }
    })
    return Array.from(sortedParticipantsWithId.values());

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -119,15 +147,16 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     styles,
     layout,
     onRenderAvatar,
-    showMuteIndicator
+    showMuteIndicator,
+    dominantSpeakers
   } = props;
   const [sortedRemoteParticipants, setSortedRemoteParticipants] = useState<VideoGalleryRemoteParticipant[]>([]);
```
prprabhu-ms, line 153: This should be `useRef` not `useState`.

You have not (correctly!) dropped the `useEffect` block, so this doesn't need to be a state variable.

State variables should track information that changes internal to this component *without any change from props*. That is not the case here. This is just caching some computed values from props.

packages/react-components/src/types/Call.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface DominantSpeakers {
+  /** A list of speakers who are dominant speakers in a call.
+   * The most dominant speaker is at index 0.
+   * The second dominant at 1 and so on.
+   * */
+  speakersList: ReadonlyArray<string>;
+  /**
+   * Timestamp of when this object was created.
+   */
+  timestamp: Date;
```
prprabhu-ms, line 18: I'm assuming this is here to potentially avoid too frequent updates?

nit: The comment above is a bit misleading. This is the timestamp when dominant speakers were last updated (not this particular object, which is cached and what not).

prprabhu-ms (approved):
LGTM, pending some non-trivial logic tweaking.

prprabhu-ms:
> Can it be removed until it is needed? UI components don't need to be driven by the same props as state so potentially won't be required (I'm thinking timestamp won't be needed if we want to keep the components using it pure and instead the speakersList would dictate order) 

I like this idea too. Bindings are where logic for avoiding frequent reorders should live. So, bindings should avoid reordering the list.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Sure, I can change it. But curious what kind of security issues it can cause.

packages/react-components/src/types/Call.ts
anjulgarg: Right. Will change the language.

packages/react-components/review/react-components.api.md
anjulgarg: Removing it.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Good Suggestion.

packages/react-components/src/types/Call.ts
anjulgarg: Infact, removing it for now. You guyz are correct. Logic should reside in bindings.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Could you point me to the source of this claim?
Replacing `useState` with `useRef` here will prevent the component from re-rendering when remote participants are sorted.
And hence, need a state variable.

prprabhu-ms: Heh: s/not/now -- I was saying that you have *correctly* dropped the `useEffect` block, but typo'ed to "not correctly". Classic :P

packages/react-components/src/components/VideoGallery.tsx
prprabhu-ms: It's a very notorious SSL bug: https://www.imperialviolet.org/2014/02/22/applebug.html

packages/calling-component-bindings/review/calling-component-bindings.api.md
```diff
@@ -194,6 +196,9 @@ export const getDeviceManager: (state: CallClientState) => DeviceManagerState;
 // @public (undocumented)
 export const getDisplayName: (state: CallClientState) => string | undefined;
 
+// @public (undocumented)
+export const getDominantSpeakers: (state: CallClientState, props: CallingBaseSelectorProps) => undefined | DominantSpeakersInfo;
```
prprabhu-ms, line 200: This shouldn't be part of the public API?

This isn't used by any of our components directly, so it shouldn't be exposed.

Actually, I think there are other base selectors that need to be hidden.

packages/react-components/src/components/VideoGallery.tsx
prprabhu-ms: Actually, my comment above is not fully correct. You don't need `useRef` either. `useMemo` already returns the same value for same inputs (so caching happens in a global context internal to `useMemo`). So you are good to go with just a local variable.

    const sortedRemoteParticipants = useMemo(() => sortParticipants(remoteParticipants, dominantSpeakers), [remoteParticipants, dominantSpeakers])

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: We don't even support older browsers for calling. The only 2 browsers supported atm are chrome and safari.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Haha. The way I used if/else would never have caused that. But added brackets anyway.

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Ok I understand what you mean now. Thanks, good suggestion.

packages/calling-component-bindings/review/calling-component-bindings.api.md
anjulgarg: So currently, everything inside baseSelector is being exported through index file.
Can create a bug for this cleanup.

PR #746, opened 2021-09-01
Alcail/fix sharing name

alcail:
# What
displayed name and mic and screen sharer VideoTile
Updated screen name and removed mic from screen sharing tile

image
image

# Why
bug where a dot was displayed on screen sharer VideoTile: https://skype.visualstudio.com/SPOOL/_workitems/edit/2550271

# How Tested
ran call sample and storybook locally
temporarily logged info to verify styles changes were done only when necessary in useMemo

# Process & policy checklist


packages/react-components/src/components/VideoTile.tsx
```diff
@@ -156,32 +156,34 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+          <Stack
+            horizontal
+            className={mergeStyles(
+              isVideoRendered ? videoHint : disabledVideoHint,
+              // when video is being rendered, the displayName has a grey-ish background, so no use of theme
+              { color: isVideoRendered ? palette.neutralPrimary : theme.palette.neutralPrimary },
+              styles?.displayNameContainer
             )}
+          >
+            <Stack>
```
prprabhu-ms, line 169: Do you need this `Stack` as there is a single item in here?

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -156,32 +156,34 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+                <Text
+                  className={mergeStyles(displayNameStyle, {
+                    color: isVideoRendered ? palette.neutralPrimary : theme.palette.neutralPrimary
+                  })}
+                >
+                  {displayName}
+                </Text>
+              )}
+            </Stack>
+            <Stack className={mergeStyles(iconContainerStyle)}>
```
prprabhu-ms, line 180: Do you need this `Stack` with a single item? You probably want a `Stack.Item` in both instances.

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -156,32 +156,34 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+              {showMuteIndicator &&
```
prprabhu-ms, line 181: You probably want to drop the surrounding `Stack` (or `Stack.Item`, see my other comments) as well when you decided not to render the `Icon`, instead of rendering an empty `Stack`.

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -156,32 +156,34 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+              // when video is being rendered, the displayName has a grey-ish background, so no use of theme
+              { color: isVideoRendered ? palette.neutralPrimary : theme.palette.neutralPrimary },
+              styles?.displayNameContainer
             )}
+          >
+            <Stack>
+              {displayName && (
+                <Text
+                  className={mergeStyles(displayNameStyle, {
+                    color: isVideoRendered ? palette.neutralPrimary : theme.palette.neutralPrimary
```
prprabhu-ms, line 173: This logic is duplicated here and on line 165.
Actually, do you need this here at all? It should be inherited from the surrounding Stack, no?

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -156,32 +156,34 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
-          <Stack>
-            {displayName && (
-              <Text
-                className={mergeStyles(displayNameStyle, {
-                  color: isVideoRendered ? palette.neutralPrimary : theme.palette.neutralPrimary
-                })}
-              >
-                {displayName}
-              </Text>
+        {(displayName || showMuteIndicator) && (
```
prprabhu-ms, line 159: The logic here is dense, and probably slightly wrong?

Should this be:

    (displayName || (showMuteIndicator && isMuted !== undefined))

It's probably more readable, and not all that wrong, to unconditionally create this `Stack`. If all the contents end up being skipped, it will be an empty `Stack`. It'll be hard to keep the conditionals correctly synchronized as it stands.

packages/react-composites/src/composites/CallComposite/ScreenShare.tsx
```diff
@@ -84,8 +84,9 @@ export const ScreenShare = (props: ScreenShareProps): JSX.Element => {
 
     return (
       <VideoTile
-        displayName={screenShareParticipant?.displayName}
-        isMuted={screenShareParticipant?.isMuted}
+        displayName={
+          screenShareParticipant?.displayName ? `${screenShareParticipant?.displayName}'s screen` : undefined
```
prprabhu-ms, line 88: This will break localization. "'s screen" is not a valid suffix beyond English.

alcail: I followed Figma's naming
image

packages/react-composites/src/composites/CallComposite/ScreenShare.tsx
```diff
@@ -99,7 +100,11 @@ export const ScreenShare = (props: ScreenShareProps): JSX.Element => {
         {videoStream && videoStream.isAvailable && videoStream.renderElement && (
           <Stack horizontalAlign="center" verticalAlign="center" className={aspectRatioBoxStyle}>
             <Stack className={aspectRatioBoxContentStyle}>
-              <VideoTile renderElement={<StreamMedia videoStreamElement={videoStream.renderElement ?? null} />} />
+              <VideoTile
+                displayName={screenShareParticipant?.displayName}
+                isMuted={screenShareParticipant?.isMuted}
+                renderElement={<StreamMedia videoStreamElement={videoStream.renderElement ?? null} />}
```
PorterNan, line 106: @prprabhu-ms should we do 
renderElement={videoStream.renderElement ? <StreamMedia videoStreamElement={ videoStream.renderElement } />} : undefined here?

I believe it is more aligned with the logic of previous PRs

Here

packages/react-components/src/components/VideoTile.tsx
alcail: If I remove this here, I end up with display name color being almost unreadable
image

And if I remove this color in line 165, it's the mic icon that is almost unreadable

packages/react-composites/src/composites/CallComposite/ScreenShare.tsx
PorterNan, line 88: Maybe put it into the localization strings? Or just show the name should be good - no one will misunderstand a screensharing stream with a name to a real video stream

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -156,32 +156,34 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+          <Stack
+            horizontal
+            className={mergeStyles(
+              isVideoRendered ? videoHint : disabledVideoHint,
+              // when video is being rendered, the displayName has a grey-ish background, so no use of theme
+              { color: isVideoRendered ? palette.neutralPrimary : theme.palette.neutralPrimary },
```
PorterNan, line 165: If this part is not removable, assign it with a const

And also wrap both mergeStyles with useMemo()

Since videotiles will be rendering a lot in the app, any perf improvement counts

PorterNan: const nametagColorOverride = { color: isVideoRendered ? palette.neutralPrimary : theme.palette.neutralPrimary };

Then

const displayNameContainerStyle = useMemo(() => mergeStyles(...));

PorterNan (approved):

packages/react-composites/src/composites/CallComposite/ScreenShare.tsx
alcail: yep, good catch, @PorterNan

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -37,6 +37,7 @@ test.describe('Call Composite E2E Tests', () => {
       await stubLocalCameraName(page);
       await page.waitForSelector(dataUiId('call-composite-device-settings'));
       await page.waitForSelector(dataUiId('call-composite-local-preview'));
+      await pages[idx].waitForSelector(`${dataUiId('call-composite-start-call-button')}[data-is-focusable="true"]`);
```
alcail, line 40: Added this to fix issue I encountered where actual snapshot was with info on device half updated 
image
So waiting for the button to be enabled , so we have a stable state for comparison

prprabhu-ms: pages[idx] could just be page

prprabhu-ms (approved):

PR #749, opened 2021-09-02
Plumb through diagnostics to show error bar for bad network quality in call

prprabhu-ms:
# What
* Add stateful plumbing for call diagnostics
* Add bindings to bubble up error when speaking while muted

KI: Currently, the error bar shown when speaking while muted can't be dimissed at all. A fix is pending a small rearchitecturing of ErrorBar.

# How Tested
storybook


prprabhu-ms:
I'm still working on fixing the problem that the new error can't be dismissed. Might be worth landing this with the actual binding disabled (so all plumbing is in place) if fixing that issue takes longer.

PTAL.

packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -15,40 +15,48 @@ import { createSelector } from 'reselect';
...
-  }
-  if (latestErrors['Call.mute'] !== undefined) {
-    activeErrors.push('muteGeneric');
-  }
-  if (latestErrors['Call.stopScreenSharing'] !== undefined) {
-    activeErrors.push('stopScreenShareGeneric');
-  }
+    // Errors reported via diagnostics are more reliable than from API method failures, so process those first.
+    if (diagnostics?.media.latest.speakingWhileMicrophoneIsMuted?.value) {
+      activeErrors.push('speakingWhileMuted');
```
prprabhu-ms, line 32: Will comment out this line if I can't fix the KI first.

packages/calling-stateful-client/src/DiagnosticsSubscriber.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+      const media = call.diagnostics?.media.latest;
+      if (media) {
+        media[args.diagnostic] = latestFromEvent(args);
+      }
+    });
+    this.logDiagnostics('media update');
+  }
+
+  private logDiagnostics(operation: string) {
```
prprabhu-ms, line 86: Will remove this before landing.

anjulgarg (approved):

anjulgarg:
LGTM
Not sure if showing an error bar is the best way to go for indicating to a user that they are speaking while their mic is off.

alcail (approved):

alkwa-msft (approved):

packages/calling-component-bindings/src/errorBarSelector.ts
prprabhu-ms: KI fixed in main.

prprabhu-ms:
> LGTM
> Not sure if showing an error bar is the best way to go for indicating to a user that they are speaking while their mic is off.

Teams shows it as a similar banner, though we may not want it to be quite the error banner.

Anyway, I had picked an easy one just to showcase the plumbing. Switched to surfacing network connection error -- the thing I actually was working towards.

PR #753, opened 2021-09-03
Add timestamps to saved errors

prprabhu-ms:
# What
* Add timestamps to errors as they are saved in stateful clients
* Bubble the latest timestamp over to the UI component data

# Why

This timestamp will be used by `ErrorBar` to track what errors have been dismissed by the user.

# How Tested
Storybook composites.

  Changes the type signatures in the API related to error reporting.


anjulgarg (approved):

alcail (approved):

prprabhu-ms:
Checked snapshot diffs locally.

PR #754, opened 2021-09-04
Track dismissed errors internally in `ErrorBar`

prprabhu-ms:
# What

* Track errors that are dismissed internall in the `ErrorBar` component.
  * The component now remembers what errors were dismissed and when.
  * It uses this information together with the active errors from the stateful clients to determine what messages to show on the UI

# Why

* Clearing an error in the stateful clients is not the same as dismissing an `ErrorBar`.
  * An `ErrorBar` is dimissed when the user no longer wants to see it.
  * The error is cleared from stateful when the error condition is gone.

Some examples where this becomes a problem:
* Some `ErrorBar` messages will be driven by call diagnostics, which can't be cleared at all in the stateful client (this is what led me to write this PR).
* Some other UI element might be driven by the same error state in the stateful client (e.g., disable `SendBox` when we know we can't send messages). In that case, dismissing the `ErrorBar` shouldn't enable the `SendBox` as a side effect.

# How Tested

* New jest tests.
 
# Process & policy checklist

  * Changes the behaviour of `ErrorBar` - dismissing errors no longer modifies stateful state.


packages/react-components/src/components/ErrorBar.tsx
```diff
@@ -123,29 +128,125 @@ export interface ActiveError {
...
+    {
+      type: toDismiss.type,
+      dismissedAt: now,
+      activeSince: toDismiss.timestamp
+    }
+  ];
+};
+
+// Returns a new Array iff contents change, to avoid re-rendering when nothing was dropped.
+const maybeDropDismissalsForInactiveErrors = (
```
alcail, line 209: why 'maybe' ?

prprabhu-ms: OK, let's do it each time :D

prprabhu-ms:
@DominikMe by popular demand :D

common/config/rush/pnpm-lock.yaml
```diff
@@ -12390,6 +12390,12 @@ packages:
         optional: true
     resolution:
       integrity: sha512-olV41bKSMm8BdnuMsewT4jqlZ8+3TCARAXjZGT9jcoSnrfUnRCqnMoF9XEeoWjbzObpqF9dRhHQj0Xb9QdF6/w==
+  /jest-react-hooks-shallow/1.5.1:
```
prprabhu-ms, line 12393: Remove. I ended up not using shallow rendering in unittests because they don't work well with React hooks.

The workarounds (and the fragility they add to tests) isn't worth shallow rendering for this simple component.

packages/react-components/src/components/ErrorBar.tsx
```diff
@@ -123,29 +128,115 @@ export interface ActiveError {
...
+    ...dismissedErrors,
+    {
+      type: toDismiss.type,
+      dismissedAt: now,
+      activeSince: toDismiss.timestamp
+    }
+  ];
+};
+
+// Returns a new Array iff contents change, to avoid re-rendering when nothing was dropped.
```
anjulgarg, line 198: if*

anjulgarg (approved):

packages/react-components/src/components/ErrorBar.tsx
prprabhu-ms: I actually mean `iff` here: https://www.merriam-webster.com/dictionary/iff

Will make it if and only if for clarity.

common/config/rush/pnpm-lock.yaml
prprabhu-ms: Done

packages/react-components/src/components/ErrorBar.tsx
```diff
@@ -123,29 +128,115 @@ export interface ActiveError {
...
+    [props.activeErrors, dismissedErrors]
+  );
+
+  const toShow = errorsToShow(props.activeErrors, dismissedErrors);
 
-  // FIXME: Memoize onDismiss callbacks.
   return (
     <Stack>
-      {props.activeErrors.map((activeError) => (
+      {toShow.map((activeError) => (
```
PorterNan, line 155: Rename 'activeError' to 'errorToShow', I thought this was an instance of props.activeErrors at the first glance

packages/react-components/src/components/ErrorBar.tsx
```diff
@@ -123,29 +128,115 @@ export interface ActiveError {
...
+};
+
+// Returns a new Array iff contents change, to avoid re-rendering when nothing was dropped.
+const dropDismissalsForInactiveErrors = (
+  activeErrors: ActiveError[],
+  dismissedErrors: DismissedError[]
+): DismissedError[] => {
+  const active = new Map();
+  for (const error of activeErrors) {
+    active[error.type] = error;
```
PorterNan, line 205: use a plain object or change this line to 
active.set(error.type, error);

This mapObj[index] won't visit it through map, it will add a property into the obj instead 
Check "Setting object properties" in page https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map for reference

packages/react-components/src/components/ErrorBar.tsx
```diff
@@ -123,29 +128,115 @@ export interface ActiveError {
...
+  if (dismissedErrors.some((dismissed) => shouldDeleteDismissal(dismissed))) {
+    return dismissedErrors.filter((dismissed) => !shouldDeleteDismissal(dismissed));
+  }
+  return dismissedErrors;
+};
+
+const errorsToShow = (activeErrors: ActiveError[], dismissedErrors: DismissedError[]): ActiveError[] => {
+  const dismissed: Map<ErrorType, DismissedError> = new Map();
+  for (const error of dismissedErrors) {
+    dismissed[error.type] = error;
```
PorterNan, line 226: Same here, plain obj or use functions in map

PorterNan (approved):
Love this! So we don't need modify state directly from state to our stateful client - which is following all the principles!

packages/react-components/src/components/ErrorBar.tsx
```diff
@@ -103,34 +106,137 @@ export interface ErrorBarStrings {
...
 
-  if (props.activeErrors.length === 0) {
-    return <></>;
-  }
+  const [dismissedErrors, setDismissedErrors] = useState<DismissedError[]>([]);
+
+  // dropDismissalsForInactiveErrors only returns a new object if `dismissedErrors` actually changes.
+  // Without this behaviour, this `useEffect` block would cause a render loop.
+  useEffect(
+    () => setDismissedErrors(dropDismissalsForInactiveErrors(props.activeErrors, dismissedErrors)),
```
PorterNan, line 147: Instead of putting error ref directly into the state, is there an error Id we could use to keep everything on track - obj reference might not be always reliable - you are handling it very well, I am a little worried about changes introduced in the future
But again, this PR is already looking great to me

packages/react-components/src/components/ErrorBar.tsx
prprabhu-ms: Renamed to just `error`. The name `errorsToShow` is already taken by a helper function.

packages/react-components/src/components/ErrorBar.tsx
prprabhu-ms: So far, I've avoided creating an error ID because the errors stem from different sources - some errors are from the API, others are translations of quality signals (see https://github.com/Azure/communication-ui-library/pull/749).

You are right that this can be finicky. That is why this component uses timestamps to track when new errors should be shown (doesn't use object references). Also the unittests do check for the intended behaviour w.r.t. dismissal and new errors ocurring.

prprabhu-ms:
(linux snapshot needs update)

prprabhu-ms:
Still need to update storybook story

alkwa-msft (approved):

PR #755, opened 2021-09-07
Runtime loading of localization

mgamis-msft:
# What

Wrapped LocalizationProvider to allow for asynchronous localization. See AsyncLocales.snippet.tsx for example. Composites will be similar and have 'localeLoader' property

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2542874

# How Tested

Local testing

# Process & policy checklist


packages/storybook/stories/snippets/localization/AsyncLocales.snippet.tsx
```diff
@@ -0,0 +1,39 @@
...
+  COMPONENT_LOCALE_JA_JP,
+  COMPONENT_LOCALE_FR_FR,
+  COMPONENT_LOCALE_EN_US,
+  ComponentLocale,
+  CameraButton
+} from '@azure/communication-react';
+import React from 'react';
+
+export const AsyncLocalesSnippet = (): JSX.Element => {
+  //  mock API functions
```
mgamis-msft, line 12: Could we hose an API service for our locales instead of exporting them?

anjulgarg: Given that we are purely a UI framework. I would discourage it.
Can we use something like azure blob storage directly without an API?

packages/react-composites/src/composites/localization/LocalizationProvider.tsx
```diff
@@ -65,3 +65,67 @@ export const LocalizationProvider = (props: LocalizationProviderProps): JSX.Elem
...
+
+  useEffect(() => {
+    const intializeLocale = async (): Promise<void> => {
+      const compositeLocale = await localeLoader(defaultLocale);
+      setCompositeLocale(compositeLocale);
+    };
+    intializeLocale();
+  }, [defaultLocale, localeLoader]);
+
+  const state = useMemo<SetLocaleContext>(
```
anjulgarg, line 96: Don't see the point of `useMemo` here.

packages/storybook/stories/snippets/localization/AsyncLocales.snippet.tsx
mgamis-msft: I see. We can also move the locales to a separate package. It seems that 'runtime loading' is a bit of a misnomer and we don't actually need to load locales at runtime. It seems the need is for users to only include locales that they need.

packages/react-composites/src/composites/localization/LocalizationProvider.tsx
mgamis-msft: It is not needed here. Thanks for catching that.

packages/communication-react/review/communication-react.api.md
```diff
@@ -1213,14 +1214,12 @@ export interface JumpToNewMessageButtonProps {
...
-// @public
-export const LocalizationProvider: (props: LocalizationProviderProps) => JSX.Element;
-
-// @public
-export type LocalizationProviderProps = {
-    locale: ComponentLocale;
+// @public (undocumented)
+export const LocalizationProvider: (props: {
+    defaultLocale?: string | undefined;
+    locale: ComponentLocale | ((locale?: string | undefined) => Promise<ComponentLocale>);
```
prprabhu-ms, line 1220: Can we take a step back and understand the use-case this addresses.

Is the intention to

1. load the locale asynchronously, but only once at the start, based on hint from browser environment
2. be able to load a different locale based on user action
3. Allow Contoso to pick a locale at *build time* and only include that locale in the bundle.

I'm guessing the answer is 1. 

This API can't handle 2 well, because the Composite implementation won't know when a new locale is available and the whole composite will have to be refreshed when that happens.

If the intention is 3, we don't need any support for dynamic loading of locales.

packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -76,10 +76,74 @@ export type LocalizationProviderProps = {
...
 };
 
 /** React hook to access locale */
 export const useLocale = (): ComponentLocale => useContext(LocaleContext);
+
+interface SetLocaleContext {
+  setLocale: (locale?: string) => Promise<void>;
+}
+
+const SetLocaleContext = createContext<SetLocaleContext>({
```
prprabhu-ms, line 91: use camelCase: `setLocaleContext`, especially because it is currently shadowing the interface definition `SetLocaleContext`

packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -76,10 +76,74 @@ export type LocalizationProviderProps = {
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function
```
prprabhu-ms, line 92: Please drop `@typescript-esling/no-empty-function` by adding a comment in the function body about why this is empty. See https://eslint.org/docs/rules/no-empty-function

You can likely drop the `no-unsed-var` as well by not including that argument:

    const setLocaleContext = createContext<SetLocaleContext>({
        setLocale: async () => { /* this is empty because ... */ }
    });

packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -76,10 +76,74 @@ export type LocalizationProviderProps = {
...
+ */
+export const useSetLocale = (): SetLocaleContext => useContext(SetLocaleContext);
+
+export const LocalizationProvider = (props: {
+  defaultLocale?: string;
+  locale: ComponentLocale | ((locale?: string) => Promise<ComponentLocale>);
+  children: React.ReactNode;
+}): JSX.Element => {
+  const { children, defaultLocale, locale } = props;
+  if (typeof locale === 'function') {
```
prprabhu-ms, line 141: I'd recommend against using typecheck to make logical decisions like this.

Since the synchronous and asynchronous locales work in very different ways, I'd argue it is better to keep the two separate in the API.
So, add two providers to the API: `AsyncLocalizationProvider` and `LocalizationProvider`. Internally, `LocalizationProvider` can make use of `AsyncLocalizationProvider` in the implementation like you do here.

packages/react-composites/src/composites/common/Composite.tsx
```diff
@@ -23,9 +23,15 @@ export interface BaseCompositeProps {
   /**
    * Locale for the composite.
    *
-   * @defaultValue English (US)
+   * @defaultValue English (US) locale
    */
-  locale?: CompositeLocale;
+  locale?: CompositeLocale | ((locale?: string) => Promise<CompositeLocale>);
```
prprabhu-ms, line 28: Similar to my comment about `LocalizationProvider` vs `AsyncLocalizationProvider`, it is better to keep the two ways to set locales separate, since they are very different things, rather than a single field that we type-switch on later on.

packages/react-composites/src/composites/common/Composite.tsx
```diff
@@ -23,9 +23,15 @@ export interface BaseCompositeProps {
...
+  /**
+   * Default string for locale prop if locale prop is type (locale?: string) => Promise<CompositeLocale>
+   *
+   * @defaultValue undefined
+   */
+  defaultLocale?: string;
```
prprabhu-ms, line 34: I am confused  by how the async locale loading works. What provides the non-default locale string to be used for calling the locale loading function?

packages/react-components/review/react-components.api.md
```diff
@@ -358,14 +358,12 @@ export interface JumpToNewMessageButtonProps {
...
-// @public
-export const LocalizationProvider: (props: LocalizationProviderProps) => JSX.Element;
-
-// @public
-export type LocalizationProviderProps = {
-    locale: ComponentLocale;
+// @public (undocumented)
+export const LocalizationProvider: (props: {
+    defaultLocale?: string | undefined;
+    locale: ComponentLocale | ((locale?: string | undefined) => Promise<ComponentLocale>);
```
prprabhu-ms, line 364: As discussed IRL, let's simplify the async loader to `() => Promise<ComponentLocale>` etc. We can also delete `useSetLocale` then.

packages/communication-react/review/communication-react.api.md
prprabhu-ms: This PR actually covers all three of those use-cases. I just want to document what the intended use-cases are.

packages/react-components/review/react-components.api.md
mgamis-msft: Thanks for the suggestion. Done.

packages/react-composites/src/composites/common/Composite.tsx
mgamis-msft: I wonder how we should expose the Composite API to cover both usages `LocalizationProvider` and `AsyncLocalizationProvider`. Two props doesn't (locale and localeLoader) doesn't feel right because both props can be provided.

packages/storybook/stories/snippets/localization/AsyncLocales.snippet.tsx
alkwa-msft: Customers may want to only use a certain number of locales but its the user that needed to choose a specific one which is done at runtime.

1. Lets make the locales available in some form like JSON. 
2. We can go one step further for our internal samples where we host it similar to our token service endpoints. 
3. We then document best practices if someone else wanted to use this JSON.
By doing it like this we have enabled customers to use a number of different localizations but they have to hook it up themselves without adding the burden onto our team for maintaining a locale providing service.

packages/react-composites/src/composites/common/Composite.tsx
prprabhu-ms: I think tagged unions are a good fit for this:

```ts

// Replace with actual type
type CompositeLocale = string;

interface AsyncLocaleLoader {
  kind: 'asyncLoader';
  loader: () => Promise<CompositeLocale>;
  /**
   * If set, the locale to use before locale can be loaded via `loader`.
   * 
   * If unset, en-US locale strings included with this library will be used.
   */
  defaultLocale?: CompositeLocale;
}

interface StaticLocale {
  kind: 'staticValue';
  value: CompositeLocale;
}

type LocaleArgument = AsyncLocaleLoader | StaticLocale;

const newComposite = (arg: LocaleArgument) => {
  switch(arg.kind) {
    case 'asyncLoader':
      console.log(`This is async loader with arguments ${arg.defaultLocale} and ${arg.loader}`);
      break;
    case 'staticValue':
      console.log(`This is static value with arguments ${arg.value}`);
  }
}
```
[playground](https://www.typescriptlang.org/play?#code/FAehAICUFMAcBsCGBja4DuBLALgC3CtgK6LzjYCes0wl14AwgPYC2sTAzjtADJPKk0AXnAdsAJ0wA7AOYBuYMGnZo4gGYo0AQQ4UpyPgPi8miACarwAb2DhwAa2lmAXOADkiXfr7nVbhXbwphbirgAUAJTgQgB84AAK4qyYHNAAPMxsnNyGgjEB4CAAVEW2dkXgAJJqotDYADTkuGhBRmjYTOBEqeAARtBqTOIt-ILgAlJ9I75m4ABumIjgAAZBvuLLAHRl4BU7FdVdUqkN4NBSALQAqgDK4K1jYpKyHODSyPBEFrNYeE0p90wvXEiHEFAwmHgZH6XVSZm2dl2IDKFg0RHg2FyxgA-K5MuwuCosdAFABfRTKVQaVDgG7YRDYTAGUbGaxlRxSFzuMQMpkANVIRGg-jKc0F0DxrAJORZJOA5NoVDQxK04hkRBY52w0XAOj0zLaPhC4AAPrT6YyDYIFMBkExjtqpNB0PjsiodWFQTJXCq1RqtVFYmy7BxfshcJ61ZsOWYojZEeNPGgPF5mes3M4doi7ccmMZNkEZGFlgAVXAAgGefX3YKWX74L3+qTYV4AEisXs2qMQ6MxstJBE54HbnbWIVJywiBQTwOgiHs08TPTcPMtAs+wszCbsOY4eegBaYRdL5deANXTPm4ohf0bmubbY7UbFG4nU7K5NJQA)

packages/react-composites/src/composites/common/Composite.tsx
prprabhu-ms: Resolved since we've dropped the argument to loader now.

packages/react-composites/src/composites/common/Composite.tsx
mgamis-msft: I tried this but it changed the usage in the composite usage to something like this:
```
<ChatComposite adapter={chatAdapter} locale={{ kind: 'asyncLoader', loader: () => COMPOSITE_LOCALE_FR_FR, defaultLocale: COMPOSITE_LOCALE_EN_US}} />}
<CallComposite adapter={callAdapter} locale={{ kind: 'staticValue', value: COMPOSITE_LOCALE_JA_JP}} />
```
I found this to be less delightful.
How about two props:
```
  /**
   * Locale for the composite.
   *
   * @defaultValue English (US) locale
   */
  locale?: CompositeLocale;
  /**
   * Asynchronous callback to change locale
   *
   * @defaultValue undefined
   */
  localeLoader?: () => Promise<CompositeLocale>;
```
Such that when `localeLoader` is defined then we use the `AsyncLocalizationProvider` with `locale` as the default locale. If `localeLoader` is not defined then `LocalizationProvider` is used with `locale`.

packages/communication-react/review/communication-react.api.md
```diff
@@ -1213,14 +1214,12 @@ export interface JumpToNewMessageButtonProps {
...
 export const lightTheme: PartialTheme & CallingTheme;
 
-// @public
-export const LocalizationProvider: (props: LocalizationProviderProps) => JSX.Element;
-
-// @public
-export type LocalizationProviderProps = {
-    locale: ComponentLocale;
+// @public (undocumented)
+export const LocalizationProvider: (props: {
```
prprabhu-ms, line 1218: Let's follow the same API pattern as the adapter here:

    interface LocalizationProviderProps {
        locale?: ComponentLocale;
        loader?: () => Promise<ComponentLocale>;
    }

And add comments that `locale` is used initially, and if `loader` is provided, it is called to asynchronously load the locale to use once it becomes available.

packages/communication-react/review/communication-react.api.md
JamesBurnside: My question is similar to commenting here, I'm confused on the functional use of the async locale loader.

Wouldn't Contoso just load the locale in an async fashion and then update the providers locale?

e.g.
```jsx

const [local, setLocale] = useState<Locale>(defaultLocale);

// inside useEffect:
(async () => {
  // fetch/load locale in an async manner
  const newLocale = await fetchOrLoadLocale('fr');
  setLocale(newLocale);
}());

<LocalizationProvider locale={choseLocale}>
  <Composite />
<LocalizationProvider />

```

mgamis-msft:
Parking this PR for after GA after team discussion. For GA, users can just wrap our composite/components with their own async block. Will reopen if there is a specific customer ask for easier async loading.

PR #393, opened 2021-05-25
Configurable max listeners limit

allenplusplus:
# What
Make the max listeners limit in Stateful Chat and Stateful Calling configurable.

# Why
Allows developers to set their own custom max listeners limit.

# How Tested
Tested manually with Chat and Calling Sample.

# Process & policy checklist

- [X] I have updated the project documentation to reflect my changes if necessary.
- [X] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.

**Is this a breaking change?**

- [x] This change causes current functionality to break.

PR #758, opened 2021-09-08
Alkwa/update calling 1.2.1 beta.1

alkwa-msft:
# What
Updating to 1.2.1.-beta.1 for our acs calling dependency

# Why
The latest update has some mobile improvements that could help our customers

# How Tested
Ran it locally and noticed a build break after CallClientOptions was removed. Removing references to it caused it to build perfect and start a call again locally

# Process & policy checklist


prprabhu-ms:
The `CallClientOptions` passthrough was added in https://github.com/Azure/communication-ui-library/pull/393 to allow Contoso to set the max event listeneres limit.
What would be the way to do that after this PR?

alkwa-msft:
> The `CallClientOptions` passthrough was added in #393 to allow Contoso to set the max event listeneres limit.
> What would be the way to do that after this PR?

The StatefulCallClientOptions had both an event listener property and took in the existing CallClientOptions. CallClientOptions was removed but our StatefulCallClientOptions still has our event listener property in it. 

I believe they removed it because the only thing done in the CallClientOptions was the AzureLogger code which doesn't need to be in CallClientOptions anyways

alcail (approved):

prprabhu-ms (approved):

PR #759, opened 2021-09-08
Add message edited tags to edited messages

PorterNan:
# What

image

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/ChatMessageComponent.tsx
```diff
@@ -162,6 +163,7 @@ export const ChatMessageComponent = (props: ChatMessageProps): JSX.Element => {
             : undefined}
         </Text>
       }
+      details={message.payload.editedOn ? <div className={chatMessageEditedTagStyle(theme)}>Edited</div> : undefined}
```
alkwa-msft, line 166: @mgamis-msft  what files do we need to add this "editted" string to for it to be using our en_us locale?

alkwa-msft (approved):

packages/react-components/src/components/ChatMessageComponent.tsx
PorterNan: Hmm, interesting question, ideally this is still under messageThread component(this component is something for code separation), we can still put it there for external customers

PorterNan: Will send another PR for localization of the whole editing feature once it is all finished

prprabhu-ms (approved):

alcail (approved):

JamesBurnside (approved):

PR #760, opened 2021-09-09
Fix screenshare cancel bug

mgamis-msft:
# What

- Changed error handler in adapter passed to CallComposite to ignore emitted errors due to starting screen share failure.
- Removed starting screen share failures from active error types of errorBar selector so that an error bar does not show 'Failed to start screen sharing' after screen sahring is cancelled.
- Refactored dependency of `useEffect` in call composite `CallScreen` 

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2564836

# How Tested

Calling sample locally

# Process & policy checklist


samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -37,6 +37,12 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
       });
       adapter.on('error', (e) => {
         console.error(e);
+        // TODO: Handle other error types from adapter event emitter. Error type does not
+        // have properties 'operation' and 'error'
+        // Do not call error handler when start screen sharing fails
+        if (e['operation'] === 'Call.startScreenSharing' && e['error']['_code'] === 400) {
```
alkwa-msft, line 43: code style wise this is a little scary to assume these variables like this.

Can we bring them into their own variables and check the correct values?

Unfortunately people do look at this code or worse just copy and paste and quickly change and we have to set an example of best practices. When possible.

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -37,6 +37,12 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
...
+          return;
+        }
```
mgamis-msft, line 45: @prprabhu-ms Is it intended for error listener to also handler events of type `{ operation: string, error: Error }`? Was not sure how to pick out the case for screen share failures the right way

samples/Calling/src/app/views/CallScreen.tsx
prprabhu-ms: Two points here.

1. We can and should introduce an interface for errors raised via events as well as in the adapter state that has the `operation` and `error` fields. I'll do it today and send a PR.
2. You will still have to dig around inside the `error` though, because we get a opaque `Error` from the base SDK -- typescript does not have a way to strongly type raised exceptions. That said, you should (a) not have to use a private field `_code`, there should be a public field; (b) use a function to duck-type and then typecast so that the rest of the code is clean. Similar to https://github.com/Azure/communication-ui-library/blob/e337a0813532da34430795921a4bfc324453d881/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts#L524 (or atleast wrap the duck-typing logic in a helper function, e.g https://github.com/Azure/communication-ui-library/blob/e48aefca725807e58d1d2e4a6ff0121f5b09437e/packages/chat-component-bindings/src/errorBarSelector.ts#L78)

samples/Calling/src/app/views/CallScreen.tsx
prprabhu-ms: Let's discuss in the comment thread above.

samples/Calling/src/app/views/CallScreen.tsx
prprabhu-ms: Here's a PR to add strong error types to adapters: https://github.com/Azure/communication-ui-library/pull/761

This renames `operation` -> `target`. So, if we like it, lets land that PR before this one.

packages/react-composites/src/composites/CallComposite/CallScreen.tsx
```diff
@@ -99,7 +99,7 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
         callErrorHandler('removed');
       }
     }
-  }, [adapter, callErrorHandler, endedCall]);
```
prprabhu-ms, line 102: The `adapeter` isn't referenced inside this `useEffect` block. Can you help me understand how this was causing the bug?

prprabhu-ms: Oh, wait. You _removed_ the adapter from the dependencies. I still don't understand why this was causing the bug though. Why was adapter changing, and if it was, why was it triggering a page change?

prprabhu-ms: nvm, this is just clean up. The fix is in CallScreen.

Got it.

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -37,6 +37,12 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
...
         onCallError(e);
```
prprabhu-ms, line 46: Great find.

This PR will handle the bug you're fixing, but I think we might want to go further (OK in a follow up PR) -- currently, *any* error from the adapter will kick us to the error screen. I think this should be an allowlist instead -- log and squelch errors here, except for a few really bad ones (I can't think of any bad enough ones at present and we haven't seen this behaviour in all our self hosting, so I expect that there aren't actually many errors where we'd want to leave the call).

prprabhu-ms (approved):

samples/Calling/src/app/views/CallScreen.tsx
mgamis-msft: Sounds good. I am just squelching based on error target being 'Call.startScreenSharing' as I can't really use the string stack intelligently.

samples/Calling/src/app/views/CallScreen.tsx
prprabhu-ms: #761 has landed. Please update this PR to use the newly added types.

alcail (approved):

PR #761, opened 2021-09-09
Strongly type errors from adapters

prprabhu-ms:
# What

* Add types for errors emitted from adapters
* Use same types for errors stored in adapter state
* Change field name: `operation` -> `target` to align with stateful error objects.

# Why

Allows for better client code.
Using the same name makes for easy conversion to-fro stateful and adapter errors and also avoids user confusion.

# How Tested

`rush test`

  Changes error related adapter API signature

### DIFF
--- a/packages/react-composites/src/composites/common/adapters.ts
+++ b/packages/react-composites/src/composites/common/adapters.ts
@@ -27,28 +27,19 @@ export interface AdapterDisposal {
 /**
  * Error reported via error events and stored in adapter state.
  */
-export class AdapterError extends Error {
+export interface AdapterError extends Error {
   /**
    * The operation that failed.
    */
-  public target: string;
+  target: string;
   /**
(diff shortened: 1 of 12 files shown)


packages/react-composites/src/composites/common/adapters.ts
```diff
@@ -23,3 +23,28 @@ export interface AdapterPages<TPage> {
...
+  target: string;
+  /**
+   * Error thrown by the failed operation.
+   */
+  inner: Error;
+  /**
+   * Timestamp added to the error in the adapter implementation.
+   */
+  timestamp: Date;
+}
```
mgamis-msft, line 43: Nice!

mgamis-msft (approved):

prprabhu-ms:
@alkwa-msft  PTAL to unblock Miguel's PR.

alcail (approved):

PR #765, opened 2021-09-10
Feature toggle for UI elements in Call and Chat Composite

anjulgarg:
# What
POC showcasing a mechanism for feature customization in Composites

# Why
We have requirements from customers where they want to disable certain features.
For example, the ability to turn their microphones on during a call.

# How Tested

# Process & policy checklist


packages/react-composites/review/react-composites.api.md
```diff
@@ -241,8 +241,19 @@ export interface CallCompositeStrings {
 }
 
 // @public
-export type CallCompositeVisualElements = {
+export type CallCompositeVisualElements = CallControlVisualElements & {
     showErrorBar?: boolean;
+    showCallControls?: boolean;
```
alkwa-msft, line 246: what happens if they do showCallControls false but then we have explicitly set showCameraButton true?

my gut feeling is that we don't show the call controls or the camera button. My concern with this is how we can potentially put ourselves in a weird state.

I am curious if we flip all of these show____ to hide____ and then we don't have an explicit hideCallControls but rather the developer has to set true to all of the hide____ for the CallControlVisualElements

Does it make more sense to call it CallControlButtonFlags

anjulgarg: If `showCallControls` has been set to false by Contoso, it is going to hide the entire control bar (not render it entirely). So it doesn't matter if individual buttons are set to true.
So, if the entire element is hidden, there is no possibility of putting ourselves in an unexpected state since children elements won't come into play at all.

The argument between `show` prefix and `hide` prefix is subjective TBH. I remember having this discussion with Prathmesh as well earlier :D
But I understand your perspective and hide might be a better prefix for elements that are visible by default. I wonder if we can get rid of this ambiguity entirely by not using a prefix like hide or show? and just using something like `endCallButton`?

Named it `CallControlVisualElements` to be inline with `CallComponentVisualElements`. Can change it if you feel strongly about `CallControlButtonFlags`

alkwa-msft: VisualElements seems pretty ambiguous .. can we add in more visual elements or also remove? Since its just showing or hiding I am curious if we can just have a prop on the composite API that just says
hiddenElements which has a flat structure of all of the areas of the composite and their underlying components and which ones are just being hidden

e.g if we wanted to hide the CameraButton it would be

`<CallComposite hiddenElements={{CallControls: { cameraButton: true }}}/>`

prprabhu-ms: `visualElements` was the best name Anjul and I could come up with, at the time :D

I like the `hiddenElements` API better -- also because I really dislike boolean flags that default to `true` (like the ones added in this PR). It's very confusing when an undefined flag is the same as setting it to true (breaks expectations about truthyness of undefined. You can't do checks like  `if(boolean(flag)) { ... flag must be true, but what if undefined? ...})`.

alkwa-msft: @prprabhu-ms Are you suggesting that its still wierd that we are checking true if we should not be showing some UI?

prprabhu-ms: I'm saying that for any boolean field, not setting it should have the same effect as setting it to false. Thus, the default value of all optional boolean fields must be false.

packages/react-composites/src/composites/CallComposite/CallScreen.tsx
```diff
@@ -172,13 +170,13 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
               </>
             )}
           </Stack.Item>
-          {props.visualElements.showCallControls && (
+          {props?.hiddenElements?.callControls !== true && (
```
prprabhu-ms, line 173: This can just be `props?.hiddenElements?.callControls || ( ... )`. Yay for falsey default values!

prprabhu-ms (approved):

packages/react-composites/src/composites/CallComposite/CallScreen.tsx
anjulgarg: Thanks, totally forgot about that shorthand. I would prefer leaving it explicit right now since it is a little hard to read. :D

alcail (approved):

packages/react-composites/src/composites/CallComposite/CallControls.tsx
```diff
@@ -37,30 +70,40 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
...
     <ControlBar layout="dockedBottom">
-      <CameraButton data-ui-id="call-composite-camera-button" {...cameraButtonProps} showLabel={!compressedMode} />
-      <MicrophoneButton
-        data-ui-id="call-composite-microphone-button"
-        {...microphoneButtonProps}
-        showLabel={!compressedMode}
-      />
-      <ScreenShareButton {...screenShareButtonProps} showLabel={!compressedMode} />
-      {props.showParticipantsControl && (
+      {hiddenElements?.cameraButton !== true && (
```
alcail, line 73: why not just !hiddenElements?.cameraButton instead?

alcail (approved):

packages/react-composites/src/composites/CallComposite/CallControls.tsx
anjulgarg: Readability. `!h` is easy to miss.
I missed it myself a couple of times.

anjulgarg: And also because it takes some mental overhead to quickly understand double negatives .

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -29,31 +30,34 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
...
+ * Optional features of the {@link CallComposite}
  */
-export type CallCompositeVisualElements = {
+export type CallCompositeHiddenElements = CallControlHiddenElements & {
   /**
    * Surface Azure Communication Services backend errors in the UI with {@link @azure/communication-react#ErrorBar}.
+   * Hidden if set to `true`
    *
-   * @defaultValue false
+   * @default false
```
JamesBurnside, line 46: This should remain `@defaultValue` I think https://tsdoc.org/pages/tags/defaultvalue/

JamesBurnside:
image

Should this visual change have happened (no longer shows the topic header in the embedjshtml example?)

packages/react-composites/review/react-composites.api.md
```diff
@@ -226,8 +232,8 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
     adapter: CallAdapter;
     // (undocumented)
     callInvitationURL?: string;
+    hiddenElements?: CallCompositeHiddenElements;
```
JamesBurnside, line 235: With it being hiddenElements (so everything is shown by default) and not visualElements (where everything has to be set to true explicitly) -what happens if we want to add an optional visual element ?
Would showing it by default be a breaking change? (advantage of visualElements over hiddenElements was that we could add it to this and it would be hidden by default)

anjulgarg: You have a valid point. Had a discussion around this with @alkwa-msft and decided to make every UI feature visible by default going forward.

Couple of ways to solve that would be to add that new feature to hidden elements with a default `true`

@alkwa-msft any suggestions?

packages/react-composites/src/composites/MeetingComposite/SidePane.tsx
```diff
@@ -106,7 +106,11 @@ export const EmbeddedChatPane = (props: {
 }): JSX.Element => {
   return (
     <SidePane hidden={props.hidden} headingText={'Chat'} onClose={props.onClose}>
-      <ChatComposite adapter={props.chatAdapter} fluentTheme={props.fluentTheme} />
+      <ChatComposite
+        adapter={props.chatAdapter}
+        fluentTheme={props.fluentTheme}
+        hiddenElements={{ participantPane: true }}
```
JamesBurnside, line 112: Meeting composite should also hide the topic header

JamesBurnside:
I don't think the typing indicator should be changed in this PR:
image
(might be related to the main pipeline failing on UI tests)

packages/react-composites/review/react-composites.api.md
JamesBurnside: The defaulting to true takes away from @prprabhu-ms's main point that `undefined === false` + @prprabhu-ms for ideas here

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
anjulgarg: Oops. Will change it all

JamesBurnside (changes requested):
Small change to ensure meeting composite is not showing chat topic and to evaluate new UI test images the CI introduced

JamesBurnside (approved):

PR #767, opened 2021-09-11
Alcail/a11y bugs on ParticipantList being inaccessible through keyboard

alcail:
# What
made ParticipantItem focusable
changed 'More Options' to be always displayed if there are some items there, so behaviour is easier for accessibility purpose
updated styles
image
image

# Why
ParticipantList and thus ParticipantItems were inaccessible through keyboard navigation, so user could not read the list of participants and thus access participant info or options
bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2562125

# How Tested
ran storybook and calling sample locally

# Process & policy checklist


packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -77,14 +73,15 @@ export interface ParticipantItemProps {
   strings?: Partial<ParticipantItemStrings>;
 }
 
+const iconsContainerDefaultWidth = '5rem';
```
alkwa-msft, line 76: can we not add this as a style in iconContainerStyle?

alcail: sure :)

packages/react-components/src/components/styles/ParticipantItem.styles.ts
```diff
@@ -14,17 +14,17 @@ export const participantItemContainerStyle = mergeStyles({
...
   alignItems: 'center',
-  height: '100%',
-  overflow: 'hidden'
+  height: '100%'
 });
 
-export const iconStyles = mergeStyles({ display: 'flex', margin: 'auto' });
+export const iconStyles = mergeStyles({
+  display: 'flex',
+  lineHeight: 0,
```
JamesBurnside, line 28: what does line height 0 do here? put a quick comment here to ensure this isn't removed in the future and the bug re-introduced
```suggestion
  lineHeight: 0, // ensure the icon ...
```

JamesBurnside (approved):

packages/react-components/src/components/styles/ParticipantItem.styles.ts
alcail: with line-height at 0
image

without line-height at 0, the "More" icon is slightly above center line
image

packages/react-components/src/components/styles/ParticipantItem.styles.ts
```diff
@@ -14,17 +14,18 @@ export const participantItemContainerStyle = mergeStyles({
 });
 
 export const menuButtonContainerStyle = mergeStyles({
-  root: {
-    '&:hover': { background: 'none' },
-    '&:active': { background: 'none' }
-  }
+  paddingTop: '0.25rem'
 });
```
mgamis-msft, line 18: I don't think we use this style anymore so we can delete this. Other than that LGTM. Thanks for awesome refactor of ParticipantItem. Looks much simpler now 👍

mgamis-msft (approved):

PR #773, opened 2021-09-13
Fix storybook package bump

prprabhu-ms:
# What
* Change type of some storybook change files to 'prerelease' from 'patch'. This was causing storybook to incorrectly upgrade storybook to `1.0.0`, i.e. off of the prerelease version scheme.
* Centralize beachball config so that only `prelease` and `none` types are allowed for all packages (including storybook)
* Fix a bug in the beachball config so that it actually considers all the packlets as a group
* Disable `bumpDeps` because it causes infinite recursion in beachball when used with groups.

# Why

* So that we can correctly version storybook with our releases.

# How Tested

Locally bump packages the way creating a release branch would.


common/config/beachball/changelog-config.ts
```diff
@@ -2,20 +2,30 @@ import { BeachballConfig } from 'beachball';
...
 export const config: BeachballConfig = {
-  disallowedChangeTypes: ['major'],
   branch: 'origin/main',
   changelog: {
     customRenderers: {
       renderHeader,
       renderEntry
     }
   },
+  tag: "next",
```
JamesBurnside, line 12: do we need tag?

JamesBurnside: oh right we had tag in the package.jsons - nvr mnd

JamesBurnside (approved):

alcail (approved):

PR #776, opened 2021-09-14
Adjust api comments - iteration 2

PorterNan:
# What

Adjust api comments - iteration 2
# Why

# How Tested

# Process & policy checklist


packages/calling-component-bindings/review/calling-component-bindings.api.md
```diff
@@ -138,24 +138,7 @@ export const cameraButtonSelector: reselect.OutputParametricSelector<CallClientS
...
-    onToggleMicrophone: () => Promise<void>;
-    onToggleScreenShare: () => Promise<void>;
-    onCreateLocalStreamView: (options?: VideoStreamOptions | undefined) => Promise<void>;
-    onCreateRemoteStreamView: (userId: string, options?: VideoStreamOptions) => Promise<void>;
-    onParticipantRemove: (userId: string) => Promise<void>;
-    onStartLocalVideo: () => Promise<void>;
-    onDisposeRemoteStreamView: (userId: string) => Promise<void>;
-    onDisposeLocalStreamView: () => Promise<void>;
-};
+export const createDefaultCallingHandlers: (callClient: StatefulCallClient, callAgent: CallAgent | undefined, deviceManager: StatefulDeviceManager | undefined, call: Call | undefined) => CallingHandlers;
```
JamesBurnside, line 141: 😍

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -516,12 +516,20 @@ export const createAzureCommunicationCallAdapter = async ({
...
 }: AzureCommunicationCallAdapterArgs): Promise<CallAdapter> => {
   const callClient = createStatefulCallClient({ userId });
-  const deviceManager = (await callClient.getDeviceManager()) as StatefulDeviceManager;
   const callAgent = await callClient.createCallAgent(credential, { displayName });
-  const adapter = new AzureCommunicationCallAdapter(callClient, locator, callAgent, deviceManager);
+  const adapter = createAzureCommunicationCallAdapterFromClient(callClient, callAgent, locator);
   return adapter;
 };
 
+export const createAzureCommunicationCallAdapterFromClient = async (
```
JamesBurnside, line 524: (non PR blocking comment) Do we need/want storybook documentation for this?

JamesBurnside (approved):

packages/communication-react/review/communication-react.api.md
```diff
@@ -606,7 +606,7 @@ export type ChatCompositeIcons = Pick<CompositeIcons, 'MessageDelivered' | 'Mess
 // @public (undocumented)
 export interface ChatCompositeProps extends BaseCompositeProps<ChatCompositeIcons> {
     adapter: ChatAdapter;
-    onRenderMessage?: (messageProps: MessageProps, defaultOnRender?: DefaultMessageRendererType) => JSX.Element;
+    onRenderMessage?: (messageProps: MessageProps, defaultOnRender?: MessageRenderer) => JSX.Element;
```
alkwa-msft, line 609: we defaultOnRenderer can we just call the parameter messageRenderer? 

our functions should still have a parameter that takes in an optional message renderer and they could pass in the default one if they want to but i dont think we should use the word default in the paramter?

alkwa-msft (approved):

alcail (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: Yea, we should have a work item for this, but not necessary for GA or azure review

packages/communication-react/review/communication-react.api.md
PorterNan: This function is a callback - so the parameter here is what we pass to developers - like message data + the default renderer for message data, return value is what developers wanna pass back to us

So defaultMessageRender is the right name cause it is passing our renderer for developers to consume

PR #780, opened 2021-09-15
Fix ControlBarButton labels

mgamis-msft:
# What

- Changed ControlButton default labels to use `Text` component from FluentUI instead of `Label` because `Label` does not seem to respond to changes to `theme.defaultFontStyle` as tested on CodePen below:
image
image
- ControlBarButton labels have regular font weight by refactoring `onRenderText` value. (see comments in changes)

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2564946

# How Tested

Tested on storybook by adding the following to the light theme:
image
and confirmed the following changes:
image
Not sure if the EndCallButton font should also be included

# Process & policy checklist


packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -55,16 +63,6 @@ export interface ControlBarButtonProps extends IButtonProps {
   onRenderOffIcon?: IRenderFunction<IButtonProps>;
 }
 
-const DefaultRenderText = (props?: ControlBarButtonProps): JSX.Element => {
-  const labelText =
-    props?.text ?? props?.strings?.label ?? (props?.checked ? props?.strings?.onLabel : props?.strings?.offLabel);
-  return (
-    <Label key={props?.labelKey} className={mergeStyles(controlButtonLabelStyles, props?.styles?.label)}>
```
mgamis-msft, line 62: I had to refactor this to the callback on line 82 because for some reason `props?.strings?.label` has some default styling including a font weight of 600 as shown below.
image
image

PR #781, opened 2021-09-15
Do not automatically clear errors from client state 

prprabhu-ms:
# What

Stop clearing latest error from client state on successful invocation of related method calls. Auto-clearing is difficult to get right (concurrent method calls / failure to send one message vs success in sending an unrelated message). Also, this is not longer needed because `ErrorBar` internally stores dismissed messages.


JamesBurnside (approved):

alcail (approved):

prprabhu-ms:
This is the KI about automation test flakiness. Except I think that the committed snapshots are the ones that happen when we snapshot too soon. So, adding label to update snapshots.

PR #783, opened 2021-09-15
Fix meeting trying to access the chat before user has joined the meeting

JamesBurnside:
# What
Ensure the chat composite is only constructed when the user has joined the call

This regression happened as part of a previous fix that set the chat pane to be hidden instead of destroyed to retain partially composed chat messages when closing and reopening the chat pane.

# Why
Meeting composite was trying to access the chat in the pre join screen:
image

BONUS: This also fixes the issue where the chat pane was still showing when the user leaves the call

# How Tested
Locally in storybook - no longer throws 403 error in pre join screen for a teams meeting.


packages/react-composites/src/composites/MeetingComposite/ChatButton.tsx
```diff
@@ -14,6 +14,7 @@ export const ChatButton = (props: ControlBarButtonProps): JSX.Element => {
   return (
     <ControlBarButton
       {...props}
+      labelKey={'chatButtonLabelKey'}
```
JamesBurnside, line 17: Quickly fixed this also:
image

packages/react-composites/src/composites/MeetingComposite/PeopleButton.tsx
```diff
@@ -15,6 +15,7 @@ export const PeopleButton = (props: ControlBarButtonProps): JSX.Element => {
     <ControlBarButton
       {...props}
       strings={strings}
+      labelKey={'peopleButtonLabelKey'}
```
JamesBurnside, line 18: Quickly fixed this also:
image

prprabhu-ms (approved):

alcail (approved):

PR #785, opened 2021-09-15
Temporarily add a delay to reduce CI flakiness

prprabhu-ms:
# What

Add a delay in CallComposite e2e tests.

# Why

Alas, life is hard.


PorterNan (approved):
Good one, from this one we can know whether if this is because timing issue or other issues - if it is still failing 3 times a lot, then we should look into this problem again

alkwa-msft (approved):

PR #786, opened 2021-09-15
Create initial meeting composite ui test + cleanup

JamesBurnside:
# What
Setup code for writing meeting composite ui tests
* Create a single `Page Load` test for meeting composite
* More tests to be added in an individual PR

Cleanup inside test folder
* Create common folder and server folder to have common files

# Why
Setup for UI regression tests for meeting composite

# How Tested
Ran locally with `node ./serve` and ran ui snapshot generation locally also. Will update the actual snapshote with `ui change` in the PR


alcail (approved):

packages/react-composites/tests/browser/meeting/MeetingComposite.test.ts
```diff
@@ -0,0 +1,28 @@
...
+import { expect, Page } from '@playwright/test';
+
+export const waitForMeetingCompositeToLoad = async (page: Page): Promise<void> => {
+  await page.waitForLoadState('load');
+
+  // @TODO
+  // We wait 3 sec here to work around flakiness due to timing.
+  // It sometimes take a while for the local video / audio streams to load in CI environments.
+  // We don't have a good way to know when the composite is fully loaded.
+  await page.waitForTimeout(3000);
```
anjulgarg, line 15: Can the fix you added to call composites be applied here as well?
That waits for video state

packages/react-composites/tests/browser/meeting/MeetingComposite.test.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { dataUiId } from '../common/utils';
+import { test } from './fixture';
+import { expect, Page } from '@playwright/test';
+
+export const waitForMeetingCompositeToLoad = async (page: Page): Promise<void> => {
+  await page.waitForLoadState('load');
+
+  // @TODO
+  // We wait 3 sec here to work around flakiness due to timing.
```
anjulgarg, line 12: I don't think we need a timeout for call composite to load. It's for chat due to this bug
image

Could we update the comments to reflect the proper reason?
If that's not the case, please do correct me.

JamesBurnside: Sigh, divergence from main again, this isn't need anymore since the fix to wait for video ready, I will remove

anjulgarg (approved):

packages/react-composites/tests/browser/meeting/MeetingComposite.test.ts
anjulgarg: Does this fix the chat bug as well?

JamesBurnside: This just loads the configuration screen (chat hasn't loaded at this point). But haven't seen any issues/failures yet with test for checking the chat messaging inside the composite

PR #789, opened 2021-09-16
Fix ControlBarButton label

mgamis-msft:
# What

- Changed ControlButton default labels to use `Text` component from FluentUI instead of `Label` because `Label` does not seem to respond to changes to `theme.defaultFontStyle` as tested on CodePen below:
image
image
- ControlBarButton labels have regular font weight by refactoring `onRenderText` value. (see comments in changes)

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2564946

# How Tested

Tested on storybook by adding the following to the light theme:
image
and confirmed the following changes:
image
Not sure if the EndCallButton font should also be included

# Process & policy checklist


alcail (approved):

change/@internal-react-composites-7c1614bf-ee32-4f62-bc78-2ba07a5aa3a9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "this change does not affect the published package in any way.",
```
JamesBurnside, line 3: This actually does impact the composites:
image

https://github.com/Azure/communication-ui-library/pull/789/files?short_path=710bcfd#diff-710bcfd462fa4c1aa7e00a93f85f0320b9b9761ad40ea523fcd25ae90fd891ec

Is the difference padding intentional / meet figma specs?

mgamis-msft: Thanks for catching that. I have added margins to fix this. But now the font weight is no longer 600 and the icon and label are tighter together which is closer to the Figma spec.

JamesBurnside (approved):

JamesBurnside:
Just to check also, there's no impact to screen reader/narrator by changing from label to span+text?

PR #790, opened 2021-09-16
[PoC] Alkwa/rem to em

alkwa-msft:
# What
Convert all uses of rem to em in the UI Library

# Why
We were using rems however we don't want to override any em usage used by an app that is using the UI Library components/composites

# How Tested

# Process & policy checklist


alcail:
Did you test all this?
Conversion from REM to EM is not that simple I think, as EM will depend on its parents font size, which can be 16px or any other value we (or FluentUI) override

prprabhu-ms:
I'm quite sure this PR can't be merged in the current form. Shall we close it for now?

alcail:
> I'm quite sure this PR can't be merged in the current form. Shall we close it for now?

sure. We need to revisit this switching from REM to EM

PR #793, opened 2021-09-16
Changed ControlBar to be affected by `theme.effects`

mgamis-msft:
# What

Stack in ControlBar has its borderRadius and boxShadow assigned based on the `layout` prop so FluentUI `theme.effects` won't take effect. So I changed it so that styling from FluentUI `theme.effects` wins out.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2564949

# How Tested

Tested on storybook by changing `theme.effects.roundedCorner6` and `theme.effects.elevation8` like so:
image
and verified the changes to be:
image

# Process & policy checklist


packages/react-components/src/components/ControlBar.tsx
```diff
@@ -70,7 +70,14 @@ export const ControlBar = (props: ControlBarProps): JSX.Element => {
   }, [layout, styles?.root, theme]);
   return (
     <div className={mergeStyles(mainDivStyle)}>
-      <Stack className={controlBarClassName}>{props.children}</Stack>
+      <Stack
+        className={mergeStyles(controlBarClassName, {
+          borderRadius: (theme as PartialTheme)?.effects?.roundedCorner6,
+          boxShadow: (theme as PartialTheme)?.effects?.elevation8
```
mgamis-msft, line 76: Do we expect Contoso to set `themes.effects.roundedCorner6` to change corners for surfaces as documented by [FluentUI](https://github.com/microsoft/fluentui/blob/dab45a2afe609709735b9b6e604a1af40d50e809/packages/theme/src/types/IEffects.ts#L52) and similarly `themes.effects.elevation8` to change elevation for command tools

packages/react-components/src/components/ControlBar.tsx
```diff
@@ -70,7 +70,14 @@ export const ControlBar = (props: ControlBarProps): JSX.Element => {
   }, [layout, styles?.root, theme]);
   return (
     <div className={mergeStyles(mainDivStyle)}>
-      <Stack className={controlBarClassName}>{props.children}</Stack>
+      <Stack
+        className={mergeStyles(controlBarClassName, {
+          borderRadius: (theme as PartialTheme)?.effects?.roundedCorner6,
```
JamesBurnside, line 75: Can these be moved into the controlBarClass in the `.styles` file?

prprabhu-ms: Pulling things out of the theme outside the component is a bit complex. You need to use `useTheme`, but you can only do that from another `useX` function.

There is a supported way for this via registered names, but I don't think we're using that uniformly (yet).

So, soft suggestion here. Do it only if it's easy.

prprabhu-ms (approved):

packages/react-components/src/components/ControlBar.tsx
JamesBurnside: We do it elsewhere with `getTheme`. But now I'm looking at it I'm not convinced this will update when the theme is updated:

https://github.com/Azure/communication-ui-library/blob/a205caeb26065412fff2e4c38b1aaeac76d6f921/packages/react-components/src/components/styles/VideoGallery.styles.ts#L14

image

JamesBurnside (approved):

packages/react-components/src/components/ControlBar.tsx
mgamis-msft: Yes, I can't move the merging of styles to a separate styles file because `getTheme` does not get the theme context from the `ThemeProvider`

JamesBurnside: Oh good to know! I've created a bug to remove uses of getTheme: https://skype.visualstudio.com/SPOOL/_workitems/edit/2591054

PR #795, opened 2021-09-16
POC - custom participant list menu items

anjulgarg:
# What
Allows devs to inject custom menu items to participant list in call and chat composite.

# Why
Part of the user story https://skype.visualstudio.com/SPOOL/_workitems/edit/2487948

# How Tested

# Process & policy checklist


packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -48,13 +76,9 @@ const onRenderParticipantDefault = (
...
-  const menuItems: IContextualMenuItem[] = [];
-  if (participant.userId !== myUserId && onParticipantRemove) {
-    menuItems.push({
-      key: 'Remove',
-      text: 'Remove',
-      onClick: () => onParticipantRemove(participant.userId)
-    });
+  let defaultMenuItems = defaultOnFetchParticipantMenuItems(participant.userId, myUserId, onParticipantRemove);
+  if (onFetchParticipantMenuItems) {
+    defaultMenuItems = onFetchParticipantMenuItems(participant.userId, myUserId, defaultMenuItems);
```
alcail, line 81: nit: we can't say it's the default MenuItem list if we are fetching from Contoso customization

packages/react-composites/src/composites/CallComposite/CallControls.tsx
```diff
@@ -1,61 +1,68 @@
...
   onEndCallClick(): void;
-  compressedMode?: boolean;
   callInvitationURL?: string;
-  hiddenElements?: CallControlHiddenElements;
+  onFetchParticipantMenuItems?: ParticipantMenuItemsCallback;
+  options?: boolean | CallControlOptions;
 };
 
-export type CallControlHiddenElements = {
+export type CallControlOptions = {
```
alcail, line 25: nit: if we go back to naming it 'Options', we should rename the boolean for buttons with 'hide' to be clear on what they do

alcail (approved):

change/@internal-react-components-b7490699-acd9-4bb3-91a1-86788782aec9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Update Feature Toggle API",
```
JamesBurnside, line 3: nit: for the changelog this doesn't explain what the change is as is likely won't be mentioned in the release notes - maybe `Allow custom menu items in ParticipantList`

packages/react-composites/src/composites/CallComposite/CallControls.tsx
anjulgarg: We can't, since the proposal is to use a shorthand approach. Which means that `CallControlOptions.endCallButton` can have these possible value
`true`
`false`
`{  color: something, position: somthing }` etc etc

Basically, an option can be true, where it will be rendered with default options. Or users can pass in custom options instead of a boolean.

change/@internal-react-components-b7490699-acd9-4bb3-91a1-86788782aec9.json
anjulgarg: Thanks. I recently acquired a higher appreciation for better changelog after I did the release on Monday. I completely agree with this...

packages/react-components/src/components/ParticipantList.tsx
anjulgarg: would changing the name to `menuItems` make it more clear?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -429,7 +429,7 @@ export type MessageThreadProps = {
    * Optional callback to override render of a message.
    *
    * @param messageProps - props of type MessageProps
-   * @param defaultOnRender - default render of type DefaultMessageRendererType
+   * @param defaultOnRender - default render of type MessageRenderer
```
JamesBurnside, line 432: nit: we should use `@link` more for referencing other parts of packages - then we can have api-extractor break when links to do not resolve correctly https://api-extractor.com/pages/messages/ae-unresolved-link/ so renames are caught

packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -28,13 +37,32 @@ export type ParticipantListProps = {
...
+const defaultOnFetchParticipantMenuItems = (
+  participantUserId: string,
+  myUserId?: string,
+  onParticipantRemove?: (userId: string) => void
+): IContextualMenuItem[] => {
+  const menuItems: IContextualMenuItem[] = [];
+  if (participantUserId !== myUserId && onParticipantRemove) {
+    menuItems.push({
+      key: 'Remove',
+      text: 'Remove',
```
JamesBurnside, line 53: Should this be a component string that is localized?

packages/react-components/src/components/ParticipantList.tsx
JamesBurnside: I like just `menuItems`

packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -93,7 +117,7 @@ const getParticipantsForDefaultRender = (
   participants: CommunicationParticipant[],
   excludeMe: boolean,
   myUserId: string | undefined
-) => {
+): CommunicationParticipant[] | [] => {
```
JamesBurnside, line 120: Does this need the `| []`? better to be strongly typed if possible. I see there's a 
```
if (participants === undefined) {
  return [];
}
```
But participants can't be undefined as a parameter so that should be able to be removed or maybe updated to:
```
if (participants.length === 0) {
  return [];
}
```
If that is what the check is meant to be for

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -23,45 +23,44 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
...
   adapter: CallAdapter;
   callInvitationURL?: string;
-
   /**
    * A callback function that can be used to provide custom data to an Avatar.
    */
   onFetchAvatarPersonaData?: AvatarPersonaDataCallback;
-
   /**
    * Flags to hide UI elements of the {@link CallComposite}.
```
JamesBurnside, line 31: Is `CallCompositeOptions` only for hiding/showing UI elements, for example could we have feature flags here also? (e.g. `newAwesomeLowerLatencyStreams: boolean`)

change/@internal-react-composites-93f823de-fb23-4096-aa20-ffb9254e29e3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Update Feature Toggle API",
```
JamesBurnside, line 3: Would be good to note here the breaking change that developers need to flip from hiddenElements to Options

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -61,18 +64,18 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
+    (userId, defaultOptions) => {
+      return <AvatarPersona userId={userId} {...defaultOptions} dataProvider={onFetchAvatarPersonaData} />;
     },
     [onFetchAvatarPersonaData]
   );
 
   return (
     <Stack className={chatContainer} grow>
-      {hiddenElements?.topic !== true && <ChatHeader {...headerProps} />}
+      {options?.topic !== false && <ChatHeader {...headerProps} />}
```
JamesBurnside, line 75: these are now double negative checks, can they be simplified to 
```suggestion
      {options?.topic && <ChatHeader {...headerProps} />}
```

packages/storybook/stories/CallComposite/snippets/CustomDataModelExampleContainer.snippet.tsx
```diff
@@ -69,13 +70,29 @@ export const CustomDataModelExampleContainer = (props: ContainerProps): JSX.Elem
       });
     });
 
+  // Custom Menu Item Callback for Participant List
+  const onFetchParticipantMenuItems: ParticipantMenuItemsCallback = (participantId, userId, defaultMenuItems) => {
+    console.log('Remote Participant', participantId);
```
JamesBurnside, line 75: nit: remove console logs

packages/storybook/stories/CallComposite/snippets/CustomDataModelExampleContainer.snippet.tsx
```diff
@@ -69,13 +70,29 @@ export const CustomDataModelExampleContainer = (props: ContainerProps): JSX.Elem
       });
     });
 
+  // Custom Menu Item Callback for Participant List
+  const onFetchParticipantMenuItems: ParticipantMenuItemsCallback = (participantId, userId, defaultMenuItems) => {
```
JamesBurnside, line 74: update params to be typed here
```suggestion
  const onFetchParticipantMenuItems: ParticipantMenuItemsCallback = (participantId: string, userId: string, defaultMenuItems: something[]) => {
```

packages/storybook/stories/ChatComposite/snippets/CustomDataModelExampleContainer.snippet.tsx
```diff
@@ -65,13 +66,29 @@ export const CustomDataModelExampleContainer = (props: CustomDataModelExampleCon
       }
     });
 
+  // Custom Menu Item Callback for Participant List
+  const onFetchParticipantMenuItems: ParticipantMenuItemsCallback = (participantId, userId, defaultMenuItems) => {
```
JamesBurnside, line 70: nit: same here on param types and console logs

packages/react-components/src/components/ParticipantList.tsx
prprabhu-ms: Yeah, please tighten this type.

I think you might just need to change line 122 to be `return [] as CommunicationParticipant[]` so typescript knows to type the empty list correctly.

packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -28,13 +37,32 @@ export type ParticipantListProps = {
   onRenderAvatar?: OnRenderAvatarCallback;
   /** Optional callback to render the context menu for each participant  */
   onParticipantRemove?: (userId: string) => void;
+  /** Optional callback to render custom menu items for each participant. */
+  onFetchParticipantMenuItems?: ParticipantMenuItemsCallback;
```
prprabhu-ms, line 41: I completely agree that `ParticipantMenuItemsCallback` is a good API at the composite layer.

Here though, it's a bit confusing.
 * Contoso can provide `onFetchParticipantMenuItems`, it will be used with the default implementation of `onRenderParticipant`.
 * Contoso can provide `onRenderParticipant`. In that case, `onFetchParticipantMenuItems` will be ignored.

 Same with other callbacks here. If `onRenderParticipant` is provided, `onRenderAvatar`, `onParticipantRemove` and `onFetchParticipantMenuItems` are all ignored. This is not obvious without looking at the implementation.

 Big question: Can we drop `onRenderParticipant` now that we have more targeted ways to customize this component?

anjulgarg: Valid concern there. This is going to be confusing only in components since we are not exposing onRenderParticipant via Composites.
To make it less confusing, we can improve the doc blocks.
TBH, I see this problem happening at multiple places throughout our components.
To handle this particular case, we could try and use the existing `onRenderParticipant` prop and pass it's attributes as defaults to the callback, but that will make the implementation really complex.

packages/communication-react/review/communication-react.api.md
```diff
@@ -337,8 +338,8 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
     adapter: CallAdapter;
     // (undocumented)
     callInvitationURL?: string;
-    hiddenElements?: CallCompositeHiddenElements;
     onFetchAvatarPersonaData?: AvatarPersonaDataCallback;
+    options?: CallCompositeOptions;
```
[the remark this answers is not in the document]
anjulgarg: Any particular reason?

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
anjulgarg: This won't work if topic is undefined or null

packages/storybook/stories/CallComposite/snippets/CustomDataModelExampleContainer.snippet.tsx
anjulgarg: Seems redundant given that we have already provided a type to `onFetchParticipantMenuItems`

packages/storybook/stories/ChatComposite/snippets/CustomDataModelExampleContainer.snippet.tsx
anjulgarg: Seems redundant given that we have already provided a type to onFetchParticipantMenuItems

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
JamesBurnside: Ah yeah you're right because if its undefined or null we want to show the ChatHeader - thanks !

packages/react-components/src/components/ParticipantList.tsx
anjulgarg: I am not in favor of removing an `onRender` function as it takes away the ability of a dev to use a custom component. And there isn't any other way to do that for participantList

packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -119,19 +121,52 @@ const getParticipantsForDefaultRender = (
...
+    myUserId,
+    participants,
+    onParticipantRemove,
+    onRenderAvatar,
+    onRenderParticipant,
+    onFetchParticipantMenuItems
+  } = props;
+
   const ids = useIdentifiers();
+  const locale = useLocale().strings.participantItem;
```
mgamis-msft, line 135: Nit. Would change this variable name to `const strings = useLocale().strings.participantItem;`. But otherwise, localization is good.

mgamis-msft (approved):

PR #734, opened 2021-08-29
Jinan/message editing

PorterNan:
# What

- Adjust Css for editing box
- Add handlers for edit message
- Update message style to flex, make it easier to caculate parent width
- Separate code of chatMessage
- Reuse logic of input box
- Adjust rtl use case
Some screenshot for edit/delete:

Context menu:
image

Editing box:
image

Responsive (expand and shrink with chat box):
responsive

### Overlay of sendbox when text is too long:
Before:
image

After:
image

### Mutiple line alignment: 
Before:
image

After
image

# Why

# How Tested

# Process & policy checklist

- [ ] I have updated the project documentation to reflect my changes if necessary.
- [ ] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.

**Is this a breaking change?**

- [ ] This change causes current functionality to break.

PR #799, opened 2021-09-17
Fix chat messages all being bold

JamesBurnside:
# What
Remove font weight override being applied to chat messages (and remove unused class)

# Why
All chat message text after #734 were bold.

# How Tested
| before | after |
| -- | -- |
| image | image |


packages/react-components/src/components/styles/ChatMessageComponent.styles.ts
```diff
@@ -3,27 +3,6 @@
 
 import { mergeStyles, IStyle, FontWeights } from '@fluentui/react';
 
-export const chatMessageStyle = mergeStyles({
-  fontWeight: '600'
-});
-
-export const chatActionStyle = mergeStyles({
```
JamesBurnside, line 10: this class is not used anywhere

prprabhu-ms (approved):

packages/react-components/src/components/styles/ChatMessageComponent.styles.ts
alkwa-msft: lets remove it and just get it in.

alkwa-msft (approved):

packages/react-components/src/components/styles/ChatMessageComponent.styles.ts
```diff
@@ -3,27 +3,6 @@
 
 import { mergeStyles, IStyle, FontWeights } from '@fluentui/react';
 
-export const chatMessageStyle = mergeStyles({
```
PorterNan, line 6: Good, this is just a copy&paste part from legacy code when I wrote this, don't know what does this for, good to see we remove it

PR #801, opened 2021-09-17
Try fix flakey video rendering results - wait for video ready state

JamesBurnside:
# What
Wait for video ready state when testing if videos have loaded

Also:
* Add awaits to all page.bringtofront and page clicks that were missing them
* Remove retries to highlight these issues quicker

# Why
We have incredibly flakey ui tests currently

**Note**: There is still flakiness where isSpeaking border still appears - despite our tests not sending any sound. This could be a logic or SDK bug 

# How Tested
Ran many times locally - before ~40% would fail. Now 100% pass locally (from ~10-20 local runs)


anjulgarg (approved):

anjulgarg:
Thanks for this @JamesBurnside 
If this works, its going to be such a great tweak.

alkwa-msft (approved):

PR #806, opened 2021-09-20
Updated styles to ensure zooming in at 400%

alcail:
# What
Updated Calling samples styles for each page (Home, Configuration, Call, End) to ensure all components are visible or accessible through keyboard when zooming in at 400% or more

# Why
a11y bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2560994

# How Tested
ran Calling Sample and zoomed in/out + resized window

Home screen at 400% on small window (before and after fix)
image

Configuration screen
image

Call screen
image
image

End call screen
image

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/CallScreen.tsx
```diff
@@ -127,7 +126,7 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     <Stack horizontalAlign="center" verticalAlign="center" styles={containerStyles} grow>
       {isInCall(callStatus ?? 'None') ? (
         <>
-          <Stack styles={bannersContainerStyles}>
+          <Stack.Item styles={bannersContainerStyles}>
             <Stack.Item>
```
JamesBurnside, line 130: Can a Stack.item be a child of a Stack.item here? (this may need updated to a Stack)

JamesBurnside (approved):
This is great work!

JamesBurnside:
Should we add a dimension to the test matrix with zoom enabled in browser tests? 100%, 200% and 400%?
(Currently I foresee `browserType x isMobile x zoomLevel x noOfTests` so something like `3 x 2 x 3 x n` which could grow rapidly 👀👀 - but maybe we add now and optimize later?)

alkwa-msft (approved):

alcail:
> atrix with zoom enabled in browser tests

@JamesBurnside, having at least 100% and 400% in test would be great indeed

JamesBurnside:
> > atrix with zoom enabled in browser tests
> 
> @JamesBurnside, having at least 100% and 400% in test would be great indeed

Seems this isn't natively supported in playwright: https://github.com/microsoft/playwright/issues/2497
As a workaround we could use zoom css property: https://developer.mozilla.org/en-US/docs/Web/CSS/zoom however it does not work in Firefox and isn't actually a standard css property (just one most browsers have chosen to implement so it may or may not work)

PR #809, opened 2021-09-21
Edit/Delete message localization

PorterNan:
# What

Localization for edit + delete messages:
- Add strings into sendbox and messageThread
- Remove some unnecessary variables
# Why

# How Tested

# Process & policy checklist


packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -60,7 +61,10 @@
     "friday": "Friday",
     "saturday": "Saturday",
     "participantJoined": "joined the chat.",
-    "participantLeft": "left the chat."
+    "participantLeft": "left the chat.",
+    "editMessage": "Edit",
+    "removeMessage": "removeMessage",
```
JamesBurnside, line 66: should the right hand side here be camel case or something like `remove message` or `delete message`?

JamesBurnside (approved):

packages/react-components/src/localization/locales/en-US/strings.json
PorterNan: What... I think that was a copy&paste problem, thanks for catching it!

change/@internal-react-components-db951d2a-5d93-4313-be20-e2336392957b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Localization for edit + delete messages: Add strings into sendbox and messageThread Remove some unnecessary variables",
```
prprabhu-ms, line 3: A bit wordy here... :)

packages/react-components/src/components/ChatMessageComponent.tsx
```diff
@@ -86,11 +87,10 @@ const GenerateTextMessageContent = (payload: ChatMessagePayload): JSX.Element =>
 };
 
 export const ChatMessageComponent = (props: ChatMessageProps): JSX.Element => {
-  const strings = useLocale().strings.messageThread;
   const ids = useIdentifiers();
   const theme = useTheme();
 
-  const { message, onUpdateMessage, onDeleteMessage, editDisabled, showDate, messageContainerStyle } = props;
+  const { message, onUpdateMessage, onDeleteMessage, editDisabled, showDate, messageContainerStyle, strings } = props;
```
prprabhu-ms, line 93: The `strings` field should always be optional and the implementation should fallback to the strings from locale. See `ParticipantItem` for an examples.

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -163,6 +163,12 @@ export interface MessageThreadStrings {
   participantJoined: string;
   /** String for participants left */
   participantLeft: string;
+  /** String for edited tag of a message */
```
prprabhu-ms, line 166: nit: Could improve the description here "Tag shown on a message that has been edited".

packages/react-components/src/components/SendBox.tsx
```diff
@@ -33,6 +32,10 @@ export interface SendBoxStrings {
    * Placeholder text in SendBox when there is no user input
    */
   placeholderText: string;
+  /**
+   * The warning message when send box text length is more than max limit
+   */
+  textExceedsLimit: string;
```
prprabhu-ms, line 38: nit: `textTooLong` might be a slightly better name here? Could it be confusing what limit you're talking about? Or is character length the only possible limit here?

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -60,7 +61,10 @@
     "friday": "Friday",
     "saturday": "Saturday",
     "participantJoined": "joined the chat.",
-    "participantLeft": "left the chat."
+    "participantLeft": "left the chat.",
```
prprabhu-ms, line 64: nit: Perhaps a separate PR: These should be sorted.

prprabhu-ms (approved):
LGTM, pending fallback to locale strings.

alcail (approved):

alkwa-msft (approved):

packages/react-components/src/components/ChatMessageComponent.tsx
PorterNan: This is not a component we are going to expose, this component is just for code/logic separation, so it is always asking for strings from messageThread component

PR #812, opened 2021-09-22
Browser localization tests

mgamis-msft:
# What

Simple browser localization tests for calling and chat composites to sanity check that the `LocalizationProvider` is doing what is intended internally

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2484984

# How Tested

Running the browser tests

# Process & policy checklist


JamesBurnside (approved):

JamesBurnside:
Make sure to add `ui change` to the PR so the snapshots are updated

packages/react-composites/tests/browser/call/Localization.test.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { test } from './fixture';
+import { dataUiId, loadCallCompositePage, waitForCallCompositeToLoad } from '../utils';
+import { expect } from '@playwright/test';
+
+test.describe('Localization tests', async () => {
```
JamesBurnside, line 8: Open question - these tests don't perform any visual snapshot; is this adding more value than the jest tests we have for localization already?
Instead of pulling the text from the html should these be a visual snapshot instead?

packages/react-composites/src/composites/CallComposite/ConfigurationScreen.tsx
```diff
@@ -37,7 +37,9 @@ export const ConfigurationScreen = (props: ConfigurationScreenProps): JSX.Elemen
       <Stack horizontal wrap horizontalAlign="center" verticalAlign="center" tokens={configurationStackTokens}>
         <LocalPreview />
         <Stack>
-          <div className={titleContainerStyle}>{title}</div>
+          <div data-ui-id="call-composite-configuration-page-title" className={titleContainerStyle}>
```
prprabhu-ms, line 40: This should be injected via `useIdentifiers()`

prprabhu-ms: The main reason for not inlining this is that it becomes part of the public API. If it's there, Contoso can use it to select the element, and then we have to keep it forever.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -93,7 +93,9 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
         {options?.participantPane !== false && (
           <Stack.Item className={participantListWrapper}>
             <Stack className={participantListStack}>
-              <Stack.Item className={listHeader}>{chatListHeader}</Stack.Item>
+              <Stack.Item data-ui-id="chat-composite-participant-list-header" className={listHeader}>
```
prprabhu-ms, line 96: Ditto, use `useIdentifiers()`

packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -16,6 +17,7 @@ const displayName = params.displayName;
 const token = params.token;
 const groupId = params.groupId;
 const userId = params.userId;
+const locale = params.locale;
```
prprabhu-ms, line 20: nit: This is a test so easy to change. In general, keep the API as tight as possible. Here, the only supported value is really `fr-FR`. This is not going to be obvious from the test code that sets this parameter.

Would be preferable to make this param `params.useFrLocale`, since that's all you need for the test anyway. If later we wanted to add more locales, we can always change this to be more generic.

prprabhu-ms (approved):
> Open question I'm not sure the answer on - these tests don't perform any visual snapshot; is this adding more value than the jest tests we have for localization already?
Instead of pulling the text from the html should these be a visual snapshot instead? 

I'd prefer these to be snapshot tests, for the reasons above. We can and should keep more detailed tests as unittests.

alcail (approved):

packages/react-composites/src/composites/CallComposite/ConfigurationScreen.tsx
mgamis-msft: Hmmm `useIdentifiers()` in only for components scope. There are instances where `data-ui-id` attribute is directly assigned to the composite like https://github.com/Azure/communication-ui-library/blob/a311b0f86dddfb868b7e5c527aa15f0bb9a65d35/packages/react-composites/src/composites/CallComposite/LocalPreview.tsx#L68. Maybe we should move `IdentifierProvider` to react-composites but some elements in our components (like MessageThread) might be hard to verify.

alkwa-msft: can we have a useCompositeIdentifiers?

alkwa-msft (approved):

packages/react-composites/tests/browser/call/Localization.test.ts
mgamis-msft: I chose to just look at text because I thought that is the main concern of localization. Also did not want to introduce more flakiness. Although 

> > Open question I'm not sure the answer on - these tests don't perform any visual snapshot; is this adding more value than the jest tests we have for localization already?
> > Instead of pulling the text from the html should these be a visual snapshot instead?
> 
> I'd prefer these to be snapshot tests, for the reasons above. We can and should keep more detailed tests as unittests.

Ok, sounds good. Changed to snapshot tests instead.

packages/react-composites/src/composites/CallComposite/ConfigurationScreen.tsx
mgamis-msft: Never mind. Changed the tests to snapshot tests so I reverted the addition of data-ui-id's.

PorterNan:
localized-call-screen-chromium.png 
This snapshot is a little bit weird, could you please take a look @mgamis-msft ?

PR #813, opened 2021-09-22
Feedback request - Composite isMobile flag

JamesBurnside:
# What
We're looking to add a flag for isMobile in the composites. Looking for feedback on what that API should look like

# Why
We are going to give composites a mobile-first view. For desktop we will be setting and documenting a min-height and width that gives an optimal experience (through customer asks we can then put in work to make this smaller and smaller). But mobile will have things like touch targets and UI elements that are optimized for mobile.

# Questions
* Is this the right place to put this flag?
* Is it OK that it defaults to false (opt-in when everything else is opt-out)

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/CallComposite.tsx
+++ b/packages/react-composites/src/composites/CallComposite/CallComposite.tsx
@@ -38,6 +38,13 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
  */
 export type CallCompositeOptions = {
   /**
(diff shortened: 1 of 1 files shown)


prprabhu-ms (approved):
+1 to option and name.

In particular, agree that Contoso should detect the device and pass this in (we probably should add a story with an example based on, say, mobile-detect.js).

alkwa-msft (approved):

alcail (approved):

JamesBurnside:
Closing this as this was for feedback on API design - implementation will follow shortly

PR #814, opened 2021-09-22
Chat sample header - People button focusable

alcail:
# What
Made People button focusable in Chat sample header

# Why
People button was not focusable so user could not navigate to it and thus turn the participant list panel on/off
bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2584873

# How Tested
ran Chat sample locally

# Process & policy checklist


samples/Chat/src/app/styles/ChatHeader.styles.ts
```diff
@@ -30,21 +30,21 @@ export const leaveIcon: IIconProps = {
   iconName: 'Leave'
 };
 
-export const panelButtonStyle = mergeStyles({
-  padding: '0px 13px',
-  cursor: 'pointer'
```
prprabhu-ms, line 35: This CSS change seems unrelated? (cursor styling got dropped?)

prprabhu-ms (approved):

samples/Chat/src/app/styles/ChatHeader.styles.ts
alcail: it is related. No need to force the cursor to pointer as it's done automatically now that we use IconButton component instead of just Icon

mgamis-msft (approved):

PR #817, opened 2021-09-23
Undo styling to body tag from FluentThemeProvider

mgamis-msft:
# What

Suppressing body styles and font faces from [`teamsTheme`](https://github.com/microsoft/fluentui/blob/bae19a786883449595089d96368e7fa23d9c416b/packages/fluentui/react-northstar/src/themes/teams/siteVariables.ts#L84) for FluentUI N* `Provider` inside our `FluentThemeProvider.` 

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2531039

# How Tested

Tested on `samples/StaticHtmlComposites` by adding text outside our composites like so:
image
and verifying that it is not font Segoe UI when rendered like so:
image

# Process & policy checklist


packages/react-components/src/theming/FluentThemeProvider.tsx
```diff
@@ -34,11 +34,21 @@ export const ThemeContext = createContext<Theme>(defaultTheme);
...
 const initialFluentNorthstarTheme = mergeNorthstarThemes(teamsTheme, {
   componentVariables: {
-    // suppressing teams theme for chat message links to get better styling from Fluent UI Link
+    // suppressing chat message links style from teamsTheme to get better styling from Fluent UI Link
     ChatMessage: {
       linkColor: undefined,
       linkColorMine: undefined
     }
+  },
+  fontFaces: [], // suppressing font faces from teamsTheme
```
prprabhu-ms, line 43: Please add a detailed comment in the code here describing why this is necessary, and link to the bug / bugs.

This is a workaround for a really pesky problem, so let's make it loud and clear in the code :)

prprabhu-ms (approved):
woohoo!

packages/react-components/src/theming/FluentThemeProvider.tsx
prprabhu-ms: Also, for my edification, can you describe why this doesn't break the font/styling for the NorthStar components themselves? Are all the body* styling just spurious (i.e., NorthStar components themselves don't even rely on those?)

anjulgarg (approved):

alkwa-msft (approved):

PR #822, opened 2021-09-23
Add `mobileView` flag to the meeting and call composite interfaces

JamesBurnside:
# What
Add is mobile flag to call composite, meeting composite and calling sample.

# Why
See #813 for more information

# How Tested
Tested sample shows true on mobile using npm package `mobile-detect`.


alkwa-msft:
Can we keep isMobile as a boolean and maybe we'll experiment with using mobile-detect externally?

prprabhu-ms (approved):

prprabhu-ms:
> Can we keep isMobile as a boolean and maybe we'll experiment with using mobile-detect externally?

Didn't get this question. This PR only adds mobile-detect to the sample. I agree with this. The composites shouldn't try to detect if we're on mobile).

alkwa-msft:
> > Can we keep isMobile as a boolean and maybe we'll experiment with using mobile-detect externally?
> 
> Didn't get this question. This PR only adds mobile-detect to the sample. I agree with this. The composites shouldn't try to detect if we're on mobile).

oh right!.. disregard my last comment.

alkwa-msft (approved):

PR #823, opened 2021-09-23
Document public API for acs-ui-common

prprabhu-ms:
For the acs-ui-common packlets:

* All API exported from modules must now have docstrings.
* This is annoying and not as useful for API exported from internal modules, but not exported from packages, so allow marking such API with @private
* All API exported from the packages must now have release tags @public, @beta etc: https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags

Release tags are a good idea anyway -- they make sure that all our package exported API is intentional, no leakages.
Coupled with the @private, it serves another purpose -- if we mark something @private to satisfy eslint, but then mark it @public to satisfy api extractor, we know something's gone wrong ;)

This PR does not yet enable the enforcement of release tags from api extractor. That can only be done once all packlets are clean.


packages/acs-ui-common/review/acs-ui-common.api.md
```diff
@@ -28,19 +28,19 @@ export type CommonProperties<A, B> = {
...
+// @public
 export const fromFlatCommunicationIdentifier: (id: string) => CommunicationIdentifier;
 
-// @public (undocumented)
+// @public
 export type FunctionWithKey<KeyT, ArgsT extends unknown[], RetT> = (key: KeyT, ...args: ArgsT) => RetT;
 
-// @public
-export const getApplicationId: () => string;
+// @internal
```
JamesBurnside, line 37: Does @internal mean used in other packages in the repo but shouldn't be exported publically?

JamesBurnside (approved):

packages/acs-ui-common/review/acs-ui-common.api.md
prprabhu-ms: Yes, for us.

According to api-extractor docs, this marks *public API* that is intended not be used except for internal purposes... https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags

In our case, I think we should simply make sure that aren't any @internal in the communication-react.api.md (it's OK to have them in the packlet APIs)

alcail (approved):

PR #827, opened 2021-09-24
Smart Dominant Speakers Function

anjulgarg:
# What
Adds a function `smartDominantSpeakers` to react-components package that will return dominant speakers gracefully ordered.
This function can be consumed inside `VideoGallery` or similar components to get order remoteParticipants on the basis of their dominance.
Works for both audio only and video participants.

# Why
Too many re-renders if we use the standard dominant speaker events without gracefully handling them.

# How Tested

# Process & policy checklist


packages/calling-component-bindings/src/videoGallerySelector.ts
```diff
@@ -129,31 +129,29 @@ const sortedRemoteParticipants = (
...
+      const idxA = dominantSpeakers[a.userId];
+      const idxB = dominantSpeakers[b.userId];
       if (idxA === undefined && idxB === undefined) return 0; // Both a and b don't exist in dominant speakers.
       if (idxA === undefined && idxB >= 0) return 1; // b exists in dominant speakers.
       if (idxB === undefined && idxA >= 0) return -1; // a exists in dominant speakers.
       return idxA - idxB;
     });
   }
 
-  return participantsWithVideo.concat(participantsWithoutVideo);
```
prprabhu-ms, line 156: nitty nit: Earlier direct return was better?

packages/communication-react/review/communication-react.api.md
```diff
@@ -1809,6 +1809,9 @@ export interface SendBoxStylesProps extends BaseCustomStylesProps {
     textField?: IStyle;
 }
 
+// @public
+export const smartDominantSpeakerParticipants: (participants?: VideoGalleryRemoteParticipant[], dominantSpeakers?: Array<string>, visibleParticipants?: VideoGalleryRemoteParticipant[], MAX_TILES?: number) => VideoGalleryRemoteParticipant[];
```
prprabhu-ms, line 1813: This shouldn't be a part of the API.

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { VideoGalleryRemoteParticipant } from '../types';
+
+/**
+ * Calculates the participants that should be renderd basis on the list of dominant speakers and currently renderd participants in a call.
+ * @param participants - Array containing all participants of a call. {@link @azure/communication-react#VideoGalleryRemoteParticipant}
+ * @param dominantSpeakers - An array containing the userId of dominant speakers in a call in the order of their dominance. 0th index is the most dominant, 1st is the second most etc
```
prprabhu-ms, line 8: nitty nit: This line is a running too long. As a guideline, avoid going over 120 char.
I can't see it on github review without scrolling right.

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @param visibleParticipants - Array containing currently rendered (visible) participants in the call. {@link @azure/communication-react#VideoGalleryRemoteParticipant}
```
prprabhu-ms, line 9: Would be nice to name this so that it's obvious that unlike `participants` and `dominantSpeakers`, this stores the *last* layout.

How about `previousParticipants`?

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @param MAX_TILES - Maximum number of tiles to calculate.
+ * @returns VideoGalleryRemoteParticipant[] {@link @azure/communication-react#VideoGalleryRemoteParticipant}
+ */
+export const smartDominantSpeakerParticipants = (
+  participants: VideoGalleryRemoteParticipant[] = [],
+  dominantSpeakers: Array<string> = [],
+  visibleParticipants: VideoGalleryRemoteParticipant[] = [],
+  MAX_TILES = 4
```
prprabhu-ms, line 17: Should just be `maxTiles`. SNAKE_CASE is reserved for global constant values.

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,61 @@
...
+): VideoGalleryRemoteParticipant[] => {
+  if (!participants) return [];
+
+  // Only use the Max allowed dominant speakers.
+  const dominantSpeakerIds = dominantSpeakers.slice(0, MAX_TILES);
+
+  // Don't apply any logic if total number of video streams is less than Max video streams.
+  if (participants.length <= MAX_TILES) return participants;
```
prprabhu-ms, line 25: nit: This should be before the logic in line 22 (we don't even need the dominant speakers, so why clip it?)

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,61 @@
...
+  // Initialize `visibleParticipants` if it is empty.
+  if (!visibleParticipants.length) {
+    visibleParticipants = participants.slice(0, MAX_TILES);
+  }
+
+  const lastVisibleSpeakerIds = visibleParticipants.map((speaker) => speaker.userId);
+  const newDominantSpeakerIds = dominantSpeakerIds.filter((id) => !lastVisibleSpeakerIds.includes(id));
+
+  // Remove participants that are no longer dominant and replace them with new dominant speakers.
+  lastVisibleSpeakerIds.forEach((id, idx) => {
```
prprabhu-ms, line 36: This updates `lastVisibleSpeakerIds`. This is a surprising side-effect of calling `smartDominantSpeakerParticipants`. I'd recommend treating input args as immutable. The caller of `smartDominantSpeakerParticipants` should explicitly assign the returned value to their stored `lastVisibleSpeakerIds` in order to be able to use it on next render.

prprabhu-ms: In order to avoid the caller having to generate the speaker IDs from the returned participant, this function could just accept the same type as the return: `previouslyVisibleParticipants: VideoGalleryRemoteParticipant[]`, and then get the IDs here as needed.

The caller is expected to do something like:

```
const previouslyVisibleParticipants = useRef([] as VideoGalleryRemoteParticipant[]);

...

const newVisibleParticipants = smartDominantSpeakerParticipants(participants, ..., previouslyVisibleParticipants.current);
previouslyVisibleParticipants.current = newVisibleParticipants;

// now render newVisibleParticipants

```

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,61 @@
...
+  // Add additional participants to the final list of visible participants if the list has less than Max visible participants.
+  if (videoParticipantsToRender.length < MAX_TILES) {
+    const diff = MAX_TILES - videoParticipantsToRender.length;
+    for (let index = 0; index < diff; index++) {
+      const filler = participants.find((p) => !videoParticipantsToRender.includes(p));
+      filler && videoParticipantsToRender.push(filler);
+    }
+  }
+
+  return videoParticipantsToRender;
```
prprabhu-ms, line 60: I'm a little concerned about rendering performance.

If there are changes to dominant speaker list of participant list but in a way that none of the visible participants change, we still rerender the gallery. Is that a concern?

In the end, should we compare the final result with `previouslyVisibleParticipants` (assuming you've updated that to cache the whole return value instead of just the IDs), and if all the contents are the same, return `previouslyVisibleParticipants` instead of the newly constructed list with the same items?

packages/react-components/src/gallery/dominantSpeaker.ts
prprabhu-ms, line 36: nitty nit `(id, idx)` is confusing.

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,61 @@
...
+  const videoParticipantsToRender = participants.filter((p) => lastVisibleSpeakerIds.includes(p.userId));
+  videoParticipantsToRender.sort((a, b) => {
+    return lastVisibleSpeakerIds.indexOf(a.userId) - lastVisibleSpeakerIds.indexOf(b.userId);
+  });
+
+  // Add additional participants to the final list of visible participants if the list has less than Max visible participants.
+  if (videoParticipantsToRender.length < MAX_TILES) {
+    const diff = MAX_TILES - videoParticipantsToRender.length;
+    for (let index = 0; index < diff; index++) {
+      const filler = participants.find((p) => !videoParticipantsToRender.includes(p));
```
prprabhu-ms, line 55: This is potentially O(N*N) because `includes` checks is linear time.

packages/react-components/src/gallery/dominantSpeaker.test.ts
```diff
@@ -0,0 +1,86 @@
...
+const visibleParticipants: VideoGalleryParticipant[] = [
+  { userId: '1' },
+  { userId: '2' },
+  { userId: '3' },
+  { userId: '4' }
+];
+
+describe('Test smartDominantSpeakerParticipants function', () => {
+  test('returns all participants when number of participants is less than or equal to number of dominant speakers', () => {
+    let result = smartDominantSpeakerParticipants(participants.slice(0, 3), dominantSpeakers, visibleParticipants);
```
prprabhu-ms, line 28: The tests that don't verify how `visibleParticipants` is handled, shoud pass in an empty list.

packages/react-components/src/gallery/dominantSpeaker.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    const result = smartDominantSpeakerParticipants(participants, dominantSpeakers, visibleParticipants, 4);
+    const resultUserIds = result.flatMap((p) => p.userId);
+    expect(resultUserIds).toEqual(['1', '5', '3', '7']);
+  });
+
+  test('maintains the ordering of new participants to match last visible participants', () => {
+    const result = smartDominantSpeakerParticipants(
+      participants,
+      ['1', '2', '3', '4'],
+      visibleParticipants.slice(2, 4),
```
prprabhu-ms, line 58: This is a bit of a matter of taste, but I like unittests to be DAMP (the opposite of Dont Repeat Yourself, but I forget what the letters stand for) -- I don't want to go understand what `visibleParticipants.slice(2, 4)` does. I'd just inline the exact object here. I'd actually also inline the `participants` list in each of these tests (perhaps making the list shorter. All these tests don't need 8 participants each. Just give them what they need, and no more).

packages/react-components/src/gallery/dominantSpeaker.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    const result = smartDominantSpeakerParticipants(
+      participants,
+      customDominantSpeakers,
+      visibleParticipants,
+      customDominantSpeakers.length
+    );
+    const resultUserIds = result.flatMap((p) => p.userId);
+    expect(resultUserIds).toEqual(customDominantSpeakers);
+  });
+});
```
prprabhu-ms, line 86: Based on a comment on the implementation: Add a test that if there are spurious changes to participant list (and (b) to dominant speaker list) that don't result in the visible participants changing, then the same list instance is returned as previous (to avoid a render)

packages/react-components/src/gallery/dominantSpeaker.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    expect(result.length).toBe(4);
+  });
+
+  test('returns participants array replacing non-dominant speakers with dominant speakers', () => {
+    const result = smartDominantSpeakerParticipants(participants, dominantSpeakers, visibleParticipants, 4);
+    const resultUserIds = result.flatMap((p) => p.userId);
+    expect(resultUserIds).toEqual(['1', '5', '3', '7']);
+  });
+
+  test('maintains the ordering of new participants to match last visible participants', () => {
```
prprabhu-ms, line 54: This might benefit from more fine-grained tests:

* When a dominant speaker that was visible is removed from participant list
* When a dominant speaker that was visible is removed from dominant speaker list, and another replaces them in dominant speaker list
* When a participant was visible even though they were not dominant, and now they become dominant


... several more along these lines.

prprabhu-ms (commented):
Good test coverage. I'd recommend a little more find-grained coverage around sorting (but there's no end to that really).

packages/calling-component-bindings/src/videoGallerySelector.ts
anjulgarg: All this code will be removed in a subsequent PR. ;)

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: I think my code formatter missed this. (Y)

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: It's not really previous participants. They are currently visible participants...

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: Just a coding style I prefer. Defining important variables up top.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: - `lastVisibleSpeakerIds` is a local variable inside the function.  I think you are talking about `visibleParticipants`
- When this is implemented, video gallery will need to consume the participants returned by this function, so can't return just ID's as then video gallery will have to convert those IDs to participant again...

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: The constant re-renders are un-avoidable now since we are also using the isSpeaking feature. Which causes dominant speakers to change and hence re-render whenever dominant speakers change.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: It is O(M*N)
Worst case M is equal to MAX_TILES. N is the number of participants.
That's not a lot but will add a map to reduce that...

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: Do read the code block again, I think you might have had a confusion.

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
...
+ * @param dominantSpeakers - An array containing the userId of dominant speakers
+ * in a call in the order of their dominance. 0th index is the most dominant, 1st is the second most etc
+ * @param visibleParticipants - Array containing currently rendered (visible)
+ * participants in the call. {@link @azure/communication-react#VideoGalleryRemoteParticipant}
+ * @param maxTiles - Maximum number of tiles to calculate.
+ * @returns VideoGalleryRemoteParticipant[] {@link @azure/communication-react#VideoGalleryRemoteParticipant}
+ */
+export const smartDominantSpeakerParticipants = (
+  participants: VideoGalleryRemoteParticipant[] = [],
+  dominantSpeakers: Array<string> = [],
```
alcail, line 19: what's the difference between Array<string> and string[]?

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
...
+  visibleParticipants: VideoGalleryRemoteParticipant[] = [],
+  maxTiles = 4
```
alcail, line 21: This number is a restriction from SDK or just our design?

alcail (approved):

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: No difference.

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -3,59 +3,78 @@
...
+  let currentParticipants = visibleParticipants.slice();
+  // Initialize `currentParticipants` if it is empty.
+  if (!currentParticipants.length) {
+    currentParticipants = participants.slice(0, maxTiles);
   }
 
-  const lastVisibleSpeakerIds = visibleParticipants.map((speaker) => speaker.userId);
+  // Only use the Max allowed dominant speakers.
+  const dominantSpeakerIds = Array.from(new Set(dominantSpeakers).values()).slice(0, maxTiles);
+  const lastVisibleSpeakerIds = currentParticipants.map((p) => p.userId);
```
prprabhu-ms, line 36: naming nit: Let's consistently use participants vs speakers -- speakers for stuff coming from dominant speakers, and participants for the various sub-lists from particiapnts.

So, this should be lastVisibleParticipantIDs (also, very nitty nit, I'd map from `visibleParticipants`)

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -3,59 +3,78 @@
...
   const newDominantSpeakerIds = dominantSpeakerIds.filter((id) => !lastVisibleSpeakerIds.includes(id));
 
   // Remove participants that are no longer dominant and replace them with new dominant speakers.
-  lastVisibleSpeakerIds.forEach((id, idx) => {
-    if (!dominantSpeakerIds.includes(id)) {
+  for (let index = 0; index < maxTiles; index++) {
+    const oldUserId = lastVisibleSpeakerIds[index];
+
+    if ((oldUserId && !dominantSpeakerIds.includes(oldUserId)) || oldUserId === undefined) {
```
prprabhu-ms, line 43: Isn't this the same as `oldUserId === undefined || !dominantSpeakerIds.includes(oldUserId)` ?
Is there a falsey value other than undefined that `oldUserId` could take?

Also nit: Even if this conditions stays the same, I'd check for `undefined` first

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -3,59 +3,78 @@
...
       const replacement = newDominantSpeakerIds.shift();
-      if (replacement) {
-        lastVisibleSpeakerIds[idx] = replacement;
-      }
+      if (!replacement) break;
```
prprabhu-ms, line 45: Style nit: Avoid hanging conditionals.

    if (!replacement) {
        break;
    }

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -3,59 +3,78 @@
...
+      lastVisibleSpeakerIds[index] = replacement;
     }
+  }
```
prprabhu-ms, line 48: This is a complex algorithm, and small naming changes can make a big difference.

e.g., below this line, `lastVisibleSpeakerIds` no longer refers to last render. Also, it refers to participants, not speakers.

The length of `lastVisibleSpeakerIds` is bounded (by maxTiles), so I'd not be worried about creating a copy. Something like:

```
const newParticipantIDs = []
for (const i = 0; i < maxTiles; i++) {
    const oldUserId = lastVisibleSpeakerIds[index];
    if ((!oldUserId || !dominanSpeakerIDs.includes(oldUserId)) && newDominantSpeakerIds.length > 0 ) {
         newParticipantIDs.push(newDominantSpeakerIds.shift());
    } else if(oldUserId) {
        newParticipantIDs.push(oldUserId);
    } else {
        break;
    }
}
```

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -3,59 +3,78 @@
...
+
+  // Search optimization
+  const lastVisibleSpeakersMap = {};
```
prprabhu-ms, line 51: Should this be `newParticipantIDsMap` since we've already updated lastVisibleSpeakerIds by this point?

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -3,59 +3,78 @@
...
-  const videoParticipantsToRender = participants.filter((p) => lastVisibleSpeakerIds.includes(p.userId));
-  videoParticipantsToRender.sort((a, b) => {
+  newSpeakers.sort((a, b) => {
     return lastVisibleSpeakerIds.indexOf(a.userId) - lastVisibleSpeakerIds.indexOf(b.userId);
   });
 
   // Add additional participants to the final list of visible participants if the list has less than Max visible participants.
-  if (videoParticipantsToRender.length < MAX_TILES) {
-    const diff = MAX_TILES - videoParticipantsToRender.length;
+  if (newSpeakers.length < maxTiles) {
```
prprabhu-ms, line 68: Might be cleaner to do this operation on the `newParticipantIDs` list above. Right now, some of the inclusion and ordering decision is made on the ids list above, and some of it is made here.

Instead, arrive at the already sorted list of IDs, *then* go find the participants for each of the IDs.

prprabhu-ms (commented):
I'm being more than usual pedantic about code minutiae because this is an involved algorithm. Readability is very important.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: Haha, yes it's the same. Guess I am failing my coding round.

anjulgarg:
> I'm being more than usual pedantic about code minutiae because this is an involved algorithm. Readability is very important.

Appreciate it. I will be gone for a month so its more important that others find it easier to work with.

packages/calling-component-bindings/src/videoGallerySelector.ts
```diff
@@ -114,7 +114,7 @@ const dominantSpeakersWithFlatId = (dominantSpeakers?: DominantSpeakersInfo): un
  */
 const sortedRemoteParticipants = (
   participants?: VideoGalleryRemoteParticipant[],
-  dominantSpeakers?: string[]
+  dominantSpeakers?: Record<string, number>
```
JamesBurnside, line 117: `<string, number>` is difficult to know what the types are, in the function can you mention this is userId -> dominantSpeakerRank
Or better yet do we have types for these? e.g. `Record<flatCommunicationIdentifier, dominantSpeakerRank>`

packages/react-components/src/gallery/dominantSpeaker.ts
JamesBurnside: nit: It's bit of a faff but could it alos be hoisted to the top of the file as `const MAX_TILES_SUPPORTED = 4` with the comment about the SDk limitation - just makes it super clear and easy to change (person changing it knows they only have to change one place and haven't missed somewhere)

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { VideoGalleryRemoteParticipant } from '../types';
+
+/**
+ * Calculates the participants that should be renderd basis on the list of dominant
```
JamesBurnside, line 6: nit sp: renderd -> rendered this line and below line

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
...
+ * {@link @azure/communication-react#VideoGalleryRemoteParticipant}
+ * @param dominantSpeakers - An array containing the userId of dominant speakers
+ * in a call in the order of their dominance. 0th index is the most dominant, 1st is the second most etc
+ * @param visibleParticipants - Array containing currently rendered (visible)
+ * participants in the call. {@link @azure/communication-react#VideoGalleryRemoteParticipant}
+ * @param maxTiles - Maximum number of tiles to calculate.
+ * @returns VideoGalleryRemoteParticipant[] {@link @azure/communication-react#VideoGalleryRemoteParticipant}
+ */
+export const smartDominantSpeakerParticipants = (
+  participants: VideoGalleryRemoteParticipant[] = [],
```
JamesBurnside, line 18: nit: I don't think participants and dominantSpeakers should default to an empty array, they should be forced to have a value passed in
```suggestion
  participants: VideoGalleryRemoteParticipant[],
```

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
...
+  dominantSpeakers: Array<string> = [],
+  visibleParticipants: VideoGalleryRemoteParticipant[] = [],
+  maxTiles = 4
+): VideoGalleryRemoteParticipant[] => {
+  if (!participants) return [];
```
JamesBurnside, line 23: nit: participants can't be falsey (empty arrays will return true here). The next condition covers `if participants.length === 0 then return early`

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  // Don't apply any logic if total number of video streams is less than Max video streams.
+  if (participants.length <= maxTiles) return participants;
+
+  let currentParticipants = visibleParticipants.slice();
+  // Initialize `currentParticipants` if it is empty.
+  if (!currentParticipants.length) {
+    currentParticipants = participants.slice(0, maxTiles);
```
JamesBurnside, line 31: Is there a reason why we do `participants.slice(0,4)` but just `visibleParticipants.slice()` (without the 0,4)?

JamesBurnside: Is it OK that visible participants might be less than `maxTiles`?

packages/react-components/src/gallery/dominantSpeaker.ts
JamesBurnside: nit: consider `active` or `previously` over `last` here (last is a bit ambiguous, it could be referring to position in the array or the last known visible speakers)

packages/react-components/src/gallery/dominantSpeaker.ts
JamesBurnside: If `oldUserId` is empty string that would be the other falsey value. It _shouldn't_ be but the function could be misused as I don't see anything guarding that from happening. (not sure what should happen in this case just making sure its being considered)

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  // Only use the Max allowed dominant speakers.
+  const dominantSpeakerIds = Array.from(new Set(dominantSpeakers).values()).slice(0, maxTiles);
+  const lastVisibleSpeakerIds = currentParticipants.map((p) => p.userId);
+  const newDominantSpeakerIds = dominantSpeakerIds.filter((id) => !lastVisibleSpeakerIds.includes(id));
+
+  // Remove participants that are no longer dominant and replace them with new dominant speakers.
+  for (let index = 0; index < maxTiles; index++) {
+    const oldUserId = lastVisibleSpeakerIds[index];
```
JamesBurnside, line 41: old has differed from last here, and is not user instead of participant or speaker. Is it `oldestVisibleParticipantId`?

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    if ((oldUserId && !dominantSpeakerIds.includes(oldUserId)) || oldUserId === undefined) {
+      const replacement = newDominantSpeakerIds.shift();
+      if (!replacement) break;
+      lastVisibleSpeakerIds[index] = replacement;
+    }
+  }
+
+  // Search optimization
```
JamesBurnside, line 50: not understanding this comment, is it saying we are using an object for search optimization?
If so update this commment to something like
```
// Convert last visible speakers in an object for easy access to elements.
```

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
...
+  const lastVisibleSpeakersMap = {};
+  lastVisibleSpeakerIds.forEach((userId) => {
+    lastVisibleSpeakersMap[userId] = true;
+  });
+
+  const newSpeakers = participants.filter((p) => lastVisibleSpeakersMap[p.userId]);
```
JamesBurnside, line 56: `lastVisibleSpeakersMap` is used solely here and it makes this function longer and more difficult reading than the potential perf benefits.
Could `lastVisibleSpeakersMap` be removed and instead this become:
```
const newVisibleParticipants = participants.filter((p) => lastVisibleSpeakerIds.includes(p.userId))
```

packages/react-components/src/gallery/dominantSpeaker.ts
JamesBurnside: I would suggest something slightly different, instead of maintaining a `newSpeakersMap`, have instead:
```
const newVisibleParticipants = participants.filter((p) => lastVisibleSpeakerIds.includes(p.userId))
const newNonVisibleParticipants = participants.filter((p) => !lastVisibleSpeakerIds.includes(p.userId))
// Ensure we have maxTiles number of visible participants
if (newVisibleParticipants.length < maxTiles) {
  newVisibleParticipants.push(...newNonVisibleParticipants.slice(0, maxTiles - newVisibleParticipants.length);
}
```

The inefficiencies should be negligible given these are tiny arrays, and given the complex function I'd put more weight on readability

packages/react-components/src/gallery/dominantSpeaker.ts
JamesBurnside: nit: sp: basis -> based

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,80 @@
...
+  const newSpeakersMap = {}; // search optimization
+  newSpeakers.forEach((p) => {
+    newSpeakersMap[p.userId] = true;
+  });
+
+  // Sort the new video participants to match the order of last visible participants.
```
JamesBurnside, line 62: I feel this sort should come after pushing new participants to the newSpeakers array. Otherwise, I think, if those participants were previously visible they will swap places

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: Shallow copying visibleParticipants using slice to avoid mutating it as its a function parameter.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: That's how it was earlier and @prprabhu-ms  advised to add a perf optimization here.
The potential perf benefits might become significant in the near future when acs supports more participants.
I think readability is important, but we need to think of performance as well. Even if that means making an algorithm more complicated in the process.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: Yes, from previous part they are guaranteed not visible.  Doesn't matter if we sort in the end or before adding more participants. Since the sorting is done bases on `lastVisibleSpeakerIds` which is not being changed.
Sorting in the end won't work as 
`lastVisibleSpeakerIds.indexOf(a.userId) - lastVisibleSpeakerIds.indexOf(b.userId);`
both these might be undefined. Newly added participants dont exist in `lastVisibleSpeakerIds`

So leaving it here is important.
Plus a bunch of test cases ensure that the ordering is maintained.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: And yes, it's ok for visibleParticipants to be less than maxTiles.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: This function needs to have it as a parameter since this limit applies to `video` tiles and not to `audio` tiles.
We can render 100s of participants that have their video turned off and a component will need to specify that to this function.

JamesBurnside: Oh yep keep it as a param, just have `maxTiles = MAX_TILES_SUPPORTED`

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: It can't/shouldn't be null ever. That would break the entire app cuz that means that one of our participants has an empty user ID
So I wouldn't worry about that here at all..

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: Doesn't matter how we use it here. With or without braces.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: I have tweaked the variable names.
But a lot of it is subjective.
I am finding the example you have pasted much harder to read TBH.

packages/react-components/src/gallery/dominantSpeaker.ts
JamesBurnside: Oh right currentParticipants are the array of currently visible participants, which might be the array _or_ (we assume it to be) the maxTiles number of currentParticipants.

Can we avoid this second slice and just allow currentParticipants to be an empty array if visibleParticipants.length === 0? (might be wrong) but if its an empty array, looks like it will first push the dominant speakers onto the array in order, then participants.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: I think you are only thinking about video tiles. But this will also be used to handle audio participants where the limits imposed by maxtiles won't apply. For example, we are talking about 10 visible audio participants in a call with 100+ audio participants.
This operation happening every time someone speaks. It can add up really fast...

packages/react-components/src/gallery/dominantSpeaker.ts
JamesBurnside: Hmm OK, I was thinking limits of 50 tiles and 1000 participants which should be trivial but suppose better to future proof

JamesBurnside: Actually better to keep the algorithm agnostic of our constraints anyway

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: I have added some optimizations and removed unneeded code based on your comments.
Should be more readable now.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: Tweaked the rest of the algo and moved it to the bottom. :)

packages/calling-component-bindings/src/videoGallerySelector.ts
anjulgarg: All this code will go away in the next PR for new layouts in video gallery :)

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: I understand your point of view for `VideoGalleryRemoteParticipant` but not for `dominantSpeakers`
Why force someone to provide an empty array?

JamesBurnside (approved):

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: Yes, it can be removed and I have removed it :)
Thanks for pointing it out.

packages/react-components/src/gallery/dominantSpeaker.ts
anjulgarg: Haha, I had it earlier as `MAX_TILES_SUPPORTED` but @prprabhu-ms  advised I change it to `maxTiles`
It's a matter of standard I guess.
I am ok either ways...

But given that it's a function argument, `maxTiles` makes more sense...

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  // Converting array to a hashmap for faster searching
+  const visibleParticipantIdsMap = {};
+  visibleParticipantIds.forEach((userId) => {
+    visibleParticipantIdsMap[userId] = true;
+  });
+
+  // Add additional participants to the final list of visible participants if the list has less than Max visible participants.
+  const emptySlots = maxTiles - visibleParticipantIds.length;
```
prprabhu-ms, line 48: This would be cleaner if we extract then append (instead of checking membership in a map being updated in the loop:

```
const fillers = participants.filter((p) => !visibleParticipantIdsMap[p.userId]);
visibleParticipantIds = visibleParticipantIds.concat(fillers.slice(0, emptySlots))
```

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -0,0 +1,67 @@
...
+        visibleParticipantIdsMap[filler.userId] = true;
+        visibleParticipantIds.push(filler.userId);
+      }
+    }
+  }
+
+  const newVisibleParticipants = participants.filter((p) => visibleParticipantIdsMap[p.userId]);
+
+  // Sort the new video participants to match the order of last visible participants.
+  newVisibleParticipants.sort((a, b) => {
```
prprabhu-ms, line 62: This sort is O(N*N) because of the use of `indexOf` on each compare.

prprabhu-ms (commented):
Thanks for the iteration. The final implementation is quite readable. Left some more small comments that can be handled as follow up.

PR #828, opened 2021-09-24
Mark IdentifierProvider and friends internal

prprabhu-ms:
# What

`IdentifierProvider` is intended for e2e testing needs. It's not part of the stable public API. So mark it @internal and rename as indicated by api-extrator.

No, 'cause @internal ;)

### DIFF
--- a/packages/react-components/review/react-components.api.md
+++ b/packages/react-components/review/react-components.api.md
@@ -323,22 +323,18 @@ export interface GridLayoutProps {
 export type GridLayoutType = 'standard';
 
 // @internal
-export const _IdentifierProvider: (props: IdentifierProviderProps) => JSX.Element;
+export const _IdentifierProvider: (props: _IdentifierProviderProps) => JSX.Element;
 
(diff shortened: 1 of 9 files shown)


alcail (approved):

packages/react-components/review/react-components.api.md
```diff
@@ -322,19 +322,19 @@ export interface GridLayoutProps {
 // @public
 export type GridLayoutType = 'standard';
 
-// @beta
-export const IdentifierProvider: (props: IdentifierProviderProps) => JSX.Element;
+// @internal
+export const _IdentifierProvider: (props: _IdentifierProviderProps) => JSX.Element;
```
JamesBurnside, line 326: Are we planning to release this as part of the API? (If so would `@beta` be better than `@internal`?)

JamesBurnside (approved):

alkwa-msft (approved):

packages/react-components/review/react-components.api.md
prprabhu-ms: This one really is @internal -- to be only used for automation tests, and we don't give API guarantees.

@beta is intended for API that will eventually become @public

PR #829, opened 2021-09-24
Make ControlBarButton styling API consistent

prprabhu-ms:
# What

Small API consistency tweak.

# Why

 Renames a public type.


packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -33,6 +34,13 @@ export interface ControlBarButtonStrings {
   offLabel?: string;
 }
 
+/**
+ * Styles for all {@link ControlBarButton} implementations.
+ *
+ * @public
+ */
+export interface ControlBarButtonStyles extends IButtonStyles {}
```
alcail, line 42: is this for later on when (if) specific styles will be added?

alcail: In the meantime, you could defined it like following to handle this issue:
export type ControlBarButtonStyles = IButtonStyles;

alcail (approved):

alkwa-msft (approved):

JamesBurnside (approved):
Not sure if linting/tsc will allow the empty interface in `export interface ControlBarButtonStyles extends IButtonStyles {}` but happy for this to be a type or interface and rest of changes all look great so approving

packages/react-components/src/components/ControlBarButton.tsx
prprabhu-ms: Lint allows empty interfaces, and this pattern seems to be common in calling SDK.

The intention, that we are extending the interface is made more explicit that way.

Also, changing `type` to `interface` later would be a breaking change, I think.

PR #830, opened 2021-09-24
Simplified component message hierarchy

prprabhu-ms:
Replaces generic type hierarchy for messages in the UI components with discriminated unions.

The new type hierarchy is easier to understand and loses none of the concrete type safety.


prprabhu-ms:
I found the message hierarchy quite confusing. PoC to see if this form is simpler and to see if I missed someething.

packages/react-components/src/types/ChatMessage.ts
```diff
@@ -21,129 +21,124 @@ export type MessageAttachedStatus = 'bottom' | 'top';
 export type MessageContentType = 'text' | 'html' | 'richtext/html' | 'unknown';
```
prprabhu-ms, line 21: This file has the key type rework. Everything else is a fairly straightforward porting over to the new types.

JamesBurnside (approved):
Nice!

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -40,56 +46,50 @@ const convertToUiChatMessage = (
...
-              (participant): CommunicationParticipant => ({
-                userId: toFlatCommunicationIdentifier(participant.id),
-                displayName: participant.displayName
-              })
-            ) ?? [],
-        type: systemMessageType,
-        messageId: message.id,
-        iconName: systemMessageType === 'participantAdded' ? 'PeopleAdd' : 'PeopleBlock'
-      }
+      messageType: 'system',
```
anjulgarg, line 67: any reason for moving away from message.type to message.messageType?
`message` prefix feels redundant.

prprabhu-ms: To disambiguate from these other fields: `systemMessageType` (which further discriminates `SystemMessage` variants) and `contentType` (a field in `ChatMessage`).

anjulgarg: Approved :)

prprabhu-ms:
Manually sanity checked ChatComposite via storybook.

Checked artifacts from the failed UI test -- the checked in snapshot for the localization test is clearly wrong for linux (shows an error bar).

alkwa-msft (approved):

Issue #605, opened 2021-07-20
Add the option to set margin values for the chat message container


aakashbhardwaj619:
**What feature are you requesting?**
Add the option to set margin values for the chat message container.

**What would be the benefit of adding this feature?**
This will give more control to style the containers as per requirement.

**What solution would be ideal for you?**
The **margin-left** and **margin-right** options should apply to the chat message container.

**What alternatives have you considered?**
I tried setting **margin-right** to **0px** using the below code but it does not work.

```typescript
const messageThreadStyle: MessageThreadStylesProps = {
  chatMessageContainer: {
    marginRight: '0px'
  }
};
```

**Attach screenshots or images add detail to your request:**
<img width="773" alt="AcsAlignMessage" src="https://user-images.githubusercontent.com/33526902/126287939-3ce4de17-38ab-4cc3-8033-febe307b8b7d.png">

JamesBurnside:
@aakashbhardwaj619 thanks for filing this issue, I agree it would be useful to be able to remove these margins and by default there doesn't look to be a way to do that through our current API - we'll look into creating a fix for this

JamesBurnside:
For more context on this issue this component is currently wrapping around fluent northstar's (N*) chat component: https://fluentsite.z22.web.core.windows.net/0.57.0/components/chat/definition
By default they add this spacing in there without an interface to remove it.

The culprit component lies in an implictly defined element between the chat container element: https://github.com/Azure/communication-ui-library/blob/2cbec06050204f200bda2c5034a02b30f638f6af/packages/react-components/src/components/MessageThread.tsx#L833 and the head element of our rendered chat messages: https://github.com/Azure/communication-ui-library/blob/2cbec06050204f200bda2c5034a02b30f638f6af/packages/react-components/src/components/MessageThread.tsx#L330

(note: it is _not_ related to the gutter property the N* chat component exposes)

For fixing this we would likely need to target the parent element of the message but we should engage with the fluent team to provide a mechanism from which to do so. We could directly override the parent element's css but this would be a brittle fix. @alkwa-msft would you be able to continue this investigation further?

PR #834, opened 2021-09-27
exposing styling through composite

alkwa-msft:
# What
We need to enable developers to customize the message styling in a message thread

# Why
We want to enable developers to be able to flexibly change their message thread styling to fit their intended UI.

# How Tested
Ran the samples locally

# Process & policy checklist


alkwa-msft:
We will now be able to change the styling of the message components in the message thread. This enables us to customize the spacing around the messages which was raised in a previous github issue
https://github.com/Azure/communication-ui-library/issues/605

Before:
image

After:
image

change/@internal-react-components-97ccb55b-fe63-4cc1-8003-18bf04a13072.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "exposing styling through composite",
```
JamesBurnside, line 3: nit: for changelog entries try making entries both specific to the package, and clear what has changed to a user who may read this. E.g. for react-components package this should probably be `Expose chat message styling through MessageThread interface`.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -38,11 +39,18 @@ export type ChatScreenProps = {
   onRenderMessage?: (messageProps: MessageProps, defaultOnRender?: MessageRenderer) => JSX.Element;
   onRenderTypingIndicator?: (typingUsers: CommunicationParticipant[]) => JSX.Element;
   onFetchParticipantMenuItems?: ParticipantMenuItemsCallback;
+  styles?: MessageThreadStylesProps;
```
JamesBurnside, line 42: This implies styles for the whole chat screen but only has styles for the MessageThread. Open to renaming this to messageThreadStyles or having something like:
```
interface ChatScreenStyles {
   messageThreadStyles: MessageThreadStylesProps
}

...

export type ChatScreenProps = {
   ...
   styles: ChatScreenStyles;
}

```

alkwa-msft: yes lets change that to align with Prathmesh's latest change I will get in on that.

change/@internal-react-components-97ccb55b-fe63-4cc1-8003-18bf04a13072.json
alkwa-msft: Would I create a new rush changelog or is it safe to just rename these entries?

packages/communication-react/review/communication-react.api.md
```diff
@@ -1499,9 +1499,10 @@ export interface MessageThreadStrings {
 }
 
 // @public
-export interface MessageThreadStylesProps extends BaseCustomStylesProps {
+export interface MessageThreadStyles extends BaseCustomStylesProps {
     chatContainer?: ComponentSlotStyle;
-    chatMessageContainer?: ComponentSlotStyle;
+    chatItemMessageContainer?: (mine: boolean) => ComponentSlotStyle;
```
prprabhu-ms, line 1504: Meta point: Each time we accept a function prop, we have to think about performance. 

Will this be called on each render, or just once?

* If it's called just once, it might be surprising to Contoso when changes in the return value (for whatever reason) aren't reflected. 
* If it is called on each render, Contoso needs to ensure that the function is cheap to call repeatedly (I guess in this case we don't need to worry about memoization because the styles will ultimated get transformed to already-memoized className via `mergeStyles`).

For style props, my recommendation would be to not accept function props. Instead, flatten out the options (of course this only works to a limited degree... if we start adding more axes of variability this won't scale)

```
localChatItemMessageCotainer?: ComponentSlotStyle
remoteChatItemMessageContainer?: ComponentSlotStyle
```

and so on.

The reason I think functions are not a good idea here is that styles aren't intended to be updated dynamically, so our API should be optimized for that.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -38,11 +39,18 @@ export type ChatScreenProps = {
   onRenderMessage?: (messageProps: MessageProps, defaultOnRender?: MessageRenderer) => JSX.Element;
   onRenderTypingIndicator?: (typingUsers: CommunicationParticipant[]) => JSX.Element;
   onFetchParticipantMenuItems?: ParticipantMenuItemsCallback;
+  styles?: MessageThreadStyles;
```
prprabhu-ms, line 42: I'm confused. This prop isn't being set by the `ChatComposite` in this PR. Nor is this exposed in some form in the `ChatComposite` API. How will this get set?

packages/communication-react/review/communication-react.api.md
alkwa-msft: ok thats a good idea. I'll separate it out and then we can swap between the two different styles in the component based on the mine property but at least we're not evaluate new style objects each time.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
alkwa-msft: should be piped through now, thanks for checking this. I think my test code was just injecting styles into the ChatComposite component which was incorrect.

JamesBurnside: I was thinking that was intentional to not be exposed on the composite API - are we supporting arbitrary styling ? (based on previous breaking changes/customizations conversations I would assume we set good defaults but not let user apply arbitrary styling inside the composite)

alcail (approved):

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -287,30 +297,42 @@ const memoizeAllMessages = memoizeFnAll(
...
+          ) : onRenderAvatar ? (
+            onRenderAvatar(payload.senderId ?? '', personaOptions)
+          ) : (
+            <Persona {...personaOptions} />
+          )
+        },
         contentPosition: payload.mine ? 'end' : 'start',
         message: {
-          className: mergeStyles({ width: 'calc(100% - 6.25rem)' }),
+          className: mergeStyles({ width: 'calc(100% - 5rem)' }),
```
JamesBurnside, line 334: nit: move into styles file and ensure these items `myChatItemMessageStyle` `chatItemMessageStyle` have it to have the styles in one place

packages/react-components/src/components/styles/InputBoxComponent.style.ts
```diff
@@ -32,6 +32,17 @@ export const inputBoxStyle = mergeStyles({
...
 
+/**
+ *
+ * @private
+ */
+export const textContainerStyle: IStyle = {
+  alignSelf: 'center',
+  position: 'relative',
+  width: '100%',
+  maxWidth: '41.25rem'
```
JamesBurnside, line 43: Does it need this maxWidth? Can it just fill its container and have the container set the desired width?

JamesBurnside: i.e. I see its the same width as the message thread:
image

JamesBurnside:
Looks like there's a regression in the chat message alignment when receiving multiple messages from the same user:
| before (https://aka.ms/acsstorybookdev ) | now |
| -- | -- |
| image | image |

packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -30,7 +30,9 @@ export const messageThreadContainerStyle = mergeStyles({
   width: '100%',
   maxHeight: '100%',
   overflow: 'auto',
-  position: 'relative'
+  position: 'relative',
+  alignSelf: 'center',
+  maxWidth: '41.25rem'
```
JamesBurnside, line 35: If we don't have maxWidth what is the experience like? Does it get **too wide** ? (I've personally only had issues when it gets too small but not too large)

packages/react-components/src/components/styles/TypingIndicator.styles.ts
```diff
@@ -17,7 +17,10 @@ export const typingIndicatorStringStyle = mergeStyles({
   fontWeight: 400,
   whiteSpace: 'pre',
   display: 'flex',
-  alignItems: 'center'
+  alignItems: 'center',
+  maxWidth: '41.25rem',
```
JamesBurnside, line 21: same here; typing indicator, messagethread and inputBox all have this property seperately. (If we keep the max width) move to a container and have these fill the containers width

JamesBurnside (changes requested):
Looks great! Just question on needing maxWidth and regression on the chat message indentation

packages/react-components/src/components/styles/InputBoxComponent.style.ts
alkwa-msft: These are changes suggested by alexp

packages/react-components/src/components/styles/MessageThread.styles.ts
alkwa-msft: Yes the maxWidth is just for when things get too wide. Lets chat with Alexp about these

alkwa-msft:
> Looks great! Just question on needing maxWidth and regression on the chat message indentation

Setting up the specific maxWidth at the composite level and figuring out the correct marginLeft on !mine messages when we aren't showing the avatar

JamesBurnside (approved):

PR #835, opened 2021-09-27
Alcail/a11y bug chat sample zoom400

alcail:
# What
Updated Chat sample styles for each screen (Home, Configuration, Chat, End) to ensure all components are visible or accessible through keyboard when zooming in at 400% or more

# Why
a11y bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2584899

# How Tested
ran chat sample and storybook locally

Before (left) / After (right)

- Home screen
image

- Configuration screen
image

- Chat screen
image

- End screen
image

- Strorybook
image

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
```diff
@@ -74,7 +74,8 @@ export const topicNameLabelStyle = mergeStyles({
  */
 export const participantListWrapper = mergeStyles({
   boxShadow: '0px 0.3px 0.9px rgba(0, 0, 0, 0.1), 0px 1.6px 3.6px rgba(0, 0, 0, 0.13)',
-  width: '15rem',
+  width: '20rem',
+  maxWidth: '50vw',
```
prprabhu-ms, line 78: Please add a comment above this line to say that the max width ensures that the message thread has enough space on narrow viewport.

packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
```diff
@@ -74,7 +74,8 @@ export const topicNameLabelStyle = mergeStyles({
  */
 export const participantListWrapper = mergeStyles({
   boxShadow: '0px 0.3px 0.9px rgba(0, 0, 0, 0.1), 0px 1.6px 3.6px rgba(0, 0, 0, 0.13)',
-  width: '15rem',
+  width: '20rem',
```
prprabhu-ms, line 77: Why the bump to 20rem?

packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
```diff
@@ -95,13 +96,13 @@ export const listHeader = mergeStyles({
...
   height: '100%'
 });
 
 /**
  * @private
  */
 export const participantListStyle = mergeStyles({
-  overflowY: 'hidden'
+  height: '100%',
+  overflow: 'auto'
```
prprabhu-ms, line 107: Do you know why we had 'hidden' earlier? Is it the case the the implementation of ParticipantList provides its own way to scroll? Or was this just a bug?

samples/Chat/src/app/styles/ConfigurationScreen.styles.ts
```diff
@@ -2,31 +2,48 @@
...
 
 export const responsiveLayoutStyle = mergeStyles({
-  display: 'flex',
-  flexDirection: 'row',
-  '@media (max-width: 37.5rem)': {
-    flexDirection: 'column'
-  }
+  height: '100%',
+  width: '100% ',
+  //half childrenGap from Stack
```
prprabhu-ms, line 14: nit: add space

prprabhu-ms (approved):

packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
alcail: children are 20rem, 15rem was truncating the pane and horizontal scrollbar was always there

packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
alcail: bug. The List of participant in chat was not working properly (no vertical scrollbar when there were a lot of people in the chat, so list truncated before)

packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
JamesBurnside: Issue with using `vw` is our composite should be independant of the viewport size - but instead be dependant on the parent container width - i.e. we don't know our composite will take up 100% of the viewport. If someone makes our composite 50% of the viewport size I think this will mean the participantListWrapper will be 100% of the composite width.

Can we use `width: 50%` here?

JamesBurnside (approved):
Looks great! Just one comment about use of `vw`

PR #837, opened 2021-09-28
Document public API for @internal/react-composites

prprabhu-ms:
For the calling-stateful-client and calling-component-bindings packlets,

* All API exported from modules must now have docstrings.
  * This is annoying and not as useful for API exported from internal modules, but not exported from packages, so allow marking such API with @private
* All API exported from the packages must now have release tags @public, @beta etc: https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags

Release tags are a good idea anyway -- they make sure that all our package exported API is intentional, no leakages.
Coupled with the @private, it serves another purpose -- if we mark something @private to satisfy eslint, but then mark it @public to satisfy api extractor, we know something's gone wrong ;)

This PR does not yet enable the enforcement of release tags from api extractor. That can only be done once all packlets are clean.

### DIFF
--- a/packages/communication-react/review/communication-react.api.md
+++ b/packages/communication-react/review/communication-react.api.md
@@ -129,7 +129,7 @@ export type AvatarPersonaData = {
 // @public
 export type AvatarPersonaDataCallback = (userId: string) => Promise<AvatarPersonaData>;
 
-// @public (undocumented)
+// @public
 export type AzureCommunicationCallAdapterArgs = {
     userId: CommunicationUserKind;
     displayName: string;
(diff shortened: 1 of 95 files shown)


packages/communication-react/review/communication-react.api.md
```diff
@@ -258,15 +258,15 @@ export interface CallAdapterSubscribers {
     // (undocumented)
     off(event: 'error', listener: (e: AdapterError) => void): void;
     // (undocumented)
-    on(event: 'participantsJoined', listener: ParticipantJoinedListener): void;
+    on(event: 'participantsJoined', listener: ParticipantsJoinedListener): void;
```
prprabhu-ms, line 261: Some callback names are updated to match the event.

packages/communication-react/review/communication-react.api.md
```diff
@@ -1186,15 +1186,6 @@ export interface _Identifiers {
     videoTile: string;
 }
 
-// @public (undocumented)
-export type IncomingCallListener = (event: {
```
prprabhu-ms, line 1190: Dropped unused export

JamesBurnside (approved):

PR #840, opened 2021-09-28
[POC] Grid Layout V2

mgamis-msft:
# What

Grid Layout is now filling in empty spaces in the grid. It is also responsive to height and width to handle both desktop and mobile and any other aspect ratios. The Grid Layout algorithm is explained in this [document](https://microsoft-my.sharepoint-df.com/:w:/p/miguelgamis/EeA1V1Qt3odMoK045IE7AvwBbfedyp9_2T8kV7e-vJrFoA?e=3osVB8).

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2595536
https://skype.visualstudio.com/SPOOL/_workitems/edit/2595537

# How Tested

On storybook

# Process & policy checklist


common/config/jest/jestSetup.js
```diff
@@ -10,3 +10,6 @@ console.error = (...args) => {
 console.warning = (...args) => {
   throw args;
 };
+
+// Need ResizeObserver for GridLayout
+global.ResizeObserver = require('resize-observer-polyfill');
```
anjulgarg, line 15: Why do we need this?

packages/react-components/src/components/GridLayout.tsx
```diff
@@ -31,43 +24,99 @@ export interface GridLayoutProps {
...
+ */
+export type BlockProps = {
+  horizontal: boolean;
+  numBlocks: number;
+};
+
+const createBlocks = (props: BlockProps & { children: React.ReactNode }): React.ReactNode => {
+  const blocks: JSX.Element[] = [];
+  // Get the percent of 1 block out of the total number of blocks rounded down.
+  const blockPercent = (Math.floor(10000 / props.numBlocks) / 100).toFixed(2);
```
anjulgarg, line 77: Isn't this equal to 
`Math.floor(100 / props.numBlocks).toFixed(2)`

packages/react-components/src/components/GridLayout.tsx
```diff
@@ -31,43 +24,99 @@ export interface GridLayoutProps {
...
+  // Split array of children into as even chunks
```
anjulgarg, line 78: nit: remove the `as`

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+  const _targetRatio = targetRatio ?? TARGET_RATIO;
+  return Math.abs(_targetRatio - A) < Math.abs(_targetRatio - B);
+};
+
+/**
+ * @private
+ */
+export const calculateBlockProps = (n: number, width: number, height: number): BlockProps => {
+  if (width <= 0) {
+    throw Error('Width provided [' + width + '] is less than or equal to 0.');
```
anjulgarg, line 19: The function should not throw errors in the UI. And simply return `numBlocks` as 0 if either height or width is 0.
Since there is no way for a client to handle these errors and they don't serve a purpose.

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+  } else if (height <= 0) {
+    throw Error('Height provided [' + height + '] is less than or equal to 0.');
+  }
+  const areaRatio = width / height;
```
anjulgarg, line 23: `aspectRatio` is the standard term used for this. Any particular reason why this is called `areaRatio`?

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const isRatioBetterThan = (A: number, B: number, targetRatio?: number): boolean => {
+  const _targetRatio = targetRatio ?? TARGET_RATIO;
+  return Math.abs(_targetRatio - A) < Math.abs(_targetRatio - B);
+};
+
+/**
+ * @private
+ */
+export const calculateBlockProps = (n: number, width: number, height: number): BlockProps => {
```
anjulgarg, line 17: Just by looking at this function definition, I can't understand what it does.
What is `n` here for example.
Unless I look at it's implementation.
Can we add some docs and a more descriptive naming for the variables?

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (width <= 0) {
+    throw Error('Width provided [' + width + '] is less than or equal to 0.');
+  } else if (height <= 0) {
+    throw Error('Height provided [' + height + '] is less than or equal to 0.');
+  }
+  const areaRatio = width / height;
+  let rows = Math.floor(Math.sqrt((TARGET_RATIO * n) / areaRatio));
```
anjulgarg, line 24: Could you add a docblock to explain what this step is doing?

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { BlockProps } from '../GridLayout';
+
+const TARGET_RATIO = 16 / 9;
```
anjulgarg, line 6: What are these ratios?
Could you add docblocks for some explanation?

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+    throw Error('Height provided [' + height + '] is less than or equal to 0.');
+  }
+  const areaRatio = width / height;
+  let rows = Math.floor(Math.sqrt((TARGET_RATIO * n) / areaRatio));
+  let cols = Math.ceil(n / rows);
+
+  // Default blocks to be horizontal
+  let horizontal = true;
+  while (rows < n) {
+    // If cell ratio is less than tolerance ratio and try more rows
```
anjulgarg, line 30: nit: and try more -> then try more

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { BlockProps } from '../GridLayout';
+
+const TARGET_RATIO = 16 / 9;
+const TOLERANCE_RATIO = 8 / 9;
+
+const isRatioBetterThan = (A: number, B: number, targetRatio?: number): boolean => {
```
anjulgarg, line 9: you can pass TARGET_RATIO as a default function parameter
`const isRatioBetterThan = (A: number, B: number, targetRatio = TARGET_RATIO): boolean => {`
you can get rid of line 10 then

anjulgarg: What is the deciding factor for which one is better?
From what I can understand, its the one that is closest to the targetRatio i.e., leaves the least amount of whitespace.
Can we document this?

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+// Licensed under the MIT license.
+
+import { BlockProps } from '../GridLayout';
+
+const TARGET_RATIO = 16 / 9;
+const TOLERANCE_RATIO = 8 / 9;
+
+const isRatioBetterThan = (A: number, B: number, targetRatio?: number): boolean => {
+  const _targetRatio = targetRatio ?? TARGET_RATIO;
+  return Math.abs(_targetRatio - A) < Math.abs(_targetRatio - B);
```
anjulgarg, line 11: Wouldn't this be the same as
`return B < A` since target ratio is the same for both of them.

lets assign a constant to `targetRatio = 100`
If `A = 20` and `B = 30`
the original function returns `false` and so does `B < A`.
When `B = 20` and `A = 30`
the original function returns `true` and so does `B < A`.

This would only fail if there is a chance of A or B being negative. But from the way they are being generated, that doesn't look to be the case.

Please correct me if I am wrong.

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+    cols = Math.ceil(n / rows);
+  }
+
+  return { horizontal: horizontal, numBlocks: horizontal ? rows : cols };
+};
+
+/**
+ * @private
+ */
+export const chunkify = <T extends unknown>(items: T[], numChunks: number): T[][] => {
```
anjulgarg, line 56: docs please

packages/storybook/stories/ParticipantList/__snapshots__/ParticipantList.stories.storyshot
```diff
@@ -2,11 +2,11 @@
```
anjulgarg, line 2: This is strange, why have all these unrelated storybook snapshots changed?

packages/react-components/src/components/GridLayout.tsx
```diff
@@ -31,43 +24,99 @@ export interface GridLayoutProps {
...
   const numberOfChildren = React.Children.count(children);
 
-  switch (layout) {
-    case 'standard': {
-      const numberOfColumns = calculateStandardLayoutColumns(numberOfChildren);
-      if (gridCol !== numberOfColumns) setGridCol(numberOfColumns);
-      const numberOfRows = calculateStandardLayoutRows(numberOfChildren, gridCol);
-      if (gridRow !== numberOfRows) setGridRow(numberOfRows);
-      break;
+  const targetRef = useRef<HTMLDivElement>(null);
```
anjulgarg, line 36: nit: `containerRef` might be a better name?

packages/react-components/src/components/GridLayout.tsx
```diff
@@ -31,43 +24,99 @@ export interface GridLayoutProps {
...
+  const [blockProps, setBlockProps] = useState<BlockProps>({
```
anjulgarg, line 37: nit: The name `blockProps` makes me think that Block is a component and these are its props.
Can we use something like `calculatedBlocks` etc?

packages/react-components/src/components/GridLayout.tsx
```diff
@@ -31,43 +24,99 @@ export interface GridLayoutProps {
...
+  for (let i = 0; i < props.numBlocks; i++) {
+    // Create block for i-th chunk children
+    const block = createBlock(props.horizontal, `block-${i}`, `${blockPercent}%`, chunks[i]);
+    // Add block to blocks
+    blocks.push(block);
   }
+  return blocks;
+};
 
+const createBlock = (
```
anjulgarg, line 89: It's a standard React component.
Let's name it as such

`Block` for example.

packages/react-components/src/components/GridLayout.tsx
```diff
@@ -31,43 +24,99 @@ export interface GridLayoutProps {
...
+/**
+ * Props to create blocks for children in Grid Layout
+ *
+ */
+export type BlockProps = {
+  horizontal: boolean;
+  numBlocks: number;
+};
+
+const createBlocks = (props: BlockProps & { children: React.ReactNode }): React.ReactNode => {
```
anjulgarg, line 74: This is a standard react component. Let's name it and use it as such
Example, `Blocks`

packages/react-components/src/components/utils/GridLayoutUtils.ts
mgamis-msft: If `target ratio = 100` and `A = 105` and `B = 90`
`B < A` is false but `|100 - A| < |100 - B|` (5 < 10) is true.
Basically I'm asking if A is closer (ie. have a smaller absolute difference) to the target ratio than B
But yes A and B can't be negative.

packages/react-components/src/components/GridLayout.tsx
mgamis-msft: Removed this because `createBlocks` was causing so much unnecessary re-rendering.

packages/react-components/src/components/GridLayout.tsx
mgamis-msft: Removed this because createBlocks was causing so much unnecessary re-rendering.

packages/storybook/stories/ParticipantList/__snapshots__/ParticipantList.stories.storyshot
mgamis-msft: Happens when you call a new mergeStyles in a *.styles.ts. Removed them now.

common/config/jest/jestSetup.js
mgamis-msft: For some reason, storybook does not have access to the typescript library's `ResizeObserver` and I get this failure when I run the snapshot tests:
image

JamesBurnside: Can you update the comment with this? Something like:
```
Add `ResizeObserver` to globals. Without this jest tests fail with "ReferenceError: ResizeObserver is not defined"
```

packages/react-components/src/components/GridLayout.tsx
```diff
@@ -31,42 +24,42 @@ export interface GridLayoutProps {
...
 
   return (
-    <div
-      className={mergeStyles(gridLayoutStyle, styles?.root)}
-      style={{
-        gridTemplateRows: `repeat(${gridRow}, minmax(0, 1fr))`,
-        gridTemplateColumns: `repeat(${gridCol}, 1fr)`
-      }}
-    >
+    <div ref={containerRef} className={mergeStyles(gridLayoutStyle, dynamicGridStyles, styles?.root)}>
```
prprabhu-ms, line 62: A bit concerned about infinite render loop here.

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+      }
+      break;
+    }
+    rows += 1;
+    columns = Math.ceil(numberOfItems / rows);
+  }
+
+  return { horizontalFill, rows, columns };
+};
+
```
mgamis-msft, line 101: Comment on why you have to do this.

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Creates a styles classname with CSS Grid related styles given GridProps and the number of items to distribute as evenly as possible.
+ * @param numberOfItems - number of items to place in grid
+ * @param gridProps - GridProps that define the number of rows, number of columns, and the fill direction
+ * @returns
+ */
+export const createGridStyles = (numberOfItems: number, gridProps: GridProps): string => {
```
prprabhu-ms, line 108: Can we try to split the children into two slices and pass the children down into two block Components.

This might avoid:
* Having to split into finer grid
* Having to style using nth-last-child

packages/react-components/src/components/utils/GridLayoutUtils.ts
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) Microsoft Corporation.
```
prprabhu-ms, line 1: nit: I'd just keep these helper functions etc in `GridLayout.tsx`

They're all used by that file, and splitting into a Utils file doesn't add much value.

alcail:
Thanks for all the explanation. Awesome work

alcail (approved):

packages/react-components/src/components/GridLayout.tsx
```diff
@@ -31,42 +24,42 @@ export interface GridLayoutProps {
...
+      }
+    };
+    const observer = new ResizeObserver(updateDynamicGridStyles);
+    if (containerRef.current) {
+      observer.observe(containerRef.current);
     }
-  }
+    updateDynamicGridStyles();
+    return () => observer.disconnect();
+  }, [numberOfChildren, containerRef.current?.offsetWidth, containerRef.current?.offsetHeight]);
```
alkwa-msft, line 57: I like that we only recalculate the layout based on the number of children so if we start doing things like dominant speaking where they may change position but the length is the same, we can reuse the exact layout

PR #843, opened 2021-09-28
Update create release branch to using machine account

JamesBurnside:
# What
Use machine account that nan had created for the update-snapshots. Use it for the create release branch action so when the branch is created a CI action will be subsequently triggered

# Why
Actions by default cannot trigger other actions to prevent infinite loops.

### DIFF
--- a/.github/workflows/create-release-branch.yml
+++ b/.github/workflows/create-release-branch.yml
@@ -19,6 +19,12 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+          # Use a machine account when checking out. This is to workaround the issue were GitHub
+          # actions, when using the default account, cannot trigger other actions. And we want this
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release-branch.yml
```diff
@@ -19,6 +19,12 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+          # Use a machine account when checking out. This is to workaround the issue were GitHub
+          # actions, when using the default account, cannot trigger other actions. And we want this
+          # action to trigger the regulaar CI pipeline on the created branch.
```
prprabhu-ms, line 24:
```suggestion
          # action to trigger the regular CI pipeline on the created branch.
```

prprabhu-ms (approved):

alcail (approved):

PR #844, opened 2021-09-28
Prune changelog for beta.6

prprabhu-ms:
Prune the changelog down for humans.

### DIFF
--- a/packages/communication-react/CHANGELOG.md
+++ b/packages/communication-react/CHANGELOG.md
@@ -1,12 +1,16 @@
 # Change Log - @azure/communication-react
 
+## [1.0.0-beta.6](https://github.com/azure/communication-ui-library/tree/@azure/communication-react_v1.0.0-beta.6)
+
+TODO
+
 This log was last generated on Mon, 13 Sep 2021 21:02:16 GMT and should not be manually modified.
 
 <!-- Start content -->
 
 ## [1.0.0-beta.5](https://github.com/azure/communication-ui-library/tree/@azure/communication-react_v1.0.0-beta.5)
 
-Mon, 13 Sep 2021 21:02:16 GMT 
+Mon, 13 Sep 2021 21:02:16 GMT
 [Compare changes](https://github.com/azure/communication-ui-library/compare/@azure/communication-react_v1.0.0-beta.4..@azure/communication-react_v1.0.0-beta.5)
 
 ### Changes
@@ -49,7 +53,7 @@ Mon, 13 Sep 2021 21:02:16 GMT
 
 ## [1.0.0-beta.4](https://github.com/azure/communication-ui-library/tree/@azure/communication-react_v1.0.0-beta.4)
 
-Mon, 16 Aug 2021 21:18:19 GMT 
+Mon, 16 Aug 2021 21:18:19 GMT
 [Compare changes](https://github.com/azure/communication-ui-library/compare/@azure/communication-react_v1.0.0-beta.3..@azure/communication-react_v1.0.0-beta.4)
 
 ### Changes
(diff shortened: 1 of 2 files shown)


packages/communication-react/CHANGELOG.md
```diff
@@ -1,12 +1,32 @@
...
+Tue, 28 Sep 2021 19:19:18 GMT
+[Compare changes](https://github.com/azure/communication-ui-library/compare/@internal/react-components_v1.0.0-beta.5..@internal/react-components_v1.0.0-beta.6)
+
+### Changes
+
+* Support editing and deleting sent messages in the `MessageThread` UI
+* Support customization to the `ParticipantList` to inject menu items for each participant
+* (breaking change) API consistency and documentation updates
+  * Add a uniform way to set cutomization options on the composites.
+* Add `MeetingAdapter` to back the `MeetingComposite`. By default, the `MeetingAdapter` is backed by the Azure Communication Services backends.
```
prprabhu-ms, line 18: @JamesBurnside Do we want to mention `MeetingComposite` in the release notes at this point?

packages/communication-react/CHANGELOG.md
```diff
@@ -1,12 +1,32 @@
...
 
-This log was last generated on Mon, 13 Sep 2021 21:02:16 GMT and should not be manually modified.
+This log was last generated on Tue, 28 Sep 2021 19:19:18 GMT and should not be manually modified.
 
 <!-- Start content -->
 
+## [1.0.0-beta.6](https://github.com/azure/communication-ui-library/tree/@internal/react-components_v1.0.0-beta.6)
+
+Tue, 28 Sep 2021 19:19:18 GMT
+[Compare changes](https://github.com/azure/communication-ui-library/compare/@internal/react-components_v1.0.0-beta.5..@internal/react-components_v1.0.0-beta.6)
```
JamesBurnside, line 10: Something for James to fix in the future, this should be something like: https://github.com/azure/communication-ui-library/compare/1.0.0-beta.5..1.0.0-beta.6

packages/communication-react/CHANGELOG.md
JamesBurnside: Yeah we should be fine too - would think about adding some sort of (BETA) but probably unecessary

JamesBurnside (approved):

packages/communication-react/CHANGELOG.md
prprabhu-ms: Actually, not sure if this is James or Prathmesh that needs a fix.

I didn't have any entries in the changelog here at all. I started by copy-pasta'ing in the changes from all the packlets. Now you know which one I copied first.

More importantly though, I think your new link won't work either: 
image

I think the problem is that we delete the release branch when we're done (so there's no https://github.com/Azure/communication-ui-library/tree/release/1.0.0-beta.5 anymore).

packages/communication-react/CHANGELOG.md
```diff
@@ -1,12 +1,32 @@
...
+
+### Changes
+
+* Support editing and deleting sent messages in the `MessageThread` UI
+* Support customization to the `ParticipantList` to inject menu items for each participant
+* (breaking change) API consistency and documentation updates
+  * Add a uniform way to set cutomization options on the composites.
```
alcail, line 17: nit: extra space to be removed

packages/communication-react/CHANGELOG.md
JamesBurnside: It smore we haven't created the beta.6 tag, its only created at release time, e.g. comparing beta-4 and beta-5 is: https://github.com/azure/communication-ui-library/compare/1.0.0-beta.4..1.0.0-beta.5

alcail (approved):

packages/communication-react/CHANGELOG.md
prprabhu-ms: Added more spaces. Intended it to be a sub-list :)

PR #818, opened 2021-09-23
UI test `fixture.ts` cleanup

JamesBurnside:
# What
Move common functionality out of fixture.ts into helper files.
Move common constants into a `defaults` file
Move common browser config into a `browsers` file (hopefully we'll be adding more of these at some point)
Made each test use the fixture created pages (some were generating a new `page`)
Made chat custom data model test display messages in all caps (was hard to debug without this)

# Why
Started doing other work and then clean up took over so putting into a separate PR.
Fixture file is hard to grasp with everything in one file - plus a lot of duplicated code.

# How Tested
Ran ui tests locally. Please review all snapshots updated in PR.

PR #847, opened 2021-09-29
Cleanup `fixture.ts` files

JamesBurnside:
# What
* Move common functionality into helper files
  * Try avoid the Utils file becoming a dumping ground
  * Ensure tests are using the same code in the same way
* Make fixture files more readable
* Make fixtures use the same browser setup (same browser args are now passed in)
  * This made small diffs to the chat text

# Why
Cleanup for incoming changes.. #818 gotten too unruly - breaking into smaller PRs

# How Tested
Stress tested in CI: passed 5 conseqcutive runs: https://github.com/Azure/communication-ui-library/actions/runs/1285350217
Another stress run after merge from main: https://github.com/Azure/communication-ui-library/runs/3793755826


packages/react-composites/tests/browser/call/fixture.ts
```diff
@@ -1,122 +1,32 @@
...
-      const pages = await Promise.all(
-        users.map(async (user) => loadPageWithPermissionsForCalls(testBrowser, serverUrl, user))
-      );
-      await use(pages);
-    },
-    { scope: 'worker' }
-  ]
+export const test = base.extend<unknown, CallWorkerFixture>({
+  /** @returns string URL for the server. */
+  serverUrl: [createTestServer({ appDir: path.join(__dirname, 'app'), serverUrl: SERVER_URL }), { scope: 'worker' }],
```
prprabhu-ms, line 22: nitty nit: I consider paths derived from `__dirname` constants, so like to see them up top: 

    const APP_DIR = path.join(__dirname, 'app');

packages/react-composites/tests/browser/common/defaults.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+export interface WorkerFixture<IdentityType> {
+  serverUrl: string;
+  testBrowser: Browser;
+  users: IdentityType[];
+  pages: Array<Page>;
+}
+
+export const CONNECTION_STRING = process.env.CONNECTION_STRING ?? '';
```
prprabhu-ms, line 39: nit: I wouldn't mind just throwing here if the connection string isn't set. What's the point of continuing? And throwing here will make the error very obvious.

packages/react-composites/tests/browser/common/fixtureHelpers.ts
```diff
@@ -0,0 +1,193 @@
...
+import { buildUrl } from './utils';
+
+/**
+ * Creates a page to be tested for each participant in a browser page.
+ * To be used in a playwright fixture's 'pages'.
+ */
+// eslint-disable-next-line no-empty-pattern, @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/explicit-function-return-type
+export const usePagePerParticipant = async ({ serverUrl, testBrowser, users }, use) => {
+  const pages = await Promise.all(users.map(async (user) => loadPage(testBrowser, serverUrl, user)));
+  await use(pages);
```
prprabhu-ms, line 26: Isn't it grand that React and Playwright give wildly different meanings to `use`...

packages/react-composites/tests/browser/common/defaults.ts
JamesBurnside: I like that better, will update

prprabhu-ms (approved):

prprabhu-ms:
All the snapshot diffs seem to be trivial :'(

JamesBurnside:
> All the snapshot diffs seem to be trivial :'(

Yeah chat now uses the same browser launch args which includes font-rendering settings so they are small but different in this PR

packages/react-composites/tests/browser/common/defaults.ts
JamesBurnside: updated

alcail (approved):

PR #848, opened 2021-09-29
Add playwright config

JamesBurnside:
# What
Add config for playwright for global settings
* Remove retries for chat tests 

# Why
Small PR for more changes incoming.
Breaking down of #818 

# How Tested
Stress test: https://github.com/Azure/communication-ui-library/actions/runs/1304949253


packages/react-composites/playwright.config.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { PlaywrightTestConfig } from '@playwright/test';
+const config: PlaywrightTestConfig = {
+  timeout: 120000, // Allow for more than the standard 30000. This ensures we get debug stack traces when await functions timeout.
```
alcail, line 6: does it need to be this long though? 2 min is quite a long time :D

alcail (approved):

packages/react-composites/playwright.config.ts
JamesBurnside: Our default timeout for all `waitForX` is still 30s - I just bumped this to 2 minutes to ensure it is well over 30s but the two minutes won't ever be hit (unless we have a bunch of waitForX that each take like 20seconds but that would be a separate problem)

(Underlying issue is sometimes the test itself times out and playwright gives no feedback what it was waiting for - by adding this it now points you to the function that it was waiting for - its a bug in playwright but updating playwright had its own issues)

packages/react-composites/playwright.config.ts
```diff
@@ -0,0 +1,13 @@
...
+  workers: 1, // Ensure tests run sequentially.
```
prprabhu-ms, line 7: It's great to set `workers` here, since all hell will break loose if we run concurrently.

Please expand on the comment to say why we need to run these tests sequentially.

JamesBurnside: Updated 👍

packages/react-composites/playwright.config.ts
JamesBurnside: (updated comments on the config)

PorterNan (approved):

PR #849, opened 2021-09-29
Remove thresholds in tests, enable accidentally excluded meeting tests, disallow .only in config

JamesBurnside:
# What
* Remove thresholds from tests - these should no longer be needed
* Enable all meeting tests (there was an accidental `.only` committed)
  * Improve robustness of meeting tests
* Disallow `.only` being committed in the playwright config

# Why
Cleanup, fixups and Breakdown of #818 

# How Tested
Stress test: https://github.com/Azure/communication-ui-library/actions/runs/1305265609


alcail (approved):

packages/react-composites/playwright.config.ts
```diff
@@ -5,6 +5,7 @@ import { PlaywrightTestConfig } from '@playwright/test';
 const config: PlaywrightTestConfig = {
   timeout: 120000, // Allow for more than the standard 30000. This ensures we get debug stack traces when await functions timeout.
   workers: 1, // Ensure tests run sequentially.
+  forbidOnly: !!process.env.CI, // Do not allow `.only` to be committed to the codebase. `.only` should only be used for diagnosing issues.
```
prprabhu-ms, line 8: very very cool!

prprabhu-ms (approved):

PR #850, opened 2021-09-29
UI test build perf: Use esbuild-loader in webpack config and remove source maps

JamesBurnside:
# What
Just for the UI test apps:
* Use esbuild-loader (should be faster than ts-loader)
  * Would be better to fully use esbuild where possible but this is a much simpler one-line change
* Don't create source maps (I can set this to only not create source maps in CI if there's value to having it locally - but I don't think many devs are debugging the UI test apps and speed improvements are preferable)

# Why
Help improve UI test build time

# How Tested
tbd: post timings from PR here
(Locally went from 2min 36s -> 1min 20s)


JamesBurnside:
Turns out this is slower in CI 🤷‍♀️ Closing.

PR #851, opened 2021-09-29
UI tests: Add missing awaits, remove testBrowser, other cleanup

JamesBurnside:
# What
Extra cleanup from #818 
* Add awaits where they are missing
* Unify uses of loadPage, loadPageWithPermissions, & loadUrlInPage 
  * Goal is to remove the individual uses of these in test files in a future PR
* Fix casing in `useFrLocale`
* Remove spawning our own browser and use the playwright setup one
* Other function simplification

# How Tested
Stress test results tbd: https://github.com/Azure/communication-ui-library/actions/runs/1310051854


alcail (approved):

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -94,35 +94,30 @@ test.describe('Chat Composite E2E Tests', () => {
...
+    await page1.waitForSelector(`[data-ui-status="seen"]`);
+    await page1.reload({ waitUntil: 'networkidle' });
+    await waitForChatCompositeToLoad(page1);
     // Fixme: We don't pull readReceipt when initial the chat again, this should be fixed in composite
-    await page.waitForSelector(`[data-ui-status="delivered"]`);
-    stubMessageTimestamps(page);
-    expect(await page.screenshot()).toMatchSnapshot('rejoin-thread.png');
-  });
-
-  test.afterAll(async ({ pages }) => {
```
PorterNan, line 115: We might still need to keep this afterAll - sometimes it just doesn't clean up well, and the process will stay in background without notice, that could cause some memory issues if we don't close all the pages

PorterNan: Ohhh I got why you remove this, then we should find a way to close all pages after all the test files running

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -39,15 +39,8 @@ export const waitForChatCompositeToLoad = async (page: Page): Promise<void> => {
...
-  await page.waitForFunction(
-    (args) => {
-      const callButton = document.querySelector(args.startCallButtonSelector);
-      const callButtonEnabled = callButton && callButton.ariaDisabled !== 'true';
-      return callButtonEnabled;
-    },
-    { startCallButtonSelector: dataUiId('call-composite-start-call-button') }
-  );
+  const startCallButton = await page.waitForSelector(dataUiId('call-composite-start-call-button'));
+  await startCallButton.waitForElementState('enabled');
```
PorterNan, line 43: Nice! I don't know about this function

packages/react-composites/tests/browser/common/fixtureHelpers.ts
```diff
@@ -153,41 +153,43 @@ export const createMeetingUsers =
...
+ * @param url URL to a running test app.
  * @returns
  */
-const loadUrlInPage = async (page: Page, serverUrl: string, user: ChatUserType | CallUserType): Promise<Page> => {
+export const loadNewPage = async (browser: Browser, url: string): Promise<Page> => {
+  const page = await browser.newPage();
   await page.setViewportSize(PAGE_VIEWPORT);
-
-  const url = buildUrl(serverUrl, user);
-  await page.goto(url, { waitUntil: 'load' });
```
PorterNan, line 164: Are we all good without option: waitUntil: 'load'
Is it a default option of page.goto ?

PorterNan (approved):

packages/react-composites/tests/browser/common/fixtureHelpers.ts
JamesBurnside: Yeah `load` is the default

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
JamesBurnside: My assumption is once the process ends the browser contexts will be disposed - but can't find documentation on this either way. Did you hit an error before where the browser/pages weren't being killed?

packages/react-composites/playwright.config.ts
```diff
@@ -21,6 +26,7 @@ const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
 
   use: {
+    viewport: DESKTOP_VIEWPORT,
```
JamesBurnside, line 29: Doesn't appear to be setting the page size judging by the snapshots, investigating

JamesBurnside: Needed to use the playwright launched browser and not spawn our own browser

JamesBurnside: When we do browser.newPage it spawns a new context that doesn't retain the viewport size, seems like a bug, updating playwright version to see if its fixed - and to give access to context in the fixture file that they added so we can use the viewport size in that

JamesBurnside: Yep, updating to 1.15.1 fixed the issue

PR #853, opened 2021-09-29
Try fix Localization call test flakiness

JamesBurnside:
# What
Try make call localization test more robust.
Wipe previous uses in a call and don't enable video for test.

# Why
Sometime this is happening:
image

# How Tested
Needs stress tested when #852 completes

PR #855, opened 2021-09-29
Try fix localization test flakiness take 2

JamesBurnside:
# What
#853 fixed that test, but is causing other tests to be more flakey: https://github.com/Azure/communication-ui-library/runs/3747592784
Try putting the call locale test in its own call instead to ensure its isolated.

My assumption here is that these tests aren't running deterministically sequentially.

# How Tested
Running stress test: https://github.com/Azure/communication-ui-library/actions/runs/1288002656

### DIFF
--- a/packages/react-composites/tests/browser/call/Localization.test.ts
+++ b/packages/react-composites/tests/browser/call/Localization.test.ts
@@ -4,21 +4,22 @@
 import { test } from './fixture';
 import { loadPageWithPermissionsForCalls, waitForCallCompositeToLoad, loadCallScreen } from '../common/utils';
 import { expect } from '@playwright/test';
+import { v1 as generateGUID } from 'uuid';
 
 test.describe('Localization tests', async () => {
-  test.beforeEach(async ({ pages }) => {
-    for (const page of pages) {
-      // Ensure any previous call users from prior tests have left the call
-      await page.reload();
-    }
-  });
-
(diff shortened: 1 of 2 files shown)


packages/react-composites/tests/browser/call/Localization.test.ts
```diff
@@ -4,21 +4,22 @@
...
-    }
-  });
-
   test('Configuration page title and participant button in call should be localized', async ({
     serverUrl,
     users,
     testBrowser
   }) => {
-    const page = await loadPageWithPermissionsForCalls(testBrowser, serverUrl, users[0], { useFrlocale: 'true' });
+    // Run this test in an isolated thread to ensure the test is totally isolated.
```
prprabhu-ms, line 15: Please add a comment here explaining why this test is so special that it needs its own call (if you know).

Feel free to link to the PR that didn't work and state that this is a hail mary.

prprabhu-ms (approved):

packages/react-composites/tests/browser/call/Localization.test.ts
JamesBurnside: Didn't help either way, still failed with the same error. Closing this PR. https://github.com/Azure/communication-ui-library/pull/855/checks?check_run_id=3747775895

JamesBurnside:
This.. didn't help, running out of options...

PR #856, opened 2021-09-29
Link to internal docs

prprabhu-ms:


alkwa-msft (approved):

alkwa-msft (changes requested):
I don't think its a good idea to talk about internal documentation.

Could we generalize these instructions and maintain the actual beta validation instructions in our internal wiki we use to validate our own beta packages?

JamesBurnside (approved):

prprabhu-ms:
It's fairly common to link to some extra "internal only" stuff from public docs (chromium does this in places where it makes sense). It's important that somebody following the public docs isn't confused by it / isn't blocked because they can't access internal links.

I'm fine with keeping this out (I just wanted a more streamlined release guidance). I do like that all the other docs are here in the public repo.

PR #857, opened 2021-09-29
removed button role re-definition to keep default

alcail:
# What
Start button in Chat sample Home page had its role defined as "main" so screen reader was announcing 'Start chat Main Landmark' instead of 'Start chat Button'.
Just removed this definition to keep the default one ("button")

# Why
bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2582638

# How Tested
ran Chat sample locally, opened Narrator , navigated to "Start chat" button and listen to screen reader

# Process & policy checklist

### DIFF
--- a/samples/Chat/src/app/HomeScreen.tsx
+++ b/samples/Chat/src/app/HomeScreen.tsx
@@ -114,7 +114,6 @@ export default (): JSX.Element => {
             </Stack>
             <PrimaryButton
               id="startChat"
-              role="main"
               aria-label="Start chat"
               className={buttonStyle}
               onClick={() => {


prprabhu-ms (approved):
Do we know why we had an explicit role set earlier?

JamesBurnside (approved):

prprabhu-ms:
> Do we know why we had an explicit role set earlier?

Nope, no history: 
image

alkwa-msft (approved):

alcail:
> Do we know why we had an explicit role set earlier?

Not sure why but it's wrong :D

PR #858, opened 2021-09-29
Horizontal Participant Gallery

anjulgarg:
# What
Adds a horizontal paginated participant gallery in video gallery.
Figma - https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/ACS-UI---POR?node-id=2896%3A17175

https://user-images.githubusercontent.com/9782747/135361480-5b564df3-1f9e-4501-a4bc-b3bef13f00bc.mp4

# Why
Part of v2 gallery design

# How Tested

# Process & policy checklist


packages/react-components/src/components/VideoGallery/VideoGallery.tsx
```diff
@@ -189,106 +233,67 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
+                showMuteIndicator={showMuteIndicator}
+              />
+            </Stack>
+          )}
+        </Stack>
+      </div>
     );
   }
-);
+};
```
JamesBurnside, line 299: This function is over 200 lines long, can we reduce it?

packages/calling-component-bindings/review/calling-component-bindings.api.md
```diff
@@ -123,6 +123,8 @@ activeErrors: ActiveError[];
 activeErrors: ActiveError[];
 }>;
 
+// Warning: (ae-incompatible-release-tags) The symbol "GetSelector" is marked as @public, but its signature references "videoGallerySelector" which is marked as @beta
```
JamesBurnside, line 126: This should be fixed.. but not sure what should happen here, only part of the GetCallingSelector is returning a beta item....

packages/react-components/src/components/utils/responsive.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  return width;
+};
+
+/**
+ * A utility hook for identifying if the current viewport size is is mobile screen.
+ * Returns updated result if parent/window resizes.
+ * @param containerRef - Ref of a parent element whose width will be used to determine the viewport size
+ */
+export const useIsMobileScreen = (containerRef: RefObject<HTMLDivElement>): boolean => {
```
JamesBurnside, line 33: can we abstract terms of `mobile` away and just have `isNarrowContainer` - the width of a screen doesn't actually determine if is a mobile screen.
Maybe even just have a helper function instead of a whole react hook
```
isNarrow = (containerSize) => containerWidth <= NARROW_WIDTH
```

packages/react-components/src/components/VideoGallery/VideoGallery.tsx
anjulgarg: This will be reduced when we merge @mgamis-msft and my layouts...

packages/react-components/src/components/utils/responsive.ts
anjulgarg: `Narrow` becomes too generic a term here. 
I am thinking `smallScreen` should be a better choice here. Thoughts?
For tablets we can use `mediumScreen`

Well that's how bootstrap does it.
`xs` for mobile, `sm` for handheld devices, `md` for low res desktop, etc...

packages/react-components/src/components/styles/HorizontalGallery.styles.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { getTheme, IStyle } from '@fluentui/react';
+
+const theme = getTheme();
```
JamesBurnside, line 6: Question: if someone updates the theme will it be updated? (i.e. does getTheme need called inside the render to prevent it being a stale value)

packages/react-components/src/components/utils/responsive.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * A utility hook for providing the width of a parent element.
+ * Returns updated width if parent/window resizes.
+ * @param containerRef - Ref of a parent element whose width will be returned.
+ */
+export const useContainerWidth = (containerRef: RefObject<HTMLDivElement>): number => {
+  const [width, setWidth] = useState(0);
+  useEffect(() => {
+    const updateWidth = (): void => {
+      const width = containerRef.current?.offsetWidth ?? 0;
```
JamesBurnside, line 15: nit: I think use `clientWidth/Height` instead of offsetWidth/Height (as we don't care if borders change size, and we want to capture size changes if a scrollbar changes the interior sizing)
image

packages/react-components/src/components/styles/HorizontalGallery.styles.ts
anjulgarg: I think it will. In storybook when we change theme, this component updates correctly.
@mgamis-msft could you verify?

packages/react-components/src/components/utils/responsive.ts
JamesBurnside: I like smallScreen etc. (though technically its not `screen` but whatever container we're in.. but screen is probably sufficient - maybe smallContainer? But I don't really like that)

packages/react-components/src/components/utils/responsive.ts
anjulgarg: For this, we need the attribute that gives us the content area including everything. OR we might run into a bug.
Kinda hard to explain in the chat. Can show you in a call if you are interested.

packages/react-components/src/components/utils/responsive.ts
anjulgarg: So for our component, it doesn't know or care about whether it's in a screen or container..
We can stick to a uniform naming assuming that we are talking about size w.r.t screen size.

packages/react-components/src/components/VideoGallery/VideoGallery.tsx
anjulgarg: So looks like we won't be able to shrink the code further even after using the new layouts. 
Do you have any recommendations?
I don't see a way to move out the code as all of it is video gallery specific UI logic.

packages/react-components/src/components/styles/HorizontalGallery.styles.ts
mgamis-msft: No, `getTheme` will not get updates to theme updates. `getTheme` is essentially a constant value.

packages/react-components/src/components/VideoGallery/HorizontalGallery.tsx
```diff
@@ -0,0 +1,208 @@
...
+    leftGutter = 8,
+    rightGutter = 8
+  } = props;
+
+  const containerRef = useRef<HTMLDivElement>(null);
+  const isSmallScreen = useIsSmallScreen(containerRef);
+  const containerWidth = useContainerWidth(containerRef);
+  const [page, setPage] = useState(0);
+  const [maxTiles, setMaxTiles] = useState(0);
+  const [tileHeight, setTileHeight] = useState(120);
```
JamesBurnside, line 61: There's a lot of values thrown around in this file: `120` `160` `88` - are these pixel values and do they represent small and large tiles?
Can we pull them out and rename accordingly, e.g
```
const TILE_SIZE_LARGE = { height: 120; width: 160 }; // Large tile size in rem
```
(convert px to rem^)

packages/react-components/src/components/VideoGallery/HorizontalGallery.tsx
```diff
@@ -0,0 +1,208 @@
...
+      maxHeight: `${tileHeight}px`
+    }),
+    [tileHeight]
+  );
+
+  const maxPages = useMemo(() => {
+    return Math.ceil(participants.length / maxTiles) - 1;
+  }, [maxTiles, participants.length]);
+
+  const changePage = (page: number): void => {
```
JamesBurnside, line 110: nit: this can use standard clamping notation:
```
const changePage = (page: number): void => setPage(Math.min(maxPages, Math.max(0, page)));
```

JamesBurnside: in fact I would remove this from here and later have
`setPage(Math.max(0, page - 1)`
and 
`setPage(Math.min(maxPages, page + 1)`

packages/react-components/src/components/VideoGallery/HorizontalGallery.tsx
```diff
@@ -0,0 +1,208 @@
...
+      minWidth: '1.75rem',
+      minHeight: `${tileHeight}px`,
+      maxWidth: '1.75rem',
+      maxHeight: `${tileHeight}px`
+    }),
+    [tileHeight]
+  );
+
+  const maxPages = useMemo(() => {
+    return Math.ceil(participants.length / maxTiles) - 1;
```
JamesBurnside, line 107: If I'm reading this correctly this isn't maxPages, but maxPageIndex. Max pages wouldn't have the -1?

JamesBurnside: e.g. 7 participants and 3 maxTiles = 3 pages of tiles (this result returns 2 and hence is referring to the index of the page and not the maximum number of pages)

packages/react-components/src/components/VideoGallery/HorizontalGallery.tsx
```diff
@@ -0,0 +1,208 @@
...
+      setPage(maxPages);
+    } else {
+      setPage(page);
+    }
+  };
+
+  const defaultOnRenderParticipants = useMemo(() => {
+    // If user provided a custom onRender function return that function.
+    if (onRenderRemoteVideoTile) {
+      return participants?.map((participant) => onRenderRemoteVideoTile(participant));
```
JamesBurnside, line 123: Could be wrong, but my assumption to `onRenderRemoteVideoTile` would be that is only overrides the rendering of the video tile and hence would still make use of the `participants?.slice` logic? (as it should still have pages of participants)

packages/react-components/src/components/VideoGallery/HorizontalGallery.tsx
```diff
@@ -0,0 +1,208 @@
...
+  return (
+    <div ref={containerRef}>
+      <Stack
+        horizontal
+        horizontalAlign="start"
+        tokens={{ childrenGap: '0.5rem' }}
+        className={mergeStyles(horizontalGalleryContainerStyle, { paddingLeft: leftGutter, paddingRight: rightGutter })}
+      >
+        {maxTiles && page > 0 && !isSmallScreen ? (
+          <DefaultButton
```
JamesBurnside, line 172: Pull these whole default button out into components defined somewhere else on this file, this big chunk should just be muchly simplified:
{showLeftButton && <LeftButton onClick={() => changePage(page - 1)} />}
{defaultOnRenderParticipants}
{showRightButton && <RightButton onClick={() => changePage(page + 1)} />}

packages/react-components/src/components/VideoGallery/VideoGallery.tsx
```diff
@@ -105,6 +114,49 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
+    );
+
+    // Create a map of visibleVideoParticipants for faster searching.
+    // This map will be used to identify overflow participants. i.e., participants
+    // that should be rendered in horizontal gallery.
+    const visibleVideoParticipantsMap = {};
+    visibleVideoParticipants.current.forEach((p) => {
+      visibleVideoParticipantsMap[p.userId] = true;
+    });
+    // @TODO: Can this possibly be done inside HorizontalGallery?
```
JamesBurnside, line 138: Resolve todo unless there's a realistic goal that someone will look at this in the future

packages/react-components/src/components/VideoGallery/VideoGallery.tsx
```diff
@@ -105,6 +114,49 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     return !!(layout === 'floatingLocalVideo' && remoteParticipants && remoteParticipants.length > 0);
   }, [layout, remoteParticipants]);
 
+  const containerRef = useRef<HTMLDivElement>(null);
+  const isMobileScreen = useIsSmallScreen(containerRef);
+  const visibleVideoParticipants = useRef<VideoGalleryParticipant[] | []>([]);
```
JamesBurnside, line 119: Shouldn't need `| []` here

packages/react-components/src/gallery/dominantSpeaker.ts
```diff
@@ -12,32 +12,42 @@ import { VideoGalleryRemoteParticipant } from '../types';
...
  * @param maxTiles - Maximum number of tiles to calculate.
+ * @param maxDominantSpeakers - Maximum number of dominant speakers to calculate.
  * @returns VideoGalleryRemoteParticipant[] {@link @azure/communication-react#VideoGalleryRemoteParticipant}
  */
 export const smartDominantSpeakerParticipants = (
   participants: VideoGalleryRemoteParticipant[],
   dominantSpeakers: Array<string> = [],
   visibleParticipants: VideoGalleryRemoteParticipant[] = [],
-  maxTiles = 4 // For video tiles, 4 is the recommended value by Calling team.
+  maxTiles = 4, // For video tiles, 4 is the recommended value by Calling team.,
```
JamesBurnside, line 22: There's too many parameters to be able to discern what is being passed in, e.g.
image
Can you update this to be a named object instead of 5 separate params

packages/react-components/src/components/utils/responsive.ts
JamesBurnside: This was marked resolved but it is still isMobile everywhere - unresolving

packages/react-components/src/components/VideoGallery/VideoGallery.tsx
JamesBurnside: Large chunks of components should be split out into isolated components, e.g.
```
return shouldFloatLocalVideo()) ?  <VideoGalleryWithFloatingLocalVideo />
 :  <VideoGalleryNAnotherName/>;

```

packages/react-components/src/components/VideoGallery/VideoGallery.tsx
```diff
@@ -105,6 +114,49 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
+  const visibleAudioParticipants = useRef<VideoGalleryParticipant[] | []>([]);
+  const [videoParticipants, setVideoParticipants] = useState<VideoGalleryParticipant[] | []>();
```
JamesBurnside, line 121: this component should be pure and not have inbuilt state - what is this state for?

packages/react-components/src/components/VideoGallery/VideoGallery.tsx
```diff
@@ -189,106 +242,67 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
+              verticalAlign="center"
+              className={gridStyle}
+              grow
+            >
+              {localParticipant && defaultOnRenderLocalVideoTile}
+            </Stack>
+            {defaultOnRenderRemoteParticipants}
+          </GridLayout>
+          {audioParticipants && audioParticipants.length > 0 && (
+            <Stack style={{ minHeight: isMobileScreen ? '6rem' : '8rem', maxHeight: isMobileScreen ? '6rem' : '8rem' }}>
```
JamesBurnside, line 292: moves styles to styles file

packages/react-components/src/components/VideoGallery/VideoGallery.tsx
```diff
@@ -105,6 +114,49 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     return !!(layout === 'floatingLocalVideo' && remoteParticipants && remoteParticipants.length > 0);
```
JamesBurnside, line 114: is there a reason this is a callback and not just a const variable? (with useMemo if we need the memoization optimization)

PR #865, opened 2021-10-01
Mobify call composite config screen

JamesBurnside:
# What
Make call composite screen work with `mobileView=true`

# Why
Work to make composites support mobile view
Reference Figma: https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/ACS-UI---POR?node-id=2753%3A19191

## Still ToDo on configuration screen

Add devices button beside turn off/on video

# How Tested
config-mobified


packages/react-composites/src/composites/CallComposite/ConfigurationScreen.tsx
```diff
@@ -13,54 +13,70 @@ import { getCallingSelector } from '@internal/calling-component-bindings';
...
     <Stack
-      horizontal
-      wrap
-      horizontalAlign="center"
+      horizontal={!mobileView}
+      horizontalAlign={mobileView ? 'stretch' : 'center'}
       verticalAlign="center"
-      tokens={configurationStackTokens}
+      tokens={mobileView ? configurationStackTokensMobile : configurationStackTokensDesktop}
       className={configurationContainer}
```
alcail, line 59: You will need to change the padding and min-width in configurationContainer as they depend on tokens value (compensation of the inner stack due to wrapping)

packages/react-composites/src/composites/CallComposite/StartCallButton.tsx
```diff
@@ -26,7 +27,7 @@ export const StartCallButton = (props: StartCallButtonProps): JSX.Element => {
     <PrimaryButton
       data-ui-id="call-composite-start-call-button"
       disabled={isDisabled}
-      className={buttonStyle}
+      className={css(buttonStyle, props.className)}
```
alcail, line 30: why not using 'mergeStyles' ?

alcail (approved):

packages/react-composites/src/composites/CallComposite/StartCallButton.tsx
JamesBurnside: They are surprisingly identical, one must be a legacy function, updating to mergeStyles to match the rest of the codebase

packages/react-composites/src/composites/CallComposite/ConfigurationScreen.tsx
JamesBurnside: Updated 👍

packages/react-composites/src/composites/CallComposite/styles/LocalPreview.styles.ts
```diff
@@ -1,19 +1,32 @@
...
   height: '18.75rem',
   padding: '0.5rem'
 });
 
+/**
+ * @private
+ */
+export const localPreviewContainerStyleMobile = mergeStyles({
+  ...localPreviewContainerStyle,
+  width: '100%',
```
prprabhu-ms, line 26: For my better understanding: Could you explain why we fill all the space on mobile, but have a fixed size on desktop? And if the explanation is at all non-trivial, consider adding it as comment in the code.

packages/react-composites/src/composites/CallComposite/ConfigurationScreen.tsx
```diff
@@ -13,54 +13,71 @@ import { getCallingSelector } from '@internal/calling-component-bindings';
...
+  titleContainerStyleDesktop,
+  titleContainerStyleMobile
 } from './styles/CallConfiguration.styles';
 import { useLocale } from '../localization';
 
 /**
  * @private
  */
 export interface ConfigurationScreenProps {
+  mobileView?: boolean;
```
prprabhu-ms, line 33: nit: For internal interfaces, I'd prefer all options to be mandatory. And the default values should be set at the top level.

This avoids us sprinkling the code base with default values. Imagine if we wanted the default value to be `true` instead. How many files would need to change? (Or, even if we can't really change the default, it increases the chances of inconsistent default value in the implementation)

prprabhu-ms (approved):
A bunch of my comments aren't necessarily modifications in this PR (just styles being moved around). Feel free to respond to them in comments and opening new PRs (or even skip them for now...my CSS is... weak)

packages/react-composites/src/composites/CallComposite/styles/LocalPreview.styles.ts
JamesBurnside: Desktop viewports could be huge (so local preview could be enormous), mobile viewports are small so we want to fill the space

packages/react-composites/src/composites/CallComposite/ConfigurationScreen.tsx
JamesBurnside: I agree, also you could easily forget to set an import optional property, will update

packages/react-composites/src/composites/CallComposite/styles/CallConfiguration.styles.ts
```diff
@@ -1,25 +1,44 @@
...
+export const configurationContainerStyleDesktop = mergeStyles({
+  ...configurationContainerStyle,
+  padding: '1rem', //half childrenGap from Stack
+  minWidth: '14.5rem' // max of min-width from stack items + padding * 2 = 12.5 + 1 * 2
+});
+
+/**
+ * @private
+ */
+export const configurationContainerStyleMobile = mergeStyles({
```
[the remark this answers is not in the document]
JamesBurnside: Merge styles appears to go into the underlying stlyes of a class and generate a new classname (https://docs.microsoft.com/en-us/javascript/api/merge-styles?view=office-ui-fabric-react-latest#mergeStyles_args_) meaning we can call mergeStyles any number of times? Is there a disadvantage to calling it multiple times? (except minor perf of merging objects)

packages/react-composites/src/composites/CallComposite/styles/CallConfiguration.styles.ts
```diff
@@ -1,25 +1,44 @@
...
+  ...configurationContainerStyle,
+  padding: '0.5rem', //half childrenGap from Stack
+  minWidth: '13.5rem' // max of min-width from stack items + padding * 2 = 12.5 + 0.5 * 2
```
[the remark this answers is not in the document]
JamesBurnside: I was just keeping what was there but given this spans the width on mobile this should be unecessary (or be 100%) - I will test and update

packages/react-composites/src/composites/CallComposite/styles/CallConfiguration.styles.ts
```diff
@@ -1,25 +1,44 @@
...
+export const configurationStackTokensMobile: IStackTokens = {
+  childrenGap: '1rem'
+};
+
+const configurationContainerStyle = {
   height: '100%',
   width: '100% ',
-  padding: '1rem', //half childrenGap from Stack
-  minWidth: '14.5rem', // max of min-width from stack items + padding * 2 = 12.5 + 1 * 2
   minHeight: 'auto'
```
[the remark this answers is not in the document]
JamesBurnside: @alcail maybe you know? Otherwise maybe fluent is setting it to something else?

PR #866, opened 2021-10-01
Add options button to config screen when in mobileview

JamesBurnside:
# What
Add options button beside 

**Note** This will become `DevicesButton` in an upcoming PR with the gear icon

# Why
Give mobile users a way to switch devices

# How Tested

| Desktop | Mobile |
| -- | -- |
| image | image |

options-btn


alkwa-msft:
This looks awesome!

A thought came to mind at which level of granularity are we exposing isMobile that makes sense?
Does it make sense to expose it at

html? ❌ - solve with media queries and css

react? ❌ - similar to html you would solve this at the css level

component? - 🚧  - it could. Another way we could do this is have the composites pass down specific styles if it was isMobile. The component would always be responsive and the composite could pass down styles to make some of the items render/not render.

composite? ✔ - yes we need to differentiate for those that just need it to work on composites 

Not blocking. If we think it makes sense then that sounds great. If we want to do something more complex that would require some changes we have more beta updates to do before we GA :) After GA it would be a breaking change.

alkwa-msft (approved):

JamesBurnside:
> This looks awesome!
> 
> A thought came to mind at which level of granularity are we exposing isMobile that makes sense? Does it make sense to expose it at
> 
> html? ❌ - solve with media queries and css
> 
> react? ❌ - similar to html you would solve this at the css level
> 
> component? - 🚧 - it could. Another way we could do this is have the composites pass down specific styles if it was isMobile. The component would always be responsive and the composite could pass down styles to make some of the items render/not render.
> 
> composite? ✔ - yes we need to differentiate for those that just need it to work on composites
> 
> Not blocking. If we think it makes sense then that sounds great. If we want to do something more complex that would require some changes we have more beta updates to do before we GA :) After GA it would be a breaking change.

Currently we only expose it at the composites layer; components should never have a global-style `isMobile` prop - they should always have specific props e.g. `increasedTouchTargets` or (as you say) allow styles to be set to make it mobile friendly

alcail (approved):

prprabhu-ms (approved):
This is awesome!

Also, the video PR descriptions rock. Need to learn this superpower.

PR #868, opened 2021-10-01
Add tsdoc for calling adapter method

PorterNan:
# What

Add tsdoc for calling adapter method

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -124,18 +124,108 @@ export type CallEndedListener = (event: { callId: string }) => void;
...
+   * This method will stop rendering a local camera view when the call is not active
+   *
+   * @public
+   */
   stopCamera(): Promise<void>;
+  /**
+   * Switch between startCamera and stopCamera according to current camera state
+   * This method will stop rendering a local camera view when the call is not active
+   *
+   * @param options - Options to control how video streams are rendered {@link @azure/communication-calling/#VideoStreamOptions }
```
PorterNan, line 161: Looks like this one is a little bit overlap with start/stop camera method, I would like to remove this one from adapter and use start/stop camera under the hook for implementation, what do you guyz think

alkwa-msft (approved):

alcail (approved):

PR #870, opened 2021-10-01
Add tsdoc to chat adapter methods

PorterNan:
# What

Add tsdoc to chat adapter methods

# Why

# How Tested

# Process & policy checklist


alkwa-msft (approved):

alcail (approved):

packages/react-composites/src/composites/ChatComposite/adapter/ChatAdapter.ts
```diff
@@ -68,20 +96,62 @@ export interface ChatAdapterThreadManagement {
  * @public
  */
 export interface ChatAdapterSubscribers {
+  /**
+   * Subscribe function for messageReceived event.
```
prprabhu-ms, line 100: Here and below, perhaps quote the event name to hint that it's a string:

```
Subscribe function for 'messageReceived' event.
```

prprabhu-ms (approved):

PR #871, opened 2021-10-01
Fix huge media gallery in meetings when a participant is screen sharing

JamesBurnside:
# Issue
If there any many participants, when another user shares their screen the media gallery becomes huge (height of all participant tiles) instead of overflowing the participant container:
image

# Fix
Tbh I'm not certain why this fixes the issue, I believe it allows the divs height to be correctly set and thus when a child does `height:100%` it correctly fills the height. This will be changed in galleryV2 so I don't wish to spend a lot more time investigating alternatives.

Screenshot after fix applied:
image


alkwa-msft:
I think this is good for now. Lets make sure if we do pass in styles from the top layer of meeting composite "one day", we merge the developer's styles with our styles in the composite layer.

If we can fix this. This will unblock us. I just tried just the call composite and it just crashed for some other instances. I will create a bug and we can look into closing this as a separate work item.

alkwa-msft (approved):

alcail (approved):

packages/react-composites/src/composites/MeetingComposite/styles/MeetingCompositeStyles.ts
```diff
@@ -6,4 +6,15 @@ import { IStackStyles } from '@fluentui/react';
 /**
  * @private
  */
-export const compositeOuterContainerStyles: IStackStyles = { root: { width: '100%' } };
+export const compositeOuterContainerStyles: IStackStyles = {
+  root: {
+    width: '100%',
+
+    // This allows the composite to correctly contain the call composite by calculating a height for the available space.
```
prprabhu-ms, line 13: <3 <3 <3 for this comment.

Can I frame this comment and put it up as required comments for all CSS that I ever approve for the rest of my life?

prprabhu-ms (approved):

PR #875, opened 2021-10-04
Remove screenshare button on mobile composites

JamesBurnside:
# What
Remove screenshot button on mobile call composite and mobile meeting composite

# Why
We want to not show a screen share button when on mobile.

# How Tested
tbd: add screenshots


alkwa-msft (approved):

alcail (approved):

packages/react-composites/src/composites/CallComposite/CallScreen.tsx
```diff
@@ -131,6 +132,18 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     );
   }
 
+  // Do not show screen share button when composite is optimized for mobile unless the developer
```
prprabhu-ms, line 135: This component is already fairly large. Please extract a helper private function:

```

const disableScreehShareIfCallControlVisible = (options: boolean | CallControlOptions | undefined): boolean | CallControlOptions => {
    ...
}
```

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -19,17 +19,23 @@ export interface MeetingCallControlBarProps {
...
   onChatButtonClicked: () => void;
   onPeopleButtonClicked: () => void;
+  mobileView?: boolean;
 }
 
 /**
  * @private
  */
 export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.Element => {
+  const callControlsOptions: CallControlOptions = {
```
prprabhu-ms, line 29: Shouldn't the `CallControlOptions` be exposed from both the `CallComposite` and `MeetingComposite`?


Perhaps override the default value for `participantsButton` to `false`, but expose the options still?

prprabhu-ms (approved):
Overall LGTM.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
JamesBurnside: They should but they weren't added to meetings when they were added to CallComposite. I'll make a future work item for this

PR #876, opened 2021-10-04
Hide labels on call controls on mobile composites 

JamesBurnside:
# What
When mobileView is enabled on call and meeting composites - ensure labels are hidden and buttons are in compact form.

# Why
When composites are in mobile we want a more compact view

# How Tested
image


alkwa-msft (approved):

alkwa-msft:
(Not blocking)
Do we have an idea of how this would scale if we enabled developers to inject in their own items into the call control bar?

If we feel there are too many questions to determine what this API would look like, can we say what is hard for us that we need to figure out first so we could answer this question above?

alcail (approved):

anjulgarg (approved):

PR #883, opened 2021-10-05
Fixed and updated calls of Primary/Default buttons for Chat, Calling and Storybook

alcail:
# What
We were not using the 'onRenderIcon' props of Primary/Default buttons but instead adding icons as child, thus not exploiting the full extend of FluentUI components.
We ended up also having styles issues with button text being defined as 'Text' child.

before
image
after fix/update
image

# Why

# How Tested
Ran Calling and chat samples locally, as well as Storybook, in Light Mode, Dark Mode and High Contrast Mode (+ 400% zoom in)
image
image
image
image
image

# Process & policy checklist


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -184,10 +185,13 @@ export interface JumpToNewMessageButtonProps {
...
   const { onClick } = props;
   return (
-    <PrimaryButton className={newMessageButtonStyle} onClick={onClick}>
-      <Icon iconName="Down" className={DownIconStyle} />
-      {NEW_MESSAGES}
-    </PrimaryButton>
+    <PrimaryButton
+      className={newMessageButtonStyle}
+      styles={buttonWithIconStyles}
+      text={NEW_MESSAGES}
```
JamesBurnside, line 191: (unrelated to this PR) This needs moved to the locale strings

packages/react-composites/src/composites/CallComposite/Error.tsx
```diff
@@ -36,10 +37,13 @@ export function Error(props: ErrorProps): JSX.Element {
...
         <div className={endCallTitleStyle}>{title}</div>
         <Stack horizontal tokens={buttonsStackTokens}>
-          <PrimaryButton className={buttonStyle} onClick={props.rejoinHandler}>
-            <Video20Filled primaryFill="currentColor" className={videoCameraIconStyle} />
-            {rejoinCall}
-          </PrimaryButton>
+          <PrimaryButton
+            className={buttonStyle}
+            styles={buttonWithIconStyles}
+            text={rejoinCall}
```
JamesBurnside, line 43: Also moved to locale strings file (maybe worth filing a bug with these instances in them)

JamesBurnside (approved):

PR #884, opened 2021-10-05
Replace the use of  `CommunicationUserKind` with `CommunicationUserIdentifier` in various constructors

prprabhu-ms:
# What
* Replace the use of `CommunicationUserKind` with `CommunicationUserIdentifier` in constroctors for stateful clients and adapters.

# Why

ARB feedback: `CommunicationUserKind` is a tagged union to be used as *output types* from the library. This helps Contoso to easily write conditional logic for specific variants of identifiers. But for input types, we should use `CommunicationUserIdentifier` so that Contoso does not have to provide the redundant `kind` field.

# How Tested

`rush build`

# Process & policy checklist

Changes public factory function arguments

### DIFF
--- a/samples/Chat/src/app/ChatScreen.tsx
+++ b/samples/Chat/src/app/ChatScreen.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { CommunicationUserIdentifier, CommunicationUserKind } from '@azure/communication-common';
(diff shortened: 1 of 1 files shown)


PorterNan (approved):
Awesome! I guess CommunicationUserKind is compatible to CommunicationUserIdentifier, so it is not a breaking change now?

alkwa-msft (approved):

alcail (approved):

PR #897, opened 2021-10-07
Alcail/a11 bug on sharing focus trap

alcail:
# What
Implemented new design for screen sharing (no more popup window, message in VideoTile and 'Stop" button with Primary style
image
image

# Why
a11y bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2562168
We had a keyboard trap in Modal component due to the absence of closing button. And even with it, user would have been blocked on the popup window and not be able to mute/unmute themselves or turn their camera on/off

# How Tested
ran Calling sample locally. Tested in Light/Dark mode. Used narrator to make sure sharing message was read.

# Process & policy checklist

added optional strings props (of type VideoGalleryStrings) to VideoGallery for Contoso to be able to have their own strings

### DIFF
--- a/packages/storybook/stories/VideoGallery/__snapshots__/VideoGallery.stories.storyshot
+++ b/packages/storybook/stories/VideoGallery/__snapshots__/VideoGallery.stories.storyshot
@@ -43,22 +43,22 @@ exports[`storybook snapshot tests Storyshots UI Components/Video Gallery Video G
             }
           >
             <div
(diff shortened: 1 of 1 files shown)


alcail:
> d the figma, can the icon be updated to be this fainter gray

sure

prprabhu-ms:
Don't think any of my comments are blocking.

The main one is whether `ScreenShareButton` should provide the default styling, the way `EndCallButton` does.

PR #898, opened 2021-10-08
Make lobby its own page in state .. and in code!

JamesBurnside:
# What
* Make Lobby a page in state in the API
* Move code out of CallPage monolith file
  * Lobby into a LobbyPage in the code
  * Move page and error handling into main CallComposite file

# Why
Allows customers to pivot on Lobby as its own page. Better separates our code.

# How Tested
Ran locally, tested:
* Start show configuration page
* Joining call shows spinner
* Joining teams meeting shows lobby screen
* Call screen is joined once call is joined
* Leaving call shows end screen
* Removed from call shows removed from call screen


alkwa-msft:
Super exciting! I remember when it was just something like

`call.state === 'Lobby' && <>Waiting to be admitted</>`

Just want to confirm we are looking into Lobby customization points in a PR at a later date?

PR #953, opened 2021-10-20
[POC] API exploration for CallControls button injection

prprabhu-ms:
A possible API for injecting buttons in the CallControls.

Note that:
* This is a backwards compatible extension of the stable API, so safe to do post GA.
* It preserves the simple way we've added to customize individual buttons (e.g., the `microphneButton` might become a 'boolean | MicrophoneButtonOptions` in the future to allow customization of that button).

Let's take a quick look, but we shouldn't land this right now.


prprabhu-ms:
Also updated sample temporarily to inject a button:
image

prprabhu-ms:
Closing this PR because there is no more discussion to be had for now.

Can always look at history when we're ready to implement this feature.

PR #955, opened 2021-10-20
Add minimum height touch targets to meny flyout items when in mobile view

JamesBurnside:
# What
* Add increasing touch targets to the api of applicable buttons
  * Wasn't sure if we should expose a boolean (way it is currently) or open the styles up more broadly - open to feedback on this
* Update composites to make use of this and increase the touch targets when mobileView is enabled

# Why
For mobile users - touch targets need to be >= 48px

# How Tested
Best is to see the UI updates when generated by PR. Should be nothing more desktop, but increases for mobile.

Locally:
| Options button (before) | Options button (after) |
| -- | -- |
| image | image |

| People flyout (before) | People flyout (after) |
| -- | -- |
| image | image |


alkwa-msft:
To simplify the naming can we just say imply Mobile means they are also using  a touchscreen and we would want to set the minHeight?

I am not sure if we want to expose the implementation details of what mobile means

alkwa-msft:
In terms of what we want to expose between Composite and Component layers.

Composite :

-   isMobile flag - used in composite api
-   designTokens - custom styles that contoso can set, terminology is geared towards end user usage and not css (potentially could be equivalent to multiple css changes)

  
  Components

- isMobile flag - each component will still use isMobile flag as well
- styles - uses FluentUI style styling for components

     
    For the context of this change, we should just be using an isMobile flag and then using mobile friendly css classes based on this prop.

PR #991, opened 2021-10-27
Allow developers to set messages count per load

PorterNan:
- Allow users to set how many number to fetch
- Set default number to 10(match sdk -request size)
# What

# Why

# How Tested

# Process & policy checklist


PorterNan:
Close because we decided to only change internal implementation without an api change, will open another PR

PR #1007, opened 2021-10-29
Fix blank messages being shown in the MessageThread

JamesBurnside:
# What
Update selector to filter insupported messages:
* participantAdd with participants with no display name
* participantRemove with participants with no display name
* topicUpdated system messages

# Why
They were causing blank messages to be shown:
image

# How Tested
After, no blank messages:
| before | after |
| -- | -- |
| image | image |


PorterNan:
Just noticed some snapshots for customMessage scenarioes have been changed (2 custom message into 1) do we know the reason why it happens?

PorterNan:
> Just noticed some snapshots for customMessage scenarioes have been changed (2 custom message into 1) do we know the reason why it happens?

Ohh! that was because some messages get filtered

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -124,9 +128,10 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
             message.type.toLowerCase() === ACSKnownMessageType.text ||
             message.type.toLowerCase() === ACSKnownMessageType.richtextHtml ||
             message.type.toLowerCase() === ACSKnownMessageType.html ||
-            message.type === ACSKnownMessageType.participantAdded ||
-            message.type === ACSKnownMessageType.participantRemoved ||
-            message.type === ACSKnownMessageType.topicUpdated ||
+            (message.type === ACSKnownMessageType.participantAdded && hasValidParticipant(message)) ||
+            (message.type === ACSKnownMessageType.participantRemoved && hasValidParticipant(message)) ||
+            // TODO: Add support for topicUpdated system messages in MessageThread component.
+            // message.type === ACSKnownMessageType.topicUpdated ||
```
PorterNan, line 134: Uncomment until we fix this TODO?

PorterNan (approved):
Looks good! I believe no one would care too much about messages with empty participant array - in the long run, we should do some filters in component, according to the final result of onRenderMessage function, could you open a work item/bug for this so that we can make it better after GA

packages/chat-component-bindings/src/messageThreadSelector.ts
JamesBurnside: We don't want it until the todo is fixed (otherwise displays empty messages), happy to delete the line instead

prprabhu-ms (approved):

alkwa-msft (approved):

mgamis-msft (approved):

PR #1008, opened 2021-10-29
Translate several User Facing Diagnostics to error bar notifications

prprabhu-ms:
# Why
Tell the user when something's wrong.

# How Tested
* Tested some of the notifications via calling sample. Not all though.
* verified UI via storybook.

Known issues / follow up:
* WifiWarning icon is too big. But there is no 16px version.
* Need to deduplicate implementation of permission banners in the calling composite with ErrorBar (follow up PR)


packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -43,8 +43,35 @@ export const errorBarSelector: ErrorBarSelector = createSelector(
     const activeErrorMessages: ActiveErrorMessage[] = [];
 
     // Errors reported via diagnostics are more reliable than from API method failures, so process those first.
-    if (diagnostics?.network.latest.networkReconnect?.value === 3) {
-      activeErrorMessages.push({ type: 'callingNetworkFailure' });
+    if (
+      diagnostics?.network.latest.networkReceiveQuality?.value === 3 ||
```
PorterNan, line 47: What is the meaning of 2 and 3, is it a rating from 0-5?

packages/react-components/src/components/ErrorBar.tsx
```diff
@@ -248,3 +296,38 @@ const errorsToShow = (
     return error.timestamp > dismissal.dismissedAt;
   });
 };
+
+const messageBarType = (errorType: ErrorType): MessageBarType => {
+  switch (errorType) {
+    case 'callNetworkQualityLow':
+    case 'callNoSpeakerFound':
+    case 'callNoMicrophoneFound':
+    case 'callMicrophoneAccessDenied':
```
PorterNan, line 305: I remember we have this error when we don't have the access to microphone, is this one different? Will there be duplicate warnings?

packages/react-components/src/components/ErrorBar.tsx
```diff
@@ -248,3 +296,38 @@ const errorsToShow = (
...
+    case 'callMicrophoneMutedBySystem':
+    case 'callMacOsMicrophoneAccessDenied':
```
PorterNan, line 307: Haha, different messages for different system, good one

packages/react-components/src/theming/icons.tsx
```diff
@@ -78,5 +82,14 @@ export const DEFAULT_COMPONENT_ICONS = {
   MessageEdit: <Edit20Regular />,
   MessageRemove: <Delete20Regular />,
   HorizontalGalleryLeftButton: <ChevronLeft20Regular />,
-  HorizontalGalleryRightButton: <ChevronRight20Regular />
+  HorizontalGalleryRightButton: <ChevronRight20Regular />,
+  errorBarCallNetworkQualityLow: <WifiWarning20Filled />,
+  errorBarCallNoSpeakerFound: <SpeakerMute16Filled />,
```
PorterNan, line 87: Noticed there are some 16 size icons and some 20 size icons - I also noticed it also happens above in existing code(only happens for once), is it on purpose or they don't provide us some 20-size icons

PorterNan (approved):

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -87,6 +87,16 @@
...
+    "stopScreenShareGeneric": "Failed to stop screen sharing",
+    "callNetworkQualityLow": "Network quality is low.",
+    "callNoSpeakerFound": "No speakers or headphones found. Connect an audio device to hear the call.",
+    "callNoMicrophoneFound": "No microphones found. Connect an audio input device.",
+    "callMicrophoneAccessDenied": "Unable to access microphone. Click the lock in the address bar to grant permission to this webpage.",
+    "callMicrophoneMutedBySystem": "You are muted by your system.",
+    "callMacOsMicrophoneAccessDenied": "Unable to access microphone. Grant microphone permission in your macOS privacy settings.",
+    "callLocalVideoFreeze": "Network bandwidth is poor. Your video may appear paused for others on the call.",
+    "callCameraAlreadyInUse": "Unable to access camera. It may already be in use by another application.",
+    "callMacOsCameraAccessDenied": "MacOS is blocking access to your camera. Update your privacy settings to allow this browser to access your camera.",
```
JamesBurnside, line 99: For this and others: I wouldn't put the OS in the user facing error - Just that access is blocked

packages/calling-component-bindings/src/errorBarSelector.ts
JamesBurnside: nit: update this to use the enum (i.e. NewtworkQuality.Poor instead of 2)

JamesBurnside (approved):

JamesBurnside:
For 
> WifiWarning icon is too big. But there is no 16px version.
Al icons should be svgs so should be independant of size / easily resizable. Is the icon poor when scaled down?

packages/react-components/src/components/ErrorBar.tsx
JamesBurnside: From PR desc - is this being handled in a follow up PR?

packages/react-components/src/components/ErrorBar.tsx
prprabhu-ms: Yeah, unfortunate but true.

packages/react-components/src/theming/icons.tsx
prprabhu-ms: We want 16px icons for the message bar where the text is also smaller. I actually need a 16px icon for `WifiWarning`, but don't know how to resize (yet).

packages/react-components/src/localization/locales/en-US/strings.json
prprabhu-ms: Strings from @pereiralex include MacOS specific error strings because some error conditions are specific to that platform.

packages/react-components/src/components/ErrorBar.tsx
prprabhu-ms: I checked that there aren't duplicate warnings after this PR.

If the microphone / camera access is denied right at the beginning, the permission banner shows up, but the UFD doesn't fire (because we fail to enumerate the devices entirely)

If the microphone / camera access is denied once the app has loaded, the UFD fires, but the permission banner doesn't show up.

I'll have a follow up PR to merge the permission banner into the `ErrorBar` component, but the two conditions and different (though their error messages should likely be made to be the same).

packages/calling-component-bindings/src/errorBarSelector.ts
prprabhu-ms: Mad the same mistake in two PRs... :(

PR #1011, opened 2021-10-29
Fix avatar initials not showing beside chat messages

JamesBurnside:
# What
Fix the fact chat icons have no avatar initials

# Why
Not sure how long this has been an issue for....

# How Tested
Ran locally and can now see avatar initials - see updated ui snapshots for change


prprabhu-ms:
From the snapshot udpates, it seems like display names are gone from the participant list?

alkwa-msft (approved):

alkwa-msft:
Should the names also be in the participant list?  I am looking at the screenshots and it looks like they are not there?

alcail (approved):

JamesBurnside:
> Should the names also be in the participant list? I am looking at the screenshots and it looks like they are not there?

Regression in the PR one down in the stack (#1010 ) just fixed and regenerating snapshots

PR #1012, opened 2021-10-29
Fix 5k+ warnings of dev mode in static html app

PorterNan:
# What

Fix 5k+ warnings of dev mode in static html app

The reason why this happens is chrome tries to load those js files with sourcemap tag inside node_modules folder (which they don't actually provide any source map)

In this fix, we embed source map into js file in dev mode, but still keep prod bundle clean
# Why

# How Tested

# Process & policy checklist


JamesBurnside (approved):

samples/StaticHtmlComposites/webpack.config.dev.js
```diff
@@ -0,0 +1,6 @@
+const config = require('./webpack.config.js');
+
+module.exports = {
+  ...config,
+  devtool: 'eval-source-map'
```
prprabhu-ms, line 5: Is it worth adding a comment about what this does. It took quite a while for us to figure out, so let's make it easy for us to remember :)

prprabhu-ms (approved):

alkwa-msft (approved):

alcail (approved):

PR #1014, opened 2021-10-31
Alkwa/compliance banner position relative

alkwa-msft:
# What
We are going to have the compliance banner and any messages take up space and not hover over the video gallery

# Why
The messages would hover and partially cover the video tiles on the top of the gallery

# How Tested
Ran it locally


alkwa-msft:
With no notifications
image

With a notification we see the video gallery have less space since the notification takes up space.
image

When the banner is dismissed the video gallery will take up the banner space as well.

PR #1015, opened 2021-10-31
notification banner should not float on top of video gallery

alkwa-msft:
# What
We are going to have the compliance banner and any messages take up space and not hover over the video gallery

# Why
The messages would hover and partially cover the video tiles on the top of the gallery

# How Tested
Ran it locally.
Join from a Teams client first with your video on
Joined a teams interop call and started recording so I could see a recording banner.
*Should be getting a number of warnings that camera cannot be activated which also reduces the space of the screen below it instead of floating on top.

### DIFF
--- a/change/@internal-react-composites-12562ee3-6774-4a9b-8683-06213c37bc4b.json
+++ b/change/@internal-react-composites-12562ee3-6774-4a9b-8683-06213c37bc4b.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "updating notifications to be relative instead of absolute",
+  "packageName": "@internal/react-composites",
+  "email": "79329532+alkwa-msft@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


alkwa-msft:
With no notifications
image

With a notification we see the video gallery have less space since the notification takes up space.
image

When the banner is dismissed the video gallery will take up the banner space as well.

JamesBurnside (approved):

prprabhu-ms:
(1) Do we want the same for `ErrorBar` as well then?
(2) Non-floating banner was the initial implementation (a few months ago). We decided to float at that time because resizing the video tile was more disruptive. Do we have a reason to believe that resizing is better? Especially because resizing the videogallery could rearrange the tiles due to change in aspect ratio etc.

alcail (approved):

alkwa-msft:
> (1) Do we want the same for `ErrorBar` as well then? (2) Non-floating banner was the initial implementation (a few months ago). We decided to float at that time because resizing the video tile was more disruptive. Do we have a reason to believe that resizing is better? Especially because resizing the videogallery could rearrange the tiles due to change in aspect ratio etc.

I think when we looked at the banner it looked like it was covering over some content.
It could potentially force a resize but considering the amount of height it takes up I don't think it will impact aspect ratio as much as we think?

mgamis-msft (approved):

prprabhu-ms (approved):

alkwa-msft:
Confirmed that we want to keep it floating and we only want it to float over the gallery. It was doing this before so I am just going to close this PR.

PR #1019, opened 2021-11-02
Add long press event to chat messages (part 1)

JamesBurnside:
# What
Add long press event to chat messages triggered after `500ms`

# Why
(Next PR) we will open the chat message context menu on long press

# How Tested
In storybook:
* In dev mode set to mobile view (so its touch events), click and hold triggers console log.
* In non-touch mode - long press does nothing, context menus function as expected.

### DIFF
--- a/packages/react-components/src/components/ChatMessageComponent.tsx
+++ b/packages/react-components/src/components/ChatMessageComponent.tsx
@@ -179,8 +179,8 @@ export const ChatMessageComponent = (props: ChatMessageProps): JSX.Element => {
   );
 
   if (messageRef.current) {
-    // Follow up PR will implement the callback for the long press
     attachLongTouchPressEvent(messageRef.current, () => {
+      // Follow up PR will implement the callback for the long press
       console.log('Long press event!');
     });
   }
--- a/packages/react-components/src/components/ChatMessageComponent.tsx
+++ b/packages/react-components/src/components/ChatMessageComponent.tsx
@@ -180,7 +180,9 @@ export const ChatMessageComponent = (props: ChatMessageProps): JSX.Element => {
 
   if (messageRef.current) {
     // Follow up PR will implement the callback for the long press
-    attachLongTouchPressEvent(messageRef.current, () => alert('long press!'));
+    attachLongTouchPressEvent(messageRef.current, () => {
+      console.log('Long press event!');
+    });
   }
 
   return chatMessage;
--- a/packages/react-components/src/components/ChatMessageComponent.tsx
+++ b/packages/react-components/src/components/ChatMessageComponent.tsx
@@ -2,7 +2,7 @@
 // Licensed under the MIT license.
 
 import { IStyle, mergeStyles, Link, ContextualMenu, DirectionalHint, IContextualMenuItem } from '@fluentui/react';
-import { Chat, Text, ComponentSlotStyle, MoreIcon, MenuProps } from '@fluentui/react-northstar';
+import { Chat, Text, ComponentSlotStyle, MoreIcon, MenuProps, Ref } from '@fluentui/react-northstar';
 import { _formatString } from '@internal/acs-ui-common';
 import { Parser } from 'html-to-react';
 import React, { useMemo, useRef, useState } from 'react';
@@ -22,6 +22,7 @@ import { formatTimeForChatMessage, formatTimestampForChatMessage } from './utils
 import { useIdentifiers } from '../identifiers/IdentifierProvider';
 import { useTheme } from '../theming';
 import { ChatMessage } from '../types';
+import { attachLongTouchPressEvent } from './utils/longPress';
 
 type ChatMessageProps = {
   message: ChatMessage;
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/ChatMessageComponent.tsx
```diff
@@ -150,29 +152,40 @@ export const ChatMessageComponent = (props: ChatMessageProps): JSX.Element => {
...
+        details={
+          message.editedOn ? <div className={chatMessageEditedTagStyle(theme)}>{strings.editedTag}</div> : undefined
+        }
+        positionActionMenu={false}
+        actionMenu={!disableEditing && message.status !== 'sending' && message.mine ? actionMenu : undefined}
+      />
+    </Ref>
   );
+
+  if (messageRef.current) {
```
JamesBurnside, line 181: Next PR: this only applied to `messages.mine`

packages/react-components/src/components/ChatMessageComponent.tsx
```diff
@@ -150,29 +152,40 @@ export const ChatMessageComponent = (props: ChatMessageProps): JSX.Element => {
...
+    attachLongTouchPressEvent(messageRef.current, () => {
```
JamesBurnside, line 182: This will need a useMemo/Callback to prevent this being assigned every render

packages/react-components/src/components/ChatMessageComponent.tsx
JamesBurnside: This will also need a useMemo/Callback to prevent this being assigned every render

packages/react-components/src/components/ChatMessageComponent.tsx
JamesBurnside, line 182: Should this be a higher-order component instead?

packages/react-components/src/components/utils/longPress.ts
```diff
@@ -0,0 +1,66 @@
...
+const LONG_PRESS_TIMEOUT_MS = 500;
+
+// Cancel the long press on movement. If this is too aggressive this should be updated to detect
+// if movement goes out of bounds of the element instead of cancelling on any movement.
+const CANCEL_ON_MOVEMENT = true;
+
+// Singleton that tracks whether there is an active touchstart event.
+let longPressTimerHandle: number | undefined = undefined;
+
+const clearLongPressTimeout = (): void => {
```
prprabhu-ms, line 18: nit: Move these helper functions below `attachLongTouchPressEvent`.

packages/react-components/src/components/utils/longPress.ts
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft Corporation.
```
prprabhu-ms, line 1: Did you look at https://github.com/minwork/use-long-press ?

The benefit is that it seems to have enough usage that it'll have better testing across platforms / browsers etc.

packages/react-components/src/components/utils/longPress.ts
```diff
@@ -0,0 +1,66 @@
...
+// However both iOS and Android allow customization of this behavior.
+// We will default to 500ms and gather feedback on this value.
+const LONG_PRESS_TIMEOUT_MS = 500;
+
+// Cancel the long press on movement. If this is too aggressive this should be updated to detect
+// if movement goes out of bounds of the element instead of cancelling on any movement.
+const CANCEL_ON_MOVEMENT = true;
+
+// Singleton that tracks whether there is an active touchstart event.
+let longPressTimerHandle: number | undefined = undefined;
```
prprabhu-ms, line 16: Don't use globals.

This should be a `use*` function, so that you can use `useRef` instead.

Similar to what the `use-long-press` package does (if we decide to not use that package for some reason).

packages/react-components/src/components/utils/longPress.ts
```diff
@@ -0,0 +1,66 @@
...
+ * The long press is cancelled on movement.
+ *
+ * @returns The passed in element with the long press event attached.
+ *
+ * @private
+ */
+export function attachLongTouchPressEvent(target: HTMLElement, callback: () => void): void {
+  target.addEventListener('touchstart', (e) => {
+    // Disable context menu that would normally show on long press
+    target.addEventListener('contextmenu', disableContextMenu);
```
prprabhu-ms, line 42: Should this be outside this function? (It's a side effect)

I like the idea of a higher order component that does both -- `useLontTouchPressEvent` and `disableContextMenu`.

packages/react-components/src/components/utils/longPress.ts
JamesBurnside: They had an open issue where their peer dependency was tied to react < 17 so I didn't want to bind us further to that issue - but they just fixed it as of yesterday!
image

prprabhu-ms (approved):

packages/react-components/src/components/utils/longPress.ts
JamesBurnside: removed, this is all handled by use-long-press

packages/react-components/src/components/utils/longPress.ts
JamesBurnside: removed, this is all handled by use-long-press

packages/react-components/src/components/utils/longPress.ts
JamesBurnside: Updated to use-long-press, much cleaner, thanks!

PR #1023, opened 2021-11-02
Make lobby screen strings and icons customizable

prprabhu-ms:
Uses the localization and icon injection APIs that we already have.

Added optional `moreDetails` strings to lobby that we don't use by default.


packages/react-composites/src/composites/common/icons.tsx
```diff
@@ -18,6 +18,8 @@ import React from 'react';
  * @public
  */
 export const COMPOSITE_ONLY_ICONS = {
+  lobbyScreenConnectingToCall: <>☕</>,
```
JamesBurnside, line 21: what is the emoji support like on all platforms? should we update this to a fluent icon? e.g. https://iconcloud.design/search/filter/Fluent%20System%20Library/Fluent%20Regular/coffee/b1c9959dd-3821b375b

JamesBurnside (approved):
We need to somehow have lobby screen UI tests...

alcail (approved):

packages/react-composites/src/composites/common/icons.tsx
prprabhu-ms: @pereiralex

PR #1024, opened 2021-11-03
Separate sample tests into to a single folder

PorterNan:
# What

Separate tests into to a single folder
Extract reusable for all samples to do their own automation tests
Get folder structure ready for component+binding tests

This PR is one of 3 small PRs to create tests for component+binding:
**Separate tests into to a single folder (This)**
Create a mini sample for component+binding
Add tests for component+binding
# Why

# How Tested

# Process & policy checklist


alcail (approved):

prprabhu-ms (approved):
Do you plan to also move the e2e tests that are currently sitting in `@internal/react-composites` and their apps? We really should...

PorterNan:
> Do you plan to also move the e2e tests that are currently sitting in `@internal/react-composites` and their apps? We really should...

Yea, I was thinking of it, but it requires more works, I believe we can open a work item and set it as post-GA

PR #1025, opened 2021-11-03
Alcail/chat message contextMenu and editBox accessible through keyboard

alcail:
# What
made Chat message contextMenu and EditBox (textField + buttons) keyboard accessible
added ariaLabel to EditBox button

# Why
bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2631640
In Chat, message contextMenu was not accessible throiugh keyboard.
Once this was fixed, I realized the EditBox component items were not accessible through keyboard, so no possibility of changing the text, cancel or submit the change.
Buttons in EditBox were not read through Navigator, so they were just read as "Button"

# How Tested
ran Chat sample locally and navigated through keyboard

# Process & policy checklist

added ariaLabel in InputBoxButtonProps and strings for EditBox buttons in MessageThreadStrings


JamesBurnside (approved):

packages/react-components/src/components/EditBox.tsx
```diff
@@ -39,8 +39,13 @@ export const EditBox = (props: EditBoxProps): JSX.Element => {
   const { onCancel, onSubmit, initialValue, strings } = props;
   const [textValue, setTextValue] = useState<string>(initialValue);
   const [textValueOverflow, setTextValueOverflow] = useState(false);
+  const editTextFieldRef = React.useRef<ITextField>(null);
   const theme = useTheme();
 
+  useEffect(() => {
+    editTextFieldRef.current?.focus();
+  }, []);
```
prprabhu-ms, line 47: Doesn't lint complain about dependencies here? `editTextFieldRef.current` should be in the dependency array.

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -130,28 +130,30 @@ export type InputBoxButtonProps = {
   onClick: (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
   className?: string;
   id?: string;
+  ariaLabel?: string;
```
prprabhu-ms, line 133: Instead, `InputBoxButtonProps` should inherit from `IButtonProps` and pass through those props to the `IconButton`.

packages/react-components/src/components/styles/InputBoxComponent.style.ts
```diff
@@ -78,14 +78,12 @@ export const textFieldStyle = (errorColor: string, hasErrorMessage: boolean, dis
  */
 export const inputButtonStyle = mergeStyles({
   color: 'grey',
-  cursor: 'pointer',
```
prprabhu-ms, line 81: nit: I would have preferred these style changes to be in a separate PR.

Most of the e2e snapshot changes are due to these, unrelated to the non-trivial code changes elsewhere in this PR.

prprabhu-ms (approved):
LGTM % using `IButtonProps`.

packages/react-components/src/components/EditBox.tsx
alcail: no complain, having an empty array will go through this only when component mount, and that's what we want to be able to have the cursor on the TextTield when EditBox opens

packages/react-components/src/components/styles/InputBoxComponent.style.ts
alcail: This is part of the PR. Styles need to be changed due to using IconButton

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -130,28 +130,30 @@ export type InputBoxButtonProps = {
...
 };
 
 /**
  * @private
  */
 export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
-  const { onRenderIcon, onClick, className, id } = props;
-  const [isMouseOverSendIcon, setIsMouseOverSendIcon] = useState(false);
+  const { onRenderIcon, onClick, ariaLabel, className, id } = props;
+  const [isMouseOverIcon, setIsMouseOverIcon] = useState(false);
```
alkwa-msft, line 141: nit.

Something feels very wrong about us holding onto the state that tells us if we are hovering over the icon.
If this makes sense to everyone else tho then lets land it

alcail: Agree, but was part of component before, so I did not want to touch that. Now that we want the props to be an extension of IButtonProps, I have to revisit this

alcail: the issue here is that we have a complete different icon when overring

alcail: left a question to FluentUI team. Will follow up

packages/react-components/src/components/InputBoxComponent.tsx
alcail: after discussion and the fact that onRenderIcon is not defined the same way between the 2 props due to rendering different icon depending on hovering state, I will not do this inheritance.
I will indeed revisit the API though but in a following PR

PR #1027, opened 2021-11-03
logging the callId whenever we are connected to a call

alkwa-msft:
# What
Logging our CallId into the console

# Why
We need the call Id whenever we want to be able to share issues with the calling team

# How Tested
Locally joined a call and saw that I could see my call id in the console when I connected to a call


alkwa-msft:
Potential new feature suggestion : Event on the adapter based on our current screen?

alcail (approved):

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -52,6 +52,10 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
       adapter.onStateChange((state: CallAdapterState) => {
         const pageTitle = convertPageStateToString(state);
         document.title = `${pageTitle} - ${webAppTitle}`;
+
+        if (state.call?.state === 'Connected') {
```
JamesBurnside, line 56: Is callId different to groupId/teamsMeetingLink? (the information in url query param?)

JamesBurnside (approved):

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -52,6 +52,10 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
...
+          console.log(`CallId: ${state.call.id}`);
```
JamesBurnside, line 57: note: this will log on every adapter state change after the call has connected - is there a nice way to only do this when the call connects?

JamesBurnside: Better event to use: calladapter.on('callIdChanged'

JamesBurnside (changes requested):
update to use adapter.on('callIdChanged'

samples/Calling/src/app/views/CallScreen.tsx
prprabhu-ms: `callIdChanged` is not what you think it is. You are the third person on our team confused about it this morning.

We really need to rethink this event. It's just piping up the `idChanged` event from the headless SDK `call`. Presumably, that fires if the ID of an *ongoing call* changes. I have no idea why that can happen, but definitely not the case we're looking for here -- we just want to know when we joined a call.

I'd recommend: Remove the `callIdChanged` event from adapter interface (can add back later) because it's confusing.
Consider adding a `callJoined` event, or `pageChanged` event, as Alex suggested.

prprabhu-ms: As for logging each time -- you should probably set a `useRef` here, and `console.log` in a `useEffect` bloc whenever that ref changes.

samples/Calling/src/app/views/CallScreen.tsx
alkwa-msft: callId is the internal call, it doesnt have anything to do with the groupid or teamsMeeting link

JamesBurnside:
> Potential new feature suggestion : Event on the adapter based on our current screen?

I agree this would be nice and very easy to add as likely customers will want to pivot off page state. (just a note for anyone looking here, customers are not blocked on this today however, they can subscribe to onStateChanged and do `oldPage !== newState.page`)

JamesBurnside (approved):

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -52,6 +60,8 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
       adapter.onStateChange((state: CallAdapterState) => {
         const pageTitle = convertPageStateToString(state);
         document.title = `${pageTitle} - ${webAppTitle}`;
+
+        setCallId(state?.call?.id);
```
JamesBurnside, line 64: nit: The reason to use useRef instead of setState is this will trigger a re-render (when all we want is to log something new). As this is the sample app wouldn't worry about it teenie optimization

prprabhu-ms (approved):

samples/Calling/src/app/views/CallScreen.tsx
alkwa-msft: Alright cool can make the change.

PR #1032, opened 2021-11-04
Fix remote video streams when screensharing is active

mgamis-msft:
# What

Before it was assumed that all tiles in horizontal gallery should have video off, but when screensharing is on that is not the case. So this refactor is to FIRST render video participants and audio participants as video tiles and THEN assign these tiles to grid and horizontal gallery depending on if screensharing is active or if there are no video participants.

# Why

Fix bug where remote video streams are not showing when someone is screen sharing

# How Tested

Local sample testing:
https://user-images.githubusercontent.com/79475487/140389437-fbd4c7fb-4d16-4acf-9bdd-87ee945bff5a.mp4

# Process & policy checklist


packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -108,6 +97,17 @@ export const horizontalGalleryStyle = (shouldFloatLocalVideo: boolean, isNarrow:
   };
 };
 
+/**
+ * @private
+ */
+export const horizontalGalleryStyle = (isNarrow: boolean): HorizontalGalleryStyles => {
```
prprabhu-ms, line 103: Should this be in `HorizontalGallery.styles.tsx` instead?

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -226,7 +214,7 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
   ]);
 
   const defaultOnRenderVideoTile = useCallback(
-    (participant: VideoGalleryRemoteParticipant, isVideoParticipant?: boolean, style?: CSSProperties) => {
+    (participant: VideoGalleryRemoteParticipant, isVideoParticipant?: boolean) => {
```
prprabhu-ms, line 217: Not for this PR: `defaultOnRenderVideoTile` should be a private free function instead of a `useCallback`.

We call it directly via `Array.map`, so there is no point memoizing to get to a fixed number of arguments.

prprabhu-ms (approved):
This PR does three things:

* Fixes the bug where we didn't render video tiles in Hz gallery
* Style file refactor for Hz gallery
* Adds keys in various arrays for React

I would have preferred three separate PRs for this.

packages/react-components/src/components/styles/VideoGallery.styles.ts
alcail: this is specific for ResponsiveHorizontalGallery inside VideoGallery, so this is placed correctly IMO (same as horizontalGalleryContainerStyle)

alcail (approved):

packages/react-components/src/components/styles/VideoGallery.styles.ts
mgamis-msft: I think it should stay in VideoGallery because this is the styles for the HorizontalGallery specific to VideoGallery

PR #1033, opened 2021-11-04
remove InputBoxButtonProps from public API

alcail:
# What
Removed InputBoxButtonProps from public API
Updated onRenderIcon in Sendbox component to remove InputBoxButtonProps dependency

# Why
This interface should not be exposed in the public API
task https://skype.visualstudio.com/SPOOL/_workitems/edit/2634719

# How Tested
Ran Chat sample and Storybook to make sure Sendbox and EditBox (dependant of InputBox) are working as expected 

# Process & policy checklist

removed InputBoxButtonProps from public API and updated onRenderIcon in SendBox component


prprabhu-ms (approved):
Perfect!
Thank you.

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -121,12 +121,12 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
 /**
- * Props for dispalying a send button besides the text input area.
+ * Props for displaying a send button besides the text input area.
  *
- * @public
+ * @private
  */
 export type InputBoxButtonProps = {
-  onRenderIcon: (props: InputBoxButtonProps, isMouseOverSendIcon: boolean) => JSX.Element;
+  onRenderIcon: (isMouseOverIcon: boolean) => JSX.Element;
```
alkwa-msft, line 129: should we say isHover?

JamesBurnside (approved):

packages/react-components/src/components/InputBoxComponent.tsx
alcail: would be better, sure. Let me change that

PR #1034, opened 2021-11-04
Add a mini examples folder for component+bindings

PorterNan:
# What

Add a mini examples folder for component+bindings

This PR is one of 3 small PRs to create tests for component+binding:
Separate tests into to a single folder
**Create a mini sample for component+binding  (This)**
Add tests for component+binding

image

# Why

# How Tested

# Process & policy checklist


samples/ComponentExamples/src/utils/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export type IdentityType = {
+  userId: string;
+  token: string;
+  endpointUrl: string;
+  displayName: string;
+  threadId: string;
+};
+
+const CONNECTION_STRING = require('../../appsettings.json')['ResourceConnectionString'];
```
PorterNan, line 15: This appsettings.json is to keep the example in a working status before I make the automation test PR (which help me split things into smaller PR) - will replace this part and use URLParam + Server request in next automtaion test PR

alcail (approved):

samples/ComponentExamples/public/index.html
```diff
@@ -0,0 +1,19 @@
+<!-- Copyright (c) Microsoft Corporation. -->
+<!-- Licensed under the MIT license. -->
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
+    <meta name="description" content="Web site created using create-react-app" />
```
prprabhu-ms, line 10: nitty nit: Please change the description.

samples/ComponentExamples/public/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
prprabhu-ms, line 3: nit: Please update the name / short_name

samples/ComponentExamples/src/examples/ComponentExample.tsx
```diff
@@ -0,0 +1,28 @@
...
+import * as React from 'react';
+
+const messageContainerStyle = {
+  height: '8rem',
+  OverflowY: 'scroll',
+  border: 'solid 2px'
+};
+
+export const ComponentExample = (): JSX.Element => {
+  // usePropsFor will get all the pros and handlers to run the component
```
prprabhu-ms, line 13: typo: pros -> props

Also, handlers are also props. So better just say "will get all the props (including handlers) to"

samples/ComponentExamples/README.md
```diff
@@ -0,0 +1,20 @@
+# UI Library Component and Binding Examples 
+
+This folder contains several mini examples of ACS UI Library component and binding.
```
prprabhu-ms, line 3: nit: mini examples -> small examples

samples/ComponentExamples/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## When should I use component instead of composite
+
+If you haven't checked our composite, we strongly recommend you take a look at our composite first. Composites are all-in-one react components which provides you a full ACS experience within several lines of code.
```
prprabhu-ms, line 7: It might be better to just have one line referring to composite, and then link to storybook directly. (avoid duplicating reasoning etc)

samples/ComponentExamples/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "start:app": "webpack-cli serve --mode=development --env development",
+    "test": "",
+    "prettier": "prettier --no-error-on-unmatched-pattern --write --config ../../.prettierrc --ignore-path=../../.prettierignore \"**/*.js\" \"**/*.jsx\" \"**/*.ts\" \"**/*.tsx\"",
+    "prettier:check": "prettier --no-error-on-unmatched-pattern --check --config ../../.prettierrc --ignore-path=../../.prettierignore \"**/*.js\" \"**/*.jsx\" \"**/*.ts\" \"**/*.tsx\"",
+    "lint": "eslint */**/*.{ts,tsx}",
+    "lint:fix": "eslint */**/*.{ts,tsx} --fix",
+    "lint:quiet": "eslint */**/*.{ts,tsx} --quiet"
+  },
+  "dependencies": {
+    "@azure/communication-chat": "1.1.0",
```
prprabhu-ms, line 20: You shouldn't need any of these. They are direct dependencies of `@azure/communication-react`. In fact, you don't want that so that you don't get versino desync.

samples/ComponentExamples/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../../common/config/tsc/tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist/dist-esm",
+    "baseUrl": "./src",
+    "paths": {
+      "@azure/communication-react": ["../../../packages/communication-react/src"]
```
prprabhu-ms, line 7: neat!

But might be worth addinng more paths for the packlets? This still means that we have to `rush build -t communication-react` for hot reloading to work.

prprabhu-ms: See tsconfig.json for samples/Chat.

prprabhu-ms (approved):

samples/ComponentExamples/README.md
PorterNan: I've checked the storybook, looks like we don't put these reasons in storybook yet, I will keep it here for now, but open a work item to add these reasons into storybook (then remove them here)

samples/ComponentExamples/package.json
PorterNan: I took a try in the repo and got this:
image

And this is the dependency folder of the project
image

So it is actually under the subfolder of @azure/communication-react, but sometimes compiler isn't that smart to seek inside of subfolder

In general, we might just need to put all dependencies we directly use in the project into package.json - if the dependencies doesn't satisfy our @azure/communication-react, we should throw a warning like what Create React App does

BTW, in this project, version desync should be fine, which will be handled by rush itself

samples/ComponentExamples/tsconfig.json
PorterNan: I also took a small experiment locally, looks like hot load is working fine - I guess it is handled by packages\communication-react\tsconfig.json - maybe we should remove those paths in chat as well

PR #1035, opened 2021-11-04
Disallow removing Teams user from chat or call

prprabhu-ms:
Can remove an ACS user:
ACSUser

But not Teams user:
TeamsUsr

# Why
Before this PR, the `remove` action seems available, but doesn't have any effect.

# How Tested
* Calling sample
* Chat via storybook


packages/chat-component-bindings/src/chatParticipantListSelector.ts
```diff
@@ -5,24 +5,28 @@ import { getUserId, getDisplayName, getParticipants, ChatBaseSelectorProps } fro
...
 const convertChatParticipantsToCommunicationParticipants = (
   chatParticipants: ChatParticipant[]
-): CommunicationParticipant[] => {
+): ParticipantListParticipant[] => {
   return chatParticipants.map((participant: ChatParticipant) => {
     return {
       userId: toFlatCommunicationIdentifier(participant.id),
-      displayName: participant.displayName
+      displayName: participant.displayName,
+      // ACS users can not remove Teams users.
```
prprabhu-ms, line 19: Havn't really tested this because there is no easy way to join a Teams thread without using the MeetingComposite. There, we use the Calling roster ignoring the chat one.

JamesBurnside (approved):

alcail (approved):

PR #1036, opened 2021-11-04
Edit buttons take a new line when width is narrow

JamesBurnside:
# What
Changes:
| edit (wide width) | edit (narrow width) |
| -- | -- |
| image | image |

# Why
Per figma - message thread edit buttons go on a new line when the width is narrow: https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library---POR?node-id=3263%3A19683

# How Tested
* Tested send box was unchagned (also makes use of this component)
* Verified keyboard navigation


PorterNan (approved):
Nice! Love the responsive design we provide for each component. I noticed the padding/margin looks a little bit different for left and right - is it just because of the word wrap?
image

alkwa-msft (approved):

JamesBurnside:
> Nice! Love the responsive design we provide for each component. I noticed the padding/margin looks a little bit different for left and right - is it just because of the word wrap? image

Yeah coincidence that second line happened to stop there, here's an example without word wrap:
image

alcail (approved):

alcail:
awesome work ❤️

PR #1038, opened 2021-11-04
Fully implement identifier to/from MRI conversions

prprabhu-ms:
It just so happens that none of our composites currently required conversions for identifiers beyond CommunicationUser. But anybody using the stateful API would be surprised / would see errors in interoperability with other types of users.

Fix it.

# Tested by

* unittests
* Verified that I can join a Teams call via calling sample, and both sides see the participants correctly.


prprabhu-ms:
@DominikMe especially looking for your nod.

packages/acs-ui-common/src/identifier.ts
```diff
@@ -8,6 +8,13 @@ import {
   isPhoneNumberIdentifier
 } from '@azure/communication-common';
 
+const COMMUNICATION_USER_PREFIX = '8:acs:';
+const PHONE_NUMBER_PREFIX = '4:';
+const TEAMS_DOD_PREFIX = '8:dod';
```
prprabhu-ms, line 13: These should all include trailing `:`. Add tests that would fail, then fix this.

packages/acs-ui-common/src/identifier.ts
```diff
@@ -37,7 +53,23 @@ export const toFlatCommunicationIdentifier = (id: CommunicationIdentifier): stri
  * @public
  */
 export const fromFlatCommunicationIdentifier = (id: string): CommunicationIdentifier => {
-  // This implementation is currently a hack that only works with ACS identifiers.
-  // TODO: Make `toFlatCommunicationIdentifier` not be lossy so this can reverse the process.
-  return { communicationUserId: id };
+  if (id.startsWith(COMMUNICATION_USER_PREFIX)) {
+    return { communicationUserId: id.substr(COMMUNICATION_USER_PREFIX.length) };
```
JamesBurnside, line 57: How come we previously didn't need to strip out the prefix? (e.g. we call `call?.removeParticipant(fromFlatCommunicationIdentifier(userId))` which will now have a different id being used)

JamesBurnside (approved):

packages/acs-ui-common/src/identifier.ts
prprabhu-ms: `toFlatCommunicationIdentifer` wasn't adding a prefix. Because it was doing a lossy stringification -- just the ID in all cases without any prefix.

packages/acs-ui-common/src/identifier.ts
```diff
@@ -8,6 +8,13 @@ import {
   isPhoneNumberIdentifier
 } from '@azure/communication-common';
 
+const COMMUNICATION_USER_PREFIX = '8:acs:';
```
alcail, line 11: nit: maybe adding a comment on where to find those prefix

alcail (approved):

packages/acs-ui-common/src/identifier.ts
prprabhu-ms: Unfortunately, it's only documented internally.

packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -19,6 +19,8 @@ import { IDS } from '../../common/constants';
 import { isMobile, verifyParamExists } from '../../common/testAppUtils';
 import memoizeOne from 'memoize-one';
 import { IContextualMenuItem, mergeStyles } from '@fluentui/react';
+import { fromFlatCommunicationIdentifier } from '@internal/acs-ui-common';
+import { CommunicationUserIdentifier } from '@azure/communication-signaling';
```
DominikMe, line 23: This should be imported from `@azure/communication-common`, not signaling.

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
```diff
@@ -38,15 +39,15 @@ function App(): JSX.Element {
         setChatAdapter(
           await createAzureCommunicationChatAdapter({
             endpoint: endpointUrl,
-            userId: { communicationUserId: userId },
+            userId: fromFlatCommunicationIdentifier(userId) as CommunicationUserIdentifier,
```
DominikMe, line 42: I see those down-casts a lot. What happens if one passes a `MicrosoftTeamsUserIdentifier`, will we throw somewhere internally?

packages/react-composites/tests/browser/call/app/index.tsx
prprabhu-ms: Yeah, changed in a follow up PR: https://github.com/Azure/communication-ui-library/pull/1039/files#diff-c662184caec88e95fe7a065186766d62d6e684b9dfb12bf6248708e748638f31R4

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
prprabhu-ms: Where we will throw isn't very well defined. We'll throw when we try to access `communicationUserId` field.

That said, we should really accept the general enum as input as well. When we start supporting `MicrosoftTeamsUserIdentifier`, we will have to do that. But that's a backwards compatible change, so I decided to not change the input signature until we have a real need.

PR #1039, opened 2021-11-05
Genearlize `*state.userId` type to be `CommunicationIdentifierKind`

prprabhu-ms:
# Why
* Consistency in our API -- all constructors / factory functions expect `CommunictionUserIdentifier` and all state fields are `CommunicationIdentifierKind`.
* Make our API future proof. Although currently we only ever work with a CommunicationUserIdentifier for the local participant, we might accept other identifiers in the future (especially Teams). While it is backward compatible to accept a broader type of participant in the constructor, changing the state.userId field later would be a backwards incompatible change.
* Reduces the need to convert the identifier between different types in our implementation because of the increased consistency across our interfaces.

# How Tested
* Chat sample
* Calling sample
* `rush test`

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/hooks/useAdaptedSelector.ts
+++ b/packages/react-composites/src/composites/CallComposite/hooks/useAdaptedSelector.ts
@@ -9,8 +9,7 @@ import memoizeOne from 'memoize-one';
 import { useAdapter } from '../adapter/CallAdapterProvider';
 import { CallAdapterState } from '../adapter/CallAdapter';
 import { CallErrors, CallState, CallClientState, DeviceManagerState } from '@internal/calling-stateful-client';
-import { CommunicationUserKind } from '@azure/communication-common';
-
+import { CommunicationIdentifierKind } from '@azure/communication-common';
 /**
  * @private
  */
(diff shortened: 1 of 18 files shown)


packages/react-composites/src/composites/CallComposite/hooks/useAdaptedSelector.ts
```diff
@@ -98,7 +98,8 @@ const memoizeCalls = memoizeOne((call?: CallState): { [key: string]: CallState }
 
 const adaptCompositeState = (compositeState: CallAdapterState): CallClientState => {
   return memoizeState(
-    compositeState.userId,
+    // xkcd: FIXME
```
JamesBurnside, line 101: dckx //

JamesBurnside (approved):
😍

packages/react-composites/src/composites/CallComposite/hooks/useAdaptedSelector.ts
prprabhu-ms: Heh, I made this change halfway through the refactor, and knew it was easy to miss ('cause the types are OK, just unnecessary casting).

And did I forget... :)

alkwa-msft (approved):

PR #1040, opened 2021-11-06
Add UI test for component + bindings

PorterNan:
# What

Add UI test for component + bindings

The last PR of 3 for component + bindings test

# Why

# How Tested

# Process & policy checklist


.github/workflows/ci.yml
```diff
@@ -352,6 +352,49 @@ jobs:
...
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Failed to pass the UI Test. If this PR is for UI change and the error is snapshot mismatch, please add "ui change" label to the PR for updating the snapshot.'
+            })
```
PorterNan, line 396: I wonder if we could extract some common logic here? And make something like a function? @JamesBurnside I guess you are the only expert of us might have some answers for this
But it is not this PR related, will need to open a work item for this

.github/workflows/update-snapshot.yml
```diff
@@ -7,6 +7,79 @@ on:
   workflow_dispatch:
 
 jobs:
+  update_examples_snapshot:
+    name: Update component examples snapshot by label
```
PorterNan, line 11: Common logic here as well, wondering what we could do for future refactoring

samples/ComponentExamples/src/examples/RenderSingleVideoTileExample.tsx
```diff
@@ -69,12 +69,14 @@ export const RenderVideoTileExample = (): JSX.Element => {
     <>
       <h3> Render single video stream in your own component </h3>
       <span>Click camera button to see what happens:</span>
-      <CameraButton {...cameraButtonProps} />
+      <CameraButton {...cameraButtonProps} id="camera-button" />
```
PorterNan, line 72: I noticed IdentifierProvider marked all data-ui-id string to required - which doesn't work very well in this case(we only need to override one component id here)

And also we are exporting it as _IdentifierProvider, anyone knows why we add a _ prefix here?

.github/workflows/ci.yml
```diff
@@ -352,6 +352,49 @@ jobs:
...
+        run: rush build -t component-examples
+      - name: Visual Regression Tests
+        id: visualregressiontests
+        run: |
+          cd samples/tests
+          rushx test:e2e:examples
+        env:
+          CONNECTION_STRING: ${{ secrets.CONNECTION_STRING }}
+      - name: Upload snapshot diff
+        if: ${{ always() && steps.visualregressiontests.outcome == 'failure' }}
```
prprabhu-ms, line 380: What's `always() &&` ?

.github/workflows/ci.yml
prprabhu-ms: GH released workflow sharing ability recently, but I've heard mixed reviews of that so far.

Could we make the `body` here (and in other UI tests) more concrete now? There are multiple kinds of failures that all add the same message.

samples/ComponentExamples/src/examples/RenderSingleVideoTileExample.tsx
prprabhu-ms: It is an `@internal` symbol exported in the API for automation testing purposes.

All `@internal` symbols must start with `_`, per api-extractor. It's `@internal` because we don't make API stability guarantees for it.

@JamesBurnside  @anjulgarg I know I initially requested `_identifierProvider`, but we only use it in `@internal/react-components` and not `@internal/react-composites`. Since we've agreed that generated DOM is not part of our public API, would we be OK removing this entirely and just using the `data-ui-id` attributes directly everywhere?

samples/ComponentExamples/src/utils/getEndpointUrl.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+let endpointUrl: string | undefined;
```
prprabhu-ms, line 4: Don't use global mutable variables.

Let the caller of `getEndpointUrl` cache the value if they want. (and I'd rename this function to `fetchEndpointUrl` to reinforce that there is a `fetch` involved.

samples/ComponentExamples/src/utils/getEndpointUrl.ts
```diff
@@ -0,0 +1,23 @@
...
+    try {
+      const getRequestOptions = {
+        method: 'GET'
+      };
+      const response = await fetch('/getEndpointUrl', getRequestOptions);
+      const retrievedendpointUrl = await response.text().then((endpointUrl) => endpointUrl);
+      endpointUrl = retrievedendpointUrl;
+      return retrievedendpointUrl;
+    } catch (error) {
+      console.error('Failed at getting environment url, Error: ', error);
```
prprabhu-ms, line 17: nit: Errors should be logged _if and only if_ they are not raised any further. Logging and re-raising leads to duplicate log statements for the same error.

samples/ComponentExamples/src/utils/getToken.ts
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT license.
+
+export type UserToken = {
+  expiresIn: number;
+  identity: string;
+  token: string;
+};
+
+// TODO: Need to revisit if cachedUserToken with respect to auto refresh token
+let cachedUserToken: UserToken | undefined = undefined;
```
prprabhu-ms, line 11: Same comment as above. Don't use global mutable variables.

samples/ComponentExamples/src/utils/loadConfigFromUrlQuery.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { IdentityType } from './utils';
+
+export const loadConfigFromUrlQuery = (): IdentityType => {
+  const urlSearchParams = new URLSearchParams(window.location.search);
+  const params = Object.fromEntries(urlSearchParams.entries());
+
```
prprabhu-ms, line 9: See @JamesBurnside method to validate args as you parse them. It is a common error to load the test page with incorrect query args, and having a clear failure message really helps in debugging.

samples/tests/ComponentExamples/ComponentExamples.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    const user = await createChatThreadAndUsers(PARTICIPANTS);
+
+    const qs = encodeQueryData(user[0]);
+    const url = `${SERVER_URL}?${qs}`;
+    await page.goto(url);
+  });
+
+  // only need to make sure it is loading correctly, the rest will be the responsibility of composite tests
+  test('Whether html page is loaded right with 2 composites', async ({ page }) => {
+    expect(await page.waitForSelector('text=Connect components to StatefulClient')).toBeTruthy();
```
prprabhu-ms, line 25: Can you replace these selectors with injected `data-ui-id` tags. The uniform way to access elements helps readability.

samples/tests/ComponentExamples/ComponentExamples.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    expect(await page.waitForSelector('text=Render single video stream in your own component')).toBeTruthy();
+    // Wait for message to finish sending
+    await page.waitForTimeout(1000);
```
prprabhu-ms, line 28: We haven't sent any messages yet? We should avoid `waitForTimeout` in tests. Why is this needed?

samples/tests/ComponentExamples/ComponentExamples.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    await page.type(dataUiId('sendbox-textfield'), 'Hello!');
+    await page.keyboard.press('Enter');
+
+    await page.waitForSelector(`[data-ui-status="delivered"]`);
+
+    await page.addScriptTag({
```
prprabhu-ms, line 34: Please extract an appropriately named function, like our other tests do.

samples/tests/ComponentExamples/ComponentExamples.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      content: `document.querySelector('[data-ui-id=message-timestamp]').innerText='timestamp';`
+    });
+
+    // Start local video using custom start button
+    await page.click(dataUiId('custom-start-button'));
+    await page.waitForFunction(() => {
+      const videoNode = document.querySelector('video');
+      return document.querySelectorAll('video').length === 1 && videoNode?.readyState === 4;
```
prprabhu-ms, line 42: Why `4`? Can you use an enum value here? If not, please document which enum, what value.

samples/ComponentExamples/src/examples/RenderSingleVideoTileExample.tsx
PorterNan: If we haven't decided whether to expose it or not, maybe we could keep it internally - we don't see a need from customer to do automation test on top of our UI SDK for now - maybe it could happen in the future? Then it is the time to improve this provider. For now, I will keep what I have as implementation of tests

.github/workflows/ci.yml
PorterNan: That means it will run even previous steps fail

samples/ComponentExamples/src/utils/getEndpointUrl.ts
PorterNan: That's a code piece copy&paste from demo server(I believe that is for caching), but yea, since caching is not necessary for the test, let me edit this part

samples/tests/ComponentExamples/ComponentExamples.test.ts
PorterNan: change it to HTMLMediaElement.HAVE_ENOUGH_DATA according to api doc from https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState

packages/react-components/src/localization/locales/de-DE/strings.json
```diff
@@ -106,4 +106,4 @@
     "screenIsBeingSharedMessage": "Sie teilen Ihren Bildschirm",
     "screenShareLoadingMessage": "Der Bildschirm von {participant} wird geladen"
   }
-}
\ No newline at end of file
+}
```
prprabhu-ms, line 109: Spurious diffs... I've had these too. The touchdown build removes trailing newlines from these files :(

samples/tests/ComponentExamples/ComponentExamples.test.ts
```diff
@@ -22,35 +24,39 @@ test.describe('Component Examples Test', () => {
...
+      return (
+        document.querySelectorAll('video').length === 1 && videoNode?.readyState === HTMLMediaElement.HAVE_ENOUGH_DATA
+      );
     });
 
     expect(await page.screenshot({ fullPage: true })).toMatchSnapshot('componentExamples-localVideoStarted.png');
 
-    // Stop local video using CameraButton from UI Library
-    await page.click('id=camera-button');
+    // TODO: page.click on cameraButton is causing some flacky test issues(but manual click is fine)
```
prprabhu-ms, line 51: nit: flacky.

Also, it's better to figure this out now rather than add 5s delay to each test. Also, adding timeout can *still* cause flakiness on slow CI bots / probabilistically.

What flakiness do you see?

prprabhu-ms (commented):
Mostly LGTM, but would like to try to remove flakiness before it lands :)

samples/tests/ComponentExamples/ComponentExamples.test.ts
PorterNan: When we click the camera button using .click(), it just doesn't work sometimes - but it works everytime when you manually click on it

PorterNan: Finally get it fixed! it is related to call status, I add an element to track that

prprabhu-ms (approved):

.github/workflows/update-snapshot.yml
JamesBurnside: yml is just the worst

samples/ComponentExamples/src/examples/RenderSingleVideoTileExample.tsx
JamesBurnside: I'm +1 removing it and just having hard-coded data-ui-ids

samples/ComponentExamples/src/utils/loadConfigFromUrlQuery.ts
JamesBurnside: Looks like it's here:
image

samples/tests/ComponentExamples/ComponentExamples.test.ts
```diff
@@ -0,0 +1,63 @@
...
+const PARTICIPANTS = ['Jack'];
+
+const dataUiId = (id: string): string => `[data-ui-id="${id}"]`;
+
+const dataUiStatus = (status: string): string => `[data-ui-status="${status}"]`;
+
+test.describe('Component Examples Test', () => {
+  test.beforeEach(async ({ page }) => {
+    // Go to the starting url before each test.
+    const user = await createChatThreadAndUsers(PARTICIPANTS);
```
JamesBurnside, line 18: this is creating a chtthread and users, is it also provisioning a call? The tests reference calling:
image
If so update this to something that indicates both chat details and call details are being created

JamesBurnside: If it does, in our other tests we have to create a new call GUID every test or we get the tests hanging trying to rejoin the same call (issue is known to SDK team)

anjulgarg:
@PorterNan I am wondering if it would be better to render one component per page.
The edge case I am worried about is when we change a component, we will need to replace the entire page screen shot.
Another component might have an undesired change that we miss upon verifying the whole page manually.
What do you think?

samples/tests/playwright.config.ts
```diff
@@ -15,8 +16,21 @@ const config: PlaywrightTestConfig = {
       name: 'chromium',
       use: {
         browserName: 'chromium',
-        viewport: { width: 1280, height: 720 }
-      }
+        launchOptions: {
+          args: [
+            '--font-render-hinting=none', // Ensures that fonts are rendered consistently.
```
JamesBurnside, line 21: we really need common place for all these things :(

JamesBurnside (approved):
LGTM! Just some small comment left

PorterNan:
> I am wondering if it would be better to render one component per page.
> The edge case I am worried about is when we change a component, we will need to replace the entire page screen shot.
> Another component might have an undesired change that we miss upon verifying the whole page manually.
> What do you think?

We can do it in a separate PR, I was thinking of moving composite tests into this folder as well - and we might want to add more component into the test, we can resolve this one when we are doing this

samples/tests/playwright.config.ts
PorterNan: I am thinking of move composite tests and these tests into a common folder (after GA), then maybe it will be the best time to extract common stuffs together :D

.github/workflows/update-snapshot.yml
prprabhu-ms: JSON was bad, so we invented YML.

JSON ain't so bad anymore ;)

PR #1042, opened 2021-11-08
Add mobileView documentation to storybook

JamesBurnside:
# What
Add some quick documentation about mobileView in storybook

# Why
Have some small documentation about this for GA. We have a larger task to update our storybook that should have more detailed information and images etc. on mobile vs desktop

# How Tested
image


anjulgarg:
Would setting `mobileView: true` change the layout to mobile immediately or start detecting for mobiles and automatically change on desktop and mobile?

packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
```diff
@@ -52,6 +56,18 @@ export const getDocs: () => JSX.Element = () => {
       </Description>
       <Source code={cssSnippet} />
 
+      <Heading>Running in a Mobile browser</Heading>
+      <Description>
+        CallComposite has been designed for both desktop views and mobile views. By default the CallComposite UI is
+        optimized for desktop views and does not do any automatic detection if it is running on a mobile. To have an
```
anjulgarg, line 62: Slightly confusing description. It is not clear if this flag changes the layout to a mobile layout immediately or starts an automatic detection and changes if it detects a mobile.

alkwa-msft:
> Would setting `mobileView: true` change the layout to mobile immediately or start detecting for mobiles and automatically change on desktop and mobile?

Changing automatically could have unintended effects. We are leaving it up to contoso to set it and we have a suggest process about how they can switch it. These changes are more than just responsive css and thats why we need to be careful about what we are commiting to.

We should definitely be open to feedback from customers here and to potentially make it automatic in the future. I think it would be easier if we 100% understand the user intent but if cant then we put ourselves at risk to making a bad experience for someone. The argument can be made that we are making the experience less magic for 100% of everyone but its on contoso to make sure they can achieve that magic moment for their users. We are just helping :)

packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
JamesBurnside: Thanks will update (it changes immediately, we do no detection ever)

anjulgarg: I love that we don't handle the detection part.
One less thing to worry about.

anjulgarg:
> > Would setting `mobileView: true` change the layout to mobile immediately or start detecting for mobiles and automatically change on desktop and mobile?
> 
> Changing automatically could have unintended effects. We are leaving it up to contoso to set it and we have a suggest process about how they can switch it. These changes are more than just responsive css and thats why we need to be careful about what we are commiting to.
> 
> We should definitely be open to feedback from customers here and to potentially make it automatic in the future. I think it would be easier if we 100% understand the user intent but if cant then we put ourselves at risk to making a bad experience for someone. The argument can be made that we are making the experience less magic for 100% of everyone but its on contoso to make sure they can achieve that magic moment for their users. We are just helping :)

I totally support this approach.

anjulgarg (approved):

alkwa-msft (approved):

packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
```diff
@@ -10,6 +10,10 @@ const containerText = require('!!raw-loader!./snippets/Container.snippet.tsx').d
 const customDataModelExampleContainerText =
   require('!!raw-loader!./snippets/CustomDataModelExampleContainer.snippet.tsx').default;
 
+const mobileViewSnippet = `
+<CallComposite options={{ mobileView: true }} />
```
prprabhu-ms, line 14: Does automation compile-check this?

prprabhu-ms (approved):

packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
JamesBurnside: It doesn't but I wanted to keep the snippet ultra small, happy to make this longer if we want the compile check but will require defining an adapter and having imports etc.

PR #1044, opened 2021-11-08
Make lobby icon not be italicized (again)

prprabhu-ms:
Before:
image

After:
image


JamesBurnside (approved):

packages/react-composites/src/composites/common/icons.tsx
```diff
@@ -11,15 +11,25 @@ import {
 } from '@fluentui/react-icons';
 import { ComponentIcons, DEFAULT_COMPONENT_ICONS } from '@internal/react-components';
 import React from 'react';
+import { mergeStyles, Text } from '@fluentui/react';
+
+const CoffeeIcon = (): JSX.Element => <Text className={mergeStyles(coffeeIconStyle)}>☕</Text>;
```
JamesBurnside, line 16: An aria-doNotReadThis thing needing added for screen reader?

anjulgarg (approved):

packages/react-composites/src/composites/common/icons.tsx
alcail: aria-hidden={true}

alcail (approved):

PR #1046, opened 2021-11-08
Alcail/call composite min max

alcail:
# What
added min width/height to Calling composite(global and by page)
* Mobile min is 312px x 336px (19.5rem x 21rem), knowing iPhone SE is 320px x 568px
image

* Desktop min is 480px x 352 px (30rem x 22rem)
image

# Why
no min were defined for our Calling composite
task https://skype.visualstudio.com/SPOOL/_workitems/edit/2630302

# How Tested
ran Call sample and played with min height/width on desktop/mobile configuration

# Process & policy checklist


JamesBurnside (approved):

anjulgarg:
@alcail 
packages/react-composites/tests/browser/meeting/MeetingComposite.test.ts-snapshots/meeting-gallery-screen-with-people-pane-Mobile-Android-Portrait-linux.png

doesn't look right

JamesBurnside (changes requested):
Resetting vote while snapshots are looked at

alcail:
> Resetting vote while snapshots are looked at (e.g. call config screen title no longer centered: image )

@JamesBurnside , yep, saw that and reverted th Stack to Text change and will revisit that later as it's not part of this PR ;)

JamesBurnside:
Any idea why this spacing changed? /is it intended?
image

alcail:
> Any idea why this spacing changed? /is it intended? image

to avoid this issue of parent not being high enough
image

JamesBurnside:
> > Any idea why this spacing changed? /is it intended?
> 
> to avoid this issue of parent not being high enough
> 
> I went from 3.875rem to 4rem 

Ideally we don't set any height and stacks should correctly size everything, did some digging and its because we use dockedBottom as our ControlBar layout which has position:absolute and thus the stack considers the space it should take up to be 0 so doesn't do the automatic sizing. I'll do a subsequent PR after this one that fixes that issue as the changes are fairly significant.

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -146,12 +147,14 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
   return (
     <BaseComposite {...props}>
       <CallAdapterProvider adapter={adapter}>
-        <MainScreen
-          callInvitationUrl={callInvitationUrl}
-          onFetchAvatarPersonaData={onFetchAvatarPersonaData}
-          onFetchParticipantMenuItems={onFetchParticipantMenuItems}
-          options={options}
-        />
+        <div className={props.options?.mobileView ? mainScreenContainerStyleMobile : mainScreenContainerStyleDesktop}>
```
alkwa-msft, line 150: minor nit. can we determine the value of classname into a variable and just use the variable in this prop when we are returning our JSX.element?

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -28,14 +29,21 @@ export interface CallArrangementProps {
...
 /**
  * @private
  */
 export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
   return (
-    <Stack horizontalAlign="center" verticalAlign="center" styles={containerStyles} grow data-ui-id={props.dataUiId}>
+    <Stack
+      horizontalAlign="center"
+      verticalAlign="center"
+      className={props.mobileView ? containerStyleMobile : containerStyleDesktop}
```
alkwa-msft, line 43: minor nit. can we determine the value of classname into a variable and just use the variable in this prop when we are returning our JSX.element?

alkwa-msft (approved):

PR #1048, opened 2021-11-08
Remove local video tile label when screen size is narrow and local tile is floating

JamesBurnside:
# What
Fix issue where name goes out of bounds in localpreview tile:
image

# Why
Per figma - No space when narrow

# How Tested
| desktop | mobile |
| -- | -- |
|image|image|


alcail:
so we don't keep the icons either?
Look fine, just curious

anjulgarg (approved):

JamesBurnside:
> so we don't keep the icons either? Look fine, just curious

Yeah, we want to have a less crowded UI and the mute button shows the current status and should be prominent enough - but definitely something we can try out and get feedback on though

PR #1052, opened 2021-11-09
Fix for preventing video resize when isSpeaker is on

anjulgarg:
# What
For bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2622523

When isSpeaking indicator appears, it changes the VideoTile div width thus resizing its child contents (video)
This happens because videoTile has a width of 100% and adding a border decreases the effective area for its children.

To avoid this, a simple workaround is to have a transparent div overlaying the video with a transparent border at all times. When isSpeaking is on, the border color changes without actually resizing anything.

# Why

# How Tested

# Process & policy checklist


alcail (approved):

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -172,11 +172,9 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
 
   const ids = useIdentifiers();
 
-  const isSpeakingStyles = isSpeaking
-    ? {
-        border: `0.25rem solid ${theme.palette.themePrimary}`
-      }
-    : {};
+  const isSpeakingStyles = {
+    border: `0.25rem solid ${isSpeaking ? theme.palette.themePrimary : 'transparent'}`
```
prprabhu-ms, line 176: Please add a comment here describing why the transparent border is needed, so it's less magic a few months down the line.

prprabhu-ms:
Just making sure these are intentional:

Shifted avatar:
Avatar shifted

ControlBar shifted:
ControlBar shifted

No border to PIP view:
No border

prprabhu-ms (approved):
% double check snapshot diffs.

JamesBurnside:
This uses quite a bit of space on the mobile local preview:
image

Instead of having it always there, we could have the border be a div positioned absolutely on top of a component?

JamesBurnside (changes requested):
I think its using too much space on the local previews and causing large bezel-like borders:
| desktop | mobile |
| -- | -- |
| image | image |

Is it possible to instead have the border be a floating element on top of the feed positioned absolutely? (which technically occludes part of the tile but only the edges so is OK?)

anjulgarg:
Thanks for pointing it out @JamesBurnside  and @prprabhu-ms 
Let me see if there is a better way to do this.

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -172,23 +173,26 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
     <Ref innerRef={videoTileRef}>
       <Stack
         data-ui-id={ids.videoTile}
         className={mergeStyles(
           rootStyles,
-          isSpeakingStyles,
-          { background: theme.palette.neutralLighter, borderRadius: theme.effects.roundedCorner4 },
+          {
+            background: theme.palette.neutralLighter,
+            borderRadius: theme.effects.roundedCorner4
```
JamesBurnside, line 184: nit: it would be nice to see this and the borderRadius below using the same const variable so its clear they're linked and if one is updated, the other is updated

JamesBurnside (approved):
Thanks for updating this!

PR #1053, opened 2021-11-09
Update call end page UI

prprabhu-ms:
Make the end call pages uniform, matching new figma spec:

image

This PR does not yet add the "rejoin call" button. Need to figure out how that will be connected into the adapter.


alkwa-msft (approved):

packages/communication-react/review/communication-react.api.md
```diff
@@ -836,7 +836,6 @@ export const COMPOSITE_LOCALE_ZH_TW: CompositeLocale;
 
 // @public
 export const COMPOSITE_ONLY_ICONS: {
-    failedToJoinCallDueToNoNetworkIcon: JSX.Element;
```
prprabhu-ms, line 839: This API file was generated incorrectly in the PR below this one. This diff will disappear once that lands and I merge it into this PR.

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/styles/NoticePage.styles.ts
```diff
@@ -3,24 +3,31 @@
 
 import { IStyle, IStackTokens } from '@fluentui/react';
 
+/**
+ * @private
+ */
+export const containerStyle: IStyle = {
+  width: '22.5rem'
```
alcail, line 10: call composite min width will be 19.5rem for mobile and 30rem for desktop, so this should be a max-width.
Wondering how this look on iPhone SE (screen width of 20rem)

alcail (approved):

prprabhu-ms:
Snapshots look good now!

PR #1055, opened 2021-11-09
calling sample: Delete dead code

prprabhu-ms:
`CallComposite` now has pages corresponding to these error conditions, and the sample uses those pages instead of showing its own. So all this code is dead code.


JamesBurnside (approved):
Thanks for cleaning this up!

alkwa-msft (approved):

alcail (approved):

PR #1058, opened 2021-11-09
Fix video tile having translucent text

JamesBurnside:
# What
Fix video tile styles to ensure display name text is not translucent

# Why
Only the background should be, currently text is as well:
image

# How Tested
| light theme | dark theme |
| -- | -- |
| image | image |


packages/react-components/src/components/styles/VideoTile.styles.ts
```diff
@@ -58,9 +58,8 @@ export const disabledVideoHint = mergeStyles({
  * @private
  */
 export const videoHint = mergeStyles(disabledVideoHint, {
-  // This will appear on top of the video stream, so no dependency on theme and thus the direct use of default palette
-  backgroundColor: palette.white,
-  opacity: 0.8,
+  // This will appear on top of the video stream, so no dependency on theme and explicitly use a translucent white
+  backgroundColor: 'rgba(255,255,255,0.8)',
```
prprabhu-ms, line 62: Lucky you. I had to jump through some hoops to make the theme color translucent. Can't just add alpha channel there ;)

prprabhu-ms (approved):

alcail (approved):

PR #1059, opened 2021-11-09
VideoTile label overflow text ellipses

mgamis-msft:
# What

VideoTile label overflow text ellipses when text exceeds available space.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2629760

# How Tested

Storybook testing:
https://user-images.githubusercontent.com/79475487/141005853-c5cd1cbf-cde5-4df9-b7eb-b8dc985afa94.mp4

Calling sample testing for mobile to cover case in bug ticket. 
https://user-images.githubusercontent.com/79475487/141006028-f7932060-346d-4aa7-88ff-2295c7cc2720.mp4
No label is shown on VideoTile as designed in Figma
image

# Process & policy checklist


packages/react-components/src/components/VideoTile.tsx
```diff
@@ -13,9 +13,9 @@ import {
   iconContainerStyle,
   overlayContainerStyles,
   rootStyles,
-  tileInfoStackItemStyle,
```
JamesBurnside, line 16: Can this style be deleted now? Or is it used elswhere?

packages/react-components/src/components/styles/VideoTile.styles.ts
```diff
@@ -37,53 +37,66 @@ export const overlayContainerStyles: IStyle = {
...
 };
 
+/**
+ * @private
+ */
+export const tileInfoContainerStyle = mergeStyles({
+  position: 'absolute',
+  bottom: '0.5rem',
+  left: '0.5rem',
+  minWidth: 'calc(100% - 1rem)',
```
JamesBurnside, line 47: is setting minWidth and maxWidth to the same value different than setting just width?

JamesBurnside (approved):

prprabhu-ms:
Is the larger overlay size intended?

image

prprabhu-ms:
> Is the larger overlay size intended?
> 
> image

Err -> smaller (it's larger before this PR)

packages/react-components/src/components/VideoTile.tsx
mgamis-msft: > > Is the larger overlay size intended?
> > image
> 
> Err -> smaller (it's larger before this PR)

Will try to maintain the old padding. Thanks.

packages/react-components/src/components/styles/VideoTile.styles.ts
mgamis-msft: I think setting minWidth and maxWidth is much more strict for the children inside. Let me verify.

packages/react-components/src/components/VideoTile.tsx
mgamis-msft: Yes . This was used for Stack.Item that I removed with style {display: 'flex'} which didn't really do anything.

mgamis-msft:
> > Is the larger overlay size intended?
> > image
> 
> Err -> smaller (it's larger before this PR)

Will try to maintain the old padding. Thanks.

anjulgarg (approved):

PR #1060, opened 2021-11-09
Add a button to rejoin call from call end pages

prprabhu-ms:
image

Figma: https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library---POR?node-id=3689%3A24815
Task: https://skype.visualstudio.com/SPOOL/_workitems/edit/2631574


JamesBurnside:
nit: the button should be slightly lower, from the figma it looks like the distance between icon -> text is equal to the distance from text -> button:
| current | figma |
|-- | -- |
|  image | image |

prprabhu-ms:
> nit: the button should be slightly lower, from the figma it looks like the distance between icon -> text is equal to the distance from text -> button:
> 
> current	figma
> image	image
> Edit: actyually in designs icon -> text is 16-ishpx, and text -> button is 24px: image image
> 
> But yeah button a little lower if possible?

Updated styles.

alcail (approved):

JamesBurnside (approved):

PR #1061, opened 2021-11-09
Alcail/chat composite min h/w

alcail:
# What
added global min width/height to Chat composite 
image

# Why
no min were defined for our Calling composite
task https://skype.visualstudio.com/SPOOL/_workitems/edit/2630303

# How Tested
ran Call sample and played with min height/width on desktop/mobile configuration

# Process & policy checklist


prprabhu-ms:
The `SendBox` is gone in the snapshot update: 
image

That doesn't look right.

alcail:
> The `SendBox` is gone in the snapshot update: image
> 
> That doesn't look right.

Yes, just saw that :(

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -42,6 +43,13 @@ export interface ChatCompositeProps extends BaseCompositeProps<ChatCompositeIcon
  * @public
  */
 export type ChatCompositeOptions = {
+  /**
+   * Choose to use the composite form optimized for use on a mobile device.
+   * @remarks This is currently only optimized for Portrait mode on mobile devices and does not support landscape.
+   * @defaultValue false
+   * @alpha
+   */
+  mobileView?: boolean;
```
JamesBurnside, line 52: We were looking to avoid the mobileView flag on chat composite because we don't have any features that present themself if it is set (i.e. if a developer sees this flag it confusing to set it to not get any different experience).
Currently the chat composite components should be responsive, is there a reason to have a different minWidth on desktop vs mobile?

JamesBurnside: Oh I see its due to the participant pane? We can ignore that in the calculations as its being disabled by default and marked beta for GA.
cc @alkwa-msft from our discussion on this recently

alcail: the participant pane, but if we don't count this pane, I can go with the same width of 19.5rem (just be aware of pane having scrollbar then when opened)

JamesBurnside: Oh I see its due to the participant pane? We can ignore that in the calculations as its being disabled by default and marked beta for GA.
cc @alkwa-msft from our discussion on this recently

packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
```diff
@@ -6,6 +6,16 @@ import { MessageThreadStyles, SendBoxStylesProps, TypingIndicatorStylesProps } f
 
 const MESSAGE_THREAD_WIDTH = '41.25rem';
 
+/**
+ * @private
+ */
+export const chatScreenContainerStyle = mergeStyles({
+  height: '100%',
+  width: '100%',
+  minHeight: '20rem',
```
JamesBurnside, line 15: As the topic header can be hidden through our API, Can this height be smaller?
image

I'm thinking the scenario where someone may want to embed this inside a chat window similar to fbs:
image

(Though seeing them side by side maybe that is a good size?)

JamesBurnside (approved):

packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
alcail: we can go smaller, but message list will be quite small I feel

mgamis-msft (approved):

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -80,13 +81,15 @@ export const ChatComposite = (props: ChatCompositeProps): JSX.Element => {
...
     <BaseComposite {...props}>
       <ChatAdapterProvider adapter={adapter}>
-        <ChatScreen
-          options={options}
-          onFetchAvatarPersonaData={onFetchAvatarPersonaData}
-          onRenderTypingIndicator={onRenderTypingIndicator}
-          onRenderMessage={onRenderMessage}
-          onFetchParticipantMenuItems={onFetchParticipantMenuItems}
-        />
+        <div className={chatScreenContainerStyle}>
```
alkwa-msft, line 84: Is it possible to have the div outside of the BaseComposite?

By changing the order then we can share that to get everything working you need to just drop this composite in and then customers can feel confident they don't need to control the dimensions from inside of the composite.

alkwa-msft (approved):

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
alcail: I will need to change this for Call composite too then :)

PR #1062, opened 2021-11-09
Fix chat message indicator size and "Message Seen" icon

JamesBurnside:
# What
* Update icons from 20px -> 12px
  * (Note: SVGs themselves are 12px x 12px but plus padding inside the Icons is 16px which is why this PR has size 16 icons)
* Update MessageSeen to be the non-filled version

# Why
Per figma designs: https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library---POR?node-id=3263%3A19683
image

# How Tested
image
image


prprabhu-ms:
Why is there a change in the size of PIP stream?

image

prprabhu-ms (approved):

JamesBurnside:
> Why is there a change in the size of PIP stream?
> 

You happened to review just as I merged from main, but before the update-snapshots had run on top of that merge haha, should be gone once that finishes

alcail (approved):

PR #1056, opened 2021-11-09
Bump headless calling dependency to 1.3.1-beta.1

prprabhu-ms:

PR #1064, opened 2021-11-10
Add an api file for finalizing our api changes

PorterNan:
# What

Let's finalize our api changes!

For any changes you wanna make (but haven't been implemented into main, and you want it to be in GA)

1. check in api changes (without implementation into this branch)
2. update api.md
3. copy it outside to this folder

# Why

# How Tested

# Process & policy checklist


review/communication-react.api.md
```diff
@@ -0,0 +1,2130 @@
...
+// @beta
+export interface TransferCallFeatureState {
+    receivedTransferRequests: TransferRequest[];
+    requestedTransfers: Transfer[];
+}
+
+// @beta
+export interface TransferRequest {
+    // (undocumented)
+    targetParticipant: CommunicationUserKind | PhoneNumberKind | MicrosoftTeamsUserKind | UnknownIdentifier;
```
alkwa-msft, line 1965: Is this fine that we are mixing UnknownIdentifier with these Kind types?

alcail (approved):

prprabhu-ms:
Need to pull in https://github.com/Azure/communication-ui-library/pull/1056/files#diff-ec0dc325bdf92a107d64cbdd0904c5aae39b00aac3236fe2bfe0b88ee332471bL386 

It's a result of API changes in latest calling headless beta.

review/communication-react.api.md
```diff
@@ -0,0 +1,2130 @@
...
+    latest: LatestNetworkDiagnostics;
+}
+
+// @public
+export type OnRenderAvatarCallback = (
+userId?: string, options?: CustomAvatarOptions,
+defaultOnRender?: (props: CustomAvatarOptions) => JSX.Element) => JSX.Element;
+
+// @public
+export const OptionsButton: (props: OptionsButtonProps) => JSX.Element;
```
JamesBurnside, line 1566: Do we want this to be DeviceSettingsButton instead of options button?

review/communication-react.api.md
```diff
@@ -0,0 +1,2130 @@
...
+
+// @public
+export type CallCompositePage = 'accessDeniedTeamsMeeting' | 'call' | 'configuration' | 'joinCallFailedDueToNoNetwork' | 'leftCall' | 'lobby' | 'removedFromCall';
+
+// @public
+export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcons> {
+    adapter: CallAdapter;
+    // (undocumented)
+    callInvitationUrl?: string;
+    options?: CallCompositeOptions;
```
PorterNan, line 304: We should do an internal optimization to avoid necessary re-render

review/communication-react.api.md
```diff
@@ -0,0 +1,2130 @@
...
+    privacyPolicy: string;
+    removedFromCallMoreDetails?: string;
+    removedFromCallTitle: string;
+    soundLabel: string;
+    startCallButtonLabel: string;
+}
+
+// @public
+export type CallControlOptions = {
+    compressedMode?: boolean;
```
DominikMe, line 352: `compressed` is a weird word that refers to file compression algorithms usually.
In UI design this is usually called `compact`, can we rename that?

Example: MS Teams
image

review/communication-react.api.md
```diff
@@ -0,0 +1,2130 @@
...
+
+// @public
+export type CallCompositeOptions = {
+    mobileView?: boolean;
```
DominikMe, line 291: It's a general API design technique to avoid boolean flags to prevent future breaking changes.
For example, one can always refactor `x: boolean` to `foo: "x"`. That way this can evolve to `foo: "x" | "y"`. With boolean flags we would have to introduce a new flag and then we would have `x: boolean; y: boolean`. The addition of additional flags then raises the question which true/false combinatorics of them are valid configurations.

With all that preface why bools are hard to evolve:
Should we consider
```typescript
formFactor?: "desktop" | "mobile";
```

DominikMe: The same bool to union type redesign should be considered everywhere where we use bool flags. (Not saying all need to be refactored, but all should get a second look).

review/communication-react.api.md
```diff
@@ -0,0 +1,2130 @@
...
+export interface MeetingAdapterUiState extends Pick<CallAdapterUiState, 'isLocalPreviewMicrophoneEnabled'> {
+    page: MeetingCompositePage;
+}
+
+// @alpha
+export const MeetingComposite: (props: MeetingCompositeProps) => JSX.Element;
+
+// @alpha
+export type MeetingCompositeOptions = {
+    mobileView?: boolean;
```
DominikMe, line 1356: Same comment as https://github.com/Azure/communication-ui-library/pull/1064/files#r750190938

review/communication-react.api.md
JamesBurnside: I like this a lot more, I'll update the mobileView flag to formFactor

JamesBurnside: https://github.com/Azure/communication-ui-library/pull/1095

review/communication-react.api.md
JamesBurnside: Should we avoid the boolean and do `formFactor: 'comfy' | 'compact'`? (maybe default instead of comfy)

JamesBurnside: PR for this change: https://github.com/Azure/communication-ui-library/pull/1096

review/communication-react.api.md
JamesBurnside: Update here: #1095

review/communication-react.api.md
JamesBurnside: Updated here: https://github.com/Azure/communication-ui-library/pull/1084

review/communication-react.api.md
DominikMe: Nope, nice catch. All of them should be the non-kind types.
```typescript
targetParticipant: CommunicationUserIdentifier | PhoneNumberIdentifier | MicrosoftTeamsUserIdentifier | UnknownIdentifier;
```
 => @prprabhu-ms 
If this is a bug in Calling, let's tell them to change it asap while it's still in beta.

DominikMe: It's a bug in calling:
image

review/communication-react.api.md
DominikMe: @JamesBurnside "formFactor" is mobile or desktop, here it's about UI density.
I see that your PR settled on `displayType` which is okay - type is a little generic, have we considered `displayDensity`?

DominikMe:
Rename `localVideoViewOption` to `localVideoViewOptions` (plural)?

PorterNan:
Close this cause we got the approval from ARB

PR #1065, opened 2021-11-10
Remove `@public` from storybook component docs

JamesBurnside:
# What
* Have descriptions set manually instead of using `<Description of=Component />`
* Quickly update participant item canvas defaults to make it nicer when landing on that component:
  * |before | after |
    | -- | -- |
    | image | image |

# Why
Before `@public` and `@link` would show which isn't ideal. Can't find a way for the docs addon to handle these
image

# How Tested
Please see storybook link created by PR and check out the changes.


alkwa-msft (approved):

prprabhu-ms (approved):
<3

PR #1066, opened 2021-11-10
Fix call screen background and call control drop shadow (and call control vertical alignment by side effect)

JamesBurnside:
# What
* Call screen and connecting/lobby screen needed neutralLightAlt background color per figma
* Call controls needed a top drop-shadow
  * To ensure the correct space was taken up I also had to fix how the control bar space is allocated (ensure it is controlled by the Stack and not be a `height: 4rem`).

# Why
Match figma designs (https://skype.visualstudio.com/SPOOL/_workitems/edit/2631979)

# How Tested
Checked connecting screen, lobby page and call gallery in mobile and desktop.
See updated snapshots for changes.


packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -159,13 +160,17 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
...
-      {microphoneButton}
-      {cameraButton}
-      {screenShareButton}
-      {participantButton}
-      {optionsButton}
-      {endCallButton}
-    </ControlBar>
+    <Stack horizontalAlign="center">
+      <Stack.Item>
+        <ControlBar layout="horizontal">
```
JamesBurnside, line 165: Note: this is updated to `horiztonal` from `dockedBottom` so the CallArrangement page correctly allocates space for the video gallery based on the height of this component. This is because `dockedBottom` has `position; absolute` and so it is floating and thus the `grow` property on the video gallery would not take its height into consideration for its sizing.

packages/react-composites/src/composites/CallComposite/styles/CallControls.styles.ts
```diff
@@ -1,19 +1,30 @@
...
+};
+
 /**
  * @private
  */
 export const groupCallLeaveButtonStyle = {
   root: {
     border: '0.125rem',
     borderRadius: 2,
-    marginRight: '.75rem',
```
JamesBurnside, line 16: anyone know why this was here? It caused our buttons to be slightly off-center

packages/react-composites/src/composites/CallComposite/styles/CallPage.styles.ts
```diff
@@ -2,39 +2,20 @@
...
-/**
- * @private
- */
-export const callControlsStyles: IStackItemStyles = {
-  root: {
-    width: '100%'
-  }
-};
+const VIDEO_GALLERY_Z_INDEX = 1;
+const CONTROL_BAR_Z_INDEX = VIDEO_GALLERY_Z_INDEX + 1;
```
JamesBurnside, line 8: Add comment here
```suggestion
// The control bar must be in a higher z-band than the video gallery so the drop shadow appears on top of the video gallery
const CONTROL_BAR_Z_INDEX = VIDEO_GALLERY_Z_INDEX + 1;
```

packages/react-composites/src/composites/CallComposite/styles/CallPage.styles.ts
```diff
@@ -2,39 +2,20 @@
...
+const NOTIFICATION_CONTAINER_Z_INDEX = Math.max(CONTROL_BAR_Z_INDEX, VIDEO_GALLERY_Z_INDEX) + 1;
```
JamesBurnside, line 9: @prprabhu-ms I think because these are sibling components we don't need to have the z-index higher than the media gallery sub component, however what errors/ufd things should I test here to make sure nothing is broken?

packages/react-composites/src/composites/CallComposite/styles/CallControls.styles.ts
```diff
@@ -1,19 +1,30 @@
...
 import { OptionsButtonStyles, ParticipantsButtonStyles } from '@internal/react-components';
 
 const MINIMUM_TOUCH_TARGET_HEIGHT_REM = 3;
 
+/** @private */
+export const controlBarContainerStyles: IStyle = {
+  paddingTop: '0.25rem',
+  paddingBottom: '0.25rem',
+
+  // @TODO: this should be exposed through a custom CallComposite Theme API that extends the fluent theme with semantic values
```
alcail, line 14: nit: rem instead of px?

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prprabhu-ms: Great justification. Please add this as comment in the code.

These kinds of "why" get lost very easily.

packages/react-composites/src/composites/CallComposite/styles/CallControls.styles.ts
alcail: got it. thanks :)

alcail (approved):

packages/react-composites/src/composites/CallComposite/styles/CallPage.styles.ts
prprabhu-ms: You can test the recording/transcription permission banner, and/or the "you're muted" notification.

They're all in the same container.

packages/react-composites/src/composites/CallComposite/styles/CallControls.styles.ts
JamesBurnside: Ah thanks! I will test it out then add this back as padding to the parent container (both left and right sides) to ensure buttons are centered

prprabhu-ms:
Only some of the snapshots show a shadow *inside* the video: 
image

prprabhu-ms (approved):

JamesBurnside:
> Only some of the snapshots show a shadow _inside_ the video: image

That's actually a really weird artifact that seems to come and go in our snapshots but never causes them to fail

packages/react-composites/src/composites/CallComposite/styles/CallPage.styles.ts
JamesBurnside: Confirmed these:
image
image

packages/react-composites/src/composites/CallComposite/styles/CallControls.styles.ts
JamesBurnside: Tested this and has correct behavior in composites and in the storybook

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
JamesBurnside: Updated

PR #1069, opened 2021-11-11
No nav buttons in HorizontalGallery for one page of participants

mgamis-msft:
# What

Changed HorizontalGallery to show page navigation buttons only if number of children greater than childrenPerPage prop. 
Changed ResponsiveHorizontalGallery to first consider if all children can fit in container without buttons to calculate childrenPerPage.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2632983

# How Tested

https://user-images.githubusercontent.com/79475487/141214360-721119d8-3f3b-47fe-9d98-fb3aaaaa22db.mp4

# Process & policy checklist


alkwa-msft (approved):

JamesBurnside (approved):

alcail (approved):

anjulgarg:
@mgamis-msft According to the attached video, there is a left arrow even when the gallery was on the first page.
If that is correct, should we also remove the left arrow if someone is on the first page? That would make it even more intuitive.

anjulgarg (changes requested):
Left arrow shouldn't be visible if the horizontal gallery is showing the first N participants.
Similarly, right arrow shouldn't be visible if the gallery is showing last N participants.

JamesBurnside:
> Left arrow shouldn't be visible if the horizontal gallery is showing the first N participants. Similarly, right arrow shouldn't be visible if the gallery is showing last N participants.

For context on this, the reason why was due to this comment: https://github.com/Azure/communication-ui-library/pull/978#discussion_r735735007 . The button shifts when clicked meaning your mouse cursor is no longer over the button:
image

However, now we've tested when always having them in and seen problems were its non-obvious that there aren't anymore participants perhaps we revert back to the previous issue? I'm happy for either option

anjulgarg:
> > Left arrow shouldn't be visible if the horizontal gallery is showing the first N participants. Similarly, right arrow shouldn't be visible if the gallery is showing last N participants.
> 
> For context on this, the reason why was due to this comment: [#978 (comment)](https://github.com/Azure/communication-ui-library/pull/978#discussion_r735735007) . The button shifts when clicked meaning your mouse cursor is no longer over the button so you can't click it again and it feels confusing: image
> 
> However, now we've tested when always having them in and seen problems were its non-obvious that there aren't anymore participants perhaps we revert back to the previous issue? I'm happy for either option

I understand what you mean. However, I thought we were doing this to get rid of the confusing arrows.
If you look at the current fix, it doesn't really help much since as soon as the number of participants grow, we are back to the same problem (non-obvious left/right buttons).

mgamis-msft:
> > > Left arrow shouldn't be visible if the horizontal gallery is showing the first N participants. Similarly, right arrow shouldn't be visible if the gallery is showing last N participants.
> > 
> > 
> > For context on this, the reason why was due to this comment: [#978 (comment)](https://github.com/Azure/communication-ui-library/pull/978#discussion_r735735007) . The button shifts when clicked meaning your mouse cursor is no longer over the button so you can't click it again and it feels confusing: image
> > However, now we've tested when always having them in and seen problems were its non-obvious that there aren't anymore participants perhaps we revert back to the previous issue? I'm happy for either option
> 
> I understand what you mean. However, I thought we were doing this to get rid of the confusing arrows. If you look at the current fix, it doesn't really help much since as soon as the number of participants grow, we are back to the same problem (non-obvious left/right buttons).

I'm okay to remove right button on the last page but the original bug was simply to remove the buttons when there is only one page of participants. Also, if we make the left button disappear on the first page, are we okay with the shift of the right button after we go to the 2nd page?

> > > Left arrow shouldn't be visible if the horizontal gallery is showing the first N participants. Similarly, right arrow shouldn't be visible if the gallery is showing last N participants.
> > 
> > 
> > For context on this, the reason why was due to this comment: [#978 (comment)](https://github.com/Azure/communication-ui-library/pull/978#discussion_r735735007) . The button shifts when clicked meaning your mouse cursor is no longer over the button so you can't click it again and it feels confusing: image
> > However, now we've tested when always having them in and seen problems were its non-obvious that there aren't anymore participants perhaps we revert back to the previous issue? I'm happy for either option
> 
> I understand what you mean. However, I thought we were doing this to get rid of the confusing arrows. If you look at the current fix, it doesn't really help much since as soon as the number of participants grow, we are back to the same problem (non-obvious left/right buttons).

I'm happy to remove the right nav button if we are on the last page but the bug ticket specifies that I should only remove the buttons when there aren't more than one page of participants.
Also if we have no left button on the first page then are we okay with the right button shift after we click it to get to the second page?
Can we get a concesus here @anjulgarg @JamesBurnside @alkwa-msft @alcail?

JamesBurnside:
> I'm happy to remove the right nav button if we are on the last page but the bug ticket specifies that I should only remove the buttons when there aren't more than one page of participants. Also if we have no left button on the first page then are we okay with the right button shift after we click it to get to the second page? Can we get a concesus here @anjulgarg Anjul Garg FTE @JamesBurnside James Burnside FTE @alkwa-msft Alex Kwan (IC3) FTE @alcail Aline Caillaud (SHE/HER) FTE?

I'm happy with either :) Don't have a strong preference either way

alkwa-msft:
> > I'm happy to remove the right nav button if we are on the last page but the bug ticket specifies that I should only remove the buttons when there aren't more than one page of participants. Also if we have no left button on the first page then are we okay with the right button shift after we click it to get to the second page? Can we get a concesus here @anjulgarg Anjul Garg FTE @JamesBurnside James Burnside FTE @alkwa-msft Alex Kwan (IC3) FTE @alcail Aline Caillaud (SHE/HER) FTE?
> 
> I'm happy with either :) Don't have a strong preference either way

I believe James, Miguel and myself had a chat at the end of the bug bash . Lets keep it simple

If there are not enough participants to require the buttons then don't include them.
If there are enough participants then keep both buttons
There are optimizations that could be done like removing certain buttons if you are at a certain side of the horizontal gallery. Lets defer this one for now

Lets modify the experience over time through customer feedback and internal usage.

alkwa-msft:
I believe James, Miguel and myself had a chat at the end of the bug bash . Lets keep it simple


  *   If there are not enough participants to require the buttons then don't include them.
  *   If there are enough participants then keep both buttons

There are optimizations that could be done like removing certain buttons if you are at a certain side of the horizontal gallery. Lets defer this one for now

Lets modify the experience over time through customer feedback and internal usage.
________________________________

anjulgarg (approved):

PR #1076, opened 2021-11-12
Display "You" instead of complete name in local video tile.

anjulgarg:
# What
Display "You" instead of complete name in local video tile.
Remove text from local video tile in mobile views.
Adds another prop to video tile for setting the initials name in case its different from display name

image

image

image

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -195,7 +195,7 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
           ) : undefined
         }
         showLabel={!(shouldFloatLocalVideo && isNarrow)}
-        displayName={localParticipant?.displayName}
+        displayName={isNarrow ? '' : 'You'}
```
alkwa-msft, line 198: can we make sure the "You" is using a localized value?

alcail (approved):

alcail:
just make sure to use Locale value for "You" ;)

packages/react-components/src/components/VideoGallery.tsx
anjulgarg: Ah, thanks for pointing that out.

JamesBurnside: Changing just display name here to `You` will cause the participant initials to be a `Y` instead of their initials:
image

I think we need to separate the logic of display name and avatar initials inside the video tile

anjulgarg: Haha yes, my memory is failing me.

alkwa-msft (approved):

PR #1077, opened 2021-11-12
Add control bar button tooltips

JamesBurnside:
# What
Add tooltips to control bar buttons, these show on hover and use the Fluent Tooltip host as exampled here under "Icon Button with Tooltip": https://developer.microsoft.com/en-us/fluentui#/controls/web/button
 
image
image
image

# Why
Per designs

# How Tested
Verified in storybook and in CallComposite (local preview screen and call page screen):
image
image

Verified control bar buttons with no tooltip content string set correctly do not show any tooltip (as is the case with the default button):
image


PorterNan (approved):
Let's check this in and froze the api change on next Tuesday👏

packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -115,12 +135,20 @@ export const ControlBarButton = (props: ControlBarButtonProps): JSX.Element => {
...
+  const tooltipId = props.tooltipId ?? props.labelKey ? props.labelKey + '-tooltip' : undefined;
+
   return (
-    <DefaultButton
-      {...props}
-      styles={componentStyles}
-      onRenderText={props.showLabel ? props.onRenderText ?? DefaultRenderText : undefined}
-      onRenderIcon={props.onRenderIcon ?? DefaultRenderIcon}
-    />
+    <ControlButtonTooltip content={tooltipContent} id={tooltipId}>
```
PorterNan, line 145: Is it possible for develoers to disable tooltips? I guess some might wanna do it

JamesBurnside: They are disabled by default and only show if a `tooltipContent` string is supplied

anjulgarg (approved):

anjulgarg:
Any plans of adding UI tests for these?

JamesBurnside:
> Any plans of adding UI tests for these?

That.. is a great question, I assume playwright can handle that, if it can I will add as a seperate PR in case they are flakey and need reverted

JamesBurnside:
> > Any plans of adding UI tests for these?
> 
> That.. is a great question, I assume playwright can handle that, if it can I will try now and add as a seperate PR in case they are flakey and need reverted

UI tests: https://github.com/Azure/communication-ui-library/pull/1079

PR #1078, opened 2021-11-12
Make all button width consistent in desktop mode

PorterNan:
# What

How it looks before:
Button_width

After:
Button_width_after

Also tested under mobile mode - it is still working great!
image

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/styles/CallControls.styles.ts
```diff
@@ -93,3 +93,12 @@ export const optionsButtonWithIncreasedTouchTargets: OptionsButtonStyles = {
     menuItemStyles: buttonFlyoutIncreasedSizeStyles
   }
 };
+
+/**
+ * @private
+ */
+export const controlButtonBaseStyle: IButtonStyles = {
+  label: {
+    width: '2.25rem'
```
PorterNan, line 102: The number we get here is from the longest text we have for those buttons ('Unmute'), so all the buttons will be always the same size, no shrink or grow

alkwa-msft: I am curious is there a language where unmute and mute are wildly different? If its greater than 2.5rem do we truncate or do we expand?

alkwa-msft (approved):

JamesBurnside:
I love the PR desc videos, super informative 👌❤️

packages/react-composites/src/composites/CallComposite/styles/CallControls.styles.ts
JamesBurnside: (Same question as alex if other langs have a longer string)
What if we allowed developers to pass longer strings that they localize by using min-width instead? It would be a worse experience (as buttons would switch size), but more flexible for developers.. not sure if we should or not...

PorterNan: Good question, lemme test "the widest text across all the exisitng locale languages"

Let's say if we have something more than this in a specific button, the button will expand

Will get back to this one when I tested all languages, and also change this to min-widtha

JamesBurnside: Ah dear classic german being long, here is unmute: `Stummschaltung aufheben`

PorterNan: Maybe we should make it multiple line? What is current existing example for long text(if that expand too long, it will be a problem, but I believe one of the main task localization team will do is to "keep text into near length as much as possible"

JamesBurnside: I like the idea of muti-line, I worry our call control bar will get too wide for narrow screens

anjulgarg: +1 to using min-width instead of using width.
+1 to multiline as long as we can enforce height consistency across buttons if one of them has multiline text...

PorterNan: Double checked what Teams is doing - they have a min-width for button so button won't be shrinking(especially when button label is very short in some languages)

Also looks like they are trying to avoid multiple word in languages and also translation (button label won't change but tooltip will be changed, still good for accessibility), which might be interesting to be considered, take French as an example:
image


Anyway this is unrelated to this PR to fix 'button too short', we might need another discussion about this

anjulgarg (approved):

JamesBurnside (approved):

PR #1080, opened 2021-11-13
Add a FakeChatClient

prprabhu-ms:
# What

* Add `FakeChatClient` that implements the ACS chat service, in memory.
* Add a couple example stories using that.

# TODO
* Cleanup exports. Currently, I ended up exposing a way to create stateful chat client that isn't actually available in storybook
* A little bit of documentation
* Update existing stories instead of new ones, if we like this.


packages/fake-backends/src/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    byPage: () => {
+      return listByPage();
+    }
+  };
+};
+
+/**
+ * An iterator that throws the given error when asynchronously iterating over items, directly or byPage.
+ */
+export const failingPagedAsyncIterator = <T>(error: Error): PagedAsyncIterableIterator<T, T[]> => {
```
prprabhu-ms, line 40: Delete

prprabhu-ms:
Closing this until I'm ready to land it.

This is actually in a fairly good state.
My preference would be to use it already for `ChatComposite` stories.

PR #1081, opened 2021-11-14
[DO NOT REVIEW] PoC - two device managers with different video devices, no call agent

prprabhu-ms:
# What

Poc to show that
* `DeviceManager` can be used without ever creating a callAgent
  * Can create video stream
* Two different `CallClient` and associated `DeviceManager` can be created in the same tab
  * Can choose different cameras in the two

image

# Why

This proves that we can create a fake `CallClient` that uses the real `DeviceManager` for all device access, but fakes all call related functionality by simply using the correct `LocalVideoStream` as the remote stream.


prprabhu-ms:
Just an experiment.

PR #1082, opened 2021-11-15
make the button width/height as large as the inner svg width/height

alkwa-msft:
# What
image

# Why
It is an odd experience on ios safari.

# How Tested
Verified on ios15.1 on iphone 13mini


JamesBurnside (changes requested):
Looks to have nudged the icon off-center for desktop:
image

alkwa-msft:
> Looks to have nudged the icon off-center for desktop: image

AlexP's designs make an assumption about a height of 36px in practice we have 38 pixels for our sendbox.
For now I think lets keep everything looking fine at 38 and we can adjust as necessary in the future

alcail:
Approved but the icon moved to the right and seems a bit too close to the right side of the TextField box
image

before:
image

alcail (approved):

packages/react-components/src/components/styles/SendBox.styles.ts
```diff
@@ -31,14 +31,17 @@ export const sendBoxStyle = mergeStyles({
  * @private
  */
 export const sendButtonStyle = mergeStyles({
-  margin: '.6rem'
+  marginTop: '0.567rem',
+  marginBottom: '0.559rem',
```
JamesBurnside, line 35: nit: these values look super specific and hence become very fragile. Is there something we can do that makes it aligned vertically automatically? (I've found often setting lineHeight to be the same as the height to work for text so might work here, or maybe wrap the icon in a stack that can vertically align its children? I've never found a consistent way to make it vertically centered so may need some random trial and error with css props)

packages/react-components/src/components/styles/SendBox.styles.ts
```diff
@@ -31,14 +31,17 @@ export const sendBoxStyle = mergeStyles({
  * @private
  */
 export const sendButtonStyle = mergeStyles({
-  margin: '.6rem'
```
JamesBurnside, line 34: (on this iteration, looks to have removed the marginLeft and marginRight)

alkwa-msft: should be changed now.

JamesBurnside: Now we've increased the width the icon takes up by ~0.2rem, can we reduce this margin, looks to have a large right margin between the icon and send box edge:
image

alcail:
now there's an issue with icons going down when there's an error message (checked storybook doc)
image

packages/react-components/src/components/styles/SendBox.styles.ts
alkwa-msft: So the correct change is to do something like marginLeft .6rem and marginRIght .4rem? @JamesBurnside

alkwa-msft:
With the latest commit we need to go back to aligning the send button from marginTop and marginRight. This looks bad for multi-line but at least it is a good experience for a majority of scenarios. We will need to look at a better solution where we can use margin-auto later on.

9px from the top so that is 0.5625rem
and it would be 9px from the right but the icon is already absolutely positioned from the right by 4px so we want to use the value 0.313rem for our marginRight.

JamesBurnside (approved):

PR #1083, opened 2021-11-15
Move mobileView prop to top level

JamesBurnside:
# What
Move mobileView prop to top level (i.e. not a sub-prop inside options)

# Why
The likes of theme and locale are top level - so I think mobileView makes more sense here

# How Tested
Verified running sample locally


alcail (approved):

alcail:
I was wondering why it was in options prop when working on composites min w/h, so really like to have this outside of it now :)

anjulgarg:
I don't have strong feelings about it.
I do like this outside the `options` submenu.
However, wondering what the ideology behind `options` is.
How do we determine what goes in and what doesn't?

anjulgarg (approved):

PR #1085, opened 2021-11-15
Update api extractor config to do beta trimming + script to set and verify correct types value in package.json

JamesBurnside:
# What

## Update to do beta trimming
Updated api-extractor to output three levels of trimmed files:
image

When an export or property of an export is marked `@beta` is won't appear in the `@public` .d.ts file:
image

**Next PR has modifications to CI to update the field in an automated manner**: https://github.com/Azure/communication-ui-library/pull/1087

# Why
Trim exports that are not part of releases.

# How Tested

Script testing:
| test | output |
| -- | -- |
| verifying types is set to public should pass | image |
| verifying types is set to public should fail | image |
| updating to beta should be successful | image |
| now verifying beta should pass | image |

(+confirmed in above tests that the package.json was rewritten correctly)


anjulgarg:
@JamesBurnside not able to wrap my head around the usage of `packages/communication-react/scripts/verifyTypeTrimmedDeclaration.js`
Why do we need it?

JamesBurnside:
> @JamesBurnside James Burnside FTE not able to wrap my head around the usage of `packages/communication-react/scripts/verifyTypeTrimmedDeclaration.js` Why do we need it?

If we release a stable/@public version we need the types field in the package.json to be `"types": "dist/communication-react-public.d.ts`
If we release a beta version we need the types field in the package.json to be `"types": "dist/communication-react-beta.d.ts`

This (`verifyTypeTrimmedDeclaration.js`) is a script that either verifies it's correct, or updates the package.json types field. This means someone doesn't manually edit the package.json types file. It also means CI can use it to automatically update this depending on the type of release (e.g. https://github.com/Azure/communication-ui-library/pull/1087). But v. open to feedback on how we should update this types field

packages/communication-react/package.json
```diff
@@ -62,6 +62,8 @@
     "test": "",
     "test:coverage": "",
     "check-deps": "node scripts/verifyDependencies",
+    "check-types-field": "node scripts/verifyTypeTrimmedDeclaration verify",
```
prprabhu-ms, line 65: nitty nit: A more contextual name here will help (`build` builds the package, `test` tests the package, `check-types-field` type checks the package (?!))

How about `validate-package-json` and `update-package-json` (it can handle the `types` field, or other automation we might need in the future (e.g., tracking different versions of the headless SDKs))

packages/communication-react/scripts/verifyTypeTrimmedDeclaration.js
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @param releaseCandidate - this must be 'public', 'beta' or 'alpha'.
+ *
+ * @examples
+ *   > node .\verifyTypeTrimmedDeclaration.js verify --rc public
+ *   > node .\verifyTypeTrimmedDeclaration.js verify --rc beta
+ *   > node .\verifyTypeTrimmedDeclaration.js update --rc public
+ *
+ * @remarks
+ * Passing the following parameters should result in the following checks:
```
prprabhu-ms, line 18: I'd drop this comment. It repeats what the code says, without adding why. It's like to bitrot.

packages/communication-react/scripts/verifyTypeTrimmedDeclaration.js
```diff
@@ -0,0 +1,111 @@
...
+ *   * 'beta' - "types": "dist/<package.name>-beta.d.ts"
+ *   * 'alpha' - "types": "dist/<package.name>-untrimmed.d.ts"
+ */
+
+const fs = require('fs');
+const path = require('path');
+const { exit } = require('process');
+const yargs = require('yargs');
+
+// Hard code constant values. These would need dynamically fetched from the package.json if
```
prprabhu-ms, line 29: nitty nit: Would drop this comment. I don't expect this script to ever be used outside of this package (and there will be other problems besides). :)

This is a very concrete package-contextual tool.

prprabhu-ms (approved):

alcail (approved):

PorterNan:
This one is awesome!! What's our strategy of internal implementation of these apis? Someone from js world instead of ts might still using our beta/alpha features, any idea of avoiding that happening?

packages/communication-react/scripts/verifyTypeTrimmedDeclaration.js
JamesBurnside: updated

packages/communication-react/scripts/verifyTypeTrimmedDeclaration.js
JamesBurnside: updated

packages/communication-react/package.json
JamesBurnside: Leaving a bit specific for now (as `validate-package-json-types` and `update-package-json-types`) and will let you alter it slightly in the future. Reason being is we have a `check-deps` that also is a check on the package.json, but this runs at build time instead of just in a branch.

JamesBurnside: Actually - perhaps ours should also run at build time, and should use the package version to dictate what to check for. i.e. if its 1.0.0-beta.7 it should validate the types field is a beta reference.

JamesBurnside:
> This one is awesome!! What's our strategy of internal implementation of these apis? Someone from js world instead of ts might still using our beta/alpha features, any idea of avoiding that happening?

Great point, asked Dom but he wasn't sure (though pointed to azure/communication-calling that does similar exports but just tags them `@hidden` to prevent them showing in docs: https://unpkg.com/browse/@azure/communication-common@1.1.0/types/communication-common.d.ts)
Have asked Azure for their opinion on this: https://teams.microsoft.com/l/message/19:344f6b5b36ba414daa15473942c7477b@thread.skype/1637167169252?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1637167169252&teamName=Azure%20SDK&channelName=Language%20%E2%80%93%20JS%E2%80%89%EF%BC%86%E2%80%89TS%20%F0%9F%90%B1%E2%80%8D%F0%9F%91%A4&createdTime=1637167169252

prprabhu-ms:
1. Please apply the commit https://github.com/Azure/communication-ui-library/pull/1064/commits/d71fb042cac35c74477f75f413da6c6c354d4085 here so we generate only the .api.md and .json files for public types only (or perhaps a version of this change that generates both public and beta separately) because that's what we need for API reviews.
2. Turns out that api-extractor *is* as smart as we first expected -- it does trim imports that are marked @beta. So, currently, the generated `communication-react-public.d.ts` file generated in this PR is unusable - some our our @public types reference types that are still @beta in headless calling, and so the imports are trimmed (try running `rush build` after merging my commit, it will fail with missing imports)

The fix for (2) is to wait for those types to be marked @public in a headless release.

JamesBurnside:
Converting to draft while we talk through release issue with ARB

JamesBurnside:
Closing this as will likely have to not only strip the types - but the actual code as well. Can reopen if we decide otherwise

PR #1091, opened 2021-11-16
Bugfix for disappearing Lobby

anjulgarg:
# What
Fix for https://skype.visualstudio.com/SPOOL/_workitems/edit/2645502

# Why

# How Tested

# Process & policy checklist


packages/calling-component-bindings/src/callUtils.ts
```diff
@@ -5,13 +5,21 @@ import { DeviceManagerState, StatefulCallClient } from '@internal/calling-statef
 import { CallState as CallStatus } from '@azure/communication-calling';
 
 /**
- * Check if the call state represents being connected to a call
+ * Check if the call state represents being in the call
  *
  * @internal
  */
 export const _isInCall = (callStatus?: CallStatus): boolean =>
   !!callStatus && !['None', 'Disconnected', 'Connecting'].includes(callStatus);
```
alcail, line 13: why keeping the "Connecting" state here if it's in "_isInLobbyOrConnecting", unless it's used elsewhere and need this state

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -104,17 +99,16 @@ export const getCallCompositePage = (
   // If the composite completes one call and joins another, the previous calls
   // will be populated, but not relevant for determining the page.
   if (call) {
-    if (_isInCall(call?.state)) {
+    // `_isInLobbyOrConnecting` needs to be checked first because `_isInCall` also returns true when call is in lobby.
+    if (_isInLobbyOrConnecting(call?.state)) {
```
JamesBurnside, line 103: Can you quickly check that the rejoin button in the call end page works? @prprabhu-ms recently changed this file (but mostly just to move `configuration` page to be the last check: https://github.com/Azure/communication-ui-library/pull/1060/files#diff-2e8a87f35c1b3e19ff0647e3431d8a292b76ad24917b9a8315ce5ecb65cb450b)

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
prprabhu-ms: Yes, please check that the "re-join call" button works from an end page (e.g., kick out a participant and then that participant should see an end call page they can rejoin from)

Looks like I flipped this order in my PR linked above. It wasn't really intentional (just wanted to move the return conditions for 'lobby' together (sub-consciously, I'm guessing).

The right thing to do here is what @alcail asked above -- overlapping checks where order matters are brittle (as this bug shows). We should make it so that order doesn't matter for checking `_isInCall` and `_isInLobbyOrConnecting`.

That said, I'm jittery of changing this so close to release, as `_isInCall` is used elsewhere: 
image

I'd recommend adding a TODO with a link to this discussion so we can come clean up in a month or so.

prprabhu-ms (approved):

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
anjulgarg: Yes, I didn't mention it here but I noticed that this was part of @prprabhu-ms PR about rejoin button. Tested rejoin as well and it was working after this fix.

packages/calling-component-bindings/src/callUtils.ts
anjulgarg: `_isInCall` is being used in many places. Can't touch it without breaking stuff...

alkwa-msft (approved):

PR #1093, opened 2021-11-16
[API update] VideoGallery V2 custom styling and docs

mgamis-msft:
# What

Updated VideoGalleryStyles type to style VideoGallery and updated documentation.

# Why

Better styling access for users of VideoGallery

# How Tested

Storybook docs examples and test sample still works.
Please check VideoGallery docs -> Custom Style section

# Process & policy checklist


packages/react-components/review/react-components.api.md
```diff
@@ -359,6 +359,18 @@ export interface GridLayoutProps {
...
 }
 
+// @public
+export interface GridLayoutStyles extends BaseCustomStyles {
+    children?: IStyle;
+}
+
+// @public
+export interface HorizontalGalleryStyles extends BaseCustomStyles {
+    children?: IStyle;
```
PorterNan, line 369: Does "children" mean videoTile? Or something else? In html, term "children" means every dom element under a parent - if that is videoTile we can call it videoTile or something more specific

PorterNan (approved):
Left a small comment about naming, looks great to me!

packages/react-components/review/react-components.api.md
PorterNan: And "children" makes me expect an array of IStyles for types as well, which doesn't match what we show here

mgamis-msft: Good question. GridLayout and HorizontalGallery each take in children prop of every dom element so it is not only limited to VideoTile. In this case, I am using applying one IStyle to all **direct** children. It would be hard to apply an array of styles to an array of children. what if the arrays don't have the same size. I think this is the most straightforward way but intered in what others think as well.

alcail (approved):

packages/react-components/review/react-components.api.md
PorterNan: Maybe a quick storybook example could explain how this works, but yep, if that is all children elements, it makes sense to name it as children

PR #1100, opened 2021-11-17
Making sure the edit/delete flyout is always next to the message and not halfway up the screen

alkwa-msft:
# What
image

image

# Why
The flyout just wants to render at the last location it was at. When we scroll message the messages are going to be at a different vertical position on the message thread. We want to make sure when this flyout shows again it is adjacent to its message.

# How Tested
Tested locally


packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -61,8 +61,12 @@ export const ChatMessageActionFlyout = (props: ChatMessageActionFlyoutProps): JS
     ]
   );
 
+  // gap space uses pixels
   return (
     <ContextualMenu
+      alignTargetEdge={true}
```
prprabhu-ms, line 67: Please add the comment you have in PR (about flyout flyin' away) above line 67, so that next time we see this we know why.

prprabhu-ms (approved):

JamesBurnside (approved):

PR #1105, opened 2021-11-17
Delete all references to `CallTransfer` feature

prprabhu-ms:
# Why

This feature will not be available in the stable headless calling SDK when we GA, and we don't (yet) have a way to conditionally exclude the dependence on this @beta headless feature from our public bundle.

It's also unused in the UI library.

# How Tested
`rush test`


JamesBurnside (approved):

JamesBurnside:
Check with Nan about changing public API if this change isn't in his branch yet

alcail (approved):

anjulgarg (approved):

alkwa-msft (approved):

PR #1106, opened 2021-11-17
Bugfix 2648152 sorted participants

anjulgarg:
# What
Adds name based sorting to participant list selector so it returns ascending list of participants.

# Why
When a large number of participants are present in a call, it becomes difficult to find a participant if they are not sorted.

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/calling-component-bindings/src/participantListSelector.ts
+++ b/packages/calling-component-bindings/src/participantListSelector.ts
@@ -43,7 +43,9 @@ const convertRemoteParticipantsToParticipantListParticipants = (
         return -1;
       } else if (nameA > nameB) {
         return 1;
-      } else return 0;
+      } else {
+        return 0;
+      }
     });
 };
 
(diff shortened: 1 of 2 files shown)


packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -18,23 +18,33 @@ import { getIdentifierKind } from '@azure/communication-common';
...
+      };
+    })
+    .sort((a, b) => {
+      const nameA = a.displayName?.toLowerCase() || '';
+      const nameB = b.displayName?.toLowerCase() || '';
+      if (nameA < nameB) {
+        return -1;
+      } else if (nameA > nameB) {
+        return 1;
+      } else return 0;
```
JamesBurnside, line 46: nit: keep consistent bracketing
```suggestion
      } else {
        return 0;
      }
```

Perhaps there's a prettifier rule we can turn on for this so its not something anyone has to care about..

JamesBurnside (approved):

packages/calling-component-bindings/src/participantListSelector.ts
anjulgarg: Yes, a prettier rule would be great.

alkwa-msft (approved):

PR #1111, opened 2021-11-18
VideoGallery UUID for its local video tile to dock to.

mgamis-msft:
# What

Assigning VideoGallery containerRef.current.id using a UUID. Then using containerRef.current.id which is assigned to root div as the Modal's hostId.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2649260

When there are multiple VideoGallery's in the same page their id's will be the same since it was hardcoded and thus all local video tiles will dock to the first VideoGallery.

# How Tested

Verified docs in storybook. Please check chromatic.

# Process & policy checklist


packages/storybook/stories/VideoGallery/__snapshots__/VideoGallery.stories.storyshot
```diff
@@ -23,11 +23,208 @@ exports[`storybook snapshot tests Storyshots UI Components/Video Gallery Video G
       <div
         className="css-85"
         data-ui-id="video-gallery"
-        id="UILibraryFloatingTileHost"
       >
         <span
           className="ms-layer"
-        />
+        >
```
mgamis-msft, line 29: This is the FluentUI Modal. Not sure why the old snapshot did not have it.

JamesBurnside (approved):
Great fix!

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -303,9 +302,14 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     />
   );
 
+  useEffect(() => {
+    if (containerRef.current) {
+      containerRef.current.id = `video-gallery-${uuidv4()}`;
```
alkwa-msft, line 307: The idea is that whenever our containerRef changes we want to use a new UUID so we don't have collisions between different video gallery instances?

alcail: curious to know too

mgamis-msft: This useEffect only happens since containerRef is instantiated only once during the lifetime of the VideoGallery component. So the UUID is only created once so it should be the same for a VIdeoGallery's lifetime.

alcail (approved):

packages/react-components/src/components/VideoGallery.tsx
prprabhu-ms: Why not simply generate an `id` once.

    containerId = useRef(`video-gallery-${uuidv4()}`)
    ...

    return (<div id={containerId.current}> .... </div>)

mgamis-msft: I like that better too. Changed. Thanks.

common/config/rollup/rollup.config.js
```diff
@@ -18,5 +19,5 @@ export default (packageJson) => ({
     format: 'cjs',
     sourcemap: true
   },
-  plugins: [commonjs(), sourcemaps(), svg(), json({ compact: true })]
+  plugins: [commonjs(), sourcemaps(), svg(), json({ compact: true }), nodeResolve({browser: true})]
```
mgamis-msft, line 22: @JamesBurnside I had to change the rollup.config.js to use uuid because Rollup could not tell which uuid library I wanted to import. I got this solution from here https://stackoverflow.com/questions/59937658/importing-uuid-in-rollup-esm-library-creates-an-import-statement-for-crypto-in. Not sure if it's the right way to go.

JamesBurnside: Oh thanks for flagging this, let me take a look tomorrow before completing this

prprabhu-ms (approved):

JamesBurnside (approved):

PR #1112, opened 2021-11-18
Alkwa/hide participants in lobby

alkwa-msft:
# What
Participants are showing up in the participant list even though they are only in the Lobby and not in the actual call.

# Why
There could be a confusion that since customers are in the participant list they are actively being a part of the call. Today that is not the case. We want to hide the users in the ACS client participant list till we have a way to allow users in the participant list to be invited into the call from the Lobby

#How it was tested


JamesBurnside (approved):

alcail (approved):

anjulgarg (approved):

prprabhu-ms (approved):

prprabhu-ms:
I noticed that we even show a video tile for participants in the lobby. That probably needs a similar filter?

prprabhu-ms:
> I noticed that we even show a video tile for participants in the lobby. That probably needs a similar filter?

filed https://skype.visualstudio.com/SPOOL/_workitems/edit/2651464 set to P1 (can reprioritize as needed)

prprabhu-ms:
Ah, I see that https://skype.visualstudio.com/SPOOL/_workitems/edit/2649505 already covers both issues. This PR is a partial fix.

alkwa-msft:
> > I noticed that we even show a video tile for participants in the lobby. That probably needs a similar filter?
> 
> filed https://skype.visualstudio.com/SPOOL/_workitems/edit/2651464 set to P1 (can reprioritize as needed)

adjusting the VideoGallery selector as well

PR #1114, opened 2021-11-18
Remove beta API in preparation for GA release

prprabhu-ms:
# What

* Hide all beta-only portions of public API
* Delete e2e tests for meetings (since the public API is removed).
* Remove references to beta-only API from storybook
* Update Chat sample to not depend on beta-only API

# Why

We're still discussing on the right way to do this going forward. But for our GA, the actual @beta surface is fairly small, and this is the safest way to remove it.


packages/communication-react/review/communication-react.api.md
```diff
@@ -582,7 +572,6 @@ export type ChatCompositeIcons = Partial<Pick<CompositeIcons, 'MessageDelivered'
 // @public
 export type ChatCompositeOptions = {
     errorBar?: boolean;
-    participantPane?: boolean;
```
prprabhu-ms, line 585: @JamesBurnside looks like api-extract isn't respective the `@beta` tag for this field of an `@public` interface.

The documentation on this is a bit unclear: 
image

from https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags

My reading of that was that we *should* be able to restrict the fields in a type to  more restrictive release tag, but that doesn't seem to be working.

We need to fix this regardless of my PoC PR here. Currently `ChatCompositeOptions.participantPane` is part of public API (though with the default value of false that we wanted)

@PorterNan our resident API change czar at the moment.

JamesBurnside: Not sure I follow, api-trimming will trim out the participant pane if its marked @beta:
image

There's a difference between the api.md file and the .d.ts file though (api.md isn't trimmed - but is only used for us reviewing things)

JamesBurnside: (though the fact it doesn't show a `//beta` beside it in the api.md file looks definitely like a bug on them)

prprabhu-ms: Ah, OK then.

See my comment on your beta trimming PR for how we should generate separate APIview files for the trimmed types as well.

prprabhu-ms (commented):
Need to also hide the `MeetingComposite` storybook pages.

JamesBurnside:
Can you remove the chat sample participant button in this also?

JamesBurnside:
> Need to also hide the `MeetingComposite` storybook pages.

Thoughts on keeping beta things in the storybook? Now they have beta banners: https://60c7ae6891f0e90039d7cd54-mjxwxkeeuq.chromatic.com/?path=/docs/composites-meeting-basicexample--basic-example

prprabhu-ms:
> Can you remove the chat sample participant button in this also?

Done.

prprabhu-ms:
> > Need to also hide the `MeetingComposite` storybook pages.
> 
> Thoughts on keeping beta things in the storybook? Now they have beta banners: https://60c7ae6891f0e90039d7cd54-mjxwxkeeuq.chromatic.com/?path=/docs/composites-meeting-basicexample--basic-example

Done.

JamesBurnside (approved):
These (`**/storybook/**/***.REMOVED` files) live on in the feature branch so happy to remove them (assuming we merge in a good way that re-adds them, which might be tricky, so also happy to keep them there)

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -67,7 +61,7 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
     onFetchAvatarPersonaData,
     onRenderMessage,
     onRenderTypingIndicator,
-    onFetchParticipantMenuItems,
+    // onFetchParticipantMenuItems,  // Removed for GA release
```
alcail, line 64: are we still keeping this comment?

alcail (approved):

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
prprabhu-ms: Yeah, I am trying to keep the diff small where I had to make intrusive changes / leave bread crumbs.

prprabhu-ms:
> These (`**/storybook/**/***.REMOVED` files) live on in the feature branch so happy to remove them (assuming we merge in a good way that re-adds them, which might be tricky, so also happy to keep them there)

I'm a bit wary of this PR. Trying to leave breadcrumbs where I can.

prprabhu-ms:
Added `ui change` to update `ChatComposite` snapshots, removing participant pane.

prprabhu-ms:
Last CI failure was a flaky `ChatComposite` e2e test.

PR #1115, opened 2021-11-18
Fix "New Messages" button not disappearing when scrolling to the bottom of the message thread

JamesBurnside:
# What
Fix the math, if this repros for someone again we can add a tolerance here also.
Example of when this failed:
image

# Why
Scrolling to the bottom didn't always dismiss the New messages button:
image

# How Tested
Storybook on web only - can no longer repro bug

### DIFF
--- a/packages/react-components/src/components/MessageThread.tsx
+++ b/packages/react-components/src/components/MessageThread.tsx
@@ -720,7 +720,6 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
         scrollToBottom();
       }
     }
-
     setIsAtBottomOfScrollRef(atBottom);
   }, [scrollToBottom, sendMessageStatusIfAtBottom]);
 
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -712,7 +712,7 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
     }
 
     const atBottom =
-      Math.floor(chatScrollDivRef.current.scrollTop) >=
+      Math.ceil(chatScrollDivRef.current.scrollTop) >=
```
prprabhu-ms, line 715: I'm bad at math. I'd prefer to check that absolute value of difference within some tolerance.

prprabhu-ms (approved):

alcail (approved):

packages/react-components/src/components/MessageThread.tsx
JamesBurnside: I was considering tolerance but didn't want to cause any unintended side effects. I'm going to leave as is to keep the delta (so hopefully any regressions) super small but I will approve anyone who wants to update this to a tolerance calculation.

PR #1116, opened 2021-11-18
Remove long press gesture on MessageThread messages

JamesBurnside:
# What
Remove the long touch press gesture on messages

# Why
After trying it we have decided to remove this after the UX isn't great.

# How Tested
Storybook - can still edi/remove messages by touch, pointer and keyboard nav


JamesBurnside:
Actually, missed a feature here, will reopen again

PR #1117, opened 2021-11-18
Fix ComplianceBanner auto-dismiss

prprabhu-ms:
# What

Fix a bug where `ComplianceBanner` was being automatically dismissed on unrelated rerenders.

This bug was introduced in the refactor in #973 Before that refactor, a `useEffect` block ensured that cached props were only updated when the value of incoming props actually changed.
This PR adds an explicit check for the same.


anjulgarg (approved):

packages/react-composites/src/composites/CallComposite/components/ComplianceBanner.tsx
```diff
@@ -12,27 +12,47 @@ export type ComplianceBannerProps = {
   callRecordState?: boolean;
 };
 
+type CachedComplianceBannerProps = {
+  new: ComplianceBannerProps;
```
alcail, line 16: nitty nit: just a personal taste on not using 'new' as an only word.
So, maybe 'coming'/'next'/'current' instead, and 'previous' instead of 'old'

alcail (approved):

PR #1123, opened 2021-11-19
Ensure TypingIndicator wraps correctly with very long names

JamesBurnside:
# What
Ensure the typing indicator can handle long names/strings by wrapping.

# Why
Caused horizontal scrollbars before this

# How Tested

Long names:
| before | after |
|--|--|
| image |  image |

Normal names:

| before | after |
|--|--|
|image|image|


packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -281,3 +283,38 @@ export const TypingIndicator = (props: TypingIndicatorProps): JSX.Element => {
...
+
+/**
+ * Create an array of span elements by replacing the pattern "\{\}" in str with the elements
+ * passed in as vars and creating inline elements from the rest
+ *
+ * @param str - The string to be formatted
+ * @param vars - Variables to use to format the string
+ * @returns formatted JSX elements
+ */
+const formatInlineElements = (str: string, vars: _IObjectMap<JSX.Element>): JSX.Element[] => {
```
alcail, line 295: why removing this from utils file and adding it here?
It might be used elsewhere in the future

alcail (approved):

packages/react-components/src/components/TypingIndicator.tsx
JamesBurnside: The reason was it wasn't used anywhere else and keeping it here would mean adding fluentui/react dependency to the acs-common package. Happy to move it back and add this dependency if people would prefer though!

JamesBurnside: (Ideally this is only used by react-components package as its a very specialized util fn, but there is a chance it could be used by composites)

prprabhu-ms (approved):

PR #1126, opened 2021-11-19
Chat sample - Config page - a11y label issue on display name TextField

alcail:
# What
* updated text field label of Configuration page of Chat sample to be a prop of this component and not a separated text so now the narrator can read it properly
* added page heading and aria-label to Avatar list so Narrator announce what the list is for

No visual change
image

# Why
bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2583860

# How Tested
ran sample Chat locally and used Narrator

# Process & policy checklist

### DIFF
--- a/samples/Chat/src/app/ConfigurationScreen.tsx
+++ b/samples/Chat/src/app/ConfigurationScreen.tsx
@@ -53,6 +53,16 @@ const CONFIGURATIONSCREEN_SHOWING_JOIN_CHAT = 2;
 const CONFIGURATIONSCREEN_SHOWING_INVALID_THREAD = 3;
 const CONFIGURATIONSCREEN_SHOWING_SPINNER_INITIALIZE_CHAT = 4;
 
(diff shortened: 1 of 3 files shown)


JamesBurnside (approved):

samples/Chat/src/app/ConfigurationScreen.tsx
```diff
@@ -197,7 +219,7 @@ export default (props: ConfigurationScreenProps): JSX.Element => {
             disabled={disableJoinChatButton}
             className={buttonStyle}
             styles={buttonWithIconStyles}
-            text={'Join chat'}
+            text={JOIN_BTN_TEXT}
```
prprabhu-ms, line 222: nitty nit: Avoid shortening words like button -> btn

prprabhu-ms (approved):

alkwa-msft (approved):

PR #1127, opened 2021-11-19
[POC] Conditional build

PorterNan:
# What

How it works:
It is a 2-pass build:
1. Babel to strip out beta part (takes 5s in my machine)
2. tsc like what we normally do

Inspired by babel-plugin-conditional-compilation and babel-plugin-transform-strip-block, I decided to write our own babel plugin which is like a combination of 2 project and have both features from then, and:

1. Support comments so it can be literally inserted into anywhere
2. Support import, export and more
2. We could add our own code to customize what we need
3. Support type features for typescripts type

There are some limitations of this:
1. Babel has a little bit weaker ability to understand typescript in some scenarios(see 2 of my minor changes in this PR), such as:
- it doesn't support type cast like <Foo>bar well (`bar as Foo` is supported and it is always recommended to do this)
- combination of ?? (double question marks operator) + type cast

2. The code to be striped out should be complete statement/expression/declaration
Support:
```
/* beta:start */
const foo = bar;
/* beta:end */
const foo1 = bar1;
```

Unsupport (the statement is cut in the middle):
```
const foo = bar
/* beta:start */
? bar1 : bar
/* beta:end */
```

- Also take participantPane as an good example of how to do a beta path for component - we should extract most of code into a single file and mark as minimum as possible code as beta - it will make it easy to remove beta tag in the future.

- File operation details: 
1. Move all original files into `communication-react/preprocess-dist/*/preprocess` (all ts file will be override by babel later, but other assets will be remaining like json files)
2. Babel will create `preprocess` folder under each subfolder of /package/
3. Move all `preprocess` contents into `communication-react/preprocess-dist`
4. Rename all `preprocess` folder names to `src` (to match the import path)

- How to run it 
run `rushx build:public` under `communication-react` folder

# Why

# How Tested

# Process & policy checklist


packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -27,7 +27,7 @@ enableMapSet();
  */
 export class ChatContext {
   private _state: ChatClientState = {
-    userId: <UnknownIdentifierKind>{ id: '' },
```
PorterNan, line 30: One of the case, Babel can't understand - <> cast, however, foo as Bar is always the recommended way.

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -244,7 +244,7 @@ const ParticipantSystemMessageComponent = ({
   if (participantsStr !== '') {
     return (
       <SystemMessageComponent
-        iconName={(message.iconName ?? '') as SystemMessageIconTypes}
```
PorterNan, line 247: This could be a bug of babel parser - a combination of ?? and type cast. This is super rare and we might also want to deal with prop type ourselves to avoid type cast anyway

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -49,13 +49,16 @@ export type ChatCompositeOptions = {
    * @defaultValue true
    */
   errorBar?: boolean;
+  /* beta:start */
```
PorterNan, line 52: image
This is how it looks after stripping, property removed while comments remain - if we wanna remove comments, there could be more edge cases to deal with to avoid false-positive removal

PorterNan: Since this is just intermediate product of the build, we should be good not to remove comments

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -133,27 +120,16 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
-                    />
-                  )}
-                  onFetchParticipantMenuItems={onFetchParticipantMenuItems}
-                />
-              </FocusZone>
-            </Stack>
-          </Stack>
-        )}
+        {
+          /* beta:start */
```
PorterNan, line 124: Extract beta component logic into a single file is the best way to minimize beta blocks - which helps us make smaller files and clearer logic

packages/react-composites/src/composites/index.ts
```diff
@@ -5,10 +5,13 @@ export type { Disposable, AdapterState } from './common/adapters';
...
 export * from './ChatComposite';
 export * from './CallComposite';
-export * from './MeetingComposite';
 export type { AvatarPersonaData, AvatarPersonaDataCallback } from './common/AvatarPersona';
 export * from './common/icons';
 export * from './localization/locales';
 export type { CompositeStrings, CompositeLocale } from './localization';
 export type { AdapterError, AdapterErrors } from './common/adapters';
 export type { BaseCompositeProps } from './common/BaseComposite';
+/* beta:start */
```
PorterNan, line 14: Export sentences are also supported

JamesBurnside:
This is very cool!

packages/communication-react/package.json
```diff
@@ -54,9 +55,15 @@
...
   "scripts": {
+    "copy-original": "cpx ../*/src/** ./preprocess-dist & renamer --find src --replace preprocessed ./preprocess-dist/**",
+    "preprocess": "babel ../*/src --out-dir ../preprocessed --extensions \".ts,.tsx\" --keep-file-extension --config-file ./.babelrc.js --relative",
+    "preprocess:public": "cross-env FLAVOR=public rushx preprocess",
+    "copy-preprocess": "cpx ../*/preprocessed/** ./preprocess-dist & renamer --find preprocessed --replace src ./preprocess-dist/**",
+    "clean:preprocess": "rimraf ../**/preprocessed && rimraf ../**/preprocess-dist",
     "build": "rushx clean && rushx check-deps && rushx build:esm && rushx build:cjs && api-extractor run --local",
     "build:cjs": "rollup -c --silent",
     "build:esm": "ttsc",
+    "build:public": "rushx copy-original && rushx preprocess:public && rushx copy-preprocess && ttsc -project tsconfig.preprocess.json && rushx build:cjs && api-extractor run --local && rushx clean:preprocess",
```
JamesBurnside, line 66: Should this be added to every package? I'm thinking for UI tests and unit tests we want them to run on on the beta-stripped code

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
alcail: feels weird though but I don't feel that it's worse trying to remove the comments too, so let's keep it that way

packages/react-composites/src/composites/localization/LocalizationProvider.tsx
```diff
@@ -69,6 +69,10 @@ export type LocalizationProviderProps = {
  */
 export const LocalizationProvider = (props: LocalizationProviderProps): JSX.Element => {
   const { locale, children } = props;
+  /* beta:start */
+  console.log('beta-test1');
```
alcail, line 73: you will need to remove those, right?

alcail (approved):

packages/communication-react/package.json
PorterNan: Good point! Yep, we need UTs and UI tests
But this command is a little bit wordy to add to each package
I guess what we could do here:
1. For UTs, we can run them inside of communication-react under preprocess-dist while everything gets generated
2. For UI tests, which we need to webpack the app - we can add this babel plugin into webpack and it will run stripping when bundling


Will do some investigation and update this POC

packages/react-composites/src/composites/localization/LocalizationProvider.tsx
PorterNan: You are right! That's why I marked this as POC, this is an example of how things work for normal statements

DominikMe:
@PorterNan Have we considered writing a TypeScript transform instead of trying to tweak babel?
That should give us faster processing speed and full type info parsing.

prprabhu-ms:
+1 to typescript transform.

A few more high-level comments / recommendations:

## Dumber conditional compilation

I'd like to dumb this down so that we force some restrictions on how we leverage conditional compilation, ensuring more manageable code. Instead of arbitrary conditional blocks, how about we support the following:
 
A single comment of the form `// @conditional-compile(beta)` or `// @conditional-compile(stable)`.

`// @conditional-compile(beta)` can be applied to:

* A type or interface definition, only at the top-level in a file.
* A field in a type of interface definition, only at the top-level in a file.
* A function definition

and `// @conditional-compile(beta)` or `// @conditional-compile(stable)` can be applied to

* An import statement

The idea is that beta-only types and fields and associated implementation can be marked as such.

Whenever we need actual conditional logic, we must provide alternative implementations for stable/beta in different files (with the beta implementations marked as `// @conditional-compile(beta)` and the correct definition imported using the a conditional import.

I feel that this scheme will force us to keep the diff between stable and beta clear and small.

## Support for local edit-compile-test

We must provide `rush` commands in each packlet to be able to build and unittests against stable and beta. I'd expect that this will be part of the local dev cycle (if not each time, definitely before PR upload / to repro failure in CI). The simpler we make to build and test compilation flavors, the easier they are to maintain. Might even be worth just using a ENV variable to chose the flavour and all `rush` commands "do the right thing" for the right flavour.

prprabhu-ms:
> @PorterNan Have we considered writing a TypeScript transform instead of trying to tweak babel? That should give us faster processing speed and full type info parsing.

Unfortunately, typescript transform allows us to add transformation passes only *after* typechecking. This is a problem because we don't get type safety in the generated AST.

To be able to typecheck after transformation, we'd need to use something like https://github.com/dsherret/ts-morph/tree/latest/packages/ts-morph

See https://stackoverflow.com/questions/63947663/transform-typescript-before-typecheck

PorterNan:
> +1 to typescript transform.
> 
> A few more high-level comments / recommendations:
> 
> ## Dumber conditional compilation
> I'd like to dumb this down so that we force some restrictions on how we leverage conditional compilation, ensuring more manageable code. Instead of arbitrary conditional blocks, how about we support the following:
> 
> A single comment of the form `// @conditional-compile(beta)` or `// @conditional-compile(stable)`.
> 
> `// @conditional-compile(beta)` can be applied to:
> 
> * A type or interface definition, only at the top-level in a file.
> * A field in a type of interface definition, only at the top-level in a file.
> * A function definition
> 
> and `// @conditional-compile(beta)` or `// @conditional-compile(stable)` can be applied to
> 
> * An import or export statement
> 
> The idea is that beta-only types and fields and associated implementation can be marked as such.
> 
> Whenever we need actual conditional logic, we must provide alternative implementations for stable/beta in different files (with the beta implementations marked as `// @conditional-compile(beta)` and the correct definition imported using the a conditional import.
> 
> I feel that this scheme will force us to keep the diff between stable and beta clear and small.
> 
> ## Support for local edit-compile-test
> We must provide `rush` commands in each packlet to be able to build and unittests against stable and beta. I'd expect that this will be part of the local dev cycle (if not each time, definitely before PR upload / to repro failure in CI). The simpler we make to build and test compilation flavors, the easier they are to maintain. Might even be worth just using a ENV variable to chose the flavour and all `rush` commands "do the right thing" for the right flavour.

If we go this way - that means our code always need a pre-compile when usage - and it creates always 2 paths for each beta/stable code

What about only going ahead with // @conditional-compile(beta), it only decorates single statement, import, functions (means it still support statement, but only one line instead of a block)

This will force people write a single line as much as possible when they write code for beta(otherwise they will have to // @conditional-compile(beta) to each line which we will block them from PR most likely), and it is still keep there is only one branch to go with (and code is still compliable even without preprocess)

PorterNan:
I am looking at some typescript transformer stuffs around, including ts-morph, looks like we still need 2-pass build to generate file  then build the package - currently pre-process only takes 10s in my machine (Babel is good at perf for this one)

I chose babel because it is a second-official way to build/transpile typescript into js - you can see people are using it all around with webpack or other tools https://learntypescript.dev/12/l2-babel/

And we get all accurate type information by @babel/plugin-transform-typescript which you can easily find all related ts node doc here: https://babeljs.io/docs/en/babel-types#typescript

and finally, babel is one of the AST tools which handles comments very well

So I would say let's make this babel version as the initial version of conditional compile - and we could start some investigation on ts transformation like ts-morph, and figure out some of undetermined questions here:
1. How's their perf (See perf issues of ts-morph)
2. How well these tools support comments/JSDoc

PorterNan:
Close this cause landing this one into another feature branch

PR #1128, opened 2021-11-19
Add linting to Storybook mdx files + fail Storybook build if there are lint warnings

JamesBurnside:
# What
* Add `eslint-plugin-mdx` to lint md files
  * Fix linting errors (mostly just unused/unoredered imports)
* Fail builds if eslint fails
  * Fix existing linting issues in tsx files (caught a couple bugs in useEffect blocks)

# Why
We're doing a lot of storybook changes and this help ensure our code is working
Not sure why failing builds isn't already turned on... (same with other packlets, separate PR for those)

### DIFF
--- a/packages/storybook/stories/ControlBar/Buttons/Devices/snippets/DevicesButtonWithKnobs.snippet.tsx
+++ b/packages/storybook/stories/ControlBar/Buttons/Devices/snippets/DevicesButtonWithKnobs.snippet.tsx
@@ -5,7 +5,7 @@ import { DevicesButton, DevicesButtonProps } from '@azure/communication-react';
 import React, { useState } from 'react';
 import { defaultControlsCameras, defaultControlsMicrophones, defaultControlsSpeakers } from '../../../../controlsUtils';
 
-export const DevicesButtonWithKnobs = (args: unknown): JSX.Element => {
+export const DevicesButtonWithKnobs = (deviceButtonProps: DevicesButtonProps): JSX.Element => {
   const [selectedCamera, setSelectedCamera] = useState<{ id: string; name: string }>(defaultControlsCameras[0]);
   const [selectedMicrophone, setSelectedMicrophone] = useState<{ id: string; name: string }>(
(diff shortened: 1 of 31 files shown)


prprabhu-ms (approved):

packages/storybook/stories/ControlBar/Buttons/Devices/snippets/DevicesButtonWithKnobs.snippet.tsx
```diff
@@ -5,7 +5,7 @@ import { DevicesButton, DevicesButtonProps } from '@azure/communication-react';
 import React, { useState } from 'react';
 import { defaultControlsCameras, defaultControlsMicrophones, defaultControlsSpeakers } from '../../../../controlsUtils';
 
-export const DevicesButtonWithKnobs = (args: any): JSX.Element => {
+export const DevicesButtonWithKnobs = (args: unknown): JSX.Element => {
```
alcail, line 8: curious about this change

alcail (approved):

alkwa-msft (approved):

packages/storybook/stories/ControlBar/Buttons/Devices/snippets/DevicesButtonWithKnobs.snippet.tsx
JamesBurnside: We shouldn't use `any` types because you can do `myVar.foo` and typescript won't complain. If we don't know the type we should use `unknown`. Doing myVar.foo will then throw an error, so to use myVar we should explicit casing (e.g. `(myVar as KnownType).foo`.

But now I'm looking at this again, this should be `DeviceButtonProps` and the caller should do any `any->unknown->type` casting. I'll update this

JamesBurnside: updated.

PR #1134, opened 2021-11-19
Added identity quickstart to composite initiate screens

ddematheu2:
# What
Added links to composite screens to point to identity quickstarts

# Why
Today developers don't necessarily know what to do when it comes to getting tokens to initiate the composites.

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/storybook/stories/CompositeStringUtils.tsx
+++ b/packages/storybook/stories/CompositeStringUtils.tsx
@@ -13,3 +13,13 @@ export const CompositeConnectionParamsErrMessage = (errors: string[]): JSX.Eleme
     })}
   </>
 );
+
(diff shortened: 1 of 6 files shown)


packages/storybook/stories/CompositeStringUtils.tsx
```diff
@@ -13,3 +13,13 @@ export const CompositeConnectionParamsErrMessage = (errors: string[]): JSX.Eleme
     })}
   </>
 );
+
+export const CreateIdentityLink = (): JSX.Element => {
```
alkwa-msft, line 17: This doesn't need to be a function it can just be a constant that is a JSX.Element

ddematheu2:
Looks like this was already fixed.

PR #1135, opened 2021-11-22
Remove useEffect from remote video tile

mgamis-msft:
# What

Remove both useEffects in RemoteVideoTile. 
1. Removing the first useEffect because it is not necessary since the code in the useEffect block depends on props anyways
2. Removing the second useEffect because it is called when component dismounts when the the HorizontalGallery page changes causing it to call `onDisposeRemoteStreamView(userId)`. As James pointed out in the [comment](https://github.com/Azure/communication-ui-library/pull/1135#discussion_r755525431), screenshare stream and video stream are disposed together, so if the screensharer's video tile dismounts, their screenshare stream also is disposed and re-rendered.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2641235

# How Tested

On local Calling sample:
https://user-images.githubusercontent.com/79475487/142915040-1bf9e6d1-b276-4fba-a508-8ed7d24b4bae.mp4

# Process & policy checklist


anjulgarg (approved):

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -39,27 +39,12 @@ export const RemoteVideoTile = React.memo(
...
-      userId
-    ]);
-
-    useEffect(() => {
-      return () => {
-        onDisposeRemoteStreamView && onDisposeRemoteStreamView(userId);
-      };
-    }, [onDisposeRemoteStreamView, userId]);
+    if (isAvailable && !renderElement) {
+      onCreateRemoteStreamView && onCreateRemoteStreamView(userId, remoteVideoViewOption);
```
JamesBurnside, line 43: note: is there a chance this will be called twice, when isAvailable but before the renderElement has been created? (not sure how long `onCreateRemoteStreamView` takes for renderElement to become defined

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -39,27 +39,12 @@ export const RemoteVideoTile = React.memo(
       showMuteIndicator
     } = props;
 
-    useEffect(() => {
-      if (isAvailable && !renderElement) {
-        onCreateRemoteStreamView && onCreateRemoteStreamView(userId, remoteVideoViewOption);
-      }
-      if (!isAvailable) {
```
JamesBurnside, line 46: I think I'm still confused on the solution here, from the PR desc, you mention that onDisposeRemoteStreamView being called is the reason for the recreation of the tile - but why does this get called?
Is it being called here when `!isAvailable` or is it being called in the `return () => { onDisposeRemoteStreamView && onDisposeRemoteStreamView(userId);`?

mgamis-msft: Sorry for the confusion. I updated the description. 
To answer the quesiton, it is being called in the second useEffect with `return () => { onDisposeRemoteStreamView && onDisposeRemoteStreamView(userId);` after the `RemoteVideoTile` at the end of its lifetime when the HorizontalGallery page changes. So when the screensharer's `RemoteVideoTile` ends its lifetime this causes a change in the VideoGalleryRemoteParticipant object of the screensharer. Since the VideoGalleryRemoteParticipant object of the screensharer is a prop of `RemoteScreenShare`, it will cause a rerender because of the change.

packages/react-components/src/components/RemoteVideoTile.tsx
mgamis-msft: Yes, that is possible. I think onCreateRemoteStreamView calls an SDK method that is idempotent but I could be wrong.

alcail (approved):

packages/react-components/src/components/RemoteVideoTile.tsx
JamesBurnside: My concern here is now, at the end of this tile's lifetime, onDisposeRemoteStreamView is not being called, so ultimately `callClient.disposeView` is never being called. Not sure what any implications that actually has, but its probably a resource we should be cleaning up.

`onDisposeRemoteStreamView ` is super unopinionated and cleans up both the remote camera stream and the screenshare stream:
image

Given we might want to create just a screenshare stream or a video feed independently (and in this case - just dispose a video feed and not a screenshare stream) - I think we need to distinguish in onCreateRemoteStreamView and onDisposeRemoteStreamView if we actually want both or just one or the other.

+ @PorterNan for opinion here

This'll likely be a change to the API, so if this change doesn't cause any performance issues (e.g. imagine a video of 40+ people, if they've each had a video feed created at some point but never disposed - will we hit problems?) I'm happy to keep this change and file a work item for after GA to handle videoFeed and screenshareFeed seperately

mgamis-msft: Thanks for pointing out that `onDisposeRemoteStreamView ` disposes both video stream and screenshare stream. I was not aware of that. That is the main reason for the re-render.

JamesBurnside (changes requested):
Just while comment is unresolved, want to make sure this won't cause any perf impact

mgamis-msft:
> Just while waiting on feedback on unresolved comment, want to make sure this won't cause any perf impact/insidious bug where streams mightn't work

No worries. I'm fine if the plan is to leave the second useEffect since it ensures resources are not wasted although it does not fix the rerender. I would need to add the dispose method on the RemoteScreenShare component as well. That too will cause rerender of the screensharer video stream if they stop screen sharing.

mgamis-msft:
Closing this to create 2 alternative PR's for 1.) a possible API change 2.) workaround to avoid disposing screenshare when participant is screensharing

PR #1138, opened 2021-11-22
Prevent dismissal of ContextMenu on window resize

prprabhu-ms:
# What

Prevent dismissal of ContextMenu for ParticipantsButton and DevicesButton on resize events.

# Why

The CallOut is dismissed whenever there is a 'resize' event for any element in the window. This happens a lot in the VideoGallery, and leads to an experience where the callout is dismissed while in use.

A side effect of this PR is that it is now possible to get into a situation where the callout is no longer at the original target, but isn't dismissed. This does require a large amount of window resize though, so it's not too bad:

image

This workaround can be removed if Fluent UI's Callout can be made to only dismiss when the window is resized (instead of any element in the window)

# How Tested
Calling sample


JamesBurnside (approved):

packages/react-components/src/components/DevicesButton.tsx
```diff
@@ -172,12 +172,28 @@ const generateDefaultMenuProps = (
...
+      // - The Callout is dismissed whenever *any child of window (inclusive)* is resized. In practice, this
+      //   happens when we change the VideoGallery layout, or even when the video stream element is internally resized
+      //   by the headless SDK.
+      // - There is a `preventDismissOnEvent` prop that we could theoretically use to only dismiss when the target of
+      //   of the 'resize' event is the window itself. But experimentation shows that setting that prop doesn't
+      //   deterministically avoid dismissal.
+      //
+      // A side effect of this workaround is that the context menu stays open when window is resized, and may
+      // get detached from original target visually. That bug is preferable to the bug when this value is not set -
+      // The Callout (freuently) gets dismissed automatically.
```
JamesBurnside, line 194: I mean there's no ambiguity to what this word meant so very pointless nit: `freuently` -> `frequently`

packages/react-components/src/components/DevicesButton.tsx
```diff
@@ -172,12 +172,28 @@ const generateDefaultMenuProps = (
...
-
-    // Confine the menu to the parents bounds.
-    // More info: https://github.com/microsoft/fluentui/issues/18835
-    calloutProps: { styles: { root: { maxWidth: '95%' } } }
+    calloutProps: {
+      styles: {
+        root: {
+          // Confine the menu to the parents bounds.
+          // More info: https://github.com/microsoft/fluentui/issues/18835
+          maxWidth: '100%'
```
JamesBurnside, line 181: keep as 95% here (and probably 95% n the others really), because the size of the actual callout is slightly bigger (drop shadow and some margins) than what this calloutProps targets

JamesBurnside: (You can git blame me on that one for not adding a good comment 😳)

alcail (approved):

packages/react-components/src/components/DevicesButton.tsx
prprabhu-ms: Thanks for catching that. Also added comment.

PR #1139, opened 2021-11-23
Bugfix 2641725 - Disable microphone toggle in lobby

anjulgarg:
# What
Disable microphone toggle in lobby.
Fix for bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2641725

# Why
Currently, toggling microphone in lobby causes console errors as it is not supported by headless sdk.
 
# How Tested

# Process & policy checklist


anjulgarg:
@prprabhu-ms @JamesBurnside 
Couldn't find a clean way to abstract the logic away inside a component level selector in `packages/react-composites/src/composites/CallComposite/components/CallControls.tsx`

Any suggestions how it can be done?

cc: @PorterNan

alcail (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -327,19 +330,17 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
 
   public async mute(): Promise<void> {
     return await this.asyncTeeErrorToEventEmitter(async () => {
-      if (!_isInCall(this.call?.state)) {
-        this.context.setIsLocalMicrophoneEnabled(false);
-      } else if (!this.call?.isMuted) {
+      this.context.setIsLocalMicrophoneEnabled(false);
```
JamesBurnside, line 333: Where isLocalMicrophoneEnabled is defined, can you add a comment about this new behavior, i.e. something like
```
IsLocalMicrophoneEnabled is used to track the user's selection of whether they want the microphone to be enabled (unmuted) or not. This may differ from the call state of the users microphone; for example when the call has not yet been joined, when the user is in the Lobby of a Teams call, or when the user is muted by another participant in the call.
```

packages/react-composites/src/composites/CallComposite/pages/CallPage.tsx
```diff
@@ -53,10 +56,25 @@ export const CallPage = (props: CallPageProps): JSX.Element => {
...
   const networkReconnectTileProps = useSelector(networkReconnectTileSelector);
+  const isLocalMicrophoneEnabled = useSelector(getLocalMicrophoneEnabled);
+  const adapter = useAdapter();
 
   // Reduce the controls shown when mobile view is enabled.
   const callControlOptions = mobileView ? reduceCallControlsForMobile(options?.callControls) : options?.callControls;
 
+  /**
+   * Since toggling microphone in lobby page doesn't actually change the microphone status in the call object,
+   * we track it as a ui state attribute `isLocalMicrophoneEnabled` and force the adapter to mute/unmute the call
```
JamesBurnside, line 67: I worry about a small condition where:
* I elect unmute in the configuration page
* I mute myself in the lobby
* I join the call and there is a small time where I am unmuted in the call before `adapter.mute();` has finished
  * Or perhaps there is an error in muting me and suddenly I am unmuted in the call

JamesBurnside: (the mitigation to this - probably leading to more spaghetti code - would be to always do joinCall({microphoneEnabled: false}), then when the call connects call adapter.unmute if the user was meant to be unmuted)

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -94,7 +97,14 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
 
   const options = typeof props.options === 'boolean' ? {} : props.options;
 
+  const callStatus = useSelector(getCallStatus);
+  const isLocalMicrophoneEnabled = useSelector(getLocalMicrophoneEnabled);
```
JamesBurnside, line 101: Moving this comment to here so it can be an inline conversation without having to use that annoying quote-reply:
> @prprabhu-ms Prathmesh Prabhu FTE @JamesBurnside James Burnside FTE
Couldn't find a clean way to abstract the logic away inside a component level selector in packages/react-composites/src/composites/CallComposite/components/CallControls.tsx

> Any suggestions how it can be done?

> cc: @PorterNan Nan Jiang FTE

Looking it seems the only way is to create a new microphoneButtonSelector and no longer use usePropsFor(MicrophoneButton)? Can usePropsFor work with the selectors defined in the callComposite?

anjulgarg: I couldn't see a clear way to make usePropsFor work with selectors in callComposite and handlers in component bindings...

packages/react-composites/src/composites/CallComposite/pages/CallPage.tsx
anjulgarg: Good point. However, the same condition can occur for unmuting the microphone (which I assume is safer).
This would require us to remove some code so I wouldn't worry about the increase in spaghetti code...

What do others think about this? @prprabhu-ms @PorterNan @alkwa-msft

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
anjulgarg: This is not a new behavior, I just moved the condition out of the `if` block as it should be called irrespective of the call status. :)

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
anjulgarg: Disabling microphone toggle as discussed in our standup

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -105,6 +121,7 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
 
   const screenShareButtonStyles = useMemo(
     () => mergeButtonBaseStyles(checkedButtonOverrideStyles(theme, screenShareButtonProps.checked)),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
JamesBurnside, line 124: we should avoid disabling the linting as much as possible - if someone changes code in the future it hides potential render errors (I've fixed a bug recently that was caused by this).

If `checkedButtonOverrideStyles` only needs themePrimary it should probably just take in that prop

JamesBurnside (approved):

PR #1148, opened 2021-11-23
Display a microphone tooltip in lobby informing users that it can not be toggled. 

anjulgarg:
# What
Display a microphone tooltip in lobby informing users that it can not be toggled. 

MicrosoftTeams-image (7)

# Why
Headless SDK doesn't support toggling mute/unmute while a call is in lobby.

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/Strings.tsx
+++ b/packages/react-composites/src/composites/CallComposite/Strings.tsx
@@ -169,4 +169,8 @@ export interface CallCompositeStrings {
    * More details text of the page shown to the user when there is intermittent network failure during a call.
    */
   networkReconnectMoreDetails: string;
+  /**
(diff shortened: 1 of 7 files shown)


packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -169,4 +169,8 @@ export interface CallCompositeStrings {
    * More details text of the page shown to the user when there is intermittent network failure during a call.
    */
   networkReconnectMoreDetails: string;
+  /**
+   * Tooltip text used to inform a user that toggling microphone in lobby is not supported.
+   */
+  microphoneToggleInLobbyNotAllowed: string;
```
prprabhu-ms, line 175: @PorterNan this is a trivial API change.

prprabhu-ms (approved):

alcail (approved):

Issue #824, opened 2021-09-24
"html" messages do not render in chat in ChatComposite


mmarinhomac:
All messages of type "html" are not rendered in chat!

With two chats open in parallel, sending a file through the first one, in the second chat in real time it is possible to see the message (tag a with link) that has just been sent, but when closing the chat and reopening all html messages are not rendered.

I'm sending a message like this:
```
const threadClient = new ChatThreadClient(endpointUrl, threadId, new AzureCommunicationTokenCredential(token));
threadClient.sendMessage({
  senderDisplayName: chatData.userName,
  content : `<a href='google.com' target='_blank'>Arquivo</a>`
}, { type : "html" });
```

Chat - Image:
![chat-without-links](https://user-images.githubusercontent.com/47289225/134737318-7693a082-d55d-420f-990f-84da3e7d1408.png)

HTML DevTool - Image:
![html-without-links](https://user-images.githubusercontent.com/47289225/134737342-fdb991d6-c243-47a9-8310-100a692712f1.png)

Response API - Image:
![response-api-with-links](https://user-images.githubusercontent.com/47289225/134737364-54038e59-d968-4267-9217-8e74d6693588.png)

- Windows on PC
- Browser [Google Chrome]
- Version lib beta.4 and beta.5 ❌

Version beta.3 working normally ✅

prprabhu-ms:
Thanks a lot for filing this bug. I'll take a look shortly.

PR #1149, opened 2021-11-24
Alkwa/dont render html messages

alkwa-msft:
# What
We want to avoid rendering messages from Teams when a meeting is being recorded

# Why
This content is not readable and creates a lot of noise in the chat thread. We will enable it again when we have a good way to display this information.

# How Tested
Tested locally in storybook

# Process & policy checklist

- [ ✔] I have updated the project documentation to reflect my changes if necessary.
- [ ✔] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.

This is fixing a bug for now.


packages/chat-stateful-client/src/EventSubscriber.ts
```diff
@@ -62,7 +62,14 @@ export class EventSubscriber {
   };
 
   private onChatMessageReceived = (event: ChatMessageReceivedEvent): void => {
+    // Today we are avoiding how to render these messages. In the future we can
+    // remove this condition and handle this message appropriately.
+    if (event.type.startsWith('RichText/')) {
```
anjulgarg: Any strong reason for avoiding these?
We can probably convert these to react components and render them.

alcail (approved):

anjulgarg (approved):

packages/chat-stateful-client/src/EventSubscriber.ts
alkwa-msft: There could be any number of control messages. RichText/Media_CallRecording just happens to be one of them. It is better to filter on what we are expecting (Text or RichText/Html)

prprabhu-ms (approved):

packages/chat-stateful-client/src/EventSubscriber.ts
```diff
@@ -62,7 +62,14 @@ export class EventSubscriber {
   };
 
   private onChatMessageReceived = (event: ChatMessageReceivedEvent): void => {
+    // Today we are avoiding how to render these messages. In the future we can
+    // remove this condition and handle this message appropriately.
+    if (event.type !== 'Text' && event.type !== 'RichText/Html') {
```
JamesBurnside, line 67: Do we need a condition for 'html'? https://github.com/Azure/communication-ui-library/pull/946/files

Looking to avoid breaking: https://github.com/Azure/communication-ui-library/issues/824

JamesBurnside: nit: convert the type we are checking to lowercase and use the enum `ACSKnownMessageType` we have for checking against

PorterNan, line 67: Is there concerns that we have to do it in stateful client? Since it is our component + selector's responsibility to render the message (and they don't support it well), messages filter should be happening in messageThreadSelector.tsx - In the other hand, stateful client should always provide accurate data directly coming from js sdk (so developers can read messages and render them themselves even without our implementation)

We already have some filterings there in messageThreadSelector.tsx, I wonder why it is not working, I guess there could be some leakings for our conditions there

PorterNan: Sync with Alex offline, the big limitation for now is we froze api changes so we are not able to support more types in convertEventType() function above - for short term, this change makes a lot of change before we release the lock of api changes
In long term, we will need to open a bug and add new api changes and move this logic into selector after first GA release

PorterNan (approved):

packages/chat-stateful-client/src/EventSubscriber.ts
PorterNan: Yea, 'html' type is coming from teams sometimes - not sure if that is also the case for notifications

PorterNan: Huge inconsistence of type name/lower/upper cases from chatThreadClient.getMessages() and these notifications, which should be resolved by our chat sdk in their further release

PorterNan:
According to my previous experience, 2 scenarioes might need to be tested for this PR around html type:

1. Send a message using teams
2. Send a message using chatThreadClient.sendMessage('Hello', {type: 'html'})

packages/chat-stateful-client/src/EventSubscriber.ts
alkwa-msft: Yep I am going to toLowerCase() and do checks against text richtext/html and html

all other messages are currently dropped.

PR #1150, opened 2021-11-24
Make playwright waitFor functions take a screenshot on failure

JamesBurnside:
# What
Wrap `page.waitForSelector`, `page.waitForFunction` and `page.click` in functions that limit their time and take a screenshot when they fail

# Why
We don't have good idea why flakey tests fail when they timeout - plawright just doesn't take a screenshot by default in these scenarios

# How Tested
Locally forcing a timeout on a waitForSelector successfully took a screenshot while failing the test: 
image


alcail (approved):

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -2,33 +2,76 @@
...
+export const PER_STEP_TIMEOUT_MS = 3000;
+
 /** Selector string to get element by data-ui-id property */
 export const dataUiId = (id: string): string => `[data-ui-id="${id}"]`;
 
+async function screenshotOnFailure<T>(page: Page, fn: () => Promise<T>): Promise<T> {
+  try {
+    return await fn();
+  } catch (e) {
+    await page.screenshot({ path: 'test-results/failure-screenshot.png' });
```
prprabhu-ms, line 18: I'd recommend adding a random suffix here. That way if we do somehow end up hitting this more than once in CI, we won't overwrite files.

prprabhu-ms (approved):

packages/react-composites/tests/browser/common/utils.ts
JamesBurnside: Good note, updated to `test-results/failure-screenshot-${generateGUID()}.png`

JamesBurnside: Actually, turned on `screenshot: 'only-on-failure'` in the settings, which now works because of the timeout

JamesBurnside: OK noo reverted that, it gives weird results on mismatches:
image

I don't know when these `only-on-failure` are actually taken but none are at the end....
| expected | actual | test-failed-1.png | test-failed-2.png |
| -- | -- | -- | -- |
|image|image|image|image|

JamesBurnside: reverted back to custom screenshot method (with guid for uniqueness)

PR #1153, opened 2021-11-24
Fix non-react html example

dmceachernmsft:
# What
Updated the cdn in the chat composite html example on Cross-framework page. renamed some variables to match the static html example we provide.

# Why
updates the cdn to call appropriate composite

https://skype.visualstudio.com/SPOOL/_backlogs/backlog/ACS%20Web%20UI/Features/?workitem=2653662

# How Tested
inspected the doc visually and ran the html example to make sure code is correct.


prprabhu-ms (approved):
w00t! Welcome aboard :)

JamesBurnside (approved):

anjulgarg (approved):

PR #1154, opened 2021-11-24
Run chat UI tests in order - prevent flakey localization test

JamesBurnside:
# What
Run tests in order
* They must be in the same file to deterministically do this: https://playwright.dev/docs/test-parallel#parallelize-tests-in-a-single-file

# Why
We have a flakey localization test where (I believe) sometimes it runs after the error bar tests and sometimes before:
image


prprabhu-ms:
srsly?

prprabhu-ms (approved):

JamesBurnside:
> srsly?

They really don't want tests run sequentially (which is great tbh) so they actively don't support running them intentionally sequentially, having them in the same file is just a side-effect than anything intentional

alcail (approved):

PR #1157, opened 2021-11-24
Removing additional options in storybook

alkwa-msft:
# What
We wanted to simplify the options in Storybook so developers could navigate and discover more easily without visual clutter.

# Why
We felt it was halting our documentation experience and we wanted to create a more simplified experience

# How Tested
Ran it locally

# Process & policy checklist


alkwa-msft:
Docs will be before Canvas
We are removing everything aligned on the left side except for Docs and Canvas.
We will leave all of the options on the right as they are.

Before: Docs
image

After: Docs
image

Before: Canvas
image

After: Canvas
image

packages/storybook/.storybook/preview.tsx
```diff
@@ -23,6 +23,7 @@ registerIcons({ icons: DEFAULT_COMPONENT_ICONS });
 
 export const parameters = {
   layout: 'fullscreen',
+  previewTabs: { 'storybook/docs/panel': { index: -1 } },
```
JamesBurnside, line 26: noticed when I go to a page from overview, I'm still take to the canvas (not sure if there's any way to overcome that)

alkwa-msft: There is some stickyness between the components. I will look into seeing how to reset the selection

JamesBurnside (approved):

packages/storybook/.storybook/main.js
```diff
@@ -15,10 +15,13 @@ module.exports = {
...
         actions: false,
-        controls: false
+        controls: false,
+        measure: false,
+        outline: false,
+        backgrounds: false,
+        viewport: false
       }
-    },
-    '@storybook/addon-controls'
```
alcail, line 21: this is removing all controls on Canvas, so possibility for users to play with our components/composites

packages/storybook/.storybook/main.js
```diff
@@ -15,10 +15,13 @@ module.exports = {
       name: '@storybook/addon-essentials',
       options: {
         actions: false,
-        controls: false
+        controls: false,
```
anjulgarg, line 18: How about add a url check here.
If current URL contains `/story/` we disable specific buttons and if it contains `/docs/` we disable some buttons.

alkwa-msft: we can do that so are we saying on doc pages we keep it clean and on canvas pages we have all of the bells and whistles?

packages/storybook/.storybook/main.js
alkwa-msft: oh i must have accidentally deleted it. Thanks ill put that back in!

packages/storybook/.storybook/main.js
anjulgarg: No 100% sure about that, the original idea was to clean up the header so that canvas/doc button are more prominent.
Can we push all the buttons to the right?
Decluttering top left might also make the canvas/doc buttons prominent.

alcail (approved):

packages/storybook/.storybook/main.js
alkwa-msft: It wasn't clear on how to move the positioning of the toolbar options to the right. I am in their discord right now trying to figure out how someone would do this. If the answer for now is to write my own toolbar to allow me to align it to the left or right then I will defer that for another time.

PR #1158, opened 2021-11-24
Bugfix 2631977 edit message not scrollable

anjulgarg:
# What
Add overflow to edit text box in chat

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/styles/InputBoxComponent.style.ts
```diff
@@ -13,6 +13,7 @@ export const inputBoxWrapperStyle = mergeStyles({
  * @private
  */
 export const inputBoxStyle = mergeStyles({
+  overflow: 'auto',
```
prprabhu-ms, line 16: Please add a comment here that by default, the `TextField` sets `overflow` to `hidden`, so we need to override it.

prprabhu-ms (approved):

JamesBurnside (approved):

alcail (approved):

PR #1159, opened 2021-11-24
Moved the adapter page into the composites section

dmceachernmsft:
# What
Moved the Adapter page to the Composites section.

# Why
Moved the page to a new home more related to its content.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2653668

# How Tested
visual inspection and tested the routing.


prprabhu-ms (approved):
Recommend getting sign-off from @ddematheu2

alkwa-msft (approved):

PR #1162, opened 2021-11-25
VideoGallery gap fixed for Safari on iPhone SE

mgamis-msft:
# What

Using Stack tokens to apply gap between Stack children. Safari does not seem to support flex display and the gap property.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2641774

# How Tested

Can someone with an IPhone SE please verify this works. I do not have an IPhone SE.

# Process & policy checklist


alkwa-msft:
lgtm
image

alkwa-msft (approved):

prprabhu-ms (approved):
There are a few other places we use `gap`. Can you also fix those (in this / follow up PR).

https://cs.github.com/Azure/communication-ui-library?&q=gap

PR #1163, opened 2021-11-25
Dock local video tile

mgamis-msft:
# What

Local video tile is no longer draggable therefore the VideoGallery id is removed since we are no longer using FluentUI Modal.
Box-shadow is applied when there is no HorizontalGallery.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2641808

# How Tested

Local calling sample and checked storybook docs example all look the same.

# Process & policy checklist


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -304,28 +296,16 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
-        <Modal
-          isOpen={true}
-          isModeless={true}
-          dragOptions={DRAG_OPTIONS}
-          styles={floatingLocalVideoModalStyle(theme, isNarrow)}
-          layerProps={{ hostId: containerRef.current?.id }}
-        >
-          {localParticipant && localVideoTile}
-        </Modal>
+      {shouldFloatLocalVideo && localParticipant && (
```
alcail, line 307: nit: this is not 'float' anymore but more 'isolate' or 'separate'

alcail (approved):

prprabhu-ms:
Verified that keyboard navigation works as expected.

But I'm confused by one thing: I thought we only wanted to dock the local video when horizontal gallery is being shown. When only the grid is being shown, we still wanted it to be a modal (so it can be moved out of the way if it's hiding something).

Can you confirm the UX here?

mgamis-msft:
> Verified that keyboard navigation works as expected.
> 
> But I'm confused by one thing: I thought we only wanted to dock the local video when horizontal gallery is being shown. When only the grid is being shown, we still wanted it to be a modal (so it can be moved out of the way if it's hiding something).
> 
> Can you confirm the UX here?

Yes, you are right and I confirmed it with Alex P. Updated to only dock if HorizontalGallery is showing.

packages/react-components/src/components/VideoGallery.tsx
mgamis-msft: Good point. But I confirmed with Alex P and it should still be floating when there is no HorizontalGallery. So I updated the PR.

JamesBurnside (approved):

packages/react-components/src/components/VideoGallery.tsx
JamesBurnside: agree floating is not ideal, lets land this PR and then maybe post in the WebUI channel if we want to change the enum and what we would change it to?

PR #1166, opened 2021-11-25
add tooltip strings for disabled buttons in control bar

alcail:
# What
add tooltip strings for disabled buttons in control bar

# Why
bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2655191

# How Tested
ran Calling sample locally with disabled buttons
* Camera
  image

* Microphone
  image

* Screen sharing
  image

# Process & policy checklist

add new string to APIs


JamesBurnside (approved):

mgamis-msft:
Since this is an API change I think we have to land this right after GA but please check with @PorterNan

alkwa-msft:
Managed to clean up the toolbar options. Will probably not move them to the right  for now since it didn't seem clear without injecting in custom css

alkwa-msft (approved):

PR #1167, opened 2021-11-25
Console log errors on chat tests, improve custom data model test output

JamesBurnside:
# What
Stripping these out of another PR that is to come:
* Console log errors on chat tests
* Improve custom data model test output
  * Display what the message was, but just in uppercase - instead of a completely new message

# Why
Very hard to tell what the messages where meant to be in the custom data model test in the event of failure.


alcail (approved):

packages/react-composites/tests/browser/chat/app/index.tsx
```diff
@@ -93,3 +99,23 @@ function App(): JSX.Element {
 }
 
 ReactDOM.render(<App />, document.getElementById('root'));
+
+function getMessageContentInUppercase(messageProps: MessageProps): string {
+  if (
```
prprabhu-ms, line 104: This is a tagged union:

```suggestion
  const message = messageProps.message;
  switch (message.messageType) {
    case 'chat':
    case 'custom':
    case 'content':
      return message.content.toUpperCase();
    case 'system':
      switch(message.systemMessageType) {
        case 'topicUpdated':
          return message.topic.toUpperCase();
        case 'participantAdded':
          return `Participants Added: ${message.participants.map((p) => p.displayName).join(',')}`.toUpperCase();
        case 'participantRemoved':
          return `Participants Removed: ${message.participants.map((p) => p.displayName).join(',')}`.toUpperCase();
        default:
          return 'CUSTOM SYSTEM MESSAGE';
      }
      default:
        return 'CUSTOM MESSAGE';
  }
```

prprabhu-ms (approved):

packages/react-composites/tests/browser/chat/app/index.tsx
JamesBurnside: Thanks, updated

PR #1171, opened 2021-11-26
Add custom icon documentation to storybook

anjulgarg:
# What
Add custom icon documentation to storybook
image

# Why

# How Tested

# Process & policy checklist


packages/storybook/stories/CustomIcons/CustomIcons.stories.mdx
```diff
@@ -0,0 +1,73 @@
...
+  icons={{
+    LocalDeviceSettingsCamera: <p>My Custom Icon</p>
+  }}
+/>
+  `}
+/>
+
+### Default Composite Icons
+
+Below is a list of all the icons you can override Composites with their names.
```
alcail, line 32: nit: ... all the default icons ... override in Composites or Components ...

packages/storybook/stories/CustomIcons/CustomIcons.stories.mdx
```diff
@@ -0,0 +1,73 @@
...
+const customIcons = {...DEFAULT_COMPONENT_ICONS, {
+  ControlButtonCameraOff: <p>My Custom Icon</p>
+}};
+registerIcons({ icons: customIcons });
+  `}
+/>
+
+### Default Component Icons
+
+Below is a list of all the icons used inside Components with their names.
```
alcail, line 69: nit ... all the default icons ...

alcail (approved):

JamesBurnside:
Can you move Custom Icons to maybe below Theming? (I feel it relates strongly to styling and theming):
image

packages/storybook/stories/CustomIcons/utils.tsx
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { mergeStyles, Stack, TextField } from '@fluentui/react';
+import React, { useState } from 'react';
+
+const iconGrid = (allIcons: Record<string, JSX.Element>, searchText?: string): JSX.Element[] => {
+  const allIconKeys = Object.keys(allIcons);
+  const filteredKeys = searchText
```
JamesBurnside, line 8: Search feature is such fun!!

packages/storybook/stories/CustomIcons/utils.tsx
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft Corporation.
```
JamesBurnside, line 1: nit: filename as `IconGrid` instead of utils?

packages/storybook/stories/CustomIcons/utils.tsx
```diff
@@ -0,0 +1,65 @@
...
+// Licensed under the MIT license.
+import { mergeStyles, Stack, TextField } from '@fluentui/react';
+import React, { useState } from 'react';
+
+const iconGrid = (allIcons: Record<string, JSX.Element>, searchText?: string): JSX.Element[] => {
```
JamesBurnside, line 6: nit: this makes sense to be a react element
```suggestion
const IconGrid = (props: { allIcons: Record<string, JSX.Element>, searchText?: string }): JSX.Element[] => {
```

then usage is 
```
<IconGrid icons={icons} searchText={searchText} />
```

packages/storybook/stories/CustomIcons/utils.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return filteredKeys.map((key, idx) => {
+    return (
+      <Stack
+        key={idx}
+        horizontal
+        style={{
+          padding: '0.5rem 0'
+        }}
+      >
+        <Stack horizontalAlign="start" className={mergeStyles({ span: { fontSize: '1rem' } })}>
```
JamesBurnside, line 21: name and icon aren't quite centered vertically here:
image

For beta banner I had the same issue, this worked for me:
image

packages/storybook/stories/CustomIcons/utils.tsx
```diff
@@ -0,0 +1,65 @@
...
+  );
+};
+
+export const IconGridWithSearch = (props: { icons: Record<string, JSX.Element> }): JSX.Element => {
+  const { icons } = props;
+  const [searchText, setSearchText] = useState<string | undefined>(undefined);
+  return (
+    <Stack>
+      <SearchBar onChange={(val) => setSearchText(val)} />
+      <div style={{ display: 'grid', gridTemplateColumns: 'auto auto auto', padding: '1rem 0' }}>
```
JamesBurnside, line 60: I keep going to read the icons top-bottom then left-right. Can you confirm which is the preferred UX?
I like the argument here that because there is more space between columns then rows, and because we read grouped things together first, the columns appear to us as a grouping and thus we read top->bottom: https://stackoverflow.com/a/3250389/16089184

However, jury is definitely out on this one - so don't mind going with either

If you did want to sort top->bottom then left->right the fluent stack can do this; see "Vertical Stack - Wrapping - Nested" in https://developer.microsoft.com/en-us/fluentui#/controls/web/stack

packages/storybook/stories/CustomIcons/utils.tsx
JamesBurnside: Though if I'm being super nitty the search function would be better as an extracted fn on its own and do
```
<IconGrid icons={filterIcons(icons, searchText)} />
```
But I won't block on that :D

JamesBurnside: Although this in its current form isn't a "grid" as the name suggestions - its just an array of icons. Would be best to move the `<div style={{ display: 'grid', gridTemplateColumns: 'auto auto auto', padding: '1rem 0' }}>` into here and have
```
<Stack>
  <SearchBar onChange={(val) => setSearchText(val)} />
  <IconGrid icons={filterIcons(icons, searchText)} />
</Stack>
```
below

packages/storybook/stories/CustomIcons/utils.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { mergeStyles, Stack, TextField } from '@fluentui/react';
+import React, { useState } from 'react';
+
+const iconGrid = (allIcons: Record<string, JSX.Element>, searchText?: string): JSX.Element[] => {
+  const allIconKeys = Object.keys(allIcons);
+  const filteredKeys = searchText
+    ? allIconKeys.filter((k) => k.toLowerCase().includes(searchText.toLowerCase()))
+    : allIconKeys;
+
+  return filteredKeys.map((key, idx) => {
```
JamesBurnside, line 12: Can you also apply a sort() here? Looks like there's a chunk that aren't in order:
image

packages/storybook/stories/CustomIcons/utils.tsx
anjulgarg: Here is possibly the most famous icon library out there. Took inspiration from it.
https://fontawesome.com/v4.7/icons/
It is following Left -> Right and then Top -> Bottom
I personally read them left to right as intended. Pretty sure once a user looks at them, they will identify the pattern and wouldn't find it hard to navigate.

That's another reason I added search

packages/storybook/stories/CustomIcons/utils.tsx
anjulgarg: Fixed

packages/storybook/stories/CustomIcons/utils.tsx
anjulgarg: Fixed. A simple fix is to wrap everything in Stack with verticalAlign center and give a line height of 0 to the icon ;)

packages/storybook/stories/CustomIcons/utils.tsx
anjulgarg: I realised the flaws. Fixed everything in the refactor. Take a look.

alkwa-msft:
I think we should be leading this doc with why this page exists

- What are developers looking for?
- What are some of the use cases they have to change these icons?
- We should state why we are supporting icon customization?

image

When I do the LocalDeviceSettingsCamera icon will it wipe away all other icons?
> Is there a best practices we can share here in our sample code?

It looks like there are a lot of similar icons between components and composites but we just show a separate list.
Do we know why? Should we address why the component icons differ from the composite icons?

For the registerIcons function this is probably really easy to miss compared to setting up icons
in composites. Have we thought about this? How do we make sure they are always calling this function
before rendering?

packages/storybook/stories/CustomIcons/CustomIcons.stories.mdx
```diff
@@ -0,0 +1,73 @@
+import { DEFAULT_COMPONENT_ICONS, DEFAULT_COMPOSITE_ICONS } from '@azure/communication-react';
+import { Meta, Canvas, Source } from '@storybook/addon-docs';
+import { IconGridWithSearch } from './IconGrid';
+
+<Meta id="custom-icons" title="Concepts/Custom Icons" />
```
JamesBurnside, line 5: I think I would just name and title this "Icons". It talks through both all the default icons, and the customizing icons

JamesBurnside (approved):

packages/storybook/stories/CustomIcons/CustomIcons.stories.mdx
anjulgarg: I changed the titles 10 times at least haha.
If we change it to `Icons` I think the layout of this doc page would need to change as well.

PR #1172, opened 2021-11-26
aria label missing in Call composite configuration page

alcail:
# What
replace mic label by a common label for both mic and speakers dropdown

# Why
bug: https://skype.visualstudio.com/SPOOL/_workitems/edit/2565537

# How Tested
Ran calling sample locally and checked with Accessibility Insights
image

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/components/LocalDeviceSettings.tsx
+++ b/packages/react-composites/src/composites/CallComposite/components/LocalDeviceSettings.tsx
@@ -125,7 +125,7 @@ export const LocalDeviceSettings = (props: LocalDeviceSettingsType): JSX.Element
         <Label
           id={'call-composite-local-sound-settings-label'}
           className={mergeStyles(dropDownStyles(theme).label)}
-          disabled={!props.microphonePermissionGranted || props.speakers.length === 0} // depends on dropdowns disabled state in the sound settings stack
+          disabled={!props.microphonePermissionGranted} // follows Start button disabled state in ConfigurationPage
         >
           {soundLabel}
         </Label>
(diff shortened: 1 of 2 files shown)


JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/LocalDeviceSettings.tsx
```diff
@@ -121,52 +121,64 @@ export const LocalDeviceSettings = (props: LocalDeviceSettingsType): JSX.Element
...
-        ) => {
-          props.onSelectSpeaker(props.speakers[index ?? 0]);
-        }}
-        onRenderTitle={(props?: IDropdownOption[]) => onRenderTitle('Speaker', props)}
-      />
+      <Stack>
+        <Label
+          id={'call-composite-local-sound-settings-label'}
+          className={mergeStyles(dropDownStyles(theme).label)}
+          disabled={!props.microphonePermissionGranted || props.speakers.length === 0} // depends on dropdowns disabled state in the sound settings stack
```
prprabhu-ms, line 128: Shouldn't this be `&&`.

The label should be disabled if *both* `DropDown` are disabled.

prprabhu-ms (approved):

packages/react-composites/src/composites/CallComposite/components/LocalDeviceSettings.tsx
alcail: why? you can have mic but not speakers, of vise-versa.
And the only condition for "Start" button is if mic permission is granted, so I should not even test the speakers if I really want to follow the "Start" button behaviour

PR #1174, opened 2021-11-27
highlight docs first and hiding canvas when possible

alkwa-msft:
# What
Renaming canvas to "Try it out"
Removing canvas when on an mdx page
Making docs be the first initially selected option, by making the first page in storybook an mdx page and hiding canvas on that page.

# Why
Developers are having challenges finding the documentation when the canvas is first

# How Tested
Test locally

# Process & policy checklist


packages/storybook/stories/Accessibility.stories.mdx
```diff
@@ -1,6 +1,10 @@
 import { Meta } from '@storybook/addon-docs';
 
-<Meta id="accessibility" title="Concepts/Accessibility" />
+<Meta
+  id="accessibility"
+  title="Concepts/Accessibility"
+  parameters={{ previewTabs: { canvas: { disable: true, hidden: true } } }}
```
JamesBurnside, line 6: nit: this is used in every pure docs file, can this be moved to a constant and imported in case it needs changed in the future
```suggestion
import { docsOnlyTabs } from '../constants';
...
  parameters={docsOnlyTabs}
```

packages/storybook/.storybook/preview.tsx
```diff
@@ -90,7 +89,14 @@ export const parameters = {
         ],
       ]
     }
-  }
+  },
+  viewMode: 'docs',
+  previewTabs: {
+    'storybook/docs/panel': { index: -1 },
+    'canvas' : {
+      title: 'Try it out'
```
JamesBurnside, line 97: (completely ignorable) suggestion: `Live Example` over Try it out?

packages/storybook/stories/Accessibility.stories.mdx
alkwa-msft: great suggestion will get on that!

packages/storybook/.storybook/preview.tsx
alkwa-msft: Sounds like a good discussion topic

JamesBurnside (approved):
Love this! Only concern here is that nobody will notice the new Canvas option suddenly appearing (edit: actually telemetry can tell us this!), and its an odd experience for an option to suddenly appear depending on the page you are on. But I think its a great solution we should definitely check in and try out!

alcail (approved):

packages/storybook/.storybook/preview.tsx
alkwa-msft: changing to "Preview"

packages/storybook/stories/Accessibility.stories.mdx
alkwa-msft: https://github.com/storybookjs/storybook/pull/11752/files

image

It looks like we need to manually pass in parameters to our stories.mdx files

PR #1176, opened 2021-11-29
Dmceachernmsft/stylingUpdates

dmceachernmsft:
# What
Updated styles to Match Alex P's demo video

# Why
Makes the documentation more readable.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2656093

# How Tested
Visually inspect the page and css to make sure are correct.


anjulgarg:
Too much gap above the heading.
image

Inconsistent with other doc pages.
image

anjulgarg:
Font weight looks different compared to the demo video
image

dmceachernmsft:
> Font weight looks different compared to the demo video image

for which font set in particular?

anjulgarg:
> > Font weight looks different compared to the demo video image
> 
> for which font set in particular?

<p/> blocks

packages/storybook/.storybook/preview-head.html
```diff
@@ -0,0 +1,78 @@
+<style>
+    .sbdocs.sbdocs-h1:first-of-type {
+        margin-top: 64px;
+    }
+
```
anjulgarg, line 5:
```suggestion
.sbdocs.sbdocs-h1, .sbdocs.sbdocs-h2, .sbdocs.sbdocs-h3, .sbdocs.sbdocs-p {
    font-family: "Segoe UI", "Helvetica Neue", Helvetica, Arial, sans-serif;
    color: #000000;
}
```

packages/storybook/.storybook/preview-head.html
```diff
@@ -0,0 +1,78 @@
...
+    tr:nth-of-type(2n) {
+        background-color: #faf9f8;
+    }
+
+    .sbdocs.sbdocs-wrapper {
+        padding: 4rem 64px;
+    }
+
+    body {
```
anjulgarg, line 69: If we are assigning a font family to the body, is it not being applied to all the h1, h2, p blocks?

dmceachernmsft: the stylesheet provided by story book is overwriting it for the h1 h2 h3 blocks I believe, I should remove the body CSS I believe since I think its getting overridden.

anjulgarg: I looked at the styleset, I believe you can optimize it a bit :)

dmceachernmsft: okay will take a look! 😊

packages/storybook/.storybook/preview-head.html
dmceachernmsft: thanks for this! made the change. 😊

packages/storybook/.storybook/preview.tsx
```diff
@@ -16,11 +16,14 @@ import { THEMES } from '../stories/themes';
 import { LOCALES } from '../stories/locales'
 import { TOC } from './TOC';
 
+
```
alcail, line 19: nit: no need for extra empty lines, just one to separate blocks is enough :)

dmceachernmsft: oh oops! I'll clear that 😊

packages/storybook/.storybook/preview-head.html
```diff
@@ -0,0 +1,100 @@
...
+    .sbdocs.sbdocs-p {
+        font-family: "Segoe UI", "Helvetica Neue", Helvetica, Arial, sans-serif;
+        color: #000000;
+    }
+
+    .sbdocs.sbdocs-h1,
+    [class*="sbdocs-title css-"],
+    .sbdocs.sbdocs-h1:first-of-type {
+        color: #000000;
+        font-size: 40px;
```
alcail, line 19: avoid using 'px', instead use 'rem' (with 1rem being 16px by default)
we can use 'em' also but we have to be aware of the nesting using 'em' implies :)

dmceachernmsft: ah okay, I was using the styles Alex P sent over, will do the math to make that work!

packages/storybook/.storybook/preview.tsx
alcail: no worries :)

anjulgarg (approved):

packages/storybook/.storybook/preview-head.html
```diff
@@ -0,0 +1,104 @@
...
+    table.docblock-argstable-body:nth-of-type(2n) {
+        background-color: #faf9f8 !important;
+    }
+
+    .sbdocs.sbdocs-preview,
+    .docblock-argstable-body,
+    .docblock-argstable-head,
+    .docblock-source {
+        box-shadow: none !important;
+        border: 1px solid rgba(0, 0, 0, .1);
```
alcail, line 78: nit: rem, please ;)

alcail (approved):

PR #1178, opened 2021-11-29
Fix control bar button label styling. Fix screenshare button hover and checked styling.

JamesBurnside:
# What

1. Fix background color and label of screenshare button
  * **NOTE: MOVED STYLING FROM COMPOSITES TO COMPONENT**
2. Fix our control bar buttons so that the label can be styled correctly
  * **Simplified our label code** - removed our custom label rendering logic, please review

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2633102

# How Tested
* Locally tested:
  * image
  * image
  * image
  * image

* Will check UI tests once they complete.
* Tested keyboard nav 
  * @alcail please comment any extra accessibility tests to check


packages/react-components/src/components/ScreenShareButton.tsx
```diff
@@ -67,3 +71,18 @@ export const ScreenShareButton = (props: ScreenShareButtonProps): JSX.Element =>
     />
   );
 };
+
+const screenshareButtonStyles = (theme: Theme): IButtonStyles => ({
```
JamesBurnside, line 75: NOTE: Now in components and not composites

packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -154,9 +133,11 @@ export const ControlBarButton = (props: ControlBarButtonProps): JSX.Element => {
       <DefaultButton
         {...props}
         styles={componentStyles}
-        onRenderText={props.showLabel ? props.onRenderText ?? DefaultRenderText : undefined}
+        onRenderText={props.showLabel && props.onRenderText ? props.onRenderText : undefined}
         onRenderIcon={props.onRenderIcon ?? DefaultRenderIcon}
-      />
+      >
+        {props.showLabel ? labelText : <></>}
```
JamesBurnside, line 139: NOTE: No longer using a custom render function (so get the styling from props correctly)

alcail (approved):

alcail:
just make sure that when focused on (navigated to through keyboard) the outline is indeed white and Narrator indeed says "Stop sharing button", but code seems great so should be no issue

change/@internal-react-composites-90f8fc03-9c00-426e-a0d9-44028a8259ed.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Moved code",
```
anjulgarg, line 3: Can we be specific in this comment about which code was removed?
Helps when we go over the combined changelog later.

packages/react-components/src/components/ScreenShareButton.tsx
```diff
@@ -67,3 +71,18 @@ export const ScreenShareButton = (props: ScreenShareButtonProps): JSX.Element =>
...
+  rootChecked: {
+    background: theme.palette.themePrimary,
+    color: DefaultPalette.white,
```
anjulgarg, line 78: curious about the usage of `DefaultPalette` for color instead of `theme`?

anjulgarg (approved):

change/@internal-react-composites-90f8fc03-9c00-426e-a0d9-44028a8259ed.json
JamesBurnside: This is a no-op for this package (nothing changes from the perspective of the composites package) so type is 'none'. When the changelog is generated it omits these

packages/react-components/src/components/ScreenShareButton.tsx
JamesBurnside: I left this as is (this code was just moved from one file to another) but also curious on this too if anyone knows why default palette is used

change/@internal-react-composites-90f8fc03-9c00-426e-a0d9-44028a8259ed.json
anjulgarg: Oh cool. Didn't know that.

packages/react-components/src/components/ScreenShareButton.tsx
anjulgarg: I think @PorterNan worked on this originally.

JamesBurnside:
> Awesome. I just checked to make sure that when focused on (navigated to through keyboard) the outline is indeed white and Narrator indeed says "Stop sharing button", and perfect :)

Confirmed both, thanks!

PR #1179, opened 2021-11-29
Html meta tag for google & bing domain search verification

anjulgarg:
# What
Html meta tag for google domain search verification.
We would need to run verification on google search console once we have merged this with main and deployed it to azure.github.io/communication-ui-library

# Why
To enable google to crawl and index storybook website.

# How Tested

# Process & policy checklist


prprabhu-ms (approved):

alcail (approved):

packages/storybook/.storybook/manager-head.html
```diff
@@ -2,6 +2,8 @@
 <meta name="description"
   content="Azure Communication Services UI Library makes it easy for you to build modern communications user experiences using Azure Communication Services. It gives you a library of production-ready UI components that you can drop into your applications.">
 
+<meta name="google-site-verification" content="RLP3tg6OFhUU6_Q7EpOLm4GWjlBr3aYxkETkWLT6OfE" />
```
JamesBurnside, line 5: Are we doing something similar for bing also?

JamesBurnside (approved):

packages/storybook/.storybook/manager-head.html
anjulgarg: Thanks for the reminder. Added bing meta as well

PR #1183, opened 2021-11-30
Workaround for incomplete thread

PorterNan:
There is a regression in api side for any pagesize < 50, there could be missing message when thread size > 200, this workaround sets default pagesize = 50

* The fix from api side will be rolled out around Dec 15th

- Correct the default number of messages to load for component layer

# What

# Why

# How Tested

# Process & policy checklist


packages/chat-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -71,7 +71,7 @@ export const createDefaultChatHandlers = memoizeOne(
         if (messageIterator === undefined) {
           // Lazy definition so that errors in the method call are reported correctly.
           // Also allows recovery via retries in case of transient errors.
-          messageIterator = chatThreadClient.listMessages();
+          messageIterator = chatThreadClient.listMessages({ maxPageSize: 50 });
```
alkwa-msft, line 74: should we be allowing this to be configured by contoso and we can default to 50?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -601,7 +601,7 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
     disableJumpToNewMessageButton = false,
     showMessageDate = false,
     showMessageStatus = false,
-    numberOfChatMessagesToReload = 0,
+    numberOfChatMessagesToReload = 5,
```
alkwa-msft, line 604: how is numberOfChatMessagesToReload related to maxPageSize?

PorterNan: This one if for rendering and maxPageSize is for network request, this one is more perf related parameter

packages/chat-component-bindings/src/handlers/createHandlers.ts
PorterNan: We can talk about this later after GA - no matter how many messages our customer set, the chat thread behavior will still fetch new messages when you scrolling up, and fullfill the thread until the minimum scroll top gap is satisfied - it's more like an internal implementation for composite, so I would say we might not want to expose this api until our customers have a real requirement on it

JamesBurnside (approved):

packages/react-components/src/components/MessageThread.tsx
anjulgarg: Why was this 0 before?

alcail (approved):

PR #1185, opened 2021-11-30
Quick fix chat sample home&config screens for mobile (for bug bash)

JamesBurnside:
# What
Super quickly fixup the chat home and config screen for mobile devices
(just removed problematic css and ensured we use fluent components to dictate layout)

# Why
We don't want erroneous bugs filed on screens we don't care about

# How Tested
| | before (iPhone SE) | after (iPhone SE) |
| -- | -- | -- |
| home | image | image |
| config |  image | image |

Non mobile is unchanged:
| home | config |
| -- | -- |
| image|image|

### DIFF
--- a/samples/Chat/src/app/ConfigurationScreen.tsx
+++ b/samples/Chat/src/app/ConfigurationScreen.tsx
@@ -21,7 +21,6 @@ import {
   leftPreviewContainerStackTokens,
   leftPreviewContainerStyle,
   namePreviewStyle,
-  responsiveLayoutStackTokens,
   responsiveLayoutStyle,
   rightInputContainerStackTokens,
   rightInputContainerStyle,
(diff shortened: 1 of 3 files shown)


alcail:
what happen if we have a small window in desktop, or at 400%?

alcail:
Could you test zooming in at 400% for screen  resolution 1280*768?

JamesBurnside:
> Could you test zooming in at 400% for screen resolution 1280*768?
Seems OK (large avatar is just a little clipped - but at such a small screen real estate we should really be hiding that whole thing):

https://user-images.githubusercontent.com/2684369/144119155-0f53409a-9b0e-4ed8-9940-a261b61a9c4e.mp4

alcail:
this is not okay for Configuration page though
At the top of the page and heading, is not displayed, as well as selected avatar is cropped

image

JamesBurnside:
> this is not okay for Configuration page though At the top of the page and heading, is not displayed, as well as selected avatar is cropped

Had a look - this is caused because we have a `<div class="wrapper>"` wrapped around our chat app that has height set to 100% (which is bad as anything flowing beyond 100% then isn't being taken into account for).

I was just doing this quickly for the bug bash (and deployed off this branch) to prevent folks filing bugs on non-chat screens so just gonna close this as I don't want to spend time on it - this can be picked up again if we are polishing samples

PR #1189, opened 2021-12-01
Publish alpha package off feature branch

prprabhu-ms:
# What

* Stop publishing nightly packages off of `main`.
* Add another nightly workflow to publish nightly packages off of `features/conditional-compilation` branch (to get beta features).

# Why

We've removed all beta features from `main` in preparation for GA. But we'd like our alpha packages to include those.

# How Tested

Prayed to the GH gods.


alkwa-msft:
It will be interesting to put a build from this conditional branch into self hosting to see if there are any unforseen bugs at runtime?

.github/workflows/nightly-ci.yml
```diff
@@ -103,44 +105,3 @@ jobs:
       # Important to check version consistency again after bumping versions.
```
anjulgarg, line 105: Curious why we don't merge this consistent version check with the new build file

prprabhu-ms: Don't understand. The comment only refers to the first step of the job.

I'm keeping the `nightly-ci` workflow around, so that we still get nightly build/test/... coverage on `main`, but without the publish steps.

anjulgarg:
Maybe it's just me but the filename `nightly-ci-beta` made me think that it was a beta release pipeline.

anjulgarg (approved):

prprabhu-ms:
> Maybe it's just me but the filename `nightly-ci-beta` made me think that it was a beta release pipeline.

Renamed.

alcail (approved):

.github/workflows/nightly-ci-alpha-release.yml
```diff
@@ -0,0 +1,152 @@
...
+  checkForChanges:
+    name: Check for new changes
+    outputs:
+      hasChanged: ${{ steps.checkChange.outputs.hasChanged }}
+    runs-on: ubuntu-latest
+    steps:
+      # Check-out repo
+      - uses: actions/checkout@v2
+        with:
+          ref: feature/conditional-compilation
```
JamesBurnside, line 27: Is there value in this being a new workflow? (and not just put this line `ref: feature/conditional-compilation` in the existing file?) We do CI after every PR to main so I don't think we want a nightly CI as well (really the `nightly-ci` workflow should be renamed `nightly-cd`)

.github/workflows/nightly-ci-alpha-release.yml
```diff
@@ -0,0 +1,152 @@
...
+
+      # Get datetime
+      - name: Get datetime for alpha release name
+        id: datetime
+        run: |
+          echo "::set-output name=datetime::$(date +'%Y%m%d.%H%M')"
+
+      # Bump alpha package versions
+      - name: Bump alpha package versions
+        run: node common/config/node_modules/beachball/bin/beachball canary --canary-name alpha-${{ steps.datetime.outputs.datetime }} --tag dev --no-publish
```
JamesBurnside, line 96: Apply https://github.com/Azure/communication-ui-library/pull/1192 here (change `...alpha-...` to `...alpha+...`)

.github/workflows/nightly-ci-alpha-release.yml
prprabhu-ms: Yeah, I realize we don't actually get any extra coverage in the nightly CI (!).

prprabhu-ms:
See the simpler https://github.com/Azure/communication-ui-library/pull/1193 instead.

PR #1196, opened 2021-12-02
GridLayout refactor

mgamis-msft:
# What

GridLayout refactor to avoid DOM tree recalculation if the `GridLayout` is resized to changed rows and columns
But also I realized that I forgot to change the type of prop `GridLayoutProps.styles` to `GridLayoutStyles`.

# Why

To potentially avoid disposing video streams when resizing

# How Tested

For Nan to test :)

# Process & policy checklist


packages/communication-react/review/communication-react.api.md
```diff
@@ -1218,7 +1218,7 @@ export const GridLayout: (props: GridLayoutProps) => JSX.Element;
 export interface GridLayoutProps {
     // (undocumented)
     children: React_2.ReactNode;
-    styles?: BaseCustomStyles;
+    styles?: GridLayoutStyles;
```
PorterNan, line 1221: I guess GridLayoutStyles is like a super set of BaseCustomStyles, so this is not a breaking change? If so, let's make this line post-GA because api has been frozen, especially for changes could be resolved after GA

mgamis-msft: Yeah we can make this post GA since it's not a breaking change. Will revert.

packages/react-components/src/components/GridLayout.tsx
```diff
@@ -67,10 +67,13 @@ export const GridLayout = (props: GridLayoutProps): JSX.Element => {
...
 
   return (
-    <div ref={containerRef} className={gridLayoutStyle}>
-      <BiGrid gridProps={gridProps} styles={styles}>
-        {children}
-      </BiGrid>
+    <div
+      ref={containerRef}
+      className={mergeStyles(gridLayoutStyle, createGridStyles(numberOfChildren, gridProps), styles?.root, {
+        '> *': styles?.children
```
JamesBurnside, line 73: we have our own styles on both sides of the `styles?.root`, when the merge happens - do params on the right hand side take precedence over params to the left if they change the same css property?
If we so can we keep all our styles on the left to ensure contoso can overwrite what they like

JamesBurnside: (and can this be merged into gridLayoutStyles?)

mgamis-msft: Yes, params on the right hand side take precedence. `styles?.root` and `styles?.children` are  GridLayoutStyles which contoso defines to override our styles. I guess it looks weird that I had to separate `styles?.root` and `styles?.children` but I could not think of another notation to have them combined.

JamesBurnside: Oh I totally missed this is passing down contoso styles, your code is correct ignore me! ^_^

JamesBurnside (approved):

PorterNan (approved):

PR #1198, opened 2021-12-03
Jinan/fix empty tile

PorterNan:
# What

- Dispose video when video tile gets unmounted 

- Add extra condition to ensure video always get disposed
- This PR still keeps the fixed behavior from #1175 

And this PR should be checked in after #1196 to make sure video shuffling won't cause any re-rendering, how it looks after #1196:
fix

Some more works need to be resolved after GA:
In power saving mode of android, when mobile browser running in background/or when the phone is in sleep/doze mode, the websocket event wouldn't be received anymore and there could be still empty tiles (different bugs which needs to be resolved after GA)
We need some kinda de-sync detection from call sdk to see if things are not up-to-date for the problem described above

Current logic of disposing - which should be decoupled after GA by adding a new optional 'flavor' param to dispose function:
1. No screen share + VideoStream on - RemoteVideoTile will dispose streams
2. screen share + VideoStream on - RemoteVideoTile won't dispose streams, but screenShareTile will
3. screen share + VideoStream off - screenShareTile will still dispose streams, but it doesn't matter because we are relocating the entire layout of VideoGallery, means all videoStreams will be re-created

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/VideoGallery/RemoteScreenShare.tsx
```diff
@@ -44,11 +35,10 @@ export const RemoteScreenShare = React.memo(
 
     useEffect(() => {
       return () => {
-        if (isVideoStreamOn) {
-          onDisposeRemoteStreamView && onDisposeRemoteStreamView(userId);
-        }
+        // TODO: Isolate disposing behaviors for screenShare and videoStream
```
anjulgarg, line 38: I don't understand this change.
What would happen when this component tries to call onDisposeRemoteStreamView but there is no remoteVideoStream?

anjulgarg:
Could you talk a bit more about `flavour` param and what it would look like?

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -45,25 +45,35 @@ export const RemoteVideoTile = React.memo(
       if (isAvailable && !renderElement) {
         onCreateRemoteStreamView && onCreateRemoteStreamView(userId, remoteVideoViewOptions);
       }
-      if (!isAvailable) {
-        onDisposeRemoteStreamView && onDisposeRemoteStreamView(userId);
-      }
+      // Always clean up element to make tile up to date and be able to dispose correctly
+      // TODO: Add an extra param to onDisposeRemoteStreamView(userId, flavor(optional)) when api lock is released
```
prprabhu-ms, line 49: nit: Just say "after GA". "when api lock is released" is very... enigmatic.

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -45,25 +45,35 @@ export const RemoteVideoTile = React.memo(
...
+      // and isolate dispose behavior between screen share and video
+      return () => {
+        if (renderElement && !isScreenSharingOn) {
```
prprabhu-ms, line 52: I'm jittery with this logic on two accounts.

1. What happens if `isScreenSharingOn` is set after this component mounts. In particular, consider -> mount (useEffect calls onCreateRemoteStreamView) -> isScreenSharingOn is set to true -> useEffect cleanup is noop -> useEffect *again* calls onCreateRemoteStreamView -> renderElement gets populated -> useEffect cleanup is noop -> useEffect is noop -> a noop useEffect is registered (because `isScreenSharingOn`). In this case, this component called `onCreateRemoteStreamView` twice, and will not cleanup ever. The ScreenShareTile _should_ cleanup.
2. The cleanup registered in `useEffect` below also somewhat races with this useEffect. In the example sequence of events above, when `isScreenSharingOn` changes, the cleanup for that useEffect will run, and call onDisposeRemoteStreamView.... 

In general, it's scary when `isScreenSharingOn` changes with this component already mounted.

prprabhu-ms (approved):
I'm still strongly in favour of hoisting all logic for stream creation / disposal up (to a dedicated hidden component?) in VideoGallery. Currently this logic is spread across components in a way that gives me very little confidence in the integration.

OK as a hotfix pre GA.

packages/react-components/src/components/VideoGallery/RemoteScreenShare.tsx
PorterNan: It's totally fine, you can call it hundred times, and it is just doing a check in the map object

packages/react-components/src/components/RemoteVideoTile.tsx
PorterNan: If the isScreenSharingOn is changed in that way, the screenShare will be disposed first, then it will cause the re-render of screenShareTile, then screenShare tile will notice renderElement is undefined, then it will create something for itself - it happens very rare, but once we have isolated function for screenShare and videoStream, we won't need this strange logic anymore

PorterNan:
> I'm still strongly in favour of hoisting all logic for stream creation / disposal up (to a dedicated hidden component?) in VideoGallery. Currently this logic is spread across components in a way that gives me very little confidence in the integration.
> 
> OK as a hotfix pre GA.

Agree that this is a hotfix, the logic is spreading across components - the main conflict I can see here is we are disposing videoStream and screenShare in just one function - which breaks the natural lifecycle binding between videoTile and screenShareTile and make this so tricky, once we could dispose only videoStream or screenShare, 2 logics will be entirely unrelated, then let's compare this true solution with hoisting logic in parent at that time

PorterNan:
> Could you talk a bit more about `flavour` param and what it would look like?

currently we have onDisposeRemoteStreamView(userId) which dispose all video streams under a user - which the main problem maker - when a videoStream turns off, the screenShare shouldn't get disposed at the same time
So the flavor would be an optional parameter at the end of this function to determine which stream you wanna dispose - videoStream or screenShareStream, it will entirely separate the nasty logic between RemoteVideoTile dispose and RemoteScreenShareTile dispose

PorterNan:
> > Could you talk a bit more about `flavour` param and what it would look like?
> 
> currently we have onDisposeRemoteStreamView(userId) which dispose all video streams under a user - which the main problem maker - when a videoStream turns off, the screenShare shouldn't get disposed at the same time So the flavor would be an optional parameter at the end of this function to determine which stream you wanna dispose - videoStream or screenShareStream, it will entirely separate the nasty logic between RemoteVideoTile dispose and RemoteScreenShareTile dispose

FYI, @mgamis-msft since you also mentioned this in your screenShare fix PR

mgamis-msft:
I could be wrong but isn't returned callbacks in `useEffect` blocks only executed once when the component dismounts? I am wondering if we need 3 `useEffect` blocks that all dispose remote stream on dismount which only happens once. You definitely played around with it more but just wondering if the first useEffect will suffice?

mgamis-msft:
> > > Could you talk a bit more about `flavour` param and what it would look like?
> > 
> > 
> > currently we have onDisposeRemoteStreamView(userId) which dispose all video streams under a user - which the main problem maker - when a videoStream turns off, the screenShare shouldn't get disposed at the same time So the flavor would be an optional parameter at the end of this function to determine which stream you wanna dispose - videoStream or screenShareStream, it will entirely separate the nasty logic between RemoteVideoTile dispose and RemoteScreenShareTile dispose
> 
> FYI, @mgamis-msft since you also mentioned this in your screenShare fix PR

Adding a 'flavor' param to the onDisposeRemoteStreamView function to choose whether to dispose all streams or just video/screenshare would definitely help fix the problem. Or we could just have another function to dispose just screenshare and just video.

PorterNan:
> I could be wrong but isn't returned callbacks in `useEffect` blocks only executed once when the component dismounts? I am wondering if we need 3 `useEffect` blocks that all dispose remote stream on dismount which only happens once. You definitely worked with remote streams more but just wondering if the first useEffect is enough?

That was the first version I wrote with 1 useEffect (then I came up with this which has 2 useEffect)

And your feeling is right about it - in most of cases (I would say more than 99.9%) only one useEffect will work perfectly

There is a very rare case here (like what I described in the code comment):
The normal flow is like this:
create videoTile -> createView -> (small time gap)-> renderElement exists -> tile dispose -> if (element) dispose();

But the rare case happens like this, when a tile gets created and immediately disposed:
create videoTile -> createView -> tile dispose -> if (element) dispose();

And then because element doesn't get enough time to be created yet, it fails to dispose the view even it has started being created

The second useEffect is used for filling that small time gap for very rare condition - thing will not be too bad without it, but still nice to guard all rare cases

mgamis-msft:
> > I could be wrong but isn't returned callbacks in `useEffect` blocks only executed once when the component dismounts? I am wondering if we need 3 `useEffect` blocks that all dispose remote stream on dismount which only happens once. You definitely worked with remote streams more but just wondering if the first useEffect is enough?
> 
> That was the first version I wrote with 1 useEffect (then I came up with this which has 2 useEffect)
> 
> And your feeling is right about it - in most of cases (I would say more than 99.9%) only one useEffect will work perfectly
> 
> There is a very rare case here (like what I described in the code comment): The normal flow is like this: create videoTile -> createView -> (small time gap)-> renderElement exists -> tile dispose -> if (element) dispose();
> 
> But the rare case happens like this, when a tile gets created and immediately disposed: create videoTile -> createView -> tile dispose -> if (element) dispose();
> 
> And then because element doesn't get enough time to be created yet, it fails to dispose the view even it has started being created
> 
> The second useEffect is used for filling that small time gap for very rare condition - thing will not be too bad without it, but still nice to guard all rare cases

I see. Thanks for the explanation. This is a really finicky thing to reproduce. Thanks for looking into this. Let's see if we can improve this after GA.

mgamis-msft (approved):

PR #1199, opened 2021-12-03
Dmceachernmsft/CommonPropsInterface

dmceachernmsft:
# What
Updated the getting started with composites overview example to match the examples in composite pages.

# Why
Creates more continuity between our examples in the composites section

https://skype.visualstudio.com/SPOOL/_workitems/edit/2635879

# How Tested
Ran code in local instance and visual inspection.


packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
```diff
@@ -8,19 +8,32 @@ import {
...
+};
+
+const compositeProps: CompositeProps = {
+  displayName: '<Display Name>',
+  userId: '<Azure Communication Services Identifier>',
+  token: '<Azure Communication Services Access Token>',
+  endpoint: '<Azure Communication Services Resource Endpoint>',
+  threadId: '<Get thread id from chat service>',
+  groupId: '<Developer generated GUID>'
+};
```
PorterNan, line 27: I got the concern of making all options/params for composite more obvious into one place, which is great and makes a lot of sense
For creating an object for all these variables, I would say this one is more like an *App* instead of a *component*, so define a type and create a single object to contain everything, I would say either make it a real component like this:

```
type AppProps = {

....

};

function App(props: AppProps): JSX.Element {


}
ReactDOM.render(<App [pass all those props here] />, domElement);
```
or have a comment block to put all parameters at the top(I prefer this one, because it is neat), creating an object is the best practice for App and might confuse our code reader (especially when they come for quick start)

PorterNan: Would like to see some opinions from others, and also listen to any other concerns to wrap parameters into an object instance

PorterNan: And one question to be discussed as well, since this is the first example developers will see, we will never be too careful. Do we want to expose all parameter in the very first example? It makes our code super long and might scare some potential customers away - we could show a super short version first and have a long version hiding somewhere if they wanna do more exploration

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
```diff
@@ -8,19 +8,32 @@ import {
 } from '@azure/communication-react';
 import React, { useEffect, useMemo, useState } from 'react';
 
+export type CompositeProps = {
```
JamesBurnside, line 11: I think we need a ChatCompositeProps and CallCompositeProps, as groupId doesn't apply to Chat and likewise threadId and endpoint doesn't apply to CallComposite

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
```diff
@@ -8,19 +8,32 @@ import {
...
+  userId: string; // '<Azure Communication Services Identifier>'
+  token: string; // '<Azure Communication Services Access Token>'
+  endpoint: string; // '<Azure Communication Services Resource Endpoint>'
+  displayName: string; // '<Display Name>'
+  groupId: string; // '<Developer generated GUID>'
```
JamesBurnside, line 16: suggestion: rename to `locatorID` (to create commonality with AzureCommunicationCallAdapterArgs locator prop) or `callID` (to give better semantic meaning to what a groupID could mean)

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
JamesBurnside: Oh I see this is "AllCompositeProps" superset kind of interface, if its only used to pass in props for all composites then leave as is 👍

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeComplete.snippet.tsx
```diff
@@ -1,28 +1,69 @@
...
+  formFactor?: 'desktop' | 'mobile'; // '<Used to set the form factor for the calling composite>'
+  fluentTheme?: PartialTheme | Theme; // '<Theming for the composites using FluentUi themes>'
+  callInvitationURL?: string; // '<Invitation Url that will appear with a invite to call button>'
+  locale?: CompositeLocale; // '<Sets Locale of buttons in the composites>'
+  options?: CallCompositeOptions; // '<object that sets flags for visual elements of the composites>'
+  errorBar?: boolean; // '<Hides and shows the Errorbar in the chat composite>'
+  topic?: boolean; // '<Allows a topic to be set for the chat composite>'
+};
+
+const compositeProps: CompositeProps = {
```
JamesBurnside, line 35: @PorterNan/@anjulgarg/@prprabhu-ms is there a way to do the webpack include to not have this defined in multiple places but be expanded code in the snippets and not an import?

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeComplete.snippet.tsx
```diff
@@ -1,28 +1,69 @@
...
   createAzureCommunicationChatAdapter,
-  fromFlatCommunicationIdentifier
+  fromFlatCommunicationIdentifier,
+  CompositeLocale,
+  CallCompositeOptions
 } from '@azure/communication-react';
+import { PartialTheme, Theme } from '@fluentui/react';
 import React, { useEffect, useMemo, useState } from 'react';
 
+export type CompositeProps = {
```
JamesBurnside, line 16: If this isn't used outside this file, remove the `export`

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeComplete.snippet.tsx
```diff
@@ -1,28 +1,69 @@
...
+  groupId: string; // '<Developer generated GUID>'
+  threadId: string; // '<Get thread id from chat service>'
+
+  // Optional fields for customizing composites.
+  formFactor?: 'desktop' | 'mobile'; // '<Used to set the form factor for the calling composite>'
+  fluentTheme?: PartialTheme | Theme; // '<Theming for the composites using FluentUi themes>'
+  callInvitationURL?: string; // '<Invitation Url that will appear with a invite to call button>'
+  locale?: CompositeLocale; // '<Sets Locale of buttons in the composites>'
+  options?: CallCompositeOptions; // '<object that sets flags for visual elements of the composites>'
+  errorBar?: boolean; // '<Hides and shows the Errorbar in the chat composite>'
```
JamesBurnside, line 31: we use options for CallCompositeOptions, but here we have expanded the ChatCompositeOptions
suggestion:
```
callCompositeOptions?: CallCompositeOptions; // '<set flags for visual elements of the composites>'
chatCompositeOptions?: ChatCompositeOptions; // '<set flags for visual elements of the composites>'
```

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
dmceachernmsft: Yeah, the idea was to make a object they could add everything to if they wanted to play with the customization options within the quick start example.

packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
dmceachernmsft: I could definitely remove all the optional args from the object if we wanted to leave those for the specific examples in the respective composite pages. Having it as a actual prop to the App.tsx file though does bring it back around to the issue of making them edit the index.tsx file. I'm not too sure myself if thats something we want them to do if we are aiming to make this as simple as possible. Maybe a comment or something about the type saying that its a example of a props type if the composites were to be children instead of on the app element?

PR #1200, opened 2021-12-03
Fix screenshare label color when hovered but not checked

JamesBurnside:
# What
Fix screenshare label color when hovered but not checked

# Why
Was white and shouldn't be:
image

Regression introduce in: #1178 

# How Tested
| state | screenshot |
| -- | -- |
| unchecked | image |
| unchecked hovered |  image |
| checked | image |
| checked hovered | image |


prprabhu-ms (approved):

packages/react-components/src/components/ScreenShareButton.tsx
```diff
@@ -83,6 +83,5 @@ const screenshareButtonStyles = (theme: Theme): IButtonStyles => ({
     color: DefaultPalette.white,
     ':focus::after': { outlineColor: `${DefaultPalette.white} !important` } // added !important to avoid override by FluentUI button styles
   },
-  labelHovered: { color: DefaultPalette.white },
   labelChecked: { color: DefaultPalette.white }
```
alcail, line 86: wow just because of an extra comma 😮

alcail (approved):

PR #1204, opened 2021-12-03
rejoin Chat directly instead of going to config

alcail:
# What
In Chat sample, rejoining a Chat through "Rejoin" button on End page will direct the participant to the Chat page (no loss of config data) instead of going through the Config page again

# Why
bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2659156
where participant was able to change their display name before rejoining a Chat after clicking "Rejoin" button on End page, creating confusion

# How Tested
ran Chat sample locally 
image

# Process & policy checklist


alkwa-msft:
I am curious if this works  if you were kicked out of a chat thread as well? You generally won't have access to join again.

If we are getting an invalid thread issue then we should consider redirecting to the configuration screen instead?

alcail:
> e are getting an invalid thread issue then we should consider redirecting to the configuration screen instead

Our current Chat sample does not offer the possibility of removing a participant, but I updated it with this possibility anyway :)
I tested by forcing removedBy to be different from userId in console log and ended in the correct page
image

JamesBurnside (approved):

alkwa-msft (approved):

PR #1206, opened 2021-12-03
[PoC] Partial PR to separate out composite strings

prprabhu-ms:
# What

# Why

# How Tested

# Process & policy checklist


packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -51,6 +51,9 @@ export class ChatContext {
   }
 
   public getState(): ChatClientState {
+    for (const threadId in this._state.threads) {
+      console.log('xkcd', threadId, this._state.threads[threadId].chatMessages);
```
alcail, line 55: is this for telemetry?

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -1,45 +1,41 @@
...
+  "complianceBannerRecordingAndTranscriptionStarted": "Recording and transcription have started.",
+  "complianceBannerRecordingAndTranscriptionStopped": "Recording and transcription have stopped.",
+  "complianceBannerRecordingSaving": "Recording is being saved.",
+  "complianceBannerRecordingStarted": "Recording has started.",
+  "complianceBannerRecordingStopped": "Recording has stopped.",
+  "complianceBannerTrancriptionStarted": "Transcription has started.",
+  "complianceBannerTranscriptionConsent": "By joining, you are giving consent for this meeting to be transcribed.",
+  "complianceBannerTranscriptionSaving": "Transcription is being saved.",
+  "complianceBannerTranscriptionStopped": "Transcription has stopped.",
+  "configurationPageTitle": "Start a call",
```
alcail, line 18: what about a config title that is different for meeting for example?
we will have to specify Call or Meeting in the name?
Wouldn't it be better to keep Chat and Call objects and have a common/shared one for all that is common or shared between composites?
Just wondering

prprabhu-ms: Yeah, I'm not sure about the right path here either.

This PoC showed, more than anything else, that there isn't a quick answer to be done pre-GA.

PR #1207, opened 2021-12-03
force aria-live as ariaLve in Spinner not working

alcail:
# What
used the AriaAttributes aria-live instead of existing ariaLive prop from Spinner.

# Why
Issue is our spinner is embedded in a Stack and in that case the ariaLive prop of Spinner component is not working properly ending in Spinner label not being read by screen reader
I checked this FluentUI component ariaLive prop in its doc and it's indeed not working properly when embedded in a Stack.

# How Tested
ran Call sample locally with 2 participants, shared the screen from 1 of them and checked the screen reader announcement for the other one.

# Process & policy checklist


JamesBurnside (approved):

packages/react-components/src/components/VideoGallery/RemoteScreenShare.tsx
```diff
@@ -60,7 +60,7 @@ export const RemoteScreenShare = React.memo(
 const LoadingSpinner = (props: { loadingMessage: string }): JSX.Element => {
   return (
     <Stack verticalAlign="center" className={loadingStyle}>
-      <Spinner label={props.loadingMessage} size={SpinnerSize.xSmall} />
+      <Spinner label={props.loadingMessage} size={SpinnerSize.xSmall} aria-live={'assertive'} />
```
alkwa-msft, line 63: is there a constant we can use here? I can imagine someone could just miss spell this

alkwa-msft (approved):

packages/react-components/src/components/VideoGallery/RemoteScreenShare.tsx
alcail: we can create one, sure

JamesBurnside: I would actually argue against creating a constant here as aria-live has a specific interface: 
image

JamesBurnside: looking at that screenshot looks like we need to use ariaLive and not aria-live to get that type checking though

alcail: will do in another PR though

alcail: @JamesBurnside 'ariaLive' prop from Spinner FluientUI component is not working properly when inside a Stack, that's why I'm using the 'aria-live' from AriaAttributes to force the behaviour of Spinner

alcail: > 

the 'ariaLive' prop of Spinner component has indeed a specific interface, but not the 'aria-live' attribute

alcail: but, yes, you're right, even the AriaAttributes have specific interface, so no need for constants here then

PR #1209, opened 2021-12-06
Bump @azure/communication-calling to new stable version

prprabhu-ms:
# What

* Bump `@azure/communication-calling` to version `1.3.2`
* Make necessary changes to use of some breaking changes in API since last beta
* Make necessary API changes in our own API as a result.


JamesBurnside (approved):

anjulgarg (approved):

packages/calling-stateful-client/src/CallAgentDeclarative.test.ts
```diff
@@ -80,10 +77,9 @@ class MockCallAgent implements CallAgent {
     return call;
   }
   join(groupLocator: GroupLocator, options?: JoinCallOptions): Call;
-  join(groupChatCallLocator: GroupChatCallLocator, options?: JoinCallOptions): Call;
-  join(meetingLocator: MeetingLocator, options?: JoinCallOptions): Call;
+  join(TeamsMeetingLinkLocator: TeamsMeetingLinkLocator, options?: JoinCallOptions): Call;
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  join(meetingLocator: any, options?: any): Call {
+  join(TeamsMeetingLinkLocator: any, options?: any): Call {
```
PorterNan, line 82: Upper case? Lower case? (For first 'T'), this one with any is just implementation, we shouldn't expose it or we should keep it as meetingLocator

JamesBurnside:
Weird, seems to have impacted the static html apps:
image

packages/calling-stateful-client/src/CallAgentDeclarative.test.ts
```diff
@@ -132,18 +125,14 @@ class MockCallAgentWithMultipleCalls implements CallAgent {
...
```
PorterNan, line 130: Same here this shouldn't be renamed

packages/calling-stateful-client/src/CallAgentDeclarative.test.ts
```diff
@@ -132,18 +125,14 @@ class MockCallAgentWithMultipleCalls implements CallAgent {
     return call;
   }
   join(groupLocator: GroupLocator, options?: JoinCallOptions): Call;
-  join(groupChatCallLocator: GroupChatCallLocator, options?: JoinCallOptions): Call;
-  join(meetingLocator: MeetingLocator, options?: JoinCallOptions): Call;
+  join(TeamsMeetingLinkLocator: TeamsMeetingLinkLocator, options?: JoinCallOptions): Call;
```
PorterNan, line 128: teamsMeetingLocator

PorterNan (approved):
One minor comment about name, since let's make it before ship the npm package

PorterNan:
Just noticed those are bot api file, so we are good to have this checked in, then open a bug item for it

packages/calling-stateful-client/src/CallAgentDeclarative.test.ts
```diff
@@ -80,8 +77,7 @@ class MockCallAgent implements CallAgent {
     return call;
   }
   join(groupLocator: GroupLocator, options?: JoinCallOptions): Call;
-  join(groupChatCallLocator: GroupChatCallLocator, options?: JoinCallOptions): Call;
-  join(meetingLocator: MeetingLocator, options?: JoinCallOptions): Call;
+  join(meetingLocator: TeamsMeetingLinkLocator, options?: JoinCallOptions): Call;
```
PorterNan, line 80: I guess this one can be called teamsMeetingLocator? Should be fine cause this is just test

PorterNan: And the one typed with any, will be just meetingLocator

prprabhu-ms: Keeping changes minimal.

PR #1213, opened 2021-12-06
Bugfix for display name input showing "long name" validation error when it is empty

anjulgarg:
# What
A previous bugfix introduced a minor bug in our calling sample where display name input is showing long display name validation error even when it is empty

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/samples/Calling/src/app/views/DisplayNameField.tsx
+++ b/samples/Calling/src/app/views/DisplayNameField.tsx
@@ -25,7 +25,7 @@ const hasValidLength = (name: string): boolean => {
 
 export const DisplayNameField = (props: DisplayNameFieldProps): JSX.Element => {
   const { setName, setEmptyWarning, isEmpty, defaultName } = props;
-  const [isInvalidLength, setIsInvalidLength] = useState<ConstrainBoolean>(!hasValidLength(defaultName ?? ''));
+  const [isInvalidLength, setIsInvalidLength] = useState<boolean>(!hasValidLength(defaultName ?? ''));
 
   const onNameTextChange = (
     event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
@@ -37,6 +37,8 @@ export const DisplayNameField = (props: DisplayNameFieldProps): JSX.Element => {
 
     if (!hasValidLength(newValue)) {
       setIsInvalidLength(true);
+      // The button below DisplayNameField is being disabled if name is empty.
+      // To ensure that the Join Call button is disabled when the name is too long, we have to clear it from the state.
       setName('');
       return;
     } else {
--- a/samples/Calling/src/app/views/DisplayNameField.tsx
+++ b/samples/Calling/src/app/views/DisplayNameField.tsx
@@ -19,13 +19,13 @@ const TEXTFIELD_PLACEHOLDER = 'Enter a name';
 const TEXTFIELD_EMPTY_ERROR_MSG = 'Name cannot be empty';
 const TEXTFIELD_EXCEEDS_MAX_CHARS = `Name cannot exceed ${DISPLAY_NAME_MAX_CHARS} characters`;
 
+const hasValidLength = (name: string): boolean => {
+  return name.length <= DISPLAY_NAME_MAX_CHARS;
+};
+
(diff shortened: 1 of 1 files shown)


prprabhu-ms (approved):

samples/Calling/src/app/views/DisplayNameField.tsx
```diff
@@ -19,13 +19,13 @@ const TEXTFIELD_PLACEHOLDER = 'Enter a name';
...
+};
+
 export const DisplayNameField = (props: DisplayNameFieldProps): JSX.Element => {
   const { setName, setEmptyWarning, isEmpty, defaultName } = props;
-  const [isInvalidLength, setIsInvalidLength] = useState<ConstrainBoolean>(true);
-
-  const hasValidLength = (name: string): boolean => {
-    return name.length <= DISPLAY_NAME_MAX_CHARS;
-  };
+  const [isInvalidLength, setIsInvalidLength] = useState<ConstrainBoolean>(!hasValidLength(defaultName ?? ''));
```
JamesBurnside, line 28: What's a ConstrainBoolean?

samples/Calling/src/app/views/DisplayNameField.tsx
```diff
@@ -19,13 +19,13 @@ const TEXTFIELD_PLACEHOLDER = 'Enter a name';
 const TEXTFIELD_EMPTY_ERROR_MSG = 'Name cannot be empty';
 const TEXTFIELD_EXCEEDS_MAX_CHARS = `Name cannot exceed ${DISPLAY_NAME_MAX_CHARS} characters`;
 
+const hasValidLength = (name: string): boolean => {
+  return name.length <= DISPLAY_NAME_MAX_CHARS;
+};
+
 export const DisplayNameField = (props: DisplayNameFieldProps): JSX.Element => {
   const { setName, setEmptyWarning, isEmpty, defaultName } = props;
```
JamesBurnside, line 27: Your original code looks like it should've worked (although this is better as it doesn't assign to isInvalidLength to true by default)

It looks like we never pass in setEmptyWarning and isEmpty: https://github.com/Azure/communication-ui-library/blob/cd5058489810b02b9eaf66d742381cd4cf36dc66/samples/Calling/src/app/views/HomeScreen.tsx#L86

So do we ever show the empty warning?

samples/Calling/src/app/views/DisplayNameField.tsx
```diff
@@ -19,13 +19,13 @@ const TEXTFIELD_PLACEHOLDER = 'Enter a name';
...
-  const [isInvalidLength, setIsInvalidLength] = useState<ConstrainBoolean>(true);
-
-  const hasValidLength = (name: string): boolean => {
-    return name.length <= DISPLAY_NAME_MAX_CHARS;
-  };
+  const [isInvalidLength, setIsInvalidLength] = useState<ConstrainBoolean>(!hasValidLength(defaultName ?? ''));
 
   const onNameTextChange = (
     event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
```
JamesBurnside, line 31: Any idea why we setName to `''` when the name is too long? https://github.com/Azure/communication-ui-library/blob/1c4e8e461e7b1c32a1048f7f0bac117a69195c10/samples/Calling/src/app/views/DisplayNameField.tsx#L40

samples/Calling/src/app/views/DisplayNameField.tsx
anjulgarg: Nope, we just disable the buttons.
But it can be passed through the props so I let that code sit there.

samples/Calling/src/app/views/DisplayNameField.tsx
anjulgarg: Good question. I will also add comments for it.
The button below display name field is being disabled if name is empty.
To ensure that the button is disabled when the name is too long, we have to clear it from the state.

samples/Calling/src/app/views/DisplayNameField.tsx
anjulgarg: I don't remember adding it. Weird. Changing it to boolean

alcail (approved):

PR #1217, opened 2021-12-06
Vertical control bar fix

mgamis-msft:
# What

Maximum width of ControlBar and ControlBarButton to 8rem after discussing with Alex P. Truncating long ControlBarButton labels after discussing with Alex P . Vertical ControlBar width set to fit-content. 

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2657885

# How Tested

On storybook canvas in German in various layouts:
https://user-images.githubusercontent.com/79475487/145080644-3772ec37-f709-4713-ade7-081450ecb62f.mp4

# Process & policy checklist


alkwa-msft:
Vertical looks like it is missing some concept of right padding?
Does it seem more correct to have the text be centered when in vertical so we have equal padding on both sides?

image

mgamis-msft:
> Vertical looks like it is missing some concept of right padding? Does it seem more correct to have the text be centered when in vertical so we have equal padding on both sides?
> 
> image

Thanks for catching that. I only checked dockedLeft and floatingLeft.

alcail (approved):

alkwa-msft (approved):

PR #1220, opened 2021-12-07
Alcail/a11y chat status notification for system messages, typing indicator and message status

alcail:
# What
added aria label and role for system messages, typing indicator and message status for Chat.

# Why
a11y bug where all those information where passive and user was not notify of those status changes
https://skype.visualstudio.com/SPOOL/_workitems/edit/2584895

# How Tested
ran Chat sample locally with Narrator on

# Process & policy checklist

added optional strings in MessageStatusIndicatorStrings for aria-label

### DIFF
--- a/packages/react-components/src/components/TypingIndicator.tsx
+++ b/packages/react-components/src/components/TypingIndicator.tsx
@@ -164,6 +164,23 @@ const getUsersElement = (
 };
 
 /**
+ * Helper function to get a string of all typing users
+ * @param typingUsers typing users
+ * @param delimiter string to separate typing users
+ * @returns string of all typing users
+ */
+const getNamesString = (typingUsers: CommunicationParticipant[], delimiter: string): string => {
+  const userNames: string[] = [];
+
+  typingUsers.forEach((user) => {
+    if (user.displayName) {
+      userNames.push(user.displayName);
+    }
+  });
+  return userNames.join(delimiter);
+};
+
+/**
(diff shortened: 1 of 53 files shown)


packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -205,6 +222,41 @@ const getSpanElements = (
...
 
+/**
+ * Helper function to get the string making up the typing indicator string
+ * @param strings TypingIndicatorStrings containing strings to create span elements
+ * @param namesString string of all typing users
+ * @param numTypingUsers number of total typing users
+ * @param numUserNotMentioned number of typing users abbreviated
+ * @returns typing indicator string
+ */
+const getIndicatorString = (
```
JamesBurnside, line 233: This would benefit from some unit tests, both to validate it's logic (as it will be v difficult to spot if this ever regresses), and in case someone makes a change they can easily know given some inputs what expected results are

JamesBurnside (approved):

packages/react-components/src/components/TypingIndicator.tsx
alcail: will add some in next PR when I will re-introduce the specific aria strings

alcail: created a task for that: https://skype.visualstudio.com/SPOOL/_workitems/edit/2666070

mgamis-msft (approved):

PR #1221, opened 2021-12-07
Add manual documentation for adapters

JamesBurnside:
# What
Add manual documentation tables for adapter interfaces and functions

# Why
* Insufficient documentation exists on DCOM
* No in-built storybook tool that will generate a table for us

# More interfaces
We need more of these:
* StatefulCallClient
* StatefulChatClient
* usePropsFor
* Any other function we export

# How Tested
Can view changes here: https://60c7ae6891f0e90039d7cd54-lvsmpyvult.chromatic.com/?path=/docs/composite-adapters--page#interfaces


alcail (approved):

alkwa-msft:
(small nit)
In the functions I imagine we could use a 3rd column for any special notes but because not many functions have any "notes" we think we just want to avoid having a 3rd column?

If it makes sense to have a 3rd column lets do it but I understand if we don't want to. Thanks for taking this work item that does not have much clarity associated with it.

alkwa-msft (approved):

prprabhu-ms (approved):

JamesBurnside:
> (small nit) In the functions I imagine we could use a 3rd column for any special notes but because not many functions have any "notes" we think we just want to avoid having a 3rd column?
> 
> If it makes sense to have a 3rd column lets do it but I understand if we don't want to. Thanks for taking this work item that does not have much clarity associated with it.

I will check this in and we can easily add a 3rd column when we need it

PR #1223, opened 2021-12-07
Bugfix for local video tile overlapping control bar when dragged

anjulgarg:
# What
Bugfix for local video tile overlapping control bar when dragged

https://user-images.githubusercontent.com/9782747/145099110-d264b8a5-36b7-402a-8e83-d7000b59f911.mp4

https://user-images.githubusercontent.com/9782747/145099272-bb0ce767-2692-4438-ba5c-fb52c523803f.mp4

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -359,6 +369,18 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
             {horizontalGalleryTiles}
           </ResponsiveHorizontalGallery>
         )}
+
+        <LayerHost
+          id={layerHostId}
+          style={{
```
alcail, line 375: should those styles not be added to the styles file?

alcail (approved):

packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -26,13 +26,6 @@ export const videoGalleryContainerStyle: IStackStyles = {
   root: { position: 'relative', height: '100%', width: '100%', padding: '0.5rem' }
 };
 
-/**
- * @private
- */
-export const videoGalleryContainerTokens = {
-  childrenGap: '0.5rem'
-};
```
mgamis-msft, line 34: I think we still need this gap when the HorizontalGallery is present otherwise it looks like this:
image

prprabhu-ms:
Let's keep this PR out of release (OK to land on `main`).

Could we size and position the `LayerHost` such that there's margin equal to the size of the PIP modal? In that way, the PIP will be draggable such that it never leaves the `VideoGallery` boundary at all.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -30,14 +39,14 @@ import {
...
   videoGalleryContainerStyle,
-  videoGalleryContainerTokens,
   videoGalleryOuterDivStyle
 } from './styles/VideoGallery.styles';
 import { isNarrowWidth, useContainerWidth } from './utils/responsive';
 import { LocalScreenShare } from './VideoGallery/LocalScreenShare';
 import { RemoteScreenShare } from './VideoGallery/RemoteScreenShare';
 import { VideoTile } from './VideoTile';
 import { v4 as uuidv4 } from 'uuid';
+import { useId } from '@uifabric/react-hooks';
```
JamesBurnside, line 49: Is there a reason we use uifabric over fluentui package? https://www.npmjs.com/package/@fluentui/react-hooks#useid

anjulgarg: Thanks for pointing this out James. Good catch

packages/react-components/src/components/styles/VideoGallery.styles.ts
anjulgarg: Solved it by adding a top padding to horizontal gallery.
The reason I needed to remove the children gap token was because it was adding an undesired padding to the layerhost causing a shift in the local video placement. This  was messing up the whole layout.

anjulgarg: > Let's keep this PR out of release (OK to land on `main`).
> 
> Could we size and position the `LayerHost` such that there's margin equal to the size of the PIP modal? In that way, the PIP will be draggable such that it never leaves the `VideoGallery` boundary at all.

That would cause the pip modal to shift towards the centre of video gallery and dragging it out will cause it to clip.

mgamis-msft (approved):

anjulgarg:
> Let's keep this PR out of release (OK to land on `main`).
> 
> Could we size and position the `LayerHost` such that there's margin equal to the size of the PIP modal? In that way, the PIP will be draggable such that it never leaves the `VideoGallery` boundary at all.

That would cause the pip modal to shift towards the centre of video gallery and dragging it out will cause it to clip.

PR #1227, opened 2021-12-07
ScreenReaderMenuItemFix

dmceachernmsft:
# What
Added aria roles to the context menu items in the options button.

# Why
Before it wasn't reading out the quantity of items in the list. now it does

https://skype.visualstudio.com/SPOOL/_workitems/edit/2561987
# How Tested
used Narrator to make sure the right output was being read on that menu.

### DIFF
--- a/packages/react-components/src/components/DevicesButton.tsx
+++ b/packages/react-components/src/components/DevicesButton.tsx
@@ -214,6 +214,7 @@ const generateDefaultMenuProps = (
           itemProps: {
             styles: menuItemStyles
           },
+          role: 'menuitem',
           canCheck: true,
           isChecked: camera.id === selectedCamera?.id,
           onClick: () => {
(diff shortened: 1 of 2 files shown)


alcail (approved):

packages/react-components/src/components/DevicesButton.tsx
```diff
@@ -268,6 +270,7 @@ const generateDefaultMenuProps = (
           itemProps: {
             styles: menuItemStyles
           },
+          role: 'menuitem',
```
prprabhu-ms, line 273: `menuitem` is the default role, per https://developer.microsoft.com/en-us/fluentui#/controls/web/contextualmenu#IContextualMenuProps

Why do we need to set it explicitly?

dmceachernmsft: Yes for the narrator to read out what # of the list the item is it needed that tag there. at least that's what I found when testing. do you have a suggestion on another way?

dmceachernmsft: Without the explicit declaration of it being a menu item the windows narrator does not read out which number in the menu item it is for some reason. so like "speaker x item 1 of 2" and so forth.

JamesBurnside (approved):

PR #1231, opened 2021-12-08
Conditional build

PorterNan:
# What

Merge conditional build into @prprabhu-ms's branch

- Add overall conditional build for communication-react package
- Add build:public and test:public command for each sub-project
- Add global bulk command - rush build:public and rush test:public
- Change comment block style to one line annotation comment

**Also put 2 updated api.md files as the example of how it looks after conditional build, free feel to remove/revert them**

TODO:
- Add conditional build into UI Tests (will be another PR)
- Add conditional build into pipeline (another PR)

# Why

# How Tested

# Process & policy checklist


common/config/rush/command-line.json
```diff
@@ -31,6 +31,25 @@
       "shellCommand": "node common/scripts/ensure-consistent-versions.js",
       "safeForSimultaneousRushProcesses": true
     },
+    {
+      "commandKind": "bulk",
+      "name": "build:public",
+      "summary": "Preprocess and build the non-beta version.",
```
prprabhu-ms, line 37: nitty nit: Reword to be positive (what's non-beta, does it mean nightly?): "Build the @public variant of all packages"

Let's also consistently choose a word to refer to this: `version` vs `variant` vs `flavour`.
I'd argue against `version` (version refers to SemVer). I prefer `variant` over `flavour` because I don't want to get into endless fights about British vs American spelling :P

prprabhu-ms: Along those lines, could we also rename this command to `build-variant:public` and `test-variant:public`? By itself, `build:public` is a bit confusing (it makes me think there's a private thing I could build instead).

common/scripts/babel-conditional-preprocess.js
```diff
@@ -0,0 +1,80 @@
+Object.defineProperty(exports, '__esModule', { value: true });
```
prprabhu-ms, line 1: Please add a top-level comment explaining what this preprocessing step does, and where it is used.

prprabhu-ms:
Partial review. Let's walk through the Babel plugin together.

We should add a doc explaining how the plugin works, and how to do beta vs. public development

common/config/rush/command-line.json
PorterNan: For build:public one, maybe we should call it build:prod? I do think build:public sounds like there is a private version, and our beta is still "public", so maybe prod or ga would be something more accurate?

PorterNan: I come up with another term build:stable that sounds like a good term for GA, so we might have alpha, beta then stable for GA

common/scripts/babel-conditional-preprocess.js
```diff
@@ -0,0 +1,80 @@
...
+
+const helper = require('@babel/helper-plugin-utils');
+const t = require('@babel/types');
+exports.default = helper.declare((_api, opts) => {
+  const {
+    annotations
+  } = opts;
+
+  function Handle(path) {
+    let { node } = path;
```
prprabhu-ms, line 11: Please add link to AST types: https://babeljs.io/docs/en/babel-types#typescript

common/scripts/babel-conditional-preprocess.js
```diff
@@ -0,0 +1,80 @@
...
+    let removed = false;
+
+    for (const annotation of annotations) {
+      const {
+        match
+      } = annotation;
+
+      if (node.leadingComments && node.leadingComments.length > 0) {
```
prprabhu-ms, line 19: If possible, link to documentation for Node type that shows `leadingComment` always exists.

common/scripts/babel-conditional-preprocess.js
```diff
@@ -0,0 +1,80 @@
...
+        for (const comment of node.leadingComments) {
+          if (comment.value.includes(match)) {
```
prprabhu-ms, line 21: Single `if`?

common/scripts/babel-conditional-preprocess.js
```diff
@@ -0,0 +1,80 @@
...
+            removed = true;
+          }
+          if(removed) {
+            comment.ignore = true;
+          }
+        }
+      }
+
+      if (removed) {
```
prprabhu-ms, line 30: Either:

* Add comments explaining what is happening here ++ only remove if `path.parentPath` has only one child.

Or:

* Just document that we don't support JSX expressions. We'd have to create private Component instead.

common/scripts/babel-conditional-preprocess.js
```diff
@@ -0,0 +1,80 @@
+Object.defineProperty(exports, '__esModule', { value: true });
+
+const helper = require('@babel/helper-plugin-utils');
+const t = require('@babel/types');
```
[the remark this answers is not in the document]
PorterNan: Agree with babelHelper, but maybe will keep this one. This 't' is like a recommendation name from babel plugin handbook - https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md

Which is also widely used by other plugin developers

common/scripts/babel-conditional-preprocess.js
```diff
@@ -0,0 +1,80 @@
...
+
+      ExportAllDeclaration(path) {
+        Handle(path);
+      },
+
+      TSPropertySignature(path) {
+        Handle(path);
+      },
+
+      Expression(path) {
```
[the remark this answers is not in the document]
PorterNan: Sync offline, we still need it to make some special cases work, but we need to avoid using it a lot

common/scripts/babel-conditional-preprocess.js
```diff
@@ -0,0 +1,80 @@
...
+
+      FunctionDeclaration(path) {
+        Handle(path);
+      },
+
+      Statement(path) {
+        Handle(path);
+      },
+
+      VariableDeclaration(path) {
```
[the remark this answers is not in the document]
PorterNan: Sync offline, will keep this to pass vscode 'used' compiler check in some cases

common/scripts/babel-conditional-preprocess.js
```diff
@@ -0,0 +1,80 @@
+Object.defineProperty(exports, '__esModule', { value: true });
+
+const helper = require('@babel/helper-plugin-utils');
+const t = require('@babel/types');
+exports.default = helper.declare((_api, opts) => {
+  const {
+    annotations
```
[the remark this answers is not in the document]
PorterNan: I was about to apply this comment, but after thinking for a while I'd say I would prefer to keep it, the main reason is people don't have to read conditional-complie.js to see/modify this magic keyword, the keyword should be able to be found within the project (this creates a direct coupling to share a string between project and plugin)- for avoiding setting the wrong annotation, I will use a common config file to put it just for once

common/scripts/babel-conditional-preprocess.js
PorterNan: I found a logic leak if we use if - we wanna remove all the rest following comments after the annotation comment, which means removed = true but value doesn't match, reverting this back

packages/acs-ui-common/package.json
```diff
@@ -11,6 +11,12 @@
     "build:cjs": "rollup -c --silent --failAfterWarnings",
     "build:watch": "rushx build",
     "clean": "rimraf dist",
+    "copy-original": "cpx ./src/** ./preprocessed",
```
prprabhu-ms, line 14: Why would we want to run `rushx copy-original`, ever?

I'd drop this command

packages/acs-ui-common/package.json
```diff
@@ -11,6 +11,12 @@
...
+    "preprocess": "rushx copy-original && babel ./src --out-dir ../preprocessed --extensions \".ts,.tsx\" --keep-file-extension --config-file ./.babelrc.js --relative && rimraf ../preprocessed",
```
prprabhu-ms, line 15: Do we ever want to run `rushx preprocess` instead of `rushx preprocess:stable`? I'd drop this one as well.

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -261,7 +261,7 @@ const DefaultSystemMessage: MessageRenderer = (props: MessageProps) => {
         case 'content':
           return (
             <SystemMessageComponent
-              iconName={(message.iconName ?? '') as SystemMessageIconTypes}
+              iconName={(message.iconName ? message.iconName : '') as SystemMessageIconTypes}
```
prprabhu-ms, line 264: Please remove this unrelated change.

samples/Calling/package.json
```diff
@@ -5,12 +5,14 @@
   "description": "This is a sample application to show how the UI Library can be used to build a calling experience. The client-side application is a React based user interface which uses React Hooks for handling complex state while leveraging Microsoft Fluent UI for how things look.",
   "scripts": {
     "build": "rushx clean && concurrently --kill-others-on-fail \"tsc\" \"webpack --mode=production --env production\"",
+    "build:stable": "echo skip",
```
prprabhu-ms, line 8: In a follow up PR: We need a way to  build these for both falvors as well.

packages/storybook/package.json
```diff
@@ -4,10 +4,12 @@
   "description": "Azure Communication Services UI Library Storybook",
   "scripts": {
     "build": "build-storybook -s stories,public,.storybook  --quiet --loglevel warn",
+    "build:stable": "echo skip",
```
prprabhu-ms, line 7: Follow up PR: We need a way to build / start these for both flavors as well.

prprabhu-ms (approved):
Let's merge this and follow up about:
* Storybook: Build, test, start for both flavors
* Samples: Build, test, start for both flavors
* e2e tests: Build, run for both flavors

PR #1235, opened 2021-12-09
Dmceachernmsft/keyboardFocusLostInChat

dmceachernmsft:
# What
Added hook and tab indexing to inputboxcomponent to force focus to text box on chat session join.

# Why
Fixed issue where when using the narrator that the first tab of the page would be the leave button instead of the chat message input box.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2584867
# How Tested
Tested using the chat sample and using the narrator to confirm the order of the navigation and read out.


dmceachernmsft:
fixes desired behavior for input box control and narration but scan mode is not finding the leave button in the narration. I wonder if the fact that the input box shows up near the bottom of the accessibility tree has the narrator unable to see the leave button which is up near the top of the tree?
image
image

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -107,6 +107,7 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
     <Stack className={mergedRootStyle}>
       <div className={mergedTextContainerStyle}>
         <TextField
+          tabIndex={1}
```
JamesBurnside, line 110: Hmm, if these are global, how does it impact a composite embedded in an App? Will Contoso be able to set the necessary tabIndex for their site or will we have "stolen" tabIndex: 1 from them ?

dmceachernmsft: if they set their tabindex to 1 on their forms if there is one on the page with the composite it will just follow the same behavior as default. I will look into if I can remove them and get it to behave correctly without them though because they should not have to worry about that.

alcail: should be tabIndex={0} instead, right?

packages/react-components/src/components/SendBox.tsx
```diff
@@ -221,6 +223,7 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
         }}
         id={'sendIconWrapper'}
         className={mergedSendButtonStyle}
+        ariaLabel="send"
```
alcail, line 226: this should be localized string from Locale

packages/react-components/src/components/SendBox.tsx
```diff
@@ -221,6 +223,7 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
         }}
         id={'sendIconWrapper'}
         className={mergedSendButtonStyle}
+        // ariaLabel={localeStrings.}
```
alcail, line 226: need to add this ;)

samples/Chat/src/app/ChatHeader.tsx
```diff
@@ -25,25 +25,27 @@ export const ChatHeader = (props: ChatHeaderProps): JSX.Element => {
...
-            styles={buttonWithIconStyles}
-            text={leaveString}
-            onClick={() => props.onEndChat()}
-            onRenderIcon={() => <Icon iconName={leaveIcon.iconName} className={leaveIconStyle} />}
-          />
-        </div>
-      </Stack>
+    <Stack horizontal={true} verticalAlign={'center'} horizontalAlign="end" className={chatHeaderContainerStyle}>
+      <DefaultButton
+        className={mergeStyles(leaveButtonStyle, largeLeaveButtonContainerStyle, {
```
alcail, line 30: add the container style first
--> className={mergeStyles(largeLeaveButtonContainerStyle, leaveButtonStyle,  { ...

samples/Chat/src/app/styles/ChatScreen.styles.ts
```diff
@@ -13,7 +13,11 @@ export const chatScreenContainerStyle = mergeStyles({
 export const chatHeaderContainerStyle = mergeStyles({
```
alcail, line 13: this one should not be used anymore so just remove it :)

alcail (approved):

packages/react-components/src/components/SendBox.tsx
```diff
@@ -45,6 +45,10 @@ export interface SendBoxStrings {
    * The warning message when send box text length is more than max limit
    */
   textTooLong: string;
+  /**
+   * Aria label for send message button
+   */
+  sendButtonText: string;
```
JamesBurnside, line 51: `sendButtonText` implies it would be a user facing text property (i.e. the text of the send button). Can this be `sendButtonAriaLabel`

change/@internal-react-components-27920e95-e44e-4cc2-91a2-50484465e1ec.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Added strings for sendBox component",
```
JamesBurnside, line 3: nit: would be useful for the changelog to mention here now the send box steals focus when mounted

packages/react-components/src/components/SendBox.tsx
dmceachernmsft: yup on it!

change/@internal-react-components-27920e95-e44e-4cc2-91a2-50484465e1ec.json
dmceachernmsft: yup can add that!

packages/react-components/src/components/SendBox.tsx
```diff
@@ -128,6 +132,8 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
 
   const sendTextFieldRef = React.useRef<ITextField>(null);
 
+  useEffect(() => sendTextFieldRef.current?.focus(), []);
```
JamesBurnside, line 135: A little concerned we blindly steal focus when this component mounts (components should be dumb, very generic components). Perhaps we should add a `focusOnMount` property or `focus` method`, then when the chat-composite can use those - don't block PR on this comment but would like to hear what you and others think?

alcail: good point. 
Focus should be on this SendBox when integrated in Chat for our case, but indeed, Contoso might want to have focus elsewhere when using this component on its own

dmceachernmsft: So do we think maybe changing it to a focus method then? or do we think the prop is the better way? im leaning towards the function more myself

dmceachernmsft: So I tried just removing the useEffect focus method from this and the narrator completely freaks out. so something I am wondering is are we wanting to take this out or give the option to override the focus? I'm just wondering because wont it go against the accessibility standards we need to meet with what we are developing if we make something that the narrator just freaks on? thoughts here?

alcail: what about adding a props 'shouldFocusOnMount: boolean;' and then use it like this

useEffect(() => {
  if (shouldFocusOnMount) {
    sendTextFieldRef.current?.focus();
  }
}, []);


and set it to true in the composite when calling Sendbox

alcail: yep, good point.
So in composite it will be a prop like 'shouldFocusOnSendBoxonMount' or something like that

JamesBurnside: I think at the composite level it could be the same api: `focusOnMount: boolean` or `focus: () => void` (and the chat composite logic can dictate that the sendbox is the thing that gets focus when called/mounted)

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -54,6 +54,10 @@ export type ChatCompositeOptions = {
    * @defaultValue true
    */
   topic?: boolean;
+  /**
+   * flag to force focus on the chat send box text field
+   */
+  hasFocusOnMount?: boolean;
```
alcail, line 60: more explicit name on composite level, something like hasFocusOnSendBoxOnMount

packages/react-components/src/components/SendBox.tsx
```diff
@@ -126,8 +134,13 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
   const [textValue, setTextValue] = useState('');
   const [textValueOverflow, setTextValueOverflow] = useState(false);
 
+  const hasFocusOnMount = props.hasFocusOnMount ?? true;
```
alcail, line 137: default behaviour is focus on SendBox.
Is that what we want? It sounds to be the proper way to go when used in Chat, but just asking

alcail (approved):

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -52,6 +52,7 @@ type InputBoxComponentProps = {
   errorMessage?: string | React.ReactElement;
   disabled?: boolean;
   styles?: InputBoxStylesProps;
+  hasFocus?: 'sendBoxTextField' | false;
```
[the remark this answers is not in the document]
dmceachernmsft: okay sounds good will update that!

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -103,10 +105,14 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
     [onEnterKeyDown, onKeyDown, supportNewline]
   );
 
+  let focusText;
+  hasFocus === 'sendBoxTextField' ? (focusText = true) : (focusText = false);
```
JamesBurnside, line 109: nit: this can be condensed:
```
const textFieldAutoFocus = hasFocus === 'sendBoxTextField';
```
and given it is a trivial boolean check, I'd move it inline instead of assinging to a new variable for brevity:
```
<TextField autoFocus={props.autoFocus === 'sendBoxTextField'}
```

dmceachernmsft: okay sounds good!

samples/tests/StaticHtmlComposites/EmbededComposite.test.ts
```diff
@@ -13,8 +13,19 @@ test.describe('JS Bundle Test', () => {
     // Go to the starting url before each test.
     const user = await createChatThreadAndUsers(PARTICIPANTS);
 
+    const styleContent = `
```
JamesBurnside, line 16: add comment here what this is doing and why this is being applied

dmceachernmsft: ah okay will do!

packages/communication-react/review/communication-react.api.md
```diff
@@ -1719,6 +1720,7 @@ export const SendBox: (props: SendBoxProps) => JSX.Element;
 // @public
 export interface SendBoxProps {
     disabled?: boolean;
+    hasFocusOnMount?: 'sendBoxTextField' | false;
```
prprabhu-ms, line 1723: Should this just be `boolean`. In the case of this component, we wouldn't want to provide multiple focus points, so keep the API simple, and just name it to `autoFocus`, as we're simply forwarding the prop to the HTML element for the TextBox, right?

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -103,10 +105,14 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
     [onEnterKeyDown, onKeyDown, supportNewline]
   );
 
+  let focusText;
```
prprabhu-ms, line 108: This can simply be `const focusText = (hasFocus === 'sendBoxTextField')`.

Even better, just inline to line 115? 

  `autoFocus={hasFocus === 'sendBoxTextField'}`

packages/communication-react/review/communication-react.api.md
dmceachernmsft: so move the boolean comparison in the inputboxComponent out to the chat screen inside the composite?

packages/react-components/src/components/InputBoxComponent.tsx
dmceachernmsft: yup this is better, will make it so!

packages/communication-react/review/communication-react.api.md
JamesBurnside: I'm happy for either boolean or `'sendBoxTextField' | false;`.
There is a send button inside the component so I'm happy to be non-opinionated and say more-than-one-component == use-the-enum.
But I'm also happy for us to be opinionated and say its a sendbox - we will want the user to be able to type something in the box before hitting send

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -54,6 +55,10 @@ export type ChatCompositeOptions = {
    * @defaultValue true
    */
   topic?: boolean;
+  /**
+   * flag to force focus on the chat send box text field
```
JamesBurnside, line 59: nit: update this comment slightly. We only set focus on mount - and could be any of the possible options not just the send box (currently its either sendbox or none but we can be agnostic of that in the comment).

```suggestion
   * Set focus on the composite when the composite first mounts.
```

JamesBurnside (approved):

PR #1248, opened 2021-12-14
Alcail/fixed typing indicator and added more tests

alcail:
# What
Added more tests to TypingIndicator.
While added those tests, I found some issue in its logic, so fixed it :)

# Why
TypingIndicator were tested for just one case (1 user typing)

# How Tested
ran tests locally

# Process & policy checklist


packages/react-components/src/components/TypingIndicator.test.tsx
```diff
@@ -9,14 +9,74 @@ import { mountWithLocalization, createTestLocale } from './utils/testUtils';
 
 Enzyme.configure({ adapter: new Adapter() });
 
+const randomText = Math.random().toString();
```
JamesBurnside, line 12: Is there a reason to use randomText and not a fixed test string? The issue with randomText is if we ever did a `expect().toMatchSnapshot` it would break. (and in general it's best to avoid math.random in tests as it introduces a tiny amount of uncertainty which could lead to flakey tests further down the line)

JamesBurnside: (Oh I see this line already existsed and was just moved - feel free to not resolve this comment)

JamesBurnside (approved):

packages/react-components/src/components/TypingIndicator.test.tsx
alcail: not particularly. I just extended the existing test.
I will change that :)

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -297,17 +297,12 @@ const IndicatorComponent = (
...
-    typingUsersMentioned.length,
-    numUserNotMentioned
-  );
+  const spanElements: JSX.Element[] = getSpanElements(strings, usersElement, typingUsers.length, numUserNotMentioned);
 
   const labelString = getIndicatorString(
     strings,
     getNamesString(typingUsersMentioned, strings.delimiter),
-    typingUsersMentioned.length,
+    typingUsers.length,
```
PorterNan, line 305: It makes a lot of sense to use typingUsers.length to set typing user count.
Just a little bit curious about what scenarios will current code cause problems? Inside of getNameString logic, it only depends on whether typingUserCounts > 1 or >0, then it looks the same to me for typingUsers.length and typingUserMentioned.length (But still we should fix this logic)

PorterNan (approved):

packages/react-components/src/components/TypingIndicator.tsx
alcail: @PorterNan 
because we passed typingUsersMentioned as the 3rd param of getSpanElements (which is numTypingUsers in line 194), we ended up always having numTypingUsers at 1 in case of numTypingUsersAbbreviated > 1 , meaning always having 'UserName is typing ...' when we should have 'UserName and 1 other are typing ...' or 'UserName and x others are typing ...'

Same issue with getNamesString as it was a copy/past of params :(

PR #1251, opened 2021-12-14
Alcail/config camera button focus

alcail:
# What
allow control bar buttons to be focusable even when disabled for LocalPreview and CallControls in Call composite

# Why
focus was lost on camera button after turning it on due to button being disabled till we get the video stream
bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2562167

# How Tested
ran call sample and navigate through keyboard

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
+++ b/packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
@@ -145,6 +145,7 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
       showLabel={!compactMode}
       styles={controlButtonBaseStyle}
       {...microphoneButtonStrings}
+      allowDisabledFocus={true}
     />
   );
 
@@ -154,6 +155,7 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
       {...cameraButtonProps}
       showLabel={!compactMode}
       styles={controlButtonBaseStyle}
+      allowDisabledFocus={true}
     />
   );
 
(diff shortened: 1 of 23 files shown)


prprabhu-ms (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -145,6 +145,7 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
       showLabel={!compactMode}
       styles={controlButtonBaseStyle}
       {...microphoneButtonStrings}
+      allowDisabledFocus={true}
```
JamesBurnside, line 148: nit: thoughts on having this property as default on the ControlBarButton.tsx:
```
return (
    <ControlButtonTooltip content={tooltipContent} id={tooltipId}>
      <DefaultButton
```
?
Pros: accessibility by default for customers
Cons: slight divergence from Fluent's default button

alcail: I was wondering about it too but decided to not go there as it should be a decision on Contoso when using components.
I agree that having it as true by default is a better behaviour, but that's something I can discuss with Alex P to have a better insights

JamesBurnside: Contoso can still decide, we'd only be setting allowDisabledFocus to be true _by default_ (they can set it to false if they don't want it). I'd still advocate for having it in the control bar button as we claim to be an a11y-centric library in our storybook but then wouldn't be applying a11y specific properties up front (Contoso has to set a quite hidden property to get the a11y feature)

alcail: true :)

alcail: I will update the storybook doc to reflect this though

JamesBurnside (approved):

PR #1252, opened 2021-12-15
Add conditional compiling to all webpack bundles

PorterNan:
Add conditional compiling to all webpack bundles, including:
1. All sample apps
2. All UI test apps

TODO:
We will need to add conditional compile to our PR/CI pipeline
# What

# Why

# How Tested

# Process & policy checklist


alcail (approved):

common/config/webpack/sampleapp.webpack.config.js
```diff
@@ -5,94 +5,110 @@ const path = require('path');
...
-  module: {
-    rules: [
-      {
-        test: /\.tsx?$/,
-        loader: 'ts-loader',
+  }
+
+  process.env['FLAVOR'] === 'stable' &&
+    config.module.rules.push({
+      test: /\.tsx?$/,
```
JamesBurnside, line 101: does this overwrite the existing rule applied to .tsx? files and say use babel-loader instead of ts-loader, or add a second rule and both are run?

If its the former, then for readability this check would be better inline:
```
const tsloaderRule = {
  test: /\.tsx?$/,
  loader: 'ts-loader',
  options: {
    transpileOnly: true
};

const babelLoaderRule = {
  ...babel load rule config...
};

   ...
    module: {
      rules: [
        process.env['FLAVOR'] === 'stable' ? babelLoaderRule : tsloaderRule,
```

PorterNan: This is adding one more rule and dealing with only conditional compile here

JamesBurnside: So will both ts-loader and babel-loader be run on tsx files?

PorterNan: Yep, babel-loader will strip conditional @beta code, then it follows our normal build process to get webpack bundles

PorterNan: Also for execution order, I guess we might have questions on it:
https://stackoverflow.com/questions/32234329/what-is-the-loader-order-for-webpack#:~:text=%7B%0A%20%20%20%20test%3A%20/%5C.css%24/%2C%0A%20%20%20%20loaders%3A%20%5B%27style%27%5D%2C%0A%7D%2C

It's following a reversed order, so the babel loader will be executed first

prprabhu-ms (approved):

PR #1253, opened 2021-12-15
Alcail/camera tooltip when waiting for stream

alcail:
# What
added Tooltip string to camera button for case of video loading
image

# Why
bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2659175

# How Tested
ran calling sample locally

# Process & policy checklist

added string to CameraButtonStrings

### DIFF
--- a/change/@internal-react-components-9a822f88-c39d-4d91-9b15-36af563aa5b7.json
+++ b/change/@internal-react-components-9a822f88-c39d-4d91-9b15-36af563aa5b7.json
@@ -1,0 +1,7 @@
+{
+  "comment": "add tooltipVideoLoadingContent to Camera strings",
+  "type": "none",
+  "packageName": "@internal/react-components",
+  "email": "alcail@microsoft.com",
+  "dependentChangeType": "none"
+}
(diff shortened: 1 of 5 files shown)


change/@internal-react-components-9a822f88-c39d-4d91-9b15-36af563aa5b7.json
```diff
@@ -0,0 +1,7 @@
+{
+  "comment": "add tooltipVideoLoadingContent to Camera strings",
+  "type": "none",
```
alcail, line 3: @JamesBurnside this is defined by default but what should we set it to now?
Same question for "dependentChangeType"

JamesBurnside: I haven't had time to update the options to minor/patch/none yet so easiest to leave as prerelease  - but I'll be going through the existing ones anyway to update them to patch or minor so wouldn't worry too much about it

JamesBurnside (approved):

dmceachernmsft (approved):

PR #1256, opened 2021-12-17
MeetingCallControlsOptions

dmceachernmsft:
# What
Added MeetingCallControlOptions interface
# Why
adds the options type to the MeetingCallControlBar to allow for the developer to hide buttons in the meeting composites calling control bar.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2606490
# How Tested
Ran locally with visual inspection with different options.


alcail (approved):

alcail:
Just need to fix GitHub issues ;)

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -43,6 +43,49 @@ export interface MeetingCompositeProps extends BaseCompositeProps<CallCompositeI
...
+/**
+ * Meeting Call controls to show or hide buttons on the calling control bar.
+ *
+ * @beta
+ */
+export interface MeetingCallControlOptions
+  extends Pick<
+    CallControlOptions,
+    /**
+     * Flag that toggles the camera button in the MeetingCallControlBar component.
```
JamesBurnside, line 61: You shouldn't need to re-document the `Pick`ed props, the interface will inherit the documentation of the original props

dmceachernmsft: okay will take those out!

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -43,6 +43,49 @@ export interface MeetingCompositeProps extends BaseCompositeProps<CallCompositeI
...
+     * Flag that toggles the Devices drop down and button in the MeetingCallControlBar component.
+     */
+    | 'devicesButton'
+    /**
+     * Selector that determines the display type to the CallControls component inside MeetingCallControlBar component, set to 'default' or 'compact'
+     */
+    | 'displayType'
+  > {
+  /**
+   * Flag for displaying the chat button inside the MeetingCallControlBar component.
```
JamesBurnside, line 82: nit: "toggling" here is ambiguous - does it mean toggle as in check/uncheck or toggle whether it is shown or not

```suggestion
   * Show or Hide Chat Button in the meeting control bar.
   * @defaultValue true
```

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -43,6 +43,49 @@ export interface MeetingCompositeProps extends BaseCompositeProps<CallCompositeI
...
+   */
+  chatButton?: boolean;
+  /**
+   * Flag for toggling the people button inside the MeetingCallControlBar component.
```
JamesBurnside, line 67: nit: "toggling" here is ambiguous - does it mean toggle as in check/uncheck or toggle whether it is shown or not

```suggestion
   * Show or Hide Chat Button in the meeting control bar.
   * @defaultValue true
```

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
JamesBurnside, line 63: nit: be sure to be clear what the default value is with `@defaultValue`

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -23,17 +23,29 @@ export interface MeetingCallControlBarProps {
...
+        : { disabled: props.disableButtonsForLobbyPage },
+    displayType: props.meetingCallControlOptions?.displayType,
+    microphoneButton: !props.meetingCallControlOptions?.microphoneButton ? false : true,
+    cameraButton: !props.meetingCallControlOptions?.cameraButton ? false : true,
+    devicesButton: !props.meetingCallControlOptions?.devicesButton ? false : true
   };
+  // Set flags for the chat button and people button specific to the meeting control bar.
+  const chatButton = props.meetingCallControlOptions?.chatButton === false ? false : true;
+  const peopleButton = props.meetingCallControlOptions?.peopleButton === false ? false : true;
+  console.log(callControlsOptions);
```
JamesBurnside, line 48: remove console log

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -43,6 +43,30 @@ export interface MeetingCompositeProps extends BaseCompositeProps<CallCompositeI
    * URL that can be used to copy a meeting invite to the Users clipboard.
    */
   meetingInvitationURL?: string;
+  /**
+   * Call control options to change what buttons show on the meeting composite control bar.
+   */
+  meetingCallControlOptions?: MeetingCallControlOptions;
```
JamesBurnside, line 49: Can we also allow for completely hiding the control bar to match the call composite callcontrols interface:
image

(though this is a good example of a change that would be better as a separate PR as although its similar, is a distinctly different feature)

prprabhu-ms:
Any chance we can wait for https://github.com/Azure/communication-ui-library/pull/1256 to land and then start contributing back in `main`?

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
dmceachernmsft: Will make separate PR for this issue.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -23,17 +23,28 @@ export interface MeetingCallControlBarProps {
...
 /**
  * @private
  */
 export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.Element => {
-  // Disable the default participants button as meetings composite has its own participants button.
+  // Set the desired control buttons from the meetings composite. particiapantsButton is always false since there is the peopleButton.
   let callControlsOptions: CallControlOptions | false = {
     participantsButton: false,
-    screenShareButton: props.mobileView ? false : { disabled: props.disableButtonsForLobbyPage }
+    screenShareButton:
```
prprabhu-ms, line 36: This can be simplified as 

const callControlOptions = {
  ...props.meetingCallControlOptions,
  screenShareButton: [CONTINUE TO OVERRIDE WHAT YOU NEED]
}

dmceachernmsft: oh fun! yeah I'll update that!

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -23,17 +23,28 @@ export interface MeetingCallControlBarProps {
...
+      props.mobileView || !props.meetingCallControlOptions?.screenShareButton
+        ? false
+        : { disabled: props.disableButtonsForLobbyPage },
+    displayType: props.meetingCallControlOptions?.displayType,
+    microphoneButton: !props.meetingCallControlOptions?.microphoneButton ? false : true,
+    cameraButton: !props.meetingCallControlOptions?.cameraButton ? false : true,
+    devicesButton: !props.meetingCallControlOptions?.devicesButton ? false : true
```
prprabhu-ms, line 43: Explicit values of `true` and `false` feel very unergonomic. Also, if/when `deviceButton` becomes an object with more internal options, this will need to be unnecessarily updated.

Why is this isn't just `deviceButton: props.meetingCallControlOptions?.deviceButton` ? If there is a reason, please add a comment in the code explaining why. It confused me, might confused the next reader as well.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -23,17 +23,28 @@ export interface MeetingCallControlBarProps {
...
   };
+  // Set flags for the chat button and people button specific to the meeting control bar.
+  const chatButton = props.meetingCallControlOptions?.chatButton === false ? false : true;
```
prprabhu-ms, line 43: I'm guessing this is trying to check if (1) the prop was explicitly set, (2) it was set to `false`.

That can simply be `!(props.meetingCallControlOptions?.chatButton === false)`

* I'd  inline that check in line 71. Separating that logic from its use isn't helpful.
* I'd also create a helper function below to make the intent clearer.

  const isEnabled(option: boolean | undefined): boolean => !(option === false); 


Then, line 71 becomes

  { isEnabled(props.meetingCallControlOptions?.chatButton) && (<ChatButton>....</ChatButton>) }

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
dmceachernmsft: Fix for previous comment also resolves this. was just making sure the correct logical value is passed down to the callcontrols component. but just passing the meetingControls prop down there directly removes the need for these explicit definitions.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
dmceachernmsft: done! thanks for the suggestion! 😊

JamesBurnside (approved):

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -23,18 +23,26 @@ export interface MeetingCallControlBarProps {
...
     participantsButton: false,
-    screenShareButton: props.mobileView ? false : { disabled: props.disableButtonsForLobbyPage }
+    screenShareButton:
+      props.mobileView || !props.meetingCallControlOptions?.screenShareButton
+        ? false
+        : { disabled: props.disableButtonsForLobbyPage }
   };
 
+  // Helper function to determine if a meeting control bar button is enabled or not.
+  const isEnabled = (option: boolean | undefined): boolean => !(option === false);
```
prprabhu-ms, line 44: nitty nit: I'd make this a file-level private function. Why create a new function object on each render?

prprabhu-ms (approved):

prprabhu-ms:
Thanks for waiting and rebasing to `main`.

PR #1258, opened 2021-12-17
Allow `minor` and `patch` change options when creating changelog files

JamesBurnside:
# What
* Allow `minor` and `patch` change options, alongside `none` and `prerelease` when creating changelog files
* Update existing changefiles pushed this week

# Why
We are no longer entirely prerelease! For now I've left prerelease in as an option for beta features - I need to play around with the beachball generate-changelog command to check this will work and bump packages as we desire.


alcail (approved):

alcail:
Just curious to understand the difference between "patch" and "minor" and when do we have "none"

common/config/beachball/changelog-config.ts
```diff
@@ -27,9 +27,7 @@ export const config: BeachballConfig = {
       name: "@azure/communication-react and its packlets",
       include: "packages/*",
       disallowedChangeTypes: [
-        "major",
-        "minor",
-        "patch"
+        "major"
```
prprabhu-ms, line 30: We should also disallow `prerelease`?

prprabhu-ms (approved):

common/config/beachball/changelog-config.ts
JamesBurnside: Mentioned this in the PR description, I'm leaving it in to see if we should it for beta stuff

PR #1262, opened 2021-12-17
Release nightly alpha package off `main` again

prprabhu-ms:
# What

Reverts #1193 

# Why

Feature branch is now gone and beta features have been merged back into `main`.

# How Tested
Will kick off action once this lands.


JamesBurnside:
Confirmed main will produce the correct alpha strings for packages (I set datetime suffix to `123`):
image

However one thing to note, I think when we create a beta release branch it will be 1.1.0-beta.0 (because minor changes) so there might be a weird drift of 1.0.1-alpha and 1.1.0-beta that we'll need to fix

JamesBurnside (approved):

anjulgarg (approved):

prprabhu-ms:
> Confirmed main will produce the correct alpha strings for packages (I set datetime suffix to `123`): image
> 
> However one thing to note, I think when we create a beta release branch it will be 1.1.0-beta.0 (because minor changes) so there might be a weird drift of 1.0.1-alpha and 1.1.0-beta that we'll need to fix

Still thinking about the most natural way to bump versions with our stable/beta release channels.
I think I'd like to see it being `1.1.0-alpha*` and `1.1.0-beta.*` at this point -- *if* I were to cut a stable release off `main` right now, what would the version be? That's the version we should have for the prerelease packages as well.

PR #1263, opened 2021-12-17
Update meeting composite UI tests with recent updates, and apply more util fns

JamesBurnside:
# What
* Ensure meeting composite has unique chat thread per test
* Update meeting composite with util functions that take a screenshot when they timeout
* Update meeting composite to use the chat util helpers
* Ensure consistent use of bringToFront() by moving into util fn (for all tests)


prprabhu-ms (approved):
Meeting is back in action!

This is why I wanted it in `main` already :)

anjulgarg (approved):

PR #1265, opened 2021-12-20
Bug 2564703 - [Call composite - control bar navigation]: Icon is not clearly visible on All buttons in aquatic mode under 'Calling window' page.

carocao-msft:
This PR is for Bug 2564703 - [Call composite - control bar navigation]: Icon is not clearly visible on All buttons in aquatic mode under 'Calling window' page.

## Issues Linked 
https://skype.visualstudio.com/SPOOL/_workitems/edit/2564703

## Changes
Added custom styling on buttons

## Checklist

Tested on Mac accessibility mode. 
Need testing on Windows

## Testing Steps
1. Open the above URL and login with v-credential.
2. 'Start or Join a call' page will be open.
3. Enter your name in 'Enter a name' edit field.
4. Navigate to 'Next Button' using tab key and activate it using 'Enter' key.
5. 'Start a call' page will be open.
6. Navigate to start call button using tab key and active it using enter key.
7. Call will be started and Calling window will be appear.
8. Navigate to 'Share' button using tab key.
9. Observe Keyboard focus boundary with respect to 'Share button in 'Aquatic' mode


packages/react-components/src/components/MicrophoneButton.tsx
```diff
@@ -56,6 +57,24 @@ export const MicrophoneButton = (props: MicrophoneButtonProps): JSX.Element => {
   const localeStrings = useLocale().strings.microphoneButton;
   const strings = { ...localeStrings, ...props.strings };
 
+  const theme = useTheme();
+  const onRenderMicOnIcon = (): JSX.Element => (
+    <Icon
```
prprabhu-ms, line 62: (1) Line 46/47 are now dead code?

(2) Defining these inline will cause extra renders for the icons. Consider using `useMemo` to only render when `them` or `props.disabled` changes.

(3) Reading the code below, it is not obvious to me why setting the theme colors fixes the high contrast issue. Could you:
  (a) Add a comment in the code explaining why the theme colors are needed (perhaps also what the default behavior was before adding these)
  (b) Extract a helper component since this is needed for all buttons (this also takes care of the extra rendering issue):

```ts

const HighContrastAwareIcon(props: { iconName: string, disabled: boolean}): JSX.Element => {
    const theme = useTheme();
    // Comment here describing why we need to set the colors from theme.
    return <Icon iconName={props.iconName} className={.... etc. }>
}
```

prprabhu-ms:
Here is the control bar looks before and after this PR in high contrast themes:

Theme | Before PR | After PR
:-: | :-: | :-: 
Aquatic | image | image
Desert | image | image
Dusk | image | image
Night Sky | image | image

dmceachernmsft (approved):

packages/react-components/src/components/HighContrastAwareIcon.tsx
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { Icon, mergeStyles } from '@fluentui/react';
+import React from 'react';
+import { useTheme } from '../theming';
+
+/**
+ * Strings of HighContrastAwareIcon that can be overridden.
```
prprabhu-ms, line 9: nit: This comment isn't quite true. The `props` include more than strings. The strings comment is more useful for publicly exported components.

packages/react-components/src/components/HighContrastAwareIcon.tsx
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @public
```
prprabhu-ms, line 11: This should be @private, and should not be exported from the `@internal/react-components` package.

prprabhu-ms (changes requested):
There is an unintentional change to the icon fill color for `EndCall`: 
image

I think we want to revert that.

packages/react-components/src/components/HighContrastAwareIcon.tsx
carocao-msft: I changed both interface and component declaration to private since this is a helper component for the Calling sample only

carocao-msft:
> There is an unintentional change to the icon fill color for `EndCall`: image
> 
> I think we want to revert that.

Reverted changes to the icon color, but changed background color from 'red' to 'redDark', not sure if this can fix the contrast issue on Windows

prprabhu-ms (approved):
Let's wait for the UI snapshot update automation to update the snapshots and make sure that the endcall button now looks correct.

Thanks for the iterations on this PR.

PR #1266, opened 2021-12-20
Dmceachernmsft/MeetingControlBarBooleanType

dmceachernmsft:
# What
Added to the MeetingCallControlOptions prop to allow a False boolean value.

# Why
Adds functionality to hide the whole control bar for the call in the meeting composite.
 added as per request in comment: 
image

# How Tested
ran sample in story book with different combinations of props.


packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -94,7 +94,17 @@ export type CallControlOptions = {
 export const CallControls = (props: CallControlsProps): JSX.Element => {
   const { callInvitationURL, onFetchParticipantMenuItems } = props;
 
-  const options = typeof props.options === 'boolean' ? {} : props.options;
+  const options =
+    typeof props.options === 'boolean'
+      ? {
```
dmceachernmsft, line 99: This seems a little verbose for the call controls but for some reason there is two different behaviors here. When the prop is passed in from the meeting composite as just false it works fine, but when the meeting control bar passes in the same value of false on line 76 of this file the whole call controls bar shows up. Thoughts here as to what might be happening?

.github/workflows/nightly-cd.yml
```diff
@@ -19,6 +19,9 @@ jobs:
     steps:
       # Check-out repo
       - uses: actions/checkout@v2
+        with:
```
prprabhu-ms, line 22: Unstage / remove these changes that crept in due to merge conflict.

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -46,7 +46,7 @@ export interface MeetingCompositeProps extends BaseCompositeProps<CallCompositeI
   /**
    * Call control options to change what buttons show on the meeting composite control bar.
    */
-  meetingCallControlOptions?: MeetingCallControlOptions;
+  meetingCallControlOptions?: boolean | MeetingCallControlOptions;
```
prprabhu-ms, line 49: Please add a comment explaining what the value of `true` or `false` vs the options struct does.

packages/communication-react/review/communication-react.api.md
```diff
@@ -1435,7 +1435,7 @@ export interface MeetingCompositeProps extends BaseCompositeProps<CallCompositeI
     formFactor?: 'desktop' | 'mobile';
     // (undocumented)
     meetingAdapter: MeetingAdapter;
-    meetingCallControlOptions?: MeetingCallControlOptions;
+    meetingCallControlOptions?: boolean | MeetingCallControlOptions;
```
prprabhu-ms, line 1438: Should we extract `MeetingCompositeOptions` similar to `CallCompositeOptions`?

dmceachernmsft: maybe extend it so we can pass those options directly to the calling composite from the meeting composite? I think the other properties in the calling composite will follow 'default' behavior otherwise?

prprabhu-ms: Yeah, let's do that as a follow up PR regardless.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -32,7 +32,7 @@ import {
 export type CallControlsProps = {
   callInvitationURL?: string;
   onFetchParticipantMenuItems?: ParticipantMenuItemsCallback;
-  options?: boolean | CallControlOptions;
+  options?: true | CallControlOptions;
```
prprabhu-ms, line 35: Now that you return `<></>` on line 139, this should continue to be `boolean`? And we should update `CallPage.tsx` and `LobbyPage.tsx` to always create and pass in `CallControlProps` (even when `CallControlOptions` is `false`?

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -135,7 +135,10 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
     () => mergeButtonBaseStyles(props.increaseFlyoutItemSize ? devicesButtonWithIncreasedTouchTargets : {}),
     [props.increaseFlyoutItemSize]
   );
-
+  // when props.options is false then we want to hide the whole control bar.
+  if (props.options === false) {
+    return <></>;
+  }
```
prprabhu-ms, line 141: I'd recommend moving line 97 below this  and changing it to

```ts
const options = props.options === true ? {} : props.options
```

to better reflect what's happening.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -47,7 +62,7 @@ export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.El
   const isEnabled = (option: boolean | undefined): boolean => !(option === false);
 
   // Reduce the controls shown when mobile view is enabled.
-  if (props.mobileView) {
+  if (props.mobileView && typeof callControlsOptions !== 'boolean') {
```
prprabhu-ms, line 65: It think it would be preferable if `reduceCallControlsForMobile` can correctly handle boolean options as well.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -62,31 +77,36 @@ export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.El
...
-          <PeopleButton
-            checked={props.peopleButtonChecked}
-            showLabel={true}
-            onClick={props.onPeopleButtonClicked}
-            data-ui-id="meeting-composite-people-button"
-            disabled={props.disableButtonsForLobbyPage}
-          />
-        )}
-      </Stack.Item>
+      {callControlsOptions !== false && (
```
prprabhu-ms, line 80: Since `CallControls` now returns `<></>` when options are `false`, this check should no longer be required (once you allow `boolean` in the props again).

samples/Chat/src/app/ChatHeader.tsx
```diff
@@ -27,17 +27,17 @@ export const ChatHeader = (props: ChatHeaderProps): JSX.Element => {
...
   return (
     <Stack horizontal={true} verticalAlign={'center'} horizontalAlign="end" className={chatHeaderContainerStyle}>
-      <IconButton
-        onRenderIcon={() => (props.isParticipantsDisplayed ? <People20Filled /> : <People20Regular />)}
-        className={mergeStyles(paneButtonContainerStyle, { color: theme.palette.neutralPrimaryAlt })}
-        onClick={() => props.setHideParticipants(props.isParticipantsDisplayed)}
-        aria-live={'polite'}
-        aria-label={peopleString}
-      />
+      {/** FIXME: Similar to #1235 */}
```
prprabhu-ms, line 33: This leaked in from a bad merge. Please revert changes to `ChatHeader.tsx`.

prprabhu-ms (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
dmceachernmsft: I'll change that back to boolean and investigate the other two pages 😊

anjulgarg (approved):

PR #1271, opened 2021-12-21
Dmceachernmsft/MeetingCompsiteAlternateProp

dmceachernmsft:
# What
remove the boolean type for the ui options props for the meeting composite and call control component
# Why
Alternate change to PR: https://github.com/Azure/communication-ui-library/pull/1266/files 

# How Tested
Ran locally in story book with different meeting composite options props.


prprabhu-ms (changes requested):
Discussed that #1266 is the right approach.

dmceachernmsft:
Other options in pull 1266 this is based off of is better. Closing as a result.

PR #1272, opened 2021-12-22
Dmceachernmsft/MeetingCompositeOptions

dmceachernmsft:
# What
add options prop to the meeting composite.

# Why
matches the meeting composites props with the call composite.

# How Tested
Work in progress, story book has some current complications with complex props like the options prop.


packages/communication-react/review/communication-react.api.md
```diff
@@ -1426,6 +1426,11 @@ export interface MeetingCallControlOptions extends Pick<CallControlOptions, 'cam
 // @beta
 export const MeetingComposite: (props: MeetingCompositeProps) => JSX.Element;
 
+// @beta
+export type MeetingCompositeOptions = {
+    meetingCallControlOptions?: boolean | MeetingCallControlOptions;
```
anjulgarg, line 1431: I think the prefix `meeting` and suffix `options` here is verbose.
I suggest naming it simply `callControls`

anjulgarg (approved):

packages/communication-react/review/communication-react.api.md
dmceachernmsft: Sounds good will make it so!

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -113,6 +124,12 @@ const MeetingScreen = (props: MeetingScreenProps): JSX.Element => {
     setShowPeople(!showPeople);
   }, [showPeople]);
 
+  const chatProps: ChatCompositeProps = useMemo(() => {
```
mgamis-msft, line 127: Minor: Is it necessary to use another `useMemo` for chatProps since there is already a `useMemo` for the chatAdapter on line 89? Maybe we can simply have it as:
```
const chatProps: ChatCompositeProps = { adapter: chatAdapter };
```

prprabhu-ms: That will still create a new object on each render.

We _should_ reduce to one `useMemo` though. Perhaps as:

```ts
const chatProps: ChatCompositeProps = useMemo(() => ({
    adapter: new MeetingBackedChatAdapter(meetingAdapter)
}), [meetingAdapter])
```

And then use `chatProps.adapter` wherever we need access to the adapter directly.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -33,18 +33,18 @@ export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.El
   // Set the desired control buttons from the meetings composite. particiapantsButton is always false since there is the peopleButton.
   let meetingCallControlOptions: MeetingCallControlOptions | undefined;
   let callControlsOptions: CallControlOptions | boolean;
-  if (typeof props.meetingCallControlOptions === 'boolean') {
+  if (typeof props.callControls === 'boolean') {
```
prprabhu-ms, line 36: Not _necessarily_ in this PR: This will read better if we (1) keep the logic for the two options separate, (2) use immutable local variables.

So:
```ts

const meetingCallControlOptions = inferMeetingCallControlOptions(props.callControls);
const callControlOptions = inferCallControlOptions(props.callControls);

....

const inferMeetingCallControlOptions = (callControls?: boolean | MeetingCallControlOptions): MeetingCallControlOptions | undefined {
    if (typeof callControls !== 'boolean') {
        return callControls;
    }
    if (callControls === true) {
        // I'm actually confused by this return value. Please add a comment why.
        return undefined;
    }
    
    // callControls === false
    return {
        chatButton: false,
        peopleButton: false,
    }
}

... etc.
```

prprabhu-ms (approved):

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
dmceachernmsft: okay i'll look at this!

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
dmceachernmsft: I will play with this as well. I dont think the logic quite matches but I agree that this is the better solution to pull the logic out for readability.

mgamis-msft (approved):

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -27,35 +27,46 @@ export interface MeetingCallControlBarProps {
...
+    return undefined;
+  }
+  // callControls === false
+  return false;
+};
+
+const inferCallControlOptions = (
+  callControls?: boolean | CallControlOptions
+): CallControlOptions | false | undefined => {
+  // let meetingParticipantButton = { participantsButton: false }; // always hide the participant button in meeting composite.
```
prprabhu-ms, line 47: Delete?

prprabhu-ms (approved):

PR #1274, opened 2021-12-22
Dmceachernmsft/meetingCompositeLocalesIcons

dmceachernmsft:
# What
Add localization Strings for meeting composite specific buttons. Break up meeting composite to make more readable when wrapped by a basic composite.

# Why
works toward larger localization support for the meeting composite.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2648151

# How Tested
Ran locally in story book with different locals to see if strings changed appropriately. More work to come for specific meeting buttons and screens.

# Process & policy checklist

### DIFF
--- a/packages/communication-react/review/communication-react.api.md
+++ b/packages/communication-react/review/communication-react.api.md
@@ -864,6 +864,7 @@ export interface CompositeLocale {
 export interface CompositeStrings {
     call: CallCompositeStrings;
     chat: ChatCompositeStrings;
+    meeting: MeetingCompositeStrings;
 }
 
 // @public
@@ -1437,6 +1438,15 @@ export interface MeetingCompositeProps extends BaseCompositeProps<CallCompositeI
     meetingAdapter: MeetingAdapter;
     meetingCallControlOptions?: boolean | MeetingCallControlOptions;
     meetingInvitationURL?: string;
+}
+
+// @public
+export interface MeetingCompositeStrings {
+    chatButtonLabel: string;
+    chatPaneTitle: string;
+    peopleButtonLabel: string;
+    peoplePaneSubHeader: string;
+    peoplePaneTitle: string;
 }
 
 // @beta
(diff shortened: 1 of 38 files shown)


packages/communication-react/review/communication-react.api.md
```diff
@@ -1439,6 +1440,15 @@ export interface MeetingCompositeProps extends BaseCompositeProps<CallCompositeI
     meetingInvitationURL?: string;
 }
 
+// @public
+export interface MeetingCompositeStrings {
+    chatButtonLabel: string;
+    chatPaneTitle: string;
+    peopleButtonLabel: string;
+    peoplePaneSubHeader: string;
```
prprabhu-ms, line 1448: nit: 
Title vs SubHeader. Should this be SubTitle?

prprabhu-ms (approved):

packages/communication-react/review/communication-react.api.md
anjulgarg: +1 to SubTitle

samples/Chat/src/app/App.tsx
```diff
@@ -16,7 +15,6 @@ console.info(`Thread chat sample using @azure/communication-chat : ${getChatSDKV
 console.info(`Build Date : ${getBuildTime()}`);
 
 initializeIcons();
-registerIcons({ icons: DEFAULT_COMPONENT_ICONS });
```
anjulgarg, line 19: Why is it not needed anymore?

anjulgarg: Is it because, composites registerIcons be default?

anjulgarg (approved):

samples/Chat/src/app/App.tsx
dmceachernmsft: Yeah the base composite registers them anyway and then takes in the icons its overriding from the prop. this causes the overriden icons to be hidden if left in.

PR #1279, opened 2021-12-24
Dmceachernmsft/singletonProvider

dmceachernmsft:
# What
Update the base composite to a singleton provider.

# Why
By changing it and adding logic to check to see if its nested in a instance of itself it helps prevent the duplication of Icon registration which can cause errors.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2676992?src=WorkItemMention&src-action=artifact_link
# How Tested

Locally ran the meeting composite and samples to make sure functionality isnt affected and errors are removed in meeting composite.


alkwa-msft:
If the developer is already using Fluent icons they will have issues when this singleton comes in and thinks its doing the right thing and its not.

can we not push this issue to contoso?

dmceachernmsft:
> If the developer is already using Fluent icons they will have issues when this singleton comes in and thinks its doing the right thing and its not.
> 
> can we not push this issue to contoso?

The issue this is trying to resolve is arising from within the MeetingComposite itself so this is something that we are also running into internally when making the rendering the different Fluent Icons.

packages/react-composites/src/composites/common/SingletonProvider.tsx
```diff
@@ -52,16 +52,23 @@ export interface BaseCompositeProps<TIcons extends Record<string, JSX.Element>>
...
  *
  * @private
  */
-export const BaseComposite = (
+export const SingletonProvider = (
   props: BaseCompositeProps<CallCompositeIcons | ChatCompositeIcons> & { children: React.ReactNode }
 ): JSX.Element => {
   const { fluentTheme, rtl, locale } = props;
 
+  const alreadyWrapped = useSingleton();
```
JamesBurnside, line 67: Curious if something like this would work where instead of having to have a singleton that checks if the composite has been wrapped we instead check if the icons themselves have already been registered:
```typescript
  // Only register icons that have not already been registered
  const icons = Object.entries({ ...DEFAULT_COMPOSITE_ICONS, ...props.icons });
  const iconsToBeRegistered = icons.filter(([iconName]) => IS_ALREADY_REGISTERED(iconName));
  registerIcons({ icons: Object.fromEntries(iconsToBeRegistered) });
```

There doesn't appear to be an immediate way from fluent's api to check if an Icon is already registered, looking at how they store their registered icons (https://github.com/microsoft/fluentui/blob/f26fbc8c9b7978efb5b3182ae586b74a5c93db77/packages/style-utilities/src/utilities/icons.ts#L60) it would be something like
```typescript
import { GlobalSettings, IIconRecord } from '@fluentui/react';
...
const previouslyRegisteredIcons = Object.keys(GlobalSettings.getValue<IIconRecords>('icons'));
const IS_ALREADY_REGISTERED = (iconName) => previouslyRegisteredIcon.includes(iconName.toLowerCase());
```
But we wouldn't want to do this unless a function exposed from fluent's library did this (we could contribute to the fluent library and add this function however)

prprabhu-ms: nuke option: 
image


https://github.com/microsoft/fluentui/wiki/Using-icons#library-icon-registration

JamesBurnside: Another options to solve this for meetings composite, is for the BaseComposite to take a prop called
```ts
interface BaseCompositeProps {
...
  registerIcons: boolean // default true
}
```
that meeting compsite can do
```tsx
<CallComposite registerIcons={false} />
...
<ChatComposite registerIcons={false} />
```

JamesBurnside:
Closing this as over a month old and is just being used for discussion. Linked inside bug so won't be lost

PR #1289, opened 2022-01-10
Fixed inability to click the horizontal gallery left/right button by updating zIndex.

prabhjot-msft:
# What
Fixed inability to click the horizontal gallery left/right button by updating zIndex.

# Why
https://skype.visualstudio.com/SPOOL/_queries/edit/2674355/?fullScreen=true&triage=true

# How Tested
Manually

# Process & policy checklist


change/@internal-react-components-7d1c8b1b-8763-476d-800e-d4e79b33c21a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for horizontal gallery left/right button click",
```
JamesBurnside, line 3: For this and PR description&title, mention what the fix is. The reason for this is we us these to generate the changelog so if you view this from the lens of a developer reading the changelog - does this help me understand what has changed in this new package version? (And liekwise as a PR reviewer - what change am I expecting this PR to solve) See more examples/info here: https://github.com/Azure/communication-ui-library/blob/main/docs/references/tips-for-writing-changelog-entries.md

This could be something like:
```suggestion
  "comment": "Fix inability to click the horizontal gallery left/right button by updating zIndex.",
```

packages/react-components/src/components/styles/HorizontalGallery.styles.ts
```diff
@@ -15,6 +15,7 @@ export const leftRightButtonStyles = (theme: Theme): IStyle => {
   return {
     background: 'none',
     padding: 0,
+    zIndex: 1,
```
JamesBurnside, line 18: css properties are very arbitrary and hard to understand why they exist (and thus changing css often causes regressions elsewhere). Add a comment here to say why zIndex is set here (i.e. that the buttons must exist in a higher z band than <insert other element that is causing this bug>)

prprabhu-ms: +1

Actually, I like to define constants side-by-side for all elements that have relative z-index.

e.g.

```ts
// Relative zIndex values to stack contents of the horizontal gallery 
HORIZONTAL_GALLERY_CONTAINER_ZINDEX = 0
HORIZONTAL_GALLERY_NEXT_BUTTON_ZINDEX = 1
HORIZONTAL_GALLERY_PREV_BUTTON_ZINDEX = 1
```

That way, it's obvious what the stacking is like.

change/@internal-react-components-7d1c8b1b-8763-476d-800e-d4e79b33c21a.json
prabhjot-msft: Done, updated the PR description&title.

packages/react-components/src/components/styles/HorizontalGallery.styles.ts
prabhjot-msft: Done, added comment and also used constant for adding higher zIndex for left/right buttons.

change/@internal-react-components-7d1c8b1b-8763-476d-800e-d4e79b33c21a.json
JamesBurnside: Can you update this change file also?

The change file is annoying to find in vscode but you can grab the name with the copy button in the github UI here:
image
the ctrl+p in vscode and paste and enter should open the file

samples/tests/playwright.config.ts
```diff
@@ -17,6 +17,8 @@ const config: PlaywrightTestConfig = {
       use: {
         browserName: 'chromium',
         permissions: ['notifications', 'camera', 'microphone'],
+        // Large enough to contain all UI elements from ComponentExamples sample
```
JamesBurnside, line 20: fine to leave in for this PR, but in future PRs try to keep changes to solving one issue. So setting this viewport would ideally be its own PR tackling the issue of "Fix UI snapshot not capturing all component examples"

samples/tests/playwright.config.ts
```diff
@@ -17,6 +17,8 @@ const config: PlaywrightTestConfig = {
...
+        viewport: { width: 900, height: 900 },
```
JamesBurnside, line 21: Instead of setting a larger viewport, could we use the playwright API to take a whole page screenshot: https://playwright.dev/docs/screenshots#full-page-screenshots
(from the documentation it appears this should capture elements outside the immediate view, but you would need to test if this works)

packages/react-components/src/components/styles/HorizontalGallery.styles.ts
mgamis-msft: Sorry to join the discussion late but would a better alternative be setting the CSS property 'pointerEvents' of the layerHost of VideoGallery to be 'none' so that it does not block any other buttons we may add to VideoGallery in the future. It might get tedious to keep applying zIndex properties in VideoGallery elements. What do you guys think?

JamesBurnside: ^if that solves it that would be a much cleaner fix, especially if we introduce video gallery tiles that are interactive

samples/tests/playwright.config.ts
prprabhu-ms: iirc, we're already taking full page screenshots...

prprabhu-ms: .. and this is a change leaked in from a PR that Edward landed this morning. Should disappear if you merge from `origin/main` :)

prabhjot-msft:
@JamesBurnside @prprabhu-ms I will create a new PR for this change to show only changes related to bug-2674355.

PR #1290, opened 2022-01-10
Added viewport size to capture contents in ComponentExamples test

edwardlee-msft:
# What
Added viewport size to capture additional contents in ComponentExamples. Moved final mouse location in test. 

# Why
Viewport was added in order to include/fit ControlBar into the ComponentExamples test for BUG 2677269. Additionally, moved final mouse location to avoid hover interaction.

# How Tested
Ran ComponentExample tests locally with and without the addition of ControlBar and compared resulting screenshots.

# Process & policy checklist


samples/tests/ComponentExamples/ComponentExamples.test.ts
```diff
@@ -58,6 +58,8 @@ test.describe('Component Examples Test', () => {
       return document.querySelectorAll('video').length === 0;
     });
 
+    await page.mouse.move(0, 0);
```
prprabhu-ms, line 61: Please add a comment above this saying why we need to move the mouse away. Something like, "Move the mouse off-screen to avoid tooltips / delayed focus from introducing flake in the snapshots).

prprabhu-ms (approved):
Thanks

samples/tests/playwright.config.ts
```diff
@@ -17,6 +17,7 @@ const config: PlaywrightTestConfig = {
       use: {
         browserName: 'chromium',
         permissions: ['notifications', 'camera', 'microphone'],
+        viewport: { width: 900, height: 900 },
```
prprabhu-ms, line 20: Add a comment above saying why this is needed, e.g.: "Large enough to contain all the UI elements added in ComponentExamples sample".

samples/tests/ComponentExamples/ComponentExamples.test.ts
edwardlee-msft: Added comment on newest commit:
// Move the mouse off-screen to avoid tooltips / delayed focus from introducing flake in the snapshots

PorterNan (approved):
Thanks for taking care of this viewport size problem - as we discussed offline, it would be nice to show the scroll bar, and that will make full-page-screenshot functionality work for playwright
Approved, since either current method or 'scroll-bar' method looks good to me for fixing the problem

PR #1291, opened 2022-01-10
Add flavor(variants) for stable/beta

PorterNan:
# What

Add flavor switch for the repo, including: 

rush switch-flavor:beta
rush switch-flavor:stable
rush update:stable
rushx current-flavor

Once run `rush switch-flavor:<stable/beta>` and `rush update<:stable>`, ou can just follow all the existing process to build the project

To try stable flavor:
`rush switch-falvor:stable` then
`rush build` or `rush test`

the only difference when you are under stable flavor is `rush update` -- run `rush update:stable` 

 
If you think you have a better name candidate, please leave a comment for those commands 😄 
# Why

# How Tested

# Process & policy checklist


common/config/env/beta.env
```diff
@@ -0,0 +1 @@
+FLAVOR=beta
```
prprabhu-ms, line 1: `FLAVOR` is too generic a name for an envvar.

Can we name this `COMMUNICATION_REACT_FLAVOR`?

common/config/rush/variants/stable/common-versions.json
```diff
@@ -0,0 +1,65 @@
...
+   * the "dependencies" field of the top-level common/temp/package.json, which influences how the package manager
+   * will calculate versions.  The specific effect depends on your package manager.  Generally it will have no
+   * effect on an incompatible or already constrained SemVer range.  If you are using PNPM, similar effects can be
+   * achieved using the pnpmfile.js hook.  See the Rush documentation for more details.
+   *
+   * After modifying this field, it's recommended to run "rush update --full" so that the package manager
+   * will recalculate all version selections.
+   */
+  "preferredVersions": {
+    /**
```
prprabhu-ms, line 22: Please add a comment here describing what the value here should be (i.e., the one used in the stable build)

common/config/rush/variants/stable/common-versions.json
```diff
@@ -0,0 +1,65 @@
...
+   * the same SemVer range for a given dependency.  However, sometimes exceptions are needed.
+   * The allowedAlternativeVersions table allows you to list other SemVer ranges that will be
+   * accepted by "rush check" for a given dependency.
+   *
+   * IMPORTANT: THIS TABLE IS FOR *ADDITIONAL* VERSION RANGES THAT ARE ALTERNATIVES TO THE
+   * USUAL VERSION (WHICH IS INFERRED BY LOOKING AT ALL PROJECTS IN THE REPO).
+   * This design avoids unnecessary churn in this file.
+   */
+  "allowedAlternativeVersions": {
+    "@azure/communication-calling": ["1.3.2"],
```
prprabhu-ms, line 55: Is this necessary? I thought that `allowedAlternativeVersions` is intended so that different *packlets* can use different versions at the same time.

common/config/rush/variants/stable/pnpmfile.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+/**
```
prprabhu-ms, line 3: Err... why is this file needed in our case? Please remove or document.

`readPackage` doesn't have any overrides at all...

prprabhu-ms (approved):

prprabhu-ms:
This is way better with fewer end-user commands per package.json

packages/acs-ui-common/package.json
```diff
@@ -6,25 +6,24 @@
...
-    "clean": "rimraf dist",
-    "copy-original": "cpx ./src/** ./preprocessed",
-    "preprocess": "rushx copy-original && babel ./src --out-dir ../preprocessed --extensions \".ts,.tsx\" --keep-file-extension --config-file ./.babelrc.js --relative && rimraf ../preprocessed",
-    "preprocess:stable": "cross-env FLAVOR=stable rushx preprocess",
-    "clean:preprocess": "rimraf ./preprocessed",
-    "build:stable": "rushx clean:preprocess && rushx preprocess:stable && tsc -project tsconfig.preprocess.json && rushx build:cjs && api-extractor run --local",
-    "test:stable": "rushx clean:preprocess && rushx preprocess:stable && jest --verbose ./preprocessed -c jest.config.preprocess.js",
+    "clean": "rimraf dist && rimraf preprocessed",
+    "current-flavor": "echo You are running under FLAVOR: && env-cmd -f ../../common/config/env/.env node -p process.env.FLAVOR",
+    "by-flavor": "rushx current-flavor && env-cmd -f ../../common/config/env/.env --use-shell",
```
prprabhu-ms, line 15: All the `by-flavor` variants are not intended to be run directly. Could we

* Add a prefix to all such commands, perhaps `plumbing-` or `internal-`
* Move all of them to the bottom of the `scripts`. Perhaps prefix with `zzPlumbing-`, so that sorting will naturally drop them to the bottom of the list.

packages/react-composites/package.json
```diff
@@ -158,7 +148,7 @@
     "webpack-dev-server": "3.11.2",
     "webpack-cli": "4.7.2",
     "@babel/cli": "~7.16.0",
-    "@babel/plugin-syntax-typescript": "~7.16.0",
-    "babel-plugin-conditional-compilation": "~0.0.1"
+    "env-cmd": "~10.1.0",
```
JamesBurnside, line 151: nit: order deps alphabetically

packages/react-composites/package.json
```diff
@@ -54,7 +44,7 @@
     "lint:quiet": "npm run lint -- --quiet"
   },
   "dependencies": {
-    "@azure/communication-calling": "1.3.2",
+    "@azure/communication-calling": "1.3.2-beta.1 || 1.3.2",
```
JamesBurnside, line 47: fyi in case this is completed first: #1294

common/config/rush/variants/stable/pnpmfile.js
PorterNan: That was required by rush variant, according to their instruction:
https://rushjs.io/pages/advanced/installation_variants/#:~:text=shrinkwrap.yaml%2C%20npm%2Dshrinkwrap,as%20your%20package%20manager

common/config/rush/variants/stable/common-versions.json
PorterNan: The rush consistent version checker only supports string comparing, so this one is a workaround - when you bump beta/stable sdk version, you need to also bump this one, so this one will be always in allow list (we have 2 different allowedVersions in stable and beta, which is good)

packages/react-composites/package.json
prprabhu-ms: Actually, @PorterNan how will rush variants work with peerDependencies and devDepenedencies?

Thanks for linking to my PR James.

prprabhu-ms: My understanding is that rush variants work at the level of pnpm. So it doesn't matter why we try to install a version, it just switches out the version actually installed. But it'd be worth checking.

packages/acs-ui-common/package.json
PorterNan: To make it short, I've changed the name with prefix "_" (how we name internal functions and variables) and they've been moved to the bottom ;D

packages/react-composites/package.json
PorterNan: Will merge the main branch and take a test then

common/config/rush/variants/stable/common-versions.json
prprabhu-ms: Please add this comment in the code.

packages/react-composites/package.json
PorterNan: Yea, it works when peerDependencies are stable version, but devDependencies with "1.3.2-beta.1 || 1.3.2"

JamesBurnside (approved):

PR #1294, opened 2022-01-11
Move dependencies we reexpose in our API to `peerDependencies` from `dependencies`

prprabhu-ms:
# What

Move `@azure/communication-calling` and `@azure/communication-chat` to peer dependencies for `@azure/communication-react` and associated packlets.

# Why

* Because we expose critical classes from these dependencies as part of our own API, they should be peer dependencies -- our client is expected to also depend on these packages in order to be able to use our API, so having private installs of the dependencies can be dangerous.... 
* .... and this breaks `@azure/communication-calling` as expected. Two installs of the package sometimes don't play well together. They definitely have caused problems when two different versions get installed due to different dependency versions.

# How Tested

<html>
<body>
<div style="box-sizing: border-box; margin: 0px; color: rgba(0, 0, 0, 0.9); font-family: &quot;Segoe UI VSS (Regular)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, Helvetica, Ubuntu, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;, &quot;Segoe UI Symbol&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">In a test app, played around with the different cases:</div><div style="box-sizing: border-box; margin: 0px; color: rgba(0, 0, 0, 0.9); font-family: &quot;Segoe UI VSS (Regular)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, Helvetica, Ubuntu, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;, &quot;Segoe UI Symbol&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><br style="box-sizing: border-box;"></div><div style="box-sizing: border-box; margin: 0px; color: rgba(0, 0, 0, 0.9); font-family: &quot;Segoe UI VSS (Regular)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, Helvetica, Ubuntu, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;, &quot;Segoe UI Symbol&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">This table shows where @azure/communication-calling is installed, for different cases of how the package.json for communication-react and the client app respectively specify the dependency.</div><div style="box-sizing: border-box; margin: 0px; color: rgba(0, 0, 0, 0.9); font-family: &quot;Segoe UI VSS (Regular)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, Helvetica, Ubuntu, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;, &quot;Segoe UI Symbol&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><br style="box-sizing: border-box;"></div><div style="box-sizing: border-box; margin: 0px; color: rgba(0, 0, 0, 0.9); font-family: &quot;Segoe UI VSS (Regular)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, Helvetica, Ubuntu, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;, &quot;Segoe UI Symbol&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">

Package.json setup Communication-react / App | Installed in node_modules/? | Installed in node_modules/communication-react/node_modules/ | Remarks
-- | -- | -- | --
dependency/<none> | ✅ | ❌ | Current intended behavior
dependency/matching-dependency | ✅ | ❌ | Current intended behavior
dependency/non-matching-dependency | ✅ | ✅ | ❌ The two versions are mismatched. Likely to cause runtime errors
peerDependency/<none> | ❌ | ❌ | ❌ Breaks the build
peerDependency/matching-dependency | ✅ | ❌ | Happy path!
peerDependency/non-matching-dependency | ✅ | ❌ | ❌ Warning about wrong version, but easy to miss in the console log

<p style="box-sizing: border-box; margin: 0in 0in 8pt; line-height: 15.6933px; font-size: 11pt; font-family: Calibri, sans-serif;"><span style="box-sizing: border-box; font-size: 8pt;"> </span></p></div>
</body>
</html>

# Is this a breaking change?

No. [This official SemVer FAQ](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api) states that patch updates of packages are allowed to require clients to update transitive dependencies.

We shouldn't do this for no reason, to avoid headaches in CI etc, but there is a legitimate reason for this change -- currently it is very easy for clients to get into non-functioning states if the `@azure/communication-calling` dependency happens to mismatch between their app and our library.


packages/communication-react/package.json
```diff
@@ -45,6 +43,8 @@
     "@fluentui/react-hooks": "~8.3.8"
   },
   "peerDependencies": {
+    "@azure/communication-calling": "1.3.2",
```
JamesBurnside, line 46: if we do this, presumably when someone does an npm install they also have to do `npm i @azure/communication-calling  @azure/communication-chat`? If so we'd need to update documentation along side this

Given we limit to a single version it's odd to have a peer dependency of a fixed version (although I do see the need). React hooks can detect if there's multiple versions of react and throws a big error - could that be a solution instead?

JamesBurnside:
Love the table in the PR description - super useful!

packages/communication-react/package.json
prprabhu-ms: If only we didn't ask our clients to use `--legacy-peer-deps` :( (I'm looking at your `react-northstar`).

Without this flag, the default [NPM behavior since v7 is to auto-install peer-dependencies](https://stackoverflow.com/questions/66239691) so that they wouldn't necessarily have to install our peer dependencies explicitly.

That said, peer dependencies are the way to go here:

* We do plan to move to a minimum-supported-version for these dependencies. At least allow patch version updates for chat/calling without needing to release a new communication-react stable version.
* Detecting multiple versions of a package and complaining -- if this is the path we take, it'd have to be the package that dislikes multiple versions (i.e., calling) that does that. Also, it feels wrong to solve this problem at runtime in the package, when NPM dependency management is already intended to solve it.

Perhaps `@azure/communication-calling` _should_ detect multiple versions and complain, especially if they decide it's a BAD thing to happen. I'd say that it should be in addition to proper dependency management though.

JamesBurnside: Can we have these allow minor versions today? 
```suggestion
    "@azure/communication-calling": ">=1.3.0 <1.4.0",
```

JamesBurnside (approved):

packages/communication-react/package.json
JamesBurnside: (I did file an issue with n* to support v17+ https://github.com/microsoft/fluentui/issues/21082)

JamesBurnside: (perhaps we could even hoist or bundle it out of the fluent repo: https://github.com/microsoft/fluentui/tree/master/packages/fluentui/react-northstar/src/components/Chat)

prprabhu-ms: Replying to the suggestion for this PR: I don't want to change the actual version supported in any way in this PR (just move the single pinned version to peer deps).

We can start supporting minor versions in the future, but let's keep that a separate PR / discussion. The main concern there is knowing what our support plan is -- folks might start using latest calling before we've integrated it / tested with our own stable release.

alkwa-msft (approved):

PR #1297, opened 2022-01-11
Fixed alignment of typing indicator in chat composite by changing flexFlow to column-reverse

prabhjot-msft:
# What
Fixed alignment of typing indicator in chat composite by changing flexFlow to column-reverse

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2662655

# How Tested
Manually

# Process & policy checklist


dmceachernmsft (approved):

prprabhu-ms (approved):

packages/react-components/src/components/styles/TypingIndicator.styles.ts
```diff
@@ -7,7 +7,7 @@ import { mergeStyles } from '@fluentui/react';
  * @private
  */
 export const typingIndicatorContainerStyle = mergeStyles({
-  minHeight: '2.125rem'
+  minHeight: '0.125rem'
```
JamesBurnside, line 10: The min-height is actually there to prevent the messages from shifting up and down when new messages appear (this should have a comment to make that clear).

e.g. when a typing indicator appears and dissapears is causes the messages to shift up and down:
image

Is it possible to keep the minHeight and instead align the text (vertically) to the bottom or center of the container so it is not so far away from the edit box?

JamesBurnside (changes requested):
Want to check if we can keep the min height or not

prabhjot-msft:
> Want to check if we can keep the min height or not

We can keep the min height and align the text to bottom by changing flexFlow to column-reverse, let me know if that works.

packages/react-components/src/components/styles/TypingIndicator.styles.ts
```diff
@@ -7,7 +7,9 @@ import { mergeStyles } from '@fluentui/react';
  * @private
  */
 export const typingIndicatorContainerStyle = mergeStyles({
-  minHeight: '2.125rem'
+  minHeight: '2.125rem',
+  // flexFlow set to column-reverse to align the text to the bottom of the container
+  flexFlow: 'column-reverse'
```
JamesBurnside, line 12: because flex boxes have a wide variety of css properties that are hard to dissociate from each other (and because when changing css its incredibly hard to ensure no other properties are affected) we want to avoid writing css as much as possible.

Instead we want to use a semantic driven approach and use Fluent's [Stack](https://developer.microsoft.com/en-us/fluentui#/controls/web/stack)
These have properties like `horiztonal` and `align=end` to avoid having to write css and have a semantical-driven approach to how components are laid out that very readable&understandable.

Instead can you do something like:
image
? (not sure if the alignment should be 'baseline' or 'end')

If not keep this as is!

JamesBurnside (approved):
Just a suggestion on avoid css, otherwise this is great!

packages/react-components/src/components/styles/TypingIndicator.styles.ts
prabhjot-msft: flex layout is taking precedence over the properties thus vertical align is not working in this case. The other property that we can change is justifyContent='end' which is again property to flex layout. So, I am going to use column-reverse in this case.

PR #1300, opened 2022-01-13
Bug 2641702: [Web][a11y][Call page] - first tab lands on horizontal gallery right arrow instead of "Unmute" button

carocao-msft:
# What
Changed tab index and assigned auto focus on microphone button
Tab order should be: control panel button from left to right, then next button

# Why
Fixed accessibility issue  "first tab lands on horizontal gallery right arrow instead of "Unmute" button"

# How Tested
1.First pull the code
2. then force the next button to show up in the calling screen using the following method:
first remove the 'showButtons' condition in Horizontal Gallery
Screen Shot 2022-01-12 at 8 12 57 PM

then remove the 'horizontalGalleryPresent' condition in VideoGallery
Screen Shot 2022-01-12 at 8 14 12 PM
3. then test tabbing orders


packages/react-components/src/components/HorizontalGallery.tsx
```diff
@@ -107,6 +107,7 @@ const HorizontalGalleryNavigationButton = (props: {
   const theme = useTheme();
   return (
     <DefaultButton
+      tabIndex={2}
```
JamesBurnside, line 110: `tabindex` is tricky as this is a global property and not something that is hierarchical. So if we set tabIndex to one, if Contoso (note: Contoso is our name for another company using our library) embeds our composite into their web page we steal the tabindex order that they may have set e.g.:
```
<body>
<button tabindex={1} />
<button tabindex={2} />
<button tabindex={3} />
<CallComposite />
<button tabindex={4} />
```
The tab index in that order won't work as Contoso expects.

Instead we added a `autofocus` property to the chat composite to allow contoso to say that the composite should take focus when this component is rendered: https://github.com/Azure/communication-ui-library/pull/1235/files#diff-ec0dc325bdf92a107d64cbdd0904c5aae39b00aac3236fe2bfe0b88ee332471b

With the call composite its more complicated as each page of the composite may want to have a different component take auto focus when loaded. This would be a good thing to bring up for discussion on the Web UI teams channel or in standup

JamesBurnside (changes requested):
Unfortunately we are unable to make use of `tabIndex` in case we interfere with what a developer would want that tabIndex to be on their web page.

packages/react-components/src/components/HorizontalGallery.tsx
carocao-msft: Removed tab index and leaving autofocus on Microphone

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -144,6 +144,7 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
 
   const microphoneButton = options?.microphoneButton !== false && (
     <MicrophoneButton
+      autoFocus
```
prprabhu-ms, line 147: Might be worth adding a comment here that this ensures that focus / first tab in is on the microphone button after the `CallComposite` loads.

In general, a nudge to understand that this attribute determines focus for the entire composite.

prprabhu-ms (approved):

prprabhu-ms:
You can test this without code changes by joining the same call many times and then turning on video in one of the sessions. :)

You can also use the Browser DevTools' device emulation to get a narrow form-factor to get the horizontal gallery soonner.

carocao-msft:
> You can test this without code changes by joining the same call many times and then turning on video in one of the sessions. :)
> 
> You can also use the Browser DevTools' device emulation to get a narrow form-factor to get the horizontal gallery soonner.

Yeah it's just making some code changes makes testing more convenient, just need to remember don't commit it  ;)

PR #1302, opened 2022-01-13
[POC] Mock Calling Adapter state for UI test

mgamis-msft:
# What

This is a very rough POC for Mocking Calling Adapter state and would like some input on how to proceed.

Challenges I am facing:
- No handlers in mocked Calling Adapter state so can't test clicking 'Start Call' button in configuration page. Can set Call Adapter page state to 'call' to get UI of call
- Can't modify state after test is initialized. I can only get a snapshot of the composite in a particular state but I cannot modify the state. There are no mechanisms in playwright that I know of to call React functions.
- Cannot import CallAdapterState type into tests module so I cannot type guard the state in the test

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2683785
In order to test UI for scenarios difficult to set up.

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    const newTestGuid = generateGUID();
+    const user = users[0];
+    user.groupId = newTestGuid;
+
+    // Load different locale for locale tests
+    const page = pages[0];
+    await page.goto(
+      buildCallingUrl(
+        serverUrl,
+        {
```
mgamis-msft, line 22: There would be a setup step for this but I need to match this type to CallAdapterState. But I cannot import it in tests. Any ideas?

packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -1,58 +1,44 @@
...
-const groupId = verifyParamExists(params.groupId, 'groupId');
-const userId = verifyParamExists(params.userId, 'userId');
+const state = JSON.parse(params.state) as CallAdapterState;
 
 // Optional params
 const useFrLocale = Boolean(params.useFrLocale);
 const showCallDescription = Boolean(params.showCallDescription);
 const injectParticipantMenuItems = Boolean(params.injectParticipantMenuItems);
 
-function App(): JSX.Element {
```
mgamis-msft, line 37: There would be another app such that the other snapshot tests are not affected. This is rough POC.

PR #1303, opened 2022-01-13
Add variant/flavor commands to samples

PorterNan:
- add variant/flavor commands to samples
- cleanup unused global script command

# What

# Why

# How Tested

# Process & policy checklist


prprabhu-ms (approved):

JamesBurnside (approved):

alkwa-msft:
So we have even more complex rush commands in our samples.

What is the flow when we have to port functionality into the hero sample repos? Are we making our developers use rush as well? Do they have to know what is the difference between beta and stable?

My goal for the hero samples is that they should snap to stable and developers should only need to  git clone, npm install, npm run start. Are we still able to do that with this change?

PorterNan:
> So we have even more complex rush commands in our samples.
> 
> What is the flow when we have to port functionality into the hero sample repos? Are we making our developers use rush as well? Do they have to know what is the difference between beta and stable?
> 
> My goal for the hero samples is that they should snap to stable and developers should only need to git clone, npm install, npm run start. Are we still able to do that with this change?

Thanks for this question, current command(without my change) is highly depending on rush and configs outside of our repo already, so my expectation is the new sample repo only shares code (everything under src folder), but not configs like package.json(we will have separated configs for the sample repo) - otherwise we can't put any beta code here in samples

prprabhu-ms:
Yeah I don't expect us to be able to copy-paste package.json out, especially the `scripts` section directly hooks into `rush` and other tooling in the monorepo.

PR #1305, opened 2022-01-14
BUG 2602657 : Compliance banner: Added unit test and simplified logic

carocao-msft:
# What
Moved computeVariant function to utils
Added unit test for computeVariant
Simplified logic for computeVariant, using 3 states: on, off and stopped to determine returned string
Screen Shot 2022-01-17 at 8 57 22 PM
Stopped state means:  Previous state is on, Current state is off 
On state means: Current state is on, regardless of what previous state is
Off state means: previous state is off and current state is still off. Off state is the default state 

# Why
The mobile team is current using the on/off/stop logic, it is more simplied than our logic (using previous and current values), also it is easier for development when mobile and desktop are using the same logic 
Mobile logic for reference: https://github.com/Azure/communication-ui-library-ios/blob/main/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewComponent/Banner/BannerViewModel.swift

# How to test
Start a teams meeting, join teams meeting using Calling sample. start and stop recording / transcribing in teams. Check if the calling sample is displaying correct messages.


packages/react-composites/src/composites/CallComposite/utils/Utils.test.ts
```diff
@@ -35,3 +35,39 @@ describe('SDKUtils tests', () => {
     });
   });
 });
+
+describe('Record and Transcription tests', () => {
+  describe('computeVariant tests', () => {
+    test('computeVariant should return expected values', () => {
```
alkwa-msft, line 41: can we have more granular unit tests. Don't just bunch up everything into one big test.

alkwa-msft: instead of having comments for each "test", we should just use separate test blocks where we can just make the comment into the string of what the test is doing.

alkwa-msft (commented):
Really great start. Awesome to see unit tests in a good unit testable area

I think we want to break apart the tests into actual tests and not just one big test.

packages/react-composites/src/composites/CallComposite/utils/Utils.test.ts
anjulgarg: @carocao-msft For example
`computeVariant should start recording`
`computeVariant should stop recording`
et etc. :)

packages/react-composites/src/composites/CallComposite/components/ComplianceBanner.tsx
```diff
@@ -56,10 +57,10 @@ export const ComplianceBanner = (props: ComplianceBannerProps): JSX.Element => {
   return <DismissableMessageBar variant={variant} strings={strings} />;
 };
 
-function DismissableMessageBar(props: { variant: number; strings: CompositeStrings }): JSX.Element {
+function DismissableMessageBar(props: { variant: string; strings: CompositeStrings }): JSX.Element {
```
JamesBurnside, line 60: can we strongly type the variant everywhere instead of `string`?
e.g.
```ts
type ComplianceBannerVariant =
  | 'NO_STATE'
  | 'TRANSCRIPTION_STOPPED_STILL_RECORDING'
  | 'RECORDING_STOPPED_STILL_TRANSCRIBING' ...etc.
```

packages/react-composites/src/composites/CallComposite/utils/Utils.test.ts
carocao-msft: Updated unit test :)

packages/react-composites/src/composites/CallComposite/components/ComplianceBanner.tsx
carocao-msft: Ok np!

packages/react-composites/src/composites/CallComposite/components/ComplianceBanner.tsx
```diff
@@ -12,197 +13,169 @@ export type ComplianceBannerProps = {
...
+    callRecordState: string;
+  };
 };
 
 /**
  * @private
  */
 export const ComplianceBanner = (props: ComplianceBannerProps): JSX.Element => {
-  const strings = useLocale().strings;
+  function determineStates(previous: string, current: boolean | undefined) {
```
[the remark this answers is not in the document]
JamesBurnside: And here as well:
image

```ts
type complianceState = 'on' | 'stopped';
...
    callTranscribeState: complianceState;
    callRecordState: complianceState;
```

JamesBurnside (approved):
Just suggestions on strongly typing variables-  otherwise LGTM!

packages/react-composites/src/composites/CallComposite/components/ComplianceBanner.tsx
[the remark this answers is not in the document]
carocao-msft: Uodated :)

dmceachernmsft (approved):

PR #1308, opened 2022-01-15
Add stable build to PR validation pipeline

PorterNan:
- add stable build as a variant of matrix 
- add snapshot update for stable build
- Remove auto rebuild when people run `rush switch-flavor` otherwise it is causing double build for PR

So switch flavor process now becomes(will add it to readme when everything is settled down):
```
rush switch-flavor:stable
rush rebuild

....(when you want to do install or update)
rush update:stable / rush install:stable
```

# What

# Why

# How Tested

# Process & policy checklist


.github/workflows/ci.yml
```diff
@@ -45,6 +54,7 @@ jobs:
         run: rush build -t @azure/communication-react
       # Verify no uncommitted api extractor changes
       - name: API Extractor Check
+        if: ${{ matrix.flavor != 'stable' }}
```
prprabhu-ms, line 57: We should separately generate API files for the two flavors of builds (to two different folders), and then check there are no uncommitted changes in both cases.

Please add a TODO here if you intend to do that once you've worked on generating the API review files separately.

.github/workflows/update-snapshot.yml
```diff
@@ -8,6 +8,11 @@ on:
 
 jobs:
   update_examples_snapshot:
+    strategy:
```
prprabhu-ms, line 11: Have we already separated the committed snapshots for the two flavors?

common/config/rush/command-line.json
```diff
@@ -37,7 +37,7 @@
       "summary": "Switch flavor to beta",
       "description": "Every build and test command will run based on beta version.",
       "autoinstallerName": "rush-cpy",
-      "shellCommand": "rush install && cpy common/config/env/beta.env common/config/env --rename .env && rush rebuild",
+      "shellCommand": "rush install && cpy common/config/env/beta.env common/config/env --rename .env",
```
prprabhu-ms, line 40: I see why you removed `&& rush rebuild` here -- so that CI won't do an extra rebuild when switching flavors.

But that step was necessary to ensure that developers don't accidentally use previously built packlets after switch flavors, right?

Perhaps add a rush command that deletes all the `.rush/temp` folders and run that so that next build will necessary build everything from scratch? See https://rushjs.io/pages/commands/rush_build/

.github/workflows/ci.yml
JamesBurnside: This could be a seperate PR but +1 to this - it would be very useful when introducing a beta feature to see the API change _only_ in the beta file and no nothing crept into the stable API

.github/workflows/update-snapshot.yml
PorterNan: Yes! This is already done ;D

common/config/rush/command-line.json
PorterNan: Done! This is actually a good idea! So we can still use rush build, but build everything in fresh

.github/workflows/update-snapshot.yml
```diff
@@ -240,9 +240,22 @@ jobs:
...
+          CONNECTION_STRING: ${{ secrets.CONNECTION_STRING }}
+      - name: Update calling snapshot
+        id: update-calling-snapshots
+        run: rushx test:e2e:call:update
+        working-directory: ./packages/react-composites
+        env:
+          CONNECTION_STRING: ${{ secrets.CONNECTION_STRING }}
+      - name: Update meeting snapshot
+        id: update-meeting-snapshots
+        if: ${{ matrix.flavor != 'stable' }}
```
prprabhu-ms, line 257: Why the special handling? Because there are 0 tests on stable, and there is no (easy) way for jest to accept an empty test suite?

Can you add a comment here that this check should be removed when MeetingComposite goes stable. We're likely to forget to update this...

common/config/rush/command-line.json
```diff
@@ -65,6 +65,14 @@
       "shellCommand": "rush install --variant stable",
       "safeForSimultaneousRushProcesses": true
     },
+    {
+      "commandKind": "global",
+      "name": "build:all",
```
prprabhu-ms, line 70: I'd suggest we name this `build:all-flavors`. Currently it is confusing if this means build all packages? (similar to make file `make all`)

packages/chat-stateful-client/package.json
```diff
@@ -12,7 +12,7 @@
     "build:watch": "rushx build",
     "clean": "rimraf dist && rimraf preprocessed",
     "preprocess": "cpx \"./src/**\" ./preprocessed && babel ./src --out-dir ../preprocessed --extensions \".ts,.tsx\" --keep-file-extension --config-file ./.babelrc.js --relative && rimraf ../preprocessed",
-    "build:by-flavor": "rushx clean && npm run build:esm && rushx build:cjs && api-extractor run --local",
+    "build:by-flavor": "rushx clean && npm run build:esm && rushx build:cjs && npm run api-extractor:by-flavor",
```
prprabhu-ms, line 15: I think this should be `_build:by-flavor` and `_api-extractor:by-flavor`, because we don't expect developers to ever run these commands directly in their workflow.

prprabhu-ms (approved):

.github/workflows/update-snapshot.yml
PorterNan: Because when we are doing conditional compiling, we don't do it for *.test.ts, so in the test.ts we need to read environment variables to determine whether it is under stable flavor, we can change something there if we want (but we don't even want to spin up the test server, so I guess skip the test for stable would be the best choice)

prprabhu-ms: I think updating the test file is a better choice, if it is possible. Having an ad-hoc condition like this is dangerous because we'll likely forget about it when the feature goes stable (and... never test it on stable)

PR #1309, opened 2022-01-15
Updated MessageThreads to remove usernames being displayed on top of messages in Storybook - UI Component - Message Thread

edwardlee-msft:
# What
Updated MessageThreads to remove usernames being displayed on top of messages in Storybook - UI Component - Message Thread

# Why
Usernames were being displayed on top of messages causing persona icons to be displayed on top rather than on the left side of the message. This was also causing duplicate username to be populated, one at top and one in the message itself.
https://skype.visualstudio.com/SPOOL/_workitems/edit/2667113/

# How Tested
Testing on local storybook and compared results with chat samples. 

# Process & policy checklist


dmceachernmsft (approved):

anjulgarg:
Check out the new test screenshots.
The avatars are missing user initials completely 
`packages/react-composites/tests/browser/chat/ChatComposite.test.ts-snapshots/received-messages-Desktop-Chrome-linux.png`

anjulgarg (changes requested):
Check out the new test screenshots.
The avatars are missing user initials completely
packages/react-composites/tests/browser/chat/ChatComposite.test.ts-snapshots/received-messages-Desktop-Chrome-linux.png

prprabhu-ms:
> Check out the new test screenshots. The avatars are missing user initials completely packages/react-composites/tests/browser/chat/ChatComposite.test.ts-snapshots/received-messages-Desktop-Chrome-linux.png

Huh. Why though? I'm stumped. ChatComposite overrides the default personal anyway...

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -322,8 +322,7 @@ const memoizeAllMessages = memoizeFnAll(
 
         const personaOptions: IPersona = {
           hidePersonalDetails: true,
-          size: PersonaSize.size32,
-          text: message.senderDisplayName
```
dmceachernmsft, line 326: Could the removal of message.senderDisplayName be responsible for the initials disappearing in the icons?

edwardlee-msft: Did some testing, and yes, removing message.senderDisplayName cause the initials to disappear.

prprabhu-ms: Can we explore using some of the props related to overriding initials:

image
image

See https://developer.microsoft.com/en-us/fluentui#/controls/web/persona

PorterNan:
image
Looks like initials/default avatars are not working according to the new snapshot

packages/react-components/src/components/MessageThread.tsx
PorterNan: Yep, @prprabhu-ms, use imageInitials instead (you might have to do some kinda string processing and I guess there are a lot of edge case to deal with, wondering if fluent export something like generating initials for us)

PorterNan: Another nice way to do it is just adding hidePersonaDetails, with text still there, which will hide the name for you, check my codepen for an example: 
image

The green one's name is hidden now:
image

edwardlee-msft: Good suggestion, hidePersonaDetails is currently already set to true but we the name are still populated on top:
image

As for using imageInitials, I have tested it out but it does require string processing to make the initials (otherwise it will use the full name inside the icon). I will continue by looking into Fluent to see if there is a way for them to generate initials for us.

edwardlee-msft: I was able to find a library in fluentui for creating initials. Now working as intended.
Tested in storybook:
Screen Shot 2022-01-21 at 10 49 20 AM


Tested in chat:
Screen Shot 2022-01-21 at 10 49 07 AM

PR #1310, opened 2022-01-15
Added entry for endpointUrl in appsettings.json instead of using connectionString twice

prabhjot-msft:
# What
Added entry for endpointUrl in appsettings.json instead of using connectionString twice

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2691366

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/storybook/stories/NonReact/NonReactComposites.stories.mdx
+++ b/packages/storybook/stories/NonReact/NonReactComposites.stories.mdx
@@ -125,10 +125,12 @@ After finishing building the bundle, you can also try the sample
 
 1. Replace the `ResourceConnectionString` in the `samples/Server/appsettings.json` file with the connection string from your Azure Communication Services resource, which is listed in the Azure Portal under _Keys_.
 
-2. Run this command to start the server:
+2. Replace the `EndpointUrl` in the `samples/Server/appsettings.json` file with the connection string and remove "endpoint=" from the beginning of the string.
+
+3. Run this command to start the server:
 
 ```bash
 rushx start
 ```
 
-3. Browse to http://localhost:3000
+4. Browse to http://localhost:3000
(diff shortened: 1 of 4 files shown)


packages/storybook/stories/NonReact/NonReactComposites.stories.mdx
```diff
@@ -125,10 +125,12 @@ After finishing building the bundle, you can also try the sample
 
 1. Replace the `ResourceConnectionString` in the `samples/Server/appsettings.json` file with the connection string from your Azure Communication Services resource, which is listed in the Azure Portal under _Keys_.
 
-2. Run this command to start the server:
+2. Replace the `EndpointUrl` in the `samples/Server/appsettings.json` file with the connection string and remove "endpoint=" from the beginning of the string.
```
prprabhu-ms, line 128: This isn't quite right. They should get the endpoint URL directly from the ACS resource.

The hack before this PR was parsing the connection string as a URL and then using just the $HOST part of it, which ignored other information in the connection string.

anjulgarg (approved):

packages/storybook/stories/NonReact/NonReactComposites.stories.mdx
anjulgarg: I don't quite understand what you mean. LGTM

JamesBurnside: Users should paste the endpoint directly from the portal instead of extracting it out of their connection string:
image
(^red box)

prabhjot-msft: Update the doc to get value from ACS Resource.

JamesBurnside (approved):

prprabhu-ms (approved):

PorterNan:
Sry, I am kinda late to the conversation, I remember we were discussing we want to use endpoint instead of endpointUrl, any chance we could make that change?

PR #1312, opened 2022-01-17
Participant button menu items height increased in mobile view

edwardlee-msft:
# What
Added mobile styling for participant button menu items in ParticipantItem.styles.ts

# Why
In mobile view, participant button menu items were too small for tapping.
https://skype.visualstudio.com/SPOOL/_workitems/edit/2668807/

# How Tested
Performed live tests by running changes on local machine and viewing/comparing on mobile and desktop view. 

# Process & policy checklist
- [x ] I have updated the project documentation to reflect my changes if necessary.
- [x ] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.


packages/react-components/src/components/styles/ParticipantItem.styles.ts
```diff
@@ -3,6 +3,8 @@
 
 import { mergeStyles } from '@fluentui/react';
 
+const MINIMUM_TOUCH_TARGET_HEIGHT_REM = 3;
```
prprabhu-ms, line 6: Can we extract this constant (and the one in ChatMessageComponent.styles.ts) to a common file to be used from all styles files in `react-components`?

packages/react-components/src/components/styles/ParticipantItem.styles.ts
```diff
@@ -13,7 +15,12 @@ export const participantItemContainerStyle = mergeStyles({
   maxWidth: '20rem',
   minWidth: '12rem',
   cursor: 'pointer',
-  alignItems: 'center'
+  alignItems: 'center',
+  // Mobile height for participant items increased from default height
+  '@media screen and (max-width: 640px)': {
```
prprabhu-ms, line 20: Components don't make these decisions internally. Instead, they expose a prop that can be used to indicate that we want larger touch targets. The composite decides when to set that prop. This ensures that (a) we don't have the `@media` queries spread out over the code-base, and composites can set this consistently, (b) Contoso when using the components directly can make these decision themselves.

See `ChatMessageActionFlyout` for an example.

edwardlee-msft:
Closing Pull request to further optimize implementation.

PR #1314, opened 2022-01-17
Add API for custom button injection into `CallComposite`

prprabhu-ms:
# What

* Add the API ( PoC in #953 ) for adding custom buttons to `CallComposite` control bar.
* Add an e2e test for the new API.

The new API is added as `@beta` and conditionally compiled away from stable releases.

# How Tested

New e2e test.

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -86,8 +97,86 @@ export type CallControlOptions = {
...
+   * @beta
+   */
+  onFetchCustomButtonProps?: CustomCallControlButtonCallback[];
 };
 
+/* @conditional-compile-remove-from(stable): custom button injection */
+/**
+ * Placement for a custom button injected in the {@link CallControls}.
+ *
+ * 'first': Place the button on the left end (right end in rtl mode).
```
JamesBurnside, line 113: If buttons like the options button is moved to the overflow menu, will this simply add the custom below the options button?

packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -161,4 +177,43 @@ function onFetchParticipantMenuItems(): IContextualMenuItem[] {
...
+  (args: CustomCallControlButtonCallbackArgs): CustomCallControlButtonProps => {
+    return {
+      showLabel: args.displayType !== 'compact',
+      // Some non-default icon that is already registered by the composites.
+      onRenderOffIcon: () => <Icon iconName="MessageSeen" />,
+      onRenderOnIcon: () => <Icon iconName="MessageSeen" />,
+      strings: {
+        label: 'custom #1'
+      },
+      placement: 'first'
```
JamesBurnside, line 190: in mobile appears to not be there:
image

packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -161,4 +177,43 @@ function onFetchParticipantMenuItems(): IContextualMenuItem[] {
   ];
 }
 
+const onFetchCustomButtonProps: CustomCallControlButtonCallback[] = [
+  (args: CustomCallControlButtonCallbackArgs): CustomCallControlButtonProps => {
+    return {
+      showLabel: args.displayType !== 'compact',
+      // Some non-default icon that is already registered by the composites.
+      onRenderOffIcon: () => <Icon iconName="MessageSeen" />,
```
JamesBurnside, line 185: missing icons for some reason
image

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -86,8 +97,86 @@ export type CallControlOptions = {
...
+ * It is an error to place the button in reference to another button that has
+ * been hidden via a {@link CallControlOptions} field.
+ *
+ * Multiple buttons placed in the same position are appended in order.
+ * E.g., if two buttons are placed 'first', they'll both appear on the left end (right end in rtl mode)
+ * in the order provided.
+ *
+ * @beta
+ */
+export type ControlBarButtonPlacement =
```
JamesBurnside, line 126: Can this be `CallCompositeControlBarButtonPlacement`? (or something like that but less verbose)
My thinking is the MeetingComposite will want the same/similar API but will have more buttons like chat and people buttons

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -86,8 +97,86 @@ export type CallControlOptions = {
...
+ * 'afterCameraButton': Place the button on the right (left in rtl mode) of the camera button.
+ * ... and so on.
+ *
+ * It is an error to place the button in reference to another button that has
+ * been hidden via a {@link CallControlOptions} field.
+ *
+ * Multiple buttons placed in the same position are appended in order.
+ * E.g., if two buttons are placed 'first', they'll both appear on the left end (right end in rtl mode)
+ * in the order provided.
+ *
```
JamesBurnside, line 123: should we `@remarks` what happens when when a button is disabled? e.g. if I remove participants button I assume it will be placed where the participants button _was_

packages/react-composites/tests/browser/call/app/index.tsx
prprabhu-ms: That snapshot is incorrect. I'm fixing the build so we get the right icon.

packages/react-composites/tests/browser/call/app/index.tsx
prprabhu-ms: Automation hasn't updated the snapshots yet.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prprabhu-ms: At least the way I was thinking about it, this is just for the top-level button.

Good point about API implications.

We could either:
1. Reject the placement value (i.e. deprecation, but not backward compatible)
2. Move it below/above in the overflow menu

As you suggested, (2) is the API compatible change, although it'll still be a breaking change from a UX perspective. (1) might actually be better in that case -- it's better to require developer action than surprisingly move their button around.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prprabhu-ms: Done. `CustomCallControlButtonPlacement` matches other naming conventions introdued in this PR.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prprabhu-ms: By removed, I guess you mean hidden via `CallControlOptions`? The remark above says that you cannot place it relative to a button that was hidden. So placement is always in relation to another button that is visible.

I think talking about "placed instead of hidden button" will add confusion because it relates the custom button placement to a button that was hidden.

packages/react-composites/tests/browser/call/app/index.tsx
prprabhu-ms: image

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
JamesBurnside: Ah yep that comment resolves this, missed it thanks!

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
JamesBurnside: I think for allowing contoso to add new custom buttons we have to have an overflow menu, currently we have the max number of buttons we can fit on mobile so they would have to remove one to add a custom button to make use of this

prprabhu-ms: There is work planned for Contoso to be able to *explicitly* add buttons into the overflow menu. That will be a separate API. Contoso will be able to add a section to the overflow menu and add anything they want in there.

But we won't have any smarts about moving the button automagically between the two. That seems like surprising magic that is likely to create more confusion / trouble for the clients than be helpful.

Forking this discussion to team channel for more eyes.

JamesBurnside (approved):
existing comment not blocking for this PR

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prprabhu-ms: Resolving this based on discussion.

This API makes sense right now. Additionally we'll add API to add explicitly to the overflow menu.

alkwa-msft (approved):

PR #1322, opened 2022-01-18
Participant button Item mobile view fixed by removing use of mergeStyleSets in CallControl.tsx

edwardlee-msft:
# What
Participant button items in mobile view increases in height for tap interface. 
Removed use of mergeStyleSets in CallControl.tsx as the styles were flattening when use mergeStyleSets. Confirmed in below picture below:

With mergeStyleSets
with_mergeStyleSets 

vs without mergeStyleSets

without_mergeStyleSets

# Why
Participant button items did not increase in height when viewed on mobile
https://skype.visualstudio.com/SPOOL/_workitems/edit/2668807/

# How Tested
Tested on chrome dev tools mobile and desktop view. Also tested on personal phone.
Confirmed size changes in dev tools and on mobile phone.

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -125,12 +125,18 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
   const hangUpButtonProps = usePropsFor(EndCallButton);
 
   const participantsButtonStyles = useMemo(
-    () => mergeButtonBaseStyles(props.increaseFlyoutItemSize ? participantButtonWithIncreasedTouchTargets : {}),
+    () =>
+      props.increaseFlyoutItemSize
+        ? { ...participantButtonWithIncreasedTouchTargets, ...controlButtonBaseStyle }
```
JamesBurnside, line 130: this won't do a deep merge, so if `participantButtonWithIncreasedTouchTargets` has 
```
{
  propertyY: {
    propertyA: 1
}
```
and `controlButtonBaseStyle` has:
```
{
  propertyY: {
    propertyB: 2
}
```
doing `{ ...participantButtonWithIncreasedTouchTargets, ...controlButtonBaseStyle }` would result in:
```
{
  propertyY: {
    propertyB: 2
}
```
instead of:
```
{
  propertyY: {
    propertyA: 1,
    propertyB: 2
}
```

JamesBurnside: Fluent exposes a `merge` function, perhaps that would work instead?

change/@internal-react-composites-ae9ef7c5-4a19-4aec-8f5d-f146b82fc48c.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "patch",
+  "comment": {
```
JamesBurnside, line 3: no comment has been added in this change file, normally they are of the format:
image

(not sure how yours got title and value fields)

Can you update this? The reason for the change files is we use them to auto generate the changelog for releases

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prprabhu-ms: In this case, we define both `participantButtonWithIncreasedTouchTargets` and `controlButtonBaseStyle`. So if `merge` doesn't work, we could just define two different styles for the two cases (instead of trying to merge).

Actually, what we want is probably [`concatStyleSets`](https://docs.microsoft.com/en-us/javascript/api/merge-styles?view=office-ui-fabric-react-latest#@fluentui-merge-styles-concatstylesets-function(2))

prprabhu-ms: Key is to avoid registering CSS classes here.

change/@internal-react-composites-ae9ef7c5-4a19-4aec-8f5d-f146b82fc48c.json
edwardlee-msft: Fixed! The comment did not stick when I did rush changelog. Will make sure to double check for next time. Thank you

change/@internal-react-composites-ae9ef7c5-4a19-4aec-8f5d-f146b82fc48c.json
```diff
@@ -1,9 +1,6 @@
 {
   "type": "patch",
-  "comment": {
-    "title": "",
-    "value": ""
-  },
+  "comment": "Participant button Item mobile view fix for touch",
```
prprabhu-ms, line 3: This could be worded better: "Increase size of participant flyout menu items for mobile view"

Think of how it will sound in a ChangeLog file.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -231,3 +225,6 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
     </Stack>
   );
 };
+
+const concatButtonBaseStyles = (styles: IButtonStyles): IButtonStyles =>
+  concatStyleSets(controlButtonBaseStyle, styles);
```
prprabhu-ms, line 230: Woohoo!

prprabhu-ms (approved):

change/@internal-react-composites-ae9ef7c5-4a19-4aec-8f5d-f146b82fc48c.json
edwardlee-msft: Makes absolute sense! Changed comment to better reflect pr in ChangeLog file.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
PorterNan: So the merge will not override properties from the first object using the second style sets? I am a little bit wondering what the logic of style merging is

prprabhu-ms (approved):
New snapshots look good

PorterNan (approved):
Approved, this is fixing the bug perfectly

I am a little bit concerning of the logic of mergeStyleSets, wish I could get some answers here: wouldn't that override the first style object style set using the second style set? Or what is the true logic of mergeStyleSets - seems like we are using it not just here (and we might be using it in the future), figuring out the right logic would be nice

PR #1323, opened 2022-01-19
Removed empty test script in Chat

edwardlee-msft:
# What
Removed empty test script in Chat

# Why
[Sample/Chat] Remove empty test script
https://skype.visualstudio.com/SPOOL/_workitems/edit/2691384

# Process & policy checklist


PorterNan (changes requested):
This is intentional, because it is a command required by the global command `rush test` (see command-line.json) which means, each project should have `npm run test` command inside, you can make it like 
"test": "echo skip"
or just simply ignore this empty script

edwardlee-msft (commented):
I re-added "test" with "echo skip". 
Thank you for catching that!

prprabhu-ms (approved):

PorterNan (approved):
Thanks for making this adjustoment!

PR #1328, opened 2022-01-19
Dmceachernmsft/meetingParticipantNamesAvatarsSidePane

dmceachernmsft:
# What
changed people pane in meeting composite to use the participantContainer component.

# Why
Allows for Avatar Customization in the people pane in the side bar. 

https://skype.visualstudio.com/SPOOL/_workitems/edit/2694576
# How Tested
Ran locally in meeting sample to see if people pane changed avatar information.


packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -1,25 +1,36 @@
...
 // Licensed under the MIT license.
 import React from 'react';
-import { participantListStack, participantListStyle, participantListWrapper, listHeader } from './styles/Chat.styles';
-import { OnRenderAvatarCallback, ParticipantList, ParticipantMenuItemsCallback } from '@internal/react-components';
+import {
+  participantListStack,
+  participantListStyle,
+  participantListWrapper,
+  listHeader
+} from '../ChatComposite/styles/Chat.styles';
```
prprabhu-ms, line 9: These styles should also move to `common`. Stuff from `common` shouldn't import from `ChatComposite`.

packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -1,25 +1,36 @@
...
+  OnRenderAvatarCallback,
+  ParticipantList,
+  ParticipantListProps,
+  ParticipantMenuItemsCallback
+} from '@internal/react-components';
 import { FocusZone, Stack } from '@fluentui/react';
 import { useLocale } from '../localization';
-import { usePropsFor } from './hooks/usePropsFor';
-import { AvatarPersona, AvatarPersonaDataCallback } from '../common/AvatarPersona';
+import { usePropsFor } from '../ChatComposite/hooks/usePropsFor';
```
prprabhu-ms, line 18: This is a bit of a red flag, that code from `common` is calling `ChatComposite/hooks/usePropsFor`.

`usePropsFor` from there might only work for chat components.  Looking at how this is used below, I'd recommend that `ParticipantContainerProps.participatnListProps` be a required field, that is populated with `usePropsFor` from inside `ChatComposite` and `MeetingComposite`.

prprabhu-ms (commented):
Will need changefiles.

packages/react-composites/src/composites/common/ParticipantContainer.tsx
dmceachernmsft: okay I will clean this up!

packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -1,42 +1,43 @@
...
   ParticipantList,
   ParticipantListProps,
   ParticipantMenuItemsCallback
 } from '@internal/react-components';
-import { FocusZone, Stack } from '@fluentui/react';
-import { useLocale } from '../localization';
-import { usePropsFor } from '../ChatComposite/hooks/usePropsFor';
+import { concatStyleSets, FocusZone, Stack, useTheme } from '@fluentui/react';
 import { AvatarPersona, AvatarPersonaDataCallback } from './AvatarPersona';
+import { peopleSubheadingStyle } from './styles/SidePane.styles';
```
prprabhu-ms, line 13: naming nit: `SidePange.styles` harkens to how this component is used by the `MeetingComposite` (but not `ChatComposite`. How about `ParticipantContainer.styles` :)

packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -1,42 +1,43 @@
...
  * @private
  */
 export const ParticipantContainer = (props: ParticipantContainerProps): JSX.Element => {
-  const { onFetchAvatarPersonaData, onFetchParticipantMenuItems } = props;
-  const participantListProps = props.participantListProps ? props.participantListProps : usePropsFor(ParticipantList);
-  const locale = useLocale();
-  const chatListHeader = locale.strings.chat.chatListHeader;
+  const { onFetchAvatarPersonaData, onFetchParticipantMenuItems, title, participantListProps } = props;
+  const theme = useTheme();
+  const subheadingStyleThemed = concatStyleSets(peopleSubheadingStyle, {
```
prprabhu-ms, line 29: Not a 100% sure why this was needed.

prprabhu-ms:
Would like to see snapshot updates (if any) before approval.

packages/react-composites/src/composites/common/ParticipantContainer.tsx
dmceachernmsft: makes sense to me! will take care of this

packages/react-composites/src/composites/common/ParticipantContainer.tsx
dmceachernmsft: the styling? just to use the styles that were there before, I am not 100% sure they worked correctly so I am waiting on the diffs to see

dmceachernmsft: > Would like to see snapshot updates (if any) before approval.

they updated the snapshots and it appears that there is a scroll bar present on the small version, something that I think is a symptom of the ParticipantContainer wrapping the ParticipantList.

prprabhu-ms (commented):
image

The shadow doesn't look right.

prprabhu-ms (commented):
image

As you noted, the scrollbar ain't wanted either.

dmceachernmsft:
> image
> 
> The shadow doesn't look right.

I think its a symptom of using the ParticipantContainer, I'll look into it!

change/@internal-react-composites-5523c033-6f94-4a66-9fb6-5910d071fb8c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
JamesBurnside, line 2: nit: going by the comment `"changed meeting peopel pane to use ParticipantContainer Component."` this should be type: none as this is just an internal refactor (so no implication on customers).

JamesBurnside (approved):

prprabhu-ms (approved):

PorterNan (approved):
Let's make a separated PR to create a single file and do all the conditional stuff there, then import it in the file (so that we can minimized /* conditional-xxx */ comments)

PR #1329, opened 2022-01-19
[Sample/Server]  Add comment around in-memory map of users on server

carocao-msft:
# What
Add comment around in-memory map of users on server


samples/Server/src/lib/chat/userIdToUserConfigMap.ts
```diff
@@ -7,4 +7,6 @@ interface UserConfig {
   id: string;
 }
 
+// This is for samples only.
+// All User information (name, profile image) are stored in memory only and deleted after refresh
```
prprabhu-ms, line 11: nitty nit: "refresh" might be confusing here. How about "when the server restarts".

Should we link to some potential storage technologies, perhaps on azure? wdyt @alkwa-msft ?

prprabhu-ms (approved):

samples/Server/src/lib/chat/userIdToUserConfigMap.ts
carocao-msft: Updated :)

alkwa-msft: The idea is to keep it simple.

"For the purpose of this sample, we opted to use an in-memory data store. This means that if the web application is restarted any information maintained would be wiped. For longer term storage solutions we suggest using [Azure Blob Storage](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction)

carocao-msft: Updated comments

JamesBurnside: We should probably be telling customers to use an Azure database (https://azure.microsoft.com/en-ca/product-categories/databases/) instead of a file storage as this has many read writes

prprabhu-ms: Heh, I had suggested yesterday that we link to https://docs.microsoft.com/en-us/azure/architecture/guide/technology-choices/data-store-decision-tree ( let Azure recommend what to use ;) )

JamesBurnside: Oh yeah that's way better ^

carocao-msft: So should I update the link in the comments to the decision tree one?

PorterNan (approved):

samples/Server/src/lib/chat/userIdToUserConfigMap.ts
carocao-msft: Updated the comment, do you mind having a quick look?

JamesBurnside (approved):

PR #1330, opened 2022-01-19
[Flakey UI tests] Try ensure fonts have loaded in e2e tests

JamesBurnside:
# What
Ensure fonts have loaded by using local fonts for segoe ui and using `preload` to ensure they have loaded.
* Local files ensures they do not have to be fetched over the network.
* `preload` ensures that the fonts load before the page continues.

# Why
Some tests are still failing because segoe ui failed to load or has not yet loaded:
image

# How Tested
Ran locally to ensure that before any composite has loaded, text is rendered using the local segoe ui fonts.

## Any Issues
If fluent update the font we won't get that update until someone updates the local files - perhaps a good thing though?


packages/react-composites/tests/browser/common/testapp.webpack.config.js
```diff
@@ -46,12 +47,21 @@ module.exports = (appDir, babelConfig) => {
...
           type: 'asset/inline'
         }
       ]
     },
-    plugins: [new HtmlWebpackPlugin({ template: path.join(__dirname, 'index.html') })],
+    plugins: [
+      new HtmlWebpackPlugin({ template: path.join(__dirname, 'index.html') }),
+      new CopyWebpackPlugin({
+        patterns: [
+          { from: path.join(__dirname, 'fonts', 'segoeui-bold.woff2'), to: 'fonts' },
```
prprabhu-ms, line 59: Can we not just copy everything inside `__dirname/'fonts'` (why list the files?)

prprabhu-ms (approved):

packages/react-composites/tests/browser/common/testapp.webpack.config.js
JamesBurnside: Purely because I weirdly kept getting:
image
For all the variations of:
```
{ from: path.join(__dirname, 'fonts', '*.woff2'), to: 'fonts' }
```
Do you know what it should be? Acoording to the documentation that wildcard should work^

prprabhu-ms: Not in the mood to fight a webpack plugin this afternoon. Will trust that you tried "long 'nuff"

PorterNan (approved):
Another Amazing change for improving flaky UI tests!
So currently it is waiting some kinds of functions like `loadFonts()` from fluentUI to load the font before we really load the page, wondering that why that fails (or delays) and is it possible we set some waitFor() function for that part?

I am imagining a very rare case that we might do something wrong to make fonts not loaded correctly in fluent in the future, and we might lose the way to test that part (or maybe we could have a single one test without this preload to test fonts?)

alkwa-msft (approved):

PR #1332, opened 2022-01-19
added missing return types.

dmceachernmsft:
# What
Add missing return types to existing functions

# Why
image

seen in PR 1328

# How Tested


prprabhu-ms (approved):

alkwa-msft (approved):

carocao-msft (approved):

JamesBurnside (approved):

change/@internal-react-composites-5f04960f-d30f-47ba-a8ca-d27896dd60a6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
JamesBurnside, line 2: note: this should be changetype `"none"` for this type of change.
The reason being is customers aren't impacted by this change - so we would have no need to include it on a changelog.
Patch is for bug fixes that don't change behavior and minor is for new features (can read more here: https://semver.org/)

PR #1336, opened 2022-01-20
Update the best practice for submitting PR for conditional build

PorterNan:
# What

Update the best pratice for submitting PR for conditional build

# Why

# How Tested

# Process & policy checklist


alkwa-msft (approved):

prprabhu-ms (approved):
We need dedicated documentation for build flavors. They're non-standard enough and can trip up developers easily.

PorterNan:
> We need dedicated documentation for build flavors. They're non-standard enough and can trip up developers easily.

Agree, we need another one just for build flavors, this is more general one for quick PR create, then we should have a more detailed one and point people there

PR #1337, opened 2022-01-20
[Sample/Chat&Call] adding scope param for token request

carocao-msft:
# What
For calling and chat ,added ?scope=chat so only the corresponding token will be minted 
Related task: https://skype.visualstudio.com/SPOOL/_workitems/edit/2691358


JamesBurnside (approved):

samples/Calling/src/app/utils/AppUtils.ts
```diff
@@ -9,7 +9,7 @@ import { v1 as generateGUID } from 'uuid';
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const fetchTokenResponse = async (): Promise<any> => {
-  const response = await fetch('/token');
+  const response = await fetch('/token?scope=voip');
```
PorterNan, line 12: Question about the PR title, this is a client-side change right? And why are we removing scope from client side request?

PorterNan: Maybe like [Sample/Chat&Call] adding scope param for token request

PorterNan (approved):

samples/Calling/src/app/utils/AppUtils.ts
PorterNan: Then add the related PR of server side in the description, so people can check then together when reviewing

carocao-msft: Updated :)

PR #1338, opened 2022-01-20
[Sample/Chat]  Pull Chat SDK version from version made in webpack and not at runtime in the Chat utils

carocao-msft:
# What
[Sample/Chat]  Pull Chat SDK version from version made in webpack and not at runtime in the Chat utils

# Why
Should not be reaching out to external folders to get version information

### DIFF
--- a/samples/Chat/src/app/utils/utils.ts
+++ b/samples/Chat/src/app/utils/utils.ts
@@ -4,13 +4,9 @@
 import { GUID_FOR_INITIAL_TOPIC_NAME } from './constants';
 
 declare let __BUILDTIME__: string; // Injected by webpack
+declare let __CHATVERSION__: string; // Injected by webpack
+export const chatSDKVersion = __CHATVERSION__;
 export const getBuildTime = (): string => __BUILDTIME__;
-
-export function getChatSDKVersion(): string {
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  return require('../../../package.json')['dependencies']['@azure/communication-chat'];
-}
-
 export const CAT = '🐱';
 export const MOUSE = '🐭';
 export const KOALA = '🐨';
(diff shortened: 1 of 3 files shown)


samples/Chat/src/app/utils/utils.ts
```diff
@@ -4,13 +4,9 @@
 import { GUID_FOR_INITIAL_TOPIC_NAME } from './constants';
 
 declare let __BUILDTIME__: string; // Injected by webpack
+declare let __CHATVERSION__: string; // Injected by webpack
```
PorterNan, line 7: Nice! Didn't know we are injecting this

samples/Chat/src/app/utils/utils.ts
```diff
@@ -4,13 +4,9 @@
...
+export const chatSDKVersion = __CHATVERSION__;
```
PorterNan, line 8: Follow the syntax of getBuildTime to make it consistent? Just make it a function and return the value

JamesBurnside (approved):

samples/Meeting/src/app/utils/utils.ts
```diff
@@ -4,13 +4,9 @@
 import { GUID_FOR_INITIAL_TOPIC_NAME } from './constants';
 
 declare let __BUILDTIME__: string; // Injected by webpack
+declare let __CHATVERSION__: string; // Injected by webpack
+export const chatSDKVersion = __CHATVERSION__;
```
PorterNan, line 8: Same here, follow getBuildTime function style

PorterNan (approved):

samples/Meeting/src/app/utils/utils.ts
carocao-msft: Updated :)

PR #1339, opened 2022-01-20
fixed call joining issues

dmceachernmsft:
# What
Fixed issue with two different groupId's.

# Why
Stops issue where the initial meeting participant would be left in seperate call.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2691659
# How Tested

Ran locally to verify problem resolution. inspected in debugger to check only 1 groupId is being made.


samples/Meeting/src/app/App.tsx
```diff
@@ -102,20 +100,20 @@ const App = (): JSX.Element => {
           joiningExistingCall={joiningExistingMeeting}
           startMeetingHandler={(callDetails) => {
             const isTeamsMeeting = !!callDetails.teamsLink;
-            setCallArgs({
+            const localCallArgs = {
               displayName: callDetails.displayName,
               callLocator: callDetails.teamsLink || getTeamsLinkFromUrl() || getGroupIdFromUrl() || createGroupId()
-            });
+            };
+            setCallArgs(callArgs);
```
prprabhu-ms, line 107: Should this be `localCallArgs` instead?

samples/Meeting/src/app/App.tsx
```diff
@@ -33,7 +31,7 @@ interface Credentials {
 }
 
 interface CallArgs {
-  callLocator: GroupLocator | TeamsMeetingLinkLocator;
+  callLocator: GroupLocator | TeamsMeetingLinkLocator | undefined;
```
JamesBurnside, line 34: instead of callLocator being able to be undefined, it would be better to do:
```
const [callArgs, setCallArgs] = useState<CallArgs | undefined>(undefined);
```

samples/Meeting/src/app/App.tsx
dmceachernmsft: fixed, oops 😅

prprabhu-ms (approved):

JamesBurnside (approved):

alkwa-msft (approved):

PR #1341, opened 2022-01-20
[sample/Calling] Logging when using a mobile device on landscape mode

edwardlee-msft:
# What
Simplified code by removing console log output when mobile or small screen is detected as mobile is now supported

# Why
Mobile is now supported. Outputting a console log when a mobile device is detected is no longer needed.
https://skype.visualstudio.com/SPOOL/_workitems/edit/2693378

# How Tested
Tested on chrome dev tools mobile view. Console log warning no longer being generated when opening on smaller or mobile screen.

# Process & policy checklist


edwardlee-msft:
Suggestion for future followup: check whether or not mobile phone is on landscape, then check against CallComposites minimum dimensions to ensure screen meets minimum dimensions.

samples/Calling/src/app/utils/AppUtils.ts
```diff
@@ -57,10 +57,14 @@ export const isOnIphoneAndNotSafari = (): boolean => {
...
 };
 
-export const isMobileSession = (): boolean =>
-  !!window.navigator.userAgent.match(/(iPad|iPhone|iPod|Android|webOS|BlackBerry|Windows Phone)/g);
+/*
+ * TODO:
+ *  Remove once landscape check is
+ */
+// export const isMobileSession = (): boolean =>
+//   !!window.navigator.userAgent.match(/(iPad|iPhone|iPod|Android|webOS|BlackBerry|Windows Phone)/g);
```
JamesBurnside, line 65: just completely remove these all this commented code and associated todo in code.
If we want to show an error in landscape/too small we should create a bug but I think its unecessary as we document this in our documentation.

alkwa-msft: As a general pattern we observe in this repo we delete blocks of commented code. We can always look in the git history for how a function looked. 

In rare occasions if its a small snippet that helps readability we do maintain some commented lines but that is generally more rare.

alkwa-msft: can't we just say .. isLandscape = (): boolean => window.innerWidth < window.innerHeight and use that?

edwardlee-msft: In addition, we can use it with isMobileSession to check if the user is on mobile and on landscape. I will give that a go and test it out.

samples/Calling/src/app/App.tsx
```diff
@@ -64,8 +64,8 @@ const App = (): JSX.Element => {
     return <UnsupportedBrowserPage />;
   }
 
-  if (isMobileSession() || isSmallScreen()) {
-    console.log('ACS Calling sample: This is experimental behaviour');
+  if (isMobileSession() || isLandscape()) {
```
JamesBurnside, line 67: and instead of or here
```suggestion
  if (isMobileSession() && isLandscape()) {
```

alkwa-msft (approved):

prprabhu-ms (approved):

PR #1343, opened 2022-01-21
made new file for the chatScreenPeoplePane

dmceachernmsft:
# What
Added new file for chatScreenPeoplePane

# Why
Allows for less Conditional Compile statements in the chat screen and seperates the Chat version of the people pane from the ChatScreen.

# How Tested
Ran locally to ensure that it worked as expected.

### DIFF
--- a/packages/react-composites/src/composites/ChatComposite/ChatScreenPeoplePane.tsx
+++ b/packages/react-composites/src/composites/ChatComposite/ChatScreenPeoplePane.tsx
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { ParticipantList, ParticipantMenuItemsCallback } from '@internal/react-components';
+import React from 'react';
+import { AvatarPersonaDataCallback } from '..';
+import { ParticipantContainer } from '../common/ParticipantContainer';
+import { useLocale } from '../localization';
+import { usePropsFor } from './hooks/usePropsFor';
+
+export type ChatScreenPeopelPaneProps = {
+  onFetchAvatarPersonaData: AvatarPersonaDataCallback | undefined;
+  onFetchParticipantMenuItems: ParticipantMenuItemsCallback | undefined;
+};
+
+export const ChatScreenPeoplePane = (props: ChatScreenPeopelPaneProps): JSX.Element => {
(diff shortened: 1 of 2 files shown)


JamesBurnside (approved):

packages/react-composites/src/composites/ChatComposite/ChatScreenPeoplePane.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { ParticipantContainer } from '../common/ParticipantContainer';
+import { useLocale } from '../localization';
+import { usePropsFor } from './hooks/usePropsFor';
+
+export type ChatScreenPeopelPaneProps = {
+  onFetchAvatarPersonaData: AvatarPersonaDataCallback | undefined;
+  onFetchParticipantMenuItems: ParticipantMenuItemsCallback | undefined;
+};
+
+export const ChatScreenPeoplePane = (props: ChatScreenPeopelPaneProps): JSX.Element => {
```
PorterNan, line 16: This is a great example of how to write clean conditional compilation code - instead of adding /* conditional compilation*/ everywhere, we need a separated module then we don't have to write multiple conditional lines. (and that's the reason of why we only strip the following line of comment). Thanks for being a pilot and doing this exploration!  @dmceachernmsft 

In the meantime, I would like to involve @prprabhu-ms to discuss this code style a little bit
- Inline the conditional component in the same file or create separated file - I believe that depends on length of code, it looks good to have it in separated file cause it's 30 lines of code

packages/react-composites/src/composites/ChatComposite/ChatScreenPeoplePane.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { ParticipantContainer } from '../common/ParticipantContainer';
+import { useLocale } from '../localization';
+import { usePropsFor } from './hooks/usePropsFor';
+
+/**
+ * Props for the chat screen people pane
+ * @private
+ */
+type ChatScreenPeoplePaneProps = {
+  onFetchAvatarPersonaData: AvatarPersonaDataCallback | undefined;
```
prprabhu-ms, line 16: This could be `onFetchAvatarPersonaData?: AvatarPersonaDataCallback` etc.

prprabhu-ms (approved):

alkwa-msft (approved):

PR #1345, opened 2022-01-21
Updated MessageThreads to remove usernames being displayed on top of messages in Storybook

edwardlee-msft:
# What
Updated MessageThreads to remove usernames being displayed on top of messages in Storybook - UI Component - Message Thread by setting imageInitials directly. 

# Why
Usernames were being displayed on top of messages causing persona icons to be displayed on top rather than on the left side of the message. This was also causing duplicate username to be populated, one at top and one in the message itself.
https://skype.visualstudio.com/SPOOL/_workitems/edit/2667113/

# How Tested
Testing on local storybook and compared results in chat samples locally.

# Process & policy checklist


dmceachernmsft (approved):

JamesBurnside (dismissed):
This causes a small issue where the color of the avatars is no longer consistent:
image

(after this change the DG of the chat message is blue, which no longer matches the DG color in the participants list)

PorterNan:
> This causes a small issue where the color of the avatars is no longer consistent: image
> 
> (after this change the DG of the chat message is blue, which no longer matches the DG color in the participants list)

I guess the color was calculated by displayname text, so to keep the color, I guess we will still need to keep displayname but hide the detailInformation - check my comment in another PR for removing displayname there

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -323,7 +333,7 @@ const memoizeAllMessages = memoizeFnAll(
         const personaOptions: IPersona = {
           hidePersonalDetails: true,
           size: PersonaSize.size32,
-          text: message.senderDisplayName
+          imageInitials: getInitials(message.senderDisplayName, false, true)
```
PorterNan, line 336: Didn't know they exported this, a great finding!

edwardlee-msft: "Another nice way to do it is just adding hidePersonaDetails (might save us tons of time of writing an initials function), with text still there, which will hide the name for you, check my codepen for an example:" - PorterNan

Issue is, the hidePersonalDetails is set to true, even in the react components when looking through devtools... I will investigate further as to why storybook continues to display the name even with hidePersonalDetails set to true.

edwardlee-msft: Found that when "text" prop gets populated, a TooltipHostBase gets generated as well. If I manually set the children to empty string "" through dev tools, the name disappears on top of the message box.

If "text" prop does not get populated, then no TooltipHostBase gets generated.

Screen Shot 2022-01-26 at 2 01 24 PM

edwardlee-msft: Resolved. Noticed there was humanerror previously. Instead of the prop "hidePersonaDetails" being used, it was misspelt previously as "hidePersona**l**Details".

prprabhu-ms:
Awesome! Oops about that typo... after all this debugging <3

image

That box shadow seems unrelated to this PR. Donald removed that recently, you probably just need to merge from `main`.

prprabhu-ms (approved):

PR #1347, opened 2022-01-22
HorizontalGallery UI tests using MockCallingAdapter

mgamis-msft:
# What

- Changed Calling playwright app such that if state is provided as a qArgs in the test then we will use the the `MockCallingAdapter`. Otherwise an actual `CallingAdapter` is created using `createAzureCommunicationCallAdapter`
will be used for the CallComposite.
- HorizontalGallery UI snapshot test for 1 page and multiple pages of participants added
- Identifiers added HorizontalGallery navigation buttons to perform click in playwright test

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2686575

# How Tested

Running HorizontalGallery.test.ts playwright tests along to check if they consistently pass

# Process & policy checklist


JamesBurnside:
Remove snapshots that were generated locally:
image

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -0,0 +1,237 @@
...
+  let participantKey = 1;
+
+  for (const remoteParticipant of mockRemoteParticipants) {
+    let view: VideoStreamRendererViewState | undefined = undefined;
+    if (remoteParticipant.isVideoStreamAvailable) {
+      const mockVideoElement = document.createElement('div');
+      mockVideoElement.innerHTML = '<span />';
+      mockVideoElement.style.width = decodeURIComponent('100%25');
+      mockVideoElement.style.height = decodeURIComponent('100%25');
+      mockVideoElement.style.background = stringToHexColor(remoteParticipant.displayName);
```
JamesBurnside, line 150: fun! can you doc comment the `createMockRemoteParticipants` function with information like this
```
/** 
 * Creates mock remote participants with mocked videos. The video is an empty html5 video element with a colored background. To ensure a consistent color the background color is the hex representation of the participant's display name.
 */
```

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+}
+
+/**
+ * Helper function to create remoteParticipants property of CallAdapterState from an array of TestRemoteParticipant
+ * @param mockRemoteParticipants - array of TestRemoteParticipant
+ * @returns Record of RemoteParticipantState
+ */
+
+const createMockRemoteParticipants = (
```
JamesBurnside, line 135: rename this function -- it is called `createMockRemoteParticipants` but takes in an already created array called `mockRemoteParticipants`.
Judging by the logic what a suggestion for this would be:
`createMockRemoteParticipantsWithVideo`?

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -0,0 +1,237 @@
...
+    return;
+  }
+}
+
+/**
+ * Helper function to create remoteParticipants property of CallAdapterState from an array of TestRemoteParticipant
+ * @param mockRemoteParticipants - array of TestRemoteParticipant
+ * @returns Record of RemoteParticipantState
+ */
+
```
JamesBurnside, line 134: Remove empty line here

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
JamesBurnside: Hmm actually this does more than just add video information, it also returns an object of keyed participants...

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -0,0 +1,237 @@
...
+const createMockRemoteParticipants = (
+  mockRemoteParticipants: TestRemoteParticipant[]
+): Record<string, RemoteParticipantState> => {
+  const remoteParticipants: Record<string, RemoteParticipantState> = {};
+
+  // Incrementing participant key starting at 1
```
JamesBurnside, line 140: Why does the key start at 1 and not zero?

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -0,0 +1,237 @@
...
+  let participantKey = 1;
+
+  for (const remoteParticipant of mockRemoteParticipants) {
```
JamesBurnside, line 143: nit: instead of using a key that incremenets on each loop, you can instead do:
```
  for (const [remoteParticipant, i] of mockRemoteParticipants) {
```

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
JamesBurnside: Consider splitting into two separate functions for better clarity and to ensure that a function only has one purpose.
Instead of:
```ts
const remoteParticipants = createMockRemoteParticipants(testState.remoteParticipants);
```
This could instead be something like:
```ts
const remoteParticipants = convertTestParticipantsToCallAdapterStateParticipants(testState.remoteParticipants);
Object.values(remoteParticipants).addMockVideo(participant);
```

packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  test('HorizontalGallery should have 1 audio participant', async ({ pages, serverUrl }) => {
+    const page = pages[0];
+    await page.goto(
+      buildUrlWithMockAdapter(serverUrl, {
+        remoteParticipants: defaultTestRemoteParticipants
+      })
+    );
+    // Click off the screen to turn away initial aria label
+    await page.mouse.click(-1, -1);
+    await waitForPageFontsLoaded(page);
```
JamesBurnside, line 36: instead of waiting for fonts to have loaded, is there a page element that could be waited for instead waitForPageFontsLoaded won't have guaranteed that react has done all the necessary render passes to have fully rendered the app

packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+    ]);
+    await page.goto(
+      buildUrlWithMockAdapter(serverUrl, {
+        remoteParticipants: testRemoteParticipants
+      })
+    );
+    // Click off the screen to turn away initial aria label
+    await page.mouse.click(-1, -1);
+    await waitForPageFontsLoaded(page);
```
JamesBurnside, line 74: likewise here, can we wait for selector instead/aswell?

packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
mgamis-msft: I tried waiting for the VideoGallery instead but often fails because the text has not yet loaded as seen in the actual snapshot:
horizontal-gallery-actual
I tried tried using your waitForPageFontsLoaded and it always passes so I figure it was the bottleneck. I could also wait for both if that is better.

packages/communication-react/review/communication-react.api.md
```diff
@@ -1291,6 +1291,8 @@ export interface _IdentifierProviderProps {
 
 // @internal
 export interface _Identifiers {
+    horizontalGalleryLeftNavButton: string;
```
prprabhu-ms, line 1294: It might have been a mistake exposing `_Identifiers` through the public API (we thought @internal was OK for this, but unsure).

Let's not add to this anymore. Just directly use `data-ui-id` in the component and directly use the `data-ui-id` from e2e tests, like we already do for composites.

Thanks for trying to stick to this pattern...

packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+test.describe('HorizontalGallery tests', async () => {
+  test('HorizontalGallery should have 1 audio participant', async ({ pages, serverUrl }) => {
+    const page = pages[0];
+    await page.goto(
+      buildUrlWithMockAdapter(serverUrl, {
+        remoteParticipants: defaultTestRemoteParticipants
+      })
+    );
+    // Click off the screen to turn away initial aria label
```
prprabhu-ms, line 34: Don't we already have a helper function for this? If not, let's create on `clickOffPage()` so that the function name serves instead of repeating this comment everywhere.

packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    await page.mouse.click(-1, -1);
+    await waitForPageFontsLoaded(page);
+    expect(await page.screenshot()).toMatchSnapshot('horizontal-gallery.png');
```
prprabhu-ms, line 37: Suggest using a more descriptive name for the snapshot (all these tests are for horizontal-gallery...)

packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  test('HorizontalGallery should have multiple audio participants spanning multiple pages. Navigation buttons should work.', async ({
+    pages,
+    serverUrl
+  }) => {
+    const page = pages[0];
+    const testRemoteParticipants = defaultTestRemoteParticipants.concat([
```
prprabhu-ms, line 45: There's little value in extending `defaultTestRemoteParticipants` here. Just add all the participants inline, it saves a reader from having to go look another place just to see half the list.

packages/react-composites/tests/browser/call/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param state - State
+ * @param qArgs - Optional args to add to the query search parameters of the URL.
+ * @returns URL string
+ */
+export const buildUrlWithMockAdapter = (
+  serverUrl: string,
+  state?: TestCallingState,
+  qArgs?: { [key: string]: string }
+): string => {
+  return `${serverUrl}?${encodeQueryData({ useMockAdapter: 'true', state: JSON.stringify(state), ...qArgs })}`;
```
prprabhu-ms, line 20: Why use two different flags for the same thing. Instead of `useMockAdapter` and `state`, I'd recommend naming `state` better -- `mockCallState`. And when that is set, the app uses a mock adapter.

packages/react-composites/tests/browser/call/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT license.
+
+import { encodeQueryData } from '../common/utils';
+import { TestCallingState } from './TestCallingState';
+
+/**
+ * Create the test URL.
+ * @param serverUrl - URL of webpage to test, this is typically https://localhost:3000
+ * @param user - Test user the props of which populate query search parameters
+ * @param state - State
```
prprabhu-ms, line 11: This `@param` is not very useful. Since this is a test function, just remove it if there's nothing beyond the type you want to say.

packages/react-composites/tests/browser/call/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param qArgs - Optional args to add to the query search parameters of the URL.
+ * @returns URL string
```
prprabhu-ms, line 13: I don't see no `URL` arg ;)

packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -20,19 +20,15 @@ import {
...
 const params = Object.fromEntries(urlSearchParams.entries());
 
-// Required params
-const displayName = verifyParamExists(params.displayName, 'displayName');
-const token = verifyParamExists(params.token, 'token');
-const groupId = verifyParamExists(params.groupId, 'groupId');
-const userId = verifyParamExists(params.userId, 'userId');
-
-// Optional params
+const state = JSON.parse(params.state) as TestCallingState;
```
prprabhu-ms, line 31: The name `state` is too generic. How about `mockCallingState`. Also see my comment below about not using `useMockAdapter` at all.

packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -43,13 +39,22 @@ function App(): JSX.Element {
...
+        const groupId = verifyParamExists(params.groupId, 'groupId');
+        const userId = verifyParamExists(params.userId, 'userId');
+
+        const newAdapter = await createAzureCommunicationCallAdapter({
+          userId: fromFlatCommunicationIdentifier(userId) as CommunicationUserIdentifier,
+          displayName,
+          credential: new AzureCommunicationTokenCredential(token),
+          locator: { groupId: groupId }
+        });
+        setCallAdapter(wrapAdapterForTests(newAdapter));
```
prprabhu-ms, line 56: Can we always `wrapAdapterForTests`, regardless of whether it's a real or mock adapter? Branches in code where wrapping happens or not make it harder to follow.

packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -43,13 +39,22 @@ function App(): JSX.Element {
...
-        userId: fromFlatCommunicationIdentifier(userId) as CommunicationUserIdentifier,
-        displayName,
-        credential: new AzureCommunicationTokenCredential(token),
-        locator: { groupId: groupId }
-      });
-      setCallAdapter(wrapAdapterForTests(newAdapter));
+      if (params.useMockAdapter) {
+        setCallAdapter(new MockCallAdapter(state));
+      } else {
+        const displayName = verifyParamExists(params.displayName, 'displayName');
```
prprabhu-ms, line 45: At this point, I'd recommend creating a helper function for line 45 - 54. The code block here makes the conditional code flow harder to follow.

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -0,0 +1,248 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { VideoStreamOptions } from '@internal/react-components';
+import { AudioDeviceInfo, PermissionConstraints, VideoDeviceInfo, Call } from '@azure/communication-calling';
+import { RemoteParticipantState, VideoStreamRendererViewState } from '@internal/calling-stateful-client';
+import { CallAdapterState, CallAdapter } from '../../../../../src';
+import { TestCallingState, TestRemoteParticipant } from '../../TestCallingState';
+
+/**
+ * Mock class that implements CallAdapter interface for UI snapshot tests
```
prprabhu-ms, line 10: Please add detailed docstring about what the mock can / can't do : You can set a snapshot of adapter state on creation, but none of the handlers are implemented, so there is no interactivity.

It'd be helpful to read at a glance what is possible / not possible.

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -0,0 +1,248 @@
...
+    const value = (hash >> (i * 8)) & 0xff;
+    colour += ('00' + value.toString(16)).substr(-2);
+  }
+  return colour;
+};
+
+/**
+ * Default call adapter state that the {@link MockCallAdapter} class is initialized with
+ */
+const defaultCallAdapterState: CallAdapterState = {
```
prprabhu-ms, line 209: I think that this setup is quite hidden / hard to follow:
- test creates a `TestCallingState`
- passes it in as qargs to app
- app creates a new MockCallAdapter
- which then populates a whole bunch of values in the adapter state from this default.

Instead, consider the following flow:
- test uses a helper method that creates a `CallAdapterState`. The helper could simply accept a list of remote participants, filling in all the rest with defaults like here: `createMockStateWithRemoteParticipants(...)`.
- app accepts a `CallAdapterState` via quargs, and directly uses that.

This flow is less magical for somebody reading the test.

anjulgarg (approved):

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -0,0 +1,248 @@
...
+  state: CallAdapterState;
+
+  addParticipant(participantKey: string, participantState: RemoteParticipantState): void {
+    if (!this.state.call) {
+      return;
+    }
+    this.state.call.remoteParticipants[participantKey] = participantState;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any
```
prprabhu-ms, line 38: Please drop these eslint directives.

Since none of the arguments are used, you can satisfy the interface with empty arg functions:

```ts
  onStateChange(): void {
    throw Error('Handlers not implemented');
  }
```

I'd also recommend throwing an error from all the handler implementations. That way, if we mistakenly call handlers in the test, the test fails instead of passing silently.

packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
JamesBurnside: Interesting there is *no* font and not just an incorrect one (i.e. that segoe ui hadn't loaded yet so fallsback to arial or something). Yeah put in a wait for both these things then

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
mgamis-msft: Sure. But I actually can't throw an error for onStateChange but I throw Exceptions for the rest and removed the eslint directives

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
mgamis-msft: Yes, I agree that the flow is a bit convoluted. But I was unable to figure out how to import `CallAdapterState` into 
'packages/react-composites/tests/browser/call/app' without causing the tests to fail with vague error 'window is not defined' when I run the tests. I will create a bug to simplify the setup. But for now I think this should suffice as the tests are working.

packages/react-composites/tests/browser/call/utils.ts
mgamis-msft: Oh it's the return value description :)

JamesBurnside:
Icon seems to have moved upwards in packages/react-composites/tests/browser/call/Localization.test.ts-snapshots/localized-call-configuration-page-Mobile-Android-Portrait-win32.png
image

JamesBurnside:
Do you know why this fixed the french translations in these screenshots?
| before | after |
| -- | -- |
| image | image |

(packages/react-composites/tests/browser/call/Localization.test.ts-snapshots/localized-call-screen-Desktop-Chrome-win32.png)

packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -154,6 +148,22 @@ const unsetSpeakingWhileMicrophoneIsMuted = (state: CallAdapterState): CallAdapt
  */
 const memoizedUnsetSpeakingWhileMicrophoneIsMuted = memoizeOne(unsetSpeakingWhileMicrophoneIsMuted);
 
+// Function to create call adapter using createAzureCommunicationCallAdapter
+const createCallAdapterWithCredentials = async (): Promise<CallAdapter> => {
```
JamesBurnside, line 152: nice tidy refactor!

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -144,6 +144,13 @@ export const loadCallPage = async (pages: Page[]): Promise<void> => {
   }
 };
 
+/**
+ * Click off the Composite page
+ */
+export const clickOffPage = async (page: Page): Promise<void> => {
```
JamesBurnside, line 150: nit: consider `clickOutsideOfPage` -- because we have on/off for camera, mic etx. `off` feels like it is turning something off 😅

JamesBurnside (commented):
Looks great! Just some new discrepancies in some existing UI tests I had questions on

mgamis-msft:
> Icon seems to have moved upwards in packages/react-composites/tests/browser/call/Localization.test.ts-snapshots/localized-call-configuration-page-Mobile-Android-Portrait-win32.png image

It seems like the CI did not the snapshots for localization tests. Because I can see from our dogfood app that the bottom of the icon aligns with the bottom of the text in our buttons. 
image
Sorry for putting in these unrelated snapshot updates but I thought they ought to be updated.

prprabhu-ms (approved):

JamesBurnside (approved):

PR #1351, opened 2022-01-24
Added comments to make it more clear when we are setting the topic from the adapter

carocao-msft:
# What
Added comments to make it more clear when we are setting the topic from the adapter

### DIFF
--- a/samples/Chat/src/app/ChatScreen.tsx
+++ b/samples/Chat/src/app/ChatScreen.tsx
@@ -58,6 +58,8 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
           endChatHandler(removedBy !== userId);
         }
       });
+      // set topic to be Your Chat sample after getting adapter from createAzureCommunicationChatAdapter
+      // this is executed everytime when any of displayName, endpointUrl, threadId, token, userId, errorHandler, endChatHandler changes
       adapter.setTopic('Your Chat sample');
       adapter.on('error', (e) => {
         console.error(e);


samples/Chat/src/app/ChatScreen.tsx
```diff
@@ -58,6 +58,8 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
           endChatHandler(removedBy !== userId);
         }
       });
+      // set topic to be Your Chat sample after getting adapter from createAzureCommunicationChatAdapter
+      // this is executed everytime when any of displayName, endpointUrl, threadId, token, userId, errorHandler, endChatHandler changes
       adapter.setTopic('Your Chat sample');
```
JamesBurnside, line 63: My concern (reason for filing the original bug) wasn't so much about this needing a comment, but more that anytime a user joins the chat, they are going to reset the topic. This line changes the topic for everyone. As an example, if Contoso added code to update the topic to something like 'Chat with x,y,x', when a new user joins this will overwrite that set topic back to 'Your Chat sample'

carocao-msft: Updated my PR

samples/Server/src/lib/chat/moderator.ts
```diff
@@ -18,7 +17,7 @@ export const createThread = async (topicName?: string): Promise<string> => {
   const chatClient = new ChatClient(getEndpoint(), credential);
 
   const request: CreateChatThreadRequest = {
-    topic: topicName ?? GUID_FOR_INITIAL_TOPIC_NAME
+    topic: 'Your Chat sample'
```
JamesBurnside, line 20: this function takes in an optional `topicName` so keep the `topicName ?? `
```suggestion
    topic: topicName ?? 'Your Chat sample'
```

samples/Server/src/lib/chat/moderator.ts
```diff
@@ -4,7 +4,6 @@
 import { AzureCommunicationTokenCredential } from '@azure/communication-common';
 import { ChatClient, CreateChatThreadOptions, CreateChatThreadRequest } from '@azure/communication-chat';
 import { getEndpoint } from '../envHelper';
-import { GUID_FOR_INITIAL_TOPIC_NAME } from '../constants';
```
JamesBurnside, line 7: Can we remove this everywhere in the codebase? It shouldn't be being used anymore anywhere

JamesBurnside (approved):

samples/Chat/src/app/ChatScreen.tsx
JamesBurnside: Thanks! Update PR title and desc to match also

samples/Server/src/lib/chat/moderator.ts
carocao-msft: Updated :)

samples/Server/src/lib/chat/moderator.ts
carocao-msft: Updated :)

samples/Chat/src/app/utils/utils.ts
```diff
@@ -47,5 +45,4 @@ export const getBackgroundColor = (avatar: string): { backgroundColor: string }
   }
 };
 
-export const existsTopicName = (topicName?: string): boolean =>
-  !!topicName && topicName !== GUID_FOR_INITIAL_TOPIC_NAME;
+export const existsTopicName = (topicName?: string): boolean => !!topicName;
```
JamesBurnside, line 48: this should also be removed as its not used at all anywhere (sorry to pile cleanup onto your PR)

carocao-msft: haha np! Updated :)

prprabhu-ms:
Please update the PR title and description before merging.

prprabhu-ms (approved):

PR #1352, opened 2022-01-24
Upgrade communication signaling version to 1.0.0.beta.12

anjulgarg:
# What
Upgrading communication-signaling to add metadata to events data

# Why
The dependency @azure/communication-signaling needs to be updated to 1.0.0.beta.12 as the one we are using (beta.5) doesn't expose metadata in the event type ChatMessageReceivedEvent.

Also update `EventSubscriber` in Chat stateful client to capture the metadata information and forward it to the state.

(https://skype.visualstudio.com/SPOOL/_workitems/edit/2706004)

# How Tested

# Process & policy checklist


JamesBurnside (approved):

prprabhu-ms (commented):
We should instead be dropping this dependency entirely. We're depending on a beta package from our stable versioned package, which is not good.

iiuc, all the functionality in `@azure/communication-signaling` _should_ be available via `@azure/communication-chat`.

anjulgarg:
> We should instead be dropping this dependency entirely. We're depending on a beta package from our stable versioned package, which is not good.
> 
> iiuc, all the functionality in `@azure/communication-signaling` _should_ be available via `@azure/communication-chat`.

Do you have an immediate solution for this?
Or suggestion for going forward?

prprabhu-ms:
> > We should instead be dropping this dependency entirely. We're depending on a beta package from our stable versioned package, which is not good.
> > iiuc, all the functionality in `@azure/communication-signaling` _should_ be available via `@azure/communication-chat`.
> 
> Do you have an immediate solution for this? Or suggestion for going forward?
> 
> Problem is that Event Types are not being exported by communication-chat

Let's bring this up with the chat team. There is no stable supported way to use Event Types right now in that case.

prprabhu-ms (approved):
Approving this package bump because it doesn't actually introduce a beta dep... just keeps one already there :/

PR #1353, opened 2022-01-24
File Card Component for File Sharing

anjulgarg:
# What
A File Card component for showing file and it's progress during file sharing.
image

FIGMA - https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?node-id=3970%3A22690

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2706271

# How Tested

# Process & policy checklist


packages/communication-react/package.json
```diff
@@ -40,7 +40,8 @@
     "react-linkify": "^1.0.0-alpha",
     "reselect": "~4.0.0",
     "uuid": "^8.1.0",
-    "@fluentui/react-hooks": "~8.3.8"
+    "@fluentui/react-hooks": "~8.3.8",
+    "@fluentui/react-file-type-icons": "~8.5.8"
```
JamesBurnside, line 44: sort alphabetically

packages/react-composites/package.json
```diff
@@ -63,7 +63,8 @@
     "memoize-one": "~5.2.1",
     "events": "~3.3.0",
     "uuid": "^8.1.0",
-    "reselect": "~4.0.0"
+    "reselect": "~4.0.0",
+    "@fluentui/react-file-type-icons": "~8.5.8"
```
JamesBurnside, line 67: sort alphabetically

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+ * @beta
+ */
+export const FileCard = (props: FileCardProps): JSX.Element => {
+  const { fileName, fileExtension, progress, actionIcon } = props;
+  const theme = useTheme();
+
+  const progressIndicator = (): JSX.Element => (
+    <ProgressIndicator
+      percentComplete={progress}
+      styles={{
```
JamesBurnside, line 44: nit: can the styles be constants defined outside the render loop

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCardGroup.tsx
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @beta
+ */
+export const FileCardGroup = (props: FileCardGroupProps): JSX.Element => {
+  const { children } = props;
+  return (
+    <Stack
+      horizontal
+      className={mergeStyles({
+        flexFlow: 'row wrap',
```
JamesBurnside, line 22: instead of defining naked css, can this be something like `<Stack horizonal wrap tokens={ { childrenGap: '0.5rem 0.5rem' } }  />` (not sure exactly what it should be)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/**
+ * @beta
+ */
+export const FileCard = (props: FileCardProps): JSX.Element => {
+  const { fileName, fileExtension, progress, actionIcon } = props;
+  const theme = useTheme();
+
+  const progressIndicator = (): JSX.Element => (
```
JamesBurnside, line 41: Can this just be inline where it's used and not define a new component with the same name?
```
{progress !== undefined && progress < 1 && <ProgressIndicator styles={progressStyles} percentComplete={progress} />
```

JamesBurnside (approved):

packages/communication-react/package.json
anjulgarg: Are we doing that every time we add a dependency?
Seems tedious :D
What's the utility for sorting these?

JamesBurnside: You can do: vs code -> highlight lines -> ctrl + shift + p -> Sort Ascending

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: I can do that for sure. However, I don't see the utility in this scenario. Am I missing something?

packages/communication-react/package.json
anjulgarg: I know I know ahhaha. I always forget tho. :P

anjulgarg: I wanted to understand the benefit of sorting these every time. :)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCardGroup.tsx
anjulgarg: Right. I was using childrenGap originally, but it doesn't work as expected for our designed. When child elements are wrapped and moved to the next line, the styling breaks unless done like this ;)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
JamesBurnside: For this component as it stands - not too much utility as presumably re-renders of this component will be negligible, but forcing this as a design pattern everywhere ensures we avoid scenarios where re-renders are expensive. This component could also grow or take in a child component in the future and the unnecessary re-renders could then be impactful.

Also, but a completely personal preference, I find it keeps the code small and functional - for this case once you hoist the styles outside component its clear this component doesn't do anything that the fluent component isn't already doing so means you no longer need to wrap the fluent component and can make this inline instead - making the code more streamlined

packages/communication-react/package.json
prprabhu-ms: Within reason, keep all lists sorted alphabetically. Avoids discussions on the "right order", and makes it trivial for tooling to enforce / auto-fix the order.

As an example, the import statements added via tooling are always sorted.

JamesBurnside: Ooo 'the utility' not 'a utility' sorry misread that lol. It just becomes a mess otherwise. If someone wants to check what dependencies are being used by a package/packlet - it's awful to look through when they are not sorted - it also means to find a package to update it you _have_ to do ctrl+f as finding manually in an unsorted list is impossible

This is definitely the sort of thing we should have prettier do though

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+   * File name.
+   */
+  fileName: string;
+  /**
+   * Extension of the file used for rendering the file icon.
+   */
+  fileExtension: string;
+  /**
+   * File upload/download progress percentage between 0 and 1.
+   * Not shown if undefined or 1.
```
prprabhu-ms, line 21: (1) what happens for invalid values: -1, 3 etc.
(2) It's a bit mysterious what is not shown if undefined. Perhaps "File transfer progress indicator is only shown when the value is  greater than 0 and less than 1.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+   */
+  progress?: number;
+  /**
+   * Icon to display for actions like download, upload, etc. on the right of file name.
```
prprabhu-ms, line 25: I wouldn't indicate where the icon is displayed ("on the right of") as that isn't really specified by the API.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+   */
+  actionIcon?: JSX.Element;
+  /**
+   * Function that runs when actionIcon is clicked
```
prprabhu-ms, line 29: Is the action run when `actionIcon` is clicked, or when the `FileCard` is clicked?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+ * @beta
+ */
+export const FileCard = (props: FileCardProps): JSX.Element => {
+  const { fileName, fileExtension, progress, actionIcon } = props;
+  const theme = useTheme();
+
+  return (
+    <Stack
+      className={mergeStyles({
+        width: '14rem',
```
prprabhu-ms, line 44: Avoid inline constants like this. Please define a styles constant that is merged in.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+        borderRadius: theme.effects.roundedCorner4,
+        boxShadow: theme.effects.elevation8
+      })}
+    >
+      <Stack
+        horizontal
+        horizontalAlign="space-between"
+        verticalAlign="center"
+        className={mergeStyles({
+          padding: progress === undefined ? '0.75rem' : '0.75rem 0.75rem 0.5rem 0.75rem'
```
prprabhu-ms, line 55: Please define a couple style constants and pick between the two.

Also, please add a comment as to why extra padding is needed when progress bar is being shown. It's very hard to understand the reason for CSS properties after the fact.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <Stack
+          style={{ cursor: 'pointer', padding: '0.25rem' }}
+          onClick={() => {
+            props.actionHandler && props.actionHandler();
+          }}
+        >
+          {actionIcon && actionIcon}
+        </Stack>
+      </Stack>
+      {progress !== undefined && progress < 1 && (
```
prprabhu-ms, line 79: What if progress is `-100`?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <ProgressIndicator
+          percentComplete={progress}
+          styles={{
+            itemProgress: {
+              height: 0,
```
prprabhu-ms, line 84: I'm confused by these styles. It'd help if you could add comments describing what they do (why is the `height` being set to 0?)

packages/communication-react/package.json
anjulgarg: Let me see if I can make prettier do it automatically ;)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCardGroup.tsx
prprabhu-ms: Please add this comment to the code for the next reader.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCardGroup.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * @beta
+ */
+export const FileCardGroup = (props: FileCardGroupProps): JSX.Element => {
+  const { children } = props;
+  return (
+    <Stack
+      horizontal
+      className={mergeStyles({
```
prprabhu-ms, line 21: Please define constants for all styles introduced here. Much easier to chase them that way ;)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: I can appreciate the personal preference you have.
However, I would like to understand the `re-render` logic here a bit more.
Are we saying that inline styles cause un-necessary re-renders?
OR that calling `mergeStyles` on every re-render is expensive?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: Right, but it let's a user know which icon this API changes?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: 1) Anything less than 0 is 0 and greater than 3 is 1.
2) Valid point. I will update the doc.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: When actionIcon is clicked.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: I want to understand why we want to avoid it?
Is there a strong reason?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: 0

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
prprabhu-ms: Could just say `Icon to display for actions like download, upload et. along with the file name`. There is only one icon in the `FileCard` :)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
prprabhu-ms: Rewording my question: *Should* the action run when `actionIcon` is clicked, or when the `FileCard` is clicked?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: So what's the difference between `along the file name` and `on the right of file name` ? :D

prprabhu-ms: It doesn't say where the icon will show up on the UI (for instance, should it be on the left in rtl mode?)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: For upload, only when the actionIcon is clicked, for download (possibly) when filecard is clicked.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
JamesBurnside: When this was reviewed it was a raw styles object and not merge styles:
image
And yeah raw styles into the styles prop causes the whole component to re-render as it creates a new reference each time https://github.com/microsoft/fluentui/wiki/Performance
Merge styles also works though to prevent the re-rendering instead of extracting constant out

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
prprabhu-ms: Same as discussion above with @JamesBurnside already. Let's just discuss it there.

First off, constants help discoverability a little bit (magic constants in the code are extremely hard to find. What if we wanted to change all `rem` to `em` (not that we should want such a thing...)). Second, it's a good practice because it allows pure components to skip a render if the other props are unchanged as well.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: Ah yes, I remember now. Thanks for sharing this doc.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
prprabhu-ms: Yeah, the docstring should document all these edge cases.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: Right. That's a good point.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: This is being handled by the underlying Fluent component.
If they have not documented it, I don't think we should either.
These edge cases are purely hypothetical and will never arise in a real scenario (unless due to a bug). Moreover, they are can be easily tested by cotoso.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: @prprabhu-ms makes sense to do it for `styles`. For `mergeStyles` I am unable to find a doc on perf impact if the styles are inline.

anjulgarg: I will do it :D
I just want to understand if there are stronger reasons other than personal preferences.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCardGroup.tsx
anjulgarg: Will do

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: Optimized the code to make it less confusing instead :)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -38,23 +48,39 @@ export const FileCard = (props: FileCardProps): JSX.Element => {
...
+  });
+
+  const actionIconClassName = mergeStyles({
+    cursor: 'pointer',
+    padding: '0.25rem'
+  });
+
+  const progressIndicatorStyles: IStyleFunctionOrObject<IProgressIndicatorStyleProps, IProgressIndicatorStyles> = {
+    itemProgress: {
+      // To make the progress indicator match the figma designs. It is a hack but works.
```
prprabhu-ms, line 73: Sorry, but I still don't understand why this is needed. Perhaps it's just my lack of understanding of CSS.

I was expecting something like: "setting the height to non-zero adds XYZ that we do not want"

We shouldn't reference figma designs in the comments / code - they're not publicly available (always), and in any case aren't versioned with the source code, so the comment isn't useful.

prprabhu-ms (approved):

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
anjulgarg: This is a hack. I don't understand why this works either to be honest. :/
But this is the only way I could make it look as per the figma specifications.
Gonna modify the comment.

anjulgarg: With that being said, I will try to find a better solution when I implement file sharing using this component.

PR #1354, opened 2022-01-24
Add ability to have local-only stories, Add scaffolding for PiPiP component to demonstrate

JamesBurnside:
# What
* Have an `INTERNAL` folder in storybook for internal-only stories
  * These only display on development environments and not on production environements
* Ensure that communication-react is not exporting _internal exports from react-components
* Add empty PiPiP component to show story and ensure exports not added to API file

# Why
We will have non-exported pure components by nature of only being able to support a subset of everything we create. However as a local dev I need to be able to run and see the component in storybook for local development

# How Tested
* `start storybook` (development env) - can see the Internal folder and the stories in side 
* `build storybook` (production env) - Internal folder is not there


packages/communication-react/review/communication-react.api.md
```diff
@@ -1270,32 +1270,6 @@ export interface HorizontalGalleryStyles extends BaseCustomStyles {
     previousButton?: IStyle;
 }
 
-// @internal
-export const _IdentifierProvider: (props: _IdentifierProviderProps) => JSX.Element;
-
-// @internal
-export interface _IdentifierProviderProps {
```
JamesBurnside, line 1277: We actually exported this from our GA package

prprabhu-ms: Would recommend splitting this change out of this PR and discussing it internally. I think this was only partially "accidental".

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import React from 'react';
+import { VideoTileProps } from '../VideoTile';
+
+/**
+ * Props for {@link _PictureInPictureInPicture} component.
+ *
+ * @internal
+ */
+export interface _PictureInPictureInPictureProps {
```
prprabhu-ms, line 12: I'm assuming this props definition is also only for demonstration purposes.

prprabhu-ms (approved):

packages/communication-react/review/communication-react.api.md
JamesBurnside: Great point, will do

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
JamesBurnside: Yeah, it'd v. much subject to change depending on implementation - I'll simplify for this PR

JamesBurnside: updated

packages/communication-react/review/communication-react.api.md
JamesBurnside: updated

anjulgarg (approved):

anjulgarg:
Awesome. Any public component will need to be exported explicitly now from communication-react?

JamesBurnside:
> Awesome. Any public component will need to be exported explicitly now from communication-react?

Yeah, the alternative would be to have two index.ts (so have `index.ts` and  have a `index.internal.ts`) and alias storybook to the internal one, but I like the explicit nature in the communication-react folder (as currently we have an `@internal` being exported -- presumably unintentionally??)

prprabhu-ms:
I realized that chromatic links don't show internal stories. That's less than ideal. Any way we could get the links there as well?

JamesBurnside:
> I realized that chromatic links don't show internal stories. That's less than ideal. Any way we could get the links there as well?

They actually do unintentionally which I'm fixing, for storybook builds we can set the env variable to allow it - I'll do a PR now fixing that

PR #1361, opened 2022-01-25
Add required peer dependencies to sample

prprabhu-ms:
# Why

These two ACS SDKs are now `peerDependencies` that must be installed by the app.
Before this PR, the build accidentally worked because `rush` links `node_modules/@azure/communication-react` to `packages/communication-react` which has the ACS SDKs installed via `devDependencies`. It ain't right though ;-)


JamesBurnside (approved):

anjulgarg:
Slightly off topic question, have we updated our docs to highlight to customers that they need to install these acs sdk separately now?

anjulgarg (approved):

alkwa-msft (approved):

PR #1368, opened 2022-01-25
File Upload Button

anjulgarg:
# What
Add a file upload button to the chat composite based on the figma below
https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?node-id=3970%3A22690

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2671542
https://skype.visualstudio.com/SPOOL/_workitems/edit/2671557

# How Tested

# Process & policy checklist


JamesBurnside:
(Probably a seperate PR or file a bug to be picked up as we have other places with missing icons) Icon is missing in the ui snapshot:
image

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -87,6 +90,7 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
   const typingIndicatorProps = usePropsFor(TypingIndicator);
   const headerProps = useAdaptedSelector(getHeaderProps);
   const errorBarProps = usePropsFor(ErrorBar);
+  const fileUploadButtonProps = useSelector(fileUploadButtonSelector);
```
JamesBurnside, line 93: mark as beta here

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -130,6 +134,7 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
               )}
             </div>
             <SendBox {...sendBoxProps} autoFocus={options?.autoFocus} styles={sendBoxStyles} />
+            <FileUploadButton {...fileUploadButtonProps} />
```
JamesBurnside, line 137: mark as beta here

packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
```diff
@@ -97,16 +90,15 @@ export const messageThreadChatCompositeStyles: MessageThreadStyles = { root: { m
 /**
  * @private
  */
-export const typingIndicatorChatCompositeStyles: TypingIndicatorStylesProps = {
-  typingString: { maxWidth: MESSAGE_THREAD_WIDTH }
```
JamesBurnside, line 101: Where you able to confirm that participants who have ludicrously long names have their names wrap correctly and not cause scrollbars or anything?

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
prprabhu-ms: `ChatScreen` is a fairly large component. I'd recommend avoiding any conditional compilation in this component.

Instead, conditionally compile inside `FileUploadButton` to return `<></>` on stable flavor builds. It'll likely end up being cleaner.

prprabhu-ms: Also, shouldn't this be `useAdapterSelector`?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+ */
+export interface FileUploadButtonProps {
+  /**
+   * The ACS user ID of the user uploading the file.
+   */
+  userId: CommunicationIdentifierKind;
+  /**
+   * The function of type {@link FileUploadHandler} for handling file uploads.
+   */
+  fileUploadHandler?: FileUploadHandler;
```
prprabhu-ms, line 21: You can define and use `FileUploadHandler` internally just fine. You'd conditionally export it through the public API only in beta flavor.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * A string containing the comma separated list of accepted file types.
+   * Similar to the `accept` attribute of the `<input type="file" />` element.
+   */
+  accept?: string;
+  /**
+   * Allows multiple files to be selected if set to `true`.
+   * Default value is `false`.
```
prprabhu-ms, line 29: Worth documenting here has well that this is similar to `multiple` attribute of `<input type="file" />`

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const { userId, fileUploadHandler, accept, multiple = false } = props;
+
+  const onChange = (files: FileList | null): void => {
+    if (!files) {
+      return;
+    }
+
+    const uploadedFiles: UploadedFile[] = [];
+
+    for (const file of files) {
```
prprabhu-ms, line 49: nit: This can simply be `const uploadedFiles = files.map(file => new UploadedFile(file));`

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * @internal
+ */
+export const FileUploadButton = (props: FileUploadButtonProps): JSX.Element => {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const theme = useTheme();
+  const { userId, fileUploadHandler, accept, multiple = false } = props;
+
+  const onChange = (files: FileList | null): void => {
```
prprabhu-ms, line 42: Use `useCallback` to cache this lambda.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+    for (const file of files) {
+      const uploadedFile = new UploadedFile(file);
+      uploadedFiles.push(uploadedFile);
+    }
+
+    fileUploadHandler && fileUploadHandler(userId, uploadedFiles);
+  };
+
+  const fileUploadButtonClassName = mergeStyles({
+    width: '1.5rem',
```
prprabhu-ms, line 58: We recommend using file-level styles constants for all these. If there are more than a handful of style objects, move them to `FileUploadButton.styles.ts`

packages/react-composites/src/composites/ChatComposite/styles/Chat.styles.ts
anjulgarg: Good point. Let me test that. \m/

anjulgarg: Yes, it's wrapping properly :)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <Icon iconName="Attach" />
+      </Stack>
+      <input
+        ref={inputRef}
+        hidden
+        multiple={multiple}
+        accept={accept}
+        type="file"
+        onChange={(e) => {
+          e.stopPropagation();
```
prprabhu-ms, line 84: Please add a comment explaining why we need to `stopPropagation`.

packages/react-composites/src/composites/ChatComposite/file-sharing/UploadedFile.tsx
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * File name to be displayed.
+   */
+  name: string;
+  /**
+   * Extension is used for rendering the file icon.
+   * An unknown extension will be rendered as a generic icon.
+   * Example: `jpeg`
+   */
+  extension: string;
```
prprabhu-ms, line 19: If we only render a known list of extensions, we should use a string literal type here `extension?: 'jpeg' | 'png'`

packages/react-composites/src/composites/ChatComposite/file-sharing/UploadedFile.tsx
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * @beta
+ */
+export const UPLOAD_CANCELLED_EVENT = 'uploadCancelled';
+
+/**
+ * Events emitted by the UploadedFile class.
+ * @beta
+ */
+export type UploadedFileEvents =
```
prprabhu-ms, line 47: Why is this not simply `UploadedFileEvent = 'uploadFailed' | 'uploadCancelled'` etc.

Actually, I'd shorten the values to `UploadedFileEvent` = 'success' | 'cancelled' | 'failed'` etc.

packages/react-composites/src/composites/ChatComposite/hooks/useSelector.ts
```diff
@@ -11,7 +11,7 @@ import { useSelectorWithAdaptation } from './useAdaptedSelector';
  */
 export const useSelector = <SelectorT extends (state: ChatAdapterState, props: any) => any>(
   selector: SelectorT,
-  selectorProps: Parameters<SelectorT>[1]
+  selectorProps?: Parameters<SelectorT>[1]
```
prprabhu-ms, line 14: Why? This doesn't seem like a change that should be part of this PR at all.

prprabhu-ms (commented):
Didn't quite finish the review, but pointed out some conditional compilation options.

packages/react-composites/src/composites/ChatComposite/hooks/useSelector.ts
anjulgarg: selectorProps needs to be an optional prop.
It is an optional prop in call composite hooks as well.
This is a part of this PR since File Upload Button uses a selector using `useSelector` that can only be used after fixing this bug.

packages/react-composites/src/composites/ChatComposite/file-sharing/UploadedFile.tsx
anjulgarg: To avoid bugs caused due to spelling mistakes.
Using constants eliminates that.
I am against using generic event names. The code becomes confusing when other objects subscribe to these events and you have no idea what a `success` or a `failed` means in that context.

packages/react-composites/src/composites/ChatComposite/file-sharing/UploadedFile.tsx
anjulgarg: The list is huge. And is dependent on the underlying Fluent UI.
Even Fluent UI doesn't have a documented list of supported icons.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
anjulgarg: We don't. Gonna remove it.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
anjulgarg: Gonna move it out to it's own files if we have more styles for this component. Currently, it seems overkill to move out 6 lines of styling code to it's own file.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
anjulgarg: Didn't get you

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
anjulgarg: This selector is only return userId and not anything related to the StatefulChatClient State.
useSelector should be fine here.
Correct me if I am wrong @PorterNan ?

anjulgarg: @prprabhu-ms Can I incorporate if/else in conditional compilation?
I don't fully understand how I can do what you are suggesting.
Is this a good example
image

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
anjulgarg: Don't see the point. `onChange` will only be called when files change...

dmceachernmsft (approved):

packages/communication-react/review/communication-react.api.md
```diff
@@ -1047,13 +1050,10 @@ export const DEFAULT_COMPOSITE_ICONS: {
     NoticePageJoinCallFailedDueToNoNetwork: JSX.Element;
     NoticePageLeftCall: JSX.Element;
     NoticePageRemovedFromCall: JSX.Element;
+    AttachFile: JSX.Element;
```
prprabhu-ms, line 1053: This name is too generic. We might have other places where we attach a file in the future. Could we name this `MessageThreadAttachFile` or something with a bit more context?

prprabhu-ms (commented):
Looks like we lost the left-alignment of `SendBox` and the typing indicator text when file attach button is missing:
image

Also, why do the snapshots for `ErrorBar` tests not have the file attach button?

prprabhu-ms:
Great PR!

Thanks for the hard work on figuring out what features should go into the stable channel vs not.

packages/communication-react/review/communication-react.api.md
anjulgarg: > Looks like we lost the left-alignment of `SendBox` and the typing indicator text when file attach button is missing: image
> 
> Also, why do the snapshots for `ErrorBar` tests not have the file attach button?

Filed a bug for the icon not rendering on some pages. It's a weird bug that @JamesBurnside  also noticed.

I am unable to find this Ui snapshot that you referenced here...
What's the name of the file?

anjulgarg: Found the image, fixed itr

prprabhu-ms (approved):

PR #1369, opened 2022-01-25
Center send button no matter how big the container is 

carocao-msft:
# What
Center the send button when send box is in multiple-line mode

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2689288

# How Tested
Screen Shot 2022-01-25 at 3 17 46 PM


JamesBurnside (approved):

change/@internal-react-components-a2357b20-610f-4d92-929a-b6507db829f2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Center send button",
```
prprabhu-ms, line 3: nitty nitty nit: Would be nice to say "vertically center send button in multi-line text box".

The comment should make sense standalone in a CHANGELOG.

prprabhu-ms (approved):

PorterNan (commented):
Ohh nice! I thought we might need a test to ensure this is always good for multiple lines but looks like this is already covered by this snapshot:
image

packages/react-components/src/components/styles/InputBoxComponent.style.ts
```diff
@@ -102,7 +102,8 @@ export const inlineButtonsContainerStyle: IStyle = {
   right: '0.3rem',
   top: '0',
   bottom: '0',
-  gap: '0.25rem'
+  gap: '0.25rem',
+  alignItems: 'center'
```
PorterNan, line 106: This is a shared css style across sendbox and editbox (when you edit a message), could you also double check the edit box button style just in case?

PorterNan (approved):
Looks great! Please double check if there is any css conflict for editbox (hover the mouse then edit), then we are good to go!

PR #1371, opened 2022-01-26
Injecting custom buttons to sidepane buttons in meeting composite

edwardlee-msft:
# What
Injecting custom buttons into meeting composite with storybook updated to reflect changes

# Why
Custom button injection feature support for MeetingComposite
https://skype.visualstudio.com/SPOOL/_workitems/edit/2686676/

# How Tested
Local test on meeting sample.

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -131,7 +131,11 @@ export type CustomCallControlButtonPlacement =
   | 'afterMicrophoneButton'
   | 'afterOptionsButton'
   | 'afterParticipantsButton'
-  | 'afterScreenShareButton';
+  | 'afterScreenShareButton'
+  // Sidepane placements
+  | 'beforeChatButton'
+  | 'afterChatButton'
+  | 'afterPeoplesButton';
```
JamesBurnside, line 138: nit peoples -> people
```suggestion
  | 'afterPeopleButton';
```

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -131,7 +131,11 @@ export type CustomCallControlButtonPlacement =
   | 'afterMicrophoneButton'
   | 'afterOptionsButton'
   | 'afterParticipantsButton'
-  | 'afterScreenShareButton';
+  | 'afterScreenShareButton'
+  // Sidepane placements
+  | 'beforeChatButton'
```
JamesBurnside, line 136: This needs to be a new type called CustomMeetingControlButtonPlacement that only applies to meeting composite instead of adding onto CustomCallControlButtonPlacement

prprabhu-ms: The biggest open question here is the handling of `'start'` and `'end'`. With the control bar and side pane being separate, what does `'end'` mean?

Perhaps the new type should prefix all values with the container name: `'controlBar.first'`, `'controlBar.afterCameraButton'`... and `'sidePane.afterChatButton'` etc.

This should also let us avoid `beforeChatButton` which is an outlier.

Thoughts @JamesBurnside

JamesBurnside: I think a sub sections make most sense.

`sidePane.<>` is a bit of a misnomer as the buttons are still in the controlbar just on the right hand side - and these two we have just happen to control the sidePane (they don't have to control a sidePane).

Do we know what overflow menu will look like - it could follow the same pattern, something like:
`CustomMeetingControlButtonPlacement = {
  controlBarCenter: { first, afterCameraBtn, afterMicBtn ... },
  controlBarRight: { first, afterChatBtn, afterPeopleBtn ... },
  overflow: { first, afterStartRecording ... }
}

To complicate further - mobile has a different layout again (it has no right-hand-side section)

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -133,6 +133,26 @@ export type CustomCallControlButtonPlacement =
...
+ * been hidden via a {@link CallControlOptions} field.
+ *
+ * Multiple buttons placed in the same position are appended in order.
+ * E.g., if two buttons are placed 'first', they'll both appear on the left end (right end in rtl mode)
+ * in the order provided.
+ *
+ * @beta
+ */
+export type CustomMeetingControlButtonPlacement = 'first' | 'last' | 'afterChatButton';
+
```
edwardlee-msft, line 155: FYI: I did extract out the sidepane custom button placements here. @JamesBurnside and @prprabhu-ms.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
edwardlee-msft: I like this layout as well. 
In the meantime, I will reimplement it like the 2nd example James provided.

edwardlee-msft: Would the pattern above be possible with type? If so, how would placement be set in <FilteredCustomButtons customButtonProps={customButtonProps} placement={} />

prprabhu-ms: For overflow, we'll allow adding a new section in the context menu (and all buttons go in there). Also those `menuItem`s will not be `ControlBarButton` subtypes. So I'm thinking of keeping that a separate API surface, something like:

`options.overflowButton.customSectionMenuItems` -- we decide whether that sections shows up and Contoso decides entirely what the `menuItem`s are.

I like the idea above of calling the two places `controlBarCenter` and `controlBarEnd` (*not* `controlBarRight` because that is wrong in RTL mode).

As for the placement type itself, do we really need a nested object? Or can we construct a flat string literal union type as I originally suggested: `'controlBarCenter.start' | 'controlBarCenter.afterCallEndButton | ... 'controlBarEnd.start' | 'controlBarEnd.afterPeopleButton'...`

I think the flat enum will be easier to use for Contoso (autocomplete etc will "just work" without having to understand any nesting structures).

prprabhu-ms: @edwardlee-msft we can talk about how to do this with typescript once we agree on the API. A lot of string magic is possible in typescript types. See for example: https://github.com/Azure/communication-ui-library/blob/main/packages/calling-stateful-client/src/TypeAssertions.ts that ensures that some string literal unions we define are consistent with other API from the SDK. We can then manipulate the incoming strings before passing them on to the `CallComposite`.

JamesBurnside: Idm between flat vs nested, but for mobile this will still be difficult, e.g. for desktop the chat button is on the `controlBarEnd` but on mobile it is on the `controlBarCenter` - (perhaps we should have a quick meeting to discuss this - seems like a difficult problem)

prprabhu-ms (commented):
Let's reparent this PR to `main` (The branch `prprabhu/meeting-inject-buttons` is defunct now. You've been owning this feature for a while now anyway :) )

You can click the "Edit" in the PR, and then there will be a dropdown for the base branch. Change that to `main`. That will show us a meaningful diff to review.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
edwardlee-msft: As per our discussion, I have implemented the button placements using "before.." and "after.." relative to default buttons.

PR #1372, opened 2022-01-27
Fix the unexpeteced removal bug

PorterNan:
# What

Fix unexpected removal when the conditional compile comment is at the first line

Before:
https://astexplorer.net/#/gist/a16004ec3eacc8328e703bb8d0d4253e/cf07cb00f395dde1fad3401a01813d4998abe9ff
After:
https://astexplorer.net/#/gist/8b81b0f34c1b4b2c309910e18dc33baf/bbdadc85eaa23de83e7a4167f13f70099994b391
# Why

# How Tested

# Process & policy checklist


JamesBurnside (approved):

common/scripts/babel-conditional-preprocess.js
```diff
@@ -69,6 +69,10 @@ function Handle(path, annotations) {
         }
         if (removed) {
           comment.ignore = true;
+          // Comment will be inherit to next line even it is set to 'ignore'
+          // this will cause some unexpected removals of code
+          // clear the comment to ensure nothing gets wrong
+          comment.value = '';
```
prprabhu-ms, line 75: I'd recommend describing an example of what incorrect removal happens without this setting. It'll be very hard for a reader in the future to understand the implications of removing this value.

prprabhu-ms (approved):

PR #1373, opened 2022-01-27
Fix handler device type which breaks type check

PorterNan:
# What

Since sdk update the type of device it makes our Common<typeA, typeB> not working well, which is breaking our auto completion of related handlers

After the fix (those handlers weren't there in auto completion since last update of sdk):
image

There were actually 2 ways of fixing this problem - adjust our component parameter or adjust handler. The reason why we choose this one is "no breaking change policy":
we are only using id+name to set the right type of device, if we add more properties to the component parameters that means it is a breaking change, 
this PR is still asking for only id+name which is backward compatible (it doesn't matter if you provide more properties than expected)
# Why

# How Tested

# Process & policy checklist


dmceachernmsft (approved):

prprabhu-ms:
This sounds to me like we're bending over backwards to maintain compatibility where `@azure/communication-calling` already broke compatibility. Worse, we might be lying -- if users provides us with an object that only has `id` and `name`, the call to `DeviceManager.selectMicrophone` etc might fail because it needs the other fields in the argument, e.g., `isSystemDefault`.

prprabhu-ms (changes requested):
Let's make sure we understand why the underlying API changed, and whether this fix is safe (because the underlying API has changed...)

anjulgarg:
Curious to understand why a stable API was changed by the SDK team.
Do we have answers?

prprabhu-ms:
did we get a resolution on this discussion? Is this a breaking change we're going to _have_ to carry at some point?

PorterNan:
> did we get a resolution on this discussion? Is this a breaking change we're going to _have_ to carry at some point?

Yep, will send a new PR next week for this, closing this one for now

PR #1374, opened 2022-01-27
File Sharing UI State and Selectors [Please Review]

anjulgarg:
# What
Adds attributes & function to adapter UI state.
Modifies types for adding metadata to ChatMessage
Adds and updates selectors to accomodate file sharing

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2706335

# How Tested

# Process & policy checklist


packages/communication-react/stable-review/communication-react.api.md
```diff
@@ -645,6 +646,8 @@ export interface ChatMessage extends MessageCommon {
     // (undocumented)
     messageType: 'chat';
     // (undocumented)
+    metadata?: Record<string, string>;
```
alkwa-msft, line 649: are we expecting to expose these in stable potentially as well?

packages/react-composites/review/react-composites.api.md
```diff
@@ -753,6 +770,62 @@ export type TopicChangedListener = (event: {
...
+// @beta
+export type UploadCanceledListener = () => void;
+
+// @beta
+export type UploadCompleteListener = (metaData: FileMetaData) => void;
+
+// @beta
+export class UploadedFile implements UploadEvents {
+    constructor(file: File);
+    cancelUpload(): void;
```
alkwa-msft, line 794: should we just rename this to Cancel?

packages/react-composites/review/react-composites.api.md
```diff
@@ -753,6 +770,62 @@ export type TopicChangedListener = (event: {
...
+    completeUpload(metaData: FileMetaData): void;
```
alkwa-msft, line 795: should this just be markComplete(metaData: FileMetaData): void?

packages/react-composites/review/react-composites.api.md
```diff
@@ -753,6 +770,62 @@ export type TopicChangedListener = (event: {
...
+    extension(): string;
```
alkwa-msft, line 796: should extension should be a property?

packages/react-composites/review/react-composites.api.md
```diff
@@ -753,6 +770,62 @@ export type TopicChangedListener = (event: {
...
+    failUpload(message: string): void;
+    file: File;
+    // (undocumented)
+    isUploaded(): boolean;
+    metaData?: FileMetaData;
+    off(event: UploadedFileEvents, listener: UploadedFileEventListener): void;
+    on(event: UploadedFileEvents, listener: UploadedFileEventListener): void;
+    progress: number;
+    truncatedName(length?: number): string;
```
alkwa-msft, line 805: when we upload does the file name length matter or is it more about impacting the UI when we need to render it? I dont think we want to remove information on the upload when it makes more sense to truncate things when rendering the files if the names are too long

packages/react-composites/review/react-composites.api.md
```diff
@@ -753,6 +770,62 @@ export type TopicChangedListener = (event: {
...
+// @beta
+export type UploadCompleteListener = (metaData: FileMetaData) => void;
+
+// @beta
+export class UploadedFile implements UploadEvents {
+    constructor(file: File);
+    cancelUpload(): void;
+    completeUpload(metaData: FileMetaData): void;
+    extension(): string;
+    failUpload(message: string): void;
```
alkwa-msft, line 797: should this just be markFailed(message: string): void?

packages/react-composites/src/composites/ChatComposite/file-sharing/index.ts
```diff
@@ -0,0 +1,20 @@
...
+export * from './FileCard';
+export * from './FileCardGroup';
+export * from './FileUploadButton';
+export * from './FileUploadHandler';
+export * from './UploadedFile';
+
+/**
+ * @internal
+ */
+export const ACSFilesMetadataKey = 'acsfiles';
```
alkwa-msft, line 12: In general unless we feel strongly about this lets not use the abbreviation "acs"

alkwa-msft: if this is just within the scope of the chat composite then we can say something like

filesharing_metadata_key instead?

anjulgarg: Was hoping to name the key something that allows partners to quickly identify that it's an ACS specific key (or UI library specific. )
`filesharing_metadata_key` seems too generic?
Also introduces snake case, which is not a standard naming practice in JS world compared to snake case.

packages/communication-react/stable-review/communication-react.api.md
anjulgarg: Api.md files currently don't respect conditional compilation. This will not be exposed in the stable builds.
Please correct me if am wrong @prprabhu-ms

packages/react-composites/review/react-composites.api.md
anjulgarg: This function is only used render the file name in file cards.
we are not truncating the file names for storing them in state or metadata

anjulgarg:
@alkwa-msft good suggestions around attribute naming
@prprabhu-ms had similar comments on a different PR.
I will collate the feedback and make changes to a single PR soon.

packages/react-composites/review/react-composites.api.md
anjulgarg: Going forward we might need to do some non trivial logic for getting file extensions. Hence, the method. Moreover, it will be used internally.

packages/react-composites/src/composites/ChatComposite/adapter/ChatAdapter.ts
```diff
@@ -15,6 +17,20 @@ export type ChatAdapterUiState = {
...
+   * @beta
+   */
+  uploadedFiles?: UploadedFile[];
+  /* @conditional-compile-remove-from(stable): FILE_SHARING */
+  /**
+   * Marks all the `uploadedFiles` as completely uploaded.
+   * Allows a message to be sent after the upload is complete.
+   * @beta
+   */
+  uploadedFilesCompleted?: boolean;
```
JamesBurnside, line 33: instead of having an explicit boolean that must be maintained, can this be calculated from `uploadedFiles.<some accumulator>.isUploaded()`?

JamesBurnside: e.g. could `getUploadedFilesCompleted` be something like 
```
getUploadedFilesCompleted => uploadedFiles.map(file => file.isUploaded()).reduce((a, b) => a && b)
```

packages/react-composites/review/react-composites.api.md
```diff
@@ -753,6 +770,62 @@ export type TopicChangedListener = (event: {
...
+    file: File;
+    // (undocumented)
+    isUploaded(): boolean;
```
JamesBurnside, line 800: is seems slightly odd to have a public method `isUploaded()` when the user of this class has to do `completeUpload` for isUploaded to be true (i.e. class owner should have no need for this method as they had to mark the file as uploaded) -- perhaps make it a private method?

packages/react-components/stable-review/react-components.api.md
```diff
@@ -92,6 +92,8 @@ export interface ChatMessage extends MessageCommon {
     // (undocumented)
     messageType: 'chat';
     // (undocumented)
+    metadata?: Record<string, string>;
```
prprabhu-ms, line 95: Again, +1 to including this in stable releases already.

(This does mean that our next release will end up being a minor bump)

packages/react-composites/review/react-composites.api.md
```diff
@@ -318,16 +319,22 @@ export interface ChatAdapterThreadManagement {
...
     loadPreviousChatMessages(messagesToLoad: number): Promise<boolean>;
     removeParticipant(userId: string): Promise<void>;
-    sendMessage(content: string): Promise<void>;
+    sendMessage(content: string, options?: SendMessageOptions): Promise<void>;
     sendReadReceipt(chatMessageId: string): Promise<void>;
     sendTypingIndicator(): Promise<void>;
     setTopic(topicName: string): Promise<void>;
     updateMessage(messageId: string, content: string): Promise<void>;
+    // @beta
+    uploadFiles?: (uploadedFiles: UploadedFile[]) => void;
```
prprabhu-ms, line 328: I guess this PR needs a merge from the other one that renamed `UploadedFile`.

packages/react-composites/src/composites/ChatComposite/adapter/ChatAdapter.ts
```diff
@@ -15,6 +17,20 @@ export type ChatAdapterUiState = {
   // FIXME(Delete?)
   // Self-contained state for composite
   error?: Error;
+  /* @conditional-compile-remove-from(stable): FILE_SHARING */
+  /**
+   * Files being uploaded by a user in the current thread.
+   * Should be set to null once the upload is complete.
+   * @beta
+   */
+  uploadedFiles?: UploadedFile[];
```
prprabhu-ms, line 26: In-line with my comment about the interface used by the file upload callback, this should not store the full `UploadedFile` object. It should _only_ contain the information that is needed in the adapter.

My assumption (right now) is that what is needed in the adapter state is:
* For each active file transfer, the name, extension (for icon), progress of the file transfer (i.e., read-only data).

I'd recommend that we create a new type that only contains those fields.

It can't directly use `UploadedFile` anyway, because all adapter state must be immutable (but `UploadedFile` is a mutable class).

packages/react-composites/src/composites/ChatComposite/adapter/ChatAdapter.ts
```diff
@@ -87,6 +103,18 @@ export interface ChatAdapterThreadManagement {
    *
    */
   loadPreviousChatMessages(messagesToLoad: number): Promise<boolean>;
+  /* @conditional-compile-remove-from(stable): FILE_SHARING */
+  /**
+   * Sets the uploaded files in ui state.
+   * @beta
+   */
+  uploadFiles?: (uploadedFiles: UploadedFile[]) => void;
```
prprabhu-ms, line 111: This method name is confusing. The adapter doesn't take any action to upload files. Should be more like `registerActiveFileUpload`.

The current setup isn't quite right -- we can't inject mutable objects into the adapter state. The adapter state is an immutable refrentially stable object.

See how `CallComposite` deals with `isLocalMicrophoneEnabled` -- `setIsLocalMicrophoneEnabled` is exposed as the only way to modify this field, and it uses `immer` to do it correctly.

prprabhu-ms: One possibility is that you add a method `registerActiveFileUpload` that *returns* an object with the interface passed to the Contoso callback. Internally, the adapter Context subscribes to events on this object to immutably update the adapter (ui) state.

packages/react-composites/src/composites/ChatComposite/adapter/ChatAdapter.ts
anjulgarg: Been thinking the same thing @JamesBurnside  :D
Good suggestion.

packages/react-composites/review/react-composites.api.md
anjulgarg: This PR is stale now, let me update it with the latest changes and this will be resolved :)

packages/react-composites/src/composites/ChatComposite/adapter/ChatAdapter.ts
```diff
@@ -87,6 +103,18 @@ export interface ChatAdapterThreadManagement {
...
+  /* @conditional-compile-remove-from(stable): FILE_SHARING */
+  /**
+   * Marks the uploaded files in ui state as complete.
+   * @beta
+   */
+  uploadsComplete?: () => void;
```
prprabhu-ms, line 117: I would recommend against this method. Each file transfer completes on its own, the UI should decide when to stop showing the file transfer UI / reenable chat messages based on whether all file transfers are complete.

Adding this method creates two ways of saying when all transfer are done (one via each individual `UploadedFile`, and one here). This can easily cause confusion (what happens if I call this before all files are transferred? What happens if I call this then call `uploadFiles` again?)

packages/chat-component-bindings/stable-review/chat-component-bindings.api.md
```diff
@@ -39,7 +40,7 @@ export type ChatClientProviderProps = {
 
 // @public
 export type ChatHandlers = {
-    onSendMessage: (content: string) => Promise<void>;
+    onSendMessage: (content: string, options?: SendMessageOptions) => Promise<void>;
```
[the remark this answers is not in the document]
anjulgarg: Right, and that's what's happening in this PR. I have not removed this change using conditional compilation :)
OR marked it as @beta

packages/react-composites/src/composites/ChatComposite/adapter/ChatAdapter.ts
anjulgarg: I implemented `setIsLocalMicrophoneEnabled` :D and even I don't remember it. Haha

packages/react-composites/package.json
```diff
@@ -64,7 +64,9 @@
     "react-aria-live": "^2.0.5",
     "react-linkify": "^1.0.0-alpha",
     "reselect": "~4.0.0",
-    "uuid": "^8.1.0"
+    "uuid": "^8.1.0",
+    "nanoid": "~3.1.20",
```
JamesBurnside, line 68: alphabetize

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -34,6 +19,11 @@ export interface FileUploadButtonProps {
    * Similar to the `multiple` attribute of the `<input type="file" />` element.
    */
   multiple?: boolean;
+
+  /**
+   * onChange handler for the file upload button.
```
JamesBurnside, line 24: Can you say in this comment what is changing in this onChange? (I'm confused how a Button has an `<input>` field)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -80,7 +61,7 @@ export const FileUploadButton = (props: FileUploadButtonProps): JSX.Element => {
         accept={accept}
```
JamesBurnside, line 61: the stack this is in doesn't have a tabindex, is it keyboard navigatable? (a11y work could also be a seperate work item for this all)

packages/react-composites/src/composites/ChatComposite/file-sharing/index.ts
```diff
@@ -0,0 +1,20 @@
...
+export * from './FileCard';
+export * from './FileCardGroup';
+export * from './FileUploadButton';
+export * from './FileUploadHandler';
+export * from './FileUpload';
+
+/**
+ * @internal
+ */
+export const FileSharingMetadataKey = 'fileSharingMetadata';
```
JamesBurnside, line 12: this isn't used outside this file, does it need exported?

packages/react-composites/src/composites/ChatComposite/file-sharing/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Metadata used for setting uploaded files by a user using chat composite in a group call.
+ * @internal
+ */
+export type FileSharingMetadata = {
```
JamesBurnside, line 18: this type seems odd to me, it's of type FileSharingMetadata, which is an object that has a key equal to `fileSharingMetadata`, isn't the file sharing metadata just the unkeyed string?

packages/react-composites/src/composites/ChatComposite/selectors/baseSelectors.ts
```diff
@@ -2,8 +2,34 @@
...
+    return undefined;
+  }
+  return Object.keys(fileUploads).map((key) => fileUploads[key]);
+};
+
+/* @conditional-compile-remove-from(stable): FILE_SHARING */
+/**
+ * @private
+ */
+export const getFileUploadsCompleted = (state: ChatAdapterState): boolean | undefined => {
```
JamesBurnside, line 29: `getFileUploads` returns an array of files, I'd expect `getFileUploadsCompleted` to return a subset of all files uploaded that are completed.

Can this be renamed to `getHaveAllFilesFinishedUploading`
Or if our selectors don't need to be named `get` just  `haveAllFilesFinishedUploading`

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
```diff
@@ -0,0 +1,158 @@
...
+    this.context.deleteFileUpload(id);
+    const fileUpload = this.findFileUpload(id);
+    if (!fileUpload) {
+      throw new Error('File upload not found');
+    }
+    this.unsubscribeAllEvents(fileUpload);
+    this.deleteFileUpload(id);
+  }
+
+  getFileSharingMetada(): FileSharingMetadata {
```
JamesBurnside, line 120: typo: `getFileSharingMetada` --> `getFileSharingMetadata`

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
```diff
@@ -0,0 +1,158 @@
...
+        delete draft?.fileUploads?.[id];
+      })
+    );
+  }
+}
+
+/**
+ * @internal
+ */
+export class FileSharingAdapter implements FileUploadAdapter {
```
JamesBurnside, line 83: Love this class!

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
```diff
@@ -0,0 +1,158 @@
...
+export interface FileUploadAdapter {
+  registerFileUploads?: (fileUploads: ObservableFileUpload[]) => void;
+  clearFileUploads?: () => void;
+  cancelFileUpload?: (id: string) => void;
+}
+
+/**
+ * @internal
+ */
+class FileSharingContext {
```
JamesBurnside, line 27: love this class!

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -137,6 +137,15 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
   const typingIndicatorStyles = Object.assign({}, styles?.typingIndicator);
   const sendBoxStyles = Object.assign({}, styles?.sendBox);
 
+  const fileUploadButtonOnChange = (files: FileList | null): void => {
+    if (!files) {
+      return;
+    }
+    const fileUploads = Array.from(files).map((file) => new FileUpload(file));
+    adapter.registerFileUploads && adapter.registerFileUploads(fileUploads);
```
JamesBurnside, line 145: If I upload 2 files, they finish uploading, then click to upload a third -- does this get all three files? If so do the registerFileUploads handle the two that have already been uploaded or will the developer be expected to call updateFileUpload to set as complete?

JamesBurnside:
This looks awesome! Just some minor questions but nothing blocking.

packages/react-composites/package.json
anjulgarg: You got it boss! :P

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
anjulgarg: For sure.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
anjulgarg: Not thinking about a11y work at this point. Will be tackling the entire file sharing UI from an a11y perspective later. :)

packages/react-composites/src/composites/ChatComposite/file-sharing/index.ts
anjulgarg: Good catch. Will remove the export

packages/react-composites/src/composites/ChatComposite/file-sharing/index.ts
anjulgarg: Right. I understand how it can be confusing to read the type.
Although an argument can also be made the the metadata is an object of the form `{[key]: string}`

packages/react-composites/src/composites/ChatComposite/selectors/baseSelectors.ts
anjulgarg: Makes sense. Will change the name

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
anjulgarg: Good catch.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
anjulgarg: Great point. Let me think about that. I don't think the current designs factors in that use case.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -0,0 +1,217 @@
...
+export interface FileUploadState {
+  /**
+   * Unique identifier for the file upload.
+   */
+  id: string;
+
+  /**
+   * Filename extracted from the {@link File} object.
+   */
+  filename: string;
```
prprabhu-ms, line 39: This duplicates `metadata.name`. If the intent is somehow different, please document how these two fields differ.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -0,0 +1,217 @@
...
+  failUpload: (message: string) => void;
+}
+
+/**
+ * An internal interface used by the Chat Composite to drive the UI for file uploads.
+ * @beta
+ */
+export interface ObservableFileUpload extends FileUploadEventEmitter {
+  /**
+   * UNiwue identifier for the file upload.
```
prprabhu-ms, line 94: nit: typo "Unique"

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * A record containing {@link FileUploadState} mapped to unique ids.
+ * @beta
+ */
+export type FileSharingUiState = Record<string, FileUploadState>;
+
+/**
+ * @beta
+ */
+export interface FileUploadAdapter {
```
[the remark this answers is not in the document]
prprabhu-ms: e.g., this file is called `FileSharingAdpater`, but the interface is `FileUploadAdpater`, which manages the `FileSharingUiState` (a record of `FileUploadState`) and uses `FileSharingContext`.

prprabhu-ms, line 18: We now expose two programmatic interfaces related to file upload -

1. The Composite options that provide the file upload callback
2. The FileUploadAdapter API

They serve two distinct purposes. (1) is the primary way Contoso provides an implementation for uploading files. (2) is an extension point (like all other adapter API) where Contoso can modify how composite interacts with the business logic of file upload.

This dinstinction is subtle, and will require careful developer education, in my opinion.

---

I like this design a lot. It is in line with existing patterns in our library. Just noting a tripping point for Contoso devs here.

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
```diff
@@ -0,0 +1,158 @@
...
+  registerFileUploads?: (fileUploads: ObservableFileUpload[]) => void;
+  clearFileUploads?: () => void;
+  cancelFileUpload?: (id: string) => void;
```
prprabhu-ms, line 21: I'm guessing this will be called from the UI when user clicks on a "cancel upload" button?

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
anjulgarg: Made a slight change in the `registerFileUploads` method and now this scenario should work
image

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -0,0 +1,217 @@
...
+ * @beta
+ */
+export interface FileUploadManager {
+  /**
+   * HTML {@link File} object for the uploaded file.
+   */
+  file: File;
+  /**
+   * Update the progress of the upload.
+   * Emits the `uploadProgressed` event.
```
prprabhu-ms, line 70: Don't reference the `uploadProgressed` and other events in this interface at all. It's an implementation detail of `class FileUpload` that this triggers the said event.

For Contoso, the more we can do to keep the `FileUploadManager` and `ObservableFileUpload` interfaces separate, the better. (So audit any references across these two interfaces and remove them).

packages/react-composites/src/composites/ChatComposite/file-sharing/index.ts
anjulgarg: Removed the const declaration for the key.
Should make things less confusing now?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -0,0 +1,217 @@
...
+   * Subscriber function for `uploadFailed` event.
+   */
+  on(event: 'uploadFailed', listener: UploadFailedListener): void;
+
+  /**
+   * File upload event subscriber.
+   * @param event - {@link FileUploadEvents}
+   * @param listener - {@link FileUploadEventListener}
+   */
+  on(event: FileUploadEvents, listener: FileUploadEventListener): void;
```
prprabhu-ms, line 197: This generic method shouldn't be part of the interface. It is only needed for the internal `class FileUpload` implementation.

That means that we can also remove the union types `FileUploadEvents` and `FileUploadEventListener` from the public API.

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -29,9 +29,13 @@ import {
   TopicChangedListener
 } from './ChatAdapter';
 import { AdapterError } from '../../common/adapters';
+/* @conditional-compile-remove-from(stable): FILE_SHARING */
+import { FileSharingAdapter } from './FileSharingAdapter';
+/* @conditional-compile-remove-from(stable): FILE_SHARING */
+import { ObservableFileUpload } from '../file-sharing';
 
 /** Context of Chat, which is a centralized context for all state updates */
```
prprabhu-ms, line 37: Can you mark this `@private` so that  it's not accidentally exported in the future.

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -92,6 +102,8 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
   private chatClient: StatefulChatClient;
   private chatThreadClient: ChatThreadClient;
   private context: ChatContext;
+  /* @conditional-compile-remove-from(stable): FILE_SHARING */
+  private fileSharingAdapter: FileSharingAdapter;
```
prprabhu-ms, line 106: This should store a `FileUploadAdapter` (the public interface) not the `FileSharingAdapter` so that it can only use the public API. **If we expose `FileUploadAdapter` as public API.

And that is a good question: Do we want to expose `FileUploadAdapter` as a public API at all? There are already new methods in `ChatAdapter` that Contoso can override. `FileUploadAdapter` can remain internal implementation for us, until we feel the need to export that separately. If we do export it,

* We will need to provide a way for dependency injection so Contoso can provide a different implementation of `FileUploadAdpater` and also a way to construct the `FileSharingAdapter`
* Also, the naming can be a bit confusing here. We should rename `FileSharingAdapter` to `AzureCommunicationFileUploadAdapter`, to follow the established pattern.


I think we should unexport `FileUploadAdapter` right now. At that point, we don't a separate interface at all (Just use `FileSharingAdapter` (renamed if needed) here).

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -168,9 +188,15 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
     this.context.offStateChange(handler);
   }
 
-  async sendMessage(content: string): Promise<void> {
+  async sendMessage(content: string, options: SendMessageOptions = {}): Promise<void> {
     await this.asyncTeeErrorToEventEmitter(async () => {
-      await this.handlers.onSendMessage(content);
+      /* @conditional-compile-remove-from(stable): FILE_SHARING */
+      options.metadata = { ...options.metadata, ...this.fileSharingAdapter.getFileSharingMetadata() };
```
prprabhu-ms, line 194: See my comment above - `getFileSharingMetadata` is not part of the `FileUploadAdapter` public interface, but is used here.

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
```diff
@@ -0,0 +1,158 @@
...
+    this.context.deleteFileUpload(id);
+    const fileUpload = this.findFileUpload(id);
+    if (!fileUpload) {
+      throw new Error('File upload not found');
+    }
+    this.unsubscribeAllEvents(fileUpload);
+    this.deleteFileUpload(id);
+  }
+
+  getFileSharingMetadata(): FileSharingMetadata {
```
prprabhu-ms, line 120: Why is this needed? This information should be available from the `state`.

packages/react-composites/src/composites/ChatComposite/selectors/baseSelectors.ts
```diff
@@ -2,8 +2,34 @@
...
 /**
  * @private
  */
 export const getUserId = (state: ChatAdapterState): CommunicationIdentifierKind => state.userId;
+
+/* @conditional-compile-remove-from(stable): FILE_SHARING */
+/**
+ * @private
+ */
+export const getFileUploads = (state: ChatAdapterState): FileUploadState[] | undefined => {
```
prprabhu-ms, line 17: These selectors use the `state` directly. This means they'll run on *every render*. But file upload state is unlikely to change that frequently.

I'd recommend that you add a selector to get just the file upload state `state?.fileUploads` and have these depend on that selector (so they won't run most of the time).

prprabhu-ms:
[done with my in-parts review]. This PR froze edge twice :D

packages/communication-react/review/communication-react.api.md
```diff
@@ -663,6 +666,8 @@ export interface ChatMessage extends MessageCommon {
     // (undocumented)
     messageType: 'chat';
     // (undocumented)
+    metadata?: Record<string, string>;
```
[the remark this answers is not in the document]
anjulgarg: This is weird. I did document this field.
image

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
anjulgarg: Done

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
prprabhu-ms, line 18: I think edge also lost some of my comments.

Possibly repeating this one: We're currently using "file upload" and "file sharing" inconsistently:

* This file is `FileSharingAdapter.ts`
* This interface is `FileUploadAdapter`, and is implemented by `FileSharingAdapter`
* The state is stored in `FileSharingUiState` that uses `FileUploadState`.

I think all of this API surface is for file upload only, so I'd recommend just sticking with "file upload" consistently in naming.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -0,0 +1,217 @@
...
+export class FileUpload implements FileUploadManager, ObservableFileUpload {
+  private _emitter: EventEmitter;
+  public id: string;
+  public file: File;
+
+  constructor(file: File) {
+    this.id = nanoid();
+    this.file = file;
+    this._emitter = new EventEmitter();
+    this._emitter.setMaxListeners(100);
```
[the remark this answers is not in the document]
anjulgarg: We are calling setMaxListeners in different packages.
The value will differ basis on the use cases.
Besides, any suggestions where this global const should sit?

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
anjulgarg: We can't unexport `FileUploadAdapter` as it is being extended by `ChatAdapter`
I think it's ok to not prefix these plugin adapters with `AzureCommunication` as it helps identify that `AzureCommunicationChatAdapter` is the main adapter and these are plugins(Extensions)

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
anjulgarg: Yes, I missed it. Good catch.

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -168,9 +188,15 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
...
+  async sendMessage(content: string, options: SendMessageOptions = {}): Promise<void> {
     await this.asyncTeeErrorToEventEmitter(async () => {
-      await this.handlers.onSendMessage(content);
+      /* @conditional-compile-remove-from(stable): FILE_SHARING */
+      options.metadata = { ...options.metadata, ...this.fileSharingAdapter.getFileSharingMetada() };
+
+      await this.handlers.onSendMessage(content, options);
+
+      /* @conditional-compile-remove-from(stable): FILE_SHARING */
+      this.fileSharingAdapter.clearFileUploads();
```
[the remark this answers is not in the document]
anjulgarg: Added comments describing the usage.

I understand the concern behind modifying a public API. 
I can think of a few scenarios where contoso might want to call this.
- If number of uploaded files exceeds the max number of allowed file by Contoso.
- If the file extensions are invalid.
- If contoso wants to implement a custom shortcut or `escape` key for clearing sendbox.

An argument can be made against other public functions in this adapter that they are useless to Contoso and are for internal implementation.
`sendTypingIndicator` `sendReadReceipt` etc for example.

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
anjulgarg: Yes.

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
anjulgarg: Right, I am gonna take it out of the adapter and make it a private util function in this file.
Doesn't need to be part of the adapter.

packages/react-composites/src/composites/ChatComposite/adapter/FileSharingAdapter.ts
anjulgarg: Haha yes, I agree.
Just realised this. Lemme wreak havoc on the code once again.

anjulgarg: Check it out. Renamed everything to be consistent

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -0,0 +1,217 @@
...
+  errorMessage?: string;
+}
+
+/**
+ * A wrapper object for a file that is being uploaded.
+ * Allows mmanaging file uploads by providing common functions for updating the
+ * upload progress, canceling an upload, completing an upload etc.
+ * @beta
+ */
+export interface FileUploadManager {
```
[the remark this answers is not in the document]
anjulgarg: I like that naming suggestion.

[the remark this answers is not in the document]
anjulgarg: Made the changes

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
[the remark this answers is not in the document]
prprabhu-ms: `acs-ui-common` package, exported as `@internal`.

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
prprabhu-ms: Ah, we're extending it. OK, that makes sense then, from an interface API perspective.

I was still thinking composition.

So `AzureCommunicationChatAdapter` implements `FileUploadAdapter` (due to `extends`) and there is no need for dependency injection.

And `FileSharingAdapter` need not implement `FileUploadAdapter` at all (it's just an internal implementation -- encapsulating the logic for file upload).

anjulgarg: Right, but it's good for ensuring there is no drift b/w FileSharingAdapter and FileUploadAdapter going forward.

anjulgarg: Any new functions added to the FileUploadAdapter need to be implemented in FileSharingAdapter.

JamesBurnside (approved):

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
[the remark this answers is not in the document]
anjulgarg: Done

prprabhu-ms (approved):

PR #1376, opened 2022-01-27
Fixup a11y properties of the PiP, allow secondary tile to be optional

JamesBurnside:
# What
Fixup a11y properties of the PiP
  * Use `aside` to indicate that the content within is tangential to the primary content as Picture-In-Picture content typically always is
  * Allow setting of an `aria-label`
  * Ensure the element is focusable by making use of `tabIndex` prop
  * Ensure the element can be clicked through keyboard using enter or space key

Also
  * Secondary tile is now optional

# Why
* Making components A11y friendly is important
* Secondary tile being optional allows for us to have the local participant take up the whole tile when there are no remote participants

# How Tested
Put into composite, could keyboard navigate to the picture in picture in picture and both enter and space worked for clicking it.


packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
```diff
@@ -48,16 +60,35 @@ export const _PictureInPictureInPicture = (props: _PictureInPictureInPictureProp
  */
 const PictureInPictureInPictureContainer = (props: {
   primaryView: ReactChild;
-  secondaryView: ReactChild;
+  secondaryView?: ReactChild;
```
anjulgarg, line 63: I don't understand why this is an optional prop now

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
```diff
@@ -48,16 +60,35 @@ export const _PictureInPictureInPicture = (props: _PictureInPictureInPictureProp
...
+  </aside>
 );
 
+/**
+ * For keyboard navigation - when this component has active focus, enter key and space keys should have the same behavior as mouse click.
+ */
+const handleKeyDown = (e: React.KeyboardEvent<HTMLElement>, onClickCallback: () => void): void => {
+  if (e.key === 'Enter' || e.key === ' ') {
+    // If the PiP is clicked through keyboard, perform the onClick behavior only.
+    e.preventDefault();
```
anjulgarg, line 85: what happens when we don't precent default behavior?

JamesBurnside: Actually this is just a div, so probably nothing, I'll remove it

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
JamesBurnside: From PR "why" desc 😉 'Secondary tile being optional allows for us to have the local participant take up the whole tile when there are no remote participants'

JamesBurnside: (It was bolted into this PR really from laziness as it was a pain to split out in the stack of PRs when I was splitting the big one up)

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
```diff
@@ -48,16 +60,35 @@ export const _PictureInPictureInPicture = (props: _PictureInPictureInPictureProp
...
   onClick?: () => void;
+  ariaLabel: string;
 }): JSX.Element => (
-  <div style={tileContainerStyles} onClick={props.onClick}>
+  <aside
+    style={tileContainerStyles}
+    onClick={props.onClick}
+    onKeyPress={(e) => props.onClick && handleKeyDown(e, props.onClick)}
+    aria-label={props.ariaLabel}
+    tabIndex={props.onClick ? 0 : -1} // Only allow focus to be set if there is a click handler
```
prprabhu-ms, line 72: <3

prprabhu-ms (approved):

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
JamesBurnside: removed

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
```diff
@@ -48,16 +60,33 @@ export const _PictureInPictureInPicture = (props: _PictureInPictureInPictureProp
  */
 const PictureInPictureInPictureContainer = (props: {
   primaryView: ReactChild;
-  secondaryView: ReactChild;
+  secondaryView?: ReactChild;
   onClick?: () => void;
+  ariaLabel: string;
 }): JSX.Element => (
-  <div style={tileContainerStyles} onClick={props.onClick}>
+  <aside
```
mgamis-msft, line 67: For my learning, why is `aside` used over `div` :)?
Nvm haha it's in the description

mgamis-msft (approved):

PR #1378, opened 2022-01-27
Chat Notification Icon Component

dmceachernmsft:
# What
Adds Chat message Notification icon

# Why
Allows for the display of unread messages to the user

https://skype.visualstudio.com/SPOOL/_workitems/edit/2698421
# How Tested
Simple test code to fake # of new messages to make sure displays properly


packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * @private
+ */
+export const NotificationIcon = (props: NotificationIconProps): JSX.Element => {
+  const { numberOfMessages } = props;
+  return (
+    <Stack horizontalAlign="center" verticalAlign="center" style={{ position: 'absolute', top: 0, right: '0.78rem' }}>
+      <Icon styles={NotificationIconStyles} iconName="NotificationCircle" />
+      <Stack styles={NotificationTextStyles}>
+        {numberOfMessages > 0 && numberOfMessages < 9 ? <>{numberOfMessages}</> : <></>}
+        {numberOfMessages > 9 ? <>+9</> : <></>}
```
JamesBurnside, line 24: `+9` or `9+`

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+export type NotificationIconProps = {
+  numberOfMessages: number;
+};
+/**
+ * @private
+ */
+export const NotificationIcon = (props: NotificationIconProps): JSX.Element => {
+  const { numberOfMessages } = props;
+  return (
+    <Stack horizontalAlign="center" verticalAlign="center" style={{ position: 'absolute', top: 0, right: '0.78rem' }}>
```
JamesBurnside, line 20: fluent recomendation is to use the `styles` prop instead of `style` (styles gets merged into a single classname that can be better reused, cached etc., style will directly put the css into the element in the html file)
https://github.com/microsoft/fluentui/wiki/Component-Styling

so this will become something like:
```suggestion
    <Stack horizontalAlign="center" verticalAlign="center" style={notificationIconContainerStyles}>
```
then at the bottom of the page as a *constant* (being a constant outside the render scope avoids creating a new object every render):
```tsx
const notificationIconContainerStyles = { position: 'absolute', top: 0, right: '0.78rem' }
```

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <Icon styles={NotificationIconStyles} iconName="NotificationCircle" />
+      <Stack styles={NotificationTextStyles}>
+        {numberOfMessages > 0 && numberOfMessages < 9 ? <>{numberOfMessages}</> : <></>}
```
JamesBurnside, line 23: nit: I would simplify this and deal with not-valid values seperately, so here have:
```
{ numberOfMessages < 9 ? <>{numberOfMessages}</> : ? <>9+</>
```

And then near the top of this function have the explicit case:
```
if (numberOfMessages < 1) return <></>
```

It helps make the render portion easier to read, and helps easily understand what invalid consitions will lead to: `<></>`

JamesBurnside: One extra pro of this is typescript will soon support narrowing of types! Meaning instead of 
```
  numberOfMessages: number;
```
we will be able to do:
```
  numberOfMessages: numberGreaterThanZero;
```

packages/react-composites/src/composites/MeetingComposite/ChatButton.tsx
```diff
@@ -15,13 +16,15 @@ export const ChatButton = (props: ControlBarButtonProps): JSX.Element => {
...
   return (
-    <ControlBarButton
-      {...props}
-      labelKey={'chatButtonLabelKey'}
-      strings={strings}
-      onRenderOnIcon={props.onRenderOnIcon ?? onRenderOnIcon}
-      onRenderOffIcon={props.onRenderOffIcon ?? onRenderOffIcon}
-      onClick={props.onClick}
-    />
+    <Stack style={{ position: 'relative' }}>
```
JamesBurnside, line 19: Add a comment why this is needed
```
// Chat Button may have an optional notification icon attached to it that must be positioned absolute inside the chat button.
// This requires the container of the notification icon to have `position relative`
```

JamesBurnside: use styles or style here also

JamesBurnside: How does the notification icon get passed into here? Would it better to simply have the notification icon defined here and be part of the ChatButton - I think it's a simple enough API for us to easily support and gives chat button a good bit of extra value

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -95,7 +95,7 @@ export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.El
         </CallAdapterProvider>
       </Stack.Item>
       {meetingCallControlOptions !== false && (
-        <Stack.Item>
+        <Stack horizontal>
```
JamesBurnside, line 98: can this be removed from this PR? What does this change do?

change/@internal-react-composites-3e8203bf-9f73-4300-8ae4-34b282a1e57a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Created new Notification Icon for the chat button in the meeting composite.",
```
anjulgarg, line 3: This changelog is for end-users.
So a tip is to avoid using terms like `created` and use `Add` or `Introduce` instead. ;)
Even `Notification Icon for the chat button in the meeting composite.` this looks good to me if I read it as an end user.

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
anjulgarg: Curious: I couldn't find any pros and cons of using `styles` over inline `style` on the link you shared.

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
anjulgarg: Another suggestion would be to abstract all this logic in a separate function and call that function here instead.
This logic might grow in the future and will make it easier to scale.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
dmceachernmsft: this ensures that the chat and people buttons are aligned horizontally with each other instead of getting stacked on top of one another

change/@internal-react-composites-3e8203bf-9f73-4300-8ae4-34b282a1e57a.json
dmceachernmsft: okay will update that!

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
dmceachernmsft: I matched it with how teams displayed it 😅

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
dmceachernmsft: okay will look at this!

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * @private
+ */
+export const NotificationIcon = (props: NotificationIconProps): JSX.Element => {
+  const { numberOfMessages } = props;
+
+  const renderNumber = (numberOfMessages): JSX.Element => {
+    if (numberOfMessages < 1) {
+      return <></>;
+    } else {
+      return <Stack styles={NotificationTextStyles}>{numberOfMessages < 9 ? <>{numberOfMessages}</> : <>+9</>}</Stack>;
```
prprabhu-ms, line 24: 1. You don't need `<></>` since these are already wrapped in a `Stack`.
2. btw, *why are they in a `Stack`*? A `Stack` only makes sense if you have more than one item. You should probably use a `Text` https://developer.microsoft.com/en-us/fluentui#/controls/web/text
3. Also, double check what aria labels to set so that announcer correctly handles the notification.

[Wrote this comment then realized it was three comments :P]

packages/react-composites/src/composites/MeetingComposite/styles/MeetingCompositeStyles.ts
```diff
@@ -18,3 +18,24 @@ export const compositeOuterContainerStyles: IStackStyles = {
     flexWrap: 'wrap'
   }
 };
+
+/**
+ * @private
+ */
+export const NotificationIconStyles: IIconStyles = {
+  root: {
+    color: '#0078D4'
```
prprabhu-ms, line 27: This should likely come from `theme`. Try dark theme in storybook and see what the notification icon looks like / should look like.

packages/react-composites/src/composites/common/icons.tsx
```diff
@@ -18,6 +18,8 @@ import { CameraSwitch24Filled } from '@fluentui/react-icons';
 import { ComponentIcons, DEFAULT_COMPONENT_ICONS } from '@internal/react-components';
 import React from 'react';
 import { mergeStyles, Text } from '@fluentui/react';
+/* @conditional-compile-remove-from(stable) */
```
prprabhu-ms, line 21: For all these conditional compilation directives, it's a good idea to use tags to group them (so we can find them easily later).

I've been using

```ts
/* conditional-compile-remove-from(stable) meeting-composite <further-details> */
```

for meeting related stuff.

packages/react-composites/src/composites/common/icons.tsx
```diff
@@ -48,6 +50,8 @@ export const COMPOSITE_ONLY_ICONS = {
   LocalPreviewPlaceholder: <VideoOff20Filled />,
   /* @conditional-compile-remove-from(stable) */
   LocalCameraSwitch: <CameraSwitch24Filled />,
+  /* @conditional-compile-remove-from(stable) */
+  notificationCircle: <Circle20Filled />,
```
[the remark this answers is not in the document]
dmceachernmsft: oops typo! how about NotificationIcon so we can re-use it for more than just chat if we want?

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
dmceachernmsft: will take a look at this!

dmceachernmsft: in reply to 2. they are in a stack so that the styles at the bottom of the page can be applied to them to fit within the icon appropriately.

dmceachernmsft: for 3. the aria labels will be added in another PR? I imagine that I will need to add strings to support the reading of other locales?

packages/react-composites/src/composites/MeetingComposite/styles/MeetingCompositeStyles.ts
JamesBurnside: yeah for this you will need to use `useTheme` and use `useTheme().palette.themePrimary`.
You must use useTheme inside the component render function however.

packages/react-composites/src/composites/MeetingComposite/styles/MeetingCompositeStyles.ts
```diff
@@ -18,3 +18,24 @@ export const compositeOuterContainerStyles: IStackStyles = {
...
+};
+
+/**
+ * @private
+ */
+export const NotificationTextStyles: IStackStyles = {
+  root: {
+    position: 'absolute',
+    top: '0.1rem',
+    color: 'white',
```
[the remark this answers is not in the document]
JamesBurnside: (likely `theme.palette.white`)

packages/react-composites/src/composites/MeetingComposite/ChatButton.tsx
```diff
@@ -15,13 +17,15 @@ export const ChatButton = (props: ControlBarButtonProps): JSX.Element => {
...
   return (
-    <ControlBarButton
-      {...props}
-      labelKey={'chatButtonLabelKey'}
-      strings={strings}
-      onRenderOnIcon={props.onRenderOnIcon ?? onRenderOnIcon}
-      onRenderOffIcon={props.onRenderOffIcon ?? onRenderOffIcon}
-      onClick={props.onClick}
-    />
+    <Stack styles={ChatButtonContainerStyles}>
```
JamesBurnside, line 20: can this style be directly passed into the ControlBarButton styles?
```tsx
<ControlBarButton
  ...
  styles={{ root: { position: absolute }} // <-- as a const to memoize it
/>
```

change/@internal-react-composites-3e8203bf-9f73-4300-8ae4-34b282a1e57a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
anjulgarg, line 2: Should be `prerelease`.

`minor` makes change to the stable build.

Correct me if I am wrong @JamesBurnside @prprabhu-ms

packages/communication-react/review/communication-react.api.md
```diff
@@ -284,6 +284,7 @@ export const CallComposite: (props: CallCompositeProps) => JSX.Element;
...
     LocalCameraSwitch?: JSX.Element;
+    NotificationIcon?: JSX.Element;
```
anjulgarg, line 287: Can we change the name to something not so generic?
Like `ChatButtonNotificationIcon`?

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
anjulgarg, line 98: Are we removing this as suggested by @JamesBurnside and @prprabhu-ms ?

packages/react-composites/src/composites/MeetingComposite/ChatButton.tsx
JamesBurnside: Discussed offline - change is related to when badge is added to the component

packages/communication-react/review/communication-react.api.md
dmceachernmsft: I called it something more generic so it could maybe be reused in the future for something. do we want to be more specific and maybe rename it in the future if we are going to use it somewhere else as well?

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
dmceachernmsft: I corrected this to the way the screenshot actually shows it not how I remembered it 😅

packages/communication-react/review/communication-react.api.md
JamesBurnside: The problem is "notification" is very loose - what is a notification -- this is actually a "badge" instead of a notification, e.g. https://mui.com/components/badges/#main-content

JamesBurnside: suggestion: ButtonBadgeIcon?

dmceachernmsft: I'll change it to that!

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -16,27 +15,51 @@ export type NotificationIconProps = {
...
+    root: {
+      // positioning to place the badge within the button appropriately.
+      position: 'absolute',
+      top: '-0.5rem',
+      right: '-0.5rem'
+    }
+  })
+);
+
+const NotificationIconStyles = memoizeFunction(
```
prprabhu-ms, line 48: nit: camelCase: `notificationIconStyles`

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -16,27 +15,51 @@ export type NotificationIconProps = {
...
+  (theme: ITheme): IIconStyles => ({
+    root: {
+      color: theme.palette.themePrimary
+    }
+  })
+);
+
+const NotificationTextStyles = memoizeFunction(
```
prprabhu-ms, line 56: nit: camelCase

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -16,27 +15,51 @@ export type NotificationIconProps = {
...
+  (theme: ITheme): IStackStyles => ({
+    root: {
+      position: 'absolute',
+      top: '0.1rem',
+      color: theme.palette.white,
+      fontSize: '0.8rem'
```
prprabhu-ms, line 62: `fontSize` should also come from the theme. '0.8rem', so you probably want `smallPlus`.

prprabhu-ms (approved):

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
dmceachernmsft: will Fix!

JamesBurnside (approved):

PR #1379, opened 2022-01-28
[Draft] Add command to move stable files and generate stable logs

PorterNan:
This is a draft shows how we bump version using beachball for stable:

- Add command to move stable files
- Update auto-install packages for rush

The main pain point and complexity is we need to maintain 2 logs and 2 folders of change entries - if we do bump manually - which we might prefer, it will help us decrease a lot of complexity here
# What

# Why

# How Tested

# Process & policy checklist


common/config/rush/command-line.json
```diff
@@ -31,12 +31,48 @@
...
+    },
+    {
+      "commandKind": "global",
+      "name": "bump-stable",
+      "summary": "Switch flavor to beta",
+      "description": "Every build and test command will run based on beta version.",
+      "autoinstallerName": "rush-file",
+      "shellCommand": "rush pre-stable-bump && node common/config/node_modules/beachball/bin/beachball bump && rush post-stable-bump",
+      "safeForSimultaneousRushProcesses": true
+    },
```
PorterNan, line 69: This is the main complexity we are introducing - we need to backup and restore the stable log and make it back

PorterNan:
Close this because we are now choosing another solution

PR #1380, opened 2022-01-28
[samples/Calling] Simplified Call Id ref by relocating console log in CallScreen.tsx

edwardlee-msft:
# What
Simplified use of ref in CallScreen.tsx by relocating Call Id console log in samples/Calling 

# Why
Simplified code as callid is a ref and won't retrigger a re-render. 
https://skype.visualstudio.com/SPOOL/_workitems/edit/2693381/

# How Tested
Tested on local chrome browser. 
Entered, exited, rejoined, and started a new call session to confirm if id gets logged upon joining a call.

# Process & policy checklist

### DIFF
--- a/samples/Calling/src/app/views/CallScreen.tsx
+++ b/samples/Calling/src/app/views/CallScreen.tsx
@@ -35,13 +35,6 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
   const { currentTheme, currentRtl } = useSwitchableFluentTheme();
   const [isMobileSession, setIsMobileSession] = useState<boolean>(detectMobileSession());
 
-  useEffect(() => {
-    if (!callIdRef.current) {
-      return;
-    }
-    console.log(`Call Id: ${callIdRef.current}`);
-  }, [callIdRef.current]);
-
   // Whenever the sample is changed from desktop -> mobile using the emulator, make sure we update the formFactor.
   useEffect(() => {
     const updateIsMobile = (): void => setIsMobileSession(detectMobileSession());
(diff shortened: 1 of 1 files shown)


samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -69,7 +62,10 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
         const pageTitle = convertPageStateToString(state);
         document.title = `${pageTitle} - ${webAppTitle}`;
 
-        callIdRef.current = state?.call?.id;
+        if (!callIdRef.current && state?.call?.id) {
```
JamesBurnside, line 65: note: this may need to also check if `callIdRef !== state?.call?.id` to log a new callId if the callId changes

edwardlee-msft: After my testing, I found that it outputs a new callid when you join, exit and rejoin, or join a new session after exiting.

But just in case, I revised the code to add your note for edge cases that were not tested. I tested the new edit and the logs are outputting as expected. New callids when joining, then rejoining a call.

JamesBurnside: Yeah the call ID may change _during_ an active call - I'm not clear when this happens, something very back-end related.

JamesBurnside (approved):

JamesBurnside:
Out of scope note: it would be useful to have this in the meeting sample also!

alkwa-msft:
I am open to this being out of scope. I think this is helpful when people want to debug why their call quality wasnt great. Having the CallId is always one of the first questions we need in order to determine what happened in the call from the service infrastructure side.

edwardlee-msft:
Created a new PR for sample/Meeting
https://github.com/Azure/communication-ui-library/pull/1383

anjulgarg (approved):

alkwa-msft (approved):

PR #1384, opened 2022-01-28
Adding verbose flag to the command rush build:all-flavors

anjulgarg:
# What
Displays build errors when running rush build:all-flavors

# Why
Locally building conditional code didn't yield errors

# How Tested

# Process & policy checklist

### DIFF
--- a/common/config/rush/command-line.json
+++ b/common/config/rush/command-line.json
@@ -70,7 +70,7 @@
       "name": "build:all-flavors",
       "summary": "Build all flavor in one command (mainly for api update)",
       "description": "Build all flavor in one command (mainly for api update). This command will switch current flavor to beta.",
-      "shellCommand": "(rush switch-flavor:stable && rush rebuild) || rush switch-flavor:beta && rush rebuild",
+      "shellCommand": "(rush switch-flavor:stable && rush rebuild -v) || rush switch-flavor:beta && rush rebuild -v",
       "safeForSimultaneousRushProcesses": true
     },
     {


common/config/rush/command-line.json
```diff
@@ -70,7 +70,7 @@
       "name": "build:all-flavors",
       "summary": "Build all flavor in one command (mainly for api update)",
       "description": "Build all flavor in one command (mainly for api update). This command will switch current flavor to beta.",
-      "shellCommand": "(rush switch-flavor:stable && rush rebuild) || rush switch-flavor:beta && rush rebuild",
+      "shellCommand": "(rush switch-flavor:stable && rush rebuild -v) || rush switch-flavor:beta && rush rebuild -v",
```
JamesBurnside, line 73: This won't fix when we do `rush build -t <pkg_name>` will it?

prprabhu-ms: The problem is when rush runs multiple builds in parallel (so `rush build -t` doesn't log verbose either). But `rushx build` will give you verbose errors (or run with `rush build -t -v`

prprabhu-ms (approved):

common/config/rush/command-line.json
JamesBurnside: but `rush build -t` isn't running multiple builds in parallel is it?

JamesBurnside: (or at least no more than it was before conditional compile)

anjulgarg: No, since it is a pure rush command and we can't override it. Only solution is to add a -v flag :)

JamesBurnside: My concern with `-v` is the output can get out of hand quickly - it's use isn't for seeing errors that should be reported as part of the normal build process it's, typically, for delving deeper into why there was a failure.
I don't think we understand why our errors are being silenced by the new build commands well enough to say this is the only solution (it might be I just haven't been convinced on it)

anjulgarg: Sure, that's a totally valid point @JamesBurnside 
This just seemed like the quickest easiest way to fix it.
We can always look for answers later if this unblocks us.

Also to your point about `-v` getting out of hand.
We use this flag extensively in our github pipelines.
I am yet to see the output get out of control :D

JamesBurnside: We talked before about our build system being scary, adding _even more_ output to that only furthers that.
For a simply mistyped variable name the error is super hidden in amongst other errors and red text:
image
(note this is only the latter half the output also)

anjulgarg: `grep "error"` FTW :D

prprabhu-ms: About the use of `-v` in other commands in the pipeline -- I recently added it, and not with a lot of thought, so that precedent isn't worth much.

Open to better options. I just landed that PR to get errors from CI where there were none.

PR #1386, opened 2022-01-28
Chat notification logic

dmceachernmsft:
# What
Adds state knowledge to meeting composite for unread messages

# Why
Allows for notification of unread messages to work

https://skype.visualstudio.com/SPOOL/_workitems/edit/2698422
https://skype.visualstudio.com/SPOOL/_workitems/edit/2698424

# How Tested
ran locally in ACS call and teams in the meeting sample.


packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+export type NotificationIconProps = {
+  numberOfMessages: number;
+};
+/**
+ * @private
+ */
+export const NotificationIcon = (props: NotificationIconProps): JSX.Element => {
+  const { numberOfMessages } = props;
+  return (
+    <Stack horizontalAlign="center" verticalAlign="center" style={{ position: 'absolute', top: 0, right: '0.78rem' }}>
```
alkwa-msft, line 20: can we avoid using inline styles unless absolutely necessary?

Moving to another file would be my suggestion.

packages/communication-react/review/communication-react.api.md
```diff
@@ -284,6 +284,7 @@ export const CallComposite: (props: CallCompositeProps) => JSX.Element;
...
     LocalCameraSwitch?: JSX.Element;
+    NotificationCircle?: JSX.Element;
```
alkwa-msft, line 287: It happens to be a circle for us but do we want to call it NotificationCircle?

is this potentially a name we want to outline in a design that we can share with the mobile folks? I am thinking about the future when they bring in chat and they need some way to indicate there are unread messages. Imagine developers implement some changes that use the term "NotificationCircle" but we refer to it as that in web and mobile choose something differently.

packages/react-composites/src/composites/MeetingComposite/ChatButton.tsx
```diff
@@ -4,24 +4,38 @@
...
 import { ControlBarButton, ControlBarButtonProps } from '@internal/react-components';
 import { Chat20Filled, Chat20Regular } from '@fluentui/react-icons';
+import { NotificationIcon } from './NotificationIcon';
+import { Stack } from '@fluentui/react';
+
+/**
+ * @private
+ */
+export interface ChatButtonProps extends ControlBarButtonProps {
+  numberOfUnreadMessages?: number;
```
alkwa-msft, line 14: what are our thoughts on unreadMessageCount

Its shorter and its clearer based on the type it represents?

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -102,20 +104,44 @@ const MeetingScreen = (props: MeetingScreenProps): JSX.Element => {
 
   const [currentMeetingState, setCurrentMeetingState] = useState<CallState>();
   const [currentPage, setCurrentPage] = useState<MeetingCompositePage>();
-  meetingAdapter.onStateChange((newState) => {
-    setCurrentPage(newState.page);
-    setCurrentMeetingState(newState.meeting?.state);
-  });
-
+  const [unreadChatMessages, setUnreadChatMessages] = useState<number>(0);
```
alkwa-msft, line 107: the naming of this state implies a collection of messages when its really state that represents the number of unread messages

lets change it to numberOfUnreadChatMessages, setNumberOfUnreadChatMessages or if we can be consistent and we used unreadChatMessagesCount then we should use 

const[ unreadChatMessagesCount, setUnreadChatMessagesCount] = useState<number>(0);

packages/communication-react/review/communication-react.api.md
```diff
@@ -850,6 +851,7 @@ export const COMPOSITE_ONLY_ICONS: {
     LocalDeviceSettingsSpeaker: JSX.Element;
     LocalPreviewPlaceholder: JSX.Element;
     LocalCameraSwitch: JSX.Element;
+    notificationCircle: JSX.Element;
```
anjulgarg, line 854: Capital Case
`notificationCircle` is very generic.
I recommend something like `ChatButtonNotificationBubble`

packages/react-composites/src/composites/MeetingComposite/ChatButton.tsx
anjulgarg: +1

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -102,20 +104,44 @@ const MeetingScreen = (props: MeetingScreenProps): JSX.Element => {
...
+      setCurrentPage(newState.page);
+      setCurrentMeetingState(newState.meeting?.state);
+    };
+    meetingAdapter.onStateChange(updateMeetingPage);
+    return () => {
+      meetingAdapter.offStateChange(updateMeetingPage);
+    };
+  }, [meetingAdapter]);
+  useEffect(() => {
+    const incrementUnreadtChatMessages = (event: { message: ChatMessage }): void => {
```
alkwa-msft, line 122: I think there is a mis-spelling here

Are we incrementing our unread chat messages or the number of unreadChatMessages. Again we want to be consistent about the naming to the type.

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import React from 'react';
+import { Icon, Stack } from '@fluentui/react';
+import { NotificationIconStyles, NotificationTextStyles } from './styles/MeetingCompositeStyles';
+
+/**
+ * @private
+ */
+export type NotificationIconProps = {
+  numberOfMessages: number;
```
alkwa-msft, line 12: can we be consistent here as well.

Messages or ChatMessages?

should we use prefix with number or a suffix with count?

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
anjulgarg: I think it's important to understand the reasoning behind why we do somethings the way we do them.
For example, a common PR comment I see from many team members is the suggestion of moving an inline style to another a constant in another file.
Which to me personally has no advantage (apart from peoples personal preference around readability).

If we were to not use `style` prop and use `className` instead with `mergeStyles`, that presents an advantage I can clearly see. For example, re-use of some class names, ability to compress css files etc.

packages/communication-react/review/communication-react.api.md
dmceachernmsft: Since we are using it as the primary part of the notification component would the name NotificationIcon be better?

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
dmceachernmsft: okay I will revisit the naming here

change/@internal-react-composites-7fa0abbc-d304-4a7d-9889-bf8d1677770c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
prprabhu-ms, line 2: This should be "prerelease".

Only PRs that affect the _stable build_ are patch / minor / major.

packages/react-composites/src/composites/MeetingComposite/ChatButton.tsx
```diff
@@ -4,24 +4,38 @@
...
+    <Stack style={{ position: 'relative' }}>
+      <ControlBarButton
+        {...props}
+        labelKey={'chatButtonLabelKey'}
+        strings={strings}
+        onRenderOnIcon={props.onRenderOnIcon ?? onRenderOnIcon}
+        onRenderOffIcon={props.onRenderOffIcon ?? onRenderOffIcon}
+        onClick={props.onClick}
+      />
+      {unreadMessageCount !== undefined && unreadMessageCount > 0 && (
```
prprabhu-ms, line 36: There is already logic inside `NotificationIcon` that handles different values of `unreadMessageCount`. In this case, I'd argue it's best to keep that logic in a single place: Always return a `<NotificaitonIcon>` and that component decides when to render as `<></>`, including when `unreadMessageCount` is `undefined` or `0`.

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -102,20 +104,44 @@ const MeetingScreen = (props: MeetingScreenProps): JSX.Element => {
...
-  meetingAdapter.onStateChange((newState) => {
-    setCurrentPage(newState.page);
-    setCurrentMeetingState(newState.meeting?.state);
-  });
-
+  const [unreadChatMessagesCount, setUnreadChatMessagesCount] = useState<number>(0);
   const [showChat, setShowChat] = useState(false);
   const [showPeople, setShowPeople] = useState(false);
 
+  useEffect(() => {
```
prprabhu-ms, line 111: Moving the `onStateChange` subscriber inside a `useEffect` seems unrelated to the intent of this PR. Please extract into its own (far far simpler) PR.

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -102,20 +104,44 @@ const MeetingScreen = (props: MeetingScreenProps): JSX.Element => {
...
+    const updateMeetingPage = (newState): void => {
+      setCurrentPage(newState.page);
+      setCurrentMeetingState(newState.meeting?.state);
+    };
+    meetingAdapter.onStateChange(updateMeetingPage);
+    return () => {
+      meetingAdapter.offStateChange(updateMeetingPage);
+    };
+  }, [meetingAdapter]);
+  useEffect(() => {
```
prprabhu-ms, line 121: Is this a temporary implementation of how to count chat messages?

We try to avoid carrying business logic in `useEffect` blocks using the adapter API. That breaks a distinction we've tried to make between ACS business logic and UI. I can't think of a simple way to merge the inputs from ACS (incoming messages) and UI (last time chat pane was opened / whether the chat pane is open right now) in this case either.

If we take this approach, couple suggestions:
* Document why we can't use selectors to achieve what we want.
* Move this logic inside `NotificationIcon` component, to keep it encapsulated.

prprabhu-ms (commented):
It would have been better to base this PR off of the other PR that introduces `NotificationIcon`. Currently, this PR contains the diff from there as well.

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
dmceachernmsft: The logic was out here incase we wanted to look at making the notification Icon a re-usable component with other buttons. So accessing the meeting state seemed best to do here at the composite level where we can subscribe to when a new message comes in and just pass a number around. I can leave a comment about why the composite holds this logic if you think thats best?

JamesBurnside: Yeah using selectors is just going to be complicated but definitely possible -- we need the composite to store a local state of "seen messages", then have a selector filter `!state.messages.includes(compositeState.seenMessages)` then on chat open/close update the `compositeState.seenMessages`.
It'll conflict a little with the messageThread's internal implementation/interpretation of "unread messages" so some potential extra work to make sure those would marry up nicely.
But I think this is a good approach for a V1 implementation -- but also some great value if we can expose unreadMessages through the adapter API

prprabhu-ms: Note: I'm not complaining about the impl here (I can't think of a good selector based impl either).

But let's encapsulate this logic better.

JamesBurnside: And the other reason this existed in the meetingcomposite file instead of the chatbutton / notification icon was that they don't have knowledge if the chat pane was opened/closed

prprabhu-ms: Let's get that knowledge about chat pane state prop-drilled down. I won't hold a hard line on this, but I do think that this logic needs to be encapsulated.

dmceachernmsft: will take a look at this!

prprabhu-ms (commented):
Could you @mention me once you've integrated the changes we discussed yesterday for encapsulating the logic?

dmceachernmsft:
> It would have been better to base this PR off of the other PR that introduces `NotificationIcon`. Currently, this PR contains the diff from there as well.

@prprabhu-ms the changes requested are in this PR: https://github.com/Azure/communication-ui-library/pull/1411

change/@internal-react-composites-7fa0abbc-d304-4a7d-9889-bf8d1677770c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Added logic to meeting composite to store unread messages",
```
prprabhu-ms, line 3: nitty nit: Could be terse, and a bit more accurate "MeetingComposite: store messages since last time ChatPane was opened"

"unread" has a different meaning, being pedantic, and we're only approximating that.

packages/communication-react/review/communication-react.api.md
```diff
@@ -284,6 +284,7 @@ export const CallComposite: (props: CallCompositeProps) => JSX.Element;
...
     LocalCameraSwitch?: JSX.Element;
+    NotificationIcon?: JSX.Element;
```
prprabhu-ms, line 287: Is this also renamed in a different PR? `NotificationIcon` still seems too generic (there could be multiple icons, and we may not want all of them to look identical). `NewMessageNotificationIcon` maybe?

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -102,20 +104,39 @@ const MeetingScreen = (props: MeetingScreenProps): JSX.Element => {
...
   meetingAdapter.onStateChange((newState) => {
     setCurrentPage(newState.page);
     setCurrentMeetingState(newState.meeting?.state);
   });
 
-  const [showChat, setShowChat] = useState(false);
-  const [showPeople, setShowPeople] = useState(false);
+  useEffect(() => {
+    const incrementUnreadtChatMessagesCount = (event: { message: ChatMessage }): void => {
+      if (!showChat && event.message.senderDisplayName !== '') {
```
prprabhu-ms, line 118: Please extract a local helper function that better states what this check does:

```ts
const isMessageFromBot = (message): bool => {
  // Some comment about why display name is a good heuristic. Is it?
  return message.senderDisplayName === '';
}
```

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -102,20 +104,39 @@ const MeetingScreen = (props: MeetingScreenProps): JSX.Element => {
...
+        if (event.message.type === 'text' || event.message.type === 'html') {
```
prprabhu-ms, line 119: I'm a bit worried about this check being repeated multiple times in our codebase. Can we export an `@internal` helper for this from `acs-ui-common`?

https://github.com/Azure/communication-ui-library/blob/f87bf0c6588a4fff9469a130d0492ce429a57779/packages/chat-component-bindings/src/messageThreadSelector.ts#L29
https://github.com/Azure/communication-ui-library/blob/cf149e82f77ce8bd7c4707b1bfc84bbfd9c9c2df/packages/chat-stateful-client/src/EventSubscriber.ts#L55

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -102,20 +104,39 @@ const MeetingScreen = (props: MeetingScreenProps): JSX.Element => {
...
     setShowPeople(false);
   }, []);
 
   const toggleChat = useCallback(() => {
+    if (!showChat) {
+      setUnreadChatMessagesCount(0);
+    }
     setShowPeople(false);
     setShowChat(!showChat);
   }, [showChat]);
```
prprabhu-ms, line 142: This dependency list should ideally also include the `set*` functions being called.

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
prprabhu-ms: Please do extract this particular one into a constant. Because that positioning is quite tricky -- it's trying to be just above the icon used by the button this overlays. This context is not at all obvious from looking at the `style` and deserves a detailed comment in the code.

packages/react-composites/src/composites/MeetingComposite/styles/MeetingCompositeStyles.ts
```diff
@@ -18,3 +18,24 @@ export const compositeOuterContainerStyles: IStackStyles = {
     flexWrap: 'wrap'
   }
 };
+
+/**
+ * @private
+ */
+export const NotificationIconStyles: IIconStyles = {
+  root: {
+    color: '#0078D4'
```
prprabhu-ms, line 27: Should come from theme.

packages/react-composites/src/composites/MeetingComposite/styles/MeetingCompositeStyles.ts
```diff
@@ -18,3 +18,24 @@ export const compositeOuterContainerStyles: IStackStyles = {
...
+};
+
+/**
+ * @private
+ */
+export const NotificationTextStyles: IStackStyles = {
+  root: {
+    position: 'absolute',
+    top: '0.1rem',
+    color: 'white',
```
prprabhu-ms, line 38: Should come from theme.

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
dmceachernmsft: This is resolved in another PR.

dmceachernmsft: ^ https://github.com/Azure/communication-ui-library/pull/1411

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
dmceachernmsft: this is now in another PR: https://github.com/Azure/communication-ui-library/pull/1411

packages/react-composites/src/composites/MeetingComposite/styles/MeetingCompositeStyles.ts
dmceachernmsft: resolved in this PR: https://github.com/Azure/communication-ui-library/pull/1378

packages/react-composites/src/composites/MeetingComposite/styles/MeetingCompositeStyles.ts
dmceachernmsft: resolved in this PR: https://github.com/Azure/communication-ui-library/pull/1378

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
dmceachernmsft: this is resolved now in this PR: https://github.com/Azure/communication-ui-library/pull/1378

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
dmceachernmsft: Added a comment about this, but in short system messages wont have a displayname so we dont have to have those notifications added.

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
dmceachernmsft: made a helper function in another PR: https://github.com/Azure/communication-ui-library/pull/1411

packages/react-composites/src/composites/MeetingComposite/MeetingComposite.tsx
```diff
@@ -114,15 +119,30 @@ const MeetingScreen = (props: MeetingScreenProps): JSX.Element => {
...
+    };
+  }, [meetingAdapter, setUnreadChatMessagesCount, showChat, unreadChatMessagesCount]);
 
   const closePane = useCallback(() => {
     setShowChat(false);
     setShowPeople(false);
   }, []);
 
   const toggleChat = useCallback(() => {
+    if (!showChat) {
```
JamesBurnside, line 143: add a comment here so this is clear -- 
```suggestion
    // When the chat pane is opened, mark all chat messages as read
    if (!showChat) {
```

packages/react-composites/src/composites/MeetingComposite/ChatButton.tsx
```diff
@@ -26,6 +41,14 @@ export const ChatButton = (props: ControlBarButtonProps): JSX.Element => {
         onRenderOffIcon={props.onRenderOffIcon ?? onRenderOffIcon}
         onClick={props.onClick}
       />
+      {}
```
JamesBurnside, line 44: remove?

JamesBurnside (approved):

packages/react-composites/src/composites/MeetingComposite/ChatButton.tsx
dmceachernmsft: oops! will do!

prprabhu-ms (approved):

PR #1388, opened 2022-01-28
Update visuals of `MeetingComposite` control bar buttons

prprabhu-ms:
# Why

Match designs.

Still pending: split buttons and overflow button. This only updates the styling.

before:
image

after:
image

# Alternatives considered

This PR updates achieves the updated styling without introducing any new API. A downside is that any custom buttons created using `ControlBarButton` will not automagically have the correct styling. The button will look consistent in `CallComposite` wihout any effort, but will require custom styling to look like other buttons in `MeetingComposite`.

The alternative would be to introduce a new field in props for all our buttons to pick the styling used by `CallComposite` vs `MeetingComposite`. The same field would then give the correct default styling for the custom button built using `ControlBarButton`. I decided against this option because the styling is somewhat arbitrary (there is nothing conceptually unique about the styling used in either).


packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -114,10 +112,58 @@ export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.El
...
-        </Stack.Item>
+        </Stack>
       )}
     </Stack>
   );
 };
+
+const buttonContainerStyle: IStyle = {
+  padding: '0.75rem',
+  columnGap: '0.5rem'
```
JamesBurnside, line 137: nit: does it look bad with slightly more space between the label and buttons? current (left) looks a little tighter spaced than the figma (right):
image

JamesBurnside: Not sure how much is padding and how much is columnGap, but maybe 0.625rem for column gap? 
image

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -114,10 +112,58 @@ export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.El
...
+};
+
+const controlBarStyles: BaseCustomStyles = {
+  root: buttonContainerStyle
+};
+
+const commonButtonStyles: ControlBarButtonStyles = {
```
JamesBurnside, line 144: Thia already looks so much nicer! My concern with not exporting these as buttons is if contoso wants to inject a custom button - there's no concrete way for them to style the buttons the same as ours

JamesBurnside: (also I'm losing faith our react-components are actually providing good value the more we choose not to export :( )

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -42,6 +42,21 @@ export type CallControlsProps = {
    * Recommended for mobile devices.
    */
   increaseFlyoutItemSize?: boolean;
+
+  /**
+   * Styles for the {@link ControlBar}.
+   */
+
```
JamesBurnside, line 49: nit: remove empty line

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -114,10 +112,58 @@ export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.El
...
+  root: {
+    border: 'solid 1px #E1DFDD',
```
JamesBurnside, line 146: note: I think these styles should only apply to desktopView

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -114,10 +112,58 @@ export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.El
...
+    borderRadius: '0.25rem',
```
anjulgarg, line 147: You can use `theme.effects.roundedCorner4` here

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -114,10 +112,58 @@ export const MeetingCallControlBar = (props: MeetingCallControlBarProps): JSX.El
...
+    flexFlow: 'row nowrap'
+  },
+  textContainer: {
+    // Override the default so that label doesn't introduce a new block.
+    display: 'inline'
+  },
+  label: {
+    // Override styling from ControlBarButton so that label doesn't introduce a new block.
+    display: 'inline',
+    fontSize: '0.875rem'
```
anjulgarg, line 160: This breaks when end user changes the default font size using theme.
I recommend using
`theme.fonts.small.size...` (something like that)

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
anjulgarg, line 146: Why are we not using a theme palette color for this border?

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
anjulgarg: I was under the impression that we were going to add more buttons to react-composites and then use them here :o

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
prprabhu-ms: Decided not to use theme here because the `ControlBarButton` explicitly styles with `borderRadius: 0` and `EndCallButton` explicitly styles with `borderRadius: 0.5rem`.

I'd like these to stay consistent (applying theme halfway would be worse, imo)

prprabhu-ms: But let me take another holistic look at using theme. I'll use theme if the base buttons are using them.

prprabhu-ms:
The biggest question here is if we're introducing a new set of buttons.

Along with (draft) https://github.com/Azure/communication-ui-library/pull/1392, we can avoid introducing new buttons entirely. I like this approach because we haven't fundamentally introduced a new concept here. We should avoid exporting multiple buttons that are almost the same, and where the differences are not conceptually well defined (er.... "narrow && sometimes splitbutton && border" -- why is that a new button again?)

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
prprabhu-ms: @JamesBurnside we _could_ export this style constant from `react-composites` as `MeetingControlButtonDefaultStyle` that they use.

I'm not convinced that stylistic differences are large enough to warrant a new set of buttons / base button.

JamesBurnside: I agree, I hate having to export new buttons just for styling differences, but I also really want someone to be able to inject a custom button that looks and behaves the exact same as ours --

JamesBurnside: Scratch my concern -- the API for custom buttons is simply props not JSX so by default we will style their buttons to match ours (they can pass in more styles for more customization)
We can work with Sophy's team to give them the necessary styles for creating their Azure Portal components

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -42,6 +42,21 @@ export type CallControlsProps = {
...
+  controlBarStyles?: BaseCustomStyles;
+  /**
+   * Styles for all buttons except {@link EndCallButton}.
```
JamesBurnside, line 52: Do these new styles apply to mobile (or how does it look with mobileView=true)

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
prprabhu-ms: Oh yeah, I had also completely misunderstood my own approach. Any custom buttons injected in the `MeetingComposite` *will* by default get the new styling.

But I'm generally very wary of the (stability of) all this CSS stacking.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
prprabhu-ms: Right you are. Done.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
prprabhu-ms: I haven't modified the default `IconButton` gaps here. Not sure I want to mess with that.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
prprabhu-ms: Done.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
prprabhu-ms: Done

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
prprabhu-ms: Done.

prprabhu-ms:
@anjulgarg  Thanks for pushing for using `theme` where possible.

PTAL.

anjulgarg:
> The biggest question here is if we're introducing a new set of buttons.
> 
> Along with (draft) #1392, we can avoid introducing new buttons entirely. I like this approach because we haven't fundamentally introduced a new concept here. We should avoid exporting multiple buttons that are almost the same, and where the differences are not conceptually well defined (er.... "narrow && sometimes splitbutton && border" -- why is that a new button again?)
> 
> @JamesBurnside @anjulgarg ^^

I guess we can always export the buttons later if there is a demand for it. :)

anjulgarg (approved):

JamesBurnside (approved):

PR #1389, opened 2022-01-28
Add File Sharing (Upload) props to ChatComposite

anjulgarg:
# What
Add File Sharing (Upload) props to ChatComposite

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2671562

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -67,6 +69,27 @@ export type ChatCompositeOptions = {
    * Set focus on the composite when the composite first mounts.
    */
   autoFocus?: 'sendBoxTextField' | false;
+  /* @conditional-compile-remove-from(stable) */
```
prprabhu-ms, line 72: These three arguments are clearly related. I'd vote for creating a nested object to keep them together. Yes JS interfaces seem to have a gazillion fields distinguished with (concatenated) prefixes. No it is not a good thing.

packages/communication-react/review/communication-react.api.md
```diff
@@ -2056,6 +2069,62 @@ export interface TypingIndicatorStylesProps extends BaseCustomStyles {
     typingUserDisplayName?: IStyle;
 }
 
+// @beta (undocumented)
+export const UPLOAD_CANCELLED_EVENT = "uploadCancelled";
```
prprabhu-ms, line 2073: Why are these events separate constants rather than a single string literal union: `export const FileUploadEvent = 'cancelled' | 'completed' | 'failed' | ...`. That is more in line with the rest of our API.

packages/react-composites/review/react-composites.api.md
```diff
@@ -753,6 +766,62 @@ export type TopicChangedListener = (event: {
...
+    constructor(file: File);
+    cancelUpload(): void;
+    completeUpload(metaData: FileMetaData): void;
+    extension(): string;
+    failUpload(message: string): void;
+    file: File;
+    // (undocumented)
+    isUploaded(): boolean;
+    metaData?: FileMetaData;
+    off(event: UploadedFileEvents, listener: UploadedFileEventListener): void;
```
prprabhu-ms, line 798: This most generic overload for `on` and `off` shouldn't be part of the type. Implement the most generic function, but don't declare it as an overload: https://www.typescriptlang.org/docs/handbook/2/functions.html#overload-signatures-and-the-implementation-signature

packages/react-composites/review/react-composites.api.md
```diff
@@ -753,6 +766,62 @@ export type TopicChangedListener = (event: {
...
+export const UPLOAD_PROGRESSED_EVENT = "uploadProgressed";
+
+// @beta
+export type UploadCanceledListener = () => void;
+
+// @beta
+export type UploadCompleteListener = (metaData: FileMetaData) => void;
+
+// @beta
+export class UploadedFile implements UploadEvents {
```
prprabhu-ms, line 788: The naming here is a bit confusing. It's called `UploadedFile`, but then has listners for file upload progress. Perhaps `FileUploadOperation`, or `ActiveFileUpload`, or even `FileUpload` (then methods could read like `FileUpload.cancel()`, `FileUpload.on('cancelled', <handler>` etc.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -92,14 +92,19 @@ export const FileUploadButton = (props: FileUploadButtonProps): JSX.Element => {
...
-export const FileUploadButtonWrapper = (): JSX.Element => {
-  return <div>{FileUploadConditionalCode()}</div>;
-};
-
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const FileUploadConditionalCode = (): any => {
+export const FileUploadButtonWrapper = (
+  // To make conditional compilation not throw errors.
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  props: Pick<FileUploadButtonProps, 'accept' | 'multiple' | 'fileUploadHandler'>
```
prprabhu-ms, line 98: You won't have to do this if you 

(1) define the type that contains file upload options (accept, multiple, handler etc) - unconditionally
(2) Use that type in this function -- unconditionally
(3) Only for beta -- add a field to ChatComposite that uses that type && export that type from index.ts

I think trying to keep the conditional compilation as close as possible the API surface is helpful -- it reduces how much conditional compilation you need, and that means that beta/stable are effectively closer together (so lower risk when features go from beta to stable)

prprabhu-ms (commented):
I ended up asking a bunch of questions on the FileUpload API that wasn't introduced in this PR. I might have missed that PR / not paid as much attention because of lack of time. Sorry about late comments.

As this PR actually adds that API to our public (beta) API, this is the second chance at getting the API right.

prprabhu-ms:
Let's quickly chat IRL about the `UploadedFile` class today. I'm not sure I understand how it's supposed to be used.

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
anjulgarg: Good idea.

anjulgarg: Changed. LMKWYT

packages/communication-react/review/communication-react.api.md
anjulgarg: constants for avoiding spelling mistakes when using these events with event emitter as emitter can not be typed.
We do export `UploadedFileEvents` which is a string literal to be in line with our API

packages/react-composites/review/react-composites.api.md
anjulgarg: Done.

anjulgarg:
@prprabhu-ms Made all the changes we discussed. Please review when you have time. :)

anjulgarg:
@alkwa-msft Renamed UploadedFile -> FileUpload and changes property names as well. 
Resolving your comments on a separate PR.

packages/communication-react/review/communication-react.api.md
```diff
@@ -1231,6 +1232,56 @@ export interface ErrorBarStrings {
...
+    constructor(file: File);
+    cancelUpload(): void;
+    completeUpload(metaData: FileMetaData): void;
+    extension(): string;
+    failUpload(message: string): void;
+    file: File;
+    // (undocumented)
+    isUploaded(): boolean;
+    metaData?: FileMetaData;
+    off(event: typeof UPLOAD_PROGRESSED_EVENT, listener: UploadProgressListener): void;
```
dmceachernmsft, line 1260: Does the API documentation also list off the events that a class subscribes to?

anjulgarg: Yes

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
anjulgarg: I think you misunderstand this comment. Stable build throws an unused var error because props is no longer used in stable. To disabled that error, I added the comment // eslint-disable-next-line @typescript-eslint/no-unused-vars

dmceachernmsft (approved):

prprabhu-ms (approved):
Assuming that a later PR will split the `FileUpload` interface so that we only export the 4 methods that Contoso needs (and no events / event types etc?)

prprabhu-ms (approved):

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -64,11 +66,37 @@ export type ChatScreenStyles = {
...
+  /**
+   * A string containing the comma separated list of accepted file types.
+   * Similar to the `accept` attribute of the `<input type="file" />` element.
+   * Accepts any type of file if not specified.
+   * @beta
+   */
+  accept?: string;
+  /**
+   * Allows multiple files to be selected if set to `true`.
+   * Similar to the `multiple` attribute of the `<input type="file" />` element.
```
JamesBurnside, line 83: <3 these comments!

JamesBurnside (approved):

PR #1390, opened 2022-01-28
Only compile files following trace of entry index.ts of a project

PorterNan:
# What

Only compile files following trace of entry index.ts of a project

- The idea is we only compile those files imported and exported by index.ts so conditional build won't be broken by stripped out code
- And it is good to force us not to forget about exporting apis in index.ts too

# Why

# How Tested

# Process & policy checklist


JamesBurnside:
Can you check if there are any errors/lack of syntax highlighting/autocomplete in the test.ts files? And also that we don't lose any type safety in them

PorterNan:
> Can you check if there are any errors/lack of syntax highlighting/lack of autocomplete in the test.ts files? And also that we don't lose any type safety in them (e.g. wrong type _should_ error) (my concern is tsc won't run over them anymore - not sure if that matters)

Good catch! I believe we might want a separated tsc command for checking tests - which makes sense cause it is not part of our production code - will update this PR for that command

PorterNan:
> > Can you check if there are any errors/lack of syntax highlighting/lack of autocomplete in the test.ts files? And also that we don't lose any type safety in them (e.g. wrong type _should_ error) (my concern is tsc won't run over them anymore - not sure if that matters)
> 
> Good catch! I believe we might want a separated tsc command for checking tests - which makes sense cause it is not part of our production code - will update this PR for that command

image


ALL GOOD!

JamesBurnside (approved):

prprabhu-ms (changes requested):
Not sure I agree with this, on two counts:

(1) It'll be near impossible to look at a file and know whether or not it'll be compiled. I worry that devs will be left head-scratching why one file breaks the build due to some import, but another doesn't. More magic.
(2) Similar, when adding new code, devs will be forced to export it from _some_ index.ts Sometimes, starting small means exporting nothing.

What was the main reason for this PR (current example where this helps?)

PorterNan:
> 

There are several strong cases now for this change - one is the meeting composite we discussed before, in stable, we don't export them in index but still compile them(but remove some types) that generates unnecessary compile error, another is Anjul's example, check https://teams.microsoft.com/l/message/19:5e150d250f724d1089c949305efdc8cc@thread.skype/1643398258190?groupId=3e9c1fc3-39df-4486-a26a-456d80e80f82 this for discussion



And previous behavior is not good because these undesired js files will be still included into our final npm package even we are not intentionally to do so - you can't import they from the root package but you might still accidentally import it using a deep path

First for getting errors from import - devs will be able to still follow the same proccess, lets say there are 2 types of import you want to do with a file, internal and external:

1. For internal one, if you import something and want to test it under samples, that means that's something under the index.ts import tree already - if you don't, your code won't work anyway - it just gets tsc compiled when build but doing nothing more - which would be dangerous to have those code piece in our npm package cause people could still import them, using a long import paht
2. For external import, export would be needed as always, otherwise other package cannot import the right file

But I got you concern that we want every file to be compiled, there are still solution to your question - I believe we can do this configuration only for stable version build - cause you are stripping out some classes, so you know you are excluding them from the compiling system, and we don't want to even generate these js files in our stable npm packages which makes sense, then all these files will be still checked by beta build cause it is building all the files - and it doesn't matter if we put all files in beta npm packages

prprabhu-ms:
Ack to restricting the build to reachable code for stable build only. For day to day development, compile everything, all the time.

PR #1393, opened 2022-01-29
Dmceachernmsft/addingCamerasToSwitcher

dmceachernmsft:
# What
Added Cameras to local video camera switcher.

# Why
Allow camera switcher button to change video feeds 

https://skype.visualstudio.com/SPOOL/_workitems/edit/2698407

# How Tested
Ran locally in meeting sample.


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -121,7 +125,12 @@ export interface VideoGalleryProps {
   onDisposeRemoteStreamView?: (userId: string) => Promise<void>;
   /** Callback to render a particpant avatar */
   onRenderAvatar?: OnRenderAvatarCallback;
-
+  /* @conditional-compile-remove-from(stable) */
+  /**
+   * Whether or not to display the camera switcher on the local video tile
+   */
+  /* @conditional-compile-remove-from(stable) */
```
prprabhu-ms, line 132: nit: duplicate directive. You only need the one before the docstring.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -220,14 +249,27 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     if (localVideoStream && !localVideoStream.renderElement) {
       onCreateLocalStreamView && onCreateLocalStreamView(localVideoViewOptions);
     }
-
+    console.log(props.cameras);
```
prprabhu-ms, line 252: remove.

packages/react-components/src/components/VideoGallery/LocalVideoCameraButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+  onSelectCamera: (device: OptionsDevice) => Promise<void>;
+}
+
+const localVideoCameraCycleButtonStyles: IButtonStyles = {
+  root: {
+    position: 'absolute',
+    right: '0.2rem',
+    top: '0.2rem',
+    color: '#FFFFFF', // only shows up on running video feed to we want to force specific colours.
+    zIndex: 5, // to layer the button above the video tile.
```
prprabhu-ms, line 26: Please add constants for all relevant zIndex together. (i.e., if this is 5, what is 4?)

packages/react-components/src/components/styles/VideoTile.styles.ts
```diff
@@ -103,5 +103,6 @@ export const isSpeakingBorderDiv: IStyle = {
   height: '100%',
   position: 'absolute',
   top: '0',
-  left: '0'
+  left: '0',
+  pointerEvents: 'none'
```
prprabhu-ms, line 107: Why is this part of the PR?

prprabhu-ms (commented):
This seems to be part of a stack of PRs, but merges into `main`.

Let's talk IRL about stacking PRs so that diffs from PRs lower in the stack doesn't show up.

packages/react-components/src/components/VideoGallery/LocalVideoCameraButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+    background: 'transparent'
+  },
+  rootFocused: {
```
JamesBurnside, line 29: for a11y reasons there will need to be some indication this component has focus - is there a border or anything that happens on focus?

packages/react-components/src/components/VideoGallery/LocalVideoCameraButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+export const LocalVideoCameraCycleButton = (props: LocalVideoCameraCycleButtonProps): JSX.Element => {
+  const { cameras, selectedCamera, onSelectCamera } = props;
+
+  return (
+    <IconButton
+      styles={localVideoCameraCycleButtonStyles}
+      iconProps={{ iconName: 'LocalCameraSwitch' }}
+      ariaLabel={'Cycle Camera Button'}
+      onClick={() => {
+        console.log('clicked');
```
JamesBurnside, line 48: remove console log

packages/react-components/src/components/VideoGallery/LocalVideoCameraButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * @private
+ */
+export const LocalVideoCameraCycleButton = (props: LocalVideoCameraCycleButtonProps): JSX.Element => {
+  const { cameras, selectedCamera, onSelectCamera } = props;
+
+  return (
+    <IconButton
+      styles={localVideoCameraCycleButtonStyles}
+      iconProps={{ iconName: 'LocalCameraSwitch' }}
+      ariaLabel={'Cycle Camera Button'}
```
JamesBurnside, line 46: aria labels need to be localizable strings instead of fixed constants

packages/react-components/src/components/VideoGallery/LocalVideoCameraButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+      onClick={() => {
+        console.log('clicked');
+        if (cameras && selectedCamera !== undefined) {
+          const index = cameras.findIndex((camera) => selectedCamera.id === camera.id);
+          console.log(index);
```
JamesBurnside, line 51: remove console log

packages/react-components/src/components/VideoGallery/LocalVideoCameraButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { cameras, selectedCamera, onSelectCamera } = props;
+
+  return (
+    <IconButton
+      styles={localVideoCameraCycleButtonStyles}
+      iconProps={{ iconName: 'LocalCameraSwitch' }}
+      ariaLabel={'Cycle Camera Button'}
+      onClick={() => {
+        console.log('clicked');
+        if (cameras && selectedCamera !== undefined) {
```
JamesBurnside, line 49: should this also include `&& camera.length > 1`?

packages/react-components/src/components/styles/VideoTile.styles.ts
JamesBurnside: (from our debugging) Assumming this is what was capturing the click event?
If so make sure to add a comment why this css prop was added
```suggestion
  // Ensure the isSpeaking element does not steal any pointer events such as onClick events
  pointerEvents: 'none'
```

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -36,16 +36,17 @@ export interface MediaGalleryProps {
...
+  isMobile?: boolean;
 }
 
 /**
  * @private
  */
 export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
   const videoGalleryProps = usePropsFor(VideoGallery);
-
+  console.log(videoGalleryProps);
```
JamesBurnside, line 48: nit: another one ;)

packages/react-components/src/components/VideoGallery/LocalVideoCameraButton.tsx
dmceachernmsft: do we have a design suggestion for this? (I'll ask Alex P but do we have a standard we apply now?) right now it was just making the background completely white over the video

packages/react-components/src/components/VideoGallery/LocalVideoCameraButton.tsx
dmceachernmsft: kk will add string!

change/@internal-calling-component-bindings-bf4edf9c-6d58-4b69-b2f8-1a9c60a77ce7.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
anjulgarg, line 2: Should `prerelease`
`minor` is used for changing the stable build.

change/@internal-react-components-d1a4dd6d-0a5d-4430-b2a9-6fba219cc527.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
anjulgarg, line 2: `prerelease`

change/@internal-react-composites-4cba40c4-0e66-45b4-9ec3-c580777b3188.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
anjulgarg, line 2: `prerelease` if it's changing/exposing props from chat-composite.
`none` if all of it is internal logic

anjulgarg (changes requested):
I recommend making changes to the `onRenderLocalVideoTile` prop in VideoGallery and have it accept device information as well.
Create a separate component and selector for it inside call composite and pass it in videogallery in call page of call composite.

Causes minimum changes to the stable build and is a cleaner implementation.
Also, easier to remove using conditional compilation

Thoughts?

JamesBurnside:
> I recommend making changes to the `onRenderLocalVideoTile` prop in VideoGallery and have it accept device information as well. Create a separate component and selector for it inside call composite and pass it in videogallery in call page of call composite.
> 
> Causes minimum changes to the stable build and is a cleaner implementation. Also, easier to remove using conditional compilation
> 
> Thoughts?

I think our localVideoTile is super heavily embedded into the God component that is the VideoGallery. Ideally 100% we would do this but we need to break apart the video gallery into sub components in many places. Maybe as a follow up work item we could create 'extract local video tile' in the same way remote video tile is its own component but I'd be worried doing it as part of this change in case we introduce the old flickering/re-rendering bugs (such as when it flickered when switching from regular grid view to horizontal gallery)

prprabhu-ms (approved):

dmceachernmsft:
> > I recommend making changes to the `onRenderLocalVideoTile` prop in VideoGallery and have it accept device information as well. Create a separate component and selector for it inside call composite and pass it in videogallery in call page of call composite.
> > Causes minimum changes to the stable build and is a cleaner implementation. Also, easier to remove using conditional compilation
> > Thoughts?
> 
> I think our localVideoTile is super heavily embedded into the God component that is the VideoGallery. Ideally 100% we would do this but we need to break apart the video gallery into sub components in many places. Maybe as a follow up work item we could create 'extract local video tile' in the same way remote video tile is its own component but I'd be worried doing it as part of this change in case we introduce the old flickering/re-rendering bugs (such as when it flickered when switching from regular grid view to horizontal gallery)

@anjulgarg I met you and James half way by making a new selector that we can use when we are ready to do the work that James is suggesting.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -237,8 +246,18 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
           localVideoStream?.renderElement ? (
             <>
               {
-                /* @conditional-compile-remove-from(stable) Local_Camera_switcher */
-                showCamerSwitcherInLocalPreview && <LocalVideoCameraCycleButton />
+                /* @conditional-compile-remove-from(stable) meeting/calling-composite <Local-Camera-Switcher> */
+                showCamerSwitcherInLocalPreview &&
```
anjulgarg, line 250: nit:
Might be cleaner to move this logic to a function like
`shouldRenderLocalVideoCameraCycleButton()`
instead of checking all the conditions inline in JSX :)

anjulgarg: Especially since VideoGallery is already becoming too complex to handle :D

anjulgarg (approved):

anjulgarg:
> > > I recommend making changes to the `onRenderLocalVideoTile` prop in VideoGallery and have it accept device information as well. Create a separate component and selector for it inside call composite and pass it in videogallery in call page of call composite.
> > > Causes minimum changes to the stable build and is a cleaner implementation. Also, easier to remove using conditional compilation
> > > Thoughts?
> > 
> > 
> > I think our localVideoTile is super heavily embedded into the God component that is the VideoGallery. Ideally 100% we would do this but we need to break apart the video gallery into sub components in many places. Maybe as a follow up work item we could create 'extract local video tile' in the same way remote video tile is its own component but I'd be worried doing it as part of this change in case we introduce the old flickering/re-rendering bugs (such as when it flickered when switching from regular grid view to horizontal gallery)
> 
> @anjulgarg I met you and James half way by making a new selector that we can use when we are ready to do the work that James is suggesting.

Thanks for making the change @dmceachernmsft 
Left a small comment around code cleanup. But otherwise, LGTM

packages/react-components/review/react-components.api.md
```diff
@@ -460,6 +460,21 @@ export type LocalizationProviderProps = {
     children: React_2.ReactNode;
 };
 
+// Warning: (ae-internal-missing-underscore) The name "LocalVideoCameraCycleButton" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal
+export const LocalVideoCameraCycleButton: (props: LocalVideoCameraCycleButtonProps) => JSX.Element;
```
prprabhu-ms, line 466: naming nit: `LocalVideoCameraCycleButton` vs `localVideoCameraSwitcherButtonProps`.

I don't remember which one we decided on, but let's stick to one.

packages/react-components/src/components/LocalVideoCameraButton.tsx
```diff
@@ -3,26 +3,26 @@
...
 export interface LocalVideoCameraCycleButtonProps {
   /** Array of cameras available to the user. */
-  cameras: OptionsDevice[];
+  cameras?: OptionsDevice[];
   /** Currently selected camera in the local video stream. */
-  selectedCamera: OptionsDevice;
+  selectedCamera?: OptionsDevice;
   /** callback function to change video feed. */
-  onSelectCamera: (device: OptionsDevice) => Promise<void>;
+  onSelectCamera?: (device: OptionsDevice) => Promise<void>;
```
prprabhu-ms, line 18: Should these all be required? 

This is an internal API, so we know that these props will always be set. So tell typescript to yell if we do it wrong.

prprabhu-ms (approved):

packages/react-components/src/components/VideoGallery.tsx
dmceachernmsft: will do! 😊

packages/react-components/src/components/LocalVideoCameraButton.tsx
dmceachernmsft: the way useHandlers is implemented does not like the guarantee that there is a callback there. 😥

packages/react-components/review/react-components.api.md
dmceachernmsft: ah thanks for catching that, I'll review the naming and make sure it aligns!

PR #1394, opened 2022-01-30
Add create beta/stable release branch actions

PorterNan:
# What

This PR introduces 2 release creation actions - beta&stable

- Release validation app will be built and deployed depending on branch name(whehter includes beta)
- Use beachball only to collect logs but not to bump the version
- Add our own workflow to bump versions
- For stable release - people will need to fill in parameter for stable release bump - whether it is a minor or patch release

# Why

# How Tested

2 generated PRs from these 2 actions:
https://github.com/Azure/communication-ui-library/pull/1402
https://github.com/Azure/communication-ui-library/pull/1401

# Process & policy checklist


common/config/package.json
```diff
@@ -18,7 +18,7 @@
     "@octokit/rest": "~18.0.6",
     "@rollup/plugin-commonjs": "~17.1.0",
     "@types/node": "^14.14.10",
-    "beachball": "~2.20.0",
+    "beachball": "2.20.0",
```
PorterNan, line 21: Ensure to lock the beachball version in order to keep beachball-changelog-only.js
Which is a work around before https://github.com/microsoft/beachball/issues/634 gets resolved

common/scripts/bump-beta-version.js
```diff
@@ -0,0 +1,23 @@
...
+
+const main = () => {
+  const packagePaths = findAllPackageJSON(PACKAGES_DIR);
+  const depNames = getAllNames(packagePaths);
+  updateAllVersions(bumpBetaVersion);
+  // Need to update all internal project dependencies using the same rule
+  updateAllDepVersions(bumpBetaVersion, depNames);
+}
+
+const bumpBetaVersion = (currentVersion) => {
```
JamesBurnside, line 14: Instead of writing our own logic, does `beachball canary` cmd work? (what we use for creating alpha packages)

PorterNan: That is a good point! Will beachball collect changelog and generate changelog files when using canary? Or it only updates all the versions in package.json

JamesBurnside: I don't think so, I think it only bumps packages

PorterNan: I tried canary method and looks like it doesn't follow our current bumping logic, since changing our bumping logic is another topic, I will keep this until canary supports this or we decide to change our bumping logic

.github/workflows/deploy-release-azure-webapps.yml
```diff
@@ -33,6 +33,11 @@ jobs:
       - name: Install Dependencies
         run: rush install
 
+      # Switch flavor to stable when it is release branch
+      - name: Switch flavor for stable build
+        if: ${{ !contains(github.ref, 'beta') }}
```
prprabhu-ms, line 38: This is brittle as it depends on the branch name directly. That might be surprising.

Can this workflow take an explicit argument instead to say which flavor build to deploy (with default to `beta`)?

PorterNan: Agree, this is following how currently we do release for branches - but we do need to iterate that in following PRs

JamesBurnside (approved):

common/scripts/beachball-changelog-only.js
```diff
@@ -0,0 +1,26 @@
+// This is a workaround for generating changelog without bumping
```
JamesBurnside, line 1: would be good to have a couple smoke unit-tests for this script

JamesBurnside: (but realize it's a non-trivial amount of work given the directory this lives in)

PR #1403, opened 2022-01-31
Fix callAdapter.startCall method

JamesBurnside:
# What
* Ensure correct state and context variables are being set and triggered to update after starting a call
* Refactor joinCall to ensure both methods call the same functions

# Why
If you try to use adapter.startCall you were stuck in configuration screen. This was because things like callID were never set so state wasn't being updated all over the place.

# How Tested
Manually set start call button in meeting composite to start a call with a known guid instead of joining a new group call.


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -246,25 +247,18 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
+        ? this.callAgent.join(this.locator as TeamsMeetingLinkLocator, {
+            audioOptions,
+            videoOptions
+          })
+        : this.callAgent.join(this.locator as GroupCallLocator, {
+            audioOptions,
+            videoOptions
+          });
+
+      this.updateContextAfterJoiningCall(call);
```
prprabhu-ms, line 260: Naming nit: This method does more than upate `this.context`. How about `processNewCall`, or `setupNewCall`

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -368,14 +362,34 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
   public startCall(participants: string[]): Call | undefined {
-    const idsToAdd = participants.map((participant) => {
-      // FIXME: `onStartCall` does not allow a Teams user.
-      // Need some way to return an error if a Teams user is provided.
-      const backendId = fromFlatCommunicationIdentifier(participant) as CommunicationUserIdentifier;
-      return backendId;
-    });
+    if (_isInCall(this.getState().call?.state ?? 'None')) {
+      throw new Error('You are already in the call.');
+    } else {
```
prprabhu-ms, line 367: nit: Drop the `else` clause. The `if` branch has an early return.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -368,14 +362,34 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
+      const idsToAdd = participants.map((participant) => {
+        // FIXME: `onStartCall` does not allow a Teams user.
+        // Need some way to return an error if a Teams user is provided.
+        const backendId = fromFlatCommunicationIdentifier(participant) as CommunicationUserIdentifier;
+        return backendId;
+      });
+
+      const call = this.handlers.onStartCall(idsToAdd);
```
prprabhu-ms, line 375: I'm a bit surprised that all these methods are synchronous. Doesn't starting a call take some time?
Same for the various operations inside `updateContextAfterJoiningCall`.

prprabhu-ms (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
JamesBurnside: I agree, callAgent.startCall and callAgent.join are the surprising ones to be synchronous -- but they are 🤷‍♀️ (most surprising is that is gets the call id for the call to join without being asynchronous)

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
JamesBurnside: updated

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
JamesBurnside: I like processNewCall -- updated

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
prprabhu-ms: Pro-tip: `callId` can change during a call -- exactly for this reason. The client picks one, but the server can assign a different one once connected.

anjulgarg (approved):

PR #1406, opened 2022-02-01
Fix PiP styling and Add option to Video Tile to set personal max and min sizing

JamesBurnside:
# What

* Add API prop to allow setting video tile min and max persona avatar size
* Fix PiP styling
  * For the secondary tile when video is not enabled, set a better minimum avatar size
  * Fix the width

# Why

* After changing to `<aside>` the width was spanning the whole container it was in and not fitting to content size
* Secondary tile was too big before (see before/after below)

# How Tested
Tile icon size:
| before | after |
| -- | -- |
| image | image |


dmceachernmsft (approved):

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
```diff
@@ -87,6 +87,7 @@ const handleKeyDown = (e: React.KeyboardEvent<HTMLElement>, onClickCallback: ()
 
 const tileContainerStyles: React.CSSProperties = {
```
dmceachernmsft, line 88: do we have any guidelines on what types these defined styles should be? There seems to be a lot of choices depending on what kind of fluent component you are dealing with

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -77,20 +77,31 @@ export interface VideoTileProps {
...
    */
   displayName?: string;
-  /** Name of the participant used to generate initials. For example, a name `John Doe` will display `JD` as initials.
+  /**
+   * Name of the participant used to generate initials. For example, a name `John Doe` will display `JD` as initials.
    * @remarks `displayName` is used if this property is not specified.
    */
   initialsName?: string;
+  /**
+   * Minimum size of the personal avatar.
```
prprabhu-ms, line 86: What unit is this in?

Should we include the unit in the name of the field (since we don't have it in the type). So `personaMinSizePx`.

[...some time later...]

Phew... jumped through 5 components to understand exactly what UI element this sizes. It's not obvious at all from the props here what the effect of setting it will be. We should add more details to docstring to make this usable.

"Sets the minimum size in pixel of the Persona coin shown when no video stream is available"

tbh, even that isn't clear to me (what is "size of a coin"? radius?, but `AvatarPersona` doesn't do a good job of documenting this either).

prprabhu-ms (approved):
No need for a roundtrip.

packages/react-components/src/components/VideoTile.tsx
JamesBurnside: updated docstring

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
JamesBurnside: In short: no. But as long as it's strongly typed we don't have a strong bias against one method or another

anjulgarg (approved):

PR #1407, opened 2022-02-01
Create tooltip: read by x of x 

carocao-msft:
# What
get read receipt information and group them according to chatmessageid, 
length of read receipt should be the number of people who have read the message.
For number of participants, length of object coming from getParticipants - 1 should be the number of participants 

[USER STORY 2689318](https://skype.visualstudio.com/SPOOL/_workitems/edit/2689318)

# Why
Show user how many people have read the message
Screen Shot 2022-02-03 at 9 36 42 PM


packages/chat-stateful-client/src/StatefulChatThreadClient.ts
```diff
@@ -105,6 +105,7 @@ class ProxyChatThreadClient implements ProxyHandler<ChatThreadClient> {
         return createDecoratedListParticipants(chatThreadClient, this._context);
       }
       case 'listReadReceipts': {
+        // check here
```
alkwa-msft, line 108: can you clarify this comment?

packages/chat-stateful-client/src/iterators/createDecoratedListReadReceipts.ts
```diff
@@ -14,6 +14,7 @@ export const createDecoratedListReadReceipts = (
   context: ChatContext
 ): ((options?: RestListReadReceiptsOptions | undefined) => PagedAsyncIterableIterator<ChatMessageReadReceipt>) => {
   const setReadReceipt = (readReceipt: ChatMessageReadReceipt, context: ChatContext): void => {
+    //check here
```
alkwa-msft, line 17: I find some of these comments arent really descriptive and they are creating more questions than anything. Lets make sure that if we write a comment it creates value for anyone reading it.

packages/chat-stateful-client/src/StatefulChatThreadClient.ts
carocao-msft: Gonna remove that too sorry

packages/chat-stateful-client/src/iterators/createDecoratedListReadReceipts.ts
carocao-msft: Sorry I left some comments for myself to debug, forgot to remove them, I will go through the changes again to make sure no debugging comments are left

packages/react-components/src/components/MessageStatusIndicator.tsx
```diff
@@ -105,7 +107,7 @@ export const MessageStatusIndicator = (props: MessageStatusIndicatorProps): JSX.
       );
     case 'seen':
       return (
-        <TooltipHost content={strings.seenTooltipText}>
+        <TooltipHost content={`Read by ${numRead} of ${numParticipants}`}>
```
JamesBurnside, line 110: This string will need localized

packages/react-components/src/types/ChatMessage.ts
```diff
@@ -60,8 +74,9 @@ export interface ChatMessage extends MessageCommon {
   mine?: boolean;
   clientMessageId?: string;
   contentType: MessageContentType;
+  readReceipts?: ReadReceipt[];
+  numParticipants?: number;
```
JamesBurnside, line 78: Add a doc comment to this with its purpose
```suggestion
/**
 * Total number of participants in the message thread. This is used to display how many participants out of a total have read a message.
 */
 numParticipants?: number;
```

Suggestion: don't shorten num in the public api `numParticipants` --> `numberOfParticipants` .. though this doesn't convey a lot of meaning -- number of participants of what? 
Consider an alternative name like `messageThreadParticipantCount`

carocao-msft: Updated

packages/react-components/src/components/MessageStatusIndicator.tsx
carocao-msft: updated.

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -22,7 +22,7 @@
     "deliveredAriaLabel": "Message sent",
     "deliveredTooltipText": "Sent",
     "seenAriaLabel": "Message seen by others",
-    "seenTooltipText": "Seen",
+    "seenTooltipText": "Read by {messageThreadReadCount} of {messageThreadParticipantCount}",
```
prprabhu-ms, line 25: Changing the string in a big way like this might be considered behaviorally breaking -- if Contoso had their own translations (which they could), those will break (the app will throw an error with trying to do string interpolation).

We should consider introducing a new string.

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -119,30 +146,50 @@ const hasValidParticipant = (chatMessage: ChatMessageWithStatus): boolean =>
  * @public
  */
 export const messageThreadSelector: MessageThreadSelector = createSelector(
-  [getUserId, getChatMessages, getLatestReadTime, getIsLargeGroup],
-  (userId, chatMessages, latestReadTime, isLargeGroup) => {
+  [getUserId, getChatMessages, getLatestReadTime, getIsLargeGroup, getReadInformation, getParticipants],
+  (userId, chatMessages, latestReadTime, isLargeGroup, readInformation, participants) => {
```
PorterNan, line 150: I would still name this as readReceipt cause it would be good to reflect the type using name(unless it is not possible) - I see why you give it a new name because you have another variable named the same - 
I would say 2 ways of naming - 
1. name parameter this as allReadReceipts, then keep the second one
2. name this parameter as readReceipts then rename the variable as readReceiptForMessage

Personally I prefer second one cause your basic selector getReadInformation(which might be renamed as well) would be reused in the future - so name it as getReadReceipts sounds more readable

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -119,30 +146,50 @@ const hasValidParticipant = (chatMessage: ChatMessageWithStatus): boolean =>
...
+    // get number of participants
+    const messageThreadParticipantCount = Object.values(participants).length - 1;
+    // for each chat message, add read receipt information when readInformation.id === chatmessage.id
+    const convertedChatMessages = Object.values(chatMessages).map(function (message) {
+      const readReceipt = Object.values(readInformation).filter((info) => info.chatMessageId === message.id);
```
PorterNan, line 155: readReceipt**s** if this is an array

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -119,30 +146,50 @@ const hasValidParticipant = (chatMessage: ChatMessageWithStatus): boolean =>
...
+      const uniqueReadReceipt = readReceipt.filter(
+        (value, index, self) =>
+          index ===
+          self.findIndex(
+            (t) =>
+              toFlatCommunicationIdentifier(t.sender) === toFlatCommunicationIdentifier(value.sender) &&
+              t.chatMessageId === value.chatMessageId
+          )
+      );
+      const convertedMessage = { message, readReceipt: uniqueReadReceipt };
```
PorterNan, line 166: And readReceipts + uniqueReadReceipts as awalys

packages/react-components/src/localization/locales/en-US/strings.json
PorterNan: Yes, we might need a new string here - this seenTooltipText would be a fallback for the readReceipt indicator under these scenarioes:
1. when Readby 0 of 0 (sometimes we failed to get the participantNum because of permission issues, use this 'seen' string to avoid showing some weird result)
2. People might not want readReceipt details all the time, so we might want to open some flags in future PR to disable this details feature - so keep this 'seen' string makes sense

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -67,6 +67,37 @@ const getLatestChatMessage = (messages: (ChatMessage | SystemMessage | CustomMes
...
+  }
+  return 1;
+};
+
+/**
+ * Get the latest read number for my last sent message from the message array.
+ *
+ * @param messages
+ */
+const getLatestReadNum = (messages: (ChatMessage | SystemMessage | CustomMessage)[]): number => {
```
PorterNan, line 90: Use Message type from packages\react-components\src\types\ChatMessage.ts instead

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -119,30 +146,50 @@ const hasValidParticipant = (chatMessage: ChatMessageWithStatus): boolean =>
...
-  (userId, chatMessages, latestReadTime, isLargeGroup) => {
+  [getUserId, getChatMessages, getLatestReadTime, getIsLargeGroup, getReadInformation, getParticipants],
+  (userId, chatMessages, latestReadTime, isLargeGroup, readInformation, participants) => {
+    // get number of participants
+    const messageThreadParticipantCount = Object.values(participants).length - 1;
+    // for each chat message, add read receipt information when readInformation.id === chatmessage.id
+    const convertedChatMessages = Object.values(chatMessages).map(function (message) {
+      const readReceipt = Object.values(readInformation).filter((info) => info.chatMessageId === message.id);
+      // readReceipt is getting duplicate information, remove duplicate here
+      const uniqueReadReceipt = readReceipt.filter(
```
PorterNan, line 157: Comment for following work items after this PR:
We might need to implement readReceipt logic here for readReceipt contextual menu - once we detect one message has been read by someone - then we assume that person read all the message above
Which means we might wanna change data structure a little bit for the next version

we might need a lastReadLine or lastReadMessage for each person, when we render readReceipt just follow the assumption we made

This is just a thought (and notes) for next work item, let's have more discussion after we finish this one

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -67,6 +67,37 @@ const getLatestChatMessage = (messages: (ChatMessage | SystemMessage | CustomMes
   return undefined;
 };
 
+/**
+ * Get the latest participant number from the message array.
+ *
+ * @param messages
+ */
+const getLatestParticipantNum = (messages: (ChatMessage | SystemMessage | CustomMessage)[]): number => {
```
PorterNan, line 75: Do we still need this? Since we are getting participantNum directly from selector according to what @prprabhu-ms commented

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -67,6 +67,37 @@ const getLatestChatMessage = (messages: (ChatMessage | SystemMessage | CustomMes
...
+ * @param messages
+ */
+const getLatestReadNum = (messages: (ChatMessage | SystemMessage | CustomMessage)[]): number => {
+  for (let i = messages.length - 1; i >= 0; i--) {
+    const message = messages[i];
+    if (message.messageType === 'chat' && message.mine && message.status === 'seen') {
+      // first find latest message that belongs to me, return read number for that message
+      return message.readReceipts ? message.readReceipts.length + 1 : 1;
+    }
+  }
```
PorterNan, line 97: The assumption we made and the logic here is a little bit complex and easy to be missed - we can't just return message.ReadReceipt number and say this is our read count - cause people wouldn't send readReceipt for each message - all we need to do is to find one person's latest read line then determine whether a message has been read or not

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -681,21 +749,36 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
...
-
     try {
       if (
         onMessageSeen &&
         lastMessage &&
         lastMessage.messageId &&
         lastMessage.messageId !== messageIdSeenByMeRef.current
       ) {
-        await onMessageSeen(lastMessage.messageId);
+        // lastMessageList contains every participant's last sent messages
```
PorterNan, line 759: As what we discussed with Chi, we should keep the previous logic in order not to hit throttling 

Lemme explain the logic a little bit again - 
1. People will only send readReceipt to the lastest message in the bottom of thread
2. We will calculate the read line of a person then assume he/she read all the messages above
3. Follow the logic #2 calculate the read count and render it in messageThread

#2 should be done either in messageThread or selector, but having it in selector will make it easier for next contextual menu work item

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -41,9 +59,16 @@ const convertToUiChatMessage = (
   message: ChatMessageWithStatus,
   userId: string,
   isSeen: boolean,
-  isLargeGroup: boolean
+  isLargeGroup: boolean,
+  readReceipts: ChatMessageReadReceipt[],
+  messageThreadParticipantCount: number
```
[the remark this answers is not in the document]
carocao-msft: Updated PR

packages/react-components/src/localization/locales/en-US/strings.json
carocao-msft: Updated PR

packages/chat-component-bindings/src/messageThreadSelector.ts
carocao-msft: Updated PR

packages/chat-component-bindings/src/messageThreadSelector.ts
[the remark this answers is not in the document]
carocao-msft: Updated PR

carocao-msft: Updated PR

packages/react-components/src/components/MessageThread.tsx
carocao-msft: Removed this function as it's not needed anymore

packages/chat-component-bindings/src/messageThreadSelector.ts
carocao-msft: Updated PR with new logic

packages/react-components/src/components/MessageThread.tsx
carocao-msft: removed

packages/react-components/src/components/MessageThread.tsx
carocao-msft: Updated PR with new logic

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -136,22 +169,47 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
...
+          /** logic: Looking at message A, how do we know it's read number?
+           * Assumption: if user read the latest message, user has read all messages before that
+           * ReadReceipt behaviour: read receipt is only sent to the last message
+           * if participant read a message that is sent later than message A, then the participant has read message A
+           * how do we check if the message is sent later than message A?
+           * we compare if the readon timestamp is later than the message A sent time
+           * if last read message id is not equal to message A's id, check the read on time stamp.
+           * if the last read message is read after the message A is sent, then user should have read message A as well */
+
+          let readNumber = 0;
```
PorterNan, line 182: This logic is working well to calculate the readNum, thanks for implementing this part! In addition, there could be some perf improvements but not necessary for this PR - so we don't need 2 loops O(n2) to get readNum:
1. Iterate messages from bottom to top and the last readMessage id for users, remember each last read Message Id for each user
2. when calculate readNum using this lastMessageID, just need to do O(20*n) because max participants number of enable readReceipt will be only 20

This improved logic doesn't have to be in this PR cause this one is working very well. Let's open another following PR after this one gets checked in

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -119,8 +130,30 @@ const hasValidParticipant = (chatMessage: ChatMessageWithStatus): boolean =>
  * @public
  */
 export const messageThreadSelector: MessageThreadSelector = createSelector(
-  [getUserId, getChatMessages, getLatestReadTime, getIsLargeGroup],
-  (userId, chatMessages, latestReadTime, isLargeGroup) => {
+  [getUserId, getChatMessages, getLatestReadTime, getIsLargeGroup, getReadReceipts, getParticipants],
+  (userId, chatMessages, latestReadTime, isLargeGroup, readReceipts, participants) => {
+    // get number of participants
+    const messageThreadParticipantCount = Object.values(participants).length - 1;
```
PorterNan, line 136: So this is the participantCount but not includes myself, I am wondering if we could give this a better name to indicate this:
When I am looking at this name, I was expecting it includes myself. Could be 2 solutions - 
1. give it a new name
2. not to -1 for this value, but -1 in messageThread component when rendering it

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -136,22 +169,47 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
...
+           * Assumption: if user read the latest message, user has read all messages before that
+           * ReadReceipt behaviour: read receipt is only sent to the last message
+           * if participant read a message that is sent later than message A, then the participant has read message A
+           * how do we check if the message is sent later than message A?
+           * we compare if the readon timestamp is later than the message A sent time
+           * if last read message id is not equal to message A's id, check the read on time stamp.
+           * if the last read message is read after the message A is sent, then user should have read message A as well */
+
+          let readNumber = 0;
+          for (const k in readReceiptForEachMessage) {
```
PorterNan, line 183: This part of logic is a little bit complicated, would like to see some UTs(in the future PRs) when we determine the whole logic, which will help us ensure no future bugs when someone is changing the code

packages/react-components/src/components/MessageStatusIndicator.tsx
```diff
@@ -105,7 +108,16 @@ export const MessageStatusIndicator = (props: MessageStatusIndicatorProps): JSX.
       );
     case 'seen':
       return (
-        <TooltipHost content={strings.seenTooltipText}>
+        <TooltipHost
+          content={
+            messageThreadReadCount === 1 || messageThreadParticipantCount === 1
```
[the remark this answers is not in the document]
carocao-msft: If participants are more than 1, message is at seen status, but the read count is only 1, it means we have a bug (read count is defaulted to be 1 since I always read my own message)

packages/chat-component-bindings/src/messageThreadSelector.ts
carocao-msft: https://skype.visualstudio.com/SPOOL/_workitems/edit/2762971 
Created work item for this

packages/chat-component-bindings/src/messageThreadSelector.ts
carocao-msft: created work item for this 
https://skype.visualstudio.com/SPOOL/_workitems/edit/2762970

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -136,22 +169,47 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
             message.clientMessageId !== undefined
         )
         .filter((message) => message.content && message.content.message !== '') // TODO: deal with deleted message and remove
-        .map((message) =>
-          memoizedFn(
+        .map((message) => {
+          /** logic: Looking at message A, how do we know it's read number?
```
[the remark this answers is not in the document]
carocao-msft: Created work item https://skype.visualstudio.com/SPOOL/_workitems/edit/2762970

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -868,8 +882,20 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
...
+    messageThreadReadCount: number,
+    messageThreadParticipantCount: number
+  ) => JSX.Element = useCallback(
+    (status: MessageStatus, messageThreadReadCount: number, messageThreadParticipantCount: number) => (
+      <MessageStatusIndicator
+        status={status}
+        messageThreadReadCount={messageThreadReadCount}
+        // if messageThreadParticipantCount is less than or equal to 0, we have a bug, set it to 0
+        // -1 because participant count does not include myself
+        participantCountNotIncludingSelf={messageThreadParticipantCount > 0 ? messageThreadParticipantCount - 1 : 0}
```
PorterNan, line 896: This logic could be covered in packages/react-components/src/components/MessageStatusIndicator.tsx
Line 118:
 participantCountNotIncludingSelf <= 0 ||

PorterNan (approved):
Excellent! Thanks for implement such a complicated logic - it's a great example of how we do a bunch of optimizations:
We are making assumption for people that if they read one message then they read all messages above, which reduces a lot of unnecessary requests for sending readReceipt

This is absolutely a precious knowledge we should show to all the team!

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -119,8 +130,22 @@ const hasValidParticipant = (chatMessage: ChatMessageWithStatus): boolean =>
  * @public
  */
 export const messageThreadSelector: MessageThreadSelector = createSelector(
-  [getUserId, getChatMessages, getLatestReadTime, getIsLargeGroup],
-  (userId, chatMessages, latestReadTime, isLargeGroup) => {
+  [getUserId, getChatMessages, getLatestReadTime, getIsLargeGroup, getReadReceipts, getParticipants],
+  (userId, chatMessages, latestReadTime, isLargeGroup, readReceipts, participants) => {
+    // get number of participants
+    // -1 because the first object in participants is not a valid participant (no display name)
+    const messageThreadParticipantCount = Object.values(participants).length - 1;
```
PorterNan, line 137: The no displayname user is the moderator from our sample app server - maybe -1 is not generic enough, let's filter all empty displayName participant either here or in component

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -136,22 +161,46 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
...
-          memoizedFn(
+        .map((message) => {
+          /** logic: Looking at message A, how do we know it's read number?
+           * Assumption: if user read the latest message, user has read all messages before that
+           * ReadReceipt behaviour: read receipt is only sent to the last message
+           * if participant read a message that is sent later than message A, then the participant has read message A
+           * how do we check if the message is sent later than message A?
+           * we compare if the readon timestamp is later than the message A sent time
+           * if last read message id is not equal to message A's id, check the read on time stamp.
+           * if the last read message is read after the message A is sent, then user should have read message A as well */
```
JamesBurnside, line 172: <3 this comment!
This is the sort of caclucation that lends itself very well to unit tests (to easily have a unit test for just this this should be its own fn `calculateReadReceipts(message, latestReadReceiptPerSender)` -- maybe could be added as part of https://skype.visualstudio.com/SPOOL/_workitems/edit/2762970

carocao-msft: Yeah I have a task create for adding unit tests

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -292,11 +292,17 @@ const memoizeAllMessages = memoizeFnAll(
...
+      status: MessageStatus,
+      messageThreadReadCount: number,
+      messageThreadParticipantCount: number
+    ) => JSX.Element,
     defaultChatMessageRenderer: (message: MessageProps) => JSX.Element,
     strings: MessageThreadStrings,
     _attached?: boolean | string,
     statusToRender?: MessageStatus,
+    messageThreadReadCount?: number,
+    messageThreadParticipantCount?: number,
```
JamesBurnside, line 305: Do we take into account anywhere the fact read receipts stop only work up to 20 participants?
https://docs.microsoft.com/en-us/azure/communication-services/concepts/chat/concepts#:~:text=readReceiptReceived%20-%20when%20another%20participant%20sends%20a%20read,thread%20is%20deleted%20by%20a%20Communication%20Services%20user.
image

JamesBurnside: Or better question - what is expected to be shown when there are more than 20 participants?

packages/chat-component-bindings/src/messageThreadSelector.ts
carocao-msft: Added filtering instead of -1

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -136,22 +161,46 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
...
+
+          let readNumber = 0;
+          for (const k in readReceiptForEachSender) {
+            const messageid = readReceiptForEachSender[k]['lastReadMessage'];
+            const readTime = readReceiptForEachSender[k]['readOn'];
+            if (messageid === message.id) {
+              readNumber += 1;
+            } else {
+              if (new Date(readTime) >= new Date(message.createdOn)) {
```
mgamis-msft, line 182: Super minor: the nested if can be pulled out as an else if:
```
else if(new Date(readTime) >= new Date(message.createdOn))
```

mgamis-msft (approved):

packages/react-components/src/components/MessageStatusIndicator.tsx
```diff
@@ -44,6 +46,8 @@ export interface MessageStatusIndicatorStrings {
 export interface MessageStatusIndicatorProps {
   /** Message status that determines the icon to display. */
   status?: MessageStatus;
+  messageThreadReadCount?: number;
```
mgamis-msft, line 49: Minor: some comment for both these new props so the api doesn't complain about this being undocumented.

JamesBurnside (approved):

PR #1409, opened 2022-02-01
[API DESIGN FEEBDACK REQUEST] Place adhoc call to teams user

JamesBurnside:
# Key Changes
* Locator can take an array of participants that are to be called
  * Currently we will only officially support a single Teams user but this API allows future expansion
* New CallAdapter fn `startOrJoinCall` to allow the adapter to decide whether callAgent->startCall or callAgent->joinCall should be called 
  * The adapter implementation decides based on the type of locator whether a call should be started or joined

# Why
For supporting Teams Adhoc 1:1 calls, Flow of how an adhoc call works in this design:
* Contoso's User clicks "Call user(s)" somewhere on their webapp 
  * Contoso's app page knows the participants IDs to call
  * This could be through a Graph search, or Contoso's web app could have this information elsewhere (irrelevant to us)
* Contoso's web app creates an ACSCallAdapter (with locator set) and shows the call composite (starts at configuration page)
* Contoso's user hits start call on configuration page
  * (Adapter calls startOrJoinCall)
* Ringing page shows (same as lobby page)
* Rest of normal call compsite flow follows

# How Tested
Draft PR to get API feedback

### DIFF
--- a/packages/react-composites/review/react-composites.api.md
+++ b/packages/react-composites/review/react-composites.api.md
@@ -85,7 +85,9 @@ export type AzureCommunicationCallAdapterArgs = {
 };
 
 // @public (undocumented)
-export type AzureCommunicationCallAdapterLocator = TeamsMeetingLinkLocator | GroupCallLocator | TeamsUserLocator;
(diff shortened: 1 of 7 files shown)


JamesBurnside:
Will discuss this in Teams, closing PR

PR #1419, opened 2022-02-03
Call composite UI tests based on page state

mgamis-msft:
# What

Call composite UI tests based on page state

# Why

Create snapshot for UI when access is denied to a Teams meeting

# How Tested

Generating UI snapshots with CI and tested that they consistently pass

# Process & policy checklist


packages/react-composites/tests/browser/call/PageState.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import { expect } from '@playwright/test';
+import { clickOutsideOfPage, waitForPageFontsLoaded, waitForSelector, dataUiId } from '../common/utils';
+import { test } from './fixture';
+import { buildUrlWithMockAdapter } from './utils';
+
+test.describe('Page state tests', async () => {
+  test('Page when waiting in lobby', async ({ pages, serverUrl }) => {
+    const page = pages[0];
+    await page.goto(
```
JamesBurnside, line 12: nit: there's a lot of duplicate code per test that isn't really related to the test itself (and would need updated everywhere if we needed to add a new condition/updated a fn name), consider having a helper function:
```ts
gotoPageWithMockAdapter = (qArgs) => {
  await page.goto(..);
  await clickOutsideOfPage(page);
  await waitForPageFontsLoaded(page);
}
```

JamesBurnside (approved):
So much YES for this!!

packages/react-composites/tests/browser/call/PageState.test.ts
prprabhu-ms: Since you have multiple PRs in flight at add tests. I'd recommend landing these and write a PR that extracts this out from all the `mockAdapter` tests.

prprabhu-ms (approved):

PR #1420, opened 2022-02-03
Calling UFD UI tests

mgamis-msft:
# What

Added UFD related UI tests for Calling Composite

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2686848

# How Tested

Generating UI snapshots with CI and tested that they consistently pass

# Process & policy checklist


packages/react-composites/tests/browser/call/app/index.tsx
```diff
@@ -47,7 +47,7 @@ function App(): JSX.Element {
   useEffect(() => {
     const initialize = async (): Promise<void> => {
       if (mockCallState) {
-        setCallAdapter(wrapAdapterForTests(new MockCallAdapter(mockCallState)));
```
mgamis-msft, line 50: Needed to remove this because [this wrapper unsets the 'speakingWhileMicrophoneIsMuted' diagnostic](https://github.com/Azure/communication-ui-library/blob/a6c7201daab7a6545520a11de24a1e7611197a5e/packages/react-composites/tests/browser/call/app/index.tsx#L110).

packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
```diff
@@ -29,7 +29,7 @@ test.describe('HorizontalGallery tests', async () => {
         remoteParticipants: testRemoteParticipants
       })
     );
-    // Click off page to turn away initial aria label
+    // Click outside of screen to turn away initial aria label
```
prprabhu-ms, line 32: To be honest, this comment no longer adds that much value beyond the function named `clickOutsideOfPage`. Yay for well named helper functions!

We can just delete (all instances of) this comment.

packages/react-composites/tests/browser/call/UFD.test.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation.
```
prprabhu-ms, line 1: Let's name this `UserFacingDiagnostics.test.ts`. The acronym `UFD` might be confusing without context.

packages/react-composites/tests/browser/call/UFD.test.ts
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT license.
+
+import { expect } from '@playwright/test';
+import { dataUiId, clickOutsideOfPage, waitForPageFontsLoaded, waitForSelector } from '../common/utils';
+import { test } from './fixture';
+import { buildUrlWithMockAdapter } from './utils';
+import { DiagnosticQuality } from './TestCallingState';
+
+test.describe('UFD tests', async () => {
```
prprabhu-ms, line 10: ditto -> User Facing Diagnostics tests

packages/react-composites/tests/browser/call/UFD.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  test('UFD test when speaking while muted ', async ({ pages, serverUrl }) => {
```
prprabhu-ms, line 11: nit: Describe what the test does "A banner is shown when user is speaking while muted"

And same for other tests. These test descriptions should be sensible when reading on the console. They should give us a reasonable idea of what failed without looking at the test code.

packages/react-composites/tests/browser/call/app/index.tsx
prprabhu-ms: Yes, that is necessary in live e2e tests because the isMuted diagnostic is flakily set in our test setup.

It makes sense that it isn't needed for mockAdapter.

prprabhu-ms:
The snapshot update seems wrong: 

image

Our logic to replace actual timestamps with `"timestamp"` is needed so that the snapshots aren't flaky.

packages/react-composites/tests/browser/call/TestCallingState.ts
```diff
@@ -19,3 +20,39 @@ export type TestRemoteParticipant = {
   isVideoStreamAvailable?: boolean;
   isScreenSharing?: boolean;
 };
+
```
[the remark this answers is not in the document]
mgamis-msft: Yes, unfortunately. I cannot seem to import modules from our Typescript modules because in playwright automatically transforms TypeScript to JavaScript and they mention it may not transpile correctly. 
Fortunately, if there are any breaking changes in the CallingSDK types then the test browser app will show errors since the `TestCallingState` is used to form the `CallAdapterState` in `MockCallAdapter`

mgamis-msft:
> The snapshot update seems wrong:
> 
> image
> 
> Our logic to replace actual timestamps with `"timestamp"` is needed so that the snapshots aren't flaky.

Where did you see this snapshot update? I don't see it in the 'Files Changed' section.

packages/react-composites/tests/browser/call/HorizontalGallery.test.ts
mgamis-msft: Will do all refactor tasks in a separate PR :)

JamesBurnside (approved):

PR #1423, opened 2022-02-03
Update meeting composite adapter args - accept just a teams url

JamesBurnside:
# What
* Update ACSMeetingAdapter API to have meetingArgs that can be one of two things:
  1. call locator + chat thread id
  2. teams meeting url
* Use util function that extracts chat thread id from teams url (hopefully we can remove this at a later date if chat can be convinced to support a teams url instead of threadId)
  * Add unit tests

# Why
Customer asks to only accept a teams link when joining a teams meeting

# How Tested
Todo:
* Meeting sample
* Storybook basic example and join example


packages/communication-react/review/communication-react.api.md
```diff
@@ -247,6 +246,14 @@ export interface CallAgentState {
     displayName?: string;
 }
 
+// @beta
+export interface CallAndChatMeetingArgs {
+    // (undocumented)
+    callLocator: GroupCallLocator;
+    // (undocumented)
+    chatThreadId: string;
```
prprabhu-ms, line 254: Please document new fields, even for beta.

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -412,23 +422,26 @@ export const createAzureCommunicationMeetingAdapter = async ({
   displayName,
   credential,
   endpoint,
-  chatThreadId,
-  callLocator
+  meetingArgs
 }: AzureCommunicationMeetingAdapterArgs): Promise<MeetingAdapter> => {
-  const callAdapter = await createAzureCommunicationCallAdapter({
+  const locator = 'meetingLink' in meetingArgs ? meetingArgs : meetingArgs.callLocator;
```
prprabhu-ms, line 427: This logic is used in at least two places, and even otherwise I'd request that we add a helper function for readability:

```ts
const isTeamsMeetingLinkLocator(locator: CallAndChatMeetingArgs | TeamsMeetingLinkLocator): locator is TeamsMeetingLinkLocator => {
  return 'meetingLink' in locator;
}

// Use as
const locator = isTeamsMeetingLinkLocator(meetingArgs) ? meetingArgs : meetingArgs.callLocator;

```

packages/react-composites/src/composites/MeetingComposite/adapter/parseTeamsUrl.test.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { getChatThreadFromTeamsLink } from './parseTeamsUrl';
+
+describe('Get chat thread from teams link', () => {
+  test('Should extract chat thread ID from teams url', async () => {
+    expect(
+      getChatThreadFromTeamsLink(
+        'https://teams.microsoft.com/l/meetup-join/19:meeting_ZTgzM2VmZWEtMXExMC00Yzk3LWE5NjUtYzg5ZDM3OTRjMzk0@thread.v2/0?context=%7B%22Tid%22:%2272f108bf-86f1-41af-91ab-2d7cd011db47%22,%22Oid%22:%222708ac84-aa11-4fab-8d13-e8586a26f6cd%22%7D'
```
prprabhu-ms, line 10: So... this is an active meeting that you're going to keep running, and the first 100 people to join enter a raffle draw?

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -397,8 +408,7 @@ export type AzureCommunicationMeetingAdapterArgs = {
   userId: CommunicationUserIdentifier;
   displayName: string;
   credential: CommunicationTokenCredential;
-  chatThreadId: string;
-  callLocator: TeamsMeetingLinkLocator | GroupCallLocator;
+  meetingArgs: CallAndChatMeetingArgs | TeamsMeetingLinkLocator;
```
prprabhu-ms, line 411: Could we call this `meetingLocator` to be consistent with `callLocator` from `CallComposite` and because "args" is too generic.

prprabhu-ms (approved):

packages/communication-react/review/communication-react.api.md
```diff
@@ -137,8 +137,7 @@ export type AzureCommunicationMeetingAdapterArgs = {
     userId: CommunicationUserIdentifier;
     displayName: string;
     credential: CommunicationTokenCredential;
-    chatThreadId: string;
-    callLocator: TeamsMeetingLinkLocator | GroupCallLocator;
+    meetingArgs: CallAndChatMeetingArgs | TeamsMeetingLinkLocator;
```
alkwa-msft, line 140: we say meetingArgs however we acknowledge we dont want to call it "Meeting" and then we use terms like "CallAndChatMeeting"args. Do we feel this makes sense? 

I think calling has probably had this conversation of Coordinates/Locators, etc. I am curious what they have found here.

Also a thought is can someone repurpose a composite for multiple different calls?

packages/react-composites/src/composites/MeetingComposite/adapter/parseTeamsUrl.test.ts
JamesBurnside: Hahah, I randomly changed a bunch of characters in this

packages/communication-react/review/communication-react.api.md
JamesBurnside: When we do the naming rechange this will like change from `meeting<Suffix>` to something else. I didn't want to put too much thought into it as we haven't firmly decided on the name yet.

Composites will hopefully be used for multiple calls in the future with a single adapter -- but the arguments going into the ACSAdapters are very geared towards the single meeting/call/chat you are planning to join -- this PR follows the same design to keep consistent.

When we do support incoming calls, multiple calls, possibly multiple chat threads we'll likely need a new adapter interface as accepting a single locator at the adapter creation time likely doesn't make sense. We have non-breaking options for this though such as creating a AzureCommunicationMultiCallAdapter or an interface like `switchCall(newlocator)`

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
JamesBurnside: Yeah I don't like args, my worry with locator is chat avoids the term locator -- but I like it more than args so will update

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
JamesBurnside: updated, I'm pleased typescript was able to play well so well with this

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
JamesBurnside: updated

change/@internal-react-composites-6e1eac11-db67-47a8-85f8-1d0788a935a8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "[BREAKING CHANGE] Update createAzureCommunicationMeetingAdapter arguments to enable accepting just a teams link without having to provide an extracted chat thread ID.",
```
anjulgarg, line 3: Why is this a breaking change if you are changing a beta API?

anjulgarg: I understand this would break the beta usage of MeetingComposite. Is that what you meant?

JamesBurnside: Yeah, I wanted to highlight in the beta notes in the changelog this is a very jarring change (customers using meeting composite _will_ be broken)

anjulgarg (approved):

packages/react-composites/src/composites/MeetingComposite/adapter/parseTeamsUrl.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Get chat thread from a teams url.
+ * As per documented on the Chat SDK: https://docs.microsoft.com/en-us/azure/communication-services/quickstarts/chat/meeting-interop?pivots=platform-web#get-a-teams-meeting-chat-thread-for-a-communication-services-user
+ *
+ * @private
+ */
+export const getChatThreadFromTeamsLink = (teamsMeetingLink: string): string => {
+  // Get the threadId from the url - this also contains the call locator ID that will be removed in the threadId.split
+  let threadId = teamsMeetingLink.replace('https://teams.microsoft.com/l/meetup-join/', '');
```
anjulgarg, line 12: Are we certain this will never change?

JamesBurnside: As sure as we can be unfortunately. I'm trying to get the chat sdk to support taking in a teams url. If they do then we will remove this manual extraction (though then the same problem shifts to them).

I discussed this with the calling team (as they do support taking the entire url) and they said they were just OK with taking the risk if it does change and have a relationship with teams to try make sure it doesn't

PR #1429, opened 2022-02-04
Meeting composite Narrator call outs

dmceachernmsft:
# What
Add aria label to new message icon and update string for local camera switcher button

# Why
Improves narrator call outs in the meeting composite.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2698410
https://skype.visualstudio.com/SPOOL/_workitems/edit/2698425
# How Tested
Ran meeting sample locally with narrator to ensure correct call outs.


packages/react-composites/src/composites/MeetingComposite/Strings.tsx
```diff
@@ -29,6 +29,10 @@ export interface MeetingCompositeStrings {
    * Side pane Chat screen title.
    */
   chatPaneTitle: string;
+  /**
+   * New Message label
+   */
+  newMessage: string;
```
JamesBurnside, line 35: nit: newMessage conflicts with the new messages string in the message thread, consider `chatButtonNewMessageLabel`

JamesBurnside: Update the doc comment also to say where exactly this string is used (so Contoso knows what this is used for)

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -9,19 +9,22 @@ import { Icon, IIconStyles, IStackStyles, ITheme, memoizeFunction, Stack, useThe
...
-  const { chatMessagesCount } = props;
+  const { chatMessagesCount, label } = props;
   const theme = useTheme();
   const renderNumber = (numberOfMessages): JSX.Element => {
     if (numberOfMessages < 1) {
       return <></>;
     } else {
       return (
-        <Stack styles={notificationTextStyles(theme)}>{numberOfMessages < 9 ? <>{numberOfMessages}</> : <>9+</>}</Stack>
+        <Stack aria-label={label} role={'alert'} styles={notificationTextStyles(theme)}>
```
JamesBurnside, line 25: Presumably with `alert` the screen reader only announces it once?

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
```diff
@@ -9,19 +9,22 @@ import { Icon, IIconStyles, IStackStyles, ITheme, memoizeFunction, Stack, useThe
...
+          {numberOfMessages < 9 ? <>{numberOfMessages}</> : <>9+</>}
```
JamesBurnside, line 26: Nit: consider wrapping in a `<Text>` and applying the `role={status}`

JamesBurnside: (not sure if this needs a seperate aria label also to say this is the number of unread messages?)

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
dmceachernmsft: with alert it announces whenever there is another new message. do we know if we want to do this? or should I take it out so that it only announces if there is a new message when they enter scan mode?

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
dmceachernmsft: I can ask Alex Pereira about this. I am not too sure myself

packages/react-composites/src/composites/MeetingComposite/NotificationIcon.tsx
JamesBurnside: My thinking is when it goes from 0 new messages -> some new messages there should be an alert announcement,
For 1 new messages -> 3 new messages that would be a `status` role.
If there's an alert everytime there's a new message that would be very distracting to someone trying paying attention to a meeting
For alert the docs say 
> Warning: Because of its intrusive nature, the alert role must be used sparingly and only in situations where the user's immediate attention is required.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Alert_Role
So we need to be careful with this role

packages/react-composites/review/react-composites.api.md
```diff
@@ -752,6 +752,7 @@ export interface MeetingCompositeProps extends BaseCompositeProps<CallCompositeI
 export interface MeetingCompositeStrings {
     chatButtonLabel: string;
     chatPaneTitle: string;
+    newMessage: string;
```
prprabhu-ms, line 755: Can we call this `newMessageLabel` for consistency with `chatButtonLabel` etc.

Even better would be `newMessageNotificationLabel`.

packages/react-composites/src/composites/MeetingComposite/Strings.tsx
```diff
@@ -29,6 +29,10 @@ export interface MeetingCompositeStrings {
    * Side pane Chat screen title.
    */
   chatPaneTitle: string;
+  /**
+   * New Message label for chat button with notification icon component.
+   */
+  newMessageNotificationLabel: string;
```
JamesBurnside, line 35: I think this still needs something that indicates chatButton in the name (https://github.com/Azure/communication-ui-library/pull/1429#discussion_r800926611).
`newMessageNotificationLabel` could be anywhere on the screen -- the doc comment definitely helps but if we were to implement the chat bubbles it's not clear if newMessageNotification is related to the chat button or the chat bubbles (and still conflicts with the message thread "New Messages" button which could be seen as a notification)
image
image


```suggestion
  chatButtonNewMessageNotificationLabel: string;
```

JamesBurnside (approved):
Not blocking on open comment regarding name - just a suggestion for better clarity to developers

PR #1430, opened 2022-02-04
Add a sample / cheet-sheat of conditional compilation directives

prprabhu-ms:
# What

An internal module showing common conditional compilation patterns.

# Why

It's not always obvious how to add conditional compilation directives for what we want.
This module also provides a playground for noting down gotchas / missing features, or trying out an idea that we think _should_ work.

# How Tested

* Build for beta flavor
* Build for stable flavor and inspect the preprocessed file to verify it matches expectations.

We might want to evolve this in the future as a check to validate that the preprocessing step works as expected, by checking in the preprocessed file and comparing with the stable flavor generated file.

### DIFF
--- a/packages/acs-ui-common/src/conditional-compilation-sample/index.tsx
+++ b/packages/acs-ui-common/src/conditional-compilation-sample/index.tsx
@@ -114,7 +114,7 @@ export function d(e: number, /* @conditional-compile-remove-from(stable) */ f: n
  */
 
 /**
- * Add a parameter to an existing function
+ * Call a function with conditional parameters.
  *
  * FIXME: This doesn't yet work. `f` is not removed from the function signature, although it is removed from the function call in the body.
  */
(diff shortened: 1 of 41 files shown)


dmceachernmsft (approved):

prprabhu-ms:
Thanks @anjulgarg This was his idea originally.

packages/acs-ui-common/src/conditional-compilation-sample/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+  props: DummyProps
+) => {
+  memoizedA: boolean;
+  memoizedB: boolean;
+  /* @conditional-compile-remove-from(stable) */
+  memoizedC: boolean;
+};
+
+/**
+ * Selector for {@link MicrophoneButton} component.
```
prprabhu-ms, line 179: Heh, now you know which selector I copied from to start this off...

alkwa-msft (approved):

PR #1431, opened 2022-02-07
Update adapters to make adhoc calls directly to participantIDs

JamesBurnside:
# What
* [beta] Add interface to ACSCallAdapter and ACSMeetingAdapter to support making adhoc calls directly to participantIDs
  * The change here is to add participantIDs as a new locator
* [beta] Update joinCall to call startCall if it's an adhoc call
  * The alternative to this is to add an more confusing startOrJoinCall API, see [example](https://github.com/Azure/communication-ui-library/pull/1409/files#diff-e45d7e6861410e5848963c2e06fdcd6931f885ebb719491f3559649b85b64015R375)

Other changes
* Fixed tiny eslint warning in meeting sample

# Why
Support for Adhoc Calls

# How Tested
Different branch I have a meeting sample that can place 1:1 adhoc calls


packages/communication-react/review/communication-react.api.md
```diff
@@ -249,7 +254,7 @@ export interface CallAgentState {
 
 // @beta
 export interface CallAndChatLocator {
-    callLocator: GroupCallLocator;
+    callLocator: GroupCallLocator | /* @conditional-compile-remove-from(stable) TEAMS_ADHOC_CALLING */ CallParticipantLocator;
```
prprabhu-ms, line 257: Looks like `CallAndChatLocator` is already marked `@beta`, do you need to add this type conditionally?

packages/communication-react/review/communication-react.api.md
```diff
@@ -1091,7 +1101,11 @@ export const DEFAULT_COMPOSITE_ICONS: {
     OptionsCamera: JSX.Element;
     OptionsMic: JSX.Element;
     OptionsSpeaker: JSX.Element;
-    ParticipantItemMicOff: JSX.Element;
+    ParticipantItemMicOff: JSX.Element; /**
+    * Icons that can be overridden for {@link CallComposite}.
```
prprabhu-ms, line 1105: This comment seems to be in a weird spot.

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -395,7 +398,9 @@ export class AzureCommunicationMeetingAdapter implements MeetingAdapter {
  */
 export interface CallAndChatLocator {
   /** Locator used by {@link createAzureCommunicationMeetingAdapter} to locate the call to join */
-  callLocator: GroupCallLocator;
+  callLocator:
+    | GroupCallLocator
+    | /* @conditional-compile-remove-from(stable) TEAMS_ADHOC_CALLING */ CallParticipantLocator;
```
prprabhu-ms, line 403: As noted before, everything under `MeetingComposite` is beta only, so no need for conditional compilation here.

OTOH, this is related to a different feature -- and we may not stabilize TEAMS_ADHOC_CALLING at the same time.

So I reverse my vote -- keep this conditional compilation directive.

prprabhu-ms (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -525,6 +531,34 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
+ *
+ * @example
+ * ```
+ * ['8:orgid:ab220efe-5725-4742-9792-9fba7c9ac458']
+ * ```
+ *
+ * @beta
+ */
+export type CallParticipantLocator = {
+  participantIDs: [string];
```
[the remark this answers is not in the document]
JamesBurnside: Oop thanks! Very bad typo

packages/communication-react/review/communication-react.api.md
JamesBurnside: Currently we don't do any beta-type trimming (https://api-extractor.com/pages/configs/api-extractor_json/#dtsrollupbetatrimmedfilepath) so our stable has all the beta types.

Do we want to conditionally compile out types, or allow them but have the beta trimming remove them?
cc @PorterNan

JamesBurnside: Yeah, different feature (`CallAndChatLocator` is for meetings, but `CallParticipantLocator` is for adhoc calls)

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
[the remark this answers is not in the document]
JamesBurnside: Fixed!

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
JamesBurnside: Yeah different feature (so when meetings composite goes stable this will still be beta)

packages/communication-react/review/communication-react.api.md
JamesBurnside: This is super strange, it's not related to this prop at all and this file isn't supposed to have block comments -- but manually removing this and rebuilding locally adds it back in (checking in CI now). Looks like an api-extractor bug but trying to figure out what in this PR has caused it...

JamesBurnside: OK is it more good or less good that my local build was the one introducing this.. might be time to nuke the local clone

mgamis-msft (approved):

packages/communication-react/review/communication-react.api.md
PorterNan: image
This is how it get tripping out (conditional compile won't work with incomplete sentences/expressions), it just keeps the | operator

packages/communication-react/review/communication-react.api.md
```diff
@@ -449,6 +454,11 @@ export type CallParticipantListParticipant = ParticipantListParticipant & {
     isSpeaking?: boolean;
 };
 
+// @beta
+export type CallParticipantLocator = {
```
anjulgarg, line 458: CallParticipantsLocator?

Reading the type name I assumed it can contain only 1 participant id.

packages/communication-react/review/communication-react.api.md
PorterNan: How would this look like when it is in stable build? It looks like it is not a complete sentence if we doing conditional compiling like this:
callLocator: GroupCallLocator | [nothing];

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -580,3 +614,8 @@ export const createAzureCommunicationCallAdapterFromClient = async (
 const isCallError = (e: Error): e is CallError => {
   return e['target'] !== undefined && e['innerError'] !== undefined;
 };
+
+/* @conditional-compile-remove-from(stable) TEAMS_ADHOC_CALLING */
+const isAdhocCall = (callLocator: CallAdapterLocator): callLocator is CallParticipantLocator => {
```
anjulgarg, line 619: Mark is at @private so we don't accidentally export it.

packages/communication-react/review/communication-react.api.md
PorterNan: One thing might be good to have a feedback - @JamesBurnside  @prprabhu-ms, do we have a good way to use complete sentence to strip these out - if not, maybe that is an indicator that, type union would be something we might wanna support

packages/communication-react/review/communication-react.api.md
```diff
@@ -449,6 +454,11 @@ export type CallParticipantListParticipant = ParticipantListParticipant & {
...
+    participantIDs: string[];
```
PorterNan, line 459: Currently we support property strip like this:
export type CallParticipantLocator = {
   /* @conditional-compile-remove-from(stable) TEAMS_ADHOC_CALLING */
    participantIDs: string[];
}

This could be a work around to not breaking the current compiling system

packages/communication-react/review/communication-react.api.md
JamesBurnside: This should strip out the `|` as well, @prprabhu-ms just recently added this: https://github.com/Azure/communication-ui-library/pull/1427

PorterNan: Ohh, did not know about this - that is great since we are supporting it now!

packages/communication-react/review/communication-react.api.md
```diff
@@ -198,6 +198,11 @@ export interface CallAdapterDeviceManagement {
     setSpeaker(sourceInfo: AudioDeviceInfo): Promise<void>;
 }
 
+// Warning: (ae-incompatible-release-tags) The symbol "CallAdapterLocator" is marked as @public, but its signature references "CallParticipantLocator" which is marked as @beta
```
PorterNan, line 201: Looks like we have a warning here for CallParticipantLocator

packages/communication-react/review/communication-react.api.md
JamesBurnside: There shouldn't be anything broken, here is the stable API file
image

PorterNan (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
JamesBurnside: For small helper fns not exported out of the file we typically don't require any modifier tags

packages/communication-react/review/communication-react.api.md
JamesBurnside: Good call, updated

PR #1432, opened 2022-02-07
Add documentation on running UI tests locally

JamesBurnside:
# What
UI test documentation

# Why
We needed some


docs/references/ui-tests.md
```diff
@@ -0,0 +1,67 @@
...
+# This example will build calling react-composite tests.
+
+# Navigate to packlet directory
+cd packages/react-composites
+
+# Build the e2e tests
+rushx build:e2e:call
+
+# Set a connection string environment variable
+$env:CONNECTION_STRING = '<connection string>' # This differs per development environment, this command is for powershell
```
dmceachernmsft, line 25: Maybe we should have how to do this step for in code spaces with Bash since a number of members on the team or working that way? I will look into it if you agree that is a good idea!

JamesBurnside: That would be great!

anjulgarg:
We have documentation in React Composites README file as well
communication-ui-library/packages/react-composites/README.md

Let's make sure the content is not contradicting. :)

mgamis-msft (approved):

prprabhu-ms (approved):

JamesBurnside:
> We have documentation in React Composites README file as well communication-ui-library/packages/react-composites/README.md
> 
> Let's make sure the content is not contradictory. :)

Thanks! Updated the readme.md to point to docs folder and sync-ed the two markdown file contents

PR #1442, opened 2022-02-09
Enable azure logging in sample apps

prprabhu-ms:
# Why

Basic debugging support. Set logging level to `'warning'`. This keeps log-spam low, though there's still some UFD related warnings from calling SDK.

For debugging, change to `'info'` or `'verbose'`.


samples/Calling/src/app/App.tsx
```diff
@@ -23,6 +24,8 @@ import { EndCall } from './views/EndCall';
 import { HomeScreen } from './views/HomeScreen';
 import { UnsupportedBrowserPage } from './views/UnsupportedBrowserPage';
 
+setLogLevel('warning');
```
JamesBurnside, line 27: Is this useful to include in the server also? (or just adding it for UFD related warnings)

JamesBurnside (approved):

samples/Calling/src/app/App.tsx
prprabhu-ms: Doesn't make sense on Server. We already use a different logger there. Azure logger is for browser logs primarily.
The good thing is, we might be able to easily send these logs elsewhere (like somewhere on the cloud) if we use this logger.

anjulgarg (approved):

PR #1443, opened 2022-02-09
Log state changes to Azure logger when verbose logging is enabled

prprabhu-ms:
Logs!
image

Gives us preliminary visibility into stateful client's internal state (updates), without being overly chatty -- each log line is related to an actual state change, and only includes the changes made.

### DIFF
--- a/packages/calling-stateful-client/src/CallContext.ts
+++ b/packages/calling-stateful-client/src/CallContext.ts
@@ -25,7 +25,6 @@ import {
   CallErrorTarget,
   CallError
 } from './CallClientState';
-import { CallStateModifier } from './StatefulCallClient';
 
 enableMapSet();
 // Needed to generate state diff for verbose logging.
--- a/packages/calling-stateful-client/src/CallContext.ts
+++ b/packages/calling-stateful-client/src/CallContext.ts
@@ -47,7 +47,7 @@ export class CallContext {
   private _batchMode: boolean;
 
   constructor(userId: CommunicationIdentifierKind, maxListeners = 50) {
-    this._logger = createClientLogger('uilib:calling:context');
+    this._logger = createClientLogger('communication-react:calling-context');
     this._state = {
       calls: {},
       callsEnded: {},
@@ -77,7 +77,8 @@ export class CallContext {
   public modifyState(modifier: (draft: CallClientState) => void): void {
     this._state = produce(this._state, modifier, (patches: Patch[]) => {
       if (getLogLevel() === 'verbose') {
-        this._logger.verbose(`State change: ${JSON.stringify(patches)}`);
+        // Log to `info` because AzureLogger.verbose() doesn't show up in console.
+        this._logger.info(`State change: ${JSON.stringify(patches)}`);
       }
     });
     if (!this._batchMode) {
--- a/packages/calling-stateful-client/src/CallContext.ts
+++ b/packages/calling-stateful-client/src/CallContext.ts
@@ -3,8 +3,9 @@
 
 import { CommunicationIdentifierKind } from '@azure/communication-common';
 import { AudioDeviceInfo, DeviceAccess, DominantSpeakersInfo, VideoDeviceInfo } from '@azure/communication-calling';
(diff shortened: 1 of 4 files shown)


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -24,9 +25,10 @@ import {
   CallErrorTarget,
   CallError
 } from './CallClientState';
-import { CallStateModifier } from './StatefulCallClient';
 
 enableMapSet();
+// Needed to generate state diff for verbose logging.
+enablePatches();
```
JamesBurnside, line 31: is there any perf issues with this? (I see it was maybe removed from being enabled by default in immer since v6 - but couldn't see why that was
image
)

prprabhu-ms: No performance implication. The implication is in bundle size, but it's small.

prprabhu-ms: cf: 

https://immerjs.github.io/immer/installation#pick-your-immer-version

image

JamesBurnside: thanks!

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -70,53 +74,18 @@ export class CallContext {
...
-  public setState(state: CallClientState): void {
-    this._state = state;
+  public modifyState(modifier: (draft: CallClientState) => void): void {
+    this._state = produce(this._state, modifier, (patches: Patch[]) => {
+      if (getLogLevel() === 'verbose') {
+        // Log to `info` because AzureLogger.verbose() doesn't show up in console.
+        this._logger.info(`State change: ${JSON.stringify(patches)}`);
+      }
+    });
     if (!this._batchMode) {
```
anjulgarg, line 84: Wouldn't this always be false now that no other function is setting it?

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -70,53 +74,18 @@ export class CallContext {
     return this._state;
   }
 
-  public setState(state: CallClientState): void {
-    this._state = state;
+  public modifyState(modifier: (draft: CallClientState) => void): void {
```
anjulgarg, line 77: I don't understand how this function handling batching now that we are getting rid of the batch method?

anjulgarg:
How do we ensure that the app is working as expected and this doesn't introduce any unwanted changes. (breaking or performance) @prprabhu-ms 
Any bug bash planned only for this PR?

packages/calling-stateful-client/src/CallContext.ts
prprabhu-ms: Nothing was using `batch()` before this PR either. The only caller of `batch` was `modifyState`, and nobody was calling `modifyState`.

`modifyState` was originally intended to be exposed as part of the public API, but we decided against it pre-GA, so I'm cleaning up that dead code. The `modifyState` I have _introduced_ in this PR is actually just `setState`, but with the call to `produce` folded in.

So, it actually reduces the boilerplate, without *any logical changes whatsoever*.

packages/calling-stateful-client/src/CallContext.ts
prprabhu-ms: Good point. Will remove it in my follow up PR.

prprabhu-ms:
> How do we ensure that the app is working as expected and this doesn't introduce any unwanted changes. (breaking or performance) @prprabhu-ms Any bug bash planned only for this PR?

Not planning for a bug bash for this particular PR. This PR should actually have no impact whatsoever. Short of a mistake on my part, this PR should have no functional effect whatsoever. The PR to look out for is https://github.com/Azure/communication-ui-library/pull/1451 For that one, I joined a call with several participants, and logged the skipped events.

packages/calling-stateful-client/src/CallContext.ts
prprabhu-ms: Done in #1451

alkwa-msft (approved):

PR #1444, opened 2022-02-09
File sharing FileDownloadHandler

anjulgarg:
# What
Add FileDownloadHandler for handling file downloads

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2672670

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
```diff
@@ -0,0 +1,36 @@
...
+ *       fileDownloadHandler: fileDownloadHandler
+ *     }}
+ *   />
+ * )
+ *
+ * ```
+ * @param userId - The {@link CommunicationIdentifierKind} ID of the user uploading the file.
+ * @param fileData - The {@link FileMetadata} containing file `url`, `extension` and `name`.
+ */
+export type FileDownloadHandler = (userId: CommunicationIdentifierKind, fileData: FileMetadata) => Promise<string>;
```
prabhjot-msft, line 36: Not a comment, just a question. Where will we implement the FileDownloadHandler or is Contoso going to handle that.

prabhjot-msft: Also, don't we need threadId as parameter for FileDownloadHandler or is it included in metadata.

prabhjot-msft (approved):

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
anjulgarg: Contoso needs to implement this function.
We don't need the threadId since we are providing a userID already.
How do you think a threadId might be used here?

prabhjot-msft: Don't we need threadId to see which chat thread the message belongs to? User might be active in more than one threads.

anjulgarg: For a file download scenario, contoso doesn't need to know about the thread ID.
They need to know which file is being downloaded and by who.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
```diff
@@ -0,0 +1,35 @@
...
+ *       fileDownloadHandler: fileDownloadHandler
+ *     }}
+ *   />
+ * )
+ *
+ * ```
+ * @param userId - The user ID of the user downloading the file.
+ * @param fileData - The {@link FileMetadata} containing file `url`, `extension` and `name`.
+ */
+export type FileDownloadHandler = (userId: string, fileData: FileMetadata) => Promise<string>;
```
prprabhu-ms, line 35: Can this return a `Promise<URL>` instead?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * A callback function for handling file downloads.
+ * The function needs to return a promise that resolves to a file download URL.
+ * If the promise is rejected, the {@link Error.message} will be used to display an error message to the user.
+ *
+ * Sample Usage:
+ * ```ts
+ * const fileDownloadHandler: FileDownloadHandler = async (userId, fileData) => {
+ *   if (isUnauthorizedUser(userId)) {
+ *     throw new Error('You don’t have permission to download this file.');
```
prprabhu-ms, line 16: This sample code makes me think that `FileDownloadHandler` should have a way of notifying us that download failed, so that we can show the appropriate message in the UI (instead of throwing an exception).

Further, we could provide a small set of error causes that can be returned 

`const fileDownloadFailureReason = 'unauthorized' | 'notFound'` etc.

prprabhu-ms (commented):
Like the direction.

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
anjulgarg: What would be the type of URL?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
anjulgarg: Yes, the current idea is to catch the thrown error via `fileDownloadHandler` and display this error message.

anjulgarg: What do you think about that?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
prprabhu-ms: https://javascript.info/url

JamesBurnside (approved):

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
```diff
@@ -0,0 +1,35 @@
...
+import { FileMetadata } from './FileUpload';
+
+/**
+ * @beta
+ *
+ * A callback function for handling file downloads.
+ * The function needs to return a promise that resolves to a file download URL.
+ * If the promise is rejected, the {@link Error.message} will be used to display an error message to the user.
+ *
+ * Sample Usage:
```
alkwa-msft, line 12: Does it make more sense to have this file sharing usage information as a .md file in the docs directory?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
anjulgarg: If contoso if throwing the error and providing an error message, why do we need to care about localization of that message?
This way allows contoso to display any error message not limited by `fileDownloadFailureReason`

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
anjulgarg: I think there is a lot of value in having descriptive usage eample as part of the docstrings.
We can identify a place to put all of the file sharing documentaion in a single place.
.md file would be a good place to start

packages/react-composites/src/composites/ChatComposite/file-sharing/FileDownloadHandler.ts
prprabhu-ms: I was thinking of showing error messages on the UI (to the end user). We do not show any strings on the UI that are not localized.

For any strings that go on our UI, one of two paths must be followed:
* We own the string -- it is localized
* Contoso owns the string -- we provide a way for Contoso to inject this UI component.

We currently have way (and no plans) for Contoso to inject error bars or similar UX surface, so I was thinking of us showing an error string in the UI (probably close to the FileCard, saying we failed to upload, and why).

prprabhu-ms: Talked IRL. We're showing strings that Contoso gives us. They need to localize if they want.

packages/communication-react/review/communication-react.api.md
```diff
@@ -1268,6 +1268,9 @@ export interface ErrorBarStrings {
 // @public
 export type ErrorType = keyof ErrorBarStrings;
 
+// @beta
+export type FileDownloadHandler = (userId: string, fileData: FileMetadata) => Promise<string>;
```
prprabhu-ms, line 1272: `URL | FileDownloadErrorMessage`

`interface FileDownloadErrorMessage { reason: string }`.

prprabhu-ms (approved):

PR #1447, opened 2022-02-09
Remove `ChatContext` error recovery that leaked intermittent state

prprabhu-ms:
# What

Delete error recovery in `batch()`

# Why

This error recovery wasn't incorrect for two reasons:

- All current users were internal, where the batch operation contained multiple headless SDK method calls. Rolling back the state in the stateful client would actually create desync between stateful client and underlying service.

# How Tested

# Process & policy checklist


prprabhu-ms:
Taking another approach.

PR #1448, opened 2022-02-09
Dependabot Fixes - static html sample

dmceachernmsft:
# What
Update azure-communicaiton-common and nanoid dependencies.

# Why
Fixes dependabot issues for static html composites

# How Tested

Built all flavors to see if anything failed


samples/StaticHtmlComposites/package.json
```diff
@@ -20,7 +20,9 @@
   "license": "MIT",
   "dependencies": {
     "@azure/communication-react": "1.0.1-beta.2",
-    "@azure/communication-common": "1.0.0",
+    "@azure/communication-common": "1.1.0",
+    "@azure/communication-calling": "1.3.2-beta.1 || 1.3.2",
```
anjulgarg, line 24: Why do we need to add these deps explicitly?

packages/chat-stateful-client/package.json
```diff
@@ -33,8 +33,8 @@
     "@internal/acs-ui-common": "1.0.1-beta.2",
     "events": "~3.3.0",
     "immer": "9.0.6",
-    "nanoid": "~3.1.20",
-    "@azure/communication-common": "1.0.0"
+    "nanoid": "~3.1.31",
+    "@azure/communication-common": "1.1.0"
```
JamesBurnside, line 37: nit: keep these deps alphabetized

samples/tests/package.json
```diff
@@ -20,7 +20,7 @@
   "license": "MIT",
   "dependencies": {
     "@azure/communication-react": "1.0.1-beta.2",
```
JamesBurnside, line 22: if this uses `@azure/communication-react` this should likely also specify the `@azure/communication-calling` and `chat` deps like the static html sample

packages/react-composites/package.json
```diff
@@ -62,7 +62,7 @@
     "events": "~3.3.0",
     "immer": "9.0.6",
     "memoize-one": "~5.2.1",
-    "nanoid": "~3.1.20",
+    "nanoid": "~3.1.31",
```
JamesBurnside, line 65: be sure to mention changes like this in the PR description if they are intended

JamesBurnside: PR specifies just `Update azure-communicaiton-common package`

JamesBurnside (approved):

packages/chat-stateful-client/package.json
dmceachernmsft: kk will go fix that!

samples/StaticHtmlComposites/package.json
dmceachernmsft: common needed a update for some of its child dependencies and calling and chat were moved out of being peer dependencies so they needed to be declared here too.

prprabhu-ms: This is correct. An oversight, all our samples need to declare chat/calling as deps.

prprabhu-ms (approved):

PR #1449, opened 2022-02-09
Log state changes in `ChatContext` when verbose logging is enabled

prprabhu-ms:
# What

Similar to https://github.com/Azure/communication-ui-library/pull/1443 but for chat.

There is still one state change that isn't logged, coming in a follow up PR to keep some logic changes separate.

# Why

Gives us some visibility into internal state updates of `StatefulChatClient`. The logs are a lot more noisy than the calling counterpart because of the state tends to update a lot more frequently (for each message, typing indicator, read receipt etc.)


packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -33,68 +34,66 @@ export class ChatContext {
...
-  public modifyState(modifier: ChatStateModifier): void {
-    this.batch(() => {
-      this.setState(
-        produce(this._state, (draft: ChatClientState) => {
-          modifier(draft);
-        })
-      );
+  public modifyState(modifier: (draft: ChatClientState) => void): void {
+    this._state = produce(this._state, modifier, (patches: Patch[]) => {
+      if (getLogLevel() === 'verbose') {
```
PorterNan, line 62: In my mind, it sounds like verbose < info < warnings

So my understanding here, we wanna still keep this one as verbose level but log the information using logger.info just because of the need of console output, but not to log anything when user set it as info level. Is this a correct understanding?

alkwa-msft (approved):

JamesBurnside (approved):

packages/chat-stateful-client/src/ChatContext.ts
prprabhu-ms: Yes. This log messages are quite verbose, but I can't just `log.verbose` because they don't show up on console by default. :(

By contrast, @azure/communication-calling spews a lot of logs on `log.info`, which I dislike quite a bit.

PR #1450, opened 2022-02-09
`StatefulChatClient`: Only emit `stateChanged` event when there is a change

prprabhu-ms:
* Only emit `stateChanged` event when there is an change.
* This allows the only place that explicitly checked for a change to also use `ChatContext.modifyState`.

# Why

* Enable logging for the last state change location
* Potentially avoid spurious `stateChanged` event.

# How Tested

Smoke tested with chat sample
unittests


packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -50,21 +50,15 @@ export class ChatContext {
...
   public modifyState(modifier: (draft: ChatClientState) => void): void {
+    const priorState = this._state;
     this._state = produce(this._state, modifier, (patches: Patch[]) => {
       if (getLogLevel() === 'verbose') {
         // Log to `info` because AzureLogger.verbose() doesn't show up in console.
         this._logger.info(`State change: ${JSON.stringify(patches)}`);
       }
     });
-    if (!this._batchMode) {
+    if (!this._batchMode && this._state !== priorState) {
```
PorterNan, line 61: Good to see we are adding one more checker to this
Curious about if there is a case that someone calls modifyState but state wouldn't change - Is there an existing case for this? Maybe like we set a string, but it is the same value as the existing one?

PorterNan (approved):

packages/chat-stateful-client/src/ChatContext.ts
anjulgarg, line 61: Does it handle complex deep changes?
My knowledge around javascript object comparison is rusty.
But what happens if a deeply nested attribute has changed in the state. Would this comparison handle that as well?

anjulgarg: Approved. However, I would appreciate a comment on this. :)

alkwa-msft (approved):

packages/chat-stateful-client/src/ChatContext.ts
prprabhu-ms: Nope. This is just a reference check.

The magic part is that the way we modify state is on line 55 via `produce`. *That* ensures that if there any change in the dag, the root node is recreated.

This is why selectors are effecient and React can skip renders when there is no change.

PR #1451, opened 2022-02-10
`StatefulCallClient`: Skip sending `stateChanged` event when there is no real change to state

prprabhu-ms:
# What

Similar to https://github.com/Azure/communication-ui-library/pull/1450 but for calling.

# Why

Bring parity between call and chat; reduces some selectors executions. The overall performance implication (improvement) should be minimal though. I saw only a handful of spurious events in a 5 minute session with several participants (logging was added for debugging only): 
image


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -44,7 +44,6 @@ export class CallContext {
   private _state: CallClientState;
   private _emitter: EventEmitter;
   private _atomicId: number;
-  private _batchMode: boolean;
```
PorterNan, line 47: We will still need batchMode - sometimes we wanna stop stateChanged event for a while in callStatefulClient - especially when we wanna do a bulk update - so needs for 'batch' and state compare are different - I believe we need both

alkwa-msft (approved):

JamesBurnside (approved):

packages/calling-stateful-client/src/StatefulCallClient.test.ts
```diff
@@ -166,7 +165,6 @@ describe('Stateful call client', () => {
         )
       ).toBe(true);
       expect(Object.keys(client.getState().calls[callId]?.remoteParticipantsEnded ?? {}).length).toBe(1);
-      // FIXME: There should be only one event triggered here.
```
prprabhu-ms, line 169: There are two events here because two changes are made: (1) `remoteParticipants` and (2) `remoteParticipantsEnded`.

We could have batched those two changes in a single event, but unlike the other places in this commit, this isn't a bug in how `stateChanged` event is fired.

PR #1452, opened 2022-02-10
Extract various button wrappers to be reused across `CallComposite` and `MeetingComposite`.

prprabhu-ms:
# What

This is the first PR in a series to break apart the `CallControls` component. That component currently shoulders several responsibilities. It's becoming unwieldy to share across composites.

This PR extracts out the logic relevant to creating each of the default buttons.

Breaking `CallControls` apart is necessary before we can introduce the chat button in `MeetingComposite`, and also swap the order of some common buttons.


packages/react-composites/src/composites/CallComposite/styles/Buttons.styles.ts
```diff
@@ -0,0 +1,114 @@
...
+  `
+};
+
+/**
+ * @private
+ */
+export const groupCallLeaveButtonStyle = {
+  root: {
+    border: '0.125rem',
+    borderRadius: 2,
```
anjulgarg, line 27: Let's not change it in this PR. But maybe we can have a follow up cleanup PR for converting hardocded border radius and box shadows to use `theme.effects.roundedCorner` and `theme.effects.elevation`

packages/react-composites/src/composites/CallComposite/types/CallControlOptions.ts
```diff
@@ -0,0 +1,138 @@
...
+ * @public
+ */
+export type CallControlOptions = {
+  /**
+   * Options to change how the call controls are displayed.
+   * `'compact'` display type will decreases the size of buttons and hide the labels.
+   *
+   * @remarks
+   * If the composite `formFactor` is set to `'mobile'`, the control bar will always use compact view.
+   *
```
anjulgarg, line 26: nit:
{@link CallControlDisplayType}

anjulgarg (approved):

PR #1453, opened 2022-02-10
Extract out logic for creating custom buttons

prprabhu-ms:
# What

This is the second PR in a series to simplify `CallControls` to the point where it need not be shared across composites.

This PR extracts out the logic for creating custom buttons. It also simplifies the conditional compilation logic quite a bit by declaring the new types unconditionally, but exporting them from the package conditionally.


packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -132,50 +111,31 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
...
               displayType={options?.displayType}
               increaseFlyoutItemSize={props.increaseFlyoutItemSize}
               styles={props.commonButtonStyles}
             />
           )}
-          {
-            /* @conditional-compile-remove-from(stable): custom button injection */
-            <FilteredCustomButtons customButtonProps={customButtonProps} placement={'afterOptionsButton'} />
-          }
+          {customButtons['afterOptionsButton']}
```
prprabhu-ms, line 122: Not for this PR: This should be named `afterDevicesButton`.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -132,50 +111,31 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
...
-  return (
-    <>
-      {props.customButtonProps
-        .filter((buttonProps) => buttonProps.placement === props.placement)
-        .map((buttonProps, i) => (
-          <ControlBarButton {...buttonProps} key={`${buttonProps.placement}_${i}`} />
-        ))}
-    </>
-  );
+const onFetchCustomButtonPropsTrampoline = (
```
anjulgarg, line 134: `onFetchCustomButtonPropsTrampoline` 
What's the story behind this name?

dmceachernmsft: I mean I like a good trampoline but its also got me wondering

anjulgarg (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prprabhu-ms: Heh, old habit. `trampoline` is a term used to refer to helper functions for FFI bindings.

image

from https://www.gnu.org/software/mit-scheme/documentation/stable/mit-scheme-ffi.pdf

prprabhu-ms: btw, 100% onboard with a more meaningful name for these kinds of functions for conditional compilation.

packages/react-composites/src/composites/CallComposite/types/CallControlOptions.ts
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-/* @conditional-compile-remove-from(stable): custom button injection */
```
PorterNan, line 4: Are we going to say we wanna check this into stable directly?

prprabhu-ms: This is defined unconditionally, but only exported for beta.

We've found that this pattern makes for cleaner code and smaller diff between the build flavors -- define the types unconditionally internally, but only export in the beta flavor.

alkwa-msft (approved):

PR #1454, opened 2022-02-10
Separate out `CallControls` and `MeetingCallControlBar`

prprabhu-ms:
# What

This is last in a sequence of PRs to separate out the arrangement of buttons in `CallComposite` and `MeetingComposite`.

After this PR, the layout for `MeetingComposite` can be changed independently. This will be done in follow up PRs.

# Why

`CallComposite` and `MeetingComposite` lay out their buttons in different orders. This refactoring allows us to share a lot of code while keeping the arrangement separate.


packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -131,6 +86,9 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
   );
 };
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const isEnabled = (option: any): boolean => option !== false;
```
anjulgarg, line 90: option can only be boolean here
any particular reason for using `any`?

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -75,54 +53,31 @@ export const CallControls = (props: CallControlsProps): JSX.Element => {
...
+        <ControlBar layout="horizontal">
           {customButtons['first']}
-          {options?.microphoneButton !== false && (
-            <Microphone
-              displayType={options?.displayType}
-              styles={props.commonButtonStyles}
-              splitButtonsForDeviceSelection={props.splitButtonsForDeviceSelection}
-            />
-          )}
+          {isEnabled(options?.microphoneButton) && <Microphone displayType={options?.displayType} />}
```
dmceachernmsft, line 58: where did all the props go for these buttons?

anjulgarg (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prprabhu-ms: ... magic ... :P

The `<Microphone>` component calls for `usePropsFor(MicrophoneButton)`. This allows us to encapsulate button specific logic in these wrappers, and makes the code flow here linear -- the only thing happening here is putting the buttons in the right sequence.

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
```diff
@@ -193,3 +225,6 @@ const getDesktopEndCallButtonStyles = (theme: ITheme): ControlBarButtonStyles =>
   };
   return concatStyleSets(getDesktopCommonButtonStyles(theme), overrides);
 };
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const isEnabled = (option: any): boolean => option !== false;
```
JamesBurnside, line 230: nit: typescript recommendation is to use `unknown` instead of `any`

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
JamesBurnside: typescript recommends to use `unknown` instead of `any` for these scenarios

packages/react-composites/src/composites/MeetingComposite/MeetingCallControlBar.tsx
alkwa-msft, line 230: similar to Anjul's comment around any. Can we avoid using "any" here as well?

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prprabhu-ms: Oh right `unknown` still allows me to do an eqeqeq check. Done.

PR #1457, opened 2022-02-10
MeetingComposite - Added new constructor that accepts stateful client arguments

edwardlee-msft:
# What
Added new constructor that accepts stateful client arguments in MeetingComposite

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2755205 

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -454,6 +463,44 @@ export const createAzureCommunicationMeetingAdapter = async ({
...
+  callClient,
+  callAgent,
+  meetingLocator,
+  chatClient,
+  threadId
+}: AzureCommunicationMeetingAdapterFromClientArgs): Promise<MeetingAdapter> => {
+  const locator = isTeamsMeetingLinkLocator(meetingLocator) ? meetingLocator : meetingLocator.callLocator;
+  const createCallAdapterPromise = createAzureCommunicationCallAdapterFromClient(callClient, callAgent, locator);
+
+  const chatThreadClient = await chatClient.getChatThreadClient(threadId);
```
PorterNan, line 497: Let's ask contoso app to call this line whenever they want to call this function - so parameters will be 
{
  callClient,
  callAgent,
  meetingLocator,
  chatClient,
  chatThreadClient
}
If anyone wants to call createAzureCommunicationMeetingAdapterFromClient, that means they wanna have full access to chatClient and chatThreadClient - which means they will need to have these 2 instances ready
Otherwise we are creating a new chatThreadClient instance in the adapter which could cause some unexpected behavior

PorterNan (approved):

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -454,6 +463,44 @@ export const createAzureCommunicationMeetingAdapter = async ({
...
+
+/**
+ * Create a {@link MeetingAdapter} using the provided {@link StatefulChatClient} and {@link StatefulCallClient}.
+ *
+ * Useful if you want to keep a reference to {@link StatefulChatClient} and {@link StatefulCallClient}.
+ * Consider using {@link createAzureCommunicationMeetingAdapter} for a simpler API.
+ *
+ * @beta
+ */
+export const createAzureCommunicationMeetingAdapterFromClient = async ({
```
PorterNan, line 487: createAzureCommunicationMeetingAdapterFromClient**s**
cause we are asking for multiple clients in parameters

PorterNan:
This is a nice alignment with call and chat adapter!

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -454,6 +464,42 @@ export const createAzureCommunicationMeetingAdapter = async ({
...
+ */
+export const createAzureCommunicationMeetingAdapterFromClients = async ({
+  callClient,
+  callAgent,
+  meetingLocator,
+  chatClient,
+  chatThreadClient
+}: AzureCommunicationMeetingAdapterFromClientArgs): Promise<MeetingAdapter> => {
+  const locator = isTeamsMeetingLinkLocator(meetingLocator) ? meetingLocator : meetingLocator.callLocator;
+  const createCallAdapterPromise = createAzureCommunicationCallAdapterFromClient(callClient, callAgent, locator);
```
alkwa-msft, line 496: Do we have a preference to have these steps or can we just await when we are creating them and then skip having to do a Promise.all later?

JamesBurnside: The `Promise.all` here allows them to complete asynchronously - so the second adapter doesn't have to wait on the first if the first has any IO work (e.g. network calls) that it takes up time waiting for to complete

packages/react-composites/src/composites/MeetingComposite/adapter/AzureCommunicationMeetingAdapter.ts
```diff
@@ -454,6 +464,42 @@ export const createAzureCommunicationMeetingAdapter = async ({
   return new AzureCommunicationMeetingAdapter(callAdapter, chatAdapter);
 };
 
+/**
+ * Arguments for {@link createAzureCommunicationMeetingAdapterFromClient}
+ *
+ * @beta
+ */
+export type AzureCommunicationMeetingAdapterFromClientArgs = {
+  meetingLocator: CallAndChatLocator | TeamsMeetingLinkLocator;
```
JamesBurnside, line 473: meeting locator isn't quite applicable to this construction as it requires a chatThreadId but we don't actually use the chatThreadId here (the chatThreadClient that is passed in here uses that).
Instead just make this a `callLoctor: CallLocator | TeamsMeetingLinkLocator`

JamesBurnside (approved):

prprabhu-ms (approved):

PR #1462, opened 2022-02-12
Add DrawerMenu component w/ story

JamesBurnside:
# What
Add component that takes in context-menu style props to create a drawer-style menu

## Follow up PRs
* Sub menu support
* A11y work for the whole component (Surface + Menu + sub menu support)
* Animation curves and some more refined styling once sub menus are completed and this is tested inside the composite

# How Tested
Storybook just so far (likely there will be small style fixes once this is added and tested inside the composite)
image


packages/react-components/src/components/Drawer/DrawerMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Props for the {@link _DrawerMenu}
+ *
+ * @internal
+ */
+export interface _DrawerMenuProps {
+  items: _DrawerMenuItemProps[];
+
+  /**
+   * Callback when the drawer's light-dismissal is triggered.
```
prprabhu-ms, line 30: From a quick search, it seems to me that "light dismisal" is a Microsoft UI surfaces specific term.

https://docs.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/flyouts

Should we define / or link to a more "standard" reference to understand what this means / when it is triggered.

packages/react-components/src/components/Drawer/DrawerMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const modifiedFirstItemStyle = useMemo(
+    () =>
+      merge(firstItemStyle ?? {}, {
+        root: { borderTopRightRadius: '0.375rem', borderTopLeftRadius: '0.375rem' }
+      }),
+    [firstItemStyle]
+  );
+
+  return (
+    <_DrawerSurface styles={props.styles?.drawerSurfaceStyles} onLightDismiss={props.onLightDismiss}>
```
prprabhu-ms, line 57: The comment on line 47 says that first item's styles should match the surfaces... but then the radius is hardcoded in line 51.

should you extract the border radius from the drawer surface style instead?

packages/react-components/src/components/Drawer/DrawerMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * @internal
+ */
+export const _DrawerMenu = (props: _DrawerMenuProps): JSX.Element => {
+  const firstItemStyle = props.items[0]?.styles;
+
+  // Ensure the first item has a border radius that matches the DrawerSurface
+  const modifiedFirstItemStyle = useMemo(
+    () =>
+      merge(firstItemStyle ?? {}, {
+        root: { borderTopRightRadius: '0.375rem', borderTopLeftRadius: '0.375rem' }
```
prprabhu-ms, line 51: Use `theme`

packages/react-components/src/components/Drawer/DrawerMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+      }),
+    [firstItemStyle]
+  );
+
+  return (
+    <_DrawerSurface styles={props.styles?.drawerSurfaceStyles} onLightDismiss={props.onLightDismiss}>
+      <Stack styles={props.styles}>
+        {props.items.slice(0, 1).map((item) => (
+          // Design TODO: first item needs border radius applied
```
prprabhu-ms, line 60: Why TODO. You already do this above, right?

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { BaseCustomStyles } from '../../types';
+import { submitWithKeyboard } from '../utils/keyboardNavigation';
+
+/**
+ * Props for the DrawerMenuItem
+ *
+ * @internal
+ */
+export interface _DrawerMenuItemProps {
+  onItemClick?: () => void;
```
prprabhu-ms, line 16: Should this callback be provided the `key` so that clients can use a single callback for all items instead of having to create individual closures.

They can still always ignore the `key` in the callback if they wish.

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Maps the individual item in menuProps.items passed in the {@link DrawerMenu} into a UI component.
+ *
+ * @private
+ */
+export const DrawerMenuItem = (props: _DrawerMenuItemProps): JSX.Element => {
+  const palette = useTheme().palette;
+  const rootStyles = mergeStyles(drawerMenuItemRootStyles(palette.neutralLight), props.styles?.root);
+  const onClick = (): void => props.onItemClick && props.onItemClick();
```
prprabhu-ms, line 34: nit: I'd inline this.

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </Stack>
+  );
+};
+
+const menuItemChildrenGap = { childrenGap: '0.5rem' };
+
+const drawerMenuItemRootStyles = (hoverBackground: string): IStyle => ({
+  height: '3rem',
+  lineHeight: '3rem',
+  fontSize: '1rem',
```
prprabhu-ms, line 70: use `theme` rather than hardcoding font size.

packages/storybook/.storybook/preview.tsx
```diff
@@ -16,10 +17,17 @@ import { THEMES } from '../stories/themes';
 import { LOCALES } from '../stories/locales'
 import { TOC } from './TOC';
 
+// icons used in stories that are not part of the default component icons
+const storyIcons = {
```
prprabhu-ms, line 21: <3

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+      onKeyPress={onKeyPress}
+      onClick={onClick}
+      tokens={menuItemChildrenGap}
+    >
+      {props.iconProps && (
+        <Stack.Item>
+          <Icon {...props.iconProps} />
+        </Stack.Item>
+      )}
+      <Stack.Item grow>{props.text}</Stack.Item>
```
prprabhu-ms, line 52: Can this be a `<Text>` instead?

packages/react-components/src/components/Drawer/DrawerMenu.tsx
JamesBurnside: MUI just has onClose and doesn't refernece the clickableOutside nature: https://mui.com/api/drawer/ (but does have a "click-away listener https://mui.com/components/click-away-listener/)

packages/react-components/src/components/Drawer/DrawerMenu.tsx
JamesBurnside: Yeah, that was me being lazy, I will update that

packages/react-components/src/components/Drawer/DrawerMenu.tsx
JamesBurnside: Yah left in from lower stack PRs, will remove

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
JamesBurnside: Ah I wondered why fluent's contextual menu props had the itemKey in the callback (I was confused as you are passing in item key as a prop and onclick, it should be redundant to have itemKey returned in the callback) -- but what you say makes sense! I will add it in

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
JamesBurnside: I updated to be used by the `onKeyPress` (in higher PR) so will leave the onClick here

packages/react-components/src/components/Drawer/DrawerMenu.tsx
JamesBurnside: Updated to use theme so left comment here that they should match

packages/react-components/src/components/Drawer/DrawerMenu.tsx
JamesBurnside: Updated to use theme

packages/react-components/src/components/Drawer/DrawerMenu.tsx
JamesBurnside: Removed

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
JamesBurnside: Added `ev` and `itemKey` as optional callback args to closely match IContextMenuProps (they take the whole item as the callback instead of key but I don't immediately see how that is done)

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
JamesBurnside: Updated in this PR to highlight, moved rootStyles inline

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
JamesBurnside: Updated

packages/react-components/src/components/Drawer/DrawerMenu.tsx
JamesBurnside: Updated to use useTheme so left in comment

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
JamesBurnside: Oh nice, updated

prprabhu-ms (approved):

packages/react-components/src/components/Drawer/DrawerMenu.tsx
JamesBurnside: As this is internal only I will leave as lightDismiss as this is inline with fluent's nomenclature. We can revisit the API naming if we decide to make this a publicly available component

mgamis-msft (approved):

PR #1464, opened 2022-02-14
Add github action to update API files

JamesBurnside:
# What
Quickly add a github action to update API files.
**note:** I am explicitly _not_ adding this as a PR label as API changes should be very intentional and we do not want accidental API changes included in PRs

# Why
I keep forgetting to update stable api files and the process is quite long:
`hop onto branch -> rush switch-flavor:stable -> rush build -t @azure/communication-react`
^ and that invalidates the build cache meaning the next build on my working branch has to rebuild _all_ projects again.

# How Tested
Mosty copy-pasta from our snapshot-update GH action. Will test once action is pushed and can be tested via workflow_dispatch in GitHub UI.


.github/workflows/update-api-files.yml
```diff
@@ -0,0 +1,78 @@
+name: Update API files
+
+on:
+  workflow_dispatch:
```
PorterNan, line 4: This is a nice one!
How do people trigger this when they want to generate new api.md? I guess they will need to run this manually and choose the right branch for themselves?

JamesBurnside: Correct, they have to run manually through the github actions UI and select their branch all manually. This is mostly just so I can make use of it haha, I still think all our documentation should say to simply build locally to build the API files -- but if we want to make PRs easier perhaps we should add a PR label

prprabhu-ms: I'd say add a PR label. Kicking via actions tab is not all that different from adding a PR label. But latter is more discoverable.

We'd still review the API changes in the same way.

prprabhu-ms (approved):

.github/workflows/update-api-files.yml
PorterNan: Agree with @prprabhu-ms, that would be much faster for contributor to run the pipeline, but we need another label for this like 'api-change'

PorterNan (approved):

.github/workflows/update-api-files.yml
JamesBurnside: I will complete this PR and we can make that a seperate PR (as we would also want things like a comment when the build fails due to api changes to highlight the label is available)

PR #1465, opened 2022-02-14
Foreign keyboard fix

### LINKED ISSUE #1425: Enter key behavior in chat composites
When I type a chat message, I press the Enter key to confirm the conversion in Japanese input, but if I do that in Chrome and Safari on the Mac, the message is sent.
In Japanese input, I have to press Enter several times to confirm the conversion, but I'm hoping that the Enter key after confirmation will send the message.
In Windows, it works as expected.

dmceachernmsft:
# What
adds isComposing check to input box

# Why
Allows for foreign keyboard inputs to compose multiple words on mac

Fixes #1425
# How Tested

Ran locally to ensure premature messages aren't sent.


packages/react-components/review/react-components.api.md
```diff
@@ -4,8 +4,6 @@
 
 ```ts
 
-/// <reference types="react" />
```
alkwa-msft, line 7: how come we are seeing these removals in this PR as well?

@PorterNan  any ideas here?

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -95,6 +95,9 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
 
   const onTexFieldKeyDown = useCallback(
     (ev: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+      if ((ev.nativeEvent as KeyboardEvent).isComposing) {
```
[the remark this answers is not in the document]
dmceachernmsft: im working on a little doc about this we can share, might be something nice for them to incorporate in their text boxes to ensure consistent behavior on all platforms.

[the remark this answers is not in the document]
dmceachernmsft: but testing with Edward this does seem to solve the problem.

JamesBurnside, line 98: Avoid `as` as much as possible as then typescript can't typeguard against misuse.
Looks like the `as` shouldn't be needed?:
image

JamesBurnside (approved):
Remove the changes to the api.md files and use of `as` -- otherwise this looks great! I had no idea this was a thing!

JamesBurnside:
Github tip, updated your PR description to say `Fixes <issues number>`, fixes and fix are special keywords that then track this PR as the fix to this issue and auto closes the issue when PR completes:
image

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -92,6 +92,9 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
 
   const onTexFieldKeyDown = useCallback(
     (ev: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+      if (ev.nativeEvent.isComposing) {
```
prprabhu-ms, line 95: Please add a comment in the code here explaining why this is needed. It's very easy to forget the reason for such checks.

prprabhu-ms (approved):

PR #1469, opened 2022-02-15
Drawer Component sub menu support

JamesBurnside:
# What
Add submenu support to drawer menu component

# How Tested

https://user-images.githubusercontent.com/2684369/154122304-3b733f91-212c-4d9e-8dc4-3c7aeed8e251.mp4


prprabhu-ms:
Quick question on approach: The alternative would be to create a new component for the submenu (instead of updating menuItems in existing surface).

Benefits of the alternative approach:
* Clearer state keeping than current impl
* Better interaction with DOM updates -- e.g. if/when we add animations, we wouldn't want menu items to disappear / appear in a way where the previous items are mixed in with the new ones.

prprabhu-ms (approved):
Actually, this is blocking other work for this sprint.

I'd prefer to land this PR, and then iterate on a different / better approach for submenus if we want.

alkwa-msft (approved):

PR #1470, opened 2022-02-15
Add adjustments in release action for stable

PorterNan:
# What

Add adjustments in release action for stable
- If there is input param, flavor will be chosen by parameter (beta/stable)
- If that is left empty, then determine the flavor by branch name
# Why

# How Tested

# Process & policy checklist


prprabhu-ms (approved):

.github/workflows/npm-release-publish.yml
```diff
@@ -32,6 +35,12 @@ jobs:
       - name: Install dependencies
         run: rush install
 
+      # Switch flavor to stable when it is release branch
+      - name: Switch flavor for stable build
+        # If no param for release flavor, get it from branch name
+        if: ${{ !github.event.inputs.branch && contains(github.ref, 'beta') || github.event.inputs.branch === 'stable' }}
```
JamesBurnside, line 41: should this be `github.event.inputs.release-flavor` instead of `github.event.inputs.branch`?

JamesBurnside: and should it be `contains(github.ref, 'stable')` instead of beta?

PorterNan: Addressed the first comment - for second one, we don't have str like 'stable' in stable branch, besides, it is also recommended to use the parameter instead of leaving it empty, so this is more like a fallback

JamesBurnside: Does 
image
not need updated to release-flavor also?

JamesBurnside: Does this say, if github.ref contains beta then do a rush-switch:stable? That seems odd

JamesBurnside: Should it not be `!contains(github.ref, 'beta')`?

PorterNan: Ohhh! That is correct, it is so sad that I can't test this myself easily, thanks for catching this

JamesBurnside:
I'd also love to see a github action run that tests this (comment out the publishing part of the task but let it upload to the azure blob store and check the blob store contents are correct). Github action syntax is not friendly or very readable and we don't want this failing when we finally do publish
(or even some action runs that are cancelled after the rush-switch is executed)

JamesBurnside (approved):

PorterNan:
> I'd also love to see a github action run that tests this (comment out the publishing part of the task but let it upload to the azure blob store and check the blob store contents are correct). Github action syntax is not friendly or very readable and we don't want this failing when we finally do publish (or even some action runs that are cancelled after the rush-switch is executed)


Thanks, James, for coming up with a smart way to test, now all cases were tested:

Without parameter, without beta in the name: https://github.com/Azure/communication-ui-library/actions/runs/1866916009[](https://github.com/Azure/communication-ui-library/actions/runs/1866916009)

Without parameter, beta in the name: https://github.com/Azure/communication-ui-library/actions/runs/1866928360[](https://github.com/Azure/communication-ui-library/actions/runs/1866928360)

Stable parameter, but beta in the name(will choose stable): https://github.com/Azure/communication-ui-library/actions/runs/1866936808[](https://github.com/Azure/communication-ui-library/actions/runs/1866936808)

beta parameter, but no beta in the name(choose beta): https://github.com/Azure/communication-ui-library/actions/runs/1866945121

PR #1475, opened 2022-02-15
Updating changelog for 1.1.0

anjulgarg:
# What

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/communication-react/CHANGELOG.md
+++ b/packages/communication-react/CHANGELOG.md
@@ -10,7 +10,7 @@ Mon, 14 Feb 2022 23:50:05 GMT
 
 This minor release contains general bug fixes and performance improvements.
 
-## Changes since 1.0.1-beta.2
+## Changes since 1.0.0
 [Compare changes](https://github.com/Azure/communication-ui-library/compare/1.0.0...1.1.0)
 
 
--- a/packages/communication-react/CHANGELOG.md
+++ b/packages/communication-react/CHANGELOG.md
@@ -4,80 +4,23 @@ This log was last generated on Mon, 14 Feb 2022 23:50:05 GMT and should not be m
 
 <!-- Start content -->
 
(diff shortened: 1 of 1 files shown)


packages/communication-react/CHANGELOG.md
```diff
@@ -4,80 +4,23 @@ This log was last generated on Mon, 14 Feb 2022 23:50:05 GMT and should not be m
...
+## Changes since 1.0.1-beta.2
+[Compare changes](https://github.com/Azure/communication-ui-library/compare/1.0.0...1.1.0)
 
-### Patches
 
-- `@internal/chat-component-bindings`
-  - Upgrade @azure/communication-signaling to 1.0.0.beta.12 ([PR #1352](https://github.com/azure/communication-ui-library/pull/1352) by anjulgarg@live.com)
-- `@internal/chat-stateful-client`
+### Changes
   - Upgrade @azure/communication-signaling to 1.0.0.beta.12 ([PR #1352](https://github.com/azure/communication-ui-library/pull/1352) by anjulgarg@live.com)
```
JamesBurnside, line 18: Is it OK to have beta package in our stable releases?

anjulgarg: Good question.
Ideally not. But we have been using the beta packages for this dep even in our past stable releases unfortunately.

alkwa-msft (approved):

packages/communication-react/CHANGELOG.md
prprabhu-ms: There is a bug tracking this. They don't have a stable package *at all*: https://www.npmjs.com/package/@azure/communication-signaling

And I think there are some types we can't get from @azure/communication-chat instead of this one.

prprabhu-ms (approved):

PR #1476, opened 2022-02-15
Revert "Only compile files following trace of entry.. (#1390)

prprabhu-ms:
This reverts commit 8061151ddd7b6b13e387b8c591d9fc11cc2c43fe.

# Why

Reduce the diff between `beta` and `stable` flavor builds. The larger this diff, the higher the chances that build failures will leak in when we stabilize a feature (i.e., remove related conditional compilation directives).

A better approach for beta-only features is to define all internal types unconditionally and only restrict the exported API.

### DIFF
--- a/change/@internal-acs-ui-common-1115c7f5-3291-4ff2-adf0-83473af56a32.json
+++ b/change/@internal-acs-ui-common-1115c7f5-3291-4ff2-adf0-83473af56a32.json
@@ -1,7 +1,0 @@
-{
-  "type": "none",
-  "comment": "Follow the import trace from every index.ts and not to compile non-exported files",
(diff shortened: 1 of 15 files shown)


anjulgarg (approved):

PorterNan:
Wondering if we get any build error after reverting this? Would like to help if we see a lot

prprabhu-ms:
> Wondering if we get any build error after reverting this? Would like to help if we see a lot

Thanks for the offer. I'm planning to really work on fixing all those errors and merge this next week.

prprabhu-ms:
Subsumed by https://github.com/Azure/communication-ui-library/pull/1551

PR #1477, opened 2022-02-15
Samples: Try harder to detect mobile sessions

prprabhu-ms:
# Why

I noticed that the control bar didn't change the displayType consistently when switching the form factor via devTools.

This is fixed if we check the userAgent string asynchronously.

# How Tested

Played around with samples && dev tools.


samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -37,7 +37,12 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
 
   // Whenever the sample is changed from desktop -> mobile using the emulator, make sure we update the formFactor.
   useEffect(() => {
-    const updateIsMobile = (): void => setIsMobileSession(detectMobileSession());
+    const updateIsMobile = (): void => {
+      // The userAgent string is sometimes not updated synchronously when the `resize` event fires.
+      setImmediate(() => {
```
JamesBurnside, line 42: Was this being tested on IE? :D
https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate
image

can do `setTimeout(() => {}, 0)` to get the same effect

alkwa-msft: I think it is ok to not support IE. The underlying calling sdk as an example does not.

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -37,7 +37,12 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
 
   // Whenever the sample is changed from desktop -> mobile using the emulator, make sure we update the formFactor.
   useEffect(() => {
-    const updateIsMobile = (): void => setIsMobileSession(detectMobileSession());
+    const updateIsMobile = (): void => {
```
alkwa-msft, line 40: I am curious why we want to call this in useEffect. useEffect is called after things are rendered.

Why cant we just use mobileDetect once at the very beginning and pass that prop through the components?

samples/Calling/src/app/views/CallScreen.tsx
JamesBurnside: no no, this is _only_ supported on IE (and edge to some extent)

samples/Calling/src/app/views/CallScreen.tsx
anjulgarg: So, this useEffect will be called only once as soon as this composite loads since it has an empty dependency array.

anjulgarg (approved):

samples/Calling/src/app/views/CallScreen.tsx
alkwa-msft: I guess my question is why do we even need to call useEffect to run mobileDetect

samples/Calling/src/app/views/CallScreen.tsx
alkwa-msft: ah that makes more sense .. yes I am also confused why we need to do this.

prprabhu-ms: Oooh.... I looked at that page. And also interpreted it as not supported on IE :D

samples/Calling/src/app/views/CallScreen.tsx
prprabhu-ms: The `useEffect` is needed to register window events, but only once.

JamesBurnside (approved):

PR #1479, opened 2022-02-15
Add best practice for adding a new conditional type and using it internally

prprabhu-ms:
# Why

This pattern has come up a couple times in the last sprint.

# How Tested
Build beta and stable flavor build, verify that preprocessed files match expectations.


packages/acs-ui-common/src/conditional-compilation-sample/index.tsx
```diff
@@ -204,49 +205,47 @@ export const myExtensibleSelector: MyExtensibleSelector = dummyCreateSelector(
...
+  impl.goodOldTestedFunctionality();
+  if (shouldIncludeUnstableFeature()) {
+    const arg: ThisIsUnstableType = 'getThisFromUser';
+    impl.newMethodForUnstableFunctionality(arg);
+  }
+}
+
+function shouldIncludeUnstableFeature(): boolean {
+  /* @conditional-compile-remove-from(stable) */
+  return true;
```
PorterNan, line 249: Can we double check if ts compiler or linter will throw error for this double return setences?

prprabhu-ms: I checked that neither does. CI is also green (runs both build and lint).

JamesBurnside (approved):

packages/acs-ui-common/src/conditional-compilation-sample/index.tsx
```diff
@@ -204,49 +205,47 @@ export const myExtensibleSelector: MyExtensibleSelector = dummyCreateSelector(
...
+ * - Call the new @beta class method from @public API, but only in beta flavored build.
+ *
+ * Following the general guidelines stated above, we achieve this as follows:
+ * - Unconditionally add the new type
+ *   - Conditionally export the new type in the public API.
+ * - Conditionally add the new class method that uses the conditionally exposed type.
+ * - In the @public method, extract a simple helper function that detects whether the @beta method should be called.
+ *   The implementation of this function uses conditional compilation.
+ *   - Call the method if needed. Because the method and type are defined unconditionally, no conditional
+ *     compliation is needed here.
```
edwardlee-msft, line 221: correct compliation to compilation.

prprabhu-ms: Ouch, auto-submit was enabled.

I can't spell to safe my life :/

PR #1480, opened 2022-02-16
remove lock files added by mistake

dmceachernmsft:
# What
Remove lock files that were added by accident

# Why
removes lock files that might confuse build pipline dependency

# How Tested

built locally.


prprabhu-ms (approved):

alkwa-msft:
In general if we do make any dependency changes we should be commiting the lock files so its building in the same way that we are building locally. I am sure we are doing the right thing there tho.

alkwa-msft (approved):

PR #1482, opened 2022-02-16
Added read by information in chat bubble contextual menu

carocao-msft:
# What
Added read by information in chat bubble contextual menu

# Why
show how many people have read the message when participant number more than 2

# How Tested
1. when less than equal to 2 participants, contextual menu remain unchanged
Screen Shot 2022-02-16 at 12 11 24 PM

2. when more than 2 participants, open contextual menu, will see read by x of y
3. when read by 0 of y, button is disabled 

Screen Shot 2022-02-16 at 12 12 45 PM

4. when read by more than 0 of y, button is enabled 
Screen Shot 2022-02-16 at 12 12 58 PM
5. have not implemented the submenu yet so I didn't add the chevron icon in 

Style: 
https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?node-id=4063%3A26046


packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -19,6 +26,8 @@ export interface ChatMessageActionFlyoutProps {
   onEditClick?: () => void;
   onRemoveClick?: () => void;
   onDismiss: () => void;
+  messageThreadReadCount?: number;
```
prprabhu-ms, line 29: I find this prop name a little misleading. It's how many participants have read the message right? It isn't related to the "thread" per say?

How about `messageReadByCount` ?

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -32,34 +41,93 @@ export interface ChatMessageActionFlyoutProps {
...
-          iconName: 'MessageRemove',
-          styles: menuIconStyleSet
+  const theme = useTheme();
+  const menuItems = useMemo((): IContextualMenuItem[] => {
+    // only show read by x of x if more than 3 participants in total including myself
+    if (
+      props.participantCountNotIncludingSelf &&
+      props.messageThreadReadCount !== undefined &&
+      props.participantCountNotIncludingSelf >= 2 &&
+      props.strings.messageReadCount
```
prprabhu-ms, line 51: We assume that strings always exist. So checking for `props.strings.messageReadCount` isn't necessary / looks off.

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -32,34 +41,93 @@ export interface ChatMessageActionFlyoutProps {
...
+    ) {
+      return [
+        {
+          key: 'Edit',
```
prprabhu-ms, line 55: Don't duplicate the menu items for "edit", and "remove" in the two branches. Instead, first create the array with those two, then add in the read count related items if the conditions are met.

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -104,6 +114,8 @@ export const ChatMessageComponentAsMessageBubble = (props: ChatMessageComponentA
           onEditClick={onEditClick}
           onRemoveClick={onRemoveClick}
           strings={strings}
+          messageThreadReadCount={message.readNumber ? message.readNumber : 0}
+          participantCountNotIncludingSelf={participantCountNotIncludingSelf ? participantCountNotIncludingSelf : 0}
```
prprabhu-ms, line 118: Why default to 0 here? Since `participantCountNotIncludingSelf` is a prop being forwarded, it would be better to default higher up where this is first defined, or lower in the component that uses it.

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -869,14 +871,16 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
           <ChatMessageComponent
             {...messageProps}
             message={messageProps.message}
+            // -1 because participant count does not include myself
```
prprabhu-ms, line 874: I'd drop this comment. The prop is named very well, so it's obvious why you need to subtract 1.

prprabhu-ms (approved):

prprabhu-ms (commented):
Actually, would be nice to see a few screenshots in the PR description.

Are there e2e tests already covering this?

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
carocao-msft: I'm checking for props.strings.messageReadCount because it is an optional param

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
carocao-msft: participantCountNotIncludingSelf is an optional prop so I need to give it a default value here, what's a better way for giving default values to optional prop?

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
```diff
@@ -22,6 +22,7 @@ type ChatMessageComponentProps = {
    * Setting to false will mean they are on a new line inside the editable chat message.
    */
   inlineAcceptRejectEditButtons: boolean;
+  participantCountNotIncludingSelf?: number;
```
alkwa-msft, line 25: I dont think this is better but what about

"numOtherParticipants"

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
alkwa-msft: can we not say its optional and self the value to 0 if we dont pass in anything? it would clean up this area of the code.
https://www.javatpoint.com/typescript-function-parameter#:~:text=TypeScript%20Function%20Parameter%201%20Optional%20Parameter.%20JavaScript%20allows,zero%20or%20more%20values%20to%20a%20function.%20
image

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
carocao-msft: updated :)

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
carocao-msft: updated :)

carocao-msft:
> Actually, would be nice to see a few screenshots in the PR description.
> 
> Are there e2e tests already covering this?

Updated PR to include screenshots

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
prprabhu-ms: Ah, new strings that we add need to be optional... because of backwards compatibility.

@mgamis-msft can correct me here. Don't we fallback to our own en-us strings for anything that Contoso doesn't provide? So there is _no way_ for Contoso to initialize the composite in a way where strings are missing?

If this is the case we should:
* Make sure all strings are *required*.
* Only require `Partial<LocalType>` in our constructors and provider exported in the API.

For this PR, @carocao-msft could you add a TODO once Miguel resolves this thread with the right way ahead. You are right that you don't have any option given the types at present.

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
prprabhu-ms: `remoteParticipantsCount` would be consistent with naming elsewhere in our library (and headless SDKs)

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -104,6 +114,8 @@ export const ChatMessageComponentAsMessageBubble = (props: ChatMessageComponentA
           onEditClick={onEditClick}
           onRemoveClick={onRemoveClick}
           strings={strings}
+          messageReadByCount={message.readNumber ? message.readNumber : 0}
```
JamesBurnside, line 117: nit, can use nullish coalescing operator here:
```suggestion
          messageReadByCount={message.readNumber ?? 0}
```

JamesBurnside (approved):

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
PorterNan: Aha, I like the idea of 'remote' which means not the local user

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
carocao-msft: updated

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
carocao-msft: Changed name to remoteParticipantsCount

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
carocao-msft: Updated !

prprabhu-ms (approved):

PR #1484, opened 2022-02-16
Fix chat and people button icons

prprabhu-ms:
# What

Use filled icon for both open/close pane when no label is shown.
Use regular icon for both open/close pane when label is shown.

# Why
Match figma design.

# How Tested

Meeting sample.

image
image
image
image


packages/react-composites/src/composites/CallWithChatComposite/ChatButtonWithUnreadMessagesBadge.tsx
```diff
@@ -27,20 +27,25 @@ export interface ChatButtonWithUnreadMessagesBadgeProps extends ControlBarButton
 const validNewChatMessage = (message): boolean =>
   !!message.senderDisplayName && (message.type === 'text' || message.type === 'html');
 
+const filledIcon = <Chat20Filled key={'chatIconKey'} primaryFill="currentColor" />;
```
dmceachernmsft, line 30: why was this logic moved out of the chat button?

JamesBurnside (approved):

packages/react-composites/src/composites/CallWithChatComposite/ChatButtonWithUnreadMessagesBadge.tsx
prprabhu-ms: It is needed here because the icon is chosen in two places -- once in `ChatButton` and once in the icon used in the `Stack` in `notificationOnIcon`.

Moved it here to deduplicate the logic about which icon to use.

dmceachernmsft: ahh okay makes sense!

dmceachernmsft (approved):

PR #1486, opened 2022-02-16
MobilePane for Chat and People panes in CallWithChat composite.

mgamis-msft:
# What

ChatAndPeoplePane added to CallWithChat composite.

Note:
I decided to add both Chat and People to this separate pane. I hope that is not too big for 1 PR. I was thinking about how to make the separate pane with 1 tab extensible to additional tabs but could not think of how to do it. So I decided to have Chat as a second tab. Let me know if there is a better component design I could do here.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2590657
https://skype.visualstudio.com/SPOOL/_workitems/edit/2757257

# How Tested

Locally through CallWithChat composite.
https://user-images.githubusercontent.com/79475487/154236764-e20b88e4-2b28-42ce-aecd-6ae7ba07a49b.mp4

# Process & policy checklist


packages/react-composites/src/composites/CallWithChatComposite/ChatAndPeoplePane.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const theme = useTheme();
+  const strings = useCallWithChatCompositeStrings();
+  const chatAndPeoplePaneButtonStylesThemed = concatStyleSets(chatAndPeoplePaneButtonStyles, {
+    rootChecked: {
+      borderBottom: `0.125rem solid ${theme.palette.themePrimary}`
+    }
+  });
+
+  return (
+    <Stack verticalFill grow className={activeTab ? chatAndPeoplePaneStyle : chatAndPeoplePaneHiddenStyle}>
```
alkwa-msft, line 53: For these different components could we make them different files? Could be stylistic but these files are getting pretty big and we are just changing what we are passing in?

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -141,18 +147,65 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
...
+      disableButtonsForLobbyPage={isInLobbyOrConnecting}
+      callControls={props.callControls}
+    />
+  );
+
+  /* @conditional-compile-remove-from(stable) */
+  if (isMobile) {
+    return (
+      <Stack className={allHeightAndWidthStyle}>
+        {callAdapter && chatProps.adapter && hasJoinedCall && (
```
alkwa-msft, line 178: what happens if hasJoinedCall is false? It seems to be required for all of the components we are going to return. If we just need it in all cases then perhaps  

If(!hasJoinedCall) then just return?

packages/react-composites/src/composites/CallWithChatComposite/ChatAndPeoplePane.tsx
```diff
@@ -0,0 +1,140 @@
...
+  chatAndPeoplePaneContentHiddenStyle,
+  chatAndPeoplePaneContentStyle,
+  chatAndPeoplePaneControlBarStyle,
+  chatAndPeoplePaneHiddenStyle,
+  chatAndPeoplePaneStyle
+} from './styles/ChatAndPeoplePaneStyles';
+import { useCallWithChatCompositeStrings } from './hooks/useCallWithChatCompositeStrings';
+import { removeParticipantFromCallWithChat } from './SidePane';
+
+export const ChatAndPeoplePane = (props: {
```
prprabhu-ms, line 25: I'm not a big fan of this combined component because it isn't cohesive.

My ideal design would be: A base `MobileEmbeddedPane` component that renders the back button, the chat/people switching buttons, and puts its children in a stack underneath.

Then, the existing `EmbeddedChatPane` and `EmbeddedPeoplePane` use the `MobileEmbeddedPane` if `mobileView` is set. That way `CallWithChatComposite` doesn't have the bit `mobileView` switch, and the chat vs people logic can stay separated as it is right now for desktop.

It does mean that both `EmbeddedChatPane` and `EmbeddedPeoplePane` will have to take an extra callback for switch to the other pane, which they don't need for desktop. But that's just one prop that's used only on mobile.

prprabhu-ms (changes requested):

packages/react-composites/src/composites/CallWithChatComposite/ChatAndPeoplePane.tsx
prprabhu-ms: # This is the big one.

Didn't look into the implementation here in detail. Let's resolve this first.

change/@internal-react-composites-4add5659-a0cc-45ad-9afb-18819cc01431.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
anjulgarg, line 2: Should be `prerelease`
`patch` and `minor` should only be used for stable changes.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -160,19 +162,22 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
   const isInLobbyOrConnecting = currentPage === 'lobby';
   const hasJoinedCall = !!(currentPage && hasJoinedCallFn(currentPage, currentCallState ?? 'None'));
   const showControlBar = isInLobbyOrConnecting || hasJoinedCall;
+  const showMobilePane = isMobile && (showChat || showPeople);
```
prprabhu-ms, line 165: `showMobilePane` is always used negated in the conditionals. Should we rename this `showCallComposite` instead?

prprabhu-ms: And then add a comment here: 

    // When `mobileView` is set, the chat and people pane take up the entire viewport

packages/react-composites/src/composites/CallWithChatComposite/MobilePane.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const MobilePane = (props: {
+  onClose: () => void;
+  onChatButtonClicked: () => void;
+  onPeopleButtonClicked: () => void;
+  children: React.ReactNode;
+  hidden: boolean;
+  dataUiId: string;
+  activeTab: MobilePaneTab;
+}): JSX.Element => {
+  // We hide the mobile pane instead of not rendering the entire pane to persist certain elements
```
prprabhu-ms, line 29: Great comment. Comments like this make our future self soooo grateful.

packages/react-composites/src/composites/CallWithChatComposite/MobilePane.tsx
```diff
@@ -0,0 +1,75 @@
...
+  // between renders. An example of this is composing a chat message - a chat message that has been
+  // typed but not sent should not be lost if the side panel is closed and then reopened.
+  const mobilePaneStyles = props.hidden ? hiddenMobilePaneStyle : mobilePaneStyle;
+  const theme = useTheme();
+  const chatAndPeoplePaneButtonStylesThemed = concatStyleSets(mobilePaneButtonStyles, {
```
prprabhu-ms, line 34: nit: Shorter name is equally good here: `buttonStyles`

prprabhu-ms: Also, since this is a `styles` object directly passed down (not via `mergeStyles`), consider `useMemo` to avoid spuriously rerendring children.

packages/react-composites/src/composites/CallWithChatComposite/styles/MobilePane.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  rootCheckedHovered: { background: 'none' }
+};
+
+export const mobilePaneButtonStyles: IButtonStyles = {
+  root: {
+    border: 'none',
+    borderBottom: '0.125rem solid transparent',
+    minWidth: '8rem',
+    maxWidth: '8rem',
+    borderRadius: 'none',
```
prprabhu-ms, line 24: Why do you have to explicitly set the `borderRadius`? This should be the default.

packages/react-composites/src/composites/CallWithChatComposite/styles/MobilePane.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  root: { border: 'none', minWidth: '2.5rem', height: '100%', background: 'none', padding: '0 1rem 0 0.5rem' },
+  rootChecked: { background: 'none' },
+  rootCheckedHovered: { background: 'none' }
+};
+
+export const mobilePaneButtonStyles: IButtonStyles = {
+  root: {
+    border: 'none',
+    borderBottom: '0.125rem solid transparent',
+    minWidth: '8rem',
```
prprabhu-ms, line 22: Isn't setting `minWidth` and `maxWidth` the same as setting `width`?

packages/react-composites/src/composites/CallWithChatComposite/styles/MobilePane.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+    borderRadius: 'none',
+    height: '100%',
+    background: 'none',
+    padding: '0'
+  },
+  rootChecked: { background: 'none' },
+  rootCheckedHovered: { background: 'none' },
+  flexContainer: { flexFlow: 'column', display: 'contents' },
+  label: {
+    fontSize: '1rem',
```
prprabhu-ms, line 33: Use `theme` to set font size.

packages/react-composites/src/composites/CallWithChatComposite/styles/MobilePane.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+    fontWeight: 100,
```
prprabhu-ms, line 34: Use theme for font if at all possible (not sure if it has `fontWeight` in some form)

packages/react-composites/src/composites/CallWithChatComposite/MobilePane.tsx
```diff
@@ -0,0 +1,75 @@
...
+  });
+  const strings = useCallWithChatCompositeStrings();
+
+  return (
+    <Stack verticalFill grow styles={mobilePaneStyles} data-ui-id={props.dataUiId}>
+      <Stack horizontal grow styles={mobilePaneControlBarStyle}>
+        <DefaultButton
+          onClick={props.onClose}
+          styles={mobilePaneBackButtonStyles}
+          onRenderIcon={() => <FontIcon iconName="ChevronLeft" />}
```
prprabhu-ms, line 47: Just use `Icon`. I think `FontIcon` is deprecated.

https://github.com/microsoft/fluentui/wiki/Using-icons

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
```diff
@@ -84,6 +85,9 @@ export const EmbeddedPeoplePane = (props: {
   callAdapter: CallAdapter;
   chatAdapter: ChatAdapter;
   onFetchAvatarPersonaData?: AvatarPersonaDataCallback;
+  onChatButtonClick: () => void;
```
prprabhu-ms, line 88: *Not for this PR*. We should really split off `EmbeddedPeoplePane` and `EmbeddedChatPane` to their own files.

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
```diff
@@ -84,6 +85,9 @@ export const EmbeddedPeoplePane = (props: {
...
+  onPeopleButtonClick: () => void;
+  isMobile?: boolean;
```
prprabhu-ms, line 90: This is called `mobileView` elsewhere. Can we continue to call it that for consistency?

prprabhu-ms (approved):

packages/react-composites/src/composites/CallWithChatComposite/ChatAndPeoplePane.tsx
prprabhu-ms: Thanks for iterating on the code organization. Like the new approach a lot.

packages/react-composites/src/composites/CallWithChatComposite/MobilePane.tsx
JamesBurnside: +1!

packages/react-composites/src/composites/CallWithChatComposite/styles/MobilePane.styles.ts
mgamis-msft: It's not needed. Removed. Thanks.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -181,6 +186,15 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
             fluentTheme={fluentTheme}
             onClose={closePane}
             onFetchAvatarPersonaData={props.onFetchAvatarPersonaData}
+            onChatButtonClick={() => {
```
JamesBurnside, line 189: nit: these need memoized

JamesBurnside: Might be good to have a `paneState: 'closed' | 'chatOpen' | 'peopleOpen'` instead of two seperate showChat and showPeople that cannot both be true

packages/react-composites/src/composites/CallWithChatComposite/MobilePane.tsx
JamesBurnside: Opposite way around, Icon is deprecated and this should indeed be FontIcon
image

JamesBurnside (approved):

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
mgamis-msft: Yep definitely necessary but will do in separate PR.

PR #1487, opened 2022-02-16
Dmceachernmsft/localCameraSwitcherE2E

dmceachernmsft:
# What
Adds new e2e test for local camera switcher button. update snapshots

# Why
test covereage to make sure the button is actionable from mobile

https://skype.visualstudio.com/SPOOL/_workitems/edit/2698411
# How Tested

ran test locally.


alkwa-msft:
I am confused about the diff around a majority of the UI snapshots? Most of them are exactly the same?

dmceachernmsft:
> I am confused about the diff around a majority of the UI snapshots? Most of them are exactly the same?

the diffs were just rebuilt with the new test included, they should be exactly the same. Next time I guess I could have just discarded their changes since the test isn't making any new shots?

JamesBurnside:
Can you mark this PR as `ui-change` to have all all the snapshots updated? Lots have been deleted and/or have not gotten new changes (e.g. "present" has regressed back to "share"):
image

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -122,6 +122,16 @@ test.describe('Call Composite E2E CallPage Tests', () => {
     }
   });
 
+  test('local camera switcher button cycles camera', async ({ pages }, testInfo) => {
+    // mobile check
+    const testName = testInfo.project.name.toLowerCase();
+    test.skip(testName.includes('desktop'));
```
prprabhu-ms, line 128: * Please add a helper `isMobileView()` or such to contain this check, also add a comment why this string comparison works. (i.e., who sets the 'desktop' in the name).
* It would be better to create a new test suite for just the mobile tests. And if possible skip all tests at the suite level.

packages/react-components/src/localization/locales/de-DE/strings.json
```diff
@@ -132,4 +132,4 @@
     "localVideoLabel": "Sie",
     "localVideoCameraSwitcherLabel": "Fahrrad-/Motorrad-Kamera"
   }
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
dmceachernmsft: ah okay, is there any particular reason why this showed up in this PR and not say in my PR for the docs on this button?

packages/react-composites/tests/browser/call/CallComposite.test.ts
dmceachernmsft: okay I will make it so!

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -30,6 +30,17 @@ const stubLocalCameraName = async (page: Page): Promise<void> => {
   });
 };
 
+/* @conditional-compile-remove-from(stable) meeting/calling-composite <Local-Camera-Switcher> */
```
prprabhu-ms, line 33: You shouldn't have to conditionally compile this. The check itself should return true/false unconditionally.

packages/react-components/src/localization/locales/de-DE/strings.json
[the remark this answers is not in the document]
prprabhu-ms: All depends on which PR is unlucky enough to be the one after TD build automation update.

prprabhu-ms (commented):
LGTM, pending sane snapshot updates :)

packages/react-composites/tests/browser/call/CallComposite.test.ts
dmceachernmsft: so should I just conditionally compile the individual test?

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -184,7 +184,7 @@ test.describe('Call Composite E2E CallPage [Mobile Only]', () => {
     await loadCallPageWithParticipantVideos(pages);
   });
 
-  test.only('local camera switcher button cycles camera', async ({ pages }, testInfo) => {
+  test('local camera switcher button cycles camera', async ({ pages }, testInfo) => {
```
prprabhu-ms, line 187: Muahahaha.

Blazingly fast e2e tests.... here they go.

dmceachernmsft: lol I caught that too when the tests passed in like 10 seconds 😂

packages/react-composites/tests/browser/call/CallComposite.test.ts
prprabhu-ms: You can conditionally compile the test suite. I just meant that this helper function doesn't need to be conditional compiled.

Is the problem that you aren't exporting the function so lint complains that it is unused? It should anyway be moved to (and exported from) a test helper file in `common/utils`

prprabhu-ms (approved):

JamesBurnside:
There's some weird issues with the end call button coloring:
image
image
cc @prprabhu-ms as he recently changed this

dmceachernmsft:
> There's some weird issues with the end call button coloring: image image cc @prprabhu-ms as he recently changed this

@prprabhu-ms in a recent PR darkened the button I believe

JamesBurnside:
> I did merge in from main, I think those changes got brought in

Ah the updatesnapshot test failed (from flakiness likely), I've hit re-run so should resolve itself soon
image

packages/react-composites/README.md
```diff
@@ -50,3 +50,32 @@ rushx build:e2e:chat
...
+The `skipTestInStableFlavor()` function is checking the environment variables of the session to check what flavor it is running in since it wont conditionally compile the test out. This function is found in the `Utils` folder under `testing` in the composites project.
+
+## Mobile Only tests
+
+If you are writing a test for only on Mobile make sure to add it to a test suite that is just for mobile. They will be marked with `[Mobile Only]` in the suite title. If there is not a suite for the page you are testing add one for that page with the `[Mobile only]` in the title.
+
+Once you have added your test to the appropriate suite use the following call to make sure it is not run on the desktop project:
+
+```Typescript
+test.only('Your test name here', async ({ pages }, testInfo) => {
```
JamesBurnside, line 75: remove `.only` here from this snippet

JamesBurnside:
> There's some weird issues with the end call button coloring: image image cc @prprabhu-ms Prathmesh Prabhu FTE as he recently changed this -- it looks to be updating to the correct coloring, but that shouldn't be happening in this PR

Ok everything passed but this is still happening -- I have no idea why...

JamesBurnside (approved):

PR #1489, opened 2022-02-16
documenting how to set a key when using loadChatComposite

alkwa-msft:
# What
We want to show how to make sure the chat-composite does not get stale when using loadChatComposite

# Why
There has a been some customer issues where when returning back to an instance of chat  composite it is out of date :(
By forcing a key, react knows that the component is cached and then tries to create a new instance.


PorterNan (approved):

samples/StaticHtmlComposites/index.html
```diff
@@ -47,7 +47,7 @@
         token: token
       },
       document.getElementById('chat-container'),
-      { visualElements: { showParticipantPane: true, showTopic: false } }
+      { options: { participantPane: true, topic: false }, key: new Date() }
```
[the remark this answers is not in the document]
alkwa-msft: yep added!

samples/StaticHtmlComposites/index.html
```diff
@@ -47,7 +47,11 @@
         token: token
       },
       document.getElementById('chat-container'),
-      { visualElements: { showParticipantPane: true, showTopic: false } }
+      {
+        options: { participantPane: true, topic: false },
+        // To prevent stale composite issues we are using a unique value (e.g the date) as a key for this component
```
JamesBurnside, line 52: What are the stale issues we are seeing? Looks like this `new Date()` is only executed once so wondering how this key causes the composite the composite to refresh, and how often this will cause the composite to re-render?

alkwa-msft: The issue was if you don't provide a key and you are doing some interesting app navigation to go from one screen to another, React will try to use the cached component. Now that we are introducing a key that by definition has a unique key then it will force a new React component to be rendered. This has implications when we want to make sure we have fresh chat messages and not stale ones. If we were using the older react component it would be thinking it is up to date and for whatever reason it is not getting new signals. By using the newer react component it will know there is more to get and will always be up to date in terms of chat messages. 

This was a stop gap that Nan came up with and perhaps we should be reviewing this as a larger feature in the future.

change/@internal-storybook-cadc1cb8-cff8-4aa6-8ad1-b8d1c2ae6854.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
prprabhu-ms, line 2: Should be `patch`. This change would go into the stable build.

alkwa-msft: Should I just update this specific changelog file?

JamesBurnside (approved):

samples/StaticHtmlComposites/index.html
prprabhu-ms: Did you mean https://gist.github.com/prprabhu-ms/39d64aa5cbb8ed8054de9b4a086aa2ec

change/@internal-storybook-cadc1cb8-cff8-4aa6-8ad1-b8d1c2ae6854.json
prprabhu-ms: Yeah, feel free to update the changefile.

prprabhu-ms (approved):

samples/StaticHtmlComposites/index.html
alkwa-msft: Yes how React chooses to reconcile components. That was a really awesome gist.

PR #1490, opened 2022-02-16
Limit image size from html

PorterNan:
# What

Try different methods, but finally find a happy path with 3 lines of code change :D

Before:
image

After:
image

And please expect more css details adjustments when we work on rich text feature of chat!
# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -105,7 +105,11 @@ export const defaultMyChatMessageContainer: ComponentSlotStyle = {
 export const defaultChatMessageContainer: ComponentSlotStyle = {
   maxWidth: '100%',
   minWidth: `${CHAT_MESSAGE_CONTAINER_MIN_WIDTH_REM}rem`,
-  marginRight: '0rem'
+  marginRight: '1rem',
+  ' img': {
```
anjulgarg, line 109: `' img` -> '> img' might be better as it will apply styles to all img tags recursively

anjulgarg (approved):

packages/react-components/src/components/styles/MessageThread.styles.ts
PorterNan: Thanks Anjul - I got you concerns about this, this is the intention we want to add it recursively. Because teams sometimes will generate nested html structure so we can't only go one level deep to its children

packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -105,7 +105,11 @@ export const defaultMyChatMessageContainer: ComponentSlotStyle = {
 export const defaultChatMessageContainer: ComponentSlotStyle = {
   maxWidth: '100%',
   minWidth: `${CHAT_MESSAGE_CONTAINER_MIN_WIDTH_REM}rem`,
-  marginRight: '0rem'
+  marginRight: '0rem',
+  ' img': {
```
JamesBurnside, line 109: What does the whitespace in `' img'` do? (is it intentional?)

JamesBurnside (approved):

alkwa-msft (approved):

packages/react-components/src/components/styles/MessageThread.styles.ts
PorterNan: Yes! This is to generate selector like "defaultContainer img" for limiting all messages inside of the chat bubble

PR #1493, opened 2022-02-17
Contextual menu showing who read my message

carocao-msft:
# What
Contextual menu showing who read my message

# How Tested
open contextual menu, hover on read by x of y, you can see a list of names who have read your message 
default icons if no avatar
Screen Shot 2022-02-16 at 8 58 21 PM

icons when having custom avatar
Screen Shot 2022-02-17 at 12 24 59 PM


packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -6,11 +6,15 @@ import {
...
   IContextualMenuItem,
+  IPersona,
+  Persona,
+  PersonaSize,
   Target,
   useTheme
 } from '@fluentui/react';
 import { _formatString } from '@internal/acs-ui-common';
 import React, { useMemo } from 'react';
+import { OnRenderAvatarCallback } from '../..';
```
JamesBurnside, line 17: avoid importing from the `src` or `components` index.ts files, instead use a more complete path here:
`../../types`

JamesBurnside: (it can cause circular dependency issues, particularly in unit tests)

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -42,6 +53,31 @@ export interface ChatMessageActionFlyoutProps {
  */
 export const ChatMessageActionFlyout = (props: ChatMessageActionFlyoutProps): JSX.Element => {
   const theme = useTheme();
+  const messageReadByList: IContextualMenuItem[] = [];
+
+  props.messageReadByNames?.forEach((person) => {
```
JamesBurnside, line 58: nit: This is a good place to use a `map` function as we are converting one-for-one each person into a contextmenuitem:
```ts
const messageReadByList: IContextualMenuItem[] | undefined = props.messageReadByNames?.map((person) => {
   ...
   return {
     key: person.name,
     ...
   };
```

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -27,12 +31,19 @@ export interface ChatMessageActionFlyoutProps {
   onRemoveClick?: () => void;
   onDismiss: () => void;
   messageReadByCount?: number;
+  messageReadByNames?: { id: string; name: string }[];
```
JamesBurnside, line 34: `Names` suffix is a misnomer here as it's an object with name and id, not just a name.
Consider just `messageReadBy` here

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -27,12 +31,19 @@ export interface ChatMessageActionFlyoutProps {
   onRemoveClick?: () => void;
   onDismiss: () => void;
   messageReadByCount?: number;
```
JamesBurnside, line 33: Is this necessary if we have `messageReadByNames` below (i.e. does `messageReadByCount === messageReadByNames.length`?

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
carocao-msft: updated

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -90,6 +126,9 @@ export const ChatMessageActionFlyout = (props: ChatMessageActionFlyoutProps): JS
             props.increaseFlyoutItemSize ? menuItemIncreasedSizeStyles : undefined
           )
         },
+        subMenuProps: {
+          items: messageReadByList
```
JamesBurnside, line 130: `messageReadByList` is inside a useMemo but I don't see it added to the dependency array -- should it? (there should be an eslint warning if it should be)

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -173,15 +178,14 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
...
            * if the last read message is read after the message A is sent, then user should have read message A as well */
-
-          let readNumber = 0;
+          const messageReadByNames: { id: string; name: string }[] = [];
           for (const k in readReceiptForEachSender) {
             const messageid = readReceiptForEachSender[k]['lastReadMessage'];
             const readTime = readReceiptForEachSender[k]['readOn'];
             if (messageid === message.id) {
-              readNumber += 1;
+              messageReadByNames.push({ id: k, name: readReceiptForEachSender[k]['name'] });
```
JamesBurnside, line 186: the `if` and `if else` execution blocks are identical, combine into one `if`:
`if (messageid === message.id || (new Date(readTime) >= new Date(message.createdOn))) {

Better, this whole for block could be written in a more functional programming manner for better clarity:
```ts
// Get the names and IDs of everyone who has read this message
const messageReadByList: { id: string; name: string }[] =
  // For each read receipt
  Object.entries(readReceiptForEachSender)
  // Filter to only read receipts that match the message and were read after the message had been created
  .filter(([_, readReceipt]) => readReceipt['lastReadMessage'] === message.id && new Date(readReceipt['readOn']) >= new Date(message.createdOn))
  // Map properties to useful array
  .map(([id, readReceipt]) => ({ id, name: readReceipt['name'] }));
```
(^ note: just a sample possibilty, this was typed in the github comment so copy-pasting it may need syntax fixes to it)

JamesBurnside: Then to reduce code all being in one function/file, this could be moved to a helper function in a seperate file:
```ts
const getParticipantsWhoHaveReadMessage(message, allReadReceipts) => 
  Object.entries(readReceiptForEachSender)
  .filter( etc...
;
```

packages/react-components/stable-review/react-components.api.md
```diff
@@ -91,6 +91,11 @@ export interface ChatMessage extends MessageCommon {
     // (undocumented)
     editedOn?: Date;
     // (undocumented)
+    messageReadByNames?: {
```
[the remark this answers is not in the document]
carocao-msft: changed name to readBy

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
carocao-msft: removed variable messageReadByCount

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
carocao-msft: changed name to messageReadBy

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
carocao-msft: changed code to use a map function

JamesBurnside (approved):

packages/chat-component-bindings/src/utils/getParticipantsWhoHaveReadMessage.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { ChatMessageWithStatus } from '@internal/chat-stateful-client';
+
+/**
+ * @private
+ *logic: Looking at message A, how do we know it's read number?
```
prprabhu-ms, line 7: I'm a bit confused by this logic.

`message.createdOn` is set when a message arrives at the ACS backend, this is before the message has been sent to a client.
`readReceipt.readOn` is set when we explicitly send a read receipt from a client.

So, it is possible that a message was sent by user 1, and user 2 sent a read receipt after that, but before it received the message from user 1. In that case, we will incorrectly report that user 2 has read the message.

Should we instead send read receipt for each message individually? Nothing in the headless SDK API seems to block us from doing that, right?

packages/chat-component-bindings/src/utils/getParticipantsWhoHaveReadMessage.ts
```diff
@@ -0,0 +1,30 @@
...
+export const getParticipantsWhoHaveReadMessage = (
+  message: ChatMessageWithStatus,
+  readReceiptForEachSender: { [key: string]: { lastReadMessage: string; readOn: Date; name: string } }
+): { id: string; name: string }[] => {
+  return (
+    Object.entries(readReceiptForEachSender)
+      // Filter to only read receipts that match the message OR the participant has read a different message after this message has been created
+      .filter(
+        ([_, readReceipt]) =>
+          readReceipt['lastReadMessage'] === message.id ||
```
prprabhu-ms, line 24: `readReceipt.lastReadMessage`, `readReceipt.readOn` etc so that you get typechecking.

packages/chat-component-bindings/src/utils/getParticipantsWhoHaveReadMessage.ts
carocao-msft: In this case since our assumption is : if user has read a message, user has read all messages before that. we can safely say that if a message is created before the readon time, the message is read

carocao-msft: readreceipt won't be sent until a message is received and seen. So the problem should not occur

prprabhu-ms: First, I was wrong -- both timestamps are computed at the backend server. That's good, so we can assume sync clocks.

prprabhu-ms: But the original concern stands:

    User1                              Server                                     User2
    ----------------------------
    Sends Message 1
                                          Message1:timestamp1
    Sends Message 2      
                                          Message2:timestamp2
                                                                                          Gets message1, swends readReceipt1
                                          ReadReceipt1:timestamp3
    Gets readReceipt1:timestamp3
                                                                                          Gets message2

Here, readReceipt1 has timestamp3 > message2's timestamp2. So User1 is shown that User2 has read message2. But that is wrong.

prprabhu-ms: Let's bring this discussion to the team channel. We can create a new work item to track fixing this race condition properly.

This PR also adds the UI surfaces for showing the detailed read report, so I don't want to block this PR on the discussion.

prprabhu-ms (approved):

packages/chat-component-bindings/src/utils/getParticipantsWhoHaveReadMessage.ts
carocao-msft: Ok sounds good! Thanks!

PR #1494, opened 2022-02-17
File Sharing: Updated the message thread component to display attached files in a message

prabhjot-msft:
# What
Updated the message thread component to display attached files in a message

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2671613
https://skype.visualstudio.com/SPOOL/_workitems/edit/2672659

# How Tested

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
```diff
@@ -22,6 +23,11 @@ type ChatMessageComponentProps = {
    * Setting to false will mean they are on a new line inside the editable chat message.
    */
   inlineAcceptRejectEditButtons: boolean;
+  /**
+   * Optional callback to render uploaded files in the message component.
+   * @beta
```
anjulgarg, line 28: no need to add a beta tag here since `ChatMessageComponentProps` are internally used.

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -23,6 +23,12 @@ type ChatMessageComponentAsMessageBubbleProps = {
   onEditClick: () => void;
   onRemoveClick?: () => void;
   strings: MessageThreadStrings;
+  userId: string;
+  /**
+   * Optional callback to render uploaded files in the message component.
+   * @beta
```
anjulgarg, line 29: Non need for beta tag as this is internally used.

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -0,0 +1,60 @@
...
+} from './file-sharing';
+import React from 'react';
+import { ChatMessage } from '@internal/react-components';
+
+/**
+ * @beta
+ */
+export interface FileDownloadCards {
+  /**
+   * User id for the message
```
prprabhu-ms, line 21: This is actually the local participant's userId, right? You can get userId of the participant that sent a message from the message itself. You need the local userId so that `downloadHandler` can take care of auth, if any.

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -0,0 +1,60 @@
...
+   */
+  userId: string;
+  /**
+   * Extension of the file used for rendering the file icon.
```
prprabhu-ms, line 25: Stale comment?

packages/react-composites/src/composites/ChatComposite/file-sharing/utils.ts
```diff
@@ -19,3 +21,12 @@ export const truncatedFileName = (fileName: string, length: number): string => {
...
 export const extension = (fileName: string): string => fileName.split('.').pop() || '';
+
+/**
+ * @private
+ *
+ * @param fileSharingMetadata
+ * @returns FileMetadata[]
+ */
+export const extractFileMetadata = (metadata: Record<string, string>): FileMetadata[] =>
+  JSON.parse(metadata.fileSharingMetadata);
```
prprabhu-ms, line 32: What happens here if there is no `fileSharingMetadata`, or if `metadata` is invalid JSON?

packages/communication-react/review/communication-react.api.md
```diff
@@ -1704,6 +1704,7 @@ export type MessageThreadProps = {
     onRenderJumpToNewMessageButton?: (newMessageButtonProps: JumpToNewMessageButtonProps) => JSX.Element;
     onLoadPreviousChatMessages?: (messagesToLoad: number) => Promise<boolean>;
     onRenderMessage?: (messageProps: MessageProps, messageRenderer?: MessageRenderer) => JSX.Element;
+    onRenderFileDownloads?: (userId: string, message: ChatMessage) => JSX.Element;
```
prprabhu-ms, line 1707: # Same design concern as for upload

Symmetric comment to my comment on the `SendBox` API extension -- this new render function is not targeted enough, and does not provide a default implementation for Contoso.

* Contoso could use this to essentially inject anything (it would be better named `onRenderMessageSuffix`)
* By default, file download doesn't work for Contoso. In order to implement it, they have to implement a lot of logic that is internal to `packages/react-composites`.

The alternative API would be:

* MessageThread gains a new prop that is `FileDownloadHandler`
  * MessageThread internally uses `FileDownloadCards`, using `FileDownloadHandler` if it is provided.

Additionally, there could be an `onRenderFileDownloads` prop as here, but it would *override* this default behavior.

@anjulgarg here as well.

prprabhu-ms (commented):
Main design concern. If we want to iterate on the API (with current impl), let's create a user story to track that work, and otherwise this PR looks good.

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
prabhjot-msft: done

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -103,7 +116,7 @@ export const ChatMessageComponentAsMessageBubble = (props: ChatMessageComponentA
           onPointerDown={() => setWasInteractionByTouch(false)}
           onKeyDown={() => setWasInteractionByTouch(false)}
           onClick={() => wasInteractionByTouch && setChatMessageActionFlyoutTarget(messageRef)}
-        />
+        ></Chat.Message>
```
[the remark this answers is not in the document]
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: User Id was supposed to be local participant's id, I have updated the comment

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: Updated the comment

packages/react-composites/src/composites/ChatComposite/file-sharing/utils.ts
prabhjot-msft: added a check for fileSharingMetadata

packages/communication-react/review/communication-react.api.md
prprabhu-ms: Moved this discussion to team channel. Let's unblock this PR because this concern is in the same vein as a discussion we already have for `SendBox`. We need to revisit this API a bit.

prprabhu-ms (approved):

prabhjot-msft:
UI Chat Composite message thread component with attached files in a sent and received message.

Screen Shot 2022-02-18 at 11 05 24 AM
Screen Shot 2022-02-18 at 11 04 40 AM
Screen Shot 2022-02-18 at 11 04 11 AM

anjulgarg (approved):

anjulgarg:
A couple of crucial things are missing in this PR.
- download handler has been used incorrectly inside file download cards
- No loader animation in file card when download handler is triggered.

Let's target those items in a new task/PR

PR #1500, opened 2022-02-17
[Bugfix][Release 1.1.0] Conditionally removing participant button from chat sample header

anjulgarg:
# What
Conditionally removing participant button from chat sample header

# Why
Participant button  stops working for stable builds as it is being conditionally compiled (meeting composite)
https://skype.visualstudio.com/SPOOL/_workitems/edit/2767067

# How Tested

# Process & policy checklist

### DIFF
--- a/samples/Chat/src/app/ChatHeader.tsx
+++ b/samples/Chat/src/app/ChatHeader.tsx
@@ -32,7 +32,7 @@ export const ChatHeader = (props: ChatHeaderProps): JSX.Element => {
     <Stack horizontal={true} verticalAlign={'center'} horizontalAlign="end" className={chatHeaderContainerStyle}>
       <div className={paneButtonContainerStyle}>
         {
-          /* @conditional-compile-remove-from(stable) meeting composite */
+          /* @conditional-compile-remove-from(stable) chat-composite-participant-pane*/
           <IconButton
             onRenderIcon={() => (props.isParticipantsDisplayed ? <People20Filled /> : <People20Regular />)}
             className={mergeStyles({ color: theme.palette.neutralPrimaryAlt })}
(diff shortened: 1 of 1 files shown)


samples/Chat/src/app/ChatHeader.tsx
```diff
@@ -31,11 +31,14 @@ export const ChatHeader = (props: ChatHeaderProps): JSX.Element => {
   return (
     <Stack horizontal={true} verticalAlign={'center'} horizontalAlign="end" className={chatHeaderContainerStyle}>
       <div className={paneButtonContainerStyle}>
-        <IconButton
-          onRenderIcon={() => (props.isParticipantsDisplayed ? <People20Filled /> : <People20Regular />)}
-          className={mergeStyles({ color: theme.palette.neutralPrimaryAlt })}
-          onClick={() => props.setHideParticipants(props.isParticipantsDisplayed)}
-        />
+        {
+          /* @conditional-compile-remove-from(stable) meeting composite */
```
prprabhu-ms, line 35: This isn't really meeting composite related.

We didn't add explicit tag to this one (we should):
image

prprabhu-ms (approved):

samples/Chat/src/app/ChatHeader.tsx
anjulgarg: What is this related to?
Would you happen to know?

anjulgarg: I am pretty sure this is CallWithChat composite related. @JamesBurnside ?

JamesBurnside: This is a chat-composite feature (so this and the participant pane field in the chat composite should be something like `/* @conditional-compile-remove-from(stable) chat-composite-participant-pane*/`)

JamesBurnside (approved):

samples/Chat/src/app/ChatHeader.tsx
JamesBurnside: (more info: https://azure.github.io/communication-ui-library/?path=/docs/composites-chat-basicexample--basic-example#participant-pane)
We made this a beta feature because it doesn't work on mobile form factors and didn't want to GA a feature that didn't have full support

anjulgarg: Done

PR #1502, opened 2022-02-17
Tdbuild 2022 feb 17 - Merge into release/1.1.0 later

anjulgarg:
# What

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/localization/locales/de-DE/strings.json
```diff
@@ -45,6 +45,8 @@
   "callWithChat": {
     "peopleButtonLabel": "Personen",
     "chatButtonLabel": "Chat",
+    "moreDrawerMicrophoneMenuTitle": "Microphone",
```
JamesBurnside, line 48: Hmm looks like we're not stripping out beta feature strings (or at least the meting composite ones) -- assume that's OK?

prprabhu-ms (approved):

packages/react-composites/src/composites/localization/locales/de-DE/strings.json
anjulgarg: I don't think we can. We add en-US strings directly yo JSON file and there is no way to conditionally compile those files (IMO)

JamesBurnside: Yeah don't think so either cc @PorterNan for awareness

PR #1503, opened 2022-02-17
Read Receipt logic change: using messageid instead of readon timestamp

carocao-msft:
# What
Using message ID to see if message A is created before message B
MessageId is the timestamp when message is created 

# Why
This will prevent potential issues mentioned here https://teams.microsoft.com/l/message/19:5e150d250f724d1089c949305efdc8cc@thread.skype/1645132543769?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e9c1fc3-39df-4486-a26a-456d80e80f82&parentMessageId=1645132543769&teamName=Azure%20Communication%20Services&channelName=Web%20UI%20-%20Team&createdTime=1645132543769


packages/chat-component-bindings/src/utils/getParticipantsWhoHaveReadMessage.ts
```diff
@@ -7,22 +7,21 @@ import { ChatMessageWithStatus } from '@internal/chat-stateful-client';
...
+  readReceiptForEachSender: { [key: string]: { lastReadMessage: string; name: string } }
 ): { id: string; name: string }[] => {
   return (
     Object.entries(readReceiptForEachSender)
       // Filter to only read receipts that match the message OR the participant has read a different message after this message has been created
-      .filter(
-        ([_, readReceipt]) =>
-          readReceipt.lastReadMessage === message.id || new Date(readReceipt.readOn) >= new Date(message.createdOn)
-      )
+      .filter(([_, readReceipt]) => parseInt(readReceipt.lastReadMessage, 10) >= parseInt(message.id, 10))
```
PorterNan, line 24: Instead of parseInt, a direct string comparing is safer, it is the same result if you compare 2 number strings:
https://javascript.info/comparison#:~:text=To%20see%20whether%20a%20string%20is%20greater%20than%20another%2C%20JavaScript%20uses%20the%20so%2Dcalled%20%E2%80%9Cdictionary%E2%80%9D%20or%20%E2%80%9Clexicographical%E2%80%9D%20order.

even the max int would be 900719925474099(very long, and would be safe enough), but the string comparation would support unlimited length

PorterNan (approved):

packages/chat-component-bindings/src/utils/getParticipantsWhoHaveReadMessage.ts
carocao-msft: Updated

packages/chat-component-bindings/src/utils/getParticipantsWhoHaveReadMessage.ts
```diff
@@ -7,22 +7,21 @@ import { ChatMessageWithStatus } from '@internal/chat-stateful-client';
...
+  readReceiptForEachSender: { [key: string]: { lastReadMessage: string; name: string } }
 ): { id: string; name: string }[] => {
   return (
     Object.entries(readReceiptForEachSender)
       // Filter to only read receipts that match the message OR the participant has read a different message after this message has been created
-      .filter(
-        ([_, readReceipt]) =>
-          readReceipt.lastReadMessage === message.id || new Date(readReceipt.readOn) >= new Date(message.createdOn)
-      )
+      .filter(([_, readReceipt]) => readReceipt.lastReadMessage >= message.id)
```
prprabhu-ms, line 24: Please add a comment why this comparison is safe:
* Message IDs are timestamps
* Timestamps are in epoch time so lecixographical ordering is the same as time ordering.

prprabhu-ms (approved):

packages/chat-component-bindings/src/utils/getParticipantsWhoHaveReadMessage.ts
carocao-msft: added comments

PR #1513, opened 2022-02-18
add deprecated fix for safari IME keyboards

dmceachernmsft:
# What
adds checks for deprecated fields in the events for KeyCode and which.

# Why
Allows for the termination of composition to be detected on safari browser

# How Tested

Tested Locally with Edward on MacOS and Windows.


packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -102,7 +102,7 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
 
   const onTexFieldKeyDown = useCallback(
     (ev: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>) => {
-      if (ev.nativeEvent.isComposing) {
+      if (ev.nativeEvent.isComposing || ev.nativeEvent.keyCode === 229 || ev.nativeEvent.which === 229) {
```
JamesBurnside, line 105: Add a comment for why these checks are here (both the `isComposing` check and for why the extra `229` check is needed

JamesBurnside (approved):

prprabhu-ms (approved):

alkwa-msft (approved):

packages/react-components/src/components/InputBoxComponent.tsx
dmceachernmsft: will do!

PR #1515, opened 2022-02-19
Mobile participant menu on CallWithChat

mgamis-msft:
# What

Using the new drawer menu of PeoplePane in mobile view.

Important notes:
1. I needed to add prop `onClick?: (props?: ParticipantItemProps) => void;` to `ParticipantItem` to be able to handle changing the drawer menu based on which participant is clicked. With that, I also added `onParticipantClick?: (participant?: ParticipantListParticipant) => void;` to `ParticipantList`. I have not yet inserted the conditional compilation needed for these 2 API changes until I have confirmation from the team on these 2 API changes.

2. In composites, I needed to remove the `removeParticipant` prop provided from the `ParticipantList` selector when `mobileView = true` in `ParticipantListWithHeading` so that no menu is generated in the `ParticipantList` component. In my opinion, we should remove the `removeParticipant` prop from `ParticipantList` and let prop `onParticipantClick?: (participant?: ParticipantListParticipant) => void;` from the previous note to handle it.

3. `ParticipantItem` only shows the "..." icon when prop `menuItems` is defined. Since we don't define menuItems to avoid ContextualMenu from appearing there is no "..." icon for `ParticipantItem` for the mobile PeoplePane like in this video
https://user-images.githubusercontent.com/79475487/154791551-4a73185f-63b1-4389-ae88-1ee033322b7e.mp4
Let me know if there should be. If so we need to change the behavior of `ParticipantItem`.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2590662

# How Tested

Local CallWithChat sample.

# Process & policy checklist


packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -182,6 +186,7 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
       onClick={() => {
         setItemHovered(true);
         setMenuHidden(false);
+        onClick?.();
```
JamesBurnside, line 189: You can do that?! (the `?.()`) I love it! No more `fn && fn()`

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -92,6 +92,10 @@ export interface ParticipantItemProps {
    * Optional strings to override in component
    */
   strings?: Partial<ParticipantItemStrings>;
+  /**
+   * Optional callback when component is clicked. If this is defined then contextual menu will not shown when component is clicked.
```
JamesBurnside, line 96: nit typo shown->show:
```suggestion
   * Optional callback when component is clicked. If this is defined then contextual menu will not show when component is clicked.
```

change/@internal-react-components-846df3f2-5947-4126-8ffc-615f1d49f540.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
JamesBurnside, line 2: this is actually a minor (or patch) change as this will be in stable builds

packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -44,6 +45,7 @@ export const ParticipantListWithHeading = (props: {
   title?: string;
   onFetchAvatarPersonaData?: AvatarPersonaDataCallback;
   onFetchParticipantMenuItems?: ParticipantMenuItemsCallback;
+  onParticipantClick?: (props?: ParticipantListParticipant) => void;
```
JamesBurnside, line 48: is `onParticipantClick` a `ParticipantListProps` now? i.e. should onParticipantClick exist inside participantListProps and not be a seperate prop here

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
```diff
@@ -89,29 +97,50 @@ export const EmbeddedPeoplePane = (props: {
...
+  const [menuItems, setMenuItems] = useState<_DrawerMenuItemProps[]>([]);
 
-  const participantListProps = useMemo(() => {
+  const participantListProps: ParticipantListProps = useMemo(() => {
     const onRemoveParticipant = async (participantId: string): Promise<void> =>
       removeParticipantFromCallWithChat(callAdapter, chatAdapter, participantId);
     return {
       ...participantListDefaultProps,
-      onRemoveParticipant
+      onRemoveParticipant: mobileView ? undefined : onRemoveParticipant
```
JamesBurnside, line 112: Add a comment here why is the onRemoveParticipant callback is not supplied in mobile view

packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -70,6 +72,7 @@ export const ParticipantListWithHeading = (props: {
             />
           )}
           onFetchParticipantMenuItems={onFetchParticipantMenuItems}
+          onParticipantClick={props.onParticipantClick}
```
JamesBurnside, line 75: Can remove this line as onParticipantList should be part of `{...participantListProps}` (when [other comment](https://github.com/Azure/communication-ui-library/pull/1515/files#r812139544) is resolved)

JamesBurnside (approved):

packages/react-components/src/components/ParticipantItem.tsx
prprabhu-ms: This comment says that contextual menu will not be shown when this is defined.

But I don't see how that's done in the logic below. `menuHidden` changes the same way as before...

packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -77,6 +77,8 @@ export type ParticipantListProps = {
   onRemoveParticipant?: (userId: string) => void;
   /** Optional callback to render custom menu items for each participant. */
   onFetchParticipantMenuItems?: ParticipantMenuItemsCallback;
+  /** Optional callback when rendered ParticipantItem is clicked */
```
prprabhu-ms, line 80: Is it true that if this is defined then `onFetchParticipantMenuItems` is ignored? If so, please note this in the docstring.

Also, what is our plan for supporting custom participant menu props set by Contoso in the composite API? They have to passed in to the drawer now, i.e., we have to convert from IContextMenuItem -> _DrawerMenuItem in the composite layer. This feels clunky -- on desktop `ParticipantList` handles the per-particient menu, but on mobile it doesn't.

I don't have a solution here, simply complaints :(

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
```diff
@@ -89,29 +97,50 @@ export const EmbeddedPeoplePane = (props: {
...
       participantListProps={participantListProps}
       onFetchAvatarPersonaData={props.onFetchAvatarPersonaData}
-      title={callWithChatStrings.peoplePaneSubTitle}
+      title={locale.strings.callWithChat.peoplePaneSubTitle}
+      onParticipantClick={(participant?: ParticipantListParticipant) => {
+        if (participant?.userId !== participantListProps.myUserId) {
+          setMenuItems([
+            {
+              itemKey: 'remove',
+              text: locale.component.strings.participantItem.removeButtonLabel,
```
prprabhu-ms, line 126: Don't use a string from `locale.component` inside composites.

Define a new one in `locale.composite.callWithChat` instead.

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
```diff
@@ -89,29 +97,50 @@ export const EmbeddedPeoplePane = (props: {
...
+              onItemClick: () => {
+                if (participant?.userId) {
+                  participantListDefaultProps.onRemoveParticipant?.(participant?.userId);
+                }
+                setMenuItems([]);
+              },
+              iconProps: {
+                iconName: 'UserRemove'
+              }
```
prprabhu-ms, line 135: This button should be disabled when the remote participant is not an ACS user. See the logic inside `ParticiantItem`.

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
```diff
@@ -143,6 +177,19 @@ export const EmbeddedPeoplePane = (props: {
   );
 };
 
+const drawerContainerStyles: IStackStyles = {
+  root: {
+    position: 'absolute',
```
prprabhu-ms, line 182: You should still make sure that this element's siblings have `position: 'relative'` set so that any zIndex from inside those elements won't compete with this one.

(So, you want to `position:relative` for `participantList`).

packages/react-components/src/components/ParticipantItem.tsx
mgamis-msft: Sorry, my mistake. I was planning on not showing the contextual menu when `onClick` is defined but decided not to in the end  because I think would be "anti-Fluent UI" behavior. Contextual menu is only shown when there are menuItems. Will fix this comment.

packages/react-components/src/components/ParticipantItem.tsx
mgamis-msft: Yup. Recently learned it :)

packages/react-components/src/components/ParticipantList.tsx
mgamis-msft: Sorry for the confusion here. `onFetchParticipantMenuItems` is not affected regardless of `onParticipantClick` being defined or not. `onParticipantClick` is meant to add additional behavior Contoso may want to add. 

If Contoso wanted to support custom participant menu props for participants in the `ParticipantList`, they would use the prop `onFetchParticipantMenuItems`. But if they wanted to use the _DrawerMenu to duplicate the custom participant menu props I think they would have to do the conversion of each IContextMenuItem to _DrawerMenuItem. Perhaps we can unify them in some way. Do we plan on exposing DrawerMenu in our API the future? @JamesBurnside If not we don't have to worry.

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
mgamis-msft: Good catch. Do we plan on having a `disabled` prop for `_DrawerMenuItem` like in `IContextualMenuItem` @JamesBurnside?

packages/react-components/src/components/ParticipantList.tsx
JamesBurnside: This is a good example of where we designed our API to be too open and hence inflexible -- lets make a task to support `onFetchParticipantMenuItems` on mobile and bring this up in feature weekly as the solution could be messy

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
JamesBurnside: If it needs it make a task and lets do it!
(Though I actually prefer the flow of the button always being enabled and if a user tries to remove a Teams user they then get a UFD-style message that they were denied due to privilages error -- this would extend better to when interop scenarios _can_ remove participants)

packages/react-components/src/components/ParticipantList.tsx
JamesBurnside: This would cause further divide in call -> callWithChat but perhaps we limit the `onFetchParticipantMenuItems` interface for call with chat composite to be the intersection of IContextMenuItem and _DrawerMenuItem interfaces (_DrawerMenuItem is based of IContextMenuItem so they are similar)

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
prprabhu-ms: I'd say let's duplicate the functionality we have so are (i.e. disabled button)

About disabled vs error message: It's best to tell the user when they can't do something as soon as we can. So if we know apriori that they can't remove a user, disable a button. If they try and then it fails, give them an error.

prprabhu-ms:
Please merge from `main` so that it's easier to review this PR again. Right now there is a large unrelated diff.

.github/workflows/npm-release-publish.yml
```diff
@@ -6,6 +6,9 @@ on:
       npm-tag:
         description: 'Npm tag for the release, e.g. latest, next, dev'
         required: true
+      release-flavor:
```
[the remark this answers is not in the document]
mgamis-msft: Fixed.

packages/react-composites/src/composites/CallWithChatComposite/SidePane.tsx
mgamis-msft: I have added the `disabled` prop in this PR. The changes are summarized in this commit: https://github.com/Azure/communication-ui-library/pull/1515/commits/1b52936cfb78c21776d5c2a01af784ef25f29706.
Here is a video of how the disabled menu item looks:
https://user-images.githubusercontent.com/79475487/155943629-9ab52766-2d7c-4d8c-be52-5be63ea1bf4a.mp4

packages/react-components/src/components/ParticipantList.tsx
mgamis-msft: I think we can avoid this by converting ContextualMenuItem to _DrawerMenuItem ourselves if mobile view. We access the menuItems of `ParticipantItem` through a callback. The change to this branch would be this commit: https://github.com/Azure/communication-ui-library/commit/335dd81e9123207afe24c809bf03a66047d3070b

mgamis-msft: Can we land this PR first @prprabhu-ms? Currently `onFetchParticipantMenuItems ` prop is not used in CallWithChat composite for the `ParticipantList`. I will add its usage to create `IContextualMenuItem` and `_DrawerMenuItem` in another PR for better readability.

packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
```diff
@@ -105,6 +122,14 @@ const drawerMenuItemRootStyles = (hoverBackground: string, fontSize: IRawStyle):
   }
 });
 
+const disabledDrawerMenuItemRootStyles = (background: string): IStyle => ({
```
prprabhu-ms, line 125: We usually just pass in `theme` and then pull out the colors you want.

prprabhu-ms (approved):

PR #1516, opened 2022-02-22
Added copy invite link to mobile people pane in CallWithChat (with refactor)

mgamis-msft:
# What

- Added copy invite link to mobile people pane in CallWithChat
- Refactored chat and people pane files.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2590660

# How Tested

On local CallWithChat sample. Please check UI test snapshot updates.

# Process & policy checklist


JamesBurnside:
Can you make sure we have a join link set in the UI test app to capture this

packages/react-composites/src/composites/CallWithChatComposite/styles/EmbeddedPeoplePane.styles.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { IButtonStyles } from '@fluentui/react';
+
+export const copyLinkButtonStyles: IButtonStyles = {
+  root: {
+    height: '3rem',
+    border: 'none',
```
JamesBurnside, line 9: Doesn't look like primary buttons have borders that differ from the buttons color: https://developer.microsoft.com/en-us/fluentui#/controls/web/button
Is this needed?

packages/react-composites/src/composites/CallWithChatComposite/EmbeddedPeoplePane.tsx
```diff
@@ -0,0 +1,107 @@
...
+        onChatButtonClicked={props.onChatButtonClick}
+        onPeopleButtonClicked={props.onPeopleButtonClick}
+      >
+        <Stack verticalFill tokens={peoplePaneContainerTokens}>
+          {participantList}
+          {inviteLink && (
+            <PrimaryButton
+              onClick={() => copy(inviteLink)}
+              styles={copyLinkButtonStyles}
+              onRenderIcon={() => <FontIcon iconName="Link" style={linkIconStyles} />}
```
JamesBurnside, line 70: This is using the old style link icon. We should use the new one https://iconcloud.design/search/filter/Fluent%20System%20Library/Fluent%20Regular/link/d0359d032-3821b375b

Will need imported from the react-icons package and made customizable by adding to the `DEFAULT_COMPOSITE_ICONS`

Probably best as a separate PR as call composite should be updated with this also

JamesBurnside (approved):

dmceachernmsft (approved):

PR #1518, opened 2022-02-22
InputBox API update

dmceachernmsft:
# What
Updates autoFocus API for the input box for more clarity

# Why
Improves clarity in the API for the input box

from ARB review
# How Tested

Tested locally.


packages/communication-react/review/communication-react.api.md
```diff
@@ -759,7 +759,7 @@ export type ChatCompositeOptions = {
     errorBar?: boolean;
     participantPane?: boolean;
     topic?: boolean;
-    autoFocus?: 'sendBoxTextField' | false;
+    autoFocus?: 'sendBoxTextField' | 'disabled';
```
JamesBurnside, line 762: Is there an advantage to having `'disabled'` instead of just `undefined`?

JamesBurnside: Do we autofocus by default if nothing is passed in here?

packages/react-components/src/components/SendBox.tsx
```diff
@@ -112,7 +112,7 @@ export interface SendBoxProps {
   /**
    * boolean to determine if the input box has focus on render or not.
```
JamesBurnside, line 113: Add what is the default behavior here
`@defaultValue 'disabled'`

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -68,7 +68,7 @@ export type ChatCompositeOptions = {
   /**
    * Set focus on the composite when the composite first mounts.
```
JamesBurnside, line 69: Add defaultValue here also in the tsdoc comment

packages/communication-react/review/communication-react.api.md
dmceachernmsft: the only thing I can think of is that it will just provide clarity of what autofocus would do. having it undefined might hint it would go do a 'default' behavior instead of explicitly say there will be no autofocus present

dmceachernmsft: we can remove it, the default is no focus

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -66,9 +66,10 @@ export type ChatCompositeOptions = {
    */
   topic?: boolean;
   /**
-   * Set focus on the composite when the composite first mounts.
+   * enumerable to determine if the input box has focus on render or not.
+   * When undefined nothing has focus on render
    */
-  autoFocus?: 'sendBoxTextField' | false;
+  autoFocus?: 'sendBoxTextField' | undefined;
```
prprabhu-ms, line 72: Do we need the explicit `| undefined` here? The field is already optional.

packages/react-components/src/components/SendBox.tsx
```diff
@@ -110,9 +110,10 @@ export interface SendBoxProps {
    */
   strings?: Partial<SendBoxStrings>;
   /**
-   * boolean to determine if the input box has focus on render or not.
+   * enumerable to determine if the input box has focus on render or not.
+   * When undefined nothing has focus on render
    */
-  autoFocus?: 'sendBoxTextField' | false;
+  autoFocus?: 'sendBoxTextField' | undefined;
```
prprabhu-ms, line 116: Do we need the explicit `| undefined` here? The field is already optional

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -62,7 +62,7 @@ type InputBoxComponentProps = {
   errorMessage?: string | React.ReactElement;
   disabled?: boolean;
   styles?: InputBoxStylesProps;
-  autoFocus?: 'sendBoxTextField' | false;
+  autoFocus?: 'sendBoxTextField' | 'disabled';
```
prprabhu-ms, line 65: I don't see this being set to `'disabled'` anywhere.

packages/communication-react/review/communication-react.api.md
anjulgarg: Adding 'disabled' now deviates from our composite config options.
As a practice, we decided if an optional prop is undefined, it means that that feature is disabled.
What do you think @PorterNan @prprabhu-ms ?

packages/react-components/src/components/InputBoxComponent.tsx
dmceachernmsft: ah I missed that one will also remove like the undefineds above!

PR #1531, opened 2022-02-23
Fill `CallWithChatCompositeIcons` type with all used customized icon

JamesBurnside:
# What
* Ensure all icons used by callwithchat composite are explicitly listed -- it is no longer a mere aggregation of call and chat


packages/react-composites/src/composites/common/icons.tsx
```diff
@@ -181,17 +181,76 @@ export type CallCompositeIcons = Partial<
   >
 >;
 
-/* @conditional-compile-remove-from(stable): meeting-composite */
```
prprabhu-ms, line 184: This type should stay in beta-only for now.

prprabhu-ms (approved):

packages/react-composites/src/composites/common/icons.tsx
JamesBurnside: It's removed from export in the index.ts now -- needed used in the CallWithChatComposite file in stable build as its still built but also just not exported

dmceachernmsft (approved):

PR #1536, opened 2022-02-24
[DO NOT SUBMIT] Demonstrate extracted custom hooks

prprabhu-ms:
A PR to discuss the idea of providing custom `use*Adapter` hooks from our API.

### DIFF
--- a/packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
+++ b/packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
@@ -1,70 +1,94 @@
 import { AzureCommunicationTokenCredential, CommunicationUserIdentifier } from '@azure/communication-common';
 import {
   CallAdapter,
-  createAzureCommunicationCallAdapter,
+  CallComposite,
   ChatAdapter,
+  ChatComposite,
+  createAzureCommunicationCallAdapter,
   createAzureCommunicationChatAdapter,
   fromFlatCommunicationIdentifier
 } from '@azure/communication-react';
 import React, { useEffect, useMemo, useState } from 'react';
 
+// Fill in the following configuration inline for this sample.
+// In a real application, these must be supplied securely to the client application.
+const DISPLAY_NAME = '<Display Name>';
+const ENDPOINT_URL = '<Azure Communication Services Resource Endpoint>';
+const TOKEN = '<Azure Communication Services Access Token>';
+const USER_ID = '<Azure Communication Services Identifier>';
+// CallComposite specific configuration:
+// For Group Id, developers can pass any GUID they can generate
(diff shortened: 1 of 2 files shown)


packages/storybook/stories/QuickStarts/snippets/QuickstartCompositeAdapter.snippet.tsx
```diff
@@ -1,70 +1,94 @@
...
           })
         );
-        setCallAdapter(
-          await createAzureCommunicationCallAdapter({
+    })();
+  }, [credential, displayName, groupId, userId]);
+  return callAdapter;
+}
+
+function useChatAdapter(
```
prprabhu-ms, line 69: There are a few problems with providing `useChatAdapter` as part of our API:

* `userId` argument is an object. We'd either have to accept flat string `userId` for this method (breaking from the API of `createAzureCommunicationChatAdapter`, or Contoso would have to memoize the passed in object to avoid unnecessary recreation of the adapter.
* `credential` is similarly an object. There is an extra complication with `credential` that Contoso needs to decide what to render if they fail to generate the `credential`. `useChatAdapter` must be called unconditionally (rule of hooks) so it will make for non-trivial logic in Contoso's component to ensure that `credential` is valid before calling the `useChatAdapter` hook.

I'd argue that while it might make sense to extract a local hook for creating the adapter, like here, it's not a good idea to introduce that in our quick-starts. The local custom hook I added in this PR requires the reader to be comfortable with custom hooks, without adding much value (if we provided the hook in our API, that'd be a value-add).

prprabhu-ms: @JamesBurnside @PorterNan

prprabhu-ms: (Note that we can't accept `token` as a string. In a real application, the provided `credential` should auto-refresh the token as needed).

JamesBurnside: Can we do any internal memorization inside the hook where we explicitly check the contents of the userId and credential objects?

JamesBurnside: I think the value with the shorter code snippets is still worth it. I think that while there is the added complexity of the useAdapter custom hook, there is a removal of complexity about managing the adapter in the snippet (i.e. useStates and useEffects lying around, not having to read about async functions that trigger rerenders).

Code for just chat is super small:
image

(userId needing memoized in some fashion, either useMemo or we could perhaps internally memoize the userId string)

Would be nice if we could not require the useMemos though

prprabhu-ms:
This was just a demo.

PR #1537, opened 2022-02-24
Modifying file sharing to allow attaching already uploaded files to message

anjulgarg:
**NOTE: Tweaked version of this PR  is https://github.com/Azure/communication-ui-library/pull/1556
Please review that instead**

# What
Modify File Sharing to allow users to attach files already uploaded to a storage. (Without using the attach file button)
For example, scenarios where file already exists in a storage like azure blob and doesn't need to be uploaded via a user.

Sample Usage by Contoso
```tsx
adapter.registerFileUploads([
  new ObservableFileUpload('uploadedFileName.pdf', { name: 'downloadFileName.pdf', extension: 'ext', url: 'downloadUrl' })
]);
```

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -87,35 +87,100 @@ export interface FileUploadManager {
...
 
 /**
- * An internal interface used by the Chat Composite to drive the UI for file uploads.
+ * A class containing {@link EventEmitter} object for emitting events and methods for
+ * subscribing and unsubscribing to file upload events.
  * @beta
  */
-export interface ObservableFileUpload extends FileUploadEventEmitter {
+export class FileUploadEventEmitter {
+  _emitter: EventEmitter;
```
JamesBurnside, line 95: nit: make private?
```suggestion
  private _emitter: EventEmitter;
```

packages/communication-react/review/communication-react.api.md
```diff
@@ -1470,8 +1472,11 @@ export interface FileUploadAdapter {
     registerFileUploads?: (fileUploads: ObservableFileUpload[]) => void;
 }
 
-// @beta (undocumented)
-export interface FileUploadEventEmitter {
+// @beta
+export class FileUploadEventEmitter {
```
JamesBurnside, line 1476: Do we expect contoso to create instances of this class? Or just of `ObservableFileUpload`?
If not we should keep this as an interface (of which we have an internal implementation that handles the events subscription)

JamesBurnside: Oh maybe ignore this, if `ObservableFileUpload` is extending this we have to publicly export this as a class

JamesBurnside: I guess to avoid exporting this as a class and keep as an interface it would have to be something like:
```
/** public */
interface FileUploadSubscriber {
  on()...
  off()...
}

/** private */
class FileUploadEventEmitter implements FileUploadSubscriber {
  private _emitter: EventEmitter
  ... impl ...
}

/** public */
class ObservableFileUpload implements FileUploadSubscriber {
  private fileUploadSubscriber: FileUploadSubscriber;

  // propagate to privately implemented FileUploadEventEmitter 
  on(event: 'uploadFailed', listener: UploadFailedListener) {
    this.fileUploadSubscriber.on(ev, listener);
  }
  etc.
}
```

Not sure if we feel strongly exporting classes. The reason to avoid is they are harder to maintain/easier to break if a customer ends up using the exported class - interfaces are much simpler

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
anjulgarg: Needs to be used by a child class `FileUpload`

packages/communication-react/review/communication-react.api.md
anjulgarg: So in your code sample, `ObservableFileUpload` will need to receive an instance of `FileUploadSubscriber` as a constructor parameter.
However, the expectation here is that Contoso will create a `new ObservableFileUpload` object for uploading files that are already uploaded to a cloud store.

PorterNan, line 1476: Any concerns of this must be a class? I guess we might want to follow the principle of "Separating implementation and interface", it is something like CallAdapter and AzureCommunicationCallAdapter - 
You can export an interface and use it everywhere, but implementation (the real class) could be created by another constructor or factory method(createAzureCommunicationAdapter), this leaves either us or developers more flexibility to implement them in different forms (it could be a different constructor, or it doesn't have to contain an  eventEmitter in the instance)

anjulgarg: @JamesBurnside created an alternative PR
https://github.com/Azure/communication-ui-library/pull/1556
I think it will address all your comments.

anjulgarg: @PorterNan 
created an alternative PR
https://github.com/Azure/communication-ui-library/pull/1556
I think it will address all your comments.

PR #1540, opened 2022-02-24
Added LocalAndRemotePIP to mobile EmbeddedPeoplePane in CallWithChat

mgamis-msft:
# What

Added LocalAndRemotePIP to mobile EmbeddedPeoplePane in CallWithChat

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2590664

# How Tested

Local CallWithChat test with 2 participants.
https://user-images.githubusercontent.com/79475487/156492337-ff24ee0a-fe57-45c5-8fe0-4561d4ead685.mp4
Notice that there are stream re-renders when switching between VideoGallery and mobile PeoplePane.

Please see UI snapshot update in 'Files changed' tab

Also I manually inserted multiple dummy participants in code to show what it looks like a lot of participants.
https://user-images.githubusercontent.com/79475487/156490507-14840325-a500-428c-a4b1-696653f790ba.mp4

# Process & policy checklist


dmceachernmsft:
seems some of the icons aren't working right? 
image

mgamis-msft:
> seems some of the icons aren't working right? image

Looks like the CI updated snapshot on failure and the icons loaded properly. But it is showing some flakiness.

packages/react-composites/src/composites/CallWithChatComposite/EmbeddedPeoplePane.tsx
```diff
@@ -62,14 +75,32 @@ export const EmbeddedPeoplePane = (props: {
...
+          {
+            // Only render LocalAndRemotePIP when this component is NOT hidden because VideoGallery needs to have
+            // possession of the dominant remote participant video stream
+            !props.hidden && (
+              <Stack horizontalAlign="end" styles={localAndRemotePIPContainerStyles}>
+                <Stack styles={localAndRemotePIPStyles}>
+                  <LocalAndRemotePIP
+                    {...pictureInPictureProps}
+                    {...pictureInPictureHandlers}
+                    // eslint-disable-next-line @typescript-eslint/no-empty-function
```
prprabhu-ms, line 88: Instead of telling eslint to complain that we don't have a comment saying why function is empty, just move the reason inside :)

```ts
onClick={() => { /* no onClick behavior in peoplePane */ }
```

Actually, more importantly, why is there an `onClick` prop that isn't used in this case? Is this for the camera switcher? Should this be added in the people pane in a future PR? In that case, please add a TODO in the comment instead.

packages/react-composites/src/composites/CallWithChatComposite/EmbeddedPeoplePane.tsx
```diff
@@ -62,14 +75,32 @@ export const EmbeddedPeoplePane = (props: {
...
-              styles={copyLinkButtonStyles}
-              onRenderIcon={() => <FontIcon iconName="Link" style={linkIconStyles} />}
-              text="Copy invite link"
-            />
+            <Stack.Item styles={copyLinkButtonContainerStyles}>
+              <PrimaryButton
+                onClick={() => copy(inviteLink)}
+                styles={copyLinkButtonStyles}
+                onRenderIcon={() => <FontIcon iconName="Link" style={linkIconStyles} />}
+                text="Copy invite link"
```
prprabhu-ms, line 101: Not for this PR: This string needs to be localized.

packages/react-composites/src/composites/CallWithChatComposite/EmbeddedPeoplePane.tsx
prprabhu-ms: Looking at this again, `LocalAndRemotePipProps.onClick` should simply be made optional.

packages/react-composites/src/composites/CallWithChatComposite/styles/EmbeddedPeoplePane.styles.ts
```diff
@@ -1,16 +1,38 @@
...
+ * @private
+ */
+export const peoplePaneContainerTokens: IStackTokens = {
+  childrenGap: '0.5rem'
+};
+
+/**
+ * @private
+ */
+export const localAndRemotePIPContainerStyles: IStackStyles = { root: { height: '100%' } };
```
prprabhu-ms, line 16: I dont understand how the PiPiP is overlaid on top of other content (where is the zIndex set? How large is the element? Does it cover the entire screen, but let's clicks go through?)

It would be useful to add a comment here explaining what area on the screen this container fills (and so, why is 100% needed).

Also, could you instead set `grow` prop in the `Stack`?

JamesBurnside:
Copy invite link and the floating video shouldn't be scrollable in my opinion, only the participant list should be

packages/react-composites/src/composites/CallWithChatComposite/EmbeddedPeoplePane.tsx
JamesBurnside: It should be localized already somewhere as call composite uses this (but yes in a separate PR would be better)

mgamis-msft: This string is localized in PR #1515

packages/react-composites/src/composites/CallWithChatComposite/styles/EmbeddedPeoplePane.styles.ts
mgamis-msft: Yes, the `grow` prop works. Removed that `localAndRemotePIPContainerStyles`. The PiPiP is not overlaid in this PR. It is just a child of the Stack. The Stack items are the PartcipantList, PiPiP, and the 'Copy Invite Link' button, in that order. Working on making it overlaid and draggable.

prprabhu-ms (approved):

prprabhu-ms:
I'm comfortable with merging this and then working on making PiPiP floating and movable in a follow up PR.

mgamis-msft:
> Copy invite link and the floating video shouldn't be scrollable in my opinion, only the participant list should be -- then the PiPiP should be movable in case it is occluding something -- making it movable though we can work on separately. Definitely ask AlexP about this



> Copy invite link and the floating video shouldn't be scrollable in my opinion, only the participant list should be -- then the PiPiP should be movable in case it is occluding something -- making it movable though we can work on separately. Definitely ask AlexP about this

Yes, you are right. Confirmed with Alex P. PiPiP is now draggable and persists position when mobile People Pane is closed and open again.

dmceachernmsft (approved):

JamesBurnside (approved):
❤️

PR #1542, opened 2022-02-24
Add new Accessibility friendly local video tile

dmceachernmsft:
# What

Add new instance of localvideo tile not in modal

# Why

Allows for keyboard navigation in and out of the local video tile for when controls are present.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2769964
https://skype.visualstudio.com/SPOOL/_workitems/edit/2769965
# How Tested

Ran locally.


JamesBurnside (changes requested):
Looks like there's some regressions regarding tile size:
image
image

packages/communication-react/review/communication-react.api.md
```diff
@@ -2292,7 +2292,7 @@ export type VideoGalleryParticipant = {
 // @public
 export interface VideoGalleryProps {
     dominantSpeakers?: string[];
-    layout?: 'default' | 'floatingLocalVideo';
+    layout?: 'default' | 'floatingLocalVideo' | /* @conditional-compile-remove-from(stable) meeting/calling-composite <Local-Camera-Switcher> */ 'controlsLocalVideo';
```
JamesBurnside, line 2295: `controlsLocalVideo` is difficult to parse meaning from - what controls and how does it differ from the other options? ...though I'm drawing a blank at a good solid name, will have a think...

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -192,6 +196,7 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
   const strings = { ...localeStrings, ...props.strings };
 
   const shouldFloatLocalVideo = !!(layout === 'floatingLocalVideo' && remoteParticipants.length > 0);
+  const shouldAddControlsLocalVideo = shouldAddControlsToLocalVideoTrampoline(layout, remoteParticipants);
```
JamesBurnside, line 199: would be good to encapsulate `shouldFloatLocalVideo` and `shouldAddControlsLocalVideo` into an enum as they are mutually exclusive.
something like: `const localTileHost = 'grid' | 'floatingDraggable' | 'floatingStatic'`

packages/communication-react/review/communication-react.api.md
dmceachernmsft: I was thinking on this name too, I could go more specific to something like `LocalCameraButtonLocalVideo` since thats what its for but went with `controlsLocalVideo` since in the future we might add different controls? does that maybe help with thinking of something better?

packages/react-components/src/components/VideoGallery.tsx
dmceachernmsft: So looking into this refactor, I will make it in a seperate PR since the styles all depend on the boolean evaluation of shouldFloatLocalVideo. the style functions will need to be updated to support the new logic if we change this.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -72,6 +73,15 @@ export interface VideoGalleryStrings {
...
 }
 
+/**
+ * @public
+ */
+export type LocalTileMode =
+  | 'default'
+  | 'floatingLocalVideo'
+  | /* @conditional-compile-remove-from(stable) meeting/calling-composite <Local-Camera-Switcher> */ 'controlsLocalVideo'
+  | undefined;
```
prprabhu-ms, line 83: The type shouldn't include `undefined`. Instead, make the field that uses this type optional. (It actually already is optional `VideoGallyerProps.layout`)

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -101,7 +111,7 @@ export interface VideoGalleryProps {
    */
   styles?: VideoGalleryStyles;
   /** Layout of the video tiles. */
-  layout?: 'default' | 'floatingLocalVideo';
+  layout?: LocalTileMode;
```
prprabhu-ms, line 114: The name of this field and type don't match. This is a smell that one or the other isn't scoped right.

Since `layout` is already part of public API, we can't rename that. I'd recommend renaming the type to be `VideoGalleryLayout`

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -72,6 +73,15 @@ export interface VideoGalleryStrings {
   localVideoCameraSwitcherLabel: string;
 }
 
+/**
+ * @public
+ */
+export type LocalTileMode =
+  | 'default'
+  | 'floatingLocalVideo'
+  | /* @conditional-compile-remove-from(stable) meeting/calling-composite <Local-Camera-Switcher> */ 'controlsLocalVideo'
```
prprabhu-ms, line 82: See my comment about scoping of this type.

`layout = 'controlsLocalVideo'` doesn't make much sense in the context of the video gallery (I read that as: the layout of the video gallery is such that it controls the local video!)

Should this be a separate prop `enableLocalCameraSwitcher?: boolean` that is only meaningful when `layout = 'floatingLocalVideo'`. In that case, we should throw an exception if it is set with `layout = 'default'`.

packages/react-components/src/components/VideoGallery.tsx
prprabhu-ms: OK with a separate PR. This becomes especially important in the light of my comment on changing the API -- we'll have two different props that are related to each other. It's best to merge them into a single enum as soon as possible in the implementation.

packages/react-components/src/components/VideoGallery.tsx
dmceachernmsft: okay will remove this!

packages/react-components/src/components/VideoGallery.tsx
dmceachernmsft: okay this makes sense! will update that!

packages/react-components/src/components/VideoGallery.tsx
dmceachernmsft: oh we have that prop already. I didn't think to use it. will make this refactor

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -381,6 +400,14 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
             {localVideoTile}
           </Modal>
         ))}
+      {
+        /* @conditional-compile-remove-from(stable) meeting/calling-composite <Local-Camera-Switcher> */
+        showCameraSwitcherInLocalPreview && localParticipant && (
```
JamesBurnside, line 405: add a comment why showCameraSwitcherInLocalPreview means we don't use the modal

JamesBurnside (approved):

mgamis-msft:
For this snapshot change: https://github.com/Azure/communication-ui-library/pull/1542/files#diff-d5323e24208d4b0b80f18a2138781fe2af7c1d94833436c234d6095727c60c4e

The local video tile looks a bit weird without a top and left border.
image

Are we ok with this?

dmceachernmsft:
> For this snapshot change: https://github.com/Azure/communication-ui-library/pull/1542/files#diff-d5323e24208d4b0b80f18a2138781fe2af7c1d94833436c234d6095727c60c4e
> 
> The local video tile looks a bit weird without a top and left border. image
> 
> Are we ok with this?

will fix this!

mgamis-msft (approved):

prprabhu-ms (approved):

PR #1543, opened 2022-02-24
Hotfix: Wrap JSON.stringify in try-catch when logging state changes

JamesBurnside:
# What
* Wrap state changes in verbose logging with a try catch

# Why
* Causing exceptions:
   image

This error was seen when updating the selfhosting app. Still investigating _why_ the selfhost app sees this but the samples in this repo are not.

# How Tested
n/a

### DIFF
--- a/packages/acs-ui-common/src/safeStringify.ts
+++ b/packages/acs-ui-common/src/safeStringify.ts
@@ -8,7 +8,7 @@
  *
  * @internal
  */
-export const _safeStringify = (
+export const _safeJSONStringify = (
   value: unknown,
   replacer?: ((this: unknown, key: string, value: unknown) => unknown) | undefined,
   space?: string | number | undefined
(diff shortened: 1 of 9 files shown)


packages/acs-ui-common/src/safeStringify.ts
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT license.
+
+/**
+ * Wrap JSON.stringify in a try-catch as JSON.stringify throws an exception if it fails.
+ *
+ * Use this only in areas where the JSON.stringify is non-critical and OK for the JSON.stringify to fail, such as logging.
+ *
+ * @internal
+ */
+export const _safeStringify = (
```
anjulgarg, line 11: might I suggest changing it to `_safeJSONStringify` just so that there is absolute clarity about what it is used for just by reading the name.

anjulgarg (approved):

packages/acs-ui-common/src/safeStringify.ts
JamesBurnside: Yep I like that better, updated

dmceachernmsft (approved):

PR #1545, opened 2022-02-24
Support per-feature conditional compilation directives

prprabhu-ms:
# What

Add support for conditional compilation directives of the form `@conditional-compile-remove(feature)` where `feature` is a predefined name of a feature.

Follow up PRs will replace existing conditional compilation directives with feature-specific ones.

This will allow us to easily switch features on/off.


anjulgarg (approved):

common/scripts/babel-conditional-preprocess.js
```diff
@@ -63,62 +65,68 @@ exports.default = babelHelper.declare((_api, opts) => {
...
-    if (removed) {
-      // We cannot remove Expression in JSXExpressionContainer cause it is not correct for AST
-      // Replacing it with jSXEmptyExpression will get us the same result
-      // There will always be only one expression under JSXExpressionContainer
-      if (path?.container?.type === 'JSXExpressionContainer') {
-        path.replaceWith(t.jSXEmptyExpression());
-      } else {
-        path.remove();
-      }
+  const shouldRemove = node.leadingComments && node.leadingComments.some((comment) => containsDirective(comment, match, featurePrefixes));
```
anjulgarg, line 93: Why is this needed?

common/scripts/babel-conditional-preprocess.js
```diff
@@ -63,62 +65,68 @@ exports.default = babelHelper.declare((_api, opts) => {
...
+    return true;
+  }
+  // Check for partial prefix first to avoid checking each feature in most cases.
+  if (!comment.value.includes(FEATURE_PREFIX)) {
+    return false;
+  }
+  if(featurePrefixes.some((match) => comment.value.includes(match))) {
+    return true;
+  }
+  throw new Error(`Unknown conditional compilation feature in: ${comment.value}`);
```
anjulgarg, line 131: Devs would benefit from actual filename and line number.
Possible to add that to the error message?

common/scripts/babel-conditional-preprocess.js
prprabhu-ms: The `forEach` just modifies each comment that contains a directive.

This check is an early return that skips other logic as well (JSX handling for now).

common/scripts/babel-conditional-preprocess.js
prprabhu-ms: I tried... but babel doesn't seem to be populating `node.loc.filename`

@PorterNan  do you know why?

In my latest commit, the filename is always `undefined`.

packages/acs-ui-common/src/conditional-compilation-sample/index.tsx
```diff
@@ -175,6 +175,35 @@ function propTrampoline(): string {
...
+ * The tagged code will no longer be removed from stable builds.
+ *
+ * Using an undeclared feature in conditional compilation directive will cause a build failure (prevents typos).
+ *
+ * The following code is not removed because it contains a stabilized feature tag.
+ */
+/* @conditional-compile-remove(stabilizedDemo) */
+export const thisIsNowStable = 42;
+
+/* But the following is removed because it contains one non-stabilized feature still. */
```
prprabhu-ms, line 196: This is not right.

If any feature is stabilized, the associated code should be included in stable build.

If a field is needed by multiple features, the first one that is stabilized needs it to be in the stable build.

prprabhu-ms: Done.

packages/acs-ui-common/src/conditional-compilation-sample/index.tsx
```diff
@@ -175,6 +175,39 @@ function propTrampoline(): string {
...
+ * The tagged code will no longer be removed from stable builds.
+ *
+ * Using an undeclared feature in conditional compilation directive will cause a build failure (prevents typos).
+ *
+ * The following code is not removed because it contains a stabilized feature tag.
+ */
+/* @conditional-compile-remove(stabilizedDemo) */
+export const thisIsNowStable = 42;
+
+/* The following is not removed because it contains one stabilized feature, along with a beta-only feature.
```
dmceachernmsft, line 196: What would a good situation to use this stack of conditional compile features be? I see what its saying but I am wondering if there is a good example of a use case?

prprabhu-ms: This is used currently in places by the camera switcher button and control bar split buttons.
That is an example of smaller features that add up to a larger feature -- we could decide to stabilize the smaller features first (just release the split buttons... because they get used elsewhere) or release the bigger feature with everything involved.

It's still up for debate whether that's a good idea or not.


Another example (doesn't exist yet) is if a field is needed by two somewhat independent features (e.g., and `options` extension to some existing API, where individual features add different fields to the options).

dmceachernmsft: Oh I see! so like how my camera switcher is dependent on the CallWithChatComposite to go live so it might still be in beta while the CallWithChatComposite is stabalized. I like it!

dmceachernmsft (approved):

common/config/babel/.babelrc.js
```diff
@@ -2,7 +2,27 @@ const plugins = [];
...
   plugins.push([
     '../../common/scripts/babel-conditional-preprocess',
-    { annotations: [{ match: '@conditional-compile-remove-from(stable)' }] }
+    {
+      // Deprecated: Use `features` instead.
+      match: '@conditional-compile-remove-from(stable)',
+      // A list of features recognized by the conditional compilation preprocessing plugin.
+      // "demo" is a special feature, used for demo purposes. For this feature,
+      // The plugin removes any AST node that is preceded by a comment that contains with the tag:
+      // @conditional-compile-remove(demo)
```
PorterNan, line 11: So we are changing from @conditional-compile-remove-from(stable) to @conditional-compile-remove(stable)
 Haha, a lot of find&replace need to be done!

common/scripts/babel-conditional-preprocess.js
```diff
@@ -63,62 +74,96 @@ exports.default = babelHelper.declare((_api, opts) => {
...
-          comment.ignore = true;
-          // Comment will be inherit to next line even it is set to 'ignore'
-          // this will cause some unexpected removals of code
-          // clear the comment to ensure nothing gets wrong
-          comment.value = '';
-        }
-      }
-    }
 
-    if (removed) {
```
PorterNan, line 112: The function scoped variable 'removed' should stay - it is intentional to remove all the following comment after the conditional compilation match

Imagine a case like this:

// conditional-compile-remove(stable)
/* This is the comment for the beta feature */
const betaFeauture = () => {}

The second comment should be also removed with the statement itself, without the function scoped the comment will still stay, and mess up our jsdoc for stable api doc

PorterNan (approved):
Looks great, might need a minor change to remove comments after conditional compiler comment to support more general cases

common/scripts/babel-conditional-preprocess.js
prprabhu-ms: This check is still there. It's just an early return (line 107 after the PR).

prprabhu-ms: Oh wait, now I understand the issue.

packages/acs-ui-common/src/conditional-compilation-sample/index.tsx
```diff
@@ -123,6 +123,14 @@ export function d(e: number, f?: number, g?: number): void {
   console.log(f, g);
 }
```
PorterNan, line 125: Would be nice to add a trailing comment here like
  console.log(f, g);
} // this is a trailing comment which is above conditional compile comment and wouldn't be removed

PR #1546, opened 2022-02-25
Use per-feature conditional compilation directives

prprabhu-ms:
# What

This PR uses the per-feature conditional compilation directives introduced in https://github.com/Azure/communication-ui-library/pull/1545

I included some cleanup:

* Some previously unclassified directives are now classified into named features
* Some duplicate tags are merged into defined features

But I did not materially modify what logic belongs to which feature. e.g., some conditional compilation comments refer to `call-with-chat-composite` as well as some smaller feature. We should likely clean those up to only use the smaller features (that `call-with-chat-composite` needs).

# Why

Stabilizing a feature is now as easy as moving that feature to `stabilizedFeatures` list!

Example PR: https://github.com/Azure/communication-ui-library/pull/1547

# How Tested

Verified that preprocessed code is unchanged:

Prep:

```bash
rm -rf ../stashed
mkdir ../stashed
rush build -t communication-react
for d in $(ls packages); do mv packages/$d/preprocessed ../stashed/$d; done
```

After this PR:

```bash
rush build -t communication-react
for d in $(ls ../stashed); do diff -qr ../stashed/$d packages/$d/preprocessed; done
```


common/config/babel/.babelrc.js
```diff
@@ -7,12 +7,34 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
       match: '@conditional-compile-remove-from(stable)',
       // A list of features recognized by the conditional compilation preprocessing plugin.
       // "demo" is a special feature, used for demo purposes. For this feature,
-      // The plugin removes any AST node that is preceded by a comment that contains with the tag:
+      // The plugin removes any AST node that is preceded by a comment that contains the tag:
       // @conditional-compile-remove(demo)
       features: [
+        'call-with-chat-composite',
```
prprabhu-ms, line 13: Food for thought: Now that this is so obviously -- we need to reduce the number of beta-only features.
And also need to know what the criteria are for each feature to be stabilized.

prprabhu-ms: @alkwa-msft

anjulgarg:
Thanks for doing this. Seems like a painful PR :D

anjulgarg (approved):

common/config/babel/.babelrc.js
PorterNan: Agree, there are too many beta only features - as what we mentioned in our chat channel - some of them could be directly developed in main branch without conditional compilation, some could be stabilized already
Conditional compilation is a powerful tool, but overuse it will be a disaster

PorterNan: My previous thoughts about what should be conditional compiled:
https://teams.microsoft.com/l/message/19:5e150d250f724d1089c949305efdc8cc@thread.skype/1644444493297?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e9c1fc3-39df-4486-a26a-456d80e80f82&parentMessageId=1644275183011&teamName=Azure%20Communication%20Services&channelName=Web%20UI%20-%20Team&createdTime=1644444493297

common/config/babel/.babelrc.js
```diff
@@ -7,12 +7,34 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
...
+        // Flag to add API only available in beta calling SDK to mocks and internal types.
+        // This feature should be stabilized whenever calling SDK is stabilized.
+        'calling-1.4.2-beta.1',
+        // Participant pane in the `ChatComposite`.
+        'chat-composite-participant-pane',
```
PorterNan, line 18: I think it's time for us to stabilize participant-pane in chat as well

I like this list - it reminds us which features are still remain to be stabilized - otherwise it is hard to track what is currently conditional compiled

PorterNan (approved):

PR #1547, opened 2022-02-25
[DO NOT REVIEW] stabilize file sharing

prprabhu-ms:
# What
Example PR for how easy it is to stabilize a feature now!

### DIFF
--- a/common/config/api-extractor/api-extractor.stable.json
+++ b/common/config/api-extractor/api-extractor.stable.json
@@ -28,7 +28,7 @@
         "addToApiReportFile": false
       },
       "ae-incompatible-release-tags": {
-        "logLevel": "error",
+        "logLevel": "warning",
         "addToApiReportFile": false
       }
     },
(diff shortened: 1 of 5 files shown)


common/config/api-extractor/api-extractor.stable.json
```diff
@@ -28,7 +28,7 @@
         "addToApiReportFile": false
       },
       "ae-incompatible-release-tags": {
-        "logLevel": "error",
+        "logLevel": "warning",
```
prprabhu-ms, line 31: This is an annoyance now :(

common/config/babel/.babelrc.js
```diff
@@ -43,6 +41,8 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         // Demo feature. Used in live-documentation of conditional compilation.
         // Do not use in production code.
         'stabilizedDemo',
+        // Ability to upload/download files in message thread.
+        'file-sharing',
```
prprabhu-ms, line 45: One line stabilization!

prprabhu-ms:
Just an example.

PR #1549, opened 2022-02-25
Document feature stabilization

prprabhu-ms:


docs/references/beta-only-features.md
```diff
@@ -79,8 +79,15 @@ feature.
 
 ## Stabilizing a feature
 
-    TODO: How to move a feature from beta-only to building unconditionally.
+Careful scoping of [defined features](../../common/config/babel/.babelrc.js) will allow you to easily stabilize the feature. When you are ready to add your feature to the stable build:
 
+- One-step stabilization: Include all associated conditionally compiled code in the stable flavored build by moving your feature to `stabilizedFeature` in the [defined features config file]((../../common/config/babel/.babelrc.js)).
+  - Switch to stable flavor and build, test and run samples to try out your stabilized feature.
+  - You will also get the updated API files for the stable build from this step. You can use this for API review.
+- Clean up: Once your feature is shipped in a stable release, remove all references to conditional compilation directive for your feature, and remove your feature from `stabilizedFeature` list.
```
anjulgarg, line 87: Why do this separately after the build has been shipped?
If this is done originally, there would be no need for adding it to stabilizedFeature

If this needs to be done anyway, I don't see the effectiveness of stabilizedFeature :P

anjulgarg: Also if `stabilizedFeature` is a list, shouldn't it be `stabilizedFeatures` :D

docs/references/beta-only-features.md
```diff
@@ -79,8 +79,15 @@ feature.
...
+  - Note: If some code has other conditional compilation directives along with your (now stable) feature, remove all directives (not only yours), so that the code is included in the stable flavor build unconditionally.
```
anjulgarg, line 88: How would this be solved by `stabilizedFeature` 
Is it intelligent enough to work around this?

For example

// @conditional-remove(bla bla)
function abc() {
  // @conditional-remove(cha cha)
  doSomethingExtra();
}

docs/references/beta-only-features.md
prprabhu-ms: The value of `stabilizedFeatures` is that it gives us a single flag flip to stabilize (and make beta-only if need be) a feature.

The cleanup will take longer, can be done in smaller chunks, and would not risk destabilizing a release.
When we clean up, it will be more than just removing the conditional compilation directives. Various `*trampoline` functions we've created should be deleted as they are no longer needed etc.

This separates the (simple, but release critical) act of releasing a feature from (more complex, but no longer on the hot-path) cleaning up the conditional compilation related logic.

docs/references/beta-only-features.md
prprabhu-ms: Oh, perhaps I should reword this. I meant same AST node with different features. e.g.:

```ts
export interface StableArgs {

  // @conditional-compile-remove(some-feature-that-needs-options)
  // @conditional-compile-remove(another-feature-that-needs-options)
  options: Options;
}

// @conditional-compile-remove(some-feature-that-needs-options)
// @conditional-compile-remove(another-feature-that-needs-options)
export interface Options {
  // @conditional-compile-remove(some-feature-that-needs-options)
  myFancyOption: boolean;
  // @conditional-compile-remove(another-feature-that-needs-options)
  yourFancyOption: boolean;
}
```

In this example, either feature can be stabilized first. In each case, `Options` will be included in the stable build, but only one of the two fields will be included (depending on which feature is being stabilized).

Perhaps I should add this example to the doc.

prprabhu-ms: Added to live example.

docs/references/beta-only-features.md
```diff
@@ -79,8 +79,15 @@ feature.
 
 ## Stabilizing a feature
 
-    TODO: How to move a feature from beta-only to building unconditionally.
+Careful scoping of [defined features](../../common/config/babel/.babelrc.js) will allow you to easily stabilize the feature. When you are ready to add your feature to the stable build:
 
+- One-step stabilization: Include all associated conditionally compiled code in the stable flavored build by moving your feature to `stabilizedFeatures` in the [defined features config file]((../../common/config/babel/.babelrc.js)).
+  - Switch to stable flavor and build, test and run samples to try out your stabilized feature.
+  - You will also get the updated API files for the stable build from this step. You can use this for API review.
```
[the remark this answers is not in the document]
JamesBurnside: actually TMI if it's here -- is there another doc with what these API.json files are and where they live/how they're generated we can link to from here

JamesBurnside (approved):

docs/references/beta-only-features.md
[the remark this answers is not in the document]
prprabhu-ms: We have it, but it's internal: https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/27654/Scheduling-an-Azure-Review-Board-(ARB)-Review?anchor=creating-the-api-diff-view-(conditionally-compiled-features)

edwardlee-msft (approved):

PR #1557, opened 2022-02-28
Compliance Banner Colour Update

dmceachernmsft:
# What
Add styles for compliance banner

# Why
Increases visibility and follows Fluent colours for warnings in the compliance banner.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2669940
# How Tested

Ran locally to verify the correct colour was present.


packages/react-composites/src/composites/CallComposite/components/ComplianceBanner.tsx
```diff
@@ -98,6 +100,7 @@ export const ComplianceBanner = (props: ComplianceBannerProps): JSX.Element => {
     <></>
   ) : (
     <MessageBar
+      className={mergeStyles(complianceBannerWarningStyles(theme))}
```
JamesBurnside, line 103: Fluent's message bar has a `messageBarType={MessageBarType.Warning}` that sets color (though potentially other things) - can we use this instead?

JamesBurnside: https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar

alkwa-msft: The problem is the default colors they choose in those warnings can be similar to the video gallery background which is why we want to align with our fluent theme palette instead

dmceachernmsft: using the semanticColours here does guarantee that it matches Alex's designs. but I can look into what happens when its set to type warning as well, is the concern that when in the dark theme the warning colour could be different in this case?

dmceachernmsft: Looking into this both Light and dark themes with James' suggestion do result in the correct colour. do we want to go with that? or should we stick with the new style added?

JamesBurnside: Better to go with the inbuilt prop that custom overrides (it may also have non-obvious bonuses like a11y related props set)

JamesBurnside (approved):

change/@internal-react-composites-d35c0135-d7f6-4b55-a505-beec7001a31c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
JamesBurnside, line 2: This is type patch as this impacts the stable call composite

mgamis-msft (approved):

PR #1560, opened 2022-02-28
Prprabhu/rework icons

prprabhu-ms:
* Rework icons to make the per-composite icon types explicit.
* Add type assertions to continue to ensure that per-composite icons are consistent with exported default icons.
* Add beta-guards for use of beta-only icons from public icon types.

### DIFF
--- a/packages/communication-react/stable-review/communication-react.api.md
+++ b/packages/communication-react/stable-review/communication-react.api.md
@@ -894,20 +894,7 @@ export const COMPOSITE_LOCALE_ZH_CN: CompositeLocale;
 export const COMPOSITE_LOCALE_ZH_TW: CompositeLocale;
 
 // @public
-export const COMPOSITE_ONLY_ICONS: {
-    LobbyScreenConnectingToCall: JSX.Element;
-    LobbyScreenWaitingToBeAdmitted: JSX.Element;
-    LocalDeviceSettingsCamera: JSX.Element;
-    LocalDeviceSettingsMic: JSX.Element;
-    LocalDeviceSettingsSpeaker: JSX.Element;
-    LocalPreviewPlaceholder: JSX.Element;
-    Muted: JSX.Element;
-    NetworkReconnectIcon: JSX.Element;
-    NoticePageAccessDeniedTeamsMeeting: JSX.Element;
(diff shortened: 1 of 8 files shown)


packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -58,3 +58,10 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
     </FileCardGroup>
   );
 };
+
+const DownloadIconTrampoline = (): JSX.Element => {
```
prprabhu-ms, line 62: I had trouble writing utility code to avoid these trampoline functions at point-of-use.

This works for now... and follows the principle of smallest conditionally compiled code.

JamesBurnside (approved):

packages/communication-react/stable-review/communication-react.api.md
```diff
@@ -839,23 +894,10 @@ export const COMPOSITE_LOCALE_ZH_CN: CompositeLocale;
...
-    NoticePageAccessDeniedTeamsMeeting: JSX.Element;
-    NoticePageJoinCallFailedDueToNoNetwork: JSX.Element;
-    NoticePageLeftCall: JSX.Element;
-    NoticePageRemovedFromCall: JSX.Element;
-};
+export const COMPOSITE_ONLY_ICONS: CompositeIcons;
 
 // @public
-export type CompositeIcons = ComponentIcons & Record<keyof typeof COMPOSITE_ONLY_ICONS, JSX.Element>;
+export type CompositeIcons = ChatCompositeIcons & CallCompositeIcons;
```
JamesBurnside, line 900: We have to for not breaking API - but is `CompositeIcons` used for much anymore?

prprabhu-ms: Yeah, only here for API. It's not useful for Contoso or us at this point...

prprabhu-ms: Retrospective note: This type ended up being useful after all: I now typeguard `COMPOSITE_ONLY_ICONS` with this. It's still a very loose type, but it still catches the very common error that somebody adds a new value to that constant without updating any of the icon types.

PR #1562, opened 2022-03-01
Participant button custom styles story fixes

dmceachernmsft:
# What
Fix style choices on sample participant buttons in storybook page

# Why

provide examples with cleaner styling for storybook page. 

https://skype.visualstudio.com/SPOOL/_workitems/edit/2667104
# How Tested

ran locally to verify fixes.


packages/storybook/stories/ControlBar/Buttons/Participants/styles/ParticipantStyles.tsx
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { ParticipantsButtonStyles } from '@azure/communication-react';
+
+/**
+ * styles to make sure that the border on the participant menu isnt overriden buy the csutom styles example
```
JamesBurnside, line 7: nit: buy -> by
csutom -> custom

packages/storybook/stories/ControlBar/Buttons/Participants/snippets/WithCustomRender.snippet.tsx
```diff
@@ -73,11 +74,12 @@ export const ParticipantsButtonWithCustomRenderExample: () => JSX.Element = () =
...
-      <Label key={'participantsCustomLabelKey'} style={{ color: 'darkviolet', fontStyle: 'italic' }}>
+      <Label key={'participantsCustomLabelKey'} style={{ color: 'darkviolet', fontStyle: 'italic', padding: '0rem' }}>
         who is on this call
       </Label>
     );
   };
+
   return (
     <FluentThemeProvider>
       <ParticipantsButton
```
JamesBurnside, line 85: Can we also remove this scroll bar? (not a regression in this PR)
image

packages/storybook/stories/ControlBar/Buttons/Participants/styles/ParticipantStyles.tsx
JamesBurnside, line 7: Is there an alternative way to fix the red border issue? If you hit "show code" on the button example this makes the snippet unable to copy paste (as they don't have this styles as an import), it also makes the snippet a little more confusing:
image

dmceachernmsft:
Alex Pereira is going to do some design work around the custom styling for the participant button so hold off on approving this 😊

packages/storybook/stories/ControlBar/Buttons/Participants/styles/ParticipantStyles.tsx
dmceachernmsft: I will keep digging into this, but near as I can tell when the custom styles version of the button is clicked it add the red border to all of the buttons. The red border is not there until after the custom styles button is selected, until that point all other participant lists render correctly.

PR #1565, opened 2022-03-01
Mobile pane option to disable back button

mgamis-msft:
# What

Added mobile pane option to disable button in CallWithChat composite

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2590659

# How Tested

Tested on CallWithChat sample.
https://user-images.githubusercontent.com/79475487/156122640-42a13d07-17a5-4d78-a258-1e9d61f2283e.mp4
Notice that when the back button is clicked that the threadId argument in the url disappears and immediately reappears. 
It is not the cleanest solution but it is the simplest without pushing state to the browser history.

# Process & policy checklist

### DIFF
--- a/packages/communication-react/review/communication-react.api.md
+++ b/packages/communication-react/review/communication-react.api.md
@@ -711,6 +711,7 @@ export type CallWithChatCompositeIcons = {
 // @beta
 export type CallWithChatCompositeOptions = {
     callControls?: boolean | CallWithChatControlOptions;
+    mobilePane?: MobilePaneOptions;
 };
 
 // @beta
@@ -1948,6 +1949,12 @@ export interface MicrophoneButtonStrings {
 // @public
 export interface MicrophoneButtonStyles extends ControlBarButtonStyles {
     menuStyles?: Partial<MicrophoneButtonContextualMenuStyles>;
+}
+
+// @beta
+export interface MobilePaneOptions {
+    // (undocumented)
+    disabledBackButton?: boolean;
 }
 
 // @public
--- a/packages/react-composites/review/react-composites.api.md
+++ b/packages/react-composites/review/react-composites.api.md
@@ -511,6 +511,7 @@ export type CallWithChatCompositeIcons = {
 // @beta
 export type CallWithChatCompositeOptions = {
     callControls?: boolean | CallWithChatControlOptions;
+    mobilePane?: MobilePaneOptions;
 };
 
 // @beta
@@ -927,6 +928,12 @@ export type MessageReceivedListener = (event: {
 
 // @public
 export type MessageSentListener = MessageReceivedListener;
+
+// @beta
+export interface MobilePaneOptions {
+    // (undocumented)
+    disabledBackButton?: boolean;
+}
 
 // @public
 export type NetworkDiagnosticChangedEvent = NetworkDiagnosticChangedEventArgs & {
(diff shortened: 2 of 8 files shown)


JamesBurnside:
Implementation looks great! But lets bring this PR up in the feature weekly meeting to chat through it and confirm this is the direction we want to take with this feature

packages/react-composites/src/composites/CallWithChatComposite/MobilePane.tsx
```diff
@@ -50,6 +51,19 @@ export const MobilePane = (props: {
   }, [theme]);
   const strings = useCallWithChatCompositeStrings();
 
+  useEffect(() => {
+    // disable back button when boolean is true and mobile pane is not hidden
+    if (props.disableBackButton && !props.hidden) {
+      window.onpopstate = function () {
+        window.history.forward();
```
JamesBurnside, line 58: Oh so this is allowing the back behavior to be triggered, but before it really happens it triggers a history.forward() to in essence ignore the back button. The alternative to this (to maybe avoid the disappearing&reappearing url parms) would be when the panel is opened do window.history.pushState... but that has more cleanup and maintainence of the history state to do so I like this implementation (but hard to predict if this will work with all customer websites)

packages/react-composites/review/react-composites.api.md
```diff
@@ -928,6 +929,12 @@ export type MessageReceivedListener = (event: {
 // @public
 export type MessageSentListener = MessageReceivedListener;
 
+// @beta
+export interface MobilePaneOptions {
+    // (undocumented)
+    disabledBackButton?: boolean;
```
JamesBurnside, line 935: avoid boolean in the API interface as it becomes impossible to extend at a future date. Consider something like: 
`browserBackButtonOverride?: 'none' | 'closePane'`

packages/communication-react/review/communication-react.api.md
```diff
@@ -711,6 +711,7 @@ export type CallWithChatCompositeIcons = {
 // @beta
 export type CallWithChatCompositeOptions = {
     callControls?: boolean | CallWithChatControlOptions;
+    mobilePane?: MobilePaneOptions;
```
alkwa-msft, line 714: I am curious now that we have an adaptive UI (e.g we different mobile and non mobile modes) should we just separate out options specific for mobile?

e.g

{
  ...
  mobile : {
      ...mobileSpecificOptions
   }
}

packages/react-composites/src/composites/CallWithChatComposite/MobilePane.tsx
mgamis-msft: Yes, pushing state once something like:
```
  useEffect(() => {
    window.history.pushState(null, document.title, location.href);
  }, []);
```
does avoid disappearing&reappearing url parms. Thanks.

This will be more robust for Contoso when they back up as it avoids that brief loading of the previous url which could be anything given that they can embed the CallWithChatComposite on their web page.

However, I could not avoid situations in testing where I was doing multiple pushes to state even or using the forward button to mess up the state expectations. Will continue find a good solution today.

packages/communication-react/review/communication-react.api.md
JamesBurnside: given isMobile is an existing prop we should actually probably avoid anything `mobile*` here to avoid prop overlap

Instead one idea here would be to make this something like: 
```
browserOverrides: {
  closeSidePaneOnBackButton: boolean;
}
```

then a developer would do:
```
options={{
  browserOverrides: {
    closeSidePaneOnBackButton: isUserOnMobile() ? true : false
  }
}}
```

It would mean we have to set as `false` by default for mobile (as desktop should likely always be false) so less out-of-box set up

PR #1569, opened 2022-03-01
Remove file upload button for beta release

JamesBurnside:
# What
Remove file upload button from composites containing chat for beta release

# Why
Feature is not ready yet

# How Tested
Ran locally to verify no file upload button in chat:
image


packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -183,11 +185,12 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
               styles={sendBoxStyles}
             />
 
-            <FileUploadButton
+            {/* REMOVE THE FILE SHARING BUTTON FOR 1.1.1-BETA.1 RELEASE -- DO NOT COMMIT THIS TO THE MAIN BRANCH */}
```
prprabhu-ms, line 188: I would have tried to make this even smaller.

```ts
const willNeverBeTrue = process.env['DO_NOT_SET_THIS'] === 'True';

...

{ willNeverBeTrue && <FileUploadButton ...>}
```

Then you don't need to remove any imports etc.

prprabhu-ms (approved):

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
JamesBurnside: Oh yah much simpler, updated with a const `const SHOW_FILE_SHARING_BUTTON = false;`

prprabhu-ms: Heh, I have a healthy respect for compiler optimizations. So if I want the compiler to keep some code, I tie it to something unguessable like environment variable :D

PR #1573, opened 2022-03-01
Prevent chat message flyout randomly closing itself

JamesBurnside:
# What
* Set `preventDismissOnResize` property on the chat message flyout (same property we set on the other flyouts we have)

# Why
Chat message flyout was randomly closing itself when certain things happened on the video grid (such as participants joining/leaving, resize events, 'Your Muted' UFD)

# How Tested
Locally, before change: easy to repro when "Your muted" ufd was shown
After change: Could never repro (tests: other participants joining and leaving, other participants turning off and on local video, Your muted UFD appearing)


dmceachernmsft (approved):

prprabhu-ms (approved):

prprabhu-ms:
...and thus a hack proliferates...

PR #1574, opened 2022-03-01
Show loader in File card until file download promise is resolved

prabhjot-msft:
# What
Show loader in File card until file download promise is resolved

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2771033

# How Tested
Manually
Screen Shot 2022-03-01 at 12 34 28 PM

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -39,7 +42,21 @@ export interface FileDownloadCards {
 export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
   const truncateLength = 15;
   const { userId, message } = props;
+  // @conditional-compile-remove(file-sharing)
```
anjulgarg, line 45: Why do we need to conditionally remove anything in this component?

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -39,7 +42,21 @@ export interface FileDownloadCards {
...
+  const [showSpinner, setSpinner] = useState(false);
```
anjulgarg, line 46: setSpinner -> setShowSpinner

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -39,7 +42,21 @@ export interface FileDownloadCards {
...
+    // @conditional-compile-remove(file-sharing)
+    if (showSpinner) {
+      // @conditional-compile-remove(file-sharing)
+      return <Spinner size={SpinnerSize.medium} aria-live={'assertive'} />;
+    } else {
+      // @conditional-compile-remove(file-sharing)
+      return <ChatCompositeIcon iconName="Download" />;
+    }
+    // Return _some_ available icon, as the real icon is beta-only.
+    return <ChatCompositeIcon iconName="EditBoxCancel" />;
```
anjulgarg, line 58: No need. Nothing should be conditional in this function

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -50,18 +67,27 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
             fileExtension={extension(file.name)}
             actionIcon={<DownloadIconTrampoline />}
             actionHandler={() => {
-              props.downloadHandler && props.downloadHandler(userId, file);
+              if (props.downloadHandler) {
+                // @conditional-compile-remove(file-sharing)
+                setSpinner(true);
+                props
+                  .downloadHandler(userId, file)
```
anjulgarg, line 74: Please take a look at the downloadHandler function type. We are not expecting it to throw errors. The behavior of UI should depend on it's return type.

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -50,18 +67,27 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
...
+                  .then((fileurl) => {
+                    // @conditional-compile-remove(file-sharing)
+                    setSpinner(false);
+                    window.open(fileurl.toString(), '_blank', 'noopener,noreferrer');
+                  })
+                  .catch((error) => {
+                    // @conditional-compile-remove(file-sharing)
+                    setSpinner(false);
```
anjulgarg, line 82: You can use a finally block to stop the spinner instead of multiple declarations :)

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -39,7 +42,21 @@ export interface FileDownloadCards {
 export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
   const truncateLength = 15;
   const { userId, message } = props;
+  // @conditional-compile-remove(file-sharing)
+  const [showSpinner, setSpinner] = useState(false);
   const fileDownloads: FileMetadata[] = message.metadata ? extractFileMetadata(message.metadata) : [];
+  const DownloadIconTrampoline = (): JSX.Element => {
```
anjulgarg, line 48: This component can be declared outside the `FileDownloadCards` component to clean up the code a bit.

anjulgarg: However, I think this component is not needed. An inline logic looks much simpler for actionIcon
`actionIcon={showSpinner ? <Spinner size={SpinnerSize.medium} aria-live={'assertive'} /> : <ChatCompositeIcon iconName="Download" /> }`

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -48,20 +52,34 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
...
+                <ChatCompositeIcon iconName="Download" />
+              )
+            }
             actionHandler={() => {
-              props.downloadHandler && props.downloadHandler(userId, file);
+              if (props.downloadHandler) {
+                setShowSpinner(true);
+                props
+                  .downloadHandler(userId, file)
+                  .then((value: URL | FileDownloadErrorMessage) => {
```
JamesBurnside, line 67: `.then` is the old style promise syntax, consider using await syntax:
```ts
setShowSpinner(true);
const url = await props.downloadHandler(userId, file);
if (value instanceof URL) {
  window.open(value.toString(), '_blank', 'noopener,noreferrer');
}
setShowSpinner(false);

File: packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx [hunk 8]
```diff
@@ -48,20 +52,34 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
...
+                    if (value instanceof URL) {
+                      window.open(value.toString(), '_blank', 'noopener,noreferrer');
```
JamesBurnside, line 69: `.toString()` is unecessary here, window.open should accept a URL type

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -48,20 +52,34 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
...
+                    } else {
+                      //TODO: implement error handling for reject
+                      console.log(value);
```
JamesBurnside, line 72: nit: remove the console log (I understand its temporary until the todo is done but for PRs they should still be removed -- typically because todo's can sometimes take a long time/forgot to be done and users my have weird messages in the console)

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -48,20 +52,34 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
...
+                    }
+                  })
+                  .finally(() => {
+                    setShowSpinner(false);
+                  });
+              }
               !props.downloadHandler && window.open(file.url, '_blank', 'noopener,noreferrer');
```
JamesBurnside, line 79: nit: update to `else { ` here to make clearer the logical pivot on props.downloadHandler

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -48,20 +52,34 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
...
+            actionIcon={
+              showSpinner ? (
+                <Spinner size={SpinnerSize.medium} aria-live={'assertive'} />
+              ) : (
+                <ChatCompositeIcon iconName="Download" />
+              )
+            }
             actionHandler={() => {
-              props.downloadHandler && props.downloadHandler(userId, file);
+              if (props.downloadHandler) {
```
JamesBurnside, line 63: This has become a long inline function, consider making this a standalone function inside the FileDownloadCards:
```
const fileDownloadHandler = useCallback(() => {
  // this big logic block
```

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -48,20 +52,34 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
             fileName={truncatedFileName(file.name, truncateLength)}
             key={file.name}
             fileExtension={extension(file.name)}
-            actionIcon={<DownloadIconTrampoline />}
+            actionIcon={
+              showSpinner ? (
+                <Spinner size={SpinnerSize.medium} aria-live={'assertive'} />
+              ) : (
+                <ChatCompositeIcon iconName="Download" />
```
JamesBurnside, line 59: Use DownloadIconTrampoline here (that was removed) as I assume stable build will fail without it

JamesBurnside (approved):
Left some code style comments, please take a look at them. Otherwise this looks great!

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: Done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: Done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: Done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: Done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
anjulgarg: @JamesBurnside I don't understand why stable build will fail without i?

JamesBurnside: `Download` is a beta-only icon and icon names are now typeguarded

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: Done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: Done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -39,7 +41,23 @@ export interface FileDownloadCards {
 export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
   const truncateLength = 15;
   const { userId, message } = props;
+  const [showSpinner, setShowSpinner] = useState(false);
   const fileDownloads: FileMetadata[] = message.metadata ? extractFileMetadata(message.metadata) : [];
+  const fileDownloadHandler = useCallback(
+    async (userId, file) => {
+      if (props.downloadHandler) {
+        setShowSpinner(true);
+        const url = await props.downloadHandler(userId, file);
```
anjulgarg, line 50: this can return a url or an error.
I suggest to rename this const as `response` or something.

anjulgarg, line 50: What will happen if props.downloadHandler throws an error?
I think it should be inside a try catch finally block.

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -39,7 +41,23 @@ export interface FileDownloadCards {
...
+        if (url instanceof URL) {
+          window.open(url.toString(), '_blank', 'noopener,noreferrer');
+        }
+        setShowSpinner(false);
+      } else {
+        window.open(file.url, '_blank', 'noopener,noreferrer');
```
anjulgarg, line 56: I recommend adding an if check at the top of this function to make the code easier to read.
```tsx
if (!props.downloadHandler) {
  return window.open(file.url, '_blank', 'noopener,noreferrer');
}
```

the rest of the code can then be written outside the conditional block.

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -39,7 +41,23 @@ export interface FileDownloadCards {
...
   const fileDownloads: FileMetadata[] = message.metadata ? extractFileMetadata(message.metadata) : [];
+  const fileDownloadHandler = useCallback(
+    async (userId, file) => {
+      if (props.downloadHandler) {
+        setShowSpinner(true);
+        const url = await props.downloadHandler(userId, file);
+        if (url instanceof URL) {
+          window.open(url.toString(), '_blank', 'noopener,noreferrer');
+        }
+        setShowSpinner(false);
```
anjulgarg, line 54: `setShowSpinner(false);` can be called immediately after `await props.downloadHandler(userId, file);`

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: done

prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -39,7 +41,29 @@ export interface FileDownloadCards {
...
   const fileDownloads: FileMetadata[] = message.metadata ? extractFileMetadata(message.metadata) : [];
+  const fileDownloadHandler = useCallback(
+    async (userId, file) => {
+      if (!props.downloadHandler) {
+        window.open(file.url, '_blank', 'noopener,noreferrer');
+      } else {
+        setShowSpinner(true);
+        try {
+          const response = await props.downloadHandler(userId, file);
+          setShowSpinner(false);
```
prprabhu-ms, line 54: This is not needed because of line 61.

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -39,7 +41,29 @@ export interface FileDownloadCards {
...
+          if (response instanceof URL) {
+            window.open(response.toString(), '_blank', 'noopener,noreferrer');
+          }
+        } catch (error) {
+          console.error(error);
```
prprabhu-ms, line 59: Shouldn't we show an error on the UI as well for the end-user?

prprabhu-ms (approved):

PR #1575, opened 2022-03-01
Do not allow sending empty messages when editing chat messages

JamesBurnside:
# What
* Check is an editted message is empty and disable the onSubmit if it is
* Perform a trim() on the message when doing this check

# Why
You are able to send empty messages currently by editting them:
image

# How Tested
Tested can no longer submit edits that are deemed empty (or too long), but still can submit OK messages using enter key and via the submit button


packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
```diff
@@ -41,15 +41,18 @@ export type ChatMessageComponentAsEditBoxProps = {
   inlineEditButtons: boolean;
 };
 
+type MessageState = 'OK' | 'too short' | 'too long';
```
prprabhu-ms, line 44: It messes with my brain to see a space in a string literal type. There's nothing wrong with it, and this is internal only.

But... it messes with me :P

prprabhu-ms (approved):

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
JamesBurnside: hehe

PR #1578, opened 2022-03-01
[Filesharing] - Expire Upload Errors after 10 seconds OR when more file are uploaded.

anjulgarg:
# What
- Add new internal component `SendBoxErrorBar` for displaying SendBox errors with timetout
- Remove a file upload error message above sendbox after 10 seconds.
- Adding more files to the sendbox will remove existing erroneous file uploads. And thus removing an error message being displayed above sendbox
- Adds tests for AzureCommunicationFileUploadAdapter

# Why
Part of filesharing error handling

# How Tested

# Process & policy checklist


packages/react-components/src/components/SendBoxErrorBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * @internal
+ */
+export interface SendBoxErrorBarProps {
+  /** Error message to render */
+  message?: string;
+  /**
+   * Timeout in ms after which the error bar disappears.
+   * If undefined, the error bar will never disappear.
+   */
+  timeout?: number;
```
JamesBurnside, line 17: Why do we remove the error message after a timeout? Do we do this to any other error messages?

packages/react-components/src/components/SendBoxErrorBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+  useEffect(() => {
+    if (timeout !== undefined) {
+      const messageTimeout = setTimeout(() => {
+        setErrorMessage(undefined);
+      }, timeout);
+      return () => {
+        clearTimeout(messageTimeout);
+      };
+    } else {
+      return;
```
JamesBurnside, line 36: nit: remove empty else

JamesBurnside (approved):

packages/react-components/src/components/SendBoxErrorBar.tsx
anjulgarg: The reason is a mix of product and design requirements and mimics the file upload error behavior in teams.

packages/react-components/src/components/SendBoxErrorBar.tsx
anjulgarg: Added it to remove the warning, `not all code paths return`

prabhjot-msft (approved):

PR #1579, opened 2022-03-01
Update Docs for Local Camera Switcher

dmceachernmsft:
# What
Update story book docs about local camera switcher doc. Introduce Storybook functionality so that we can have warning banners with features. 

# Why

Lets users know about the effects of the local camera switcher and whether it can be dragged with a warning: 
image

# How Tested

inspected docs locally to verify banner shows properly.


packages/storybook/stories/BetaBanners/StoryBookBanner.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <Stack className={bannerContainerStyles(palette.background)}>
+      <Stack.Item>
+        {props.children && (
+          <Stack style={{ display: 'inline-block' }}>
+            <ImportantHeading color={palette.header} />
+            {' - '}
+            {props.children}
+          </Stack>
+        )}
+        {!props.children && (
```
JamesBurnside, line 31: This shouldn't pivot off prop.children, but be it's own component that uses StorybookBanner, i.e. something like
```
const SingleLineBetaBanner () => (
  <StoryBookBanner>
    <>
      <Text>This feature is currently in public preview and not recommended for production use. </Text>
      <Link  ... >
    </>
  </StorybookBanner>
);
```

packages/storybook/stories/BetaBanners/StoryBookBanner.tsx
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @private
+ */
+export interface StoryBookBannerProps {
+  palette?: BannerPalette;
+  children?: JSX.Element;
+}
+
+/** @private */
+export const StoryBookBanner = (props: StoryBookBannerProps): JSX.Element => {
```
JamesBurnside, line 19: nit: StoryBook -> Storybook (+props and filename)

packages/storybook/stories/VideoGallery/VideoGallery.stories.tsx
```diff
@@ -129,6 +131,15 @@ const getDocs: () => JSX.Element = () => {
         button will cycle through the different camera's in the users device. Typical usage is to enable this button on
         mobile devices. This button is enabled through the use of the `showCameraSwitcherInLocalPreview` prop.
       </Description>
+      <StoryBookBanner
+        // eslint-disable-next-line react/no-children-prop
+        palette={yellowBannerPalette}
+      >
+        <p style={{ display: 'inline-block' }}>
```
JamesBurnside, line 138: use fluent components over raw html elements
```suggestion
        <Text>
```

packages/storybook/stories/BetaBanners/StoryBookBanner.tsx
dmceachernmsft: ah okay I see, will make this change!

JamesBurnside (approved):

dmceachernmsft:
> image
> 
> The icons look like they are missing in the newer version

I believe this is from a issue when it was updating the snapshots, will run the jobs again. Nothing I wrote for this touches this..

dmceachernmsft:
> > image
> > The icons look like they are missing in the newer version
> 
> I believe this is from a issue when it was updating the snapshots, will run the jobs again. Nothing I wrote for this touches this so nothing should have changed. I think we have seen this before in past PR's lately

yeah looks like is a CI thing, will watch the snapshots before I merge 
image


> > image
> > The icons look like they are missing in the newer version
> 
> I believe this is from a issue when it was updating the snapshots, will run the jobs again. Nothing I wrote for this touches this so nothing should have changed. I think we have seen this before in past PR's lately

yeah looks like this is a CI issue, will watch snapshots before I merge
image

packages/storybook/stories/VideoGallery/VideoGallery.stories.tsx
```diff
@@ -129,6 +132,14 @@ const getDocs: () => JSX.Element = () => {
         button will cycle through the different camera's in the users device. Typical usage is to enable this button on
         mobile devices. This button is enabled through the use of the `showCameraSwitcherInLocalPreview` prop.
       </Description>
+      <StorybookBanner
+        // eslint-disable-next-line react/no-children-prop
```
prprabhu-ms, line 136: Why do you need this? I don't see any explicit `children` prop here.

prprabhu-ms (approved):

packages/storybook/stories/VideoGallery/VideoGallery.stories.tsx
dmceachernmsft: oops I think I forgot to remove that!

dmceachernmsft: fixed!

alkwa-msft (approved):

PR #1582, opened 2022-03-02
Fix styling when unread chat messages exceeds 9 (and remove ControlBarButtonBadgeIcon as a customizable icon)

JamesBurnside:
# What
* Remove `ControlBarButtonBadgeIcon` as a customizable icon - this is just a circular background not a custom icon
* Update to use CSS as badge background to fix styling when chat messages exceed 9
* Increase the distance between control bar buttons text and icon in split button designs to match figma (10px)
  * This gives enough distance for the chat notification to display `9+` without occluding the `Chat` text

# Why
Not ideal UIwhen `9+` was showing:
image

New Figma designs: https://www.figma.com/file/JVOzw64OnLDrNjb9AScLN3/Web-UI-Library---Explorations?node-id=1275%3A43068

Bug link: https://skype.visualstudio.com/SPOOL/_workitems/edit/2777078

# How Tested
Locally running call with chat

Zoomed in:
image
image

Zoomed out:
image
image


dmceachernmsft (approved):

packages/react-composites/src/composites/CallWithChatComposite/NotificationIcon.tsx
```diff
@@ -32,39 +31,37 @@ export const NotificationIcon = (props: NotificationIconProps): JSX.Element => {
...
   );
 };
 
+const notificationIconPaddingREM = 0.225;
+
 const notificationIconContainerStyles = memoizeFunction(
-  (): IIconStyles => ({
+  (theme: ITheme): IIconStyles => ({
     root: {
+      borderRadius: '1rem',
```
anjulgarg, line 45: Please use `theme.effects.roundedCorner` for this.

anjulgarg:
@JamesBurnside Is the change in control bar button width and border radius intentional?
I can see some UI snapshots have changed.

JamesBurnside:
> @JamesBurnside James Burnside FTE Is the change in control bar button width and border radius intentional? I can see some UI snapshots have changed.

Button width yes, border radius might just look different because it's longer? It shouldn't have changed

packages/communication-react/review/communication-react.api.md
```diff
@@ -1396,7 +1395,6 @@ export const DEFAULT_COMPOSITE_ICONS: {
     VideoTileMicOff: JSX.Element;
     LocalCameraSwitch?: JSX.Element | undefined;
     ChevronLeft?: JSX.Element | undefined;
-    ControlBarButtonBadgeIcon?: JSX.Element | undefined;
```
PorterNan, line 1399: Nice, happy to see we save one api slot and implement it in css!

PorterNan (approved):

packages/react-composites/src/composites/CallWithChatComposite/NotificationIcon.tsx
JamesBurnside: We explicitly use `1rem` here because we're not a UI theme-tied border radius, but instead CSS to make a circle, i.e. these should always be the curvature and not be tied to theme:
image

I'll add a comment here to make this more clear

JamesBurnside: Added some comments!

prprabhu-ms (approved):

PR #1583, opened 2022-03-02
Fix: Edit box should have the same focus/unfocus behavior as sendBox

JamesBurnside:
# What
Update edit box to use the same styles as send box

# Why
Currently EditBox only has blue borders, and it never changes, it should have the same behavior as the sendbox (be gray when not active)

# How Tested
Locally:
| active | inactive |
| -- | -- |
| image | image |


packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
```diff
@@ -90,7 +86,16 @@ export const ChatMessageComponentAsEditBox = (props: ChatMessageComponentAsEditB
   }, [theme.palette.themePrimary]);
 
   return (
-    <Stack className={mergeStyles(borderAndBoxShadowStyle(theme))}>
+    <Stack
+      className={mergeStyles(
+        borderAndBoxShadowStyle({
+          theme,
+          errorColor: isDarkThemed(theme) ? '#f1707b' : '#a80000',
```
alkwa-msft, line 93: we added more than just theme. In this case are we doing special overrides that should just be in theme?

packages/react-components/src/components/SendBox.tsx
```diff
@@ -218,7 +218,12 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
   return (
     <Stack
       className={mergeStyles(
-        borderAndBoxShadowStyle(theme, isDarkThemed(theme) ? '#f1707b' : '#a80000', !!errorMessage, !!disabled),
+        borderAndBoxShadowStyle({
+          theme,
+          errorColor: isDarkThemed(theme) ? '#f1707b' : '#a80000',
```
alkwa-msft, line 223: I am curious  it looks like errorColor we want to use fixed values and it seems to be consistent. Should we just make it somehow part of "theme" by default?

Also we are being very specific about the colors that are used. Is there a palette color we can align to instead?

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
JamesBurnside: I didn't want to change any of that logic in this PR, this just does the exact same logic that sendbox does to ensure consistency (maybe I should pull into into a component they both use..)
But ideally we should be using some theme-based semantic color that defines an error:
image

Filed a bug for this: [Bug 2779691](https://skype.visualstudio.com/SPOOL/_workitems/edit/2779691): [Web] Can sendbox error coloring use a theme value?

packages/react-components/src/components/SendBox.tsx
JamesBurnside: Yeah not a fan of these hard coded colors either, filed a bug: [Bug 2779691](https://skype.visualstudio.com/SPOOL/_workitems/edit/2779691): [Web] Can sendbox error coloring use a theme value?

prabhjot-msft (approved):

anjulgarg (approved):

PR #1588, opened 2022-03-03
Fix img size for style defined element

PorterNan:
# What

There is a case that img from iPhone client size wouldn't be limited because it is directly defined in element
This fix uses an important tag to override it

Before:
image

After(because we don't have the actual file, so it img is tented to be showed as a small icon instead of full size, which is correct):
image

Thanks @alkwa-msft for help in testing!
# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/styles/MessageThread.styles.ts
+++ b/packages/react-components/src/components/styles/MessageThread.styles.ts
@@ -107,8 +107,8 @@ export const defaultChatMessageContainer: ComponentSlotStyle = {
   minWidth: `${CHAT_MESSAGE_CONTAINER_MIN_WIDTH_REM}rem`,
   marginRight: '0rem',
   ' img': {
-    maxWidth: '100%',
-    height: 'auto'
+    maxWidth: '100% !important', // Add !important to make sure it won't be overridden by style defined in element
+    height: 'auto !important'
   },
   ' p': {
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -107,8 +107,8 @@ export const defaultChatMessageContainer: ComponentSlotStyle = {
   minWidth: `${CHAT_MESSAGE_CONTAINER_MIN_WIDTH_REM}rem`,
   marginRight: '0rem',
   ' img': {
```
JamesBurnside, line 109: (non-blocking for this PR) the `' img` and `' p'` here look to be causing these issues: [Bug 2777093](https://skype.visualstudio.com/SPOOL/_workitems/edit/2777093): [Web] logspam from fela is making it impossible to see any other logs

Is there a different css selector we can use that fela won't complain about?

JamesBurnside (approved):

packages/react-components/src/components/styles/MessageThread.styles.ts
PorterNan: Thanks for reminding this! Will take some experiments and take that item if I find the reason after this PR

prprabhu-ms (approved):

PR #1589, opened 2022-03-03
[CallWithChat][Bug] Mobile Participants now extend all the way to the right hand side in mobile pane

edwardlee-msft:
# What
Mobile participant items now extend all the way to the right hand side in mobile pane, taking full width.
 

# Why
Participant items only expanded up to 20rem, which is around 3/4 of the screen width.
https://skype.visualstudio.com/SPOOL/_workitems/edit/2777090
 

# How Tested
Tested locally with devtools. Tested on mobile android phone.
Test long names and short names.

# Process & policy checklist


packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -75,3 +72,16 @@ export const ParticipantListWithHeading = (props: {
...
 };
+
+const participantListDesktopStyle = mergeStyles({
+  maxWidth: '20rem',
+  height: '100%',
+  overflowY: 'auto',
+  overflowX: 'hidden'
+});
+const participantListMobileStyle = mergeStyles({
+  maxWidth: '100vw',
```
JamesBurnside, line 83: don't use `vw` as that is the entire viewport width. While our mobile composite is likely to fill the screen it's not gauranteed. So we need the width of this list to fill the container it is in (100%) over being sized to the viewport

change/@internal-react-components-8b40a12e-a568-41b7-85f8-0021a25489f6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Mobile Participant pane now spans full sidepane width",
```
JamesBurnside, line 3: nit: this actually impacts the ParticipantItem public component we export. Cna you gear this change file to be what changed in this package. e.g. `Removed 20rem max-width restriction on ParticipantList component. Max-width is now set to 100%`

change/@internal-react-composites-68cb0c3b-d3dc-49c3-81d5-14de3eb57b73.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
JamesBurnside, line 2: nit: for this package the change type is prerelease

JamesBurnside (changes requested):
remove use of vw

change/@internal-react-components-8b40a12e-a568-41b7-85f8-0021a25489f6.json
edwardlee-msft: I've tried different types of combinations, but without setting a width the participant item overflows off screen when names are very long.


Also found that ellipsis has a couple of restraints for it to work:
```
text-overflow:ellipsis; only works when the following are true:

The element's width must be constrained in px (pixels). Width in % (percentage) won't work.
The element must have overflow:hidden and white-space:nowrap set.
```

JamesBurnside: You should be able to use `max-width: 100%`, the problem with `max-width: 100%` is each parent element will also need this constraint - i.e. what is the 100%? if its 100% of a div that can be any width, then it can be any width. If you walk up the DOM tree and continue to give each parent element `max-width: 100%` I imagine this will eventually give you what you want.
Let me know if you want to pair on this!

edwardlee-msft: Thank you James!!
I put 100% max-width on all divs going up one by one by one and then noticed that one specific stack parent div was the key.

JamesBurnside (approved):

PR #1593, opened 2022-03-04
Show error above sendbox if message is sent before all files are uploaded.

anjulgarg:
# What
Show error above sendbox if message is sent before all files are uploaded.
FIGMA: https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?node-id=3970%3A22690

https://user-images.githubusercontent.com/9782747/156677880-d48ae30a-08a0-4639-89b4-224a1c25e17a.mp4

# Why
A user should not be allowed to send a message until all the files have been uploaded.

# How Tested

# Process & policy checklist


packages/react-components/src/components/SendBox.tsx
```diff
@@ -65,6 +65,11 @@ export interface ActiveFileUpload {
    * Error message to be displayed to the user if the upload fails.
    */
   errorMessage?: string;
+
+  /**
+   * `true` means that the upload is completed.
+   */
+  isUploaded?: boolean;
```
JamesBurnside, line 72: is isUploaded the same as progress having value of 1?

anjulgarg: Yes and metadata should be populated

packages/react-components/stable-review/react-components.api.md
```diff
@@ -37,6 +37,7 @@ export interface ActiveFileUpload {
     errorMessage?: string;
```
JamesBurnside, line 37: Should this whole interface be not included in the stable api build?

alkwa-msft:
> # What
> Show error above sendbox if message is sent before all files are uploaded. FIGMA: https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?node-id=3970%3A22690
> 
>  fileInProgressErrorDemo.mp4 
> # Why
> A user should not be allowed to send a message until all the files have been uploaded.
> 
> # How Tested
> # Process & policy checklist
> * [ ]  I have updated the project documentation to reflect my changes if necessary.
> * [ ]  I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.
> 
> **Is this a breaking change?**
> 
> * [ ]  This change causes current functionality to break.

I wish we had this connected to the whole experience for this demo video. When the error  pops up does it shift everything up?

alkwa-msft:
When the error pops up does it shift all of the content (e.g message thread) up?

packages/react-components/src/components/SendBoxErrorBar.tsx
```diff
@@ -23,15 +23,19 @@ export interface SendBoxErrorBarProps {
...
+    clearSetTimeout();
     if (timeout !== undefined) {
-      const messageTimeout = setTimeout(() => {
+      setTimeoutRef.current = setTimeout(() => {
         setErrorMessage(undefined);
       }, timeout);
-      return () => {
-        clearTimeout(messageTimeout);
-      };
+      return clearSetTimeout;
```
alkwa-msft, line 38: we called clearSetTimeout on line 33 which forces us to be in a clean state.

Then after a certain amount of time we set the errorMessage to be undefined and then we clean the setTimeout again?

I am curious do we need to clear it the first time on line 33?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileCard.tsx
```diff
@@ -49,7 +49,7 @@ export const FileCard = (props: FileCardProps): JSX.Element => {
   const { fileName, fileExtension, progress, actionIcon } = props;
   const theme = useTheme();
 
-  const showProgressIndicator = (): boolean => progress !== undefined && progress < 1;
+  const showProgressIndicator = (): boolean => progress !== undefined && progress < 1 && progress > 0;
```
alkwa-msft, line 52: nit.

Can we swap the order to be progress > 0 && progress < 1 so we can read it  left to right that we want to show progress from 0 -> 1?

packages/react-components/src/components/SendBoxErrorBar.tsx
```diff
@@ -23,15 +23,19 @@ export interface SendBoxErrorBarProps {
...
     } else {
       return;
```
alkwa-msft, line 40: if else is just a return do we need to even write line 39-41?

packages/react-components/src/components/SendBoxErrorBar.tsx
alkwa-msft: timeout as the name of the prop for the length in time is a hard to read when we are using functions called setTimeout and clearTimeout, the numerical value passed in for the length of time is often referred to as "delay".

We have also exposed technical details of what the `clearSetTimeout `is doing (in the name). Ultimately what is this solving and can we make it easier to understand? `clearTimer` or something like that?

Does it seem correct that we are creating a const function inside of the `useEffect `and could we do that outside instead?

packages/react-components/src/components/SendBoxErrorBar.tsx
anjulgarg: This is for preventing warning:
Not all code paths return

anjulgarg: But it's just a warning so I am gonna remove it.

packages/react-components/src/components/SendBoxErrorBar.tsx
anjulgarg: You can create a const function inside useEffect and use it.
If not, you need to create it outside as a `useCallback` function.
I still moved the function out to make the logic easier to read plus a bunch of naming changes and refactoring.
LMKWYT

packages/react-components/stable-review/react-components.api.md
anjulgarg: Removed it. :)

packages/react-components/src/components/SendBox.tsx
anjulgarg: We can't rely on just the upload progress, since it can be 1 without metadata being populated.
In our case, metadata being populated marks an upload as complete irrespective of it's upload progress

packages/react-components/src/components/SendBox.tsx
```diff
@@ -85,6 +90,11 @@ export interface SendBoxStrings {
    * Aria label for send message button
    */
   sendButtonAriaLabel: string;
+  /* @conditional-compile-remove(file-sharing) */
+  /**
+   * Error message indicating that all file uploads are not complete.
+   */
+  fileUploadsPendingError: string;
```
prabhjot-msft, line 97: Do we need this as a prop for sendbox as this is a standard message(Uploading..please wait)

packages/react-components/src/components/SendBox.tsx
prabhjot-msft: Not able to understand why progress is not reliable, isn't it updated to 1 when file upload is complete.

packages/react-components/src/components/SendBox.tsx
```diff
@@ -217,12 +242,36 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
...
   /* @conditional-compile-remove(file-sharing) */
-  const renderFileUploadErrorMessage: JSX.Element = useMemo(() => {
+  const uploadInProgressErrorBar = useMemo(() => {
+    const renderError = !!(
+      props.activeFileUploads?.length &&
+      !props.activeFileUploads
+        .filter((fileUpload) => !fileUpload.errorMessage)
+        .every((fileUpload) => fileUpload.isUploaded)
+    );
+    return renderError && <SendBoxErrorBar message={strings.fileUploadsPendingError} />;
```
prabhjot-msft, line 252: Can we update latesterror on line 258 to fileUploadsPendingError rather than adding SendBoxErrorBar here.

packages/react-components/src/components/SendBox.tsx
prprabhu-ms: Naming nit: `uploadComplete` is more assertive that something completed.

I agree that it's better to have this prop as separate from `progress`. Progress is used just for the progress bar. Through the file sharing API, the two are sperate (different API to notify file upload completion, separate state etc.)

packages/react-components/src/components/SendBox.tsx
```diff
@@ -186,13 +196,28 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
   const [textValue, setTextValue] = useState('');
   const [textValueOverflow, setTextValueOverflow] = useState(false);
 
+  /* @conditional-compile-remove(file-sharing) */
+  const [showFileUploadsPendingError, setShowFileUploadsPendingError] = useState(false);
```
prprabhu-ms, line 200: I haven't looked at the code below yet, but I'll bet managing this state is complex. Can you extract this into a React component that can encapsulate the logic?

prprabhu-ms: That way, you won't need so many conditional compilation directives either. You can simply add that component in the send box conditionally (but define the component unconditionally).

packages/react-components/src/components/SendBox.tsx
anjulgarg: It is but it doesn't guarantee that metadata is populated.

anjulgarg: progress is controlled by Contoso's logic, so there is no way for us to ensure that progress of 1 equals metadata being populated.

packages/react-components/src/components/SendBox.tsx
anjulgarg: We allow contoso to override all the strings. They might want to display a different error message.

packages/react-components/src/components/SendBox.tsx
```diff
@@ -217,12 +242,36 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
   };
 
   /* @conditional-compile-remove(file-sharing) */
-  const renderFileUploadErrorMessage: JSX.Element = useMemo(() => {
+  const uploadInProgressErrorBar = useMemo(() => {
```
prprabhu-ms, line 245: I'd recommend limiting this to a boolean `hasIncompleteUploads`. You can also add a helper file-level function for this to minimize the conditional compilation (that function will always return `false` on stable builds).

Then, you can use that boolean value to decide both whether to show the `SendBoxErrorBar` (hopefully encpasulated in its own component to handle all the timeouts) and whether to skip sending the file.

packages/react-components/src/components/SendBox.tsx
anjulgarg: The logic is quite complex for these error states.
Upload pending error is only shown when someone tries to send a message as opposed to fileUpload error that should be visible as soon as it's in state. So `latestError` won't work here.

prabhjot-msft (approved):

packages/react-components/src/components/SendBox.tsx
```diff
@@ -217,12 +242,36 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
...
     const latestError = fileUploads.filter((fileUpload) => fileUpload.errorMessage).pop();
-    return <SendBoxErrorBar message={latestError?.errorMessage} timeout={10 * 1000} />;
+    return latestError && <SendBoxErrorBar message={latestError.errorMessage} timeoutDelay={10 * 1000} />;
   }, [props.activeFileUploads]);
 
+  /* @conditional-compile-remove(file-sharing) */
+  const renderSendBoxError = () => {
+    if (showFileUploadsPendingError) {
+      fileUploadsPendingErrorTimeoutRef.current && clearTimeout(fileUploadsPendingErrorTimeoutRef.current);
+      fileUploadsPendingErrorTimeoutRef.current = setTimeout(() => {
```
PorterNan, line 266: What is the difference between this timeout and the timeoutDelay in SendBoxErrorBar?

packages/react-components/src/components/SendBox.tsx
prprabhu-ms: I see that you already have `SendBoxErrorBar`. So I guess, I'm wondering if we can somehow move the logic for `fileUploadPendingErrorTimeoutRef` inside that component.

Something like:
* Each time user clicks the send button, if there is a pending upload, render `SendBoxErrorBar` with the *current timestamp* (So in this component, we'll need state to track the last time send was clicked)
* Inside `SendBoxErrorBar`, have the logic to return `<></>` if it's been long enough since the last click (last click is provided via prop).

All of this can be unconditional. The only conditionally compiled bit will be the computation of whether there is any pending upload (see my comment below as well)

packages/react-components/src/components/SendBox.tsx
PorterNan, line 245: SendBox logic is getting complicated now - I am worried that it could become next messageThread - which is hard to read and maintain since code pieces are flying all around in the same file. And there is a lot /* @conditional-compile-remove(file-sharing) */ which is hard to follow the stable version logic now

I believe we could separate this logic and all hooks, put them into one single component maybe name it FileSharingSendBoxErrorComponent or so, then the hierarchy is:
SendBox -> FileSharingSendBoxErrorComponent (choose which fileUploadError to be rendered) -> SendBoxErrorBar

The idea is keeping each part small and easy to read and easy for refactoring

packages/react-components/src/components/SendBox.tsx
PorterNan, line 252: It looks like all the errorBars are sharing the same underlying component - SendBoxErrorBar, is it possible to only change error message string instead of rendering different component? (Could be easier if we separate logic out into a single error component like what I mentioned)

packages/react-components/src/components/SendBox.tsx
```diff
@@ -217,12 +242,36 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
...
+  const fileUploadErrorBar = useMemo(() => {
     const fileUploads: ActiveFileUpload[] = props.activeFileUploads || [];
     const latestError = fileUploads.filter((fileUpload) => fileUpload.errorMessage).pop();
-    return <SendBoxErrorBar message={latestError?.errorMessage} timeout={10 * 1000} />;
+    return latestError && <SendBoxErrorBar message={latestError.errorMessage} timeoutDelay={10 * 1000} />;
   }, [props.activeFileUploads]);
 
+  /* @conditional-compile-remove(file-sharing) */
+  const renderSendBoxError = () => {
+    if (showFileUploadsPendingError) {
```
PorterNan, line 264: Aha, I kinda get why this logic is so hard to write - we only want to show 1 error bar above sendbox so we put a condition to choose "this" and "that"
I am thinking of maybe it is not bad to show multiple error bar - pending error message won't eliminate those file upload messages - cause they are totally different messages and one should not override another - if we could show them all togehter instead of just one - things will be easier (But maybe we should ask AlexP for his suggestion to see which is the right experience)

packages/react-components/src/components/SendBox.tsx
```diff
@@ -217,12 +242,36 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
...
+      fileUploadsPendingErrorTimeoutRef.current && clearTimeout(fileUploadsPendingErrorTimeoutRef.current);
+      fileUploadsPendingErrorTimeoutRef.current = setTimeout(() => {
+        setShowFileUploadsPendingError(false);
```
PorterNan, line 267: If this fileUploadsPendingErrorTimeoutRef is just because we want to set some state outside - then we can reuse timeoutDelay logic in SendBoxErrorBar:
1. expose an onDismiss callback in SendBoxErrorBar props, 
2. pass setShowFileUploadsPendingError to it, so it gets called when the errorBar gets dismissed

At this we could save one more timeout object from our component (the less timeout logic we have, the better)

packages/react-components/src/components/SendBox.tsx
JamesBurnside: I guess the confusion is that `progress === 1` does not actually mean the file upload progress has truly completed - there is an extra processing step (performed by the server/place it is uploaded to?) where metadata is populated and added to the file.
If I've understood that correctly I would suggest either marking this property not as optional but as required and rename to something like `readyToSend`, or have two interfaces:
`ActiveFileUploads` and `CompletedFileUploads`.

The reason I would opt for two different interfaces is I can set the progress to 50% here, but also set uploadComplete as true which doesn't make sense

packages/react-components/src/components/SendBox.tsx
anjulgarg: @pereiralex thoughts?

packages/react-components/src/components/SendBox.tsx
anjulgarg: This timeout is used to hide the upload in progress error after 10 seconds.
We can't use sendboxerror timeout for this because not only do we need to hide the error after 10 seconds, we also need to ensure that it is not rendered again until someone tries to send the message.
(there is a state variable for that `showFileUploadsPendingError` which is set to false after 10 seconds using this timeout)

packages/react-components/src/components/SendBox.tsx
anjulgarg: @PorterNan and you both have good suggestions around moving this logic into a different component.
Instead of modifying `SendBoxErrorBar` component (which can be used in the future for other scenarios as well), I would like to introduce a new component `SendBoxErrors` that encapsulates all the complex logic in SendBox

packages/react-components/src/components/SendBox.tsx
anjulgarg: Will do

packages/react-components/src/components/SendBox.tsx
anjulgarg: I am afraid the logic is not as trivial. I think you understand that already. Let me know if you need me to explain more.

packages/react-components/src/components/SendBox.tsx
anjulgarg: Love that idea. Let me try that.

packages/react-components/src/components/SendBoxErrorBar.tsx
```diff
@@ -11,31 +11,33 @@ export interface SendBoxErrorBarProps {
   /** Error message to render */
   message?: string;
   /**
-   * Timeout in ms after which the error bar disappears.
+   * Timeout delay in ms after which the error bar disappears.
    * If undefined, the error bar will never disappear.
    */
-  timeout?: number;
+  timeoutDelay?: number;
```
alkwa-msft, line 17: nit. 

I am curious can we just call this delay?

timeoutDelay sounds redundant however I dont have a strong opinion.

anjulgarg: `SendBoxError.delay`
Just by looking at the property in association with the component name, I can't understand what it does.
`timeoutDelay` is also not much better in this case.
How about `hideAfterMs`?

packages/react-components/src/components/SendBox.tsx
anjulgarg: I think making `isUploaded` a required prop might be a good start.
Multiple interfaces will add complexity.

JamesBurnside: Sounds good, maybe just add a comment in the tsdoc comment to progress mentioning how their related.
```
/** @remarks if isUploaded is true, progress bar will not be shown */
progress
```

packages/react-components/src/components/SendBoxErrors.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * @internal
+ */
+export interface SendBoxErrorsProps {
+  fileUploadsPendingError?: SendBoxError;
+  onDismissFileUploadsPendingError?: () => void;
+  fileUploadError?: SendBoxError;
+}
+
+/**
+ * @internal
```
JamesBurnside, line 25: if it's not used outside of the react-components package, mark as `@private`

packages/react-components/src/components/SendBox.tsx
anjulgarg: Spoke to AlexP, we want to replace errors and only show 1 at a time.

packages/react-components/src/components/SendBoxErrors.tsx
anjulgarg: Thanks.

packages/react-components/src/components/SendBox.tsx
```diff
@@ -58,13 +58,24 @@ export interface ActiveFileUpload {
 
   /**
    * A number between 0 and 1 indicating the progress of the upload.
+   * This is unrelated to the `uploadComplete` property.
```
prprabhu-ms, line 61: <3

packages/react-components/src/components/SendBox.tsx
prprabhu-ms: Also, this is the only way we can localize the string.

packages/react-components/src/components/SendBox.tsx
```diff
@@ -188,11 +204,34 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
 
   const sendTextFieldRef = React.useRef<ITextField>(null);
 
+  /* @conditional-compile-remove(file-sharing) */
```
prprabhu-ms, line 207: This state can be added uncondtionally (but will stay `false` all the time in stable builds).

See below.

packages/react-components/src/components/SendBox.tsx
```diff
@@ -188,11 +204,34 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
...
+  const [showFileUploadsPendingError, setShowFileUploadsPendingError] = useState(false);
+
+  /* @conditional-compile-remove(file-sharing) */
+  const hasIncompleteFileUploads = useCallback(() => {
```
prprabhu-ms, line 211: Why is this a `useCallback`, and not a `useMemo`? The only use of `hasIncompleteFileUploads` is on line 230, where it is called instead of being passed as a callback.

Also, I'd suggest that only the logic be conditional, and need not even be memoized (it's a boolean value that isn't that costly to compute). It can just be a file-private helper function, that's directly called where needed:

```ts

const hasIncompleteFileUploads = (props) => {
 /* @conditional-compile-remove-from-stable */
return props.activeFileUploads .....;
// Always return false on stable so that message can be sent and on error is shown
return false;
}
```

This would be the *only* conditional logic in SendBox, everything else can be unconditional, in stable build `hasIncompleteFileUploads` will always compute to false, and hence the error will never be shown.

packages/react-components/src/components/SendBox.tsx
```diff
@@ -256,11 +288,24 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
     [mergedSendIconStyle, onRenderIcon]
   );
 
+  /* @conditional-compile-remove(file-sharing) */
```
prprabhu-ms, line 291: Move the conditional compilation directive inside the `useMemo` (and `return ''` in stable build), this way your don't need the conditional compilation directive on line 299 at all. You can *always* have the `SendBoxErrors` component, it just will never be shown in stable build.

packages/react-components/src/components/SendBoxErrorBar.tsx
```diff
@@ -2,40 +2,49 @@
...
+  onDismiss?: () => void;
 }
 
 /**
- * @internal
+ * @private
  */
-export const SendBoxErrorBar = (props: { message?: string; timeout?: number }): JSX.Element => {
-  const { message, timeout } = props;
+export const SendBoxErrorBar = (props: SendBoxErrorBarProps): JSX.Element => {
```
prprabhu-ms, line 27: Have you tried multiple error messages:

* Set things up so that we see an error message
* Wait 10 seconds for message to dissapper
* Cause another error message
* Wait another 10 seconds

I am worried that the latter message will never be dismissed. The problem is that React may keep the `SendBoxErrorBar` component in the react component tree even after the 10 second dismissal. So, when the next error is shown (with the next message), none of `clearTimeoutRef`, `dismissAfterMs` or `onDismiss` have changed. Thus, the dismissal logic will not be triggered.


Another issue this might cause:

* Show an error
* After 5 seconds, show a different error
* The message will be cleared after 5 seconds (i.e. 10 seconds after the _first_ message was shown)

Again, because the dismissal logic is unrelated to the message being shown.

packages/react-components/src/components/SendBox.tsx
```diff
@@ -256,11 +288,24 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
...
+    return props.activeFileUploads?.filter((fileUpload) => fileUpload.errorMessage).pop()?.errorMessage;
+  }, [props.activeFileUploads]);
+
   return (
     <Stack className={mergeStyles(sendBoxWrapperStyles)}>
       {
         /* @conditional-compile-remove(file-sharing) */
-        renderFileUploadErrorMessage
+        <SendBoxErrors
+          {...{
```
PorterNan, line 301: Why do we use the spread operator instead of put all properties directly into the component?

packages/react-components/src/components/SendBoxErrors.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }, [fileUploadError]);
+
+  if (showFileUploadsPendingError && fileUploadsPendingError) {
+    return (
+      <SendBoxErrorBar
+        message={fileUploadsPendingError.message}
+        dismissAfterMs={10 * 1000}
+        onDismiss={() => {
+          setShowFileUploadsPendingError(false);
+          onDismissFileUploadsPendingError && onDismissFileUploadsPendingError();
```
PorterNan, line 47: So for all the other file upload error, we don't need a onDismiss for them
I am thinking of if we could make it more generic for any future needs 

Something like onDismissError(errorType: 'fileUploadPending' | 'uploadFailed') => ()

So we don't have to add another onDismissError for other errors in the future, and we can always do conditional logic to determine what to do

packages/react-components/src/components/SendBoxErrorBar.tsx
```diff
@@ -2,40 +2,49 @@
...
+  const { message, dismissAfterMs, onDismiss } = props;
   const [errorMessage, setErrorMessage] = React.useState(message);
+  // Using `any` because `NodeJS.Timeout` here will cause `declaration error` with jest.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const timeoutRef = React.useRef<any>();
```
PorterNan, line 32:
```suggestion
  const timeoutRef = React.useRef<NodeJS.Timeout>();
```

anjulgarg: That won't work. Checkout my comment.

prprabhu-ms (approved):

packages/react-components/src/components/SendBox.tsx
anjulgarg: Updated the code to make it more sensible. :)

packages/react-components/src/components/SendBoxErrorBar.tsx
anjulgarg: Had a meeting with @prprabhu-ms  and verified that the implementation works. Added some tweaks to make it stable

PR #1600, opened 2022-03-04
Fix eslint warnings - Final (react-composites)

JamesBurnside:
# What
Fix eslint warnings and fail builds on eslint warnings

**NOTE:** please review the changes to the react hook dependency arrays closely, these can easily cause regressions.

# Why
Continuation of #1597 and #1598 

# How Tested
I haven't but if please inspect the useEffect dependency changes closely


alkwa-msft (approved):

packages/react-composites/src/composites/CallWithChatComposite/PeopleButton.tsx
```diff
@@ -24,7 +24,7 @@ export const PeopleButton = (props: ControlBarButtonProps): JSX.Element => {
         },
         props.styles ?? {}
       ),
-    [theme]
+    [props.styles, theme.palette.neutralLight]
```
prprabhu-ms, line 27: nit: I tend to think of `theme` as an atomic unit. I'd not expect Contoso to modify parts of the theme so it's OK to add `theme` as the dependency (especially if the list of theme fields we reference starts growing)

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
```diff
@@ -55,16 +55,17 @@ export interface MoreDrawerProps extends MoreDrawerDevicesMenuProps {
...
+      const selected = speakers?.find((speaker) => speaker.id === itemKey);
       if (selected) {
         // This is unsafe - we're only passing in part of the argument to the handler.
         // But this is a known issue in our state.
-        props.onSelectSpeaker(selected as AudioDeviceInfo);
+        onSelectSpeaker(selected as AudioDeviceInfo);
       }
     },
-    [props.speakers, props.onSelectSpeaker]
+    [speakers, onSelectSpeaker]
```
prprabhu-ms, line 68: Why did you need to do this?

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -101,6 +101,7 @@ export const ChatComposite = (props: ChatCompositeProps): JSX.Element => {
    * @TODO Remove this function and pass the props directly when file-sharing is promoted to stable.
    * @private
    */
+  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
prprabhu-ms, line 104: You can use `FileSharingOptions` return type. The type is defined unconditionally (the field `ChatCompositeOptions.fileSharing` is conditionally defined.

https://github.com/Azure/communication-ui-library/blob/e9622a0f52b0d3a3218d9507d6f2dc5f0d08d6ee/packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx#L77

prprabhu-ms (approved):

PR #1603, opened 2022-03-07
Flatten pick and omit APIs in CallWithChatComposite

JamesBurnside:
# What
Flatten following CallWithChatComposite APIs that were using Pick and Omit:
* CallWithChatAdapterManagement
* CallWithChatAdapterUiState
* CallWithChatClientState
* CallWithChatControlOptions

# Why
Azure Review Board guidance -- doing so improves intellisense, improves doc generation and improves dev experience when looking at the API in the .d.ts file.

# How Tested
n/a

### DIFF
--- a/packages/react-composites/review/react-composites.api.md
+++ b/packages/react-composites/review/react-composites.api.md
@@ -353,48 +353,30 @@ export interface CallWithChatAdapter extends CallWithChatAdapterManagement, Adap
 
 // @beta
 export interface CallWithChatAdapterManagement {
-    // @public
     askDevicePermission(constrain: PermissionConstraints): Promise<void>;
-    // @public
     createStreamView(remoteUserId?: string, options?: VideoStreamOptions): Promise<void>;
     deleteMessage(messageId: string): Promise<void>;
-    // @public
     disposeStreamView(remoteUserId?: string, options?: VideoStreamOptions): Promise<void>;
     fetchInitialData(): Promise<void>;
-    // @public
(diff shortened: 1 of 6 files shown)


packages/react-composites/review/react-composites.api.md
```diff
@@ -539,8 +569,15 @@ export interface CallWithChatCompositeStrings {
 
 // @beta
 export interface CallWithChatControlOptions extends Pick<CallControlOptions, 'cameraButton' | 'microphoneButton' | 'screenShareButton' | 'displayType'> {
```
prprabhu-ms, line 571: I'm confused. So we now have *both* `extends Pick<CallControlOptions, 'cameraButton'...` and an explicit `cameraButton?: boolean` field? Worse, the two can diverge without build failure.

What I was expecting here:
* Drop the `Pick<...>` entirely.
* Add internal build-time check that `CallWithChatControlOptions` indeed extends `Pick <..>` etc.

JamesBurnside: oh nope, that was not removed by accident! but I really like that internally we still use Pick, I will update with that

prprabhu-ms: See https://github.com/Azure/communication-ui-library/pull/1560/files#diff-db700faf40319449d13fa41bfdf20a7dc853a55c7f6a8e6f8dd3bb09d85a51d5 for an example.

JamesBurnside:
Closing as need to make an update to this

PR #1605, opened 2022-03-07
Add `useAzureCommunicationCallWithChatAdapter`

prprabhu-ms:
# What

* Introduce a new React hook to create `CallWithChatAdapter`
* Use this hook in storybook documentation, sample, e2e test app.

# Why

Feedback from API review.

# How Tested
* Storybook, sample, e2e test.


packages/react-composites/tests/browser/callwithchat/app/index.tsx
```diff
@@ -71,18 +62,18 @@ function App(): JSX.Element {
...
-  } else if (!callWithChatAdapter) {
+  } else if (!adapter) {
     return <h3>Initializing call and chat adapters...</h3>;
   }
 
   return (
     <>
-      {!callWithChatAdapter && 'Initializing call-with-chat adapter...'}
-      {callWithChatAdapter && (
+      {!adapter && 'Initializing call-with-chat adapter...'}
```
JamesBurnside, line 71: Can remove this, should be covered in the `else if (!adapter) { return ...` above

JamesBurnside (approved):

packages/react-composites/tests/browser/callwithchat/app/index.tsx
prprabhu-ms: Being lazy as this is unrelated to current PR.

prprabhu-ms:
Please take another look @JamesBurnside because there are non-trivial updates to the new hook.

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -499,6 +500,97 @@ export const createAzureCommunicationCallWithChatAdapter = async ({
...
+      (async () => {
+        if (adapter) {
+          if (beforeDisposeRef.current) {
+            await beforeDisposeRef.current(adapter);
+          }
+          adapter.dispose();
+        }
+      })();
+    };
+  }, [adapter, beforeDisposeRef]);
```
JamesBurnside, line 590: Is it OK that we construct, assing, use a new adapter before disposing the old one?
i.e.
- adapter is created
- displayName updated
- new adapter created
  - afterCreateCalled()
- ... some async time passes 
- oldAdapterDispose async function is triggered
- ... some async time passes 
- oldAdapter is disposed

JamesBurnside: Should we call dispose in front of `let newAdapter = await createAzureCommunicationCallWithChatAdapter({` also?

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -499,6 +500,97 @@ export const createAzureCommunicationCallWithChatAdapter = async ({
...
+      (async () => {
+        let newAdapter = await createAzureCommunicationCallWithChatAdapter({
+          credential,
+          displayName,
+          endpoint,
+          locator,
+          userId
+        });
+        if (afterCreateRef.current) {
+          newAdapter = await afterCreateRef.current(newAdapter);
```
JamesBurnside, line 564: what is the `await afterCreateRef.current(newAdapter);`
Could this not be awaited and just be
`afterCreateRef.current = newAdapter`;

And what does assigning it back to newAdapter do?

JamesBurnside: never mind I see now!

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
prprabhu-ms: Great catch. You are right, we aren't allowed to have to adapters (actually two `CallAgent` instances) at the same time.

I added a dumb rename loop in the sample app to trigger this behavior: 
image

and did see failure in creating adapter.

The new commit fixes this issue. We need another `useRef` so that we don't cause infinite adapter creation loop.

Look at my very stupid sample app that keeps renaming the local participant :D 

https://user-images.githubusercontent.com/82062616/157300969-7ffb15e6-7d71-43ab-85a4-97fc1a8207ae.mp4

prprabhu-ms: Of course, if you have a new adapter, you reset the `CallWithChatComposite`, so that you end up back on the configuration page.

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -499,6 +500,97 @@ export const createAzureCommunicationCallWithChatAdapter = async ({
...
+  afterCreateRef.current = afterCreate;
+  beforeDisposeRef.current = beforeDispose;
+
+  useEffect(
+    () => {
+      if (!credential || !displayName || !endpoint || !locator || !userId) {
+        return;
+      }
+      (async () => {
+        if (adapterRef.current) {
```
PorterNan, line 560: We are expecting to create a new adapter whenever user put new adapter arguments into the hook, nice!
I am a little bit curious on calling scenarios - we met some problems when create call client twice in the same window (even we dispose it), now sure if the problem still exists

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -499,6 +500,97 @@ export const createAzureCommunicationCallWithChatAdapter = async ({
...
+  // are used in the `useEffect` closures below.
+  // Using a Ref ensures that new values for the callbacks do not trigger the
+  // useEffect blocks, and a new adapter creation / distruction is not triggered.
+  afterCreateRef.current = afterCreate;
+  beforeDisposeRef.current = beforeDispose;
+
+  useEffect(
+    () => {
+      if (!credential || !displayName || !endpoint || !locator || !userId) {
+        return;
```
PorterNan, line 557: Do we want to throw errors here? I guess it is hard for developers to debugs this if they are fetching things from their server side

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
JamesBurnside: Hehe love the video!

change/@internal-react-composites-a06cd2c2-cdfc-4e84-beb0-a318c49c41d5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Add hook to create AzureCommunicationCallWithChatAdapter",
```
anjulgarg, line 3: adding the name of the hook would make for a great changelog. :)

change/@internal-storybook-c102ceb1-4852-4c61-8fb0-a04c49e1e0c9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Use react hook to create AzureCommunicationCallWithChatAdapter",
```
anjulgarg, line 3: adding the name of the hook would make for a great changelog. :)

packages/react-composites/tests/browser/callwithchat/app/index.tsx
```diff
@@ -33,31 +33,25 @@ const threadId = verifyParamExists(params.threadId, 'threadId');
...
-    };
-
-    initialize();
-
-    return () => {
-      callWithChatAdapter && callWithChatAdapter.dispose();
-    };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  const userIdArg = useMemo(() => fromFlatCommunicationIdentifier(userId) as CommunicationUserIdentifier, []);
```
PorterNan, line 36: We've mentioned we wanted to useMemo in the jsdoc of useAzureCommunicationCallWithChatAdapter, but would be nice to still add some comments here of why we call "useMemo", I believe there would be a lot developers only read sample code without looking at api doc (like me sometimes)

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
prprabhu-ms: See my video above at  https://github.com/Azure/communication-ui-library/pull/1605#discussion_r821946184 you can create multiple call clients safely after this PR.

packages/react-composites/tests/browser/callwithchat/app/index.tsx
PorterNan: Ohh, this is test code not code snippet, ignore this

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -499,6 +500,97 @@ export const createAzureCommunicationCallWithChatAdapter = async ({
...
+   * Allows arguments to be undefined so that you can respect the rule-of-hooks and pass in arguments
+   * as they are created. The adapter is only created when all arguments are defined.
+   */
+  args: Partial<AzureCommunicationCallWithChatAdapterArgs>,
+  /**
+   * Optional callback to modify the adapter once it is created.
+   *
+   * If set, must return the modified adapter.
+   */
+  afterCreate?: (adapter: CallWithChatAdapter) => Promise<CallWithChatAdapter>,
```
anjulgarg, line 530: The idea is to allow devs to monkey patch this adapter?
If yes, then it is no different from replacing object properties on the adapter resolved by this function.
This callback seems redundant.
What am I missing?

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
prprabhu-ms: Nope. See docstring. This is intentional.

Contoso must call `useAzureCommunicationCallWithChatAdapter` unconditionally due to rule-of-hooks. But they may not have all the arguments available, so they can provide arguments as they become available and the adapter will be created then.

e.g., this is needed in our storybook examples, where the `credential` is created by a hook, and may not be set (we show a different error if the user provided a corrupt token and so we failed to create the `credential`).

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
prprabhu-ms: This directly replaces some `useEffect` calls in our code.

The problem is: they have to monkey patch the adapter only once -- when it is created. So, by providing this argument, they get to do it in a single hook, instead of using another `useEffect` to observe when `adapter` changes.

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -499,6 +500,97 @@ export const createAzureCommunicationCallWithChatAdapter = async ({
...
+  beforeDispose?: (adapter: CallWithChatAdapter) => Promise<void>
+): CallWithChatAdapter | undefined => {
+  const { credential, displayName, endpoint, locator, userId } = args;
+
+  // State update needed to rerender the parent component when a new adapter is created.
+  const [adapter, setAdapter] = useState<CallWithChatAdapter | undefined>(undefined);
+  // Ref needed for cleanup to access the old adapter created asynchronously.
+  const adapterRef = useRef<CallWithChatAdapter | undefined>(undefined);
+
+  const afterCreateRef = useRef(afterCreate);
```
anjulgarg, line 545: This is misleading/confusing.
We are initializing `afterCreateRef` with `afterCreate` and later setting `afterCreateRef.current = afterCreate` 
I recommend we initialize the refs with undefined values.

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -499,6 +500,97 @@ export const createAzureCommunicationCallWithChatAdapter = async ({
...
+  const beforeDisposeRef = useRef(beforeDispose);
+  // These refs are updated on *each* render, so that the latest values
+  // are used in the `useEffect` closures below.
+  // Using a Ref ensures that new values for the callbacks do not trigger the
+  // useEffect blocks, and a new adapter creation / distruction is not triggered.
+  afterCreateRef.current = afterCreate;
```
anjulgarg, line 551: Why do we need to assign the values to ref again when we have already done it above?

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
PorterNan: Ohh yea, the super annoying unconditional hook rules, makes a lot of sense

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
anjulgarg: Right. I realised it when I looked at the implementation in the last file.

samples/CallWithChat/src/app/views/CallScreen.tsx
```diff
@@ -30,37 +30,16 @@ export interface CallScreenProps {
...
+  );
 
-      const adapter = await createAzureCommunicationCallWithChatAdapter({
-        userId,
-        displayName,
-        credential: createAutoRefreshingCredential(toFlatCommunicationIdentifier(userId), token),
-        endpoint,
-        locator: locator
-      });
+  const afterAdapterCreate = useCallback(
```
PorterNan, line 41: This looks good - I am thinking of what is the best practice to use afterAdapterCreate vs useAzureCommunicationCallWithChatAdapter + useEffect

Imagine a piece of code like this, it seems to do the same job:

```
const adapter = useAdapter(...);
useEffect(()=> {
  if(adapter) {
    adapter.on('error', (e) => {...});
    decoratedAdapter = decorateAdapter(adapter);
  }
}, [adapter]);
```

Is there a specific case we want to ask developers to use afterAdapterCreate over useAzureCommunicationCallWithChatAdapter + useEffect?

PorterNan: Would be nice to document this if there is any preference

PorterNan (approved):
Good to see we are making it simpler! I am wondering if this "react way" is also a pattern we want to apply to chat and calling adapter

samples/CallWithChat/src/app/views/CallScreen.tsx
prprabhu-ms: `afterCreate` is a direct replacement for the extra `useEffect`. Less is more, so use this not `useEffect` :)

As for documentation, we use this in our samples after this PR, and we will also use `beforeDispose` in storybook for `CallAdapter` (follow up PR).

prprabhu-ms:
> Good to see we are making it simpler! I am wondering if this "react way" is also a pattern we want to apply to chat and calling adapter

Totally. But getting this right wasn't easy so I'm merging this first so I can copy-pasta :D

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
prprabhu-ms: Updated the initializer to `undefined`. As the comment here states, we need to update the cached value on each render.

PR #1606, opened 2022-03-07
Enhanced chat tooltips

dmceachernmsft:
# What
Add strings for num of messages in tool tips
image

# Why

Adds more information about how many messages are waiting in the tool tip

# How Tested

Ran and inspected locally


packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -47,6 +47,7 @@
     "peopleButtonTooltipContent": "Open People Pane",
     "chatButtonLabel": "Chat",
     "chatButtonTooltipContent": "Open Chat Pane",
+    "chatButtonEnhancedToolTipContent": "new",
```
JamesBurnside, line 50: This should just be it's own totally unique string value, something like:
```
chatButtonTooltipContentWithUnreadMessages: "Open Chat Pane - You have new messages! 😲"
```

then tooltip becomes something like:
```
let tooltip = unreadMessages  && strings.chatButtonTooltipContentWithUnreadMessages ||  strings.chatButtonTooltipContent;
```

packages/react-composites/src/composites/CallWithChatComposite/ChatButtonWithUnreadMessagesBadge.tsx
```diff
@@ -38,6 +39,23 @@ export const ChatButtonWithUnreadMessagesBadge = (props: ChatButtonWithUnreadMes
   const [unreadChatMessagesCount, setUnreadChatMessagesCount] = useState<number>(0);
 
   const baseIcon = props.showLabel ? regularIcon : filledIcon;
+  const callWithChatStrings = useCallWithChatCompositeStrings();
+
+  const numberOfMsgToolTip =
+    props.strings?.tooltipOffContent && unreadChatMessagesCount > 0
+      ? props.strings?.tooltipOffContent +
+        ' (' +
```
JamesBurnside, line 47: This shouldn't use string building like this but instead the string in the strings.json should be something like `Show chat ({unreadMessagesCount} unread)` then use _formatString function. See other string building examples in the repo for how this is done such as the readBy tooltip

JamesBurnside (changes requested):

packages/react-composites/src/composites/CallWithChatComposite/ChatButtonWithUnreadMessagesBadge.tsx
dmceachernmsft: will do!

packages/react-composites/review/react-composites.api.md
```diff
@@ -525,6 +525,7 @@ export interface CallWithChatCompositeProps extends BaseCompositeProps<CallWithC
 
 // @public
 export interface CallWithChatCompositeStrings {
+    chatButtonEnhancedToolTipContent: string;
```
prprabhu-ms, line 528: This name will confuse a Contoso dev trying to override strings (enhanced in what way?)

How about, `chatButtonMessageCountToolTipContent`

(aside: I don't like the "Content" suffix in all these names. All strings are "contents", we could just say `chatButtonMessageCountToolTip` and it would convey the same thing. I know we're just following Fluent's naming here, so it's good to stick with the longer name)

dmceachernmsft: okay will change these strings so that the names are more appropriate

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -44,11 +44,12 @@
...
   "callWithChat": {
     "peopleButtonLabel": "People",
-    "peopleButtonTooltipContentOpen": "Show Participants",
-    "peopleButtonTooltipContentClose": "Hide Participants",
+    "peopleButtonTooltipOpen": "Show participants",
+    "peopleButtonTooltipClose": "Hide participants",
     "chatButtonLabel": "Chat",
-    "chatButtonTooltipContentOpen": "Show Chat",
-    "chatButtonTooltipContentClose": "Hide Chat",
+    "chatButtonNewMessageCountToolTip": " ({unreadMessagesCount} unread)",
```
JamesBurnside, line 50: I'd like to hear what others think, but I feel strings should be self contained and not appended to each other.
So this would be `chatButtonWithNewMessageCountTooltipOpen: "Show chat NewMessageCount"` -- but I'd liek to know what others think

prprabhu-ms: I agree with that. I actually thought that Donald was making that change, that's why haven't reviewed this PR today.

dmceachernmsft: ahhh okay, I will make that adjustment and keep that in mind for the future 😊

JamesBurnside (approved):

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -44,11 +44,12 @@
...
   "callWithChat": {
     "peopleButtonLabel": "People",
-    "peopleButtonTooltipContentOpen": "Show Participants",
-    "peopleButtonTooltipContentClose": "Hide Participants",
+    "peopleButtonTooltipOpen": "Show participants",
+    "peopleButtonTooltipClose": "Hide participants",
     "chatButtonLabel": "Chat",
-    "chatButtonTooltipContentOpen": "Show Chat",
-    "chatButtonTooltipContentClose": "Hide Chat",
+    "chatButtonNewMessageCountToolTip": "Show chat ({unreadMessagesCount} unread)",
```
prprabhu-ms, line 50: suggestion: This string is actually related to `chatButtonTooltipOpen` (it adds stuff on top).
A name like `chatButtonTooltipOpenWithMessageCount` would better reflect that than `chatButtonNewMessageCountTooltip`.

prprabhu-ms (approved):

packages/react-composites/src/composites/localization/locales/en-US/strings.json
dmceachernmsft: kk will update the name! names are hard 😊

PR #1609, opened 2022-03-08
Update split button border radius

dmceachernmsft:
# What

Increases split button Border radius
# Why

Makes the border radius of the buttons match with the Figma designs

https://skype.visualstudio.com/SPOOL/_workitems/edit/2781977
# How Tested

Ran locally with visual inspection.


change/@internal-react-composites-093c2064-8b62-442d-bf87-d075572716be.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
JamesBurnside, line 2: this should be minor as it visually impacts stable builds (even more so if contoso has set their own theme.effect.roundercorner2/4 value)

change/@internal-react-composites-093c2064-8b62-442d-bf87-d075572716be.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Update styles around split button border radius",
```
JamesBurnside, line 3: Try be more specific here; `Increase control bar button border radius in Composites with Call features`

edwardlee-msft (approved):

PR #1610, opened 2022-03-08
Bump `@azure/communication-calling` to `1.4.4`

prprabhu-ms:
# What

Bump stable headless calling SDK.

# Why

Has loads of bug fixes and improvements that we'd want our customers to get in our own stable release next.

# How Tested

CI, local calling sample smoke test.

### DIFF
--- a/packages/calling-component-bindings/package.json
+++ b/packages/calling-component-bindings/package.json
@@ -36,12 +36,12 @@
     "reselect": "~4.0.0"
   },
   "peerDependencies": {
-    "@azure/communication-calling": "1.4.3-beta.1 || 1.4.3",
+    "@azure/communication-calling": "1.4.3-beta.1 || 1.4.4",
     "@types/react": ">=16.8.0 <18.0.0",
     "react": ">=16.8.0 <18.0.0"
   },
   "devDependencies": {
-    "@azure/communication-calling": "1.4.3-beta.1 || 1.4.3",
+    "@azure/communication-calling": "1.4.3-beta.1 || 1.4.4",
     "@babel/cli": "~7.16.0",
     "@babel/core": "~7.16.0",
     "@microsoft/api-documenter": "~7.12.11",
--- a/packages/calling-component-bindings/package.json
+++ b/packages/calling-component-bindings/package.json
@@ -36,12 +36,12 @@
     "reselect": "~4.0.0"
   },
   "peerDependencies": {
-    "@azure/communication-calling": "1.4.3-beta.1 || 1.3.2",
+    "@azure/communication-calling": "1.4.3-beta.1 || 1.4.3",
     "@types/react": ">=16.8.0 <18.0.0",
     "react": ">=16.8.0 <18.0.0"
   },
   "devDependencies": {
-    "@azure/communication-calling": "1.4.3-beta.1 || 1.3.2",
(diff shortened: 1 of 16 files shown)


packages/calling-stateful-client/src/TestUtils.ts
```diff
@@ -95,15 +94,15 @@ export class MockRecordingCallFeatureImpl implements RecordingCallFeature {
   public recordings;
   public emitter = new EventEmitter();
   on(event: 'isRecordingActiveChanged', listener: PropertyChangedEvent): void;
-  /* @conditional-compile-remove(calling-1.4.3-beta.1) */
-  on(event: 'recordingsUpdated', listener: CollectionUpdatedEvent<RecordingInfo>): void;
+  // TODO: revert to using `RecordingInfo`.
```
prprabhu-ms, line 97: Working with the calling team to  be able to undo this before this PR is merged.

packages/calling-stateful-client/package.json
```diff
@@ -35,10 +35,10 @@
     "immer": "9.0.6"
   },
   "peerDependencies": {
-    "@azure/communication-calling": "1.4.3-beta.1 || 1.3.2"
+    "@azure/communication-calling": "1.4.3-beta.1 || 1.4.3"
```
JamesBurnside, line 38: I've asked this before but forget, can these be peer deps be `^1.4.3` so customers can use newer calling sdk versions without having to upgrade?

JamesBurnside (approved):

packages/calling-stateful-client/package.json
prprabhu-ms: Definitely not in this PR.

For stable dep, we can move to minimum-supported-sdk-version. The key decision there is about testing coverage and support expectations -- our CI can only test against some versions (in fact we'll want to test against the oldest supported version at any point) so we have to make sure we set expectations accordingly.

prprabhu-ms: I'd like to do this for stable deps. @alkwa-msft  thoughts?

packages/calling-component-bindings/package.json
```diff
@@ -36,12 +36,12 @@
     "reselect": "~4.0.0"
   },
   "peerDependencies": {
-    "@azure/communication-calling": "1.4.3-beta.1 || 1.3.2",
+    "@azure/communication-calling": "1.4.3-beta.1 || 1.4.3",
```
JamesBurnside, line 39: 1.4.3-beta.1 is behind 1.4.3, should this just be  
`"@azure/communication-calling": "1.4.3",`

or to make it simpler when we upgrade again
`"@azure/communication-calling": "1.4.3 || 1.4.3",` lol

prprabhu-ms: Yeah, it wouldn't work. Our tooling for changing the version for release branches is primitive. Also, @azure/communication-calling doesn't follow SemVer properly for beta packages (`1.4.3-beta.1` has lots of things not in `1.4.3`) so it is more "correct" to explicitly state that both are allowed.

JamesBurnside: Oh yeah good point it'll have beta functionality

edwardlee-msft (approved):

alkwa-msft (approved):

PR #1611, opened 2022-03-08
Update e2e tests to use local icons

JamesBurnside:
# What
Update e2e tests to use local icons instead of those from Fluent's CDN

# Why
Prevent flakiness in tests.
This is a stopgap fix to prevent flakiness today. The real fix is stop using icons from Fluent's CDN everywhere (tracked in another bug)

# How Tested
See PR for updated snapshots.
Icon styling isn't great (e.g. chevrondown icon is not vertically centered, see onion-skin diff below), but this would be tackled in the actual fix to stop using fluent's font icons
image


prprabhu-ms:
I'm worried that we'll start getting lots of review comments about misaligned icons in the snapshots, or worse learn to ignore these misalignments and let real bugs slip through.

JamesBurnside:
> I'm worried that we'll start getting lots of review comments about misaligned icons in the snapshots, or worse learn to ignore these misalignments and let real bugs slip through.

I agree. This basically is a real bug though -- it's not easy for contoso to set icons that are aligned correctly, but having flakiness is blocking having landscape tests so I'd like to check in regardless.
One mitigation would be to only set the people and chevronRight icons as they are the flakiness culprits

prprabhu-ms:
> > I'm worried that we'll start getting lots of review comments about misaligned icons in the snapshots, or worse learn to ignore these misalignments and let real bugs slip through.
> 
> I agree. This basically is a real bug though -- it's not easy for contoso to set icons that are aligned correctly, but having flakiness is blocking having landscape tests so I'd like to check in regardless. One mitigation would be to only set the people and chevronRight icons as they are the flakiness culprits. But really we need the proper fix to remove the use of fluent's font icons, discussion here: [teams.microsoft.com/l/message/...](https://teams.microsoft.com/l/message/19:5e150d250f724d1089c949305efdc8cc@thread.skype/1646760430458?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e9c1fc3-39df-4486-a26a-456d80e80f82&parentMessageId=1646760430458&teamName=Azure%20Communication%20Services&channelName=Web%20UI%20-%20Team&createdTime=1646760430458)

LGTM so we can get landscape tests. Let's see what we can do about those misaligned icons.

prprabhu-ms (approved):

anjulgarg:
I am confused.
I thought fluent doesn't contain any icons and all of them needed to be downloaded from a CDN?

JamesBurnside:
> I am confused. I thought fluent doesn't contain any icons and all of them needed to be downloaded from a CDN?

We use a package called `@fluentui/react-icons` for most of our icons, these are local SVGs that are bundled. However, we still have some instances where we use the fluent CDN font icons

PR #1612, opened 2022-03-08
[CallWithChat] Switch from Beta to Public and Stable

edwardlee-msft:
# What
Switch CallWithChat from Beta to Public and Stable

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2770679/

# How Tested
Test changes on local system on CallWithChat sample. Stable flavour

### DIFF
--- a/change/@internal-calling-component-bindings-c32a3bbb-b6fe-47b1-8ff5-319058087b27.json
+++ b/change/@internal-calling-component-bindings-c32a3bbb-b6fe-47b1-8ff5-319058087b27.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Switch CallWithChat from Beta to Public and Stable",
+  "packageName": "@internal/calling-component-bindings",
+  "email": "edwardlee@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/communication-react/stable-review/communication-react.api.md
+++ b/packages/communication-react/stable-review/communication-react.api.md
@@ -132,6 +132,24 @@ export type AzureCommunicationCallAdapterArgs = {
 };
 
 // @public
+export type AzureCommunicationCallWithChatAdapterArgs = {
+    endpoint: string;
+    userId: CommunicationUserIdentifier;
+    displayName: string;
+    credential: CommunicationTokenCredential;
+    locator: CallAndChatLocator | TeamsMeetingLinkLocator;
+};
+
+// @public
+export type AzureCommunicationCallWithChatAdapterFromClientArgs = {
+    callLocator: CallAdapterLocator | TeamsMeetingLinkLocator;
+    callAgent: CallAgent;
+    callClient: StatefulCallClient;
+    chatClient: StatefulChatClient;
+    chatThreadClient: ChatThreadClient;
+};
+
+// @public
 export type AzureCommunicationChatAdapterArgs = {
     endpoint: string;
     userId: CommunicationUserIdentifier;
@@ -250,6 +268,12 @@ export interface CallAgentState {
 }
 
 // @public
+export interface CallAndChatLocator {
+    callLocator: GroupCallLocator;
+    chatThreadId: string;
+}
+
+// @public
 export const CallClientProvider: (props: CallClientProviderProps) => JSX.Element;
 
 // @public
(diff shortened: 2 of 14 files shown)


change/@internal-calling-component-bindings-c32a3bbb-b6fe-47b1-8ff5-319058087b27.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
JamesBurnside, line 2: This is now a minor type!

packages/communication-react/stable-review/communication-react.api.md
```diff
@@ -1363,6 +1630,14 @@ export type LocalizationProviderProps = {
     children: React_2.ReactNode;
 };
 
+// @public (undocumented)
```
prprabhu-ms, line 1633: This shouldn't be part of the public API

@dmceachernmsft

prprabhu-ms:
1. Can you please merge in https://github.com/Azure/communication-ui-library/pull/1605 and make that the new hook is also exported in stable.

2. I expected to see more UI test snapshots for `CallWithChatComposite` for stable build. Looks like we've disabled those tests in CI. Can you please also find all instances of this and add stable flavor (like the other e2e tests):
3. 
image
(This is from .github/workflows/ci.yaml)

JamesBurnside:
> 2. I expected to see more UI test snapshots for `CallWithChatComposite` for stable build. Looks like we've disabled those tests in CI. Can you please also find all instances of this and add stable flavor (like the other e2e tests):

Also in update-snapshot.yml, remove this if:
image

packages/communication-react/stable-review/communication-react.api.md
dmceachernmsft: @prprabhu-ms This is going out at a different time than CallWithChat? that because of it showing up in the calling composite?

JamesBurnside:
Can you remove the beta flag in storybook also!
image

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -49,7 +49,7 @@ import { VideoTile } from './VideoTile';
 import { useId } from '@fluentui/react-hooks';
 /* @conditional-compile-remove(local-camera-switcher) */
 import { LocalVideoCameraCycleButton, LocalVideoCameraCycleButtonProps } from './LocalVideoCameraButton';
-/* @conditional-compile-remove(call-with-chat-composite) @conditional-compile-remove(local-camera-switcher) */
```
prprabhu-ms, line 52: Removing all these conditional compilation directives themselves was unnecessary. Just moving the feature to `stablizedFeatures` has the same effect.

It's OK that you've done it now, but the PR can be simpler (lower risk) by not removing these at the same time.

prprabhu-ms (approved):

prprabhu-ms:
Let's merge this early today so we have a bit of a lead time rest of the week before we cut the stable release branch.

Thanks for taking care of this!

packages/react-components/src/components/VideoGallery.tsx
edwardlee-msft: Having both the conditional compile for call-with-chat-composite and the local-camera-switcher was having issues when trying to keep only the local-camera-switcher in beta. 

When running `switch-flavor:stable && rush build -t communication-react`, it would complain about the apis for local-camera-switcher when both the conditional-compile-remove for call-with-chat-composite and local-camera-switcher were present.


I tried different combinations like:

```
/* @conditional-compile-remove(call-with-chat-composite) */
/* @conditional-compile-remove(local-camera-switcher) */

```


```
/* @conditional-compile-remove(call-with-chat-composite) @conditional-compile-remove(local-camera-switcher) */
```


```
/* @conditional-compile-remove(local-camera-switcher) */
/* @conditional-compile-remove(call-with-chat-composite) */
```

But all of them had issues for local-camera-switcher when placing call-with-chat-composite in stable until I removed the conditional-compile-remove(call-with-chat-composite).

edwardlee-msft: Reverted changes. Issue resolved in https://github.com/Azure/communication-ui-library/pull/1621

JamesBurnside (approved):

PR #1614, opened 2022-03-08
Narrator Reads mobile chat and More buttons

dmceachernmsft:
# What

Add strings for More button on mobile.
update ControlBarButton to read aria-labels from other strings
# Why

Allows for better narration of the main controls in the CallWIthChat

https://skype.visualstudio.com/SPOOL/_workitems/edit/2779666
# How Tested

Ran locally testing with windows narrator in mobile mode.


JamesBurnside (approved):

change/@internal-react-components-88cf0df0-80da-4850-97cd-484c6bca2969.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
JamesBurnside, line 2: nit: this is a patch change, affects stable build

JamesBurnside: (other changefile stays prerelease)

packages/react-composites/src/composites/CallWithChatComposite/Strings.tsx
```diff
@@ -45,6 +45,14 @@ export interface CallWithChatCompositeStrings {
    * Only used with `mobileView` set to `true`.
    */
   moreDrawerSpeakerMenuTitle: string;
+  /**
+   * {@Link CallWithChatComposite} Mobile Control bar More drawer label for narrator
```
prprabhu-ms, line 49: The comment doesn't seem to match the field name / intent. This is also the label that is shown on the UI, right?

packages/react-composites/src/composites/CallWithChatComposite/Strings.tsx
```diff
@@ -45,6 +45,14 @@ export interface CallWithChatCompositeStrings {
...
+   */
+  moreDrawerButtonLabel: string;
+  /**
+   * {@Link CallWithChatComposite} Mobile Control bar More drawer label for narrator
```
prprabhu-ms, line 53: The comment doesn't seem to match intent. This is used in the tooltip that is shown in the UI on hover, right?

prprabhu-ms (approved):

packages/react-composites/src/composites/CallWithChatComposite/Strings.tsx
dmceachernmsft: Right, I labelled it in the context I was working on when its really a side effect! will change this to be more appropriate 😅

PR #1617, opened 2022-03-08
Added PiPiP in Chat pane of CallWithChat composite (mobile view)

mgamis-msft:
# What

Added PiPiP in Chat pane.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2783448

# How Tested

CallWithComposite sample testing:
https://user-images.githubusercontent.com/79475487/157643190-5b2dd5ca-5177-48ad-bb21-57dfc33913f1.mp4

# Process & policy checklist


packages/react-composites/src/composites/CallWithChatComposite/EmbeddedChatPane.tsx
```diff
@@ -46,16 +65,23 @@ export const EmbeddedChatPane = (props: {
...
-        onClose={props.onClose}
-        activeTab="chat"
-        onChatButtonClicked={props.onChatButtonClick}
-        onPeopleButtonClicked={props.onPeopleButtonClick}
+        modalLayerHostId={props.modalLayerHostId}
+        styles={pipStyles}
       >
-        {chatComposite}
-      </MobilePane>
+        <MobilePane
```
JamesBurnside, line 74: Implementation is throwing me a little here and it causes the position shift between chat and people panes. Can you check with AlexP if the placement of the PiPiP should remain unchanged when shifting between people and chat pane? it seems odd to have it suddenly move. I believe the implementation is also causing the PiPiP refresh between changing side panes.

I think we need to have the modal live outside of the specific chat and people panes, and inside a generic component such that is doesn't need to be recreated if the user switches between panes

Rather than having two seperate panes as we do now:
```
- CompositeContainer
  | - Chat Side Pane
  | - People Side Pane
```

I would suggest something more along the lines of:
```
- CompositeContainer
 | - GenericSidePane
    | - Side Pane Header
    | - Side Pane Contents (i.e. PeoplePaneContents / ChatPaneContents)
    | - PiPiP
```
In jsx an impl could be something like:
```
<SidePane props: { selectedHeader: 'chat' | 'people'; children } >
  <SidePaneHeader selectedHeader={props.selectedHeader />
  <SidePaneContents>{props.children}</SidePaneContents>
  <PiPiP />
</SidePane>
```
(also other ways to structure this, just a quick suggestion, the key idea being the seperation of sidepane contents and the PiPiP so they are siblings and changes to one doesn't cause a re-render of the other)

This means when switching between people pane and chat pane, only the pane contents should refresh -- the PiPiP should remain in place and with no render changes

packages/react-composites/src/composites/CallWithChatComposite/EmbeddedChatPane.tsx
```diff
@@ -16,17 +18,34 @@ import { FileSharingOptions } from '../ChatComposite';
...
   fileSharing?: FileSharingOptions;
 }): JSX.Element => {
   const callWithChatStrings = useCallWithChatCompositeStrings();
+  const theme = useTheme();
+
+  const pipStyles: WithLocalAndRemotePIPStyles = useMemo(
+    () => ({
+      modal: {
+        main: {
+          top: '3.25rem',
```
JamesBurnside, line 40: For this:

> Bounds of Modal is not working when the starting position is set through styling such as { main: { top: '3.25rem' } }

Seems like a bad bug. Why do we need to set it for ChatPane and not for PeoplePane?

JamesBurnside (changes requested):

packages/react-composites/src/composites/CallWithChatComposite/EmbeddedChatPane.tsx
mgamis-msft: Talked with Alex P. and you are right there should only be one PiPiP. Simplified the pane logic similar to what you have proposed. Thanks.

packages/react-composites/src/composites/CallWithChatComposite/EmbeddedChatPane.tsx
mgamis-msft: No longer using the `main.top` style property so the [bad bug](https://github.com/microsoft/fluentui/issues/20122) shouldn't be an issue now.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -127,8 +126,7 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
 
   const [currentCallState, setCurrentCallState] = useState<CallState>();
   const [currentPage, setCurrentPage] = useState<CallCompositePage>();
-  const [showChat, setShowChat] = useState(false);
-  const [showPeople, setShowPeople] = useState(false);
+  const [activePane, setActivePane] = useState<'chat' | 'people' | 'none'>('none');
```
JamesBurnside, line 129: ❤️

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatPane.tsx
```diff
@@ -0,0 +1,153 @@
...
+  chatCompositeProps: Partial<ChatCompositeProps>;
+  callAdapter: CallAdapter;
+  chatAdapter: ChatAdapter;
+  onClose: () => void;
+  onFetchAvatarPersonaData?: AvatarPersonaDataCallback;
+  onFetchParticipantMenuItems?: ParticipantMenuItemsCallback;
+  onChatButtonClicked: () => void;
+  onPeopleButtonClicked: () => void;
+  modalLayerHostId: string;
+  activePane: 'chat' | 'people' | 'none';
```
JamesBurnside, line 38: nit `'chat' | 'people' | 'none'` is listed in a couple places, consider making a type 
`type CallWithChatPaneOptions = 'none' | 'chat' | 'people'`

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatPane.tsx
```diff
@@ -0,0 +1,153 @@
...
+      styles={paneStyles}
+      data-ui-id={
+        props.activePane === 'chat' ? 'call-with-chat-composite-chat-pane' : 'call-with-chat-composite-people-pane'
+      }
+    >
+      {header}
+      <Stack.Item verticalFill grow styles={paneBodyContainer}>
+        <Stack horizontal styles={scrollableContainer}>
+          <Stack.Item verticalFill styles={scrollableContainerContents}>
+            <Stack styles={props.activePane === 'chat' ? availableSpaceStyles : hiddenStyles}>
```
JamesBurnside, line 90: nit, the body of this is v. long and the nesting makes it hard to visualize, might be worth splitting components out and having something like:
```
<Stack>
  {header}
  <Stack.Item>
    <Stack>
      <Stack.Item>
        {chatPane}
        {peoplePane}
     </Stack.Item>
    </Stack>
   </Stack.Item>
   {modalPiPiP}
</Stack> 
```

So it's easier to visualize how these components live relative to each other

JamesBurnside (changes requested):
Everything looks great! Thanks for making those layout changes. Some non-blocking minor nits, but then one blocking item: the Desktop side pane looks too wide:
image

mgamis-msft:
> Everything looks great! Thanks for making those layout changes. Some non-blocking minor nits, just one blocking item: the Desktop side pane looks too wide: image

Thanks for spotting that. Fixed.

JamesBurnside (approved):
One small regression, small amount of padding needed between the PivotHeader and the contents:
image
Otherwise looks great!

packages/react-composites/src/composites/CallWithChatComposite/ModalLocalAndRemotePIP.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const localAndRemotePIP = useMemo(
+    () => <LocalAndRemotePIP {...pictureInPictureProps} {...pictureInPictureHandlers} />,
+    [pictureInPictureProps, pictureInPictureHandlers]
+  );
+
+  const modalStylesThemed = concatStyleSets(modalStyle, props.styles?.modal);
+
+  return (
+    <Stack styles={rootStyles}>
+      <Modal
```
dmceachernmsft, line 54: Maybe think about using ModalClone?

packages/react-composites/src/composites/CallWithChatComposite/PeoplePaneContent.tsx
```diff
@@ -150,70 +110,35 @@ export const EmbeddedPeoplePane = (props: {
...
+    <Stack tokens={peoplePaneContainerTokens}>
+      {inviteLink && (
+        <DefaultButton
+          text={strings.copyInviteLinkButtonLabel}
+          onRenderIcon={() => <LinkIconTrampoline />}
+          onClick={() => copy(inviteLink)}
+          styles={copyLinkButtonStylesThemed}
+        />
+      )}
+      {participantList}
```
dmceachernmsft, line 140: Why is this not wrapped in a stack?

dmceachernmsft (approved):

change/@internal-react-composites-d8da4cf2-00e1-48b9-84d0-0e7dfcd0f453.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
prprabhu-ms, line 2: In this case it doesn't matter too much (since we're going to merge back into release/1.2.0).

This would be a 'minor' change in my opinion even though there is no API change because it is a large UX change.

packages/react-composites/src/composites/CallWithChatComposite/BasicHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT license.
+import { CommandBarButton, Stack } from '@fluentui/react';
+import { useTheme } from '@internal/react-components';
+import React from 'react';
+import { sidePaneHeaderStyles } from '../common/styles/ParticipantContainer.styles';
+
+/**
+ * @private
+ */
+export const BasicHeader = (props: { headingText: string; onClose: () => void }): JSX.Element => {
```
prprabhu-ms, line 11: nit: Adjectives like `Basic`, `Simple` etc are a code smell.

It doesn't tell me why this is `Basic` (for that I need to know that there exists a `TabHeader`, and this is the non-tab version).

`SidePaneHeader` would be better as this is the header used by the side pane.

packages/react-composites/src/composites/CallWithChatComposite/BasicHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const theme = useTheme();
+  const sidePaneCloseButtonStyles = {
```
prprabhu-ms, line 13: memoize.

packages/react-composites/src/composites/CallWithChatComposite/BasicHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+    root: { minWidth: '1.5rem', padding: 0 },
+    icon: { color: theme.palette.neutralSecondary },
+    iconHovered: { color: theme.palette.neutralSecondary },
+    iconPressed: { color: theme.palette.neutralSecondary }
+  };
+  return (
+    <Stack horizontal horizontalAlign="space-between">
+      <Stack.Item styles={sidePaneHeaderStyles}>{props.headingText}</Stack.Item>
+      <CommandBarButton styles={sidePaneCloseButtonStyles} iconProps={{ iconName: 'cancel' }} onClick={props.onClose} />
```
prprabhu-ms, line 22: memoize `iconProps`?

prprabhu-ms (approved):

packages/react-composites/src/composites/CallWithChatComposite/BasicHeader.tsx
mgamis-msft: It's a hardcoded value so I think there's no need to memoize.

packages/react-composites/src/composites/CallWithChatComposite/ModalLocalAndRemotePIP.tsx
mgamis-msft: Will do that in a separate PR because I would need to export `ModalClone` from 'react-composites'

PR #1619, opened 2022-03-09
`useAzureCommunicationCallWithChatAdapter`: Dispose adapter on unmount

prprabhu-ms:
# Why

Discovered bug while working on `useAzureCommunicationCallAdapter` -- if the parent component unmounts, we need to dispose the adapter.

# How Tested

`rush build`


dmceachernmsft (approved):

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -610,6 +610,21 @@ export const useAzureCommunicationCallWithChatAdapter = (
     [adapterRef, afterCreateRef, beforeDisposeRef, credential, displayName, endpoint, locator, userId]
   );
 
+  // Dispose any existing adapter when the component unmounts.
+  useEffect(() => {
+    return () => {
+      (async () => {
+        if (adapterRef.current) {
```
JamesBurnside, line 617: it's for reviews like these I love the javascript is single threaded 😅

PR #1622, opened 2022-03-09
Initial UI tests for filesharing in chat composite

anjulgarg:
# What
- Add tests for file upload button and file cards in sendbox 
- Fix file shraing UI styling based on Figma https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?node-id=4022%3A24730

# Why

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/chat/FileSharing.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  test('shows file cards for uploaded files', async ({ serverUrl, users, page }) => {
+    await page.goto(
+      buildUrl(serverUrl, users[0], {
+        useFileSharing: 'true',
+        uploadedFiles: JSON.stringify([
+          {
+            name: 'SampleFile.pdf',
+            extension: 'pdf',
+            url: 'https://sample.com/SampleFile.pdf'
```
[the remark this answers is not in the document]
anjulgarg: Unfortunately not. :/

change/@internal-react-components-dfd490b8-84bc-430e-bf0f-e32b4877904a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Update internal FileCard component to not show progress bar when progress is 0",
```
prprabhu-ms, line 3: nitty nitty nit: Shorter change comment means less editing later: `FileCard: Do not show progress bar when progress is 0`.

change/@internal-react-components-dfd490b8-84bc-430e-bf0f-e32b4877904a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
prprabhu-ms, line 2: This type should be `prerelease` since it changes production behavior of a prerelease feature.

packages/acs-ui-common/src/cssUtils.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ * @internal
+ */
+export const _pxToRem = (px: number): string => `${px / 16}rem`;
```
prprabhu-ms, line 7: Please document this as it will be used from other packlets (soft boundary)

prprabhu-ms: Oh, and also <3

packages/react-components/src/components/FileCard.tsx
```diff
@@ -49,13 +50,13 @@ export const FileCard = (props: FileCardProps): JSX.Element => {
   const { fileName, fileExtension, progress, actionIcon } = props;
   const theme = useTheme();
 
-  const showProgressIndicator = (): boolean => progress !== undefined && progress < 1;
+  const showProgressIndicator = (): boolean => progress !== undefined && progress > 0 && progress < 1;
```
prprabhu-ms, line 53: nit: Perf papercut warning: This can be a file-level private function. That will avoid creating a new closure on every render. Coming from a static languages background, I balk at how freely we create new closures, but I don't have a real understandig of how costly closures are in JavaScript.

packages/react-composites/tests/browser/chat/app/index.tsx
```diff
@@ -30,8 +32,11 @@ const userId = verifyParamExists(params.userId, 'userId');
 // Optional params
 const useFrLocale = Boolean(params.useFrLocale);
 const customDataModel = params.customDataModel;
+const useFileSharing = Boolean(params.useFileSharing);
+const uploadedFiles = params.uploadedFiles ? JSON.parse(params.uploadedFiles) : [];
 
 // Needed to initialize default icons used by Fluent components.
+initializeFileTypeIcons();
```
prprabhu-ms, line 39: So... we need to update our documentation to ask Contoso to do this as well? :(

I think we should export our own `initializeIcons` that does both of this then.

prprabhu-ms:
Two high-level comments:

* Could you split this PR into two (1) add ui tests; (2) change the styling etc. Then we can see the change caused by (2) in snapshot updates!
* The snapshots currently don't look right - (1) No file attach icon is visible when there should be one; (2) the snapshots with the attached files are blank. Perhaps you need to update the `waitForChatCompositeToLoad` function now?

packages/react-components/src/components/FileCard.tsx
anjulgarg: Makes sense. Updated the code.

packages/react-composites/tests/browser/chat/app/index.tsx
anjulgarg: Yes. I agree, time has come to export our own util function.
Maybe we can go one step further and allow our function to block the loading of composite before icons are loaded. To help with flaky tests.

anjulgarg:
> Two high-level comments:
> 
> * Could you split this PR into two (1) add ui tests; (2) change the styling etc. Then we can see the change caused by (2) in snapshot updates!
> * The snapshots currently don't look right - (1) No file attach icon is visible when there should be one; (2) the snapshots with the attached files are blank. Perhaps you need to update the `waitForChatCompositeToLoad` function now?

Feeling too lazy for the 1st one. :P 
Fixed the incorrect snapshots

packages/react-composites/tests/browser/chat/app/index.tsx
JamesBurnside: We should be able to remove initializeIcons() when we move away from using fluent CDN font icons also (so then goes back to just `initializeFileTypeIcons()`)

prprabhu-ms:
Need to stub out timestamps: 
image

Also, is the increased height of SendBox in the ComponentExamples snapshots intentional?

anjulgarg:
> Need to stub out timestamps: image
> 
> Also, is the increased height of SendBox in the ComponentExamples snapshots intentional? image

Good catch. Fixed :)

prprabhu-ms (approved):
Is the icon change intentional?
image

anjulgarg:
> Is the icon change intentional? image

Not caused by this PR

anjulgarg:
> Noticed that there are a lot of snapshots with a vertical scroll bar, is this intended? Desktop View:  Mobile view: 

You are looking at the old snapshots. New snapshots don't have vertical scroll bar ;)

edwardlee-msft (approved):

alkwa-msft (commented):
This looks OK from me but I just want to make sure we are removing the file upload button in the new snapshot? I thought the point was to have file attachments so we can use them in SH?

PR #1623, opened 2022-03-09
Create clone of fluent modal with fluent trap zone disabled

JamesBurnside:
# What
* Create clone of fluent modal with fluent trap zone disabled
 

# Why
Don't hate us but...

@dmceachernmsft and I debugged the fluent modal for some time and basically the FocusTrapZone inside it _needs_ to be disabled.

We have a ship-blocking MAS blocking bug where the fluent should be able to be moved with keyboard so we must allow focus on the modal -- with Fluent's current modal once it gets focus it never relinquishes it if you are using the keyboard.

This is a direct copy paste of the [Fluent Modal](https://github.com/microsoft/fluentui/blob/b7f17e976f9e058f39c9fce4f0f9bb6eb4dfa577/packages/react/src/components/Modal/Modal.base.tsx) and [Fluent Draggable Zone](https://github.com/microsoft/fluentui/blob/b7f17e976f9e058f39c9fce4f0f9bb6eb4dfa577/packages/react/src/utilities/DraggableZone/DraggableZone.tsx) but with the FocusTrapZone `disabled` prop set to `true`.

## Is there a better fix?

Ideally Fluent provides a fix for: https://github.com/microsoft/fluentui/issues/18924 and when they do we should 100% remove this. However a fix (+package release with fix) will not be ready in time for shipping call with chat composite.

Very open to creative ideas on how else we can solve this

# How Tested
Can successfully tab in and out of modal, as well as drag it etc as normal.


prprabhu-ms:
<Prathmesh slowly backs away into the hedge>

prprabhu-ms (approved):

JamesBurnside:
> Do we have a bug filed to clean this up if/when we can get traction with fluent upstream? 

Good call, filed an internal bug: https://skype.visualstudio.com/SPOOL/_workitems/edit/2787059

dmceachernmsft (approved):

PR #1625, opened 2022-03-10
Add error bar to show file download error message in the chat screen

prabhjot-msft:
# What
Add error bar to show file download error message in the chat screen

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2681601

# How Tested
Screen Shot 2022-03-09 at 5 12 20 PM

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
```diff
@@ -54,6 +58,9 @@ export const FileDownloadCards = (props: FileDownloadCards): JSX.Element => {
           setShowSpinner(false);
           if (response instanceof URL) {
             window.open(response.toString(), '_blank', 'noopener,noreferrer');
+          } else {
+            console.log(response);
```
alkwa-msft, line 62: do we really want to console.log here?

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -165,8 +167,21 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
+            fileDownloadErrorMessage={downloadErrorMessage}
+            /* @conditional-compile-remove(file-sharing) */
+            onDismissFileDownloadErrorMessage={() => {
+              setDownloadErrorMessage('');
+            }}
+            /* @conditional-compile-remove(file-sharing) */
             onRenderFileDownloads={(userId, message) => (
-              <FileDownloadCards userId={userId} message={message} downloadHandler={fileSharing?.downloadHandler} />
+              <FileDownloadCards
+                userId={userId}
```
alkwa-msft, line 178: for here can we simplify this since the prop has the same name as the value?

change/@internal-react-components-5c4e27c8-a1aa-4f57-a5fa-3176b703b9ad.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Add error bar to show file download error message in the chat screen.",
```
anjulgarg, line 3: `chat screen` for `react-components` package, there is no such thing as chat screen.
Let's use MessageThread or the appropriate component name that we updated.

packages/react-components/src/components/FileDownloadErrorBar.tsx
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
```
anjulgarg, line 1: @prprabhu-ms We have an ErrorBar component in already in our components.
Do you think we can re-use it for this?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -517,6 +518,16 @@ export type MessageThreadProps = {
    * @beta
    */
   onRenderFileDownloads?: (userId: string, message: ChatMessage) => JSX.Element;
+  /**
+   * Optional file download error message to render.
+   * @beta
+   */
+  fileDownloadErrorMessage?: string;
+  /**
+   * Optional callback to dismiss the file download error.
```
anjulgarg, line 527: Misleading comment. 
This callback runs when someone dismisses an error bar.

Comment should be something like
`Optional callback. Runs whenever a download error message is dismissed`

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -998,17 +1009,24 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
...
-            {onRenderJumpToNewMessageButton ? (
-              onRenderJumpToNewMessageButton({ text: strings.newMessagesIndicator, onClick: scrollToBottom })
-            ) : (
-              <DefaultJumpToNewMessageButton text={strings.newMessagesIndicator} onClick={scrollToBottom} />
-            )}
-          </div>
-        )}
+      <Stack grow>
+        {/* @conditional-compile-remove(file-sharing) */}
+        <FileDownloadErrorBar
```
anjulgarg, line 1014: This is not working. If it was, you would get an error for unused import of `FileDownloadErrorBar` :P
To remove this you would need to change it to 
```tsx
{
  <FileDownloadErrorBar ... </FileDownloadErrorBar>
}
```

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -517,6 +518,16 @@ export type MessageThreadProps = {
    * @beta
    */
   onRenderFileDownloads?: (userId: string, message: ChatMessage) => JSX.Element;
+  /**
```
anjulgarg, line 521: Conditionally remove this prop

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -517,6 +518,16 @@ export type MessageThreadProps = {
...
+   * Optional file download error message to render.
+   * @beta
+   */
+  fileDownloadErrorMessage?: string;
+  /**
+   * Optional callback to dismiss the file download error.
+   * @beta
+   */
+  onDismissFileDownloadErrorMessage?: () => void;
```
anjulgarg, line 530: Conditionally remove this prop

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -39,6 +39,7 @@ import { SystemMessage as SystemMessageComponent, SystemMessageIconTypes } from
 import { ChatMessageComponent } from './ChatMessage/ChatMessageComponent';
 import { useLocale } from '../localization/LocalizationProvider';
 import { isNarrowWidth, useContainerWidth } from './utils/responsive';
+import { FileDownloadErrorBar } from './FileDownloadErrorBar';
```
anjulgarg, line 42: Conditionally remove this import

packages/react-components/stable-review/react-components.api.md
```diff
@@ -609,6 +609,8 @@ export type MessageThreadProps = {
     onLoadPreviousChatMessages?: (messagesToLoad: number) => Promise<boolean>;
     onRenderMessage?: (messageProps: MessageProps, messageRenderer?: MessageRenderer) => JSX.Element;
     onRenderFileDownloads?: (userId: string, message: ChatMessage) => JSX.Element;
+    fileDownloadErrorMessage?: string;
```
anjulgarg, line 612: these props should not be part of the stable API

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
anjulgarg, line 62: No need to log it to console here.
Contoso returns this error so they already know about it.

packages/react-components/review/react-components.api.md
```diff
@@ -635,6 +635,8 @@ export type MessageThreadProps = {
     onLoadPreviousChatMessages?: (messagesToLoad: number) => Promise<boolean>;
     onRenderMessage?: (messageProps: MessageProps, messageRenderer?: MessageRenderer) => JSX.Element;
     onRenderFileDownloads?: (userId: string, message: ChatMessage) => JSX.Element;
+    fileDownloadErrorMessage?: string;
```
prprabhu-ms, line 638: This API seems to be scoped too tight -- If we want to add a way for `MessageThread` to show an error, it should be agnostic of the error being shown (i.e., `errorBarMessage` and `onDismissErrorBar`).

Even then, this feature doesn't seem to fit a `MessageThread`. Is there any reason that the `ChatComposite` can't show a `MessageBar` on top of the `MessageThread` (just like we show varius banners for call recording, `ErrorBar` for errors from ACS services etc.)

packages/react-components/src/components/FileDownloadErrorBar.tsx
prprabhu-ms: I thought about that a bit. I don't think that this is a good fit for `ErrorBar`.

`ErrorBar` does two things: (1) Holds business logic for how to interpret ACS service errors (via the selector) (2) Provide localized strings for shown errors.

In this case, we want only (2), but (1) is fundamental to the purpose of `ErrorBar`.

That said, see my comment above. I think we should use a `MessageBar` directly in `ChatComposite` with a localized string instead of piping error messages into `MessageThread`. (So this component will move to `react-composites`)

prprabhu-ms (commented):
Let's discuss the high-level comments first. I'll take a more detailed look later.

packages/react-components/src/components/MessageThread.tsx
prprabhu-ms: This is a fairly big gotcha. @prabhjot-msft can you also update https://github.com/Azure/communication-ui-library/blob/main/packages/acs-ui-common/src/conditional-compilation-sample/index.tsx#L155 with how *not* to remove JSX elements

(Separate PR if you do that) 

Thanks.

change/@internal-react-components-5c4e27c8-a1aa-4f57-a5fa-3176b703b9ad.json
prabhjot-msft: Done

packages/react-components/src/components/FileDownloadErrorBar.tsx
prabhjot-msft: Done

packages/react-components/src/components/MessageThread.tsx
prabhjot-msft: Done

packages/react-components/stable-review/react-components.api.md
prabhjot-msft: Done

packages/react-composites/src/composites/ChatComposite/FileDownloadCards.tsx
prabhjot-msft: Done

change/@internal-react-components-5c4e27c8-a1aa-4f57-a5fa-3176b703b9ad.json
```diff
@@ -0,0 +1,7 @@
+{
```
anjulgarg, line 1: Remove this file. We are not changing anything in the react-components package anymore

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -160,13 +166,29 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
       <Stack className={chatArea} tokens={participantListContainerPadding} horizontal grow>
         <Stack className={chatWrapper} grow>
           {options?.errorBar !== false && <ErrorBar {...errorBarProps} />}
+          {
+            /* @conditional-compile-remove(file-sharing) */
+            <FileDownloadErrorBar
```
anjulgarg, line 171: nit: Slightly more terse code

```tsx
<FileDownloadErrorBar
  onDismissDownloadErrorMessage={() => {
    setDownloadErrorMessage('');
  }}
  fileDownloadErrorMessage={downloadErrorMessage || ''}
/>
```

anjulgarg (approved):

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
prabhjot-msft: done

change/@internal-react-components-5c4e27c8-a1aa-4f57-a5fa-3176b703b9ad.json
prabhjot-msft: Done

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
```diff
@@ -0,0 +1,43 @@
...
+  fileDownloadErrorMessage?: string;
+}
+
+/**
+ * @private
+ */
+export const FileDownloadErrorBar = (props: FileDownloadErrorBarProps): JSX.Element => {
+  const { fileDownloadErrorMessage, onDismissDownloadErrorMessage } = props;
+  const messageBarIconProps = (): IIconProps | undefined => {
+    const iconName = { iconName: 'ProtectedDocument' };
```
prprabhu-ms, line 23: Add a uniquely named icon to `ChatComopsiteIcons` so that Contoso can override the icon shown.

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import { IIconProps, MessageBar, MessageBarType, Stack } from '@fluentui/react';
+import React from 'react';
+
+/**
+ * @private
+ */
+export interface FileDownloadErrorBarProps {
+  /**callback to dismiss the download error message */
+  onDismissDownloadErrorMessage?: () => void;
```
prprabhu-ms, line 12: This prop should be required. In that case you also don't need to check that this prop exists in the impl below.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -160,13 +166,29 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
+              onDismissDownloadErrorMessage={() => {
```
prprabhu-ms, line 172: Memoize this closure with `useCallback` to avoid rendering `FileDownloadErrorBar` on each render pass.

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
```diff
@@ -0,0 +1,43 @@
...
+  /** Error message to render */
+  fileDownloadErrorMessage?: string;
```
prprabhu-ms, line 14: This error should be required (in trivial cases, it is set to `''`)

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * @private
+ */
+export const FileDownloadErrorBar = (props: FileDownloadErrorBarProps): JSX.Element => {
+  const { fileDownloadErrorMessage, onDismissDownloadErrorMessage } = props;
+  const messageBarIconProps = (): IIconProps | undefined => {
```
prprabhu-ms, line 22: Why is this is a callback (It's called on line 34). This can just be a file-private constant since the value never changes.

prprabhu-ms (commented):
Thanks for moving the logic to composites. Remaining comments are of a smaller scope.

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
prabhjot-msft: done

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
prabhjot-msft: cannot add this to ChatComopsiteIcons as it returns JSX.element, but we are passing messageBarIconProps i.e just the icon name

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
```diff
@@ -3,35 +3,34 @@
...
  * @private
  */
 export const FileDownloadErrorBar = (props: FileDownloadErrorBarProps): JSX.Element => {
   const { fileDownloadErrorMessage, onDismissDownloadErrorMessage } = props;
-  const messageBarIconProps = (): IIconProps | undefined => {
-    const iconName = { iconName: 'ProtectedDocument' };
-    return iconName;
-  };
-  if (fileDownloadErrorMessage && fileDownloadErrorMessage !== '') {
+  const messageBarIconProps = { iconName: 'ProtectedDocument' };
```
prprabhu-ms, line 22: This const will still create a new object on each render pass and cause MessageBar to be rerendered each time. I mean to add a file-level constant.

packages/react-composites/src/composites/ChatComposite/FileDownloadErrorBar.tsx
```diff
@@ -3,35 +3,34 @@
...
+
+  if (fileDownloadErrorMessage !== '') {
```
prprabhu-ms, line 24: Prefer early returns to reduce conditional nesting.

```ts

if (fileDownloadErrorMessage === '') {
  return <></>;
}
return (
  <Stack. ...
);
```

prprabhu-ms (approved):

PR #1628, opened 2022-03-10
Split button A11y fixes

dmceachernmsft:
# What

- add strings for button types -> `split button` OR `button`
  - this Follows Fluents example on their docs here: https://developer.microsoft.com/en-us/fluentui#/controls/web/button
- add default keyboard behavior by assigning a Keycode to the `menuTriggerKeyCode` prop

# Why

Allows for better narration around the split buttons, as well as better keyboard navigation.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2782073
# How Tested

Ran locally and tested with windows narrator. tested keyboard navigation behavior.


packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -137,6 +144,8 @@ export const ControlBarButton = (props: ControlBarButtonProps): JSX.Element => {
         onRenderIcon={props.onRenderIcon ?? DefaultRenderIcon}
         ariaLabel={props.ariaLabel ?? tooltipContent ?? labelText}
         allowDisabledFocus={props.allowDisabledFocus ?? true}
+        aria-roledescription={'split button'}
```
JamesBurnside, line 147: should this be being set here? (was this left in?)

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -170,6 +180,11 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
       menuProps={props.menuProps ?? generateDefaultDeviceMenuPropsTrampoline(props, strings)}
       menuIconProps={props.menuIconProps ?? !enableDeviceSelectionMenuTrampoline(props) ? { hidden: true } : undefined}
       split={props.split ?? enableDeviceSelectionMenuTrampoline(props)}
+      ariaDescription={
```
JamesBurnside, line 183: Can you explain more in the PR description what this is doing and why is it needed? Does fluent not provide the aria roles for their buttons already?

JamesBurnside: Wait -- is there a difference between aria-roledescriptiona and aria-description?

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -48,6 +48,16 @@ export interface CameraButtonStrings {
    * Tooltip of camera menu
    */
   cameraMenuTooltip: string;
+  /* @conditional-compile-remove(call-with-chat-composite) @conditional-compile-remove(control-bar-split-buttons) */
```
prprabhu-ms, line 51: These should only reference `control-bar-split-buttons` feature as the split buttons in `react-components` can be stabilized independently.

And any case, I think both of these features are now stabilized, so don't add conditional compilation at all: https://github.com/Azure/communication-ui-library/blob/main/common/config/babel/.babelrc.js#L30

prprabhu-ms:
I'm confused how this fixes keyboard navigation. Comments in the code explaining what affects keyboard navigation would be helpful.

packages/react-components/src/components/ControlBarButton.tsx
dmceachernmsft: oops will fix it!

packages/react-components/src/components/CameraButton.tsx
dmceachernmsft: they dont provide the explicit narration of 'split button' which I think would provide more context to say the drop down is attached to this button, will update for more clarity 😊

packages/react-components/src/components/CameraButton.tsx
dmceachernmsft: okay will remove these comments! 😊

packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -137,6 +144,8 @@ export const ControlBarButton = (props: ControlBarButtonProps): JSX.Element => {
         onRenderIcon={props.onRenderIcon ?? DefaultRenderIcon}
         ariaLabel={props.ariaLabel ?? tooltipContent ?? labelText}
         allowDisabledFocus={props.allowDisabledFocus ?? true}
+        aria-roledescription={props.ariaDescription}
```
JamesBurnside, line 147: why `aria-roledescription` over `aria-description` ?

packages/react-components/src/components/MicrophoneButton.tsx
```diff
@@ -166,6 +176,11 @@ export const MicrophoneButton = (props: MicrophoneButtonProps): JSX.Element => {
       menuProps={props.menuProps ?? generateDefaultDeviceMenuPropsTrampoline(props, strings)}
       menuIconProps={props.menuIconProps ?? !enableDeviceSelectionMenuTrampoline(props) ? { hidden: true } : undefined}
       split={props.split ?? enableDeviceSelectionMenuTrampoline(props)}
+      ariaDescription={
+        enableDeviceSelectionMenuTrampoline(props)
+          ? strings.microphoneButtonSplitRoleDescription
+          : strings.microphoneButtonDefaultRoleDescription
```
JamesBurnside, line 182:
```suggestion
         : undefined;
```

JamesBurnside (approved):

prprabhu-ms (approved):

PR #1630, opened 2022-03-10
Flatten pick and omit APIs in CallWithChatComposite

JamesBurnside:
# What
Flatten following CallWithChatComposite APIs that were using Pick and Omit:
* CallWithChatAdapterManagement
* CallWithChatAdapterUiState
* CallWithChatClientState
* CallWithChatControlOptions

# Why
Azure Review Board guidance -- doing so improves intellisense, improves doc generation and improves dev experience when looking at the API in the .d.ts file.

# How Tested
Type assertion file has been added to ensuring consistency across CallWithChat <--> Call & Chat interfaces


packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatAdapter.ts
```diff
@@ -26,55 +23,243 @@ import {
...
+  /**
+   * Join the call with microphone initially on/off.
+   *
+   * @param microphoneOn - Whether microphone is initially enabled
+   *
+   * @public
+   */
+  joinCall(microphoneOn?: boolean): Call | undefined;
+  /**
+   * Leave the call
```
prprabhu-ms, line 56: Add trailing period.

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatTypeAssertions.ts
```diff
@@ -0,0 +1,141 @@
...
+import { AdapterErrors } from '../../common/adapters';
+import { CallWithChatControlOptions } from '../CallWithChatComposite';
+import { CallWithChatAdapterUiState, CallWithChatClientState } from '../state/CallWithChatAdapterState';
+import { CallWithChatAdapterManagement } from './CallWithChatAdapter';
+/* @conditional-compile-remove(file-sharing) */
+import { FileUploadAdapter } from '../../ChatComposite';
+
+/// IMPORTANT
+///
+/// EACH INTERFACE DEFINED IN HERE IS THE PICK&OMIT VERSION OF THE FLATTENED INTERFACE WE EXPORT.
```
prprabhu-ms, line 24: Does the whole comment have to be in ALL CAPS? :)

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatTypeAssertions.ts
```diff
@@ -0,0 +1,141 @@
...
+): Required<CallWithChatAdapterManagementInternal> => value;
+
+CallWithChatAdapterManagementTypeAssertion;
+CallWithChatAdapterManagementRequiredTypeAssertion;
+
+/// CallWithChatControlOptions
+
+interface CallWithChatControlOptionsInternal
+  extends Pick<CallControlOptions, 'cameraButton' | 'microphoneButton' | 'screenShareButton' | 'displayType'> {
+  chatButton?: boolean;
```
prprabhu-ms, line 87: All the extra fields here are unnecessary because we are only asserting one-way inclusion -- that the publicly exported API correctly extends the `*Internal` version.

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatTypeAssertions.ts
```diff
@@ -0,0 +1,97 @@
...
+  value: Required<CallWithChatAdapterUiState>
+): Required<CallWithChatAdapterUiStateInternal> => value;
+
+CallWithChatAdapterUiStateTypeAssertion;
+CallWithChatAdapterUiStateRequiredTypeAssertion;
+
+/// CallWithChatClientState
+
+interface CallWithChatClientStateInternal
+  extends Omit<CallAdapterClientState, 'displayName' | 'endedCall' | 'latestErrors' | 'userId'> {}
```
prprabhu-ms, line 88: This brought up a real question. Why is `displayName` required in `CallWithChatComopsite`, but not in `CallComposite`?

prprabhu-ms: @JamesBurnside

prprabhu-ms, line 88: Omit style assertions are easier to maintain -- whenever we add a new field to the base type, it will break this assertion, and we need to explicitly Omit it if we intend to not add to CallWithChat.

prprabhu-ms (approved):

mgamis-msft (approved):

PR #1639, opened 2022-03-14
Fetch read receipt when participant join the chat, also not show read receipt when having more than 20 participants

carocao-msft:
# What
Fetch read receipt when participant join the chat, also not show read receipt when having more than 20 participants

# Why
When participant leave and rejoin the chat,should fetch the most up to date read receipt

# How Tested
When having more than 20 people in the chat, hide contextual menu read receipt
Get userid, threadid and tokens by using breakpoints in sample chat, put them in to storybook chat sample, check if you are getting the correct read information


packages/chat-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -73,20 +74,42 @@ export const createDefaultChatHandlers = memoizeOne(
...
           }
+
           // We have traversed all messages in this thread
           if (message.done) {
             isAllChatMessagesLoaded = true;
             break;
           }
         }
+        // keep fetching read receipts until read receipt time < earlist message time
+        let continueFetchingReadReceipt = true;
```
PorterNan, line 103: Embed continueFetchingReadReceipt condition into while(...)

PorterNan (approved):
Nice to see some changes from our snapshots, exciting to see we finally fix a long-existing bug!

One trap is this could introduce a raise condition for ui test that highly depends on which window is opened first (whether the message is read or not), would be nice to confirm if this scenario exists or not

packages/chat-component-bindings/src/handlers/createHandlers.ts
PorterNan:
```suggestion
        let readReceipt;
        while (readReceipt = readReceiptIterator.next()! && !readReceipt.done && parseInt(readReceipt?.value?.chatMessageId) >= earliestTime);
```

PorterNan: Don't forget to remove last "s" in readReceipts name :P

packages/chat-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -73,20 +74,36 @@ export const createDefaultChatHandlers = memoizeOne(
...
           messageIterator = chatThreadClient.listMessages({ maxPageSize: 50 });
         }
-
+        if (readReceiptIterator === undefined) {
+          readReceiptIterator = chatThreadClient.listReadReceipts();
+        }
+        // get the earliest message time
         let remainingMessagesToGet = messagesToLoad;
         let isAllChatMessagesLoaded = false;
+        let earliestTime = Number.MAX_SAFE_INTEGER;
```
dmceachernmsft, line 83: Why are we using this constant? maybe leave a comment as to why?

dmceachernmsft (approved):

PR #1640, opened 2022-03-14
Add new modal clone to release

dmceachernmsft:
# What

Add new modal to the release
# Why

because we want to test in bug bash

# How Tested

Ran locally to make sure the modal works appropriately in release.


prprabhu-ms:
Please make sure CI passes after https://github.com/Azure/communication-ui-library/pull/1642 is merged (to stop testing `beta` flavor entirely on this release branch).

anjulgarg:
Is this because we have a draggable modal for showing local preview?
I am thinking if it is worth having that modal...

dmceachernmsft:
> Is this because we have a draggable modal for showing local preview? I am thinking if it is worth having that modal...

This is to fix a number of A11y issues that are caused by the fluent trap zone found inside the modal. we will remove it when the issue is fixed on the fluent side since that is the best solution.

mgamis-msft (approved):

PR #1645, opened 2022-03-14
Bugfix for delay in removing file card after a message is sent

anjulgarg:
# What
Bugfix for delay in removing file card after a message is sent

# Why

# How Tested

# Process & policy checklist


edwardlee-msft (approved):

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -204,15 +204,17 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
...
-       * This also cleans the state for new file uploads with a fresh message.
+       * All the current uploads need to be clear from the state before a message has been sent.
+       * This ensures the following behavior:
+       * 1. File Upload cards are removed from sendbox at the same time text in sendbox is removed.
+       * 2. any component rendering these file uploads doesn't continue to do so.
+       * 3. Cleans the state for new file uploads with a fresh message.
        */
       this.fileUploadAdapter.clearFileUploads();
+
+      await this.handlers.onSendMessage(content, options);
```
prprabhu-ms, line 217: what happens if `onSendMessage` fails here?

anjulgarg: image

This method ultimately calls the stateful chat client send message and handles errors inside that method implementation.
We can not catch errors here and thus can't do anything if this fails

anjulgarg: Also, our current implementation clears the sendbox as soon as a message is sent. (Irrespective of it's success)
Hence, we should keep the logic same with file upload cards

alkwa-msft (approved):

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
anjulgarg: What would you suggest?

prprabhu-ms: Two thoughts here:
* This becomes more important when we add the "resend message" functionality. @PorterNan Should we continue to show that the files are attached (but not sent) @pereiralex for the UX.
* Currently, all the attached files will be lost and user will have to attach them again. This is not too different than what happens when a single file fails to upload (before sending message), right?

The "wish list" UX here would be - files remain attached, but greyed out. Then, if user resends the message, it is sent with the metadata at which point the already uploaded files show up in the message thread as usual.

anjulgarg: For this to work, we would also need to leave the message text in the sendbox.
So, makes sense to grey out and disable the entire sendbox while sendMessage is in flight?

PR #1651, opened 2022-03-14
File Sharing Code Cleanup

anjulgarg:
# What
- Remove duplicate file sharing files in react-composites package
- export file sharing components from `react-components` as internal components
- optimize file card filename truncation logic
- remove unneeded code
- Adds filecard component story in internal stories

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/FileUploadCards.tsx
```diff
@@ -30,7 +30,7 @@ export interface FileUploadCardsProps {
  * @beta
  */
 export const FileUploadCards = (props: FileUploadCardsProps): JSX.Element => {
-  const truncateLength = 15;
+  const truncateLength = 12;
```
prprabhu-ms, line 33: How are we chosing this number. Also, can't we instead use CSS `overflow:hidden text-overflow:ellipses overflow-wrap:anywhere` here?

anjulgarg: Oh yesss. Completely forgot about that CSS property Haha

prprabhu-ms (approved):
image

THEY LOOK THE SAME TO ME!

edwardlee-msft (approved):

PR #1168, opened 2021-11-25
Immediately fetch initial chat data in adapter (+fix customDataModel flakey UI test)

JamesBurnside:
# What
When creating chatAdapter, immediately call `chatAdapter->fetchInitialData()`

# Why
Otherwise you can create a chat composite with missing information
* Chat composite will load with no topic header (then suddenly populate it)
* Chat composite will load with no previous messages (then suddenly populate them)
* We have had flakey UI tests from this for over a month now and tripped us up - likely will trip up others.

## Caveats
* Chat adapter will now takes longer to construct as it has to wait for fetchInitialData to complete

# Fixes
Flakey ui test where previous messages hadn't loaded before screenshot was taken:
image

PR #1652, opened 2022-03-15
Fix race condition of "not in chat"

PorterNan:
# What

https://skype.visualstudio.com/SPOOL/_queries/query/8f724fc4-2b1b-4030-9471-9a64e1c8a7d0/

We are pulling the initial data when we are creating the chat adapter, which should not be part of adapter(adapter itself doesn't control when to call functions, someone external should do it)

remove the fetchInitialData() function call in creation of chat adapter, and we will highly depend on ChatScreen or external caller to do this (what we are currently doing already)

The above error is happening because:

1. we create callAdapter and chatAdapter when we are not part of the thread (the user hasn't been admitted yet)
2. createChatAdapter calls fetchInitialData, which throws "not in chat" error
3. the user is admitted to chat later when someone in teams click the button

This scenario means we should not make any request when we are creating adapter(we should not assume we have all the permission), we should depend on external caller to make the final request

# How Tested

# Process & policy checklist


PorterNan:
One more interesting topic to be brought up, looks like we are adding this to avoid flaky tests(which is working pretty well), and now we might need some other way to work this out 
https://github.com/Azure/communication-ui-library/pull/1168

PorterNan:
image
Looks like the sendBox style is auto adjusted here (because of delay loading of chat title), which more like a right behavior as non-localized version, updated the ui snapshot

image

anjulgarg:
image

Icon missing

mgamis-msft (approved):

PorterNan:
> I don't understand why this PR fixes the issue -- why are we sure that `ChatScreen` will only render _after_ the user is admitted to the chat thread?


We can't, but at least it leaves you the flexibility to do this, and our meeting composite needs this gap not to fetch the data when chat adapter is created

prprabhu-ms (approved):

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -418,7 +418,6 @@ export const createAzureCommunicationChatAdapter = async ({
   await chatClient.startRealtimeNotifications();
 
   const adapter = await createAzureCommunicationChatAdapterFromClient(chatClient, chatThreadClient);
-  await adapter.fetchInitialData();
```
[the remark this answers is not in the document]
PorterNan: Yep, ideally this should be done by composite instead of adapter(And we assume in most case adapter will be used with composite) - if people are using adapter directly, they will need to simulate what composite is doing, including this fetchInitial data

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -121,6 +121,7 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
   const adapter = useAdapter();
 
   useEffect(() => {
+    // Initial data should be always fetched by the composite(or external caller) instead of the adapter
     adapter.fetchInitialData();
```
alkwa-msft, line 125: (feel free to disregard)
thoughts on "loadChatContent"?

"initial" is tough because it denotes some sense of "time" like only once and at the beginning.
Do we call this once at the beginning or can we call this at  anytime?

PorterNan: We only call this once when we want to start the chat, and it wouldn't be called again (there is fetch message function to serve this purpose)

PR #1668, opened 2022-03-15
Hide `TabHeader` buttons in `CallWithChatComposite` on mobile when chat/people are hidden

prprabhu-ms:
# What

When the chat or people button are hidden via `CallWithChatCompositeProps.options.callControls`, we now hide the tab headers shown for switching between people and chat panes on `mobileView`.

# Why

The intention of hiding those buttons is to hide the way to open the people and chat pane, so we hide all the UX that opens those tabs.

# How Tested

https://user-images.githubusercontent.com/82062616/158490919-7744edfb-e0db-4f1f-85ee-7e683bd2c6f5.mp4

### DIFF
--- a/change/@internal-react-composites-d0c51282-4a08-46b4-bb38-706d6319af52.json
+++ b/change/@internal-react-composites-d0c51282-4a08-46b4-bb38-706d6319af52.json
@@ -1,0 +1,10 @@
+{
+  "type": "patch",
+  "comment": {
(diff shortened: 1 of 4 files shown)


mgamis-msft:
I'm good with this unless Alex P. says otherwise.

mgamis-msft (approved):

dmceachernmsft (approved):

PR #1669, opened 2022-03-16
wait for tooltips to show in CallComposite tests

dmceachernmsft:
# What

adds wait on the tooltips aria-describedby label to show up before proceeding with the test
# Why

waits for the tool tip to appear since this is the new default state of the app when it runs since keyboard focus is no longer getting stuck in the modal.

# How Tested

ran tests locally and verified generated snapshots.


packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -110,8 +110,8 @@ test.describe('Call Composite E2E CallPage Tests', () => {
 
       await page.goto(buildUrl(serverUrl, user));
       await waitForCallCompositeToLoad(page);
-      // Move the mouse off-screen to avoid tooltips / delayed focus from introducing flake in the snapshots
-      await page.mouse.move(0, 0);
+      // we need to dismiss any tool tips if there are any so we press the keyboard.
+      await page.keyboard.down('Escape');
```
prprabhu-ms, line 114: Shouldn't this be https://playwright.dev/docs/api/class-keyboard#keyboard-press

Also, isn't this racy: The escape might be pressed before the tooltip shows, and then fail to dismiss the tooltip later?

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -110,8 +110,8 @@ test.describe('Call Composite E2E CallPage Tests', () => {
 
       await page.goto(buildUrl(serverUrl, user));
       await waitForCallCompositeToLoad(page);
-      // Move the mouse off-screen to avoid tooltips / delayed focus from introducing flake in the snapshots
-      await page.mouse.move(0, 0);
+      // we need to dismiss any tool tips if there are any so we press the keyboard.
```
prprabhu-ms, line 113: Stale comment.

dmceachernmsft: oops! will remove.

edwardlee-msft (approved):

dmceachernmsft:
Closing this because unsure this fixes this. the end result is not needed anymore. will look into why the tooltips appeared in the release branch and not in main.

PR #1670, opened 2022-03-16
add escape key press to dismiss tooltips.

dmceachernmsft:
# What

adds test fix that clears tooltips
# Why

allows for consistent start state of the call page for testing

# How Tested

ran locally


alkwa-msft (approved):

prprabhu-ms:
iiuc, we don't need this anymore, right?

dmceachernmsft:
Closing because not needed will look into why the snapshots suddenly differed.

PR #1671, opened 2022-03-16
[CallWithChat] Left Align Participant pane items in desktop and mobile view

edwardlee-msft:
# What
Left align items in people pane/sidepane in desktop and mobile view

# Why
Many items in the people pane in mobile and desktop were not aligned
https://skype.visualstudio.com/SPOOL/_workitems/edit/2632003/

# How Tested

# Process & policy checklist
Tested locally on desktop and mobile view


prprabhu-ms (commented):
The context menu in desktop also got extra margins. Don't think that is intended: 
image

dmceachernmsft:
should you have the extra space between the participants in the CallWithChatComposite? 
image

edwardlee-msft:
> should you have the extra space between the participants in the CallWithChatComposite? image

Yes, I believe so. 
When looking at the figma, there should actually be space between participants in mobile view, at least in CallWithChat. (But that is out of scope for this PR).

dmceachernmsft (approved):

packages/react-components/src/components/styles/ParticipantList.styles.ts
```diff
@@ -8,8 +8,8 @@ import { ParticipantListItemStyles } from '../ParticipantList';
  * @private
```
anjulgarg, line 8: Unnecessary change?
Just flipping the order?

anjulgarg (approved):

PR #1673, opened 2022-03-16
Bugfix for file upload button and sendbox errors

anjulgarg:
# What
- Same file can be picked again using file upload button
- File Upload Errors contain message and timestamp
- SendBox Errors for file upload now use timestamp to choose which error to display instead of complex state logic

# Why
- Same file couldn't be picked earlier since it didn't fire onChange again
- Sendbox file upload errors had inconsistent behavior

# How Tested
Manually

https://user-images.githubusercontent.com/9782747/158499521-63b72adb-d27c-44d8-aa2f-2233f84abd56.mp4

# Process & policy checklist


packages/react-components/src/components/SendBox.tsx
```diff
@@ -288,15 +288,21 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
     [mergedSendIconStyle, onRenderIcon]
   );
 
+  /* @conditional-compile-remove(file-sharing) */
+  // Ensure that errors are cleared when there are no files in sendbox
+  React.useEffect(() => {
+    if (!props.activeFileUploads?.filter((upload) => !upload.error).length) {
```
prprabhu-ms, line 294: 1. Why do you need to filter out files that have upload errors? Is the user allowed to send the message if some of the files failed to upload?

2. You shouldn't need this `useEffect`. A change in `props.activeFileUploads` always causes a rerender, so instead you can change line 303 to be `containsPendingUpload(props.activeFileUploads) && fileUploadsPendingError`.

prprabhu-ms (approved):

packages/react-components/src/components/SendBox.tsx
anjulgarg: 1. Yes. Users are allowed to send message even if there are failed uploads. (Those uploads don't show in sendbox)
2. How do I set `setFileUploadsPendingError(undefined);` using your approach? `fileUploadsPendingError` will always be true

edwardlee-msft (approved):

packages/react-components/src/components/SendBox.tsx
prprabhu-ms: Ah right, you actually need to clear internal state whenever `activeFileUploads` becomes empty.

PR #1674, opened 2022-03-16
Fix PIPIP overlap in chat pane

dmceachernmsft:
# What

increase the z-index of the PIPIP in the chat and people panes
# Why

ensures that the PIPIP is above the items in the chat and people panes.
Before: 
image
After:
image

https://skype.visualstudio.com/SPOOL/_workitems/edit/2791126
# How Tested

Ran locally and visually inspected.


packages/react-composites/src/composites/CallWithChatComposite/ModalLocalAndRemotePIP.tsx
```diff
@@ -98,7 +98,7 @@ const modalStyle: IStyleFunctionOrObject<IModalStyleProps, IModalStyles> = {
     minWidth: 'min-content',
     minHeight: 'min-content',
     position: 'absolute',
-    zIndex: 1,
+    zIndex: 3, // set the z-index to keep the PIPIP above the other items in the chat and people panes.
```
prprabhu-ms, line 101: Magic numbers are hard to find. What's 1 and what's 2?

At the least I'd like us to add constants like:

```ts
MESSAGE_THREAD_Z_INDEX = 2;
PIPIP_MODEL_Z_INDEX = 3;
```

etc in a single place so that we know what the competition is like.

prprabhu-ms: More critically though, it seems like this zIndex competes with something outside `ModalLocalAndRemotePIP`. This introduces some non-local logic that is hard to understand and maintain.

We should instead set the zIndex via `prop.styles` in the component that contains both `ModalLocalAndRemotePIP` and the `MessageThread`. Essentially, try to make the competing zIndices be local to _some_ container component.

prprabhu-ms:
I'll refrain from approving this PR since I contributed to it.

carocao-msft (approved):

mgamis-msft (approved):

PR #1676, opened 2022-03-16
Resend Button for Chat

carocao-msft:
[](url)# What
Added resend button to contextual menu, will show up when message is failed to send

https://skype.visualstudio.com/SPOOL/_workitems/edit/2689329

https://user-images.githubusercontent.com/96077406/158687414-2fea7c87-5327-4c8d-bd4c-6da3da7fd3e1.mov

# How Tested
first comment out this line (setPage) in app.tsx
Screen Shot 2022-03-16 at 11 57 50 AM

Then use chrome devtools to block message sent request
Then unblock and click on resend button:

expected behavior: failed message is deleted and message is sent out successfully


anjulgarg:
@carocao-msft Please provide screenshots or video showcasing this change

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
```diff
@@ -48,12 +53,16 @@ export const ChatMessageComponent = (props: ChatMessageComponentProps): JSX.Elem
...
   const onEditClick = useCallback(() => setIsEditing(true), [setIsEditing]);
 
-  const { onDeleteMessage, message } = props;
+  const { onDeleteMessage, onSendMessage, message } = props;
   const onRemoveClick = useCallback(() => {
     if (onDeleteMessage && message.messageId) {
       onDeleteMessage(message.messageId);
     }
   }, [message.messageId, onDeleteMessage]);
+  const onResendClick = useCallback(() => {
```
alkwa-msft, line 62: (Potentially a small thing, not blocking)
i realize I am just looking at this ____Click function now its kind of wierd the on____ functions have Click at the end. Is this right?

alkwa-msft (approved):

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
carocao-msft: I think the function name is trying to say on click resend? I was just following the naming patterns of other existing functions (onremoveclick, onupdateclick ...) but yeah now you mentioned it, it does seems a bit weird

PorterNan:
image
Looks like we are not updating any ui here, but we only introduce wrong ui snapshots this time, let's revert the wrong snapshot update and remove ui change label (we don't have snapshot for contextual menu for now)

BTW, do we know why this one gets updated? Is it already wrong in our main branch?
image

PorterNan (approved):
It's exciting to see all functionalities are working - according to our figma, there is still something visual changes we might wanna introduce(including some color change and tooltip text change), let's do it in another PR

image

dmceachernmsft (approved):

PR #1686, opened 2022-03-17
Use single quotes in event expression literal

prprabhu-ms:
# What
Fix the syntax used for string literal in github action event.

# Why

String translation workflow is currently broken.

# How Tested

Prayers.

### DIFF
--- a/.github/workflows/run-td-build.yml
+++ b/.github/workflows/run-td-build.yml
@@ -34,4 +34,4 @@ jobs:
           azure-pipeline-name: 'Azure.communication-ui-library-touchdown-build'
           azure-devops-token: '${{ secrets.AZURE_TOUCHDOWN_BUILD_PIPELINE_SECRET }}'
           # Only commit updated files back when explicitly requested.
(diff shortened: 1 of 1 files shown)


.github/workflows/run-td-build.yml
```diff
@@ -34,4 +34,4 @@ jobs:
           azure-pipeline-name: 'Azure.communication-ui-library-touchdown-build'
           azure-devops-token: '${{ secrets.AZURE_TOUCHDOWN_BUILD_PIPELINE_SECRET }}'
           # Only commit updated files back when explicitly requested.
-          azure-pipeline-variables: '{"commit": ${{ github.event_name == "workflow_dispatch" && github.event.inputs.commit_strings }} }'
+          azure-pipeline-variables: "{'commit': ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.commit_strings }} }"
```
dmceachernmsft, line 37: Is the inversion here to help with readability?

dmceachernmsft (approved):

.github/workflows/run-td-build.yml
prprabhu-ms: No, github action event string literals can only use '' and not "".

edwardlee-msft (approved):

PR #1687, opened 2022-03-17
Use dummy telemetry tag on `main`

prprabhu-ms:
# What

* Set a dummy telemetry tag on `main`.
* Update github actions such that the tag is only updated to the correct value on branch creation.

# Why

We want to avoid polluting the telemetry data with events from our main CI. There is no way to do this via our package's API.
After this PR, the main CI, on `main` will report a dummy version. The actual version will still be reported from a CI on a release branch, but that is a short-lived branch in the run-up to a release, so the impact of events from there is far lower.


common/scripts/ensure-consistent-versions.js
```diff
@@ -46,16 +45,6 @@ function ensurePackageVersionsAreIdentical(packages, versions) {
     }
 }
 
-function ensureTelemetryVersionMatches(version) {
```
prprabhu-ms, line 49: Dropped this check so that `ensure-consistent-versions` can continue to be run as part of CI.

On release branches, the telemetry version is anyway always updated by another script right before we check.

packages/acs-ui-common/src/telemetryVersion.js
```diff
@@ -3,4 +3,9 @@
 
 // GENERATED FILE. DO NOT EDIT MANUALLY.
 
-module.exports = '1.2.1-beta.0';
+// The telemetry verision in development branches is
```
alkwa-msft, line 6: version

alkwa-msft (approved):

dmceachernmsft (approved):

PR #1691, opened 2022-03-18
Optimize Read receipt

carocao-msft:
# What
Optimize read receipt logic to improve performance
instead of calculating read receipt information in selectors, pass read receipt array down to component level and calculate read number/read list when hover the tooltip/ open the contextual menu


packages/communication-react/stable-review/communication-react.api.md
```diff
@@ -993,13 +993,6 @@ export interface ChatMessage extends MessageCommon {
     // (undocumented)
     mine?: boolean;
     // (undocumented)
-    readBy?: {
```
JamesBurnside, line 996: We will need to revert this, this breaks stable API

JamesBurnside (changes requested):

packages/communication-react/stable-review/communication-react.api.md
PorterNan: Hmmm... looks like it is already in 1.1.0, as long as we don't need it anymore, maybe we could just leave the api, but not put anything into it

PorterNan: I just revert this and still make it open to our customers, so for our customer they have 2 ways to render message readBy:
1. Provide an array of readReceipt for the whole messageThread, then the component will calculate each RR for each message and this is what we are doing (most optimized way)
2. Override RR for each single message by providing this property - not optimized in perf, but good to have this flexibility

JamesBurnside: We could also mark the existing fields as deprecated and point them to the new location (but we'll need to get populating the fields until a breaking change)

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -171,19 +164,12 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
...
-            readReceiptForEachSender
-          );
-
           return memoizedFn(
             message.id ?? message.clientMessageId,
             message,
             userId,
             message.createdOn <= latestReadTime,
-            isLargeGroup,
-            readBy.length,
```
JamesBurnside, line 185: We will still need to populate these fields even if we do not use them to prevent any breaking change (will be breaking if a customer relies on this prop but it is suddenly undefined)

carocao-msft: > We will still need to populate these fields even if we do not use them to prevent any breaking change (will be breaking if a customer relies on this prop but it is suddenly always undefined)
readBy.length and readBy are two fields I added when building read by information, just removing them now so it will not introduce new perf regression

packages/communication-react/stable-review/communication-react.api.md
JamesBurnside: Actually wait, I'm getting very confused.
In our APIView for 1.1.0 we had them sign off on readBy and readNumber: see `export interface ChatMessage MessageCommon` in https://apiview.dev/Assemblies/Review/42fe51ace1e24184a42351ff89044352?revisionId=174cc81091854531884d2326cbc15d6c&doc=False

**However**, I do not see that in our API in our release version: https://github.com/Azure/communication-ui-library/blob/d523bb3380d88e3097ba73864b8f998c4e3db822/packages/communication-react/review/communication-react.api.md?plain=1#L811

Did we submit the wrong file to review somehow in 1.1.0?

packages/chat-component-bindings/src/messageThreadSelector.ts
JamesBurnside: I am getting confused on an issue where it looks like we got sign off on something incorrect: https://github.com/Azure/communication-ui-library/pull/1691#discussion_r837937832

Hold off on this comment for a bit 😅

packages/communication-react/stable-review/communication-react.api.md
PorterNan: Hmmm... Look like so, I was a little confused by that as well cause we didn't have this feature in 1.1.0

PorterNan: If this is the case, go ahead and remove this part @carocao-msft

PorterNan: Let's remove this since this api wouldn't be needed in any following stable version, and let's grab a new quick approval from ARB team if necessary ( should be good cause we just take an api back, which doesn't actually exist in previous version)

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -193,7 +179,8 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
       userId,
       showMessageStatus: !isLargeGroup,
       messages: convertedMessages,
-      messageThreadParticipantCount
+      messageThreadParticipantCount,
```
PorterNan, line 182: Check if this name exists in 1.1.0 api file, if not, please rename it to participantCount, cause this is messageThreadSelector, so the name prefix here is a little redundant

packages/communication-react/review/communication-react.api.md
```diff
@@ -1879,7 +1879,16 @@ export const MessageStatusIndicator: (props: MessageStatusIndicatorProps) => JSX
 
 // @public
 export interface MessageStatusIndicatorProps {
+    // (undocumented)
+    message?: ChatMessage;
     messageThreadReadCount?: number;
+    // (undocumented)
+    readReceiptForEachSender?: {
```
PorterNan, line 1886: Can we pass in a tooltip string instead of this? readReceiptsForEachSender sounds like we are passing too much info for just one indicator

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -909,27 +936,44 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
...
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [new Date().toDateString(), isNarrow, messageThreadParticipantCount, onRenderAvatar]
+    [
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      new Date().toDateString(),
+      isNarrow,
+      messageThreadParticipantCount,
+      onRenderAvatar,
+      onActionButtonClickMemo,
+      props.onRenderFileDownloads,
```
JamesBurnside, line 952: ❤️ for updating this prop list!

packages/communication-react/review/communication-react.api.md
```diff
@@ -1879,7 +1879,16 @@ export const MessageStatusIndicator: (props: MessageStatusIndicatorProps) => JSX
 
 // @public
 export interface MessageStatusIndicatorProps {
+    // (undocumented)
+    message?: ChatMessage;
```
PorterNan, line 1883: I guess we could do the calculation out side of MessageStatusIndicator in MessageThread, message object seems too big for the indicator

JamesBurnside (approved):

packages/react-components/src/components/MessageStatusIndicator.tsx
```diff
@@ -127,6 +145,9 @@ export const MessageStatusIndicator = (props: MessageStatusIndicatorProps): JSX.
                   remoteParticipantsCount: `${remoteParticipantsCount}`
                 })
           }
+          onTooltipToggle={() => {
+            setIsToolTipHovered(!isToolTipHovered);
```
PorterNan, line 149: This is onToolTipsToggle instead of tooltips hover, the behavior is hover on the element to toghle the tooltip, not hover on the tool tip

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -300,16 +301,17 @@ const memoizeAllMessages = memoizeFnAll(
...
+      readReceiptForEachSender: { [key: string]: { lastReadMessage: string; name: string } },
       messageThreadParticipantCount: number
     ) => JSX.Element,
     defaultChatMessageRenderer: (message: MessageProps) => JSX.Element,
     strings: MessageThreadStrings,
     _attached?: boolean | string,
     statusToRender?: MessageStatus,
-    messageThreadReadCount?: number,
     messageThreadParticipantCount?: number,
+    readReceiptForEachSender?: { [key: string]: { lastReadMessage: string; name: string } },
```
PorterNan, line 314: I am thinking of whether ForEachSender suffix is necessary or not... The type {string: {name, lastMessage}} indicates it is for sender ids, maybe name it directly as readReceipt or readReceiptByUserId (sender either means sender of the message or sender of the readReceipt, which is more confusing)

Also the property "name" is a little confusing, I would like to make it displayName so people know it is for user not something else

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -909,27 +936,44 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
...
   );
 
   const localeStrings = useLocale().strings.messageThread;
   const strings = useMemo(() => ({ ...localeStrings, ...props.strings }), [localeStrings, props.strings]);
 
   const defaultStatusRenderer: (
+    message: ChatMessage,
     status: MessageStatus,
-    messageThreadReadCount: number,
+    readReceiptForEachSender: { [key: string]: { lastReadMessage: string; name: string } },
```
PorterNan, line 964: This type shows up for so many times, we can give it a type like message, maybe name it ReadReceiptsBySenderId

packages/react-components/src/components/MessageThread.tsx
PorterNan: Sorry readReceiptsBySenderId

PorterNan: @prprabhu-ms @JamesBurnside as genius of naming, do you have any ideas of better name of this one?

packages/react-components/src/components/MessageThread.tsx
JamesBurnside: I like `ReadReceiptsBySenderId`! I would make the property match also
``` suggestion
    readReceiptsBySenderId: ReadReceiptsBySenderId
```

Or perhaps just:
``` suggestion
    readReceipts: ReadReceiptsBySenderId
```

packages/react-components/src/components/MessageThread.tsx
JamesBurnside: Yep I like
```
readReceipts: ReadReceiptsBySenderId
```

packages/chat-component-bindings/src/messageThreadSelector.ts
JamesBurnside: This is now resolved.

change/@internal-chat-component-bindings-49aec90d-63f1-4ce6-bf2f-3c241a0af248.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
prprabhu-ms, line 2: The change type here is wrong. All of these clearly affect the stable build.

change/@internal-react-composites-ec7e402b-767e-4419-b47a-0546059b1958.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "optimization",
```
prprabhu-ms, line 3: I'd recommend adding a more informative comment. Especially because when we next release, we'll want to know "oh we already cherry-picked that one into the last release.

packages/chat-component-bindings/src/messageThreadSelector.ts
prprabhu-ms: This does not exist in `1.1.0`.

Let's rename.

packages/react-components/src/components/MessageThread.tsx
prprabhu-ms: Can we also remove the eslint-disable on line 946 now?

packages/react-components/src/components/MessageThread.tsx
carocao-msft: There is another variable named readreceipts in the selectors, how about naming it to readReceiptsBySenderId instead?

packages/chat-stateful-client/src/types/ReadReceiptsBySenderId.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ * An array of key value pairs that stores each sender's display name and last read message
+ *
+ * @public
+ */
+export type ReadReceiptsBySenderId = { [key: string]: { lastReadMessage: string; displayName: string } };
```
PorterNan, line 9: This type is not part of stateful client, the principle is statefulClient and components have their own types to keep themselves decoupled - if a type is shared between statefulClient and components then it is a stateful client type, otherwise it is only component type, move this to packages/react-components/src/types/

packages/react-components/src/components/MessageStatusIndicator.tsx
```diff
@@ -114,40 +109,23 @@ export const MessageStatusIndicator = (props: MessageStatusIndicatorProps): JSX.
...
+            readCount === 0 ||
             (remoteParticipantsCount && remoteParticipantsCount <= 1) ||
-            !messageThreadReadCount ||
+            !readCount ||
             !remoteParticipantsCount ||
             strings.readByTooltipText === undefined
               ? strings.seenTooltipText
               : _formatString(strings.readByTooltipText, {
-                  messageThreadReadCount: `${messageThreadReadCount}`,
+                  messageThreadReadCount: `${readCount}`,
```
PorterNan, line 125: I can still see perf issue here, and I would say even readCount and remoteParticipant count parameter isn't necessary, expose a toolTipText and onHover function will make this component more generic

PorterNan (approved):

PR #1693, opened 2022-03-18
bugfix: Track callId changes for in-flight video render operations

prprabhu-ms:
`callId` can change for a call. When this happens, we update internal maps that use `callId` as a key. but any asynchronous operations that already captured the old `callId` do not see this update.

This PR adds internal `callId` history to correctly handle the asynchronous operations that span the `callId` update.


PorterNan (commented):
💯, love this one. I also proposed a fix for this yesterday:
https://github.com/Azure/communication-ui-library/tree/jinan/stream-util-fix and also I add some kinda de-loop there. Great thoughts always are always similar!

If ids are flipping too fast, this will improve the scenarios a bit, make renderInfo into not found to 'rendered' state, maybe another fix need to be introduced to fix that rendered part

PorterNan (commented):
Just noticed another place needs something similar in our code for StreamUtils, let's have some chats after FHL (or in FHL)

prprabhu-ms:
The immediate problem was fixed with a server-side change. Longer term fix needs work beyond the UI library.

PR #1698, opened 2022-03-18
Memoize chat bubble to avoid unnecessary re-render

PorterNan:
# What

It's a perf regression that we lost the React.memo for chat bubble (It was there but somehow removed by accident, when we moved code around)
Use React.memo to ensure there is no re-rendering when no props changed for chat bubble

After apply this fix (these gray bricks are all solid before we apply this, and 1ms for each BOX rendering, now it is 0.1ms)
image

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -164,3 +164,6 @@ export const ChatMessageComponentAsMessageBubble = (props: ChatMessageComponentA
 
   return chatMessage;
 };
+
+/** @private */
+export const ChatMessageComponentAsMessageBubble = React.memo(MessageBubble);
```
alkwa-msft, line 169: should it be memoized based on its props?

alkwa-msft (approved):

dmceachernmsft (approved):

JamesBurnside (approved):

PR #1700, opened 2022-03-18
[FileSharing] - Resolving ARB Feedback

anjulgarg:
# What
- Abstracts away the implementation of file sharing and removes unneeded interfaces from public API
- Introduces methods in Chat Adapter for allowing contoso complete control of file sharing directly through adapter.
- Improves testability of file sharing features directly using adapter
- API Changes to make it more intuitive to use file sharing features

# Why
- Resolves a bunch of ARB comments
- Reduces public API interface
- Improved testability of code

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -308,8 +316,12 @@ export class AzureCommunicationCallWithChatAdapter implements CallWithChatAdapte
     return await this.chatAdapter.deleteMessage(messageId);
   }
   /* @conditional-compile-remove(file-sharing) */
-  public registerFileUploads(fileUploads: ObservableFileUpload[]): void {
-    this.chatAdapter.registerFileUploads(fileUploads);
+  public registerActiveFileUploads(files: File[]): FileUploadManager[] {
```
prabhjot-msft, line 319: why did we change the type of uploads from `ObservableFileUpload` to `File`

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -174,34 +167,34 @@ export class FileUpload implements FileUploadManager, ObservableFileUpload {
  * Events emitted by the FileUpload class.
  * @beta
  */
-export type FileUploadEvents = 'uploadProgressChange' | 'uploadComplete' | 'uploadFail';
+type FileUploadEvents = 'uploadProgressChange' | 'uploadComplete' | 'uploadFail';
```
prabhjot-msft, line 170: Don't we need to export these event types and listeners anymore?

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationFileUploadAdapter.test.ts
```diff
@@ -110,21 +106,12 @@ const createChatContext = (): ChatContext =>
   );
 
 /* @conditional-compile-remove(file-sharing) */
-const generateObservableFileUploads = (quantity: number): ObservableFileUpload[] => {
-  const fileUploads: ObservableFileUpload[] = [];
+const generateFiles = (quantity: number): File[] => {
```
JamesBurnside, line 109: super nit (feel free to ignore, this is a test file, but thought it was quite neat), you can remove the for loop and reduce this to one line using an array generator!
```ts
const generateFiles = (quantity: number): File[] =>
  Array.from({ length: quantity }, () => new File([], nanoid(), { type: 'text/plain' }));
```

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -96,41 +100,30 @@ export interface FileUploadManager {
 }
 
 /**
- * An internal interface used by the Chat Composite to drive the UI for file uploads.
- * @beta
+ * A wrapper object for a file that is being uploaded.
+ * Provides common functions for updating the upload progress, canceling an upload etc.
+ * @internal
```
JamesBurnside, line 105: If this is just used in this package mark as `private`, if its exported from this package but not publically keep as @internal but the name should be `_FileUpload`

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -96,41 +100,30 @@ export interface FileUploadManager {
...
- */
-export class FileUpload implements FileUploadManager, ObservableFileUpload {
-  private _emitter: EventEmitter;
-  public id: string;
-  public file: File;
-  fileName: string;
+  public metadata?: FileMetadata;
 
-  constructor(file: File, maxListeners = _MAX_EVENT_LISTENERS) {
+  constructor(data: { file?: File; metadata?: FileMetadata }) {
```
JamesBurnside, line 120: finding it hard to discern; will this ever be constructed with both a file and metadata?
If not can this instead be `constructor(uploadData: File | FileMetaData)` and the property be `public readonly fileUpload: File | FileMetaData` ?

JamesBurnside: Or alternatively, should `notifyUploadCompleted` be mutating and setting the metadata field?

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -96,41 +100,30 @@ export interface FileUploadManager {
...
-  id: string;
+export class FileUpload implements FileUploadManager, FileUploadEventEmitter {
+  private _emitter: EventEmitter;
+  public id: string;
+  public file?: File;
   /**
    * Filename to be displayed in the UI during file upload.
    */
-  fileName: string;
+  public fileName: string;
```
JamesBurnside, line 114: instead of a property I think this should be a getter helper function
`filename(): string | undefined =>  data.file?.name || data.metadata?.name;`

as currently there are three different fields that will give you the filename (file.name, metadata.name and fileName). Alternative if we combine file and metadata into one field called uploadData then filename would be myUploadFile.uploadData.name

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
anjulgarg: Observable File Upload was an un-necessary interface that exposed attributes in the public API that were of no use to end users.
By using an HTML interface `File` we keep the API simple and eliminate the need for devs to deal with custom interfaces

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
anjulgarg: Not anymore, no exported interface is using these types and hence we can keep them internal

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationFileUploadAdapter.test.ts
anjulgarg: Neat. Thanks

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
```diff
@@ -96,41 +100,30 @@ export interface FileUploadManager {
...
+ * @internal
  */
-export interface ObservableFileUpload extends FileUploadEventEmitter {
-  /**
-   * Unique identifier for the file upload.
-   */
-  id: string;
+export class FileUpload implements FileUploadManager, FileUploadEventEmitter {
+  private _emitter: EventEmitter;
+  public id: string;
```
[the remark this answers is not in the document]
anjulgarg: Nice. I didn't even know you could do that in ts. Thanks

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
anjulgarg: Simplified the constructor which in turn simplified it's underlying usage.
Check it out :)

packages/react-composites/src/composites/ChatComposite/file-sharing/FileUpload.ts
anjulgarg: Slight correction, `fileName` contains data from `file.name` OR `metadata.name` so there are 2 fields that give us filename.
Since `file.name` takes precedence over `metadata.name` initializing `fileName` with the correct data is sufficient.

prabhjot-msft (approved):

PR #1701, opened 2022-03-18
Dmceachernmsft/ContextMenuTestAlignmentFix

dmceachernmsft:
# What

Add sub icon style set
# Why

fixes alignment of the subicon in the flyout
Figma:
image
Code:
image

https://skype.visualstudio.com/SPOOL/_workitems/edit/2791165
# How Tested

Ran locally and compared against the figmas


JamesBurnside:
Can you post a screenshot of the result? (or add a UI test that captures this)

packages/react-components/src/components/styles/ChatMessageComponent.styles.ts
```diff
@@ -90,3 +90,13 @@ export const menuIconStyleSet = {
     width: '1.25rem'
   }
 };
+
+/**
+ * @private
+ */
+export const menuSubIconStyleSet = {
+  root: {
+    height: 'calc(100% - 10px)',
```
JamesBurnside, line 99: Why `- 10px`? Is 10px this the size of the current seen icon? Contoso is able to override icons so this fix may not scale well

dmceachernmsft: So the idea here is to use a similar offset to the main Icon to have it match in alignment. I agree that this will possibly cause issues with custom icons but if set to use the same styles as the main Icon for the context menu it appears slightly higher.

JamesBurnside (approved):

edwardlee-msft (approved):

PR #1704, opened 2022-03-22
2nd fix for call id, set right view element to context for client state

PorterNan:
# What

Similar to the fix for internal context, this fix corrects the right behavior for view element in internal context when there is asynchronous problem happening

Tested in local environment - Combine these 2 fixes together - we are good to facing rapid changing call id issue - the rendering perf is not optimized, but the behavior looks all correct

# Why

# How Tested

# Process & policy checklist


JamesBurnside (approved):

JamesBurnside:
(Looks like base branch PR was closed (#1693) but I would still complete this PR into the base branch in case we need to revive the PR

PorterNan:
Close this one cause we are going to make other PRs to fix this one

PR #1706, opened 2022-03-22
Prprabhu/esbuild todos

prprabhu-ms:
- Add path maps for monorepo
- Add Proxy server to redirect API calls.

SVG assets still aren't loading correctly, but I'm going to ignore that for now.


prprabhu-ms:
Hmm... need to figure out a way to not spam everybody as reviewers for these FHL PRs.

Sorry.

PR #1708, opened 2022-03-22
[DO NOT REVIEW] esbuild perf experiments

prprabhu-ms:
Please do not review this PR. It exists as a way for me to run CI as a play around with esbuild.


prprabhu-ms:
FHL be done

PR #1716, opened 2022-03-23
Local video tile Keyboard navigation fix

dmceachernmsft:
# What

Add a wrapper stack to the local video tile
# Why

This allows us to add a tab index to the local video tile making it keyboard movable in the new modal that we added.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2784616
https://skype.visualstudio.com/SPOOL/_workitems/edit/2784616
# How Tested

Ran locally in moved with keyboard interactions

https://user-images.githubusercontent.com/94866715/160441272-c959b314-b40f-41a0-8346-429da4504d59.mp4


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -259,28 +259,30 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
-        }
-        showLabel={!(shouldFloatLocalVideo && isNarrow)}
-        displayName={isNarrow ? '' : strings.localVideoLabel}
-        initialsName={localParticipant.displayName}
-        styles={localVideoTileStylesThemed}
-        onRenderPlaceholder={onRenderAvatar}
-        isMuted={localParticipant.isMuted}
-        showMuteIndicator={showMuteIndicator}
-      />
+      <Stack tabIndex={0}>
```
JamesBurnside, line 262: should this have a role or narration? (or what does narrator read out when it's focused on?)

JamesBurnside, line 262: I think at all times when something has focus there needs to be a change in this component that indicates this (typically a border being applied/made thicker). I can't find where something defines this but typically you are always able to spot when something has focus -- should we apply a border of some kind to this Stack (PIP) when it has focus?

dmceachernmsft: I was wondering this as well, I will get in touch with Alex P about it since it should probably have different display than a button since it has a hidden drop down? or should we just add the button selections to it?

dmceachernmsft: Good point, this should probably have a aria-label/role saying it has a drop down, will add that!

JamesBurnside: If you are adding strings can you do that as a seperate PR, we may be too late to add new string changes to get the localized in time (this should be a sev2 or sev3 issue so not crucial to fix this)

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -259,28 +259,35 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
-        initialsName={localParticipant.displayName}
-        styles={localVideoTileStylesThemed}
-        onRenderPlaceholder={onRenderAvatar}
-        isMuted={localParticipant.isMuted}
-        showMuteIndicator={showMuteIndicator}
-      />
+      <Stack
+        tabIndex={0}
+        aria-label={'Tile keyboard movement Dropdown Hidden'}
+        aria-role={'drop down'}
```
JamesBurnside, line 265: drop down doesn't highlight the whole role of this widget. It has a hidden drop down to allow moving the PiP but the PiP itself also shows useful information beyond the drop down.
Consider `diaglog` as the role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

The something in the aria-description should mention about the available dropdown to move the dialog about

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -259,28 +259,35 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
-        displayName={isNarrow ? '' : strings.localVideoLabel}
-        initialsName={localParticipant.displayName}
-        styles={localVideoTileStylesThemed}
-        onRenderPlaceholder={onRenderAvatar}
-        isMuted={localParticipant.isMuted}
-        showMuteIndicator={showMuteIndicator}
-      />
+      <Stack
+        tabIndex={0}
+        aria-label={'Tile keyboard movement Dropdown Hidden'}
```
JamesBurnside, line 264: Similar to comment on aria-role, this component is more than just the Dropdown. A screen reader hitting this should convey this is a movable tile that is showing the user their curent video&mic state, not just that there's a hidden dropdown to move it around.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -259,28 +259,35 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
+        aria-role={'drop down'}
+        aria-live={'polite'}
```
JamesBurnside, line 266: Just want to double check this is needed as the documentation implies to try avoid this (well to make sure it's used correctly):
image

What content is changed and read out inside this that needs the polite property?

packages/react-components/src/components/VideoGallery.tsx
JamesBurnside: (unless a11y info from the video tile child is accessible by the narrator somehow?)

packages/react-components/src/components/VideoGallery.tsx
dmceachernmsft: oh I will remove it, my understanding was that the narration of this control would be read out in standard order when scan mode is active but I will remove it based on this.

packages/react-components/src/components/VideoGallery.tsx
dmceachernmsft: Changed this label for now to: 'movable local video tile' happy to change it again if needed! 😊

packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -182,3 +191,18 @@ export const localVideoCameraCycleButtonStyles: IButtonStyles = {
...
   }
 };
+
+/**
+ * Styles for the local video tile modal when it is focused, will cause keyboard move icon to appear over video
+ * @private
+ */
+export const localVideoModalStyles: IModalStyles = {
+  keyboardMoveIconContainer: {
+    zIndex: 6 // zIndex to set the keyboard movement Icon above the other layers in the video tile.
```
JamesBurnside, line 201: always avoid raw zIndex values, this should be 
```
const keyboardMoveIconContainerZIndex = n + 1;
```
where n is an import from another file

dmceachernmsft: kk I think I know just the constant to use 😊

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -259,28 +259,30 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
-        }
-        showLabel={!(shouldFloatLocalVideo && isNarrow)}
-        displayName={isNarrow ? '' : strings.localVideoLabel}
-        initialsName={localParticipant.displayName}
-        styles={localVideoTileStylesThemed}
-        onRenderPlaceholder={onRenderAvatar}
-        isMuted={localParticipant.isMuted}
-        showMuteIndicator={showMuteIndicator}
-      />
+      <Stack tabIndex={0} aria-label={'Movable Local Video Tile'} role={'dialog'}>
```
JamesBurnside, line 262: @prprabhu-ms should we leave an english string or have no string here?

packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -182,3 +191,18 @@ export const localVideoCameraCycleButtonStyles: IButtonStyles = {
...
+  },
+  root: undefined,
```
JamesBurnside, line 203: Are you able to avoid assigning all these undefineds by changing the return type from `IModalStyles` to `Partial<IModalStyles>`

dmceachernmsft: yeah without the undefined sections of it the linter complains, we could try disabling the check for it?

JamesBurnside (approved):
Nice work on this fix!

packages/react-components/src/components/styles/VideoGallery.styles.ts
JamesBurnside: even with wrapping the return type in `Partial<>`? 
```ts
export const localVideoModalStyles: Partial<IModalStyles> = {
```

Where does it complain?

JamesBurnside: Otherwise try this type:
`IStyleFunctionOrObject<IModalStyleProps, IModalStyles>`

anjulgarg (approved):

packages/react-components/src/components/VideoGallery.tsx
edwardlee-msft: Would this be a better as a localized string? @dmceachernmsft

prprabhu-ms (approved):
Please remove the `aria-label` before merging.

packages/react-components/src/components/VideoGallery.tsx
dmceachernmsft: okay, I do have a follow PR up with the localized string @edwardlee-msft string so I will remove the string for the cherry pick.

PR #1717, opened 2022-03-23
A11y Chat pane keyboard navigation fix

dmceachernmsft:
# What

Add timeoout logic to force focus on chat composite send box once the chat pane is visible.
# Why

to allow better A11y behavior when opening the chat pane.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2790789
# How Tested

Ran locally and tested behavior.

https://user-images.githubusercontent.com/94866715/159813695-35673ea2-948b-4ab2-8913-b29ecaa35ae6.mp4


packages/react-composites/src/composites/CallWithChatComposite/CallWithChatPane.tsx
```diff
@@ -102,8 +103,8 @@ export const CallWithChatPane = (props: {
       <Stack.Item verticalFill grow styles={paneBodyContainer}>
         <Stack horizontal styles={scrollableContainer}>
           <Stack.Item verticalFill styles={scrollableContainerContents}>
-            <Stack styles={props.activePane === 'chat' ? availableSpaceStyles : hiddenStyles}>{chatContent}</Stack>
-            <Stack styles={props.activePane === 'people' ? availableSpaceStyles : hiddenStyles}>{peopleContent}</Stack>
+            {props.activePane === 'chat' && <Stack styles={availableSpaceStyles}>{chatContent}</Stack>}
```
JamesBurnside, line 106: The comment got lost through refactors but we previously hide the chat pane instead of not rendering it as an easy fix to prevent partially composed chat messages being lost (https://github.com/Azure/communication-ui-library/pull/662) -- we'll need to either find a different fix for maintaining partially composed messages or find a different way to trigger the autofocus.

I worry about not rendering the chat pane as quite a large change that could have potential regressions before right release.

I think this exposes that we really should have a function in the adapter that can set focus but the API design was tough so we settled on just the autofocus prop /:

JamesBurnside (changes requested):

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatPane.tsx
dmceachernmsft: Ahh I see, I did think it was interesting that it was just hidden and not rendered, so is the issue in that if it is de-rendered when messages come in it can lose some of the messages in the thread because the composite isn't present, and when it shows up it only grabs some of the messages in the thread?

JamesBurnside: Not that messages in the message thread is lost -- its that anything typed into the sendbox _but not sent yet_ is lost.
i.e. open chat pane > type a message (do not hit send) > close chat pane > reopen chat pane > typed message is gone when it shouldn't be

JamesBurnside (approved):

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -177,6 +177,10 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
       setActivePane('none');
     } else {
       setActivePane('chat');
+      // timeout is required to give the window time to render the sendbox so we have something to send focus to.
+      setTimeout(() => {
+        (document.querySelector(`[data-ui-id="sendbox-textfield"]`) as HTMLTextAreaElement)?.focus();
+      }, 10);
```
anjulgarg, line 183: How did we decide that 10 ms is enough to render sendbox?

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -177,6 +177,10 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
       setActivePane('none');
     } else {
       setActivePane('chat');
+      // timeout is required to give the window time to render the sendbox so we have something to send focus to.
+      setTimeout(() => {
+        (document.querySelector(`[data-ui-id="sendbox-textfield"]`) as HTMLTextAreaElement)?.focus();
```
prprabhu-ms, line 182: `data-ui-id` is *strictly* for e2e testing purposes. We should be able to change `data-ui-id` arbitrarily without breaking production code.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
prprabhu-ms: We shouldn't add non-deterministic behaviour to production code.

To me it seems like the `'chat'` pane should auto-focus when it is shown. Similar to how `SendBox` does it (b.t.w, doesn't this conflict with the auto-focus on `SendBox`?

prprabhu-ms (changes requested):
Two problems with the approach:

* Select HTML elements directly to modify 
* Using `data-ui-id` in production code

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
dmceachernmsft: Okay I will change how it is selected. 

so the major problem here is that the chat pane and people pane are hidden with just CSS right now. this is a issue because the autofocus option is dependent on rendering, so since the chat pane is only rendered once but hidden the autofocus prop does not work

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
dmceachernmsft: I can make it longer but the 10ms works out because its CSS changing. if I make it longer it dosen't focus right away and if I make it shorter it wont focus at all so 10ms worked out for it to appear natural.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
prprabhu-ms: So this is a case where we externally want to refocus on the SendBox after initial render.

This would be a good argument to simply implement `.focus()` on `ChatComposite` (and probably `CallWithChatComposite`), where `.focus()` does the right thing:
* For `ChatComposite`, it focuses on the `SendBox` text field
* For `CallWithChatComposite`, it focuses on `MuteButton` if `ChatPane` is hidden, and on `SendBox` if it is shown.

This would be a non-trivial feature though.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -177,6 +177,15 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
       setActivePane('none');
     } else {
       setActivePane('chat');
+      // timeout is required to give the window time to render the sendbox so we have something to send focus to.
+      // TODO: look into another way to select this/ adjust composite to select this better.
```
prprabhu-ms, line 181: Nit: could you word this comment to be more explicit: `Selecting elements in the DOM via attributes is not stable. We should expose an API from ChatComposite to be able to focus on the TextBox`.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -177,6 +177,15 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
...
+      const chatFocusTimeout = setInterval(() => {
+        const textBox = document.querySelector(`[id="sendbox"]`) as HTMLTextAreaElement;
+        if (textBox !== null) {
+          textBox.focus();
+          clearInterval(chatFocusTimeout);
```
prprabhu-ms, line 186: You also want an overall timeout after which you `clearInterval`. If, for some reason, the chat pane isn't shown at all, this will keep polling forever.

dmceachernmsft: will do!

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
dmceachernmsft: yup! will update

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
dmceachernmsft: should we make a user story around this?

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -177,6 +177,18 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
       setActivePane('none');
     } else {
       setActivePane('chat');
+      // timeout is required to give the window time to render the sendbox so we have something to send focus to.
+      // TODO: Selecting elements in the DOM via attributes is not stable. We should expose an API from ChatComposite to be able to focus on the TextBox.
+      const chatFocusTimeout = setInterval(() => {
+        const textBox = document.querySelector(`[id="sendbox"]`) as HTMLTextAreaElement;
```
JamesBurnside, line 183: nit: sendbox over textBox for variable name for contextual clarity of what textbox focus is being set (also comments are query refer to sendbox so its nice to keep the same name for consistency)

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
JamesBurnside: There's a bug that exists from when we implemented the original autofocus prop: [Bug 2672276: [Web] Contoso cannot call .focus() on composites or components such as sendbox](https://skype.visualstudio.com/SPOOL/_workitems/edit/2672276)

JamesBurnside (changes requested):
Selector narrowing

prprabhu-ms (approved):

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -172,13 +172,29 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
     setActivePane('none');
   }, [setActivePane]);
 
+  /** Constant setting of id for the parent stack of the composite */
+  const compositeParentDivId = 'callWithChatCompositeParentDiv-internal';
```
JamesBurnside, line 176: nit: use `root` instead of `parent`. To Contoso the parent of callwithchat is whatever they have put it in

JamesBurnside (approved):

PR #1718, opened 2022-03-24
update strings for modal

dmceachernmsft:
# What

Adds aria strings for localization for local video tile
# Why

Allows for proper narration announcing there are controls in the local video tile

# How Tested

Ran locally listening to the narrator.


alkwa-msft (approved):

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -137,6 +137,8 @@
     "screenIsBeingSharedMessage": "You are sharing your screen",
     "screenShareLoadingMessage": "Loading {participant}'s screen",
     "localVideoLabel": "You",
-    "localVideoCameraSwitcherLabel": "Cycle camera"
+    "localVideoCameraSwitcherLabel": "Cycle camera",
+    "localVideoMovementLabel": "Movable Local Video Tile",
+    "localVideoMovementRole": "Dialog"
```
JamesBurnside, line 142: Dialog is a standard aria role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role
Does it need localized?

change/@internal-react-components-6507a903-0818-4048-8755-ef850d53c2bf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add tab index to local video tile through a wrapper to allow keyboard navigation of the modal controls.",
```
JamesBurnside, line 3: After merging in from main this should just be updating the string as main already has the tabindex property set

JamesBurnside (approved):

mgamis-msft (approved):

PR #1722, opened 2022-03-25
Add aria role and narration priority to Back button [Mobile]

dmceachernmsft:
# What

Add aria-role and aria-live props to return to call button. subsequent PR coming with localization strings for aria-label
# Why

Helps with A11y interactions with screen readers

https://skype.visualstudio.com/SPOOL/_workitems/edit/2790265
# How Tested

Ran locally and inspected with screen reader - will test with team member and Iphone before completing the issue.


packages/react-composites/src/composites/CallWithChatComposite/TabHeader.tsx
```diff
@@ -56,6 +56,8 @@ export const TabHeader = (props: TabHeaderProps): JSX.Element => {
   return (
     <Stack horizontal grow styles={mobilePaneControlBarStyle}>
       <DefaultButton
+        aria-role={'button'}
+        aria-live={'polite'}
```
JamesBurnside, line 60: Why do we need aria live here? (Can you add a comment what content is being live updated that requires this?)

JamesBurnside (changes requested):
I believe you mentioned this issue was fixed instead by setting appropriate aria-labels instead. Should be able to close this

PR #1727, opened 2022-03-28
Ensure all required strings are localized

mgamis-msft:
# What

All required strings should not be optional so that we ensure that they are localized.

# Why

Ensure the code will NOT compile if a required string is missing in the en_US locale file so that we ensure that the required string is localized.

# How Tested

Not tested. Relying on CI.

# Process & policy checklist

### DIFF
--- a/change/@internal-react-components-37e07b23-05a9-4a5e-98b5-a97c055638e7.json
+++ b/change/@internal-react-components-37e07b23-05a9-4a5e-98b5-a97c055638e7.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "All component strings interface properties changed to required.",
+  "packageName": "@internal/react-components",
+  "email": "miguelgamis@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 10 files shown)


change/@internal-react-components-37e07b23-05a9-4a5e-98b5-a97c055638e7.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
JamesBurnside, line 2: nit: this should patch

packages/communication-react/review/communication-react.api.md
```diff
@@ -1254,20 +1254,20 @@ export interface ControlBarButtonProps extends IButtonProps {
...
 // @public
 export interface ControlBarButtonStrings {
-    label?: string;
-    offLabel?: string;
-    onLabel?: string;
-    tooltipContent?: string;
-    tooltipDisabledContent?: string;
-    tooltipOffContent?: string;
-    tooltipOnContent?: string;
+    label: string;
```
JamesBurnside, line 1264: This looks like a breaking change to the stable API (going from optional -> required is breaking)

JamesBurnside (changes requested):

packages/communication-react/review/communication-react.api.md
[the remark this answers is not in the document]
mgamis-msft: Will reopen this PR and bring up if this breaking change should be done after release.

PR #1728, opened 2022-03-28
Disable RR details in meeting interop

PorterNan:
# What

Disable RR details in meeting interop 

This is to not introduce problematic behavior into 1.2.0 when it is teams interop

Currently we can't get display name when it is meeting interop - because teams user are not joining using ACS, so the display name roster doesn't include displayName for teams user.

There are 2 solutions we might want to consider for future work:
1. Get display name from calling side when it is callWithChat component - no extra setup, but will only work for meeting scenarios
2. Provide our customers snippet on render custom display name and connect teams user name using AAD (extra set up and coding needed, but a more general solution)
Why

# How Tested

# Process & policy checklist

### DIFF
--- a/change/@internal-chat-component-bindings-82459451-1502-436b-bf85-94bc70a93fc2.json
+++ b/change/@internal-chat-component-bindings-82459451-1502-436b-bf85-94bc70a93fc2.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Disable RR in meeting interop",
+  "comment": "Disable Read Receipt details in meeting interop",
   "packageName": "@internal/chat-component-bindings",
   "email": "jiangnanhello@live.com",
   "dependentChangeType": "patch"
--- a/change/@internal-chat-component-bindings-82459451-1502-436b-bf85-94bc70a93fc2.json
+++ b/change/@internal-chat-component-bindings-82459451-1502-436b-bf85-94bc70a93fc2.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Disable RR in meeting interop",
+  "packageName": "@internal/chat-component-bindings",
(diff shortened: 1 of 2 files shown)


packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -136,11 +136,15 @@ const hasValidParticipant = (chatMessage: ChatMessageWithStatus): boolean =>
...
+
     // get number of participants
     // filter out the non valid participants (no display name)
-    const messageThreadParticipantCount = Object.values(participants).filter(
-      (p) => p.displayName && p.displayName !== ''
-    ).length;
+    // Read Receipt details will be disabled when participant count is 0
+    const messageThreadParticipantCount = isTeamsInterop
+      ? 0
+      : Object.values(participants).filter((p) => p.displayName && p.displayName !== '').length;
```
JamesBurnside, line 148: Further down on L178 the `readBy` will be incorrect as this also filters on `.name !== ''` inside `getParticipantsWhoHaveReadMessage`, this should also be set to undefined or 0.

PorterNan: Updated:D will update the cherry pick branch after merging this in

change/@internal-chat-component-bindings-82459451-1502-436b-bf85-94bc70a93fc2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Disable RR in meeting interop",
```
prprabhu-ms, line 3: Please fully spell out read receipt to ease changelog grooming later.

JamesBurnside (approved):

prprabhu-ms (approved):

PR #1736, opened 2022-03-30
Local video tile Keyboard navigation fix (#1716)

dmceachernmsft:
# What

update new modal tabs to activate keyboard navigation. add styles to new modal
# Why

allows for keyboard controls to move around the local video tile, and highlight when it has focus

https://skype.visualstudio.com/SPOOL/_workitems/edit/2790804
# How Tested

Ran locally - see video in this PR: https://github.com/Azure/communication-ui-library/pull/1716


JamesBurnside (approved):

packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -54,7 +54,16 @@ export const floatingLocalVideoModalStyle = (
     {
       main: localVideoTileContainerStyle(theme, isNarrow)
     },
-    { main: { boxShadow: theme.effects.elevation8 } }
+    {
+      main: {
+        boxShadow: theme.effects.elevation8,
+        ':focus-within': {
+          boxShadow: theme.effects.elevation16,
+          border: `2px solid ${theme.palette.neutralPrimary}`
```
alkwa-msft, line 62: do we want to use rem here instead?

alkwa-msft (approved):

packages/react-components/src/components/styles/VideoGallery.styles.ts
dmceachernmsft: updated this here to use rem, will add a task to update this on main as well.

PR #1738, opened 2022-03-30
CallWithChat new modal update styles

dmceachernmsft:
# What

Update new modal styles to use rem not px.
# Why

Matches our convention for using `rem` over `px`.

# How Tested

Ran locally and visually inspected.


JamesBurnside (approved):

packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -60,7 +60,7 @@ export const floatingLocalVideoModalStyle = (
         boxShadow: theme.effects.elevation8,
         ':focus-within': {
           boxShadow: theme.effects.elevation16,
-          border: `2px solid ${theme.palette.neutralPrimary}`
+          border: `0.125rem solid ${theme.palette.neutralPrimary}`
```
anjulgarg, line 63: there is a utility function `_pxToRem` in common package. (In case you want to use that :))

anjulgarg (approved):

PR #1740, opened 2022-03-30
Storybook: Update references to the correct @azure/communication-calling version

prprabhu-ms:
# Why

Our `peerDependencies` still refer to a specific headless SDK dependency, so our documentation should follow suite.


prprabhu-ms:
I think we still do need to include instructions for installing these peerDependencies. NPM v6 is still fairly common. Ubuntu LTS still installs v6 for example.

(NPM v7+ will auto install peer dependencies, so that these instructions are redundant)

JamesBurnside (approved):

alkwa-msft (approved):

PR #1746, opened 2022-03-31
Correctly set readBy count when `mobileView` is set

prprabhu-ms:
# What
image

# Why

Fixes a bug where we always showed that 0 remote participants have read a message.

# How Tested
Local `CallWithChat` sample.

### DIFF
--- a/packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
+++ b/packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
@@ -148,8 +148,13 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
             if (!wasInteractionByTouch) {
               return;
             }
-            props.onActionButtonClick(message, setMessageReadBy);
+            // If the message was touched via touch we immediately open the menu
+            // flyout (when using mouse the 3-dot menu that appears on hover
+            // must be clicked to open the flyout).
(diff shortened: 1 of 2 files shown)


carocao-msft (approved):

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -144,7 +144,13 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
           onTouchStart={() => setWasInteractionByTouch(true)}
           onPointerDown={() => setWasInteractionByTouch(false)}
           onKeyDown={() => setWasInteractionByTouch(false)}
-          onClick={() => wasInteractionByTouch && setChatMessageActionFlyoutTarget(messageRef)}
+          onClick={() => {
+            if (!wasInteractionByTouch) {
+              return;
+            }
+            props.onActionButtonClick(message, setMessageReadBy);
+            setChatMessageActionFlyoutTarget(messageRef);
```
JamesBurnside, line 152: I had to remember what was happening here... can you do better than me and add a comment haha?
```ts
// If the message was touched via touch we immediately open the menu flyout (when using mouse the 3-dot menu that appears on hover  must be clicked to open the flyout).
// In doing so here we set the target of the flyout to be the message and not the 3-dot menu button to position the flyout correctly.
```

JamesBurnside (approved):

PR #1748, opened 2022-04-01
ModalPiPiP in mobile CallWithChat set to higher position and fixed bounds.

mgamis-msft:
# What

- Initial y position of ModalPiPiP set higher.
- Imported ModalClone with changes from https://github.com/Azure/communication-ui-library/pull/1725 to react-composites so that ModalPiPiP bounds can be set properly when starting position of Modal is not at the center.
- Applying rtl property of CallWIthChatComposite

# Why

- https://skype.visualstudio.com/SPOOL/_workitems/edit/2791137 (ModalPiPiP should be higher so that the latest messages are not covered)
- ModalPiPiP can be dragged off screen.
- rtl property of CallWIthChatComposite is not applying

# How Tested

Tested on local CallWithChat sample:
https://user-images.githubusercontent.com/79475487/161246430-2b33c2ab-abb5-4472-827e-b4692fd4ab9a.mp4
With rtl:
https://user-images.githubusercontent.com/79475487/161247324-74496b2f-7478-43c8-b399-7ffe37f584d6.mp4

# Process & policy checklist


packages/react-composites/src/composites/CallWithChatComposite/CallWithChatPane.tsx
```diff
@@ -87,6 +90,7 @@ export const CallWithChatPane = (props: {
     </CallAdapterProvider>
   );
 
+  const modalLayerHost = document.getElementById(props.modalLayerHostId);
```
JamesBurnside, line 93: Can we instead use callWIthChatCompositeRootDiv.getElementById() instead of document?
If Contoso ever has more than one composite on the page this will break.
Don recently added a reference to this rootDiv to do the same thing

JamesBurnside (approved):

dmceachernmsft (approved):

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatPane.tsx
mgamis-msft: I tested this with two composites side by side and it works fine. 
https://user-images.githubusercontent.com/79475487/161810214-6cdb126f-2042-4cc0-94a8-940c50e4d316.mp4
This is because the id of the modal layer host is uniquely generated in the CallWithChat composite with `useId` from '@fluentui/react-hooks'
I could change reference the CallWithChat composite div but it would be another prop that needs to be added to CallWIthChatPane.

JamesBurnside: Oh nice! Nope what you have sounds good!

PR #1752, opened 2022-04-01
Add telemetry for rendering problems

PorterNan:
# What

Add telemetry for rendering problems

This will require change from https://skype.visualstudio.com/SCC/_git/internal_dogfood_self-hosted-e2e/pullrequest/612214 to get logs in selfhost app
# Why

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -34,21 +35,27 @@ async function createViewRemoteVideo(
   } else {
     participantKey = toFlatCommunicationIdentifier(participantId);
   }
+  const streamLogInfo = `callId: ${callId}, participantKey: ${participantKey}, streamId: ${streamId}`;
+
+  callingStatefulLogger.info(`Start creating view for remote video! ${streamLogInfo}`);
```
JamesBurnside, line 40: From @alkwa-msft 's comment on not impacting network bandwidth during calls -- when is this telemetry sent? Is there a way to batch-and-wait ?

JamesBurnside, line 40: Extra question (related to my other question) is this enabled by default for all customers?

PorterNan: This will only output things into console if you call logger.setLevel("info"), it wouldn't do anything without anyone call this line
And it will only be redirected into app insight once contoso app calls the connector api(currently only in selfhost app, and network bandwidth friendly implementation will be implemented there)

JamesBurnside: Nice!

JamesBurnside: Can you add something in our Docs > References somewhere that mentions this -- it's really hard to find info online about the logger

PorterNan: Sure, we will have more docs around this, and we've used this a lot already in state change. Before we check in any telemetry api into this repo, let's treat this as an internal conditional console.log because this is how it behaves for now.
And the behavior will only change once we have the logger connector api, and that would be the time we add docs, currently it would be a little confusing if we add anything (cause no api or implementation is provided)

JamesBurnside, line 40: nit: remove exclamation marks, particularly for non-errors as these aren't too unexpected

JamesBurnside (approved):

alkwa-msft:
I am curious how can we use these logs to help determine that something wrong is going on? Can you share a scenario about how we could see something odd is going on and how we can use the logs to visibly see this?

alkwa-msft (approved):

PR #1758, opened 2022-04-04
Build only storybook when deploying

prprabhu-ms:
# Why

On stable release branches, we test that storybook builds, but *do not test that the other packlets build in beta flavored build*.
When deploying, we can not be sure the other packages build.

# How Tested

Deployed storybook for `1.2.0`.

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -28,8 +28,8 @@ jobs:
       - name: Install Dependencies
         run: rush install
 
-      - name: Build Storybook dependencies
-        run: rush build -v -T storybook
+      - name: Build Storybook
+        run: rush build -v -o storybook
 
       - name: Run Storybook tests
         run: rush test -o storybook


JamesBurnside (approved):

.github/workflows/deploy-storybook.yml
```diff
@@ -28,8 +28,8 @@ jobs:
       - name: Install Dependencies
         run: rush install
 
-      - name: Build Storybook dependencies
-        run: rush build -v -T storybook
+      - name: Build Storybook
+        run: rush build -v -o storybook
```
dmceachernmsft, line 32: What does the `-o` flag do here?

dmceachernmsft (approved):

edwardlee-msft (approved):

.github/workflows/deploy-storybook.yml
edwardlee-msft: https://rushjs.io/pages/commands/rush_build/
short for "--only"
telling rush to only build storybook

PR #1759, opened 2022-04-04
Added UI tests for read receipt and resend button

carocao-msft:
# What
Added UI tests for read receipt and resend button


edwardlee-msft:
ignoring ignoring ignoring

carocao-msft:
> ignoring ignoring ignoring

lol

change/@internal-react-components-3f6426b0-9fb2-4856-aff3-40edc7924dfb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "add id",
```
JamesBurnside, line 3: Update comment so it can be used in our release changelog. Added what id to what?

I also see:
image

Which looks like a functional change we should include in the changelog

change/@internal-react-components-3f6426b0-9fb2-4856-aff3-40edc7924dfb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
JamesBurnside, line 2: Going by changes this should be a patch based on the reordering of participants in the read receipts

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -59,6 +59,10 @@ export interface ChatMessageActionFlyoutProps {
 export const ChatMessageActionFlyout = (props: ChatMessageActionFlyoutProps): JSX.Element => {
   const theme = useTheme();
   const messageReadByCount = props.messageReadBy?.length;
+
+  props.messageReadBy?.sort(function (a, b) {
```
JamesBurnside, line 63: I don't love that this mutates the original array. If this array is used by another component it will be impacted by this.
consider sorting before passing in, or here do:
```ts
const sortedMessageReadyByList = [...props.messageReadBy].sort((a,b) => a.displayName.localeCompare(b.displayName));
```

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -147,7 +151,8 @@ export const ChatMessageActionFlyout = (props: ChatMessageActionFlyoutProps): JS
         },
         submenuIconProps: {
           iconName: 'HorizontalGalleryRightButton',
-          styles: menuIconStyleSet
+          styles: menuIconStyleSet,
+          id: 'chat-composite-message-contextual-menu-read-info'
```
JamesBurnside, line 155: if ids are added for UI tests, use data-ui-id. This likely won't work for adding to the icon of the menu item - can you add to the menu item itself?

JamesBurnside: Something like this should work:
image

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -59,7 +59,10 @@ export interface ChatMessageActionFlyoutProps {
 export const ChatMessageActionFlyout = (props: ChatMessageActionFlyoutProps): JSX.Element => {
   const theme = useTheme();
   const messageReadByCount = props.messageReadBy?.length;
-  const messageReadByList: IContextualMenuItem[] | undefined = props.messageReadBy?.map((person) => {
+
+  const sortedMessageReadyByList = props.messageReadBy?.sort((a, b) => a.displayName.localeCompare(b.displayName));
```
JamesBurnside, line 63: `.sort` will still mutate the original array here, you will need to clone the array first then call `.sort`:
```suggestion
  const sortedMessageReadyByList = [...(props.messageReadBy ?? [])].sort((a, b) => a.displayName.localeCompare(b.displayName));
```

Spread operator here will clone `[...props.messageReadBy]`, you could also use `Array.from(props.messageReadBy)` to clone

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -205,6 +209,7 @@ export const ChatMessageActionFlyout = (props: ChatMessageActionFlyoutProps): JS
   // gap space uses pixels
   return (
     <ContextualMenu
+      id="chat-composite-message-contextual-menu"
```
JamesBurnside, line 212: update to data-ui-id here also

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
JamesBurnside: We chatted offline and found there isn't a good solution for this, lets keep id

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -47,14 +47,43 @@ test.describe('Chat Composite E2E Tests', () => {
...
+    const page0 = pages[0];
+    await sendMessage(page0, testMessageText);
+    await waitForMessageDelivered(page0);
+    await stubMessageTimestamps(page0);
+    expect(await page0.screenshot()).toMatchSnapshot('sent-messages.png');
+
+    const page1 = pages[1];
+    await waitForMessageWithContent(page1, testMessageText);
+
+    // It could be too slow to get typing indicator here, which makes the test flakey
```
JamesBurnside, line 69: This is now in more than one place, can this be a util fn: `await waitForTypingIndicatorHidden(page)`?

JamesBurnside (approved):
🚀🚀🚀

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
carocao-msft: updated ! :)

packages/react-composites/tests/browser/chat/ChatComposite.test.ts
```diff
@@ -38,23 +39,41 @@ test.describe('Chat Composite E2E Tests', () => {
...
-    expect(await page0.screenshot()).toMatchSnapshot('read-message-status.png');
+    expect(await page0.screenshot()).toMatchSnapshot('sent-messages.png');
+
+    const page1 = pages[1];
+    await waitForMessageWithContent(page1, testMessageText);
+    await waitForTypingIndicatorHidden(page1);
+    await stubMessageTimestamps(page1);
+    await page0.locator(dataUiId('chat-composite-message')).click();
+    await page0.locator(dataUiId('chat-composite-message-action-icon')).click();
+    await page0.waitForSelector('[id="chat-composite-message-contextual-menu"]');
```
edwardlee-msft, line 68: I don't see this as a blocker, but just wanted to know whether or not we are okay with setting an "id" for specific components (if necessary).
Would love to hear your thoughts @JamesBurnside

carocao-msft: Oh yeah we had a discussion on this and reached the agreement that we can put an id here. Basically because it's not possible to add data ui id on that component and we want the ui test to keep track of that component

edwardlee-msft (approved):

PR #1763, opened 2022-04-04
[CallWithChat][A11y] Add aria-label for dismiss button in SidePaneHeader

edwardlee-msft:
# What
aria-label added to the dismiss button in SidePaneHeader
Screen Shot 2022-04-04 at 4 06 18 PM

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2790819/

# How Tested
Tested locally on MacOS with VoiceOver

# Process & policy checklist


JamesBurnside (approved):

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -61,6 +61,7 @@
     "pictureInPictureTileAriaLabel": "Video Feeds. Click to return to call screen.",
     "removeMenuLabel": "Remove",
     "copyInviteLinkButtonLabel": "Copy invite link",
+    "dismissSidePaneButton": "Close",
```
dmceachernmsft, line 64: Should this maybe say `close side pane`? just saying close the user might think its closes or leaves the app?

edwardlee-msft: Sounds like that would be the right wording. But A11y recommended to write "Close". 
https://skype.visualstudio.com/SPOOL/_workitems/edit/2790819/

PR #1765, opened 2022-04-05
File Sharing UI - Making components consistent with Figma

anjulgarg:
# What
Modifies file sharing  components to make them consistent with FIGMA 
https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?node-id=4022%3A24730

# Why

# How Tested
Manually

# Process & policy checklist


packages/react-components/src/components/SendBox.tsx
```diff
@@ -305,18 +305,18 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
...
-    [props]
-  );
+  const onRenderFileUploads = useCallback(() => {
+    if (props.activeFileUploads?.length === 0) {
+      return null;
+    }
+    return props.onRenderFileUploads ? (
+      props.onRenderFileUploads()
+    ) : (
+      <div style={{ margin: '0.25rem' }}>
```
JamesBurnside, line 315: From the snapshot of the sendbox, this looks to always be being applied even when no fileUploads are present?
image

This looks to be causing the sendbox being slightly too big when used just as a component (outside of a composite):
image

anjulgarg: Yes, pushed another commit to fix this. :)

anjulgarg: Fixed

packages/react-components/src/components/SendBox.tsx
```diff
@@ -305,18 +305,18 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
...
-    [props]
-  );
+  const onRenderFileUploads = useCallback(() => {
+    if (!props.activeFileUploads?.length) {
+      return null;
+    }
+    return props.onRenderFileUploads ? (
+      props.onRenderFileUploads()
+    ) : (
+      <div style={{ margin: '0.25rem' }}>
```
JamesBurnside, line 315: Looks like we should avoid setting this margin when there are active file uploads, but they failed (or there is some other logic needed to remove them from props.activeFileUploads):
image

anjulgarg: Fixed, thanks for pointing out.

JamesBurnside (approved):

PR #1767, opened 2022-04-05
Update styles for local camera video switcher

dmceachernmsft:
# What

Update styles for local camera video switcher.
# Why

Helps with visibility of the button when present on white backgrounds. Fixes other styling issues like position as well.

image

https://skype.visualstudio.com/SPOOL/_workitems/edit/2791163
# How Tested

Ran locally and visually  inspected styles.


packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -185,20 +185,34 @@ export const layerHostStyle: IStyle = {
...
+      display: 'flex',
+      width: _pxToRem(36),
+      height: _pxToRem(32),
+      flexDirection: 'row',
+      left: _pxToRem(32),
+      top: '0rem',
+      color: '#FFFFFF', // only shows up on running video feed to we want to force specific colours.
+      zIndex: 2, // shows the button directly over the local video feed.
+      background: 'rgba(0,0,0,0.4)',
+      borderRadius: theme.effects.roundedCorner2
```
edwardlee-msft, line 201: bottom left corner of the local video camera switcher button BorderRadius looks different from the border radius of the pip. Is this intentional?

dmceachernmsft: yes, the figma has it with a smaller radius than the pip. I can check with Alex P that its supposed to be that way though.

edwardlee-msft (approved):

packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -185,20 +185,34 @@ export const layerHostStyle: IStyle = {
...
-  }
+export const localVideoCameraCycleButtonStyles = (theme: Theme): IButtonStyles => {
+  return {
+    root: {
+      position: 'absolute',
+      display: 'flex',
+      width: _pxToRem(36),
+      height: _pxToRem(32),
+      flexDirection: 'row',
+      left: _pxToRem(32),
```
JamesBurnside, line 196: Not understanding these positioning styles -- why `left` over `right` here? And what is flex&flexDirection needed for?
Consider wrapping the icon in a stack and stack item having the stack item have `align="end"` to avoid css

packages/react-components/src/components/styles/VideoGallery.styles.ts
```diff
@@ -185,20 +185,34 @@ export const layerHostStyle: IStyle = {
...
+      // styles to remove the unwanted white highlight and blue colour after tapping on button.
+      color: '#FFFFFF',
+      background: 'rgba(0,0,0,0.4)' // sets opacity of background to be visible on all backdrops in video stream.
+    },
+    icon: {
+      padding: '0rem',
+      margin: 0
+    },
+    flexContainer: {
+      height: _pxToRem(24) // sets height to center Icon in button.
```
JamesBurnside, line 213: I'm surprised this is needed to center, and worried this won't center icons if Contoso supplies a different icon.
Would simply setting padding around the icon (currently set to `0rem`) work instead? (I also know the icons are finicky to work with so might not work)

packages/react-components/src/components/styles/VideoGallery.styles.ts
dmceachernmsft: I can switch it but I used styles from the Figma to match the design.

packages/react-components/src/components/styles/VideoGallery.styles.ts
dmceachernmsft: I try playing with the padding here and it just pushed it further down, right now if the height isnt set here there is a big overhead on the Icon pushing the button further into the pip than wanted.

JamesBurnside (approved):

PR #1771, opened 2022-04-06
Added chat test setup with per user query arguments & UI tests for file download card

prabhjot-msft:
# What
https://skype.visualstudio.com/SPOOL/_workitems/edit/2814688
https://skype.visualstudio.com/SPOOL/_workitems/edit/2814688

# Why
Added chat test setup with per user query arguments
Tests for file download cards in message thread

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/common/chatTestHelpers.ts
```diff
@@ -35,6 +35,28 @@ export const chatTestSetup = async ({
   await Promise.all(pageLoadPromises);
 };
 
+export const chatTestSetupWithPerUserArgs = async ({
```
JamesBurnside, line 38: To reduce duplicate code, can `chatTestSetup` (fn above this) call this function:
```ts
export const chatTestSetup = async (pages, users, serverUrl, qArgs) =>
  await chatTestSetupWithPerUserArgs(pages, users.map(user => ({ user, qArgs }), serverUrl);
```
(^code totally untested, just an impression of my thought)

JamesBurnside (commented):
This looks great! Waiting to quickly have a look at the UI images when they update before hitting approve

packages/react-composites/tests/browser/chat/FileSharing.test.ts
```diff
@@ -194,3 +200,54 @@ test.describe('Filesharing SendBox Errorbar', async () => {
     expect(await page.screenshot()).toMatchSnapshot('filesharing-sendbox-file-upload-in-progress-error.png');
   });
 });
+
+test.describe('Filesharing Message Thread', async () => {
+  test.skip(isTestProfileStableFlavor());
+
+  test.beforeEach(async ({ pages, users, serverUrl }) => {
+    const usersWithArgs: { user: ChatUserType; qArgs?: { [key: string]: string } }[] = [];
+    for (let idx = 0; idx < users.length; idx++) {
```
prprabhu-ms, line 209: This can be simplified to:

```
const firstUserArgs = { user: users[0] qargs: { /* qargs for the first user */ }};
const otherUsersArgs = users.slice(1).map(user => ({user, qArgs: { /* qargs for other users */ }));
const usersArgs = [firstUserArgs, ...otherUsersArgs];

File: packages/react-composites/tests/browser/common/chatTestHelpers.ts [hunk 3]
```diff
@@ -35,6 +35,28 @@ export const chatTestSetup = async ({
...
+  pages,
+  usersWithArgs,
+  serverUrl
+}: {
+  pages: Page[];
+  usersWithArgs: { user: ChatUserType; qArgs?: { [key: string]: string } }[];
+  serverUrl: string;
+}): Promise<void> => {
+  const newUsers = await createChatThreadAndUsers(TEST_PARTICIPANTS);
```
prprabhu-ms, line 47: This function repeats the bug on line 24: it accepts `users` but then creates new users ignoring the ones passed in. That is very surprising, and we lost a few hours last week debugging it.

We should instead fix this bug both here and on line 24. If the function accepts `users`, it must respect those users. There is no need to recreate users and thread, just reload the page for the given users with the args you want.

JamesBurnside: The reason the thread is recreated is to ensure each test is using it's own thread. Ideally we should remove the fixture code that initially sets up a thread as this is only run before all tests (not before every test which is what we need)

JamesBurnside: (But agree that it is bad we pass in users and then recreate new ones)

packages/react-composites/tests/browser/chat/fixture.ts
```diff
@@ -14,17 +14,17 @@ const APP_DIR = path.join(__dirname, 'app');
...
- * A worker-scoped test fixture for ChatComposite browser tests.
+ * A test-scoped test fixture for ChatComposite browser tests.
  *
  * The @returns values are available to reference in tests.
  */
-export const test = base.extend<unknown, ChatWorkerFixture>({
+export const test = base.extend<ChatWorkerFixture>({
   /** @returns string URL for the server. */
-  serverUrl: [createTestServer({ appDir: APP_DIR, serverUrl: SERVER_URL }), { scope: 'worker' }],
+  serverUrl: [createTestServer({ appDir: APP_DIR, serverUrl: SERVER_URL }), { scope: 'test' }],
```
JamesBurnside, line 23: nice!!

packages/react-composites/tests/browser/chat/FileSharing.test.ts
```diff
@@ -194,3 +201,59 @@ test.describe('Filesharing SendBox Errorbar', async () => {
...
+    });
+  });
+  test('contains File Download Card', async ({ pages }) => {
+    const testMessageText = 'Hello!';
+    const page0 = pages[0];
+
+    await waitForChatCompositeToLoad(page0);
+    await sendMessage(page0, testMessageText);
+    await waitForMessageDelivered(page0);
+    await page0.waitForTimeout(1000);
```
JamesBurnside, line 243: what is the wait for 1000ms for? Can this be changed to waiting for a specific expected element?

packages/react-composites/tests/browser/chat/FileSharing.test.ts
```diff
@@ -194,3 +201,59 @@ test.describe('Filesharing SendBox Errorbar', async () => {
...
+
+    await stubMessageTimestamps(page0);
+    expect(await page0.screenshot()).toMatchSnapshot('filesharing-file-download-card-in-sent-messages.png');
+
+    const page1 = pages[1];
+
+    // It could be too slow to get typing indicator here, which makes the test flakey
```
JamesBurnside, line 250: This is now in more than one place, can this be moved to a util fn: `waitForTypingIndicatorHidden(page)`

prprabhu-ms (approved):
w00t! Happy to see that this PR makes all the chat tests more hermetic.

I'd have recommended splitting into two PRs. First one just converts the fixture to a per-test one, and the second one adds new tests. That way, it'd be easier to find the PR to blame if we see increased test flakiness / service side throttling from Chat etc.

But let's let this one slide.

packages/react-composites/tests/browser/chat/FileSharing.test.ts
```diff
@@ -194,3 +201,54 @@ test.describe('Filesharing SendBox Errorbar', async () => {
...
+    });
+  });
+  test('contains File Download Card', async ({ pages }) => {
+    const testMessageText = 'Hello!';
+    const page0 = pages[0];
+
+    await waitForChatCompositeToLoad(page0);
+    await sendMessage(page0, testMessageText);
+    await waitForMessageDelivered(page0);
+    await page0.waitForSelector(dataUiId('file-download-card-group'));
```
JamesBurnside, line 243: use `waitForSelector(page, ...)` instead of `page.waitForSelector`. waitForSelector(page, ..) is our custom function that wraps the selector in our desired timeout and screenshots on failure

JamesBurnside (approved):

PR #1772, opened 2022-04-06
Add announcer to camera and mic buttons

dmceachernmsft:
# What

Adds a announcer component and add it to camera and microphone buttons
# Why

Allows for us to introduce announcments when actions are taken on the page with a screen reader turned on.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2790284
# How Tested

Ran locally and tested with windows narrator and Android talk back.


packages/react-components/src/components/Announcer.tsx
```diff
@@ -0,0 +1,41 @@
...
+ *  aria props for this hidden component.
+ */
+export type AnnouncerProps = {
+  announcementString: string;
+  ariaLive: 'off' | 'polite' | 'assertive' | undefined;
+};
+
+/**
+ * @private
+ * Announcer component to make Aira announcments on actions
```
JamesBurnside, line 18: sp: announcments -> announcements

packages/communication-react/review/communication-react.api.md
```diff
@@ -848,6 +848,8 @@ export type CameraButtonSelector = (state: CallClientState, props: CallingBaseSe
 
 // @public
 export interface CameraButtonStrings {
+    cameraActionTurnedOffAnnouncment: string;
```
JamesBurnside, line 851: sp: announcment -> announcement (here and line below)
```suggestion
    cameraActionTurnedOffAnnouncement: string;
```

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -139,12 +148,15 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
   const [waitForCamera, setWaitForCamera] = useState(false);
   const localeStrings = useLocale().strings.cameraButton;
   const strings = { ...localeStrings, ...props.strings };
-  const onRenderCameraOnIcon = (): JSX.Element => (
-    <HighContrastAwareIcon disabled={props.disabled || waitForCamera} iconName="ControlButtonCameraOn" />
-  );
-  const onRenderCameraOffIcon = (): JSX.Element => (
-    <HighContrastAwareIcon disabled={props.disabled || waitForCamera} iconName="ControlButtonCameraOff" />
-  );
+  const [announcerString, setAnnouncerString] = useState<string>(strings.cameraActionTurnedOffAnnouncment);
```
JamesBurnside, line 151: (From Teams chat) Can we avoid the set state and have this be
```ts
const announcerString = (props?.checked && !waitForCamera) ? strings.cameraActionTurnedOnAnnouncment : strings.cameraActionTurnedOffAnnouncment
```
Although in writing this again I see it may be called on first render which we won't want

JamesBurnside: Having it inside the `onRenderCameraOnIcon` though isn't ideal as if Contoso supplies their own icon they lose this feature

JamesBurnside: Maybe we keep the setState but put it in `onToggleClick`?

packages/communication-react/review/communication-react.api.md
JamesBurnside: Oh and with microphoneActionTurnedOffAnnouncment

packages/react-components/src/components/Announcer.tsx
```diff
@@ -0,0 +1,41 @@
...
+ */
+export const Announcer = (props: AnnouncerProps): JSX.Element => {
+  const { announcementString, ariaLive } = props;
+
+  return <Stack aria-label={announcementString} aria-live={ariaLive} styles={announcerStyles}></Stack>;
+};
+
+/**
+ * Styles to hide the announcer from view but still displaying it so that narration can happen.
```
JamesBurnside, line 27: nit: instead of `displaying` say something like `exists in the DOM tree` to make it clear nothing is visible/being displayed to the user's screen

packages/react-components/src/components/CameraButton.tsx
dmceachernmsft: I will move it there and implement for the microphone button as well.

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -139,44 +148,62 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
...
     if (onToggleCamera) {
+      // allows for the setting of narrator strings triggering the announcer when camera is turned on or off.
+      if (announcerString === strings.cameraActionTurnedOffAnnouncement) {
+        setAnnouncerString(strings.cameraActionTurnedOnAnnouncement);
+      } else {
+        setAnnouncerString(strings.cameraActionTurnedOffAnnouncement);
+      }
       setWaitForCamera(true);
       try {
         await onToggleCamera(localVideoViewOptions ?? defaultLocalVideoViewOptions);
```
JamesBurnside, line 173: Move the `toggleAnnouncerString` to after `await onToggleCamera` as it should be announced after the action has completed successfully.
e.g. if this fails to turn on the video it shouldn't have announced that the video was turned on.

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -139,44 +148,62 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
...
+  };
   if (waitForCamera && strings.tooltipVideoLoadingContent) {
     strings.tooltipDisabledContent = strings.tooltipVideoLoadingContent;
   }
 
   const onToggleClick = useCallback(async () => {
     // Throttle click on camera, need to await onToggleCamera then allow another click
     if (onToggleCamera) {
+      // allows for the setting of narrator strings triggering the announcer when camera is turned on or off.
+      if (announcerString === strings.cameraActionTurnedOffAnnouncement) {
```
JamesBurnside, line 166: This toggle will be incorrect if the component starts with the camera already enabled. Instead of checking the string, can we check the state of the button? (maybe props.checked?)

JamesBurnside: Difficulty here is that we want it to almost be the opposite of props.checked...
I'll let you explore what else in the interface could be best for this, but one suggestion is to revert back the idea of having:

```ts
const announcerString = props.checked ? strings.cameraActionTurnedOnAnnouncement : strings.cameraActionTurnedOffAnnouncement
```

Then to fix the issue of it being read on first render have:
```ts
const [buttonClicked, setButtonClicked] = useState(false);

// Only set the announcer string if the button has been clicked at least once
const announcerString = !buttonClicked ? undefined : props.checked ? strings.cameraActionTurnedOnAnnouncement : strings.cameraActionTurnedOffAnnouncement

onToggleClick() => {
  ...
  await onToggleCamera();
  setButtonClicked(true);
}
```

JamesBurnside: Actually that won't work if the button is clicked and is in the waitForCamera state... I'll stop trying to post solutions I haven't tested...

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -139,44 +148,62 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
...
+        setAnnouncerString(strings.cameraActionTurnedOnAnnouncement);
+      } else {
+        setAnnouncerString(strings.cameraActionTurnedOffAnnouncement);
+      }
       setWaitForCamera(true);
       try {
```
JamesBurnside, line 172: If this fails should we read out: 'Failed to start/stop camera'?

JamesBurnside: Actually error bar will show and be read out so we shouldn't need to

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -139,44 +148,62 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
...
   const localeStrings = useLocale().strings.cameraButton;
   const strings = { ...localeStrings, ...props.strings };
-  const onRenderCameraOnIcon = (): JSX.Element => (
-    <HighContrastAwareIcon disabled={props.disabled || waitForCamera} iconName="ControlButtonCameraOn" />
-  );
-  const onRenderCameraOffIcon = (): JSX.Element => (
-    <HighContrastAwareIcon disabled={props.disabled || waitForCamera} iconName="ControlButtonCameraOff" />
-  );
+  const [announcerString, setAnnouncerString] = useState<string>(strings.cameraActionTurnedOffAnnouncement);
+  const onRenderCameraOnIcon = (): JSX.Element => {
```
JamesBurnside, line 152: nitty nit nit: remove changes to these onRender fns from changelist

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -149,34 +161,49 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
...
   const onToggleClick = useCallback(async () => {
     // Throttle click on camera, need to await onToggleCamera then allow another click
     if (onToggleCamera) {
+      // allows for the setting of narrator strings triggering the announcer when camera is turned on or off.
       setWaitForCamera(true);
       try {
         await onToggleCamera(localVideoViewOptions ?? defaultLocalVideoViewOptions);
       } finally {
         setWaitForCamera(false);
+        toggleAnnouncerString();
```
JamesBurnside, line 181: anything in the `finally` block will always execute, (whether the try block threw an error or not), but if onToggleCamera fails we won't want to update the announcer string -- can this be placed after the `await onToggleCamera` inside the try catch?

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -149,34 +161,49 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
...
+      announcerString === strings.cameraActionTurnedOffAnnouncement
+        ? strings.cameraActionTurnedOnAnnouncement
+        : strings.cameraActionTurnedOffAnnouncement
+    );
+  }, [announcerString, strings.cameraActionTurnedOffAnnouncement, strings.cameraActionTurnedOnAnnouncement]);
+
   const onToggleClick = useCallback(async () => {
     // Throttle click on camera, need to await onToggleCamera then allow another click
     if (onToggleCamera) {
+      // allows for the setting of narrator strings triggering the announcer when camera is turned on or off.
```
JamesBurnside, line 175: more/remove comment

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -139,6 +148,9 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
   const [waitForCamera, setWaitForCamera] = useState(false);
   const localeStrings = useLocale().strings.cameraButton;
   const strings = { ...localeStrings, ...props.strings };
+  const [announcerString, setAnnouncerString] = useState<string>(
+    props.checked ? strings.cameraActionTurnedOnAnnouncement : strings.cameraActionTurnedOffAnnouncement
```
JamesBurnside, line 152: my feeling is this needs to be undefined on first render, otherwise what stops this being read out on the first render pass?
```ts
const [announcerString, setAnnouncerString] = useState<string | undefined>();
```

packages/react-components/src/components/CameraButton.tsx
dmceachernmsft: yes will move 😊

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -149,34 +161,49 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
     strings.tooltipDisabledContent = strings.tooltipVideoLoadingContent;
   }
 
+  const toggleAnnouncerString = useCallback(() => {
```
JamesBurnside, line 164: toggling may work most of the time, but in theory could be out of sync. If we don't account for all edge cases then this could be out of sync with the button state (thinking of things like our app is backgrounded and another app uses the camera thus making our app lose it, then the user returns to the app, will this become out of sync? more over -- can we catch every edge case that might happen?).

I think this needs to read it out based on the button state. Perhaps could be as simply as checking the props.checked.

Maybe this should be `const updateAnnouncerString = useCallback((cameraOn: boolean) => {`

Then in `onToggleClick`:
```ts
await onToggleCamera(...);
// Camera state is now the opposite of the checked state as we have just successfully toggled the camera
const cameraNowOn = !props.checked;
updateAnnouncerString(cameraNowOn);
```

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -149,34 +159,54 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
     strings.tooltipDisabledContent = strings.tooltipVideoLoadingContent;
   }
 
+  const cameraOn = props.checked;
+
+  const toggleAnnouncerString = useCallback(
+    (isCameraOn: boolean) => {
+      setAnnouncerString(
+        announcerString === (strings.cameraActionTurnedOffAnnouncement || undefined) || !isCameraOn
```
JamesBurnside, line 167: do you need the `(strings.cameraActionTurnedOffAnnouncement || undefined)` in the check?

If announcerString currently equals `cameraActionTurnedOffAnnouncement` and the camera is turnedOff this if will equate to true and announce the camera is turned on

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -149,34 +159,54 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
...
+  );
+
   const onToggleClick = useCallback(async () => {
     // Throttle click on camera, need to await onToggleCamera then allow another click
     if (onToggleCamera) {
       setWaitForCamera(true);
       try {
         await onToggleCamera(localVideoViewOptions ?? defaultLocalVideoViewOptions);
+        // allows for the setting of narrator strings triggering the announcer when camera is turned on or off.
+        toggleAnnouncerString(cameraOn ? cameraOn : false);
```
JamesBurnside, line 182: if camera was On, button will be checked (confirm this), and was just toggled so we want to announce it is turned off -- should this be `toggleAnnouncerString(!cameraOn)`?

dmceachernmsft: so the thought here was to abuse the undefined state of that boolean like a `false` but I will investigate this.

change/@internal-react-components-5e7b0e61-f149-4204-9438-9428ca7f6143.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Introduce announcer component which is to be used for announcing when actions are made  ",
```
JamesBurnside, line 3: nit: as this is for the changelog, the part customers will care about is now some control bar buttons announce to screen readers state change
```suggestion
  "comment": "Announce when microphone is turned on/off and video is turned on/off to screen readers and narrator tools",
```

packages/react-components/src/components/Announcer.tsx
```diff
@@ -0,0 +1,42 @@
...
+ *  aria props for this hidden component.
+ */
+export type AnnouncerProps = {
+  announcementString?: string;
+  ariaLive: 'off' | 'polite' | 'assertive' | undefined;
+};
+
+/**
+ * @private
+ * Announcer component to make Aira announcements on actions
```
alkwa-msft, line 18: aria

packages/react-components/src/components/MicrophoneButton.tsx
```diff
@@ -150,31 +159,62 @@ export interface MicrophoneButtonProps extends ControlBarButtonProps {
...
-      ariaDescription={
-        enableDeviceSelectionMenuTrampoline(props) ? strings.microphoneButtonSplitRoleDescription : undefined
+  const onToggleClick = useCallback(async () => {
+    if (onToggleMicrophone) {
+      try {
+        await onToggleMicrophone();
+        // allows for the setting of narrator strings triggering the announcer when microphone is turned on or off.
+        toggleAnnouncerString(!isMicOn);
+      } catch {
+        setAnnouncerString(announcerString);
```
JamesBurnside, line 193: I would remove this announcement in the catch. It will announce something has changed when it hasn't.

JamesBurnside (approved):

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -149,34 +159,54 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
     strings.tooltipDisabledContent = strings.tooltipVideoLoadingContent;
   }
 
+  const cameraOn = props.checked;
+
+  const toggleAnnouncerString = useCallback(
+    (isCameraOn: boolean) => {
+      setAnnouncerString(
+        announcerString === strings.cameraActionTurnedOffAnnouncement || !isCameraOn
```
edwardlee-msft, line 167: Non blocking question as I tested on android and seems to be working correctly (except for when focusing on a local video tile then back to the button):
I'm confused as to why announcerString is being compared to strings.cameraActionTurnedOffAnnouncement.
Would it not always set announcerString to the off string if announcerString is already the off string? It would not reach the !isCameraOn condition.

anjulgarg (approved):

edwardlee-msft:
Tested on safari and chrome on mac and also tested on Android. Working as intended. Seems like the cycling to local video tile and losing the announcer talkback bug is resolved as well.

PR #1773, opened 2022-04-07
Wait for timestamps have been stubbed in UI tests

mgamis-msft:
# What

- Using `textContent` instead of `innerHTML` because `innerHTML` is slow because HTML tags have to be searched and parsed into DOM nodes. 
- Wait for timestamps to stubbed

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2779986

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -238,7 +238,17 @@ export const waitForPiPiPToHaveLoaded = async (page: Page, videosEnabledCount: n
 export const stubMessageTimestamps = async (page: Page): Promise<void> => {
   const messageTimestampId: string = dataUiId(IDS.messageTimestamp);
   await page.evaluate((messageTimestampId) => {
-    Array.from(document.querySelectorAll(messageTimestampId)).forEach((i) => (i.innerHTML = 'timestamp'));
+    Array.from(document.querySelectorAll(messageTimestampId)).forEach((i) => (i.textContent = 'timestamp'));
+    (function checkForStubbedTimestamps(i) {
+      setTimeout(function () {
```
JamesBurnside, line 243: avoid our own setTimeouts, instead use `waitForFunction`

so something like:
```ts
// Set timestamps in the DOM
await page.evaluate((messageTimestampId) => {
  Array.from(document.querySelectorAll(messageTimestampId)).forEach((i) => (i.textContent = 'timestamp'));
}, messageTimestampId);

// Wait for timestamps to have been updated in the DOM
await waitForFunction(page, (args: any) => {
  const timestampNodes = document.querySelectorAll(args.messageTimestampId);
  return timestampNodes.every(node => node.innerHTML === 'timestamp');
}, {
  messageTimestampId: messageTimestampId
})
```
(^disclaimer: completely untested code written in the github browser comment window, likely has syntax and logical errors)

Using waitForFunction both allows avoiding custom timeouts and uses the global timeout system, but is also wrapped in a takeScreenshotOnFailure in the event of a timeout happening.

JamesBurnside: waitForFunction will also fail the test here if the timestamps never updated

mgamis-msft: Thanks for the code snippet. It worked for me. Before I commit this, I was wondering does `waitForFunction` retry when `timestampNodes.every(node => node.innerHTML === 'timestamp')` returns false?

JamesBurnside: It should do yep. It should poll every x seconds (or every animationFrameRequest depending on what the default polling type is) and then has an overall timeout: https://playwright.dev/docs/api/class-page#page-wait-for-function (which we limit to 5 seconds)
image

JamesBurnside (approved):

PR #1774, opened 2022-04-07
Responsive attach file icon

anjulgarg:
# What
- Introducing a `formFactor` prop to chat composite similar to call composite
- Attach file icon position changes basis on the form factor

# Why
To ensure that composite matches figma designs
image

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -205,23 +230,25 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
-                onChange={fileUploadButtonOnChange}
-              />
-            )}
+            <Stack horizontal={formFactor === 'mobile'} horizontalAlign="start">
+              {formFactor === 'mobile' && (
+                <Stack verticalAlign="center">
+                  <AttachFileButton />
+                </Stack>
+              )}
+              <div style={{ width: '100%' }}>
```
JamesBurnside, line 239: can this be a `StackItem` with `grow` property?
`width: 100%` is implying it should fill `100%` of the parents width but it really wants to be 100% minus any other sibling elements

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -205,23 +230,25 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
-            />
-
-            {fileSharing?.uploadHandler && (
-              <FileUploadButton
-                accept={fileSharing?.accept}
-                multiple={fileSharing?.multiple}
-                onChange={fileUploadButtonOnChange}
-              />
-            )}
+            <Stack horizontal={formFactor === 'mobile'} horizontalAlign="start">
```
JamesBurnside, line 233: Why `horizontalAlign="start"`? (this implies left alignment)

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -205,23 +230,25 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
+              {formFactor === 'mobile' && (
+                <Stack verticalAlign="center">
+                  <AttachFileButton />
+                </Stack>
+              )}
+              <div style={{ width: '100%' }}>
+                <SendBox
```
JamesBurnside, line 240: Are we bringing any of these features to the sendbox component we export?

anjulgarg: Not 100% sure about the best way to do it.
Since the placement of file attach icon is so different for mobile and desktop.

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -33,11 +33,17 @@ export interface ChatCompositeProps extends BaseCompositeProps<ChatCompositeIcon
...
    * Flags to enable/disable visual elements of the {@link ChatComposite}.
    */
   options?: ChatCompositeOptions;
+  /* @conditional-compile-remove(file-sharing) */
+  /**
+   * Optimizes the composite form factor for either desktop or mobile.
+   * @remarks `mobile` is currently only optimized for Portrait mode on mobile devices and does not support landscape.
+   * @defaultValue 'desktop'
+   */
+  formFactor?: 'desktop' | 'mobile';
```
JamesBurnside, line 46: when this goes stable we'll have to somehow remember to update storybook docs...

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
JamesBurnside: Yeah not sure either, but can always make this decision when we support rich text as that's when many more changes will be made cc @PorterNan

anjulgarg: Right. That's why I am keeping it simple for now.

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
anjulgarg: I have made a note.

JamesBurnside (approved):

dmceachernmsft (approved):

PR #1777, opened 2022-04-08
No re-render of Call composite if chat/people pane is opened and closed

mgamis-msft:
# What

Hiding inner call composite in CallWithChat composite instead of unmounting and re-rendering if chat/people pane is opened and closed. 

# Why

Dismissed compliance banners will reappear if chat/people pane is opened and closed.
Before: 
https://user-images.githubusercontent.com/79475487/162343097-7321bce7-4859-4291-96a9-7b6822e66208.mp4
After:
https://user-images.githubusercontent.com/79475487/162343196-195d4cee-3610-4394-a1f2-aeb22d4ca9d2.mp4

# How Tested

Locally

# Process & policy checklist


packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -242,17 +243,15 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
   return (
     <Stack verticalFill grow styles={compositeOuterContainerStyles} id={compositeParentDivId}>
       <Stack horizontal grow>
-        {!isMobileWithActivePane && (
```
JamesBurnside, line 245: Out of interest, if keyboard navigating, where does focus go when the side pane is closed?

JamesBurnside, line 245: Can you test also, in mobile, does the PiPiP still work as expected? I'm worried the video tiles on the gallery will still have handles to the streams and maybe won't load the streams of the PiPiP correctly

prprabhu-ms:
waiting for answers to pending questions before I approve.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
mgamis-msft: The Call local stream stops working after opening and closing the chat/people pane.
https://user-images.githubusercontent.com/79475487/164388812-4d1c3dbc-66af-4ab9-98ca-5e988ffde635.mp4
The fix will to be done to the streams in the Call without rerendering the Call. 

Closing this PR. 

Logged a bug for the Compliance Banner reappearing https://skype.visualstudio.com/SPOOL/_search?text=Dismissed*&type=workitem&lp=workitems-Project&filters=Projects%7BSPOOL%7D&pageSize=25.

PR #1779, opened 2022-04-08
Change tooltip color in dark mode

carocao-msft:
# What
Changed tooltip color in dark mode to "neutralLighter"(#383735) 

# Why
(https://skype.visualstudio.com/SPOOL/_workitems/edit/2658072)

# How Tested
Tooltip color in light mode should not be changed  (still verifying with Alex P)
Screen Shot 2022-04-08 at 9 47 32 AM
Screen Shot 2022-04-08 at 9 47 32 AM

tooltip color in dark mode is changed to  "neutralLighter"(#383735) 
Screen Shot 2022-04-08 at 9 47 45 AM

# Process & policy checklist

no


packages/react-components/src/components/ControlButtonTooltip.tsx
```diff
@@ -18,6 +23,15 @@ const hostStyles: Partial<ITooltipHostStyles> = { root: { display: 'inline-block
...
  */
 export const ControlButtonTooltip = (props: ITooltipHostProps): JSX.Element => {
+  const theme = useTheme();
+  const calloutStyle: Partial<ICalloutContentStyles> = { root: { padding: 0 }, calloutMain: { padding: '0.5rem' } };
+
+  // Place callout with no gap between it and the button.
+  const calloutProps = {
+    gapSpace: 0,
+    styles: calloutStyle,
+    backgroundColor: isDarkThemed(theme) ? theme.palette.neutralLighter : ''
```
JamesBurnside, line 33: Does this background color need to be updated for every tooltip we have? Or just control bar buttons. (Do we mind having inconsistent tooltip backgrounds?)

carocao-msft: Hmmm good question, let me ask Alex P

JamesBurnside: Note: (we may not care about this but) if a customer is using fluent components alongside our components/composites, the existing tooltips on their site will not be consistent with ours if we update them all

carocao-msft: Updated background color for all tooltips

JamesBurnside (approved):

edwardlee-msft (approved):

PR #1786, opened 2022-04-08
Custom Data Model Fix Video Gallery

dmceachernmsft:
# What

Adds use of `onFetchAvatarPersonaData` Composite prop to be used in the video gallery.
# Why

Allows for the custom data model to override the display name of the remote users based on contoso's function.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2694662
# How Tested

Ran locally with Custom data model to override name.
image


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -309,6 +313,14 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
   const defaultOnRenderVideoTile = useCallback(
     (participant: VideoGalleryRemoteParticipant, isVideoParticipant?: boolean) => {
       const remoteVideoStream = participant.videoStream;
+      // Async call to retrieve custom data model information to replace display name in remote tiles with camera on.
+      const newParticipantName = async (): Promise<void> => {
```
anjulgarg, line 317: Might be cleaner this way
```tsx
(async () {
  if (onFetchAvatarPersonaData) {
    const newPaticipantData = await onFetchAvatarPersonaData(participant.userId);
    participant.displayName = newPaticipantData.text;
  }
})();
```

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -336,6 +340,15 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
   const defaultOnRenderVideoTile = useCallback(
     (participant: VideoGalleryRemoteParticipant, isVideoParticipant?: boolean) => {
       const remoteVideoStream = participant.videoStream;
+      // Async call to retrieve custom data model information to replace display name in remote tiles with camera on.
+      if (onFetchAvatarPersonaData) {
+        (async () => {
+          const newPaticipantData = await onFetchAvatarPersonaData(participant.userId);
+          participant.displayName = newPaticipantData.text;
```
JamesBurnside, line 347: this mutates the participant displayName prop which is bad practice (if this is used elsewhere the value of the property is suddenly changed).
Instead we'll need to call a setState(participantsMergedWithFetchedData) once this returns with the new data. This also means we should have no name until the name is fetched from the onFetchAvatarPersonaData

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -336,6 +340,15 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
   const defaultOnRenderVideoTile = useCallback(
     (participant: VideoGalleryRemoteParticipant, isVideoParticipant?: boolean) => {
       const remoteVideoStream = participant.videoStream;
+      // Async call to retrieve custom data model information to replace display name in remote tiles with camera on.
+      if (onFetchAvatarPersonaData) {
+        (async () => {
```
JamesBurnside, line 345: this function is resonably complex and is nested in a complex structure (a tileRender inside a video gallery). Pull this out as a helper function
```suggestion
        fetchAvatarPersonaDataAsync()
```

JamesBurnside, line 345: we should memoize this result so it doesn't get called on every render

packages/react-components/src/components/VideoGallery.tsx
JamesBurnside: Without a useState or similar there's also nothing to tell this component to rerender when this async block completes. (imagine it took 10 seconds to complete and everything in the app was stable and unchanging -- nothing here would tell the component to rerender to update the name)

JamesBurnside, line 347: AvatarPersonaData has many fields, but here we only overwrite the displayName prop - is there a plan to support the other props in this interface?

dmceachernmsft: All of the other properties of the avatarPersonaData are only related to the avatar that is present when the camera is off and in the participant list in its many forms. the key issue here unfortunately is that the display name that the video gallery is placing on each tile isn't getting the new `text` or name of the participant that shows in the participant list.

anjulgarg (approved):

dmceachernmsft:
Closing this PR as found a better solution to this keeping the changes to the composites

PR #1787, opened 2022-04-08
Test file download error bar

prabhjot-msft:
# What
https://skype.visualstudio.com/SPOOL/_workitems/edit/2807695

# Why
Test file download error bar

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/chat/app/index.tsx
```diff
@@ -73,6 +79,16 @@ function App(): JSX.Element {
     }
   }, [adapter]);
 
+  const fileDownloadHandler: FileDownloadHandler = (userId, fileData): Promise<URL | FileDownloadError> => {
+    return new Promise((resolve) => {
+      if (fileData.name !== 'Unauthorized.pdf') {
```
prprabhu-ms, line 84: Keying off of filename here feels unclean. Can you add a new field `fileData.shouldFailDownload` or something that makes this more robust?

prprabhu-ms: Oh wait, `fileData` comes from the sent message, so it's part of the exported API (can't add fields to that that).
In that case, I'd recommend adding a new `qArg` to the test sample app that says "fail all file downloads". And then load `page1` with that `qArg`.

JamesBurnside:
This is showing the onHover action menu:
image

I worry this could be flakey... we have a helper function, something like `clickOutsideOfPage()`, can you add this before taking the screenshot?

mgamis-msft (approved):

anjulgarg (approved):

prprabhu-ms (approved):

PR #1788, opened 2022-04-08
Show Error Bar to user when joining a call fails

JamesBurnside:
# What
* Add ErrorBar to configuration page
  * This is surfaced at the top of the page.
  * **non goal:** surface error bar in a more relevant position (i.e. beside start call for failed calls and in the local preview box for failed camera start)
* Update error bar to handle callagent:join failures
* Tee adapter.joinCall errors to state

# Why
Currently nothing happens if there is an invalid teams link, you click join call and nothing happens (an error is logged to console but the end user sees nothing happen)

https://skype.visualstudio.com/SPOOL/_workitems/edit/2763425

# How Tested
Desktop:
image

Mobile:
image


edwardlee-msft (approved):

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -133,7 +133,9 @@
     "callCameraAlreadyInUse": "Unable to access camera. It may already be in use by another application.",
     "callMacOsCameraAccessDenied": "MacOS is blocking access to your camera. Update your privacy settings to allow this browser to access your camera.",
     "callMacOsScreenShareAccessDenied": "MacOS is blocking screen sharing. Update your privacy settings to allow this browser to record your screen.",
-    "dismissButtonAriaLabel": "Close"
+    "dismissButtonAriaLabel": "Close",
+    "failedToJoinCallGeneric": "Failed to join call.",
+    "failedToJoinCallInvalidMeetingLink": "Unable to join Meeting. Invalid Link."
```
anjulgarg, line 138: What's the reasoning behind using `call` and `Meeting` for separate errors.

anjulgarg (approved):

packages/react-components/src/localization/locales/en-US/strings.json
JamesBurnside: I used "Meeting" over call in one of them as this error is specific for invalid Teams Meetings links (we could update this string for Teams Meeting).

Then there's two strings here:
1. Failed to join from an invalid teams link -- this felt actionable by the end user as often meetings can't be joined because someone was sent or copy pasted the link incorrectly.
2. Otherwise we just have a callAgent.Join error which is very generic catch-all error message as this could be due to anything so unless they specifically can't join because of an invalid teams link we just show "Failed to join call."

packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -89,6 +89,17 @@ export const errorBarSelector: ErrorBarSelector = createSelector(
     appendActiveErrorIfDefined(activeErrorMessages, latestErrors, 'Call.startVideo', 'startVideoGeneric');
     appendActiveErrorIfDefined(activeErrorMessages, latestErrors, 'Call.unmute', 'unmuteGeneric');
 
+    if (latestErrors['CallAgent.join']?.message === 'CallAgent.join: Invalid meeting link') {
```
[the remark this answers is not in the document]
JamesBurnside: Yeah, unfortunately we don't get anything more specific back from the call agent. I believe it is part of the SDK version and not a backend message that is propagated so hopefully shouldn't break from a service change, only an SDK update. If it does fail we will still show an error message "Failed to join call." just not the more specific/helpful error message "Unable to join Meeting. Invalid Link."

PR #1790, opened 2022-04-11
Default Renderer for File downloads in Message Thread

anjulgarg:
# What
- Adds default renderer for File Downloads in Message Thread
- Removed duplicate file sharing code from `react-composites`

# Why
To allow file sharing to be used without chat composite

# How Tested
Manually tested.

# Process & policy checklist


packages/react-components/src/theming/icons.tsx
```diff
@@ -74,6 +76,10 @@ export const DEFAULT_COMPONENT_ICONS = {
   ControlButtonParticipants: <People20Filled />,
   ControlButtonScreenShareStart: <ShareScreenStart20Filled />,
   ControlButtonScreenShareStop: <ShareScreenStop20Filled />,
+  /* @conditional-compile-remove(file-sharing) */
+  Cancel: <FontIcon iconName="Cancel" />,
```
JamesBurnside, line 80: Can we use icons from `'@fluentui/react-icons'` here to move towards having consistently bundled icons and not pull from fluent's CDN (and avoid this: [Bug 2821183](https://skype.visualstudio.com/SPOOL/_workitems/edit/2821183): [Web] Flaky tests: File sharing download icon)

JamesBurnside: (not blocking for this PR, just suggestion)

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -572,6 +574,20 @@ export type MessageThreadProps = {
...
+   * Optional function called when someone clicks on the file download icon.
+   */
+  fileDownloadHandler?: FileDownloadHandler;
+  /* @conditional-compile-remove(file-sharing) */
+  /**
+   * @beta
+   * Property name that contains information about file downloads in `message.metadata` object.
+   * @defaultValue fileSharingMetadata
+   */
+  fileDownloadMetadataKey?: string;
```
JamesBurnside, line 590: Could this be different per message? Or guaranteed to be consistent across the whole message thread?

packages/react-components/src/theming/icons.tsx
anjulgarg: Absolutely. I overlooked that. Thanks for the headsup

packages/react-components/src/components/MessageThread.tsx
anjulgarg: Consistent for the whole message thread since this prop is set at the message thread level

JamesBurnside: Might Contoso want a different key for different messages? Struggling to think of a good reason why they would do this.. perhaps they updated to a new key as the json structure changed but want to support old key for backwards compat reasons? Just don't want to limit the API today to one `fileDownloadMetadataKey` per message thread and find out we need one per message in the future

packages/react-components/src/theming/icons.tsx
anjulgarg: Fixed

packages/react-components/src/components/MessageThread.tsx
anjulgarg: Fair point. One way to do it would be to embed the key information in `message.metadata` itself.
```tsx
metadata = {
  fileDownloadMetadataKey: "fileDownloads",
  fileDownloads: []
}
```

anjulgarg: But this makes the concept much more harder for Contoso to grasp.

packages/react-components/src/theming/icons.tsx
```diff
@@ -74,6 +78,10 @@ export const DEFAULT_COMPONENT_ICONS = {
   ControlButtonParticipants: <People20Filled />,
   ControlButtonScreenShareStart: <ShareScreenStart20Filled />,
   ControlButtonScreenShareStop: <ShareScreenStop20Filled />,
+  /* @conditional-compile-remove(file-sharing) */
+  Cancel: <Dismiss16Regular />,
+  /* @conditional-compile-remove(file-sharing) */
+  Download: <ArrowDown16Regular />,
```
JamesBurnside, line 84: I think this should be `<ArrowDownload16Regular>`: 
https://iconcloud.design/browse/Fluent%20System%20Library/Fluent%20Regular/60e878c1e-3821b375b

Instead of `ArrowDown16Regular`:
https://iconcloud.design/browse/Fluent%20System%20Library/Fluent%20Regular/80d07ae9f-3821b375b

anjulgarg: Oh what would I do without you @JamesBurnside :D

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -0,0 +1,174 @@
...
+        try {
+          const response = await props.downloadHandler(userId, file);
+          setShowSpinner(false);
+          if (response instanceof URL) {
+            window.open(response.toString(), '_blank', 'noopener,noreferrer');
+          } else {
+            props.onDownloadErrorMessage && props.onDownloadErrorMessage(response.errorMessage);
+          }
+        } catch (error) {
+          console.error(error);
```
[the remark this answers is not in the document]
anjulgarg: Ideally not. The expectation is for contoso to handle all the errors in their function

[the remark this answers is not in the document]
anjulgarg: The exception is not part of the error message that needs to be displayed to the Users.
Contoso might not want to display internal errors to users.

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -572,6 +574,20 @@ export type MessageThreadProps = {
    * Optional strings to override in component
    */
   strings?: Partial<MessageThreadStrings>;
+
+  /* @conditional-compile-remove(file-sharing) */
+  /**
+   * @beta
+   * Optional function called when someone clicks on the file download icon.
+   */
+  fileDownloadHandler?: FileDownloadHandler;
```
prabhjot-msft, line 583: Why are we adding fileDownloadHandler as a separate prop when it is being handled by onRenderFileDownload Prop?

packages/react-components/src/components/MessageThread.tsx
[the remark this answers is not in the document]
anjulgarg: `Property name that contains information about file downloads in `message.metadata` object.`
Contoso can use any key to set file download metadata in message.metadata.
They need to provide the name of that key using this prop.
for example
```tsx
metadata = {
  someContosoKey: [ ...FileMetaData ]
}
```

So, `fileDownloadMetadataKey` would be `someContosoKey`

anjulgarg: @PorterNan @prprabhu-ms thoughts?

JamesBurnside: Would also love to get @PorterNan or @prprabhu-ms input here, my feeling is a download is contained inside a message so I feel it make more sense for the message props to encapsulate the key to retrieve the download from the message data.
I think I would have it at the `ChatMessage` interface level:
```tsx
interface ChatMessage extends MessageCommon {
  messageType: 'chat';
  content?: string;
  contentType: MessageContentType;
  ...
  metadata?: Record<string, string>;
  // key inside the metadata field that houses file attachment information
  fileMetadataKey: string;

  // in theory other attachments could then be added:
  powerpointMetadatKey; // specifically a onedrive hosted powerpoint link to embed a live preview in message
  arcadeMetadata; // information on a arcade component that is to be live embedded in a chat message
  // ^ though perhaps at this point its not a ChatMessage and its something larger 
}
```

packages/react-composites/src/composites/common/icons.tsx
```diff
@@ -49,11 +49,11 @@ const coffeeIconStyle = {
  */
 export const COMPOSITE_ONLY_ICONS: CompositeIcons = {
   /* @conditional-compile-remove(file-sharing) */
-  Cancel: <FontIcon iconName="Cancel" />,
+  CancelFileUpload: <Dismiss16Regular />,
```
JamesBurnside, line 52: Can this be removed not that they live in the component (in react-components)?

packages/react-components/src/components/MessageThread.tsx
anjulgarg: In contrast to changing the interface.
I would recommend storing the `fileMetadataKey` in the `message.metadata` itself.
@JamesBurnside keys other than `metadata` won't be persisted with the message. The info will be lost when the message is sent...

JamesBurnside, line 590: semantic nit: the file isn't a fileDownload currently, it's a fileAttachment (but would rather keep fileDownload if it drives consistency)

anjulgarg: Let's see what others have to say about this API and we can revisit the name basis on what path we take :)

JamesBurnside: There's going to have to be a contract in place in Contoso's application that defines what the metadata key is regardless, so I think that information is lose-able when messages are sent.

If the fileMetadataKey is stored in message.metadata, the receiver will need to know to lookup `fileMetadataKey` as a fixed prop anyway (so shifts the contract to being defined by us instead of consoto, at which point we may as well have a fixed fileMetadataKey that is always called fileMetadataKey and not let contoso use a custom key)

PorterNan: Is this file download metadata syntax defined by us or our chat sdk?

If that is defined by us, it would be great to add a "type" property into the metadata type, and do some typescript magic
If type="fileDownload" then we limit the metadata to be file metadata type, so we are good to put the key into metadata

anjulgarg: @PorterNan I love that idea. It can be done.

JamesBurnside: > How does that handle the scenario where each message has different key for attached files if the contract is statically defined in Contoso's application?

Not a hardcoded constant, just a contract between contoso applications. There could be if-else's in the code to choose different keys, or perhaps a service updates that key, but agree that is a contrived example.

I think I see message thread a our container for chat messages, it arranges and styles messages, but chat messages are isolated components that should be able to exist outside of a message thread. We don't export a ChatMessage component so we don't have to support that, but in terms of component encapsulation it should be possible to have a chat message with a file download elsewhere. Imagine if we implemented the full chat notifications and allowed a download button inside one of these:
image

packages/communication-react/review/communication-react.api.md
```diff
@@ -959,8 +959,6 @@ export type ChatCompositeIcons = {
     SendBoxSend?: JSX.Element;
     SendBoxSendHovered?: JSX.Element;
     SendBoxAttachFile?: JSX.Element;
-    Download?: JSX.Element;
-    Cancel?: JSX.Element;
```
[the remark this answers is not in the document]
JamesBurnside: these are beta only, they're being moved to the COMPONENT_ICONS instead of COMPOSITE_ICONS

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/**
+ * @beta
+ *
+ * A callback function for handling file downloads.
+ * The function needs to return a promise that resolves to a file download URL.
+ * If the promise is rejected, the {@link Error.message} will be used to display an error message to the user.
+ *
+ * Sample Usage:
```
alkwa-msft, line 48: Is it appropriate to have this sample code in the code?

packages/react-components/src/components/FileDownloadCards.tsx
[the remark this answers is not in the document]
alkwa-msft: Would it not seem more appropriate to throw this error? rather than console.error?
We only use console.error in storybook + samples (things we own)

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -38,15 +39,13 @@ import { participantListContainerPadding } from '../common/styles/ParticipantCon
 import { ChatScreenPeoplePane } from './ChatScreenPeoplePane';
 import { toFlatCommunicationIdentifier } from '@internal/acs-ui-common';
 /* @conditional-compile-remove(file-sharing) */
-import { FileDownloadCards } from './FileDownloadCards';
-/* @conditional-compile-remove(file-sharing) */
 import { fileUploadsSelector } from './selectors/fileUploadsSelector';
 /* @conditional-compile-remove(file-sharing) */
 import { useSelector } from './hooks/useSelector';
 /* @conditional-compile-remove(file-sharing) */
-import { useState } from 'react';
```
alkwa-msft, line 47: we remove this import and then we are using it directly?

packages/react-components/src/components/FileDownloadCards.tsx
anjulgarg: Quite a common practice in a lot of open source libraries :)

packages/react-components/src/components/FileDownloadCards.tsx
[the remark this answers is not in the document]
anjulgarg: I think you and nan are right. Updated the code

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
anjulgarg: To reduce the surface of conditionally compiled code.
Since `React` is already in scope. importing `useState` separately means that we now have to conditionally remove it.
`React.useState` saves us that trouble

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -46,6 +49,17 @@ const memoizedAllConvertChatMessage = memoizeFnAll(
...
 );
 
+/* @conditional-compile-remove(file-sharing) */
+const extractAttachedFilesMetadata = (metadata: Record<string, string>): FileMetadata[] => {
+  const fileMetadata = metadata['fileSharingMetadata'];
+  if (!fileMetadata) return [];
+  try {
+    return JSON.parse(fileMetadata);
+  } catch (e) {
+    return [];
```
JamesBurnside, line 61: can we console log the error here? I'm always scared of simply swallowing errors

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -46,6 +49,19 @@ const memoizedAllConvertChatMessage = memoizeFnAll(
   }
 );
 
+/* @conditional-compile-remove(file-sharing) */
+const extractAttachedFilesMetadata = (metadata: Record<string, string>): FileMetadata[] => {
+  const fileMetadata = metadata['fileSharingMetadata'];
```
JamesBurnside, line 54: nit: pull out `'fileSharingMetadata'` to a constant at the top of the file
```
const FILE_SHARING_METADATA_KEY = 'fileSharingMetadata';
```

packages/communication-react/review/communication-react.api.md
```diff
@@ -1932,6 +1934,7 @@ export type MessageThreadProps = {
     onSendMessage?: (messageId: string) => Promise<void>;
     disableEditing?: boolean;
     strings?: Partial<MessageThreadStrings>;
+    fileDownloadHandler?: FileDownloadHandler;
```
JamesBurnside, line 1937: dont hate me but.. what if this was also on the ChatMessage interface?
(thinking of the idea of if a chat message could exist outside of the message thread)

anjulgarg: If contoso is willing to provide a different `fileDownloadHandler` function per message.
I am sure that logic can be wrapped inside a single `fileDownloadHandler` too. :P

JamesBurnside: Ignore this actually. You have already added this to the ChatMessageAsAWhatever which makes sense

packages/chat-component-bindings/src/messageThreadSelector.ts
anjulgarg: Oh yes.

packages/chat-component-bindings/src/messageThreadSelector.ts
anjulgarg: Trying to reduce the number of lines I have to conditionally remove.

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -0,0 +1,162 @@
...
```
JamesBurnside, line 47: nit: use [`@example`](https://tsdoc.org/pages/tags/example/) so inteliisense/docs things picks this up nicely

JamesBurnside (approved):

packages/chat-component-bindings/src/messageThreadSelector.ts
prprabhu-ms: `console.exception` so we know what caused the 'Failed to parse }" error.

prprabhu-ms (approved):

PR #1791, opened 2022-04-12
Filter read receipt and participant list to not include myself

carocao-msft:
# What
Filter read receipt and participant list to not include myself

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2811663


JamesBurnside (approved):

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -138,20 +138,24 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
     // Read Receipt details will be disabled when participant count is 0
     const participantCount = isTeamsInterop
```
JamesBurnside, line 139: nit: update variable name to indicate this is not the total participant count, but `participantCountNotIncludingSelf` or `otherParticipantsCount`

JamesBurnside: Hmmm, actually message thread property is:
image
Which indicates all participants including self

JamesBurnside: Perhaps message thread should also have a `remoteParticipantsCount`? @PorterNan @prprabhu-ms

prprabhu-ms: I'd vote against both of the options:
* Change the meaning of `participantCount` without rename (this PR as of now)
* Add another field `remoteParticipantCount` (why have both?)

How about we instead update the documentation for `participantCount` to note that it includes the local participant, and then subtract one in the component itself? I know this moves some business logic to the UI component, but backwards compatibility etc.

For `readReceipts`, we have two (complimentary) options:
* We already compute the counts in the component, so can update there to skip local user: https://github.com/Azure/communication-ui-library/blob/9af6a7a6a6b0f0b5fa3847161b199c6bf8622396/packages/react-components/src/components/utils/getParticipantsWhoHaveReadMessage.ts#L24
* Add fields for read count and add selectors for this. This is a (potentially) costly computation that currently happens in the component code repeatedly.

PorterNan: +1 to keep minor one in the component logic

1. participantCount will have its own meaning
2. That value will be still reusable in some other scenarioes in chatThread

PorterNan: For readReceipts - we need them both in Read x by y (only read count)
And detail readReceipts in contextual menu

I would say computing the readCount for each message is a huge, currently we don't do this computation for each messages. we've already done some optimizations that delay compute one of message only when you hover on one message, to compute it in selector is introducing computation back to every message which is not perf friendly

carocao-msft: I have updated the PR: keep the logic to filter the read receipt at selector level, removed the filtering for participant and revert it back to -1 @JamesBurnside @prprabhu-ms @PorterNan

JamesBurnside (approved):

PorterNan (approved):

PR #1793, opened 2022-04-12
Log the function called from stateful client

PorterNan:
# What

The stateful client is following these pattern

1. stateful client function call (usually triggered by UI) ->
2.  send request to service (done by sdk) ->
3.  get result -> 
4. pump it to state
While we've already log step 2 and 4, the initiate step 1 isn't logged, to clearly see what is happening, it is necessary to see when a stateful client function was called by users

# Why

# How Tested

# Process & policy checklist


change/@internal-calling-stateful-client-c3355bc5-9d82-4dca-8a07-ce9d1d0f8798.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
JamesBurnside, line 2: are these prerelease? As the change executes on stable builds

JamesBurnside (approved):

change/@internal-calling-stateful-client-c3355bc5-9d82-4dca-8a07-ce9d1d0f8798.json
anjulgarg: Shouldn't this be none?
Doesn't affect public api

packages/chat-stateful-client/src/ChatContext.ts
```diff
@@ -346,6 +347,7 @@ export class ChatContext {
   ): (...args: Args) => R {
     return (...args: Args): R => {
       try {
+        chatStatefulLogger.info(`Stateful client target function called: ${target}`);
```
anjulgarg, line 350: The above comment says `Calling stateful client`
and this comment says `Stateful client`
Do we want to make it consistent?

anjulgarg (approved):

PR #1794, opened 2022-04-12
[CallWithChat][A11y] Update label and indicate selected camera in LocalVideoCameraButton 

edwardlee-msft:
# What
Indicate which camera is currently selected in LocalVideoCameraButton using ariaDescription.
Update "Cycle camera" string to "Switch camera" as indicated by A11y bug.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2791846/

# How Tested
Test locally on MacOS using VoiceOver

# Process & policy checklist

### DIFF
--- a/packages/communication-react/stable-review/communication-react.api.md
+++ b/packages/communication-react/stable-review/communication-react.api.md
@@ -1670,6 +1670,7 @@ export type LocalizationProviderProps = {
 // @public (undocumented)
 export interface LocalVideoCameraCycleButtonProps {
     cameras?: OptionsDevice[];
+    description?: string;
     label?: string;
     onSelectCamera?: (device: OptionsDevice) => Promise<void>;
     selectedCamera?: OptionsDevice;
@@ -2398,6 +2399,7 @@ export interface VideoGalleryStrings {
     localVideoCameraSwitcherLabel: string;
     localVideoLabel: string;
     localVideoMovementLabel: string;
+    localVideoSelectedDescription: string;
     screenIsBeingSharedMessage: string;
     screenShareLoadingMessage: string;
 }
--- a/packages/communication-react/stable-review/communication-react.api.md
+++ b/packages/communication-react/stable-review/communication-react.api.md
@@ -1671,8 +1671,8 @@ export type LocalizationProviderProps = {
 
 // @public (undocumented)
 export interface LocalVideoCameraCycleButtonProps {
+    ariaDescription?: string;
     cameras?: OptionsDevice[];
-    description?: string;
     label?: string;
     onSelectCamera?: (device: OptionsDevice) => Promise<void>;
     selectedCamera?: OptionsDevice;
--- a/packages/react-components/src/components/VideoGallery.tsx
+++ b/packages/react-components/src/components/VideoGallery.tsx
@@ -67,6 +67,8 @@ export interface VideoGalleryStrings {
   localVideoCameraSwitcherLabel: string;
   /** String for announcing the local video tile can be moved by keyboard controls */
   localVideoMovementLabel: string;
+  /** String for announcing the selected camera */
+  localVideoSelectedDescription: string;
 }
 
 /**
@@ -238,6 +240,9 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
               selectedCamera={localVideoCameraCycleButtonProps.selectedCamera}
               onSelectCamera={localVideoCameraCycleButtonProps.onSelectCamera}
               label={strings.localVideoCameraSwitcherLabel}
+              description={
+                localVideoCameraCycleButtonProps.selectedCamera.name + ' ' + strings.localVideoSelectedDescription
+              }
             />
           )}
       </Stack>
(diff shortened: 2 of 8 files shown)


packages/communication-react/stable-review/communication-react.api.md
```diff
@@ -1672,6 +1672,7 @@ export type LocalizationProviderProps = {
 // @public (undocumented)
 export interface LocalVideoCameraCycleButtonProps {
     cameras?: OptionsDevice[];
+    description?: string;
```
JamesBurnside, line 1675: I think this should be `ariaDescription` instead of just description to be clear what this description is used for

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -238,6 +240,9 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
               selectedCamera={localVideoCameraCycleButtonProps.selectedCamera}
               onSelectCamera={localVideoCameraCycleButtonProps.onSelectCamera}
               label={strings.localVideoCameraSwitcherLabel}
+              description={
+                localVideoCameraCycleButtonProps.selectedCamera.name + ' ' + strings.localVideoSelectedDescription
```
JamesBurnside, line 244: Use string builder here: https://github.com/Azure/communication-ui-library/blob/4c0a64adec1576372ecc417ab2c82d55d40b55f6/packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx#L34

packages/communication-react/stable-review/communication-react.api.md
anjulgarg: @JamesBurnside we are using just `label` prop right below this one though.
Can we look at other components to see wht kind of convention we are following?

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -141,7 +141,8 @@
     "screenIsBeingSharedMessage": "You are sharing your screen",
     "screenShareLoadingMessage": "Loading {participant}'s screen",
     "localVideoLabel": "You",
-    "localVideoCameraSwitcherLabel": "Cycle camera",
-    "localVideoMovementLabel": "Movable Local Video Tile"
+    "localVideoCameraSwitcherLabel": "Switch camera",
+    "localVideoMovementLabel": "Movable Local Video Tile",
+    "localVideoSelectedDescription": "selected"
```
anjulgarg, line 146: nit: Should `s` be capital?

anjulgarg (approved):

packages/react-components/src/localization/locales/en-US/strings.json
edwardlee-msft: The string comes after the camera name. 
Lowercase makes sense in this case. However, we can change it to uppercase for consistency. What do you think?

Would look like:
<Camera name> **selected**.
vs.
<Camera name> **Selected**.

packages/communication-react/stable-review/communication-react.api.md
JamesBurnside: Good point. For control bar buttons we definitely use label, but they have a UI component that shows the label (we just use that for the aria-label for simplicity), so this component should probably not have had `label` but instead had `ariaLabel` (too late to change now).

My vote would be for `ariaDescription` but happy to keep description if we mention in the doc comment this prop is used for the aria-description

packages/react-components/src/localization/locales/en-US/strings.json
JamesBurnside: I think lowercase will make more sense with the string builder as this string will become something like `{cameraName} selected`

packages/communication-react/stable-review/communication-react.api.md
edwardlee-msft: Changed to ariaDescription :)

packages/react-components/src/components/VideoGallery.tsx
edwardlee-msft: Used string builder :)

packages/react-components/src/localization/locales/en-US/strings.json
edwardlee-msft: Will keep it as lowercase as it makes more sense in this case.

JamesBurnside, line 146: This string should have the `{cameraName}` inside it:
```suggestion
    "localVideoSelectedDescription": "{cameraName} selected"
```

It allows Contoso to understand this string is used in a string builder and how the string is built.
E.g. on L142 we have: `"screenShareLoadingMessage": "Loading {participant}'s screen",`

JamesBurnside: (TD must do some magic to make use of this, @mgamis-msft would have more info on that)

edwardlee-msft: Love the magic! Changed it up to use {cameraName} and changed up ariaDescription to follow suit.

JamesBurnside (approved):

packages/react-components/src/localization/locales/en-US/strings.json
JamesBurnside: Thanks for making those changes!

PR #1797, opened 2022-04-12
Added a min width of 320px to Message thread to render file cards completely in mobile view

prabhjot-msft:
# What
Added a min width of 320px to Message thread to render file cards completely in mobile view

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2806013

# How Tested

# Process & policy checklist


packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -31,7 +31,8 @@ export const messageThreadContainerStyle = mergeStyles({
   maxHeight: '100%',
   overflow: 'auto',
   position: 'relative',
-  alignSelf: 'center'
+  alignSelf: 'center',
+  minWidth: '320px'
```
JamesBurnside, line 35: Our current minWidth for the chat composite is 280px:
image
https://60c7ae6891f0e90039d7cd54-actiqxaqcq.chromatic.com/?path=/docs/composites-chat-basicexample--basic-example

Can we reduce the space the file cards need?

JamesBurnside (changes requested):

packages/react-components/src/components/styles/MessageThread.styles.ts
anjulgarg: @pereiralex thoughts?

anjulgarg: This dimension came from AlexP.
@prabhjot-msft let's reach out to AlexP and get his opinion here.

JamesBurnside: note: for context this was reduced to support the galaxy fold

PR #1800, opened 2022-04-13
Responsive contextual menu

mgamis-msft:
# What

Added utility hooks to observe height and width of containers of `CallControlBar` and `CallWithChatControlBar` in `CallComposite` and `CallWithChatComposite`, respectively.
Added ContainerRectProps to props of  `CallControlBar` and `CallWithChatControlBar` so that changes to container height and width will cause a rerender to buttons with contextual menus.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2794403

# How Tested

Tested local samples
CallWithChat:
https://user-images.githubusercontent.com/79475487/163097116-74c826cf-aff1-4bc4-b3ce-86de85a7af8b.mp4

Calling:
https://user-images.githubusercontent.com/79475487/163097149-d1b04d2f-91ef-4e10-b9b9-b431e6f97011.mp4

# Process & policy checklist


packages/react-composites/src/composites/common/responsive.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Returns updated width if parent/window resizes.
+ * @param containerRef - Ref of a parent element whose height will be returned.
+ * @private
+ */
+export const useContainerWidth = (containerRef: RefObject<HTMLDivElement>): number | undefined => {
+  const [width, setWidth] = useState<number | undefined>(undefined);
+
+  const observer = useRef(
+    new ResizeObserver((entries) => {
+      const { height } = entries[0].contentRect;
```
anjulgarg, line 47: Shouldn't this be width?
Otherwise the entire logic in useContainerHeight and this function is similar.

anjulgarg:
I don't understand what's happening in the demo video attached. Since it doesn't relate with the logic in this PR.
How is this piece of code making contextual menu responsive?
image

packages/react-composites/src/composites/common/responsive.ts
mgamis-msft: Thanks for catching that. Will fix this.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -31,13 +32,21 @@ export type CallControlsProps = {
...
  */
-export const CallControls = (props: CallControlsProps): JSX.Element => {
+export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX.Element => {
   const options = useMemo(() => (typeof props.options === 'boolean' ? {} : props.options), [props.options]);
   const customButtons = useMemo(
     () => generateCustomButtons(onFetchCustomButtonPropsTrampoline(options), options?.displayType),
     [options]
   );
 
+  // This useState and useEffect is to rerender the control bar when the container height and/or width change
```
JamesBurnside, line 42: Is this necessary? My understanding is if the props passed in are changed the component should re-render (perhaps react is smart and avoids the re-render if the props are unused inside the component..)

mgamis-msft: Yes. React is smart enough. Thanks

anjulgarg (approved):

edwardlee-msft (approved):

PR #1801, opened 2022-04-13
Added UI test for Chat unread badge 

carocao-msft:
# What
Added ui test for CallwithChat unread message badge
This test includes testing for <9 and >9 unread messages
To prevent flakiness (sometimes there are 10 unread messages but the badge didn't finish updating before the UI test take a snapshot, so it displays <10 unread messages in the snapshot.), to test >9 unread messages scenario, sent 10 messages before taking a screenshot.

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2780032


packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,61 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+    await sendMessage(pages[0], 'Another message!');
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+    await sendMessage(pages[0], 'Another message!');
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+    await sendMessage(pages[0], 'Another message!');
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+    await sendMessage(pages[0], 'Another message!');
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+    await sendMessage(pages[0], 'Another message!');
+    await waitForMessageDelivered(pages[0]);
```
JamesBurnside, line 113: Looking at the implementation for this it will only wait for any message to have been delivered and not all of these messages:
https://github.com/Azure/communication-ui-library/blob/536ac874e857a7570525b44ef4584e746f051070/packages/react-composites/tests/browser/common/chatTestHelpers.ts#L66

Is there a way we can check a specific message has been delivered? (i.e. the last one)
Or perhaps we can ensure there is _only_ a 'delivered'/'seen' icon and no 'sending' icon?

JamesBurnside: Otherwise each `sendMessage` should be `sendMessageAndWaitForDelievered` to ensure each message delivers successfully

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,61 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+    await pages[1].waitForTimeout(1000); // ensure badge has had time to appear
+    await waitForSelector(pages[1], dataUiId('call-with-chat-composite-chat-button-unread-icon'));
+    expect(await pages[1].screenshot()).toMatchSnapshot(`call-with-chat-gallery-screen-with-one-unread-messages.png`);
+  });
+
+  test('Unread chat message button badge are displayed correctly for >9 messages', async ({ pages }, testInfo) => {
+    // Open chat pane on page 0 and send 16 messages
+    await pageClick(pages[0], dataUiId('call-with-chat-composite-chat-button'));
+    await waitForSelector(pages[0], dataUiId('call-with-chat-composite-chat-pane'));
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
```
JamesBurnside, line 97: nit: for loop? ;)

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,61 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+    await sendMessage(pages[0], 'Another message!');
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+    await sendMessage(pages[0], 'Another message!');
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+    await sendMessage(pages[0], 'Another message!');
+    await waitForMessageDelivered(pages[0]);
+
+    // Ensure typing indicator has disappeared to prevent flakey test
+    await pages[0].bringToFront();
+    const typingIndicator = await pages[0].$(dataUiId(IDS.typingIndicator));
```
JamesBurnside, line 117: We should be checking on page1 here as we are waiting for user0's typing indicator to have disappeared

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,61 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+    typingIndicator && (await typingIndicator.waitForElementState('hidden'));
+
+    if (!isTestProfileDesktop(testInfo)) {
+      await waitForPiPiPToHaveLoaded(pages[0], 2);
```
JamesBurnside, line 121: This check doesn't need to happen as we're not taking a screenshot of this page

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,61 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+    // Open chat pane on page 0 and send 16 messages
+    await pageClick(pages[0], dataUiId('call-with-chat-composite-chat-button'));
+    await waitForSelector(pages[0], dataUiId('call-with-chat-composite-chat-pane'));
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+    await sendMessage(pages[0], 'Another message!');
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+    await sendMessage(pages[0], 'Another message!');
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+    await sendMessage(pages[0], 'Another message!');
+    await sendMessage(pages[0], 'Call with Chat composite is awesome!');
```
dmceachernmsft, line 103: to cut down on time waiting for these messages should we only send 10?

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,48 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+      // timeout between each messages to prevent chat throttling
+      await pages[0].waitForTimeout(100);
+    }
+
+    // Ensure typing indicator has disappeared to prevent flakey test
+    await pages[1].bringToFront();
+    const typingIndicator = await pages[1].$(dataUiId(IDS.typingIndicator));
+    typingIndicator && (await typingIndicator.waitForElementState('hidden'));
+
+    await waitForMessageSeen(pages[0]);
```
dmceachernmsft, line 104: Is this `waitForMessagesSeen` function supposed to be about the messages being seen by the recipient of the messages? or is this one about the message sent indicator?

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,48 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+    let seenMessages = await pages[0].$$('[data-ui-status="seen"]');
```
dmceachernmsft, line 105: should this maybe be named `recievedMessages`?

dmceachernmsft (approved):

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
carocao-msft: In this test even though page[1] didn't see the message, the messages sent by pages[0] arrives at the seen state. This waitForMessagesSeen is just another layer of prevention against flaky tests to make sure there are seen icons in page[0]

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
carocao-msft: Same as above,  the messages sent by pages[0] arrives at the seen state even though they are not seen by page[1]

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,48 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+    // make sure all messages are delivered and received before taking a screenshot
+    while (seenMessages.length < 10) {
```
JamesBurnside, line 107: this does not have a timeout so could get stuck in an infinite loop if there are never `>=10` .
Instead can you use the `waitForFunction`?

waitForFunction sends a function for the headless browser page to execute, and continually sends that function (at some small interval) until it returns true or the global timeout elapses.

something like:
```tsx
waitForFunction(page, () => {
  // fun fact: everything inside this function body is executed by the headless browser page and not the playwright node process
  const seenMessages = document.querySelectorAll(`[data-ui-status="seen"]`);
  return seenMessages.length >= 10
});
```

Or you make a generic function for other tests to use:
```tsx
const waitForNSeenMessages(n: number): Promise<void> => {
  return await waitForFunction(page, (args) => {
  // fun fact: everything inside this function body is executed by the headless browser page and not the playwright node process
  const seenMessages = document.querySelectorAll(`[data-ui-status="seen"]`);
  return seenMessages.length >= args.seenMessageCountToLookFor
},
// args parameter to pass from the node process to the headless browser page
{
  seenMessageCountToLookFor: n
});
};

```

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,48 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+
+  test('Unread chat message button badge are displayed correctly for >9 messages', async ({ pages }) => {
+    // Open chat pane on page 0 and send 10 messages
+    await pageClick(pages[0], dataUiId('call-with-chat-composite-chat-button'));
+    await waitForSelector(pages[0], dataUiId('call-with-chat-composite-chat-pane'));
+
+    for (let i = 0; i < 10; i++) {
+      await sendMessage(pages[0], 'Call with Chat composite is awesome!');
+      // timeout between each messages to prevent chat throttling
+      await pages[0].waitForTimeout(100);
```
JamesBurnside, line 96: Skeptical 🤨 but I'll soft allow it 😄
Interested though if `await waitForNSeenMessages(i);` could be used here instead of waitForTimeout (see other comment for an implementation of waitForNSeenMessages)

packages/react-composites/tests/browser/callwithchat/CallWithChatComposite.test.ts
```diff
@@ -70,6 +70,48 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
+    }
+
+    // Ensure typing indicator has disappeared to prevent flakey test
```
JamesBurnside, line 99: nit: use your helper fn you added for this!

JamesBurnside: `waitForTypingIndicatorHidden`

JamesBurnside (approved):

PR #1814, opened 2022-04-20
Allow sending blank message when a file is attached in sendbox

anjulgarg:
# What
Allow sending blank message when a file is attached in sendbox
Use deletedOn property of a message to detect deleted messages 

# Why
When a message contains a file, user should be able to send it without any text

# How Tested
Manually

# Process & policy checklist


packages/react-components/src/components/SendBox.tsx
```diff
@@ -233,12 +233,20 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
     }
 
     // we dont want to send empty messages including spaces, newlines, tabs
-    if (!EMPTY_MESSAGE_REGEX.test(textValue)) {
-      onSendMessage && onSendMessage(textValue);
+    // Message can be empty if there is a valid file upload
+    if (!EMPTY_MESSAGE_REGEX.test(textValue) || hasFile) {
+      // Chat SDK doesn't send messages with empty content. For sending files without text message,
```
anjulgarg, line 238: I don't have 100% confidence in this statement.
In conversation with the SDK team to understand the exact behavior.
This is based on my tests.

packages/react-components/src/components/SendBox.tsx
```diff
@@ -233,12 +233,20 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
...
 
     // we dont want to send empty messages including spaces, newlines, tabs
-    if (!EMPTY_MESSAGE_REGEX.test(textValue)) {
-      onSendMessage && onSendMessage(textValue);
+    // Message can be empty if there is a valid file upload
+    if (!EMPTY_MESSAGE_REGEX.test(textValue) || hasFile()) {
+      // Chat SDK doesn't send messages with empty content. For sending files without text message,
+      // we need to send a message with a space.
+      if (EMPTY_MESSAGE_REGEX.test(textValue)) {
+        onSendMessage && onSendMessage(' ');
```
PorterNan, line 241: quick question: should we send metadata from here(onSendMessage) with text all together? It looks more readable if user want to implement a file upload from component level

anjulgarg: Thought about it.
Metadata is not really a part of the message or sendbox from a pure UI perspective.
It is an underlying implementation detail that contoso needn't care about.

Also, if contoso were to use this component without composite and implement file sharing, they would again not need metadata here. :)

PorterNan: I am thinking of a scenario where contoso wants to implement this with component + stateful client (still use our acs), I believe the recommended from us is still using metadata right? I am thinking if we could provide an optional metadata prop in sendBox component, then sendMessage will be sending that metadata if user provides any
But yea, this is not related to this PR, let's discuss the topic when we are writing a fully functional component-level uploading sample for our contoso

packages/react-components/src/components/SendBox.tsx
```diff
@@ -233,12 +233,20 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
     }
 
     // we dont want to send empty messages including spaces, newlines, tabs
-    if (!EMPTY_MESSAGE_REGEX.test(textValue)) {
-      onSendMessage && onSendMessage(textValue);
+    // Message can be empty if there is a valid file upload
+    if (!EMPTY_MESSAGE_REGEX.test(textValue) || hasFile()) {
+      // Chat SDK doesn't allow sending messages with NULL content. For sending files without text message,
+      // we need to send a message with an empty string.
+      if (EMPTY_MESSAGE_REGEX.test(textValue)) {
```
PorterNan, line 240: I don't think we need if else from 240-242, textValue will be always '' if you don't put anything there
Or we can do
onSendMessage && onSendMessage(textValue ?? '');

anjulgarg: We want to avoid sending spaces (as per my understanding).
If we remove this if else check, someone would be able to send `'               '` which will make the message appear strange.
This logic ensures that if someone has entered a bunch of spaces, we only send blank message

prprabhu-ms: These conditionals have become a bit convoluted.

Helper method would make the flow easier to follow, I think.

```
function sendSanitizedText(sendFunction: (sanitizedText: string) => void, textValue: string, hasFile) {
  if (!EMPTY_MESSAGE_REGEX.test(textValue)) {
    sendFunction(textValue);
    return;
  }
  if (hasFile) {
    // Send an empty message if there is a valid file to send.
    sendFunction('');
  }
}

...

// and here in the onClick callback
onSendMessage && sendSanitizedText(onSendMessage, textValue, hasFile());
```

packages/react-components/src/components/SendBox.tsx
```diff
@@ -266,16 +274,23 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
   const mergedStyles = useMemo(() => concatStyleSets(styles), [styles]);
 
   const hasText = !!textValue;
+  const hasFile = (): boolean => {
```
prprabhu-ms, line 277: I would extract this as a file private helper (trampoline) function rather than creating a new lambda on each render.

prprabhu-ms (approved):

prprabhu-ms:
All minor comments. Approved.

Please re-assign if you need my attention on this PR again.

packages/react-components/src/components/SendBox.tsx
anjulgarg: Looks more complicated to me now. :D

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -188,7 +188,8 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
             // message.type === ACSKnownMessageType.topicUpdated ||
             message.clientMessageId !== undefined
         )
-        .filter((message) => message.content && message.content.message !== '') // TODO: deal with deleted message and remove
+        // Filter out deleted messages
```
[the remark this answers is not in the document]
anjulgarg: How does it not apply to the whole filter?
`.filter((message) => message.content && !message.deletedOn)` does just what it says
  // Filter out deleted messages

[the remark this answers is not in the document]
anjulgarg: We should not be checking for content here anymore. I will update the code

packages/react-components/src/components/SendBox.tsx
anjulgarg: That's a great point @PorterNan 
Can I create a fast follow up PR for this?

PorterNan (approved):

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -217,3 +217,13 @@ const sanitizedMessageContentType = (type: string): MessageContentType => {
...
+
+const messagesWithContentOrFileSharingMetadata = (message: ChatMessageWithStatus) => {
+  if (message.deletedOn) {
+    return false;
+  }
+  if (message.metadata?.['fileSharingMetadata']) {
+    return true;
+  }
+  return message.content?.message !== '';
+};
```
PorterNan, line 229: This condition is a bit different from previous one - the message without content will be still shown
We should also hide those messgaes with message.content===undefined

PorterNan: Especially in teams interop scenarioes, teams is sending a lot noise message(at least we don't handle them for now) which message.content===undeifined

PR #1818, opened 2022-04-21
Update the right call.id into state

PorterNan:
# What

Update the call.id in state when call id gets changed
# Why

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -168,6 +168,7 @@ export class CallContext {
   public setCallId(newCallId: string, oldCallId: string): void {
     this.modifyState((draft: CallClientState) => {
       const call = draft.calls[oldCallId];
+      call.id = newCallId;
```
alkwa-msft, line 171: So this means the Call object that we get may not be the most updated call object and we want to apply on the correct Id before we set the Call object to the correct key?

alkwa-msft (approved):

packages/calling-stateful-client/src/CallContext.ts
PorterNan: The id value in state.call.id has never been updated before this fix... Although we only use it in a very extreme case

Since this will be run at for the first callIdChange event, so this code will execute ahead of all the other state change, we should be good to assume customer will always get the right value in state

prprabhu-ms (approved):

prprabhu-ms:
Unassigning to remove from attention set. Please reassign if you need another review.

PR #1820, opened 2022-04-21
Fix whole composite rerender

PorterNan:
# What

Fix whole composite re-render when the callId gets changed

The callId change event will happen first in stateful client then in adapter, then it generates a small gap that with 2 conflict values in stateful client and adapter, which will make call object totally undefined during this gap.

In this PR we try to sync call.id directly from sdk original call object when there is a state change, which ensure adapter can always find the right call id value.

The diagram to explain why callId gets out of sync(before the fix):
image

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -170,6 +170,11 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
         callClient.offStateChange(onStateChange);
         return;
       }
+      // Always use current call.id value to access client state
+      // Since call.id could be change during the call
+      if (this.call?.id) {
```
alkwa-msft, line 175: does this mean whenever we have a state change and there is a call.id then we should set the callId? This sounds like it would potentially fire way more often and cause even more re-renders than only when the id has changed?

prprabhu-ms:
Please add *a lot* of context in the PR description. This is a very hard bug to understand and I'm sure I'll come digging git history to this PR 6 months from today :)

I would even soft-suggest leaving a reference to the PR in the comment in the code.

prprabhu-ms:
Hey, any blockers for this PR?

PorterNan:
> Hey, any blockers for this PR?

Just updated the PR, it's good to go now :D

alkwa-msft (approved):

prprabhu-ms (approved):

PR #1821, opened 2022-04-21
Rename chatComposite userId prop

dmceachernmsft:
# What

Rename `userid` prop to `userIdentifier`
# Why

Helps with clarity around naming of the chat composite adapter props. 

https://skype.visualstudio.com/SPOOL/_workitems/edit/2804059
# How Tested

Ran Chat sample and CallWithChat sample to make sure they work

Checked story book chat snippets to make sure they match

### DIFF
--- a/packages/communication-react/stable-review/communication-react.api.md
+++ b/packages/communication-react/stable-review/communication-react.api.md
@@ -2210,7 +2210,7 @@ export interface StatefulChatClient extends ChatClient {
 
 // @public
 export type StatefulChatClientArgs = {
-    userIdentifier: CommunicationUserIdentifier;
+    userId: CommunicationUserIdentifier;
     displayName: string;
     endpoint: string;
     credential: CommunicationTokenCredential;
--- a/packages/communication-react/stable-review/communication-react.api.md
+++ b/packages/communication-react/stable-review/communication-react.api.md
@@ -144,7 +144,7 @@ export type AzureCommunicationCallWithChatAdapterFromClientArgs = {
 // @public
 export type AzureCommunicationChatAdapterArgs = {
     endpoint: string;
-    userId: CommunicationUserIdentifier;
+    userIdentifier: CommunicationUserIdentifier;
     displayName: string;
     credential: CommunicationTokenCredential;
     threadId: string;
@@ -901,7 +901,7 @@ export const ChatComposite: (props: ChatCompositeProps) => JSX.Element;
 
 // @public
 export type ChatCompositeClientState = {
-    userId: CommunicationIdentifierKind;
+    userIdentifier: CommunicationIdentifierKind;
     displayName: string;
     thread: ChatThreadClientState;
     latestErrors: AdapterErrors;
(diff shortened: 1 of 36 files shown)


packages/communication-react/stable-review/communication-react.api.md
```diff
@@ -144,7 +144,7 @@ export type AzureCommunicationCallWithChatAdapterFromClientArgs = {
 // @public
 export type AzureCommunicationChatAdapterArgs = {
     endpoint: string;
-    userId: CommunicationUserIdentifier;
+    userIdentifier: CommunicationUserIdentifier;
```
JamesBurnside, line 147: we unfortunately cannot change this here as it will break the stable API. I think it's best to constrain these changes to storybook only

samples/Chat/src/app/ChatScreen.tsx
```diff
@@ -43,28 +43,28 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
     },
-    [endChatHandler, userId]
+    [endChatHandler, userIdentifier]
   );
 
   const adapterArgs = useMemo(
     () => ({
       endpoint: endpointUrl,
-      userId: fromFlatCommunicationIdentifier(userId) as CommunicationUserIdentifier,
+      userIdentifier: fromFlatCommunicationIdentifier(userIdentifier) as CommunicationUserIdentifier,
```
JamesBurnside, line 62: I think part of the confusion from reading the original bug is that we had `userId` as a string and `userId` as a CommunicationUserIdentifier.
I think anywhere a string a is used, userId is fine.
Anywhere `CommunicationIdentifier` is used it should be `userIdentifier` or `communicationUserIdentifier` (as userIdentifier just is userId expanded)

JamesBurnside: Alternatively anywhere the string is used could be `identity` or `userIdentity` as then it more closely matches the azure portal:

image

packages/communication-react/stable-review/communication-react.api.md
dmceachernmsft: ah okay this was something I was wondering, this seemed like a pretty major change for a bug.

samples/Chat/src/app/ChatScreen.tsx
dmceachernmsft: to resurface this are we wanting to actually rename this? its breaking change and by the logic of the comment above to rename the `CommunicationIdentifierKind` types we need to change the names on the `AzureCommunicationsChatAdapter` which is a major change.

JamesBurnside: This comment may not be left in the correct space, but renaming the storybook props is not a breaking change -- renaming an arguments/props into the composite interface is a breaking change.
My understanding of what we can achieve without a breaking change is simply ensure in storybook we never have two variables with the same name being two different types.

dmceachernmsft:
Closing this as proper fix is up

PR #1822, opened 2022-04-21
Display name in Video Gallery override but better.

dmceachernmsft:
# What

Media gallery will edit remote user's displaynames 
# Why

Fix issue where remote participants `displayName` are not edited when a new one is provided using the custom data model.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2694662
# How Tested

Ran locally with custom data model provided.
image


anjulgarg:
@dmceachernmsft Joshua Weissman fan?

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -60,6 +65,25 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
     };
   }, [cameraSwitcherCallback, cameraSwitcherCameras]);
 
+  // Async function to retrieve custom data model information to replace display name in remote tiles with camera on.
+  const fetchAvatarPersonaDataAsync = memoizeFunction(
+    async (participant: VideoGalleryRemoteParticipant): Promise<void> => {
+      if (props.onFetchAvatarPersonaData) {
+        const newParticipantData = await props.onFetchAvatarPersonaData(participant.userId);
+        participant.displayName = newParticipantData.text ? newParticipantData.text : participant.displayName;
```
JamesBurnside, line 73: Two things here:
1. This won't trigger a re-render to update the display name. It may be working when testing but only because other things will cause rerenders to occur
2. This mutates the participant in `videoGalleryProps.remoteParticipants` which is bad practice. (Mutating variables makes code very difficult to understand and debug as values could have changed randomly at any time, by anyone)

Instead consider having the merged participants in their own state in this component that can trigger re-renders:
```tsx
const [participants, setParticipants] = useState<VideoGalleryRemoteParticipant[]>([]);
```

Then when the participants update, fetch their props asynchronously:
```tsx
useEffect(() => {
  // in async block:
  const augmentedParticipants = await generateVideoParticipantsFromCustomData(videoGalleryProps.remoteParticipants, props.onFetchAvatarPersonaData);
  setParticipants(augmentedParticipants);

  // **note:** This is just a quick example but a poor implementation. If two renders are triggered due to 
  // remoteParticipants changing, there's no guarantee in what order the setParticipants will occur
}, [videoGalleryProps.remoteParticipants]
```

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -60,11 +65,57 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
+
+  // Async function to retrieve custom data model information to replace display name in remote tiles with camera on.
+  const fetchAvatarPersonaDataAsync = memoizeFunction(
+    async (
+      onFetchAvatarPersonaData: AvatarPersonaDataCallback,
+      participants: VideoGalleryRemoteParticipant[]
+    ): Promise<VideoGalleryRemoteParticipant[] | undefined> => {
+      if (props.onFetchAvatarPersonaData) {
+        participants.forEach(async (participant: VideoGalleryRemoteParticipant) => {
+          const newParticipantData = await onFetchAvatarPersonaData(participant.userId);
```
anjulgarg, line 80: - async/await doesn't work inside `forEach` loop. In this case it's just firing off multiple async calls but not really awaiting their results.
- You can use `for (const participant of participants)` instead.
- Better yet, use `Promise.all` which is much better than awaiting each async invocation
```tsx
await Promise.all(participants.map(async (participant) => {
  const newParticipantData = await onFetchAvatarPersonaData(participant.userId);
  participant.displayName = newParticipantData.text ? newParticipantData.text : participant.displayName;
}));
```

anjulgarg: And +1 to what @JamesBurnside  said. Mutating state variables inside components goes against our unidirectional data flow.
Better to introduce a component local state variable for `augmentedParticipants`

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -60,11 +65,58 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
+
+  // if we have the onFetchAvatarPersonaData callback set go through and edit the remote participant data.
+  useEffect(() => {
+    // flag to stop race conditions caused by participants joining the call
+    let newestFetch = true;
+    const augmentDisplayName = async (): Promise<void> => {
+      if (props.onFetchAvatarPersonaData) {
+        const tempParticipants = await fetchAvatarPersonaDataAsync(
+          props.onFetchAvatarPersonaData,
+          videoGalleryProps.remoteParticipants as VideoGalleryRemoteParticipant[]
```
anjulgarg, line 99: No need to typecast here. 
`videoGalleryProps.remoteParticipants` should be of time `VideoGalleryRemoteParticipant[]`

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -60,11 +65,58 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
+    undefined
+  );
+
+  // Async function to retrieve custom data model information to replace display name in remote tiles with camera on.
+  const fetchAvatarPersonaDataAsync = memoizeFunction(
+    async (
+      onFetchAvatarPersonaData: AvatarPersonaDataCallback,
+      participants: VideoGalleryRemoteParticipant[]
+    ): Promise<VideoGalleryRemoteParticipant[] | undefined> => {
+      if (props.onFetchAvatarPersonaData) {
```
anjulgarg, line 78: Why are we checking `props.onFetchAvatarPersonaData` here when `onFetchAvatarPersonaData` is a function argument?

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -60,11 +65,58 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
     };
   }, [cameraSwitcherCallback, cameraSwitcherCameras]);
 
+  const [augmentedParticipants, setAugmentedParticipants] = useState<VideoGalleryRemoteParticipant[] | undefined>(
+    undefined
+  );
+
+  // Async function to retrieve custom data model information to replace display name in remote tiles with camera on.
+  const fetchAvatarPersonaDataAsync = memoizeFunction(
```
anjulgarg, line 73: Let's move this function out of the `MediaGallery` component and avoid creating a new memoized lambda on every re-render

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -60,11 +65,58 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
+        return participants;
+      }
+      return;
+    }
+  );
+
+  // if we have the onFetchAvatarPersonaData callback set go through and edit the remote participant data.
+  useEffect(() => {
+    // flag to stop race conditions caused by participants joining the call
+    let newestFetch = true;
```
anjulgarg, line 94: I don't understand this `newestFetch` flag.
Seems like weird magic to me :D

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
dmceachernmsft: If you don't type cast here it thinks its type `never` for some reason

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
dmceachernmsft: good point 😅 will update

anjulgarg (approved):

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -184,3 +186,28 @@ export const computeVariant = (
     return 'NO_STATE';
   }
 };
+
+/**
+ * Edits the display name of the reomte participant based on the onFetchAvatarPersonaData function given to the call composite
```
prprabhu-ms, line 191: typo: reomte

prprabhu-ms:
I have some performance related comments so assigning myself. Will talk through this IRL after stand-up.

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -117,3 +129,29 @@ export const useLocalVideoStartTrigger = (isLocalVideoAvailable: boolean, should
     }
   }, [shouldTransition, isButtonStatusSynced, isPreviewCameraOn, isLocalVideoAvailable, mediaGalleryHandlers]);
 };
+
+/**
+ * Hook to fetch new participant avatar and video gallery displayName information if it is present.
+ * @param onFetchAvatarPersonaData
+ */
+const useOnFetchAvatarPersonaData = (
```
prprabhu-ms, line 137: Rename to better reflect what this does: `useRemoteParticipantsWithCustomDisplayNames` (or something shorter if you can think one up).

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -117,3 +129,29 @@ export const useLocalVideoStartTrigger = (isLocalVideoAvailable: boolean, should
...
+  remoteParticipants: VideoGalleryRemoteParticipant[],
+  onFetchAvatarPersonaData?: AvatarPersonaDataCallback
+): VideoGalleryRemoteParticipant[] => {
+  // Sort the remote participants so that they are always in order
+  remoteParticipants.sort((a, b) => (a.userId < b.userId ? 1 : -1));
```
prprabhu-ms, line 142: I'd recommend against reordering the `remoteParticipants`. This modifies an argument to the function in a surprising way. Side effects like this are hard to understand from the callsite.

I know we talked about sorting the `userIds` passed into `useDataProvider` to better memoize the results, but in this case we use the return value from `useDataProvider` to simply update the incoming `remoteParticipants`. So there is no need to sort the `userIds` here.

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -117,3 +129,29 @@ export const useLocalVideoStartTrigger = (isLocalVideoAvailable: boolean, should
...
+
+  const userIds = useMemo(() => {
+    return remoteParticipants.map((p) => p.userId);
+  }, [remoteParticipants]);
+
+  const avatarPersonaData = useDataProvider(userIds, onFetchAvatarPersonaData);
+  remoteParticipants.forEach((p, i) => {
```
prprabhu-ms, line 149: Reads better:

```
const newName = avatarPersonaData[i]?.text;
if (newName) {
  p.displayName = newName
}
```
Or you can chain list processing:

```
avatarPersonaData.map(data => data?.text).forEach((name, i) => name && remoteParticipants[i].displayName = name);
```

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -184,3 +186,28 @@ export const computeVariant = (
...
+ *
+ * @param onFetchAvatarPersonaData Callback function for mutating persona data
+ * @param participants Array of remote participants
+ * @returns Array of participants with their display name altered
+ *
+ * @private
+ */
+export const fetchAvatarPersonaDataAsync = async (
```
prprabhu-ms, line 199: Delete this function now?

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -87,3 +82,58 @@ export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
     />
   );
 };
+
+/**
+ * Hook to override the avatar persona data of the users
+ * @private
+ */
+export const useDataProvider = (
```
prprabhu-ms, line 90: Let's rename this to better reflect what it does: `useCustomAvatarPersonaData` or some such.

Also, should this move to a utils module. We'd want to use it each time we need custom avatar data anywhere.

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -87,3 +82,58 @@ export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
...
+  userIds: (string | undefined)[],
+  dataProvider?: AvatarPersonaDataCallback
```
prprabhu-ms, line 92: nit: rename `dataProvider`. Even just `callback` reads better than `dataProvider` to me.

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -87,3 +82,58 @@ export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
     />
   );
 };
+
+/**
+ * Hook to override the avatar persona data of the users
+ * @private
```
prprabhu-ms, line 88: We should document what this does a bit better:

* Calls `AvatarPersonaDataCallback` *on each render* for provided `userIds`.
* Returns an array of the same length as `userIds`. Entries in the array may be `undefined` if there is no data to return.

Aside: We currently return an empty list when `dataProvider` is not set. This sort of works in the uses so far, but we really should return an array of the same length as `userIds`, full of `undefined`. This ensures that caller can always zip the input and output arrays.

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -87,3 +82,58 @@ export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
...
+ */
+export const useDataProvider = (
+  userIds: (string | undefined)[],
+  dataProvider?: AvatarPersonaDataCallback
+): (AvatarPersonaData | undefined)[] => {
+  const [data, setData] = React.useState<(AvatarPersonaData | undefined)[]>([]);
+  useEffect(() => {
+    (async () => {
+      if (dataProvider) {
```
prprabhu-ms, line 97: Add early return to reduce nesting.

```
if (!dataProvider) {
  // Always return the same array full of no-data values.
  if (!data) {
    setData(userIds.map(() => undefined);
  }
  return;
}

... rest of the code without if block ...
```

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -87,3 +82,58 @@ export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
...
+        const newData = await Promise.all(
+          userIds.map(async (userId: string | undefined) => {
+            if (!userId) {
```
prprabhu-ms, line 100: This can be simplified to `return userId && await dataProvider(userId)`

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -87,3 +82,58 @@ export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
...
+ */
+const shouldUpdate = (
+  currentData: (AvatarPersonaData | undefined)[],
+  newData: (AvatarPersonaData | undefined)[]
+): boolean => {
+  let newDataPresent = false;
+  if (currentData.length !== newData.length) {
+    return true;
+  }
+  currentData.forEach((p, i) => {
```
prprabhu-ms, line 128: I'd recommend extracting a helper `avatarDeepEqual(a: AvatarPersonaData, b:AvatarPersonaData): boolean`.

This can then be simplified to a oneliner: `return currentData.some((data, i) => avatarDeepEqual(data, newData[i]));`

prprabhu-ms (commented):
Most comments are minor code readability nits.

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -117,3 +129,29 @@ export const useLocalVideoStartTrigger = (isLocalVideoAvailable: boolean, should
...
+    if (avatarPersonaData[i]?.text === undefined) {
+      return;
+    }
+    p.displayName = avatarPersonaData[i]?.text;
```
JamesBurnside, line 153: Similar to the side effects @prprabhu-ms mentioned above with sorting, this is mutating (changing) the displayName on the participant of the passed in remoteParticipants. This means the caller who uses `useOnFetchAvatarPersonaData` has just changed the participants display name for everyone using the remoteParticipants property.
Either this should be expected to mutate properties in place and not return anything (typically this is bad for usability, readability and debuggability of code)
Or (better) it should be returning a new set of remote participants with the augmentations in place (i.e. be a pure function with no side effects).

dmceachernmsft: okay will be sure to update this!

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -134,24 +135,22 @@ export const useLocalVideoStartTrigger = (isLocalVideoAvailable: boolean, should
...
  */
-const useOnFetchAvatarPersonaData = (
+const useRemoteParticipantsWithCustomDisplayNames = (
   remoteParticipants: VideoGalleryRemoteParticipant[],
   onFetchAvatarPersonaData?: AvatarPersonaDataCallback
 ): VideoGalleryRemoteParticipant[] => {
-  // Sort the remote participants so that they are always in order
-  remoteParticipants.sort((a, b) => (a.userId < b.userId ? 1 : -1));
-
+  const newParticipants = remoteParticipants;
```
prprabhu-ms, line 142: Assigning `remoteParticipants` doesn't copy the items in the list.

To ensure that you don't update arguments, you'd have to update the logic on line 148 to create a copy of each participatn. Something like 
```
remoteParticipants.map((participant, i) => {
   const newName = avatarPersonaData[i]?.text;
   if (!newName) {
      return participant;
   };
   return {...participant, displayName: newName}; 
});
```

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -188,26 +188,73 @@ export const computeVariant = (
...
-  if (onFetchAvatarPersonaData) {
-    await Promise.all(
-      participants.map(async (participant) => {
-        const newParticipantData = await onFetchAvatarPersonaData(participant.userId);
-        participant.displayName = newParticipantData.text ? newParticipantData.text : participant.displayName;
-      })
-    );
-    return participants;
+export const useCustomAvatarPersonaData = (
+  // move this to utils
```
prprabhu-ms, line 197: Remove TODO comment.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -188,26 +188,73 @@ export const computeVariant = (
...
+  userIds: (string | undefined)[],
+  callBack?: AvatarPersonaDataCallback
+): (AvatarPersonaData | undefined)[] => {
+  const [data, setData] = React.useState<(AvatarPersonaData | undefined)[]>([]);
+  useEffect(() => {
+    (async () => {
+      if (callBack) {
```
prprabhu-ms, line 204: Use early return to make this more readable.

```
if (!callback) {
  ... default return ...
}
const newData = ...
```

Also, the default return right now is `[]`. We should return a list the same length as `userIds` where all entries are `undefined`.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -188,26 +188,73 @@ export const computeVariant = (
...
+};
+
+/**
+ * @private
+ */
+const avatarDeepEqual = (
+  currentData: (AvatarPersonaData | undefined)[],
+  newData: (AvatarPersonaData | undefined)[]
+): boolean => {
+  let newDataPresent;
```
prprabhu-ms, line 248: You can avoid this captured variable update by using `Array.some` (see my comment earlier).

prprabhu-ms (commented):
Almost there :)

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -139,18 +139,18 @@ const useRemoteParticipantsWithCustomDisplayNames = (
...
-  const newParticipants = remoteParticipants;
   const userIds = useMemo(() => {
-    return newParticipants.map((p) => p.userId);
-  }, [newParticipants]);
+    return remoteParticipants.map((p) => p.userId);
+  }, [remoteParticipants]);
 
   const avatarPersonaData = useCustomAvatarPersonaData(userIds, onFetchAvatarPersonaData);
-  newParticipants.forEach((p, i) => {
+  const newParticipantData = remoteParticipants.map((p, i) => {
```
prprabhu-ms, line 147: nitty nit: Avoid suffixes like `data`, `info` etc that don't add anything to the meaning beyond what the rest of the name already says.

I'd just name this `newParticipants`.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -184,3 +186,70 @@ export const computeVariant = (
...
+ */
+export const useCustomAvatarPersonaData = (
+  userIds: (string | undefined)[],
+  callBack?: AvatarPersonaDataCallback
+): (AvatarPersonaData | undefined)[] => {
+  const [data, setData] = React.useState<(AvatarPersonaData | undefined)[]>([]);
+  useEffect(() => {
+    (async () => {
+      if (!callBack) {
+        const undefinedData: undefined[] = [];
```
prprabhu-ms, line 204: Another little (entirely optional) list comprehension trick: `const data = userIds.map(() => undefined);`

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -184,3 +186,70 @@ export const computeVariant = (
...
+        for (let i = 0; i < userIds.length; i++) {
+          undefinedData.push(undefined);
+        }
+        setData(undefinedData);
```
prprabhu-ms, line 208: This one not a nit.

This sets a new array on each render when `callback` is `undefined`. Instead, similar to the non-default case, we should return the same array each time so that the component that uses this hook doesn't render again.

My suggestion: `shouldUpdate` already handles when elements in the array are `undefined`. Just use that here as well before calling `setData`.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -184,3 +186,70 @@ export const computeVariant = (
...
+ * @private
+ */
+export const useCustomAvatarPersonaData = (
+  userIds: (string | undefined)[],
+  callBack?: AvatarPersonaDataCallback
+): (AvatarPersonaData | undefined)[] => {
+  const [data, setData] = React.useState<(AvatarPersonaData | undefined)[]>([]);
+  useEffect(() => {
+    (async () => {
+      if (!callBack) {
```
JamesBurnside, line 203: nit: should be able to avoid this if block,
simply intialize the data to an array of `[undefineds]`
```
const [data, setData] = React.useState<(AvatarPersonaData | undefined)[]>(Array(userIds.length).fill(undefined));
```

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -184,3 +186,70 @@ export const computeVariant = (
...
+        const undefinedData: undefined[] = [];
+        for (let i = 0; i < userIds.length; i++) {
+          undefinedData.push(undefined);
```
JamesBurnside, line 206: instead of a for loop, can instead do: `Array(<length>).fill(undefined)`

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
prprabhu-ms: Nada. That don't work. :D 

Consider the case when `callback` was defined on first render but was then set to `undefined` (Why would Contoso do such a thing? I don't know, but I don't want to have reason globally when reading this function. It _can_ happen).

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
prprabhu-ms: TIL.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
JamesBurnside: This would then be caught later on under 
```tsx
        userIds.map(async (userId: string | undefined) => {
          if (!userId) {
            return undefined;
          }
          return await callBack(userId);
        })
```

a callback would now be undefined so typescript would complain calling it and this would be
```tsx
userIds.map(async (userId: string | undefined) => {
          if (!userId || !callBack) {
            return undefined;
          }
          return await callBack(userId);
        })
```

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -184,3 +186,70 @@ export const computeVariant = (
...
+
+/**
+ * Hook to override the avatar persona data of the users
+ * - Calls {@link AvatarPersonaDataCallback} on each render for provided `userIds`
+ * - Returns an array of the same length as `userIds`. Entries in the array may be undefined if there is no data to return.
+ * @private
+ */
+export const useCustomAvatarPersonaData = (
+  userIds: (string | undefined)[],
+  callBack?: AvatarPersonaDataCallback
```
JamesBurnside, line 198: nitty nit: callback is one word so camel case doesn't apply here, just `callback`

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -184,3 +186,70 @@ export const computeVariant = (
...
+        }
+        setData(undefinedData);
+        return;
+      }
+      const newData = await Promise.all(
+        userIds.map(async (userId: string | undefined) => {
+          if (!userId) {
+            return undefined;
+          }
+          return await callBack(userId);
```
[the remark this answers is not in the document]
JamesBurnside: Ah it cannot, quick test react does not like an indeterminant number of useCallbacks:
```tsx
props.userIds.map((userId) => {
    useCallback(() => {
      callbackFn(userId);
    }, [userId]);
  }, []);
```
image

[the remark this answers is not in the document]
JamesBurnside: Is there any other memoizing util we can make use of here?

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
JamesBurnside: I snuck onto Donald's machine and made the change, let me know if I missed something haha

JamesBurnside (approved):

prprabhu-ms (approved):

packages/react-composites/src/composites/common/CustomDataModelUtils.tsx
```diff
@@ -61,10 +54,10 @@ const shouldUpdate = (
  * @private
  */
 const avatarDeepEqual = (currentData?: AvatarPersonaData, newData?: AvatarPersonaData): boolean => {
-  return currentData?.text !== newData?.text &&
-    currentData?.imageUrl !== newData?.imageUrl &&
-    currentData?.initialsColor !== newData?.initialsColor &&
-    currentData?.imageInitials !== newData?.imageInitials &&
+  return currentData?.text !== newData?.text ||
```
prprabhu-ms, line 57: Whoops!

Please add a couple simple unit-tests for this function, now that we've bugged it once already :P

packages/react-composites/src/composites/common/CustomDataModelUtils.tsx
```diff
@@ -0,0 +1,64 @@
...
+  if (currentData.length !== newData.length) {
+    return true;
+  }
+  return currentData.some((data, i) => avatarDeepEqual(data, newData[i]));
+};
+
+/**
+ * @private
+ */
+const avatarDeepEqual = (currentData?: AvatarPersonaData, newData?: AvatarPersonaData): boolean => {
```
JamesBurnside, line 56: update fn name to not equal:
```suggestion
const avatarDeepDifferenceCheck = (currentData?: AvatarPersonaData, newData?: AvatarPersonaData): boolean => {
```

(Or update fn to check if they are equal and then above have `!avatarDeepEqual(data, newData[i])`)

packages/react-composites/src/composites/common/CustomDataModelUtils.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return currentData?.text !== newData?.text ||
+    currentData?.imageUrl !== newData?.imageUrl ||
+    currentData?.initialsColor !== newData?.initialsColor ||
+    currentData?.imageInitials !== newData?.imageInitials ||
+    currentData?.initialsTextColor !== newData?.initialsTextColor
+    ? true
```
JamesBurnside, line 62: nit remove `? true : false` -- this will return true or false by default

packages/react-composites/src/composites/common/CustomDataModelUtils.tsx
dmceachernmsft: will do!

PR #1825, opened 2022-04-22
Add aria label to chat message so the content and author and time will be announced by voice over 

carocao-msft:
# What
Add aria label to chat message so the content and author and time will be announced by voice over 

# Why
Voice over on Mac on Safari is not annoucing chat message content
https://skype.visualstudio.com/SPOOL/_workitems/edit/2791702

# How Tested
Tested on Mac on safari and chrome, it would be the best if someone can help testing on iphone as well :)

### DIFF
--- a/change/@internal-react-components-9bd1280a-be6b-48f8-8728-0069bf2bbc66.json
+++ b/change/@internal-react-components-9bd1280a-be6b-48f8-8728-0069bf2bbc66.json
@@ -1,5 +1,5 @@
 {
-  "type": "none",
+  "type": "patch",
   "comment": "added aria label to chat message",
   "packageName": "@internal/react-components",
   "email": "carolinecao@microsoft.com",
--- a/change/@internal-react-components-9bd1280a-be6b-48f8-8728-0069bf2bbc66.json
+++ b/change/@internal-react-components-9bd1280a-be6b-48f8-8728-0069bf2bbc66.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "added aria label to chat message",
+  "packageName": "@internal/react-components",
+  "email": "carolinecao@microsoft.com",
+  "dependentChangeType": "none"
+}
--- a/packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
+++ b/packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
@@ -130,6 +130,11 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     <>
       <div ref={messageRef}>
         <Chat.Message
+          aria-label={`${message.content}${message.senderDisplayName}${
+            showDate
+              ? formatTimestampForChatMessage(message.createdOn, new Date(), strings)
+              : formatTimeForChatMessage(message.createdOn)
+          }`}
           data-ui-id="chat-composite-message"
           className={mergeStyles(messageContainerStyle as IStyle)}
           styles={messageContainerStyle}
--- a/packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
+++ b/packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
@@ -130,11 +130,15 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     <>
       <div ref={messageRef}>
         <Chat.Message
-          aria-label={`${message.content}${message.senderDisplayName}${
-            showDate
-              ? formatTimestampForChatMessage(message.createdOn, new Date(), strings)
-              : formatTimeForChatMessage(message.createdOn)
-          }`}
+          aria-label={_formatString(strings.messageContentDescription, {
+            displayName: `${message.mine ? strings.myself : message.senderDisplayName}`,
+            timeStamp: `${
+              showDate
+                ? formatTimestampForChatMessage(message.createdOn, new Date(), strings)
+                : formatTimeForChatMessage(message.createdOn)
+            }`,
+            content: `${message.content}`
+          })}
           data-ui-id="chat-composite-message"
           className={mergeStyles(messageContainerStyle as IStyle)}
           styles={messageContainerStyle}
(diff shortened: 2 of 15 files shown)


packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -130,6 +130,11 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     <>
       <div ref={messageRef}>
         <Chat.Message
+          aria-label={`${message.content}${message.senderDisplayName}${
```
prprabhu-ms, line 133: Should we have a localized string here that reasonably separates the different parts?

Something like "Message sent by ${message.senderDisplayName} at ${....timestamp}. ${message.senderDisplayName} says ${message.content}"

change/@internal-react-components-9bd1280a-be6b-48f8-8728-0069bf2bbc66.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
JamesBurnside, line 2: nit: this is a patch change

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
carocao-msft: When tab focus on the message, the content and display name and time stamp is announced

prprabhu-ms: cool, that sounds reasonable. Let's see if we can improve what's announced using the templated localized string as I suggested above.

prprabhu-ms: Also, what about messages sent by the local participant. Should we say "message sent by you" instead of announcing the local participant's name?

carocao-msft: Updated PR, what I put down is message sent by myself

carocao-msft: when the message is sent by myself, it gets an empty display name

change/@internal-react-components-9bd1280a-be6b-48f8-8728-0069bf2bbc66.json
carocao-msft: updated :)

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -106,7 +106,9 @@
     "editBoxCancelButton": "Cancel",
     "editBoxSubmitButton": "Submit",
     "messageReadCount": "Read by {messageReadByCount} of {remoteParticipantsCount}",
-    "actionMenuMoreOptions": "More Options"
+    "actionMenuMoreOptions": "More Options",
+    "messageContentDescription": "Message sent by {displayName} at {timeStamp}. {displayName} says {content}",
```
prprabhu-ms, line 110: We use the suffix `ariaLabel` for such strings in other places. Perhaps call this `messageContentAriaLabel` for consistency?

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -106,7 +106,9 @@
...
+    "myself": "myself"
```
prprabhu-ms, line 111: Could this key be `localSenderAriaLabel`. I'm conflicted whether it's better to announce "message sent by myself" or "message sent by you". In the UI, we show "(you)" in the participant list for the local user.
image

so I'm leaning towards "message sent by you" for the announcer as well.

In both cases (myself vs you), the key `localSenderArialLabel` sounds sensible, which validates that key name.

dmceachernmsft: I also learn towards 'message set by you' since I don't think we want the narrator talking in the first person for the user.

carocao-msft: Updated :)

edwardlee-msft (approved):

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -130,6 +130,15 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     <>
       <div ref={messageRef}>
         <Chat.Message
+          aria-label={_formatString(strings.messageContentAriaLabel, {
```
JamesBurnside, line 133: can you build this string outside of this large body of nested html tags and memoize with useMemo also?

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -130,6 +130,15 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
...
+            displayName: `${message.mine ? strings.localSenderAriaLabel : message.senderDisplayName}`,
+            timeStamp: `${
+              showDate
+                ? formatTimestampForChatMessage(message.createdOn, new Date(), strings)
+                : formatTimeForChatMessage(message.createdOn)
+            }`,
+            content: `${message.content}`
```
JamesBurnside, line 140: I'm worried content is not always a nice string, what if it contains a gif? or other html elements:
image

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
carocao-msft: updated

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
carocao-msft: added tostring() method

prprabhu-ms: Great catch. I hadn't thought of that at all. Perhaps we should only announce the content when type is `'text'` and instead announce a generic message for the other types: "A non-text message was sent by <display_name>".

Later on, perhaps we can start parsing html to get alt text and get fancier :)

prprabhu-ms:
Unassigning myself because we decided to not pursue this approach for now.

Please reassign if this PR needs attention.

JamesBurnside (changes requested):
Blocking this due to discussions as unfortunately we cannot pursue this method as we must ensure the screen reader reads the chat contents naturally (and not forced into an aria label) as we cannot simply stringify rich/html content.
We should engage with the a11y team to find a path forward here but this feels like a bug in the voice over tool so best to close this PR for now without merging.

PR #1829, opened 2022-04-25
Dmceachernmsft/SplitButtonFix

dmceachernmsft:
# What

remove menu role from the control bar button. add ariasplitbuttonlabels to the camera and mic buttons.
# Why

On mac it fixes a issue where the role is announced twice. On windows it fixes the major issue of not being able to interact with the buttons at all.

# How Tested

Ran locally on windows to verify interaction issue. Forwarded port to @edwardlee-msft to verify mac fix.


packages/communication-react/stable-review/communication-react.api.md
```diff
@@ -814,7 +814,6 @@ export type CameraButtonSelector = (state: CallClientState, props: CallingBaseSe
 
 // @public
 export interface CameraButtonStrings {
-    cameraButtonSplitRoleDescription?: string;
```
JamesBurnside, line 817: JAMES BOT: THIS IMPACTS THE STABLE API FILE, HAVE YOU CONFIRMED THIS IS NOT PART OF ANY STABLE RELEASE?

dmceachernmsft: Yes I believe it did go out when we released CallWithChatComposite. How would we go about removing this string?

JamesBurnside: First check on stable if this issue repros (git checkout the 1.2.0 tag and run locally), if not do we have to remove it?

packages/react-components/src/components/MicrophoneButton.tsx
```diff
@@ -171,9 +176,12 @@ export const MicrophoneButton = (props: MicrophoneButtonProps): JSX.Element => {
       menuProps={props.menuProps ?? generateDefaultDeviceMenuPropsTrampoline(props, strings)}
       menuIconProps={props.menuIconProps ?? !enableDeviceSelectionMenuTrampoline(props) ? { hidden: true } : undefined}
       split={props.split ?? enableDeviceSelectionMenuTrampoline(props)}
-      ariaDescription={
+      aria-roledescription={
```
alkwa-msft, line 179: should this be ariaRoleDescription? Have we used hyphens in other react props?

dmceachernmsft: Not all aria tags have a camel case reflection in react, in this case we have to use the default one.

packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -133,19 +133,19 @@ export const ControlBarButton = (props: ControlBarButtonProps): JSX.Element => {
       ? props?.strings?.tooltipOnContent
       : props?.strings?.tooltipOffContent);
 
+  const splitButtonAriaLabel = props.splitButtonAriaLabel ? tooltipContent + props.splitButtonAriaLabel : undefined;
```
JamesBurnside, line 136: This shouldn't combine the tooltipContent and the aria label, instead just have the aria label be `Choose camera and camera options`.

JamesBurnside: e.g. while the video is loading and the button is disabled the tooltip is `video is loading`

packages/react-components/src/components/ControlBarButton.tsx
```diff
@@ -133,19 +133,19 @@ export const ControlBarButton = (props: ControlBarButtonProps): JSX.Element => {
...
       <DefaultButton
         {...props}
-        role={props.split ? 'menu' : undefined}
         styles={componentStyles}
         onRenderText={props.showLabel && props.onRenderText ? props.onRenderText : undefined}
         onRenderIcon={props.onRenderIcon ?? DefaultRenderIcon}
-        ariaLabel={props.ariaLabel ?? tooltipContent ?? labelText}
+        ariaLabel={!splitButtonAriaLabel ? props.ariaLabel ?? tooltipContent ?? labelText : splitButtonAriaLabel}
         allowDisabledFocus={props.allowDisabledFocus ?? true}
-        aria-roledescription={props.ariaDescription}
```
JamesBurnside, line 148: Revert this? Or is this intentionally left deleted?

JamesBurnside (changes requested):

packages/communication-react/review/communication-react.api.md
```diff
@@ -851,6 +851,8 @@ export type CameraButtonSelector = (state: CallClientState, props: CallingBaseSe
 export interface CameraButtonStrings {
     cameraActionTurnedOffAnnouncement: string;
     cameraActionTurnedOnAnnouncement: string;
+    cameraButtonDisabledSplitAriaLabel?: string;
```
JamesBurnside, line 854: Reading it now this will need to use `on`/`off` nomenclature to match the other props (and disabled for the camera button means unclickable instead of camera being "off".
In this interface it doesn't need to be prefixed with cameraButton either as these are the camera button strings:
```suggestion
    offSplitAriaLabel
    onSplitAriaLabel
```

packages/react-components/src/components/ControlBarButton.tsx
JamesBurnside: Actually I agree to removing this. With this when `aria-roledescription` is passed into ControlBarButton it is applied twice; once as role-description and once as ariaDescription. So sounds OK to remove it.. but what about other buttons that are using ariaDescription?

JamesBurnside (changes requested):
renames and `aria-roledescription` question

packages/react-components/src/components/ControlBarButton.tsx
dmceachernmsft: this was intentionally deleted since we pass it in from the camera and microphone button. It didn't make sense to have this work this way since the control bar button and the camera and mic buttons both have `aria-roledescription` and `ariaDescription` props and we cross where that information is coming from here.

JamesBurnside: Ah no other buttons use this yet! Sounds good 👍

JamesBurnside: We do have some other places we may need to update if aria-roledesciption is preferred over ariaDescription
(but _not_ in this PR!)
image

edwardlee-msft (approved):

PR #1830, opened 2022-04-26
Use rushx to enable error report for rush build

PorterNan:
# What

Previously `rushx` doesn't bubble up the return code of the child process, but now, with -use-shell parameter (which we added long ago), seems it is behaving correctly, change all npm run to rushx so that `rush build` can receive the right error message.

Side Effect: you might see multiple `build error code return x` when build fails, that's because rushx is bubbling up all the error code to the top

image

# Why

# How Tested

# Process & policy checklist


JamesBurnside:
Does this mean we can remove the `-v` everywhere??

JamesBurnside (approved):

prprabhu-ms (approved):

PR #1841, opened 2022-04-29
Fix local preview in the VideoGallery not re-rendering when localVideoViewOptions is changed - Part 1

JamesBurnside:
# What
* Fix dependencies of the useMemo
* Memoize the localCameraSwitcherButton in isolation for those deps

**Side Note**: we should fully ban disabling `react-hooks/exhaustive-deps` in our codebase

# Why
If someone changed the `localVideoViewOptions` (or a couple other props missed in the dep array) then the change would not be reflected to the end user as no re-render was triggered due to incorrect useMemo dep array.

**Update**: Found another issue where we do not recreate the local stream when the prop changes, will be a subsequent PR (this change is still required however)

# How Tested
Not extensively tested, please review thoroughly, just tested when running locally that video preview correctly showed and could turn it off and on correctly.


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -248,11 +248,12 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     maxDominantSpeakers: MAX_AUDIO_DOMINANT_SPEAKERS
   });
 
-  /* @conditional-compile-remove(call-with-chat-composite) @conditional-compile-remove(local-camera-switcher) */
-  const localCameraCycleButton = (localVideoCameraCycleButtonProps): JSX.Element => {
-    const ariaDescription = _formatString(strings.localVideoSelectedDescription, {
-      cameraName: localVideoCameraCycleButtonProps.selectedCamera.name
-    });
+  const LocalCameraCycleButton = useCallback((): JSX.Element => {
```
prprabhu-ms, line 251: I would discourage using `CapitalCase` name for a local variable, even if it behaves like a `React.Component`.

Two options:
* Extract into an actual component below in this file and use that.
* Name `camelCase` like any other variable.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -248,11 +248,12 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
+    const ariaDescription =
+      localVideoCameraCycleButtonProps?.selectedCamera &&
+      _formatString(strings.localVideoSelectedDescription, {
+        cameraName: localVideoCameraCycleButtonProps.selectedCamera.name
+      });
     return (
       <Stack horizontalAlign="end">
         {showCameraSwitcherInLocalPreview &&
```
prprabhu-ms, line 259: Not for this PR: We should ideally not render the `Stack` when all these conditions are not met (instead of rendering an empty `Stack`).

prprabhu-ms (approved):

packages/react-components/src/components/VideoGallery.tsx
JamesBurnside: I had it extracted as a component but made the changelist more confusing to review -- hard to spot that just a useCallback is now added. I have a follow up PR that will be moving this (and the whole local video tile) to its own file so will be cleaned up shortly 👍

mgamis-msft (approved):

PR #1844, opened 2022-04-29
Storybook participant item now showing participant name and wrong font

carocao-msft:
# What
participant item storybook example was using on render avatar wrong, patched that error.
Made sure the (you) tag is the right font
make sure participant items are not clickable 
Screen Shot 2022-04-29 at 3 11 05 PM

# Why
[BUG 2667118](https://skype.visualstudio.com/SPOOL/_workitems/edit/2667118)


packages/storybook/stories/ParticipantItem/snippets/CustomIcon.snippet.tsx
```diff
@@ -14,13 +14,16 @@ export const CustomIconExample: () => JSX.Element = () => {
 
   const containerStyle = mergeStyles({ width: '15rem' });
 
+  const participantItemStyle: ParticipantItemStyles = { me: { fontFamily: 'Segoe UI' } };
```
JamesBurnside, line 17: This shouldn't be required -- do you know why we need to set this? This should be the default font.

JamesBurnside: Actually, can you check where the `(you)` text is in the ParticipantItem, from a quick glance it might need wrapped in a `<Text>` block (Text imported from fluentui)

packages/storybook/stories/ParticipantItem/snippets/CustomAvatar.snippet.tsx
```diff
@@ -1,19 +1,14 @@
...
   const onRenderAvatar = (): JSX.Element => {
     return (
-      <img
-        src="https://static2.sharepointonline.com/files/fabric/office-ui-fabric-react-assets/persona-female.png"
-        width="32px"
-        height="32px"
-        style={avatarStyle}
+      <Persona
+        size={PersonaSize.size32}
+        text="Annie Lindqvist"
```
JamesBurnside, line 10: The participant item takes in a `displayName` so I would assume the avatar should just be the image and _not_ be the image + participantName. (In which case we have to make a significant change the participantItem component to not render the fluent persona text but instead have our own `<Text>{displayName)</Text>` item.
I'm not sure what regressions that would cause however so makes it a more risky fix.

carocao-msft: In the code for participant item, display name is only used here 
Screen Shot 2022-04-29 at 4 30 05 PM
So in this case with onrenderavatar, it needs to pass in a <persona> not an image

packages/storybook/stories/ParticipantItem/snippets/CustomIcon.snippet.tsx
carocao-msft: This is where the (you) text is 
Screen Shot 2022-04-29 at 4 31 33 PM

carocao-msft: displayed names are generated using fluent ui <persona>, which explains why they are different font

carocao-msft: Tried wrapping it in a <text> block, the color changed to the same color as participant name. We still need to apply extra styling to change the text color.

carocao-msft: Tried wrapping it in a Text block, it works! Updated the PR :)

packages/storybook/stories/ParticipantItem/snippets/CustomAvatar.snippet.tsx
carocao-msft: Updated the code to use options.text in onrenderavatar instead of directly passing in display name. This matches with how we use onrenderavatar elsewhere in the code space. :)

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -180,7 +181,7 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
       ref={containerRef}
       role={'menuitem'}
       data-is-focusable={true}
-      className={mergeStyles(participantItemContainerStyle(me), styles?.root)}
+      className={mergeStyles(participantItemContainerStyle(me, menuItems ? true : false), styles?.root)}
```
prprabhu-ms, line 184: can be simply `!!menuItem`

Also, I generally advice against flag arguments: https://moderatemisbehaviour.github.io/clean-code-smells-and-heuristics/functions/f3-flag-arguments.html?msclkid=1506d94ac83c11ecb82a9b5b09dc259b

Consider creating two different functions, `participantItemContainerStyleNonClickable` and `participantItemContainerStyleClickable`.

OR

[my preferred], keep the flag argument, but make the arguments easier to understand at the callsite by using an options bag: 
```
participantItemContainerStyle({localParticipant: me, clickable: !!menuItems))

const participantItemContainerStyle = (options: {localparticipant: boolean, clickable: boolean}) => {
...
}
```

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -194,7 +195,11 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
         className={mergeStyles({ width: `calc(100% - ${menuButtonContainerStyle.width})`, alignItems: 'center' })}
       >
         {avatar}
-        {me && <Stack className={meTextStyle}>{isMeText}</Stack>}
+        {me && (
+          <Stack>
```
prprabhu-ms, line 199: Do you need this `Stack`, now that this is wrapped in a `Text`?

prprabhu-ms (approved):

packages/react-components/src/components/ParticipantItem.tsx
carocao-msft: Updated

packages/react-components/src/components/ParticipantItem.tsx
carocao-msft: udpated

JamesBurnside (approved):

PR #1845, opened 2022-04-29
Chat client error handling

dmceachernmsft:
# What

Introduce the use of `asyncTeeErrorToEventEmitter` to the `fetchInitialData` function 
# Why

so that the errors coming from the `fetchInitialData` are properly bubbled. These errors should be thrown for Contoso to catch. Our samples should possibly know how to catch these samples and behave appropriately.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2457314
# How Tested

Ran locally by crashing storybook to get these errors to show.


packages/chat-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -72,6 +72,7 @@ export const createDefaultChatHandlers = memoizeOne(
         if (messageIterator === undefined) {
           // Lazy definition so that errors in the method call are reported correctly.
           // Also allows recovery via retries in case of transient errors.
+
```
alkwa-msft, line 75: do we need to have this newline?

packages/chat-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -81,28 +82,36 @@ export const createDefaultChatHandlers = memoizeOne(
...
-        let readReceipt = await readReceiptIterator.next();
-        while (!readReceipt.done && parseInt(readReceipt?.value?.chatMessageId) >= earliestTime) {
-          readReceipt = await readReceiptIterator.next();
+        try {
+          let readReceipt = await readReceiptIterator.next();
+          while (!readReceipt.done && parseInt(readReceipt?.value?.chatMessageId) >= earliestTime) {
+            readReceipt = await readReceiptIterator.next();
+          }
+        } catch (e) {
+          console.log(e);
```
alkwa-msft, line 114: hesitant around console.log in library code. 

Can we throw some kind of exception instead?

dmceachernmsft: this is how we catch the other errors coming out of the chat thread in the adapter we do this for `getProperties` and `listParticipants` in `AzureCommunicationsChatAdapter.ts` should we throw the errors here and then catch them in the adapter as well?

PorterNan: The handler layer is more like for those users who is building on top of component+selector level, swallowing the error and not throwing it sounds a little dangerous and it's actually a behavior change.
If this is just to stop the iteration, then the best way is just to stop the iteration/loop then throw the error following previous behavior

dmceachernmsft: Will make that change!

packages/chat-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -104,6 +104,7 @@ export const createDefaultChatHandlers = memoizeOne(
         while (!readReceipt.done && parseInt(readReceipt?.value?.chatMessageId) >= earliestTime) {
           readReceipt = await readReceiptIterator.next();
         }
+
```
PorterNan, line 107: Extra line here to remove

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -176,22 +176,18 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
   }
 
   async fetchInitialData(): Promise<void> {
-    try {
+    // If get properties fails we dont want to try to get the participants after.
+    await this.asyncTeeErrorToEventEmitter(async () => {
       await this.chatThreadClient.getProperties();
```
PorterNan, line 181: After thinking about this a bit, I believe we could merge 2 async () functions together, so it will be:

```
 await this.asyncTeeErrorToEventEmitter(async () => {
      await this.chatThreadClient.getProperties();
      for await...
});
```

PorterNan (approved):

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
dmceachernmsft: I was thinking about this as well since if the first function fails the error will surface. something I am wondering though (will test and report back) is does the emitter support calling 2 functions inside it?

prprabhu-ms: Emitter doesn't care about what you do in the closure. It just pipes the thrown exception, if any.

prprabhu-ms (approved):

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
dmceachernmsft: Thanks for the clarification will update

PR #1846, opened 2022-05-02
Remove stabilized feature flag `control-bar-split-buttons`

prprabhu-ms:
# What

The feature was stabilized in the `1.2.0` release.

This PR removes the feature flags and cleans up any logic that existed specifically to support this feature flag.

### DIFF
--- a/change/@internal-calling-component-bindings-b4315fda-dc84-4d8f-9b26-ce98d82f29f3.json
+++ b/change/@internal-calling-component-bindings-b4315fda-dc84-4d8f-9b26-ce98d82f29f3.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Remove stabilized feature flag control-bar-split-buttons",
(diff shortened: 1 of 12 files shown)


PorterNan (approved):
This is a good example of how to entirely remove a feature flag - especially for xxxTrampoline part, we should mention one of these examples somewhere in the document then it makes everything eaiser for people who wants to remove a feature flag

dmceachernmsft (approved):

prprabhu-ms:
> This is a good example of how to entirely remove a feature flag - especially for xxxTrampoline part, we should mention one of these examples somewhere in the document then it makes everything eaiser for people who wants to remove a feature flag

Good idea.

This is also an example why it's a good idea to do this clean up after the release, when there is no deadline. It is easy to introduce bugs: https://github.com/Azure/communication-ui-library/pull/1846/commits/260bfef7a62b79fc19201fc42b3623099af500e6

Good thing we have those e2e tests, eh?

PR #1849, opened 2022-05-03
[DRAFT] WIP: Perf counter for re-rendering 

PorterNan:
# What

This is the prototype of perf counter:
1. it records how many times the binding level hook "useSelector" being called for each specific selector- which means components which are calling the selector are rerendered.
2. It records how many times real state changed are happening for the selector, which means a re-render triggered by state itself
# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/selectors/localPreviewSelector.ts
```diff
@@ -1,17 +1,20 @@
...
 
+import { DeviceManagerState } from '@internal/calling-stateful-client';
 import * as reselect from 'reselect';
 import { getDeviceManager } from './baseSelectors';
 
-/**
- * @private
- */
-export const localPreviewSelector = reselect.createSelector([getDeviceManager], (deviceManager) => {
+const selectLocalPreview = (deviceManager: DeviceManagerState) => {
```
PorterNan, line 8: We need to make it non-anonymous function so we can get function.name from this when we count for perf

packages/react-composites/src/composites/common/Logger.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { createClientLogger } from '@azure/logger';
+
+/**
+ * @private
+ */
+export const perfLogger = createClientLogger('communication-react:composite-perf-counter');
```
PorterNan, line 8: We use a specific perfLogger to log perf event

packages/react-composites/tests/browser/call/CallComposite.test.ts-snapshots/composite pages load completely
```diff
@@ -0,0 +1 @@
+{"getPage-parent rerender":1,"-parent rerender":126,"selectLocalPreview-parent rerender":29,"getLocalMicrophoneEnabled-parent rerender":29,"-return state changed":24,"selectLocalPreview-return state changed":4}
```
PorterNan, line 1: This is the re-render snapshot for each selector-event - some selectors are anonymous so we can't track the name, so we need to make changes like in localPreviewSelector to give them a name

packages/react-composites/tests/browser/call/CallComposite.test.ts-snapshots/local device buttons should show tooltips on hover
```diff
@@ -0,0 +1 @@
+{"getPage-parent rerender":3,"-parent rerender":378,"selectLocalPreview-parent rerender":87,"getLocalMicrophoneEnabled-parent rerender":87,"-return state changed":72,"selectLocalPreview-return state changed":12}
```
PorterNan, line 1: Tested this for several times - it looks pretty stable while I ran 10+ test updates

PR #1851, opened 2022-05-03
Add util method for hiding and enabling tooltips during tests

anjulgarg:
# What
Adding util method for hiding and enabling tooltips.
This helps us fix the following bug
**https://skype.visualstudio.com/SPOOL/_sprints/taskboard/ACS%20Web%20UI/SPOOL/CY2022-Q2/Sprint%2062%20(Ends%20May%2015)?workitem=2826640**

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-composites/tests/browser/call/CallComposite.test.ts
+++ b/packages/react-composites/tests/browser/call/CallComposite.test.ts
@@ -10,7 +10,9 @@ import {
   isTestProfileStableFlavor,
   waitForCallCompositeToLoad,
   waitForFunction,
-  waitForSelector
+  waitForSelector,
+  disableTooltips,
+  enableTooltips
 } from '../common/utils';
 import { test } from './fixture';
 import { expect, Page } from '@playwright/test';
@@ -51,7 +53,9 @@ test.describe('Call Composite E2E Configuration Screen Tests', () => {
   test('composite pages load completely', async ({ pages }) => {
     const page = pages[0];
     await stubLocalCameraName(page);
+    await disableTooltips(page);
     expect(await page.screenshot()).toMatchSnapshot(`call-configuration-page.png`);
+    await enableTooltips(page);
   });
 
   test('local device settings can toggle camera & audio', async ({ pages }) => {
@@ -63,7 +67,9 @@ test.describe('Call Composite E2E Configuration Screen Tests', () => {
       return !!videoNode && videoNode.readyState === 4 && !videoNode.paused && videoNode;
     });
     await stubLocalCameraName(page);
+    await disableTooltips(page);
     expect(await page.screenshot()).toMatchSnapshot(`call-configuration-page-camera-enabled.png`);
+    await enableTooltips(page);
   });
 
   test('local device buttons should show tooltips on hover', async ({ pages }) => {
--- a/packages/react-components/src/components/FileDownloadCards.tsx
+++ b/packages/react-components/src/components/FileDownloadCards.tsx
@@ -4,6 +4,7 @@ import { Icon, Spinner, SpinnerSize } from '@fluentui/react';
 import React, { useCallback, useState } from 'react';
 import { _FileCard } from './FileCard';
 import { _FileCardGroup } from './FileCardGroup';
+import { extension } from './utils';
 
 /**
  * Meta Data containing information about the uploaded file.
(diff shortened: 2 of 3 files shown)


packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -51,7 +53,9 @@ test.describe('Call Composite E2E Configuration Screen Tests', () => {
   test('composite pages load completely', async ({ pages }) => {
     const page = pages[0];
     await stubLocalCameraName(page);
+    await disableTooltips(page);
```
JamesBurnside, line 56: Just a note: for other tooltip issues we've used `clickOutsideOfPage` in common/utils.ts
Both solutions should work and I'm happy to keep this the way it is, just highlighting for awareness that that fn exists

JamesBurnside: (thinking some more: an alternative solution would be to add `clickOutsideOfPage` to waitForCallCompositeToLoad so new tests for the configuration page shouldn't hit this be default)

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -51,7 +53,9 @@ test.describe('Call Composite E2E Configuration Screen Tests', () => {
   test('composite pages load completely', async ({ pages }) => {
```
JamesBurnside, line 53: CallWithChatComposite also has a pre-join screen experience that should probably get this tooltip mitigation also

JamesBurnside (approved):

packages/react-composites/tests/browser/call/CallComposite.test.ts
anjulgarg: > Just a note: for other tooltip issues we've used `clickOutsideOfPage` in common/utils.ts Both solutions should work and I'm happy to keep this the way it is, just highlighting for awareness that that fn exists

Basis on your comment in the bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2826640
Looks like clickOutsideOfPage doesn't always work?

JamesBurnside: Lol! Looks like clickOutsideOfPage is not being used in these tests
But yeah I think what I was getting at is it's tough when someone adds something new we suddenly might have flakiness from tooltips which the OG developer shouldn't have to think about... So maybe we use your functions and disable tooltips by default and only enable them when testing? But we can cross that bridge another time

packages/react-composites/tests/browser/call/CallComposite.test.ts
```diff
@@ -63,7 +67,9 @@ test.describe('Call Composite E2E Configuration Screen Tests', () => {
       return !!videoNode && videoNode.readyState === 4 && !videoNode.paused && videoNode;
     });
     await stubLocalCameraName(page);
+    await disableTooltips(page);
```
prprabhu-ms, line 70: Should we add a screenshot helper instead? `page.screenshotWithStubbedElements()` that internally disables tooltips and stubs out local camera name while taking the screenshot?

packages/react-composites/tests/browser/call/CallComposite.test.ts
prprabhu-ms: It's confusing to have two ways to do things.

Let's pick one of the two ways to disable tooltips and follow that uniformly.

packages/react-composites/tests/browser/call/CallComposite.test.ts
anjulgarg: Good idea.
However, might end up being used for scenarios where we need tooltips to be shown as well.
Someone new in the team might be unaware of what's happening behind the scenes and could end up wasting a lot of time debugging...

prprabhu-ms: Either way is fine, as long as we modify this PR to keep one way of avoiding tooltips instead of two.

prprabhu-ms:
> @JamesBurnside and @prprabhu-ms If we already have a `clickOutside...` function and if it works, I am gonna abandon this PR. No point introducing a new way of doing things...
> 
> What say you?

Sounds good. You'd still need to call that function from the few tests we forgot, right?
Do that in a separate PR? (Just to keep conversation straight)? I'm OK with cannibalizing this PR for that as well.

PR #1856, opened 2022-05-04
Add a more reliable method to avoid flaky tooltips in e2e snapshots

prprabhu-ms:
The UFD tests were already dismissing tooltips by clicking off screen, but we still saw a small rate of flakiness in the screenshots.

This PR introduces an alternative method, suggested by @anjulgarg, in #1851 that is likely to work better - it disables the tooltips via CSS. We know that the new styles have been loaded by the browser when the function returns.

The plan is to merge this PR and observe for a week or so, if the UFD tests are found to be stable, we can start replacing other uses of the now deprecated `clickOutsideOfPage` function with the new `stableSnapshot` helper.

### DIFF
--- a/packages/react-composites/tests/browser/common/utils.ts
+++ b/packages/react-composites/tests/browser/common/utils.ts
@@ -333,7 +333,7 @@ export async function stableScreenshot(
     await disableTooltips(page);
   }
   try {
-    return page.screenshot(screenshotOptions);
+    return await page.screenshot(screenshotOptions);
   } finally {
     if (stubOptions?.tooltips != false) {
       await enableTooltips(page);
--- a/packages/react-composites/tests/browser/common/utils.ts
+++ b/packages/react-composites/tests/browser/common/utils.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT license.
 
 import { IDS } from './constants';
-import { ElementHandle, JSHandle, Page, TestInfo } from '@playwright/test';
+import { ElementHandle, JSHandle, Page, PageScreenshotOptions, TestInfo } from '@playwright/test';
 import { ChatUserType, CallUserType, CallWithChatUserType } from './fixtureTypes';
 import { v1 as generateGUID } from 'uuid';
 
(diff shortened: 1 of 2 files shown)


packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -302,3 +303,53 @@ export const isTestProfileStableFlavor = (): boolean => {
...
+  // Disable tooltips on all buttons in the call control bar.
+  dismissTooltips?: boolean;
+}
+
+/**
+ * A helper to take stable screenshots.
+ *
+ * USE THIS INSTEAD OF page.screenshot()
+ */
+export async function stableScreenshot(
```
anjulgarg, line 319: The whole comment and naming of this function is misleading.
At a glance, it looks like this should be the defacto method for taking screenshots. However, what if my test requires the tooltip to be visible?

prprabhu-ms: You still use this method.

Tooltips are only dismissed if you call it as `stableScreenshot(page, {disableTooltips: true})`.
Tooltips are not dismissed if you call it as `stableScreenshot(page, {})`.

The only thing it definitely does every time is wait for fonts to load (we might wait for more such universal things in the future).

anjulgarg: didn't notice that optional param. (Y)

anjulgarg (approved):

edwardlee-msft (approved):

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -34,9 +34,7 @@ async function screenshotOnFailure<T>(page: Page, fn: () => Promise<T>): Promise
 export const pageClick = async (page: Page, selector: string): Promise<void> => {
   await page.bringToFront();
   await screenshotOnFailure(page, async () => await page.click(selector, { timeout: PER_STEP_TIMEOUT_MS }));
-
-  // Move the mouse off the screen
-  await page.mouse.move(-1, -1);
+  clickOutsideOfPage(page);
```
JamesBurnside, line 37: this won't work for interacting with context menus, (I'm assumming) they will be clicked then instantly dismissed by the clickOutsideOfPage
I don't remember the context of `await page.mouse.move(-1, -1);` (and I'm too lazy to look at the hisotry) but my assumption is that it was added to remove flaky hover styling/behavior -- perhaps that could be worked into your stableScreenshot fn

prprabhu-ms: Oh oops. This change is also entirely unrelated to the purpose of this PR.

... and this is the reason e2e tests have been failing, I think.

My saviour!

PR #1860, opened 2022-05-05
Enable size change compare in CI

PorterNan:
# What

This is to enable bundle size compare in CI
# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -537,7 +537,7 @@ jobs:
         app: [Chat, Calling, CallWithChat]
     steps:
       - id: matrix-lowercase
-        run: INPUT=${{ matrix.app }} echo "::set-output name=value::${INPUT,,}"
+        run: INPUT=${{ matrix.app }} echo "::set-output name=app::${INPUT,,}"
       - uses: actions/checkout@v2
         with:
           ref: ${{ github.base_ref }}
@@ -551,7 +551,7 @@ jobs:
       - name: Install dependencies
         run: rush install
       - name: Build
-        run: rush build -o ${{steps.matrix-lowercase.outputs.value}}
+        run: rush build -o ${{steps.matrix-lowercase.outputs.app}}
       - name: Upload ${{matrix.app}} stats.json
         uses: actions/upload-artifact@v1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -557,7 +557,7 @@ jobs:
       - name: Upload ${{matrix.app}} stats.json
         uses: actions/upload-artifact@v1
         with:
-          name: ${matrix.app}}-base-stats
+          name: ${{matrix.app}}-base-stats
           path: samples/${{matrix.app}}/dist/build/stats.json
 
   compare_bundle-sizes:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -537,7 +537,9 @@ jobs:
         app: [Chat, Calling, CallWithChat]
     steps:
       - id: lowercase
-        run: INPUT=${{ matrix.app }}; echo "::set-output name=app::${INPUT,,}"
+        run: |
+          INPUT=${{ matrix.app }}
+          echo "::set-output name=app::${INPUT,,}"
       - uses: actions/checkout@v2
         with:
           ref: ${{ github.base_ref }}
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -345,6 +345,12 @@ jobs:
         run: rush build -o calling
       - name: Tests
         run: rush test -o calling
+      - name: Upload current calling stats.json
+        uses: actions/upload-artifact@v1
+        if: ${{ matrix.flavor == 'beta' && github.event_name == 'pull_request' }}
```
prprabhu-ms, line 350: Why only on beta? Might as well upload for both flavors with different file names.

It'd actually be great to see that diff between stable/beta as our calling dependency might be different.

PorterNan: This is a little heavy cause it needs to build all previous code so we are adding 3 more apps to build if we build stable - but I agree it would be nice to see the size of stable, let's start with beta, and if everything works great, then I can open a new PR for stable (need to work on some matrix code)

alkwa-msft:
I want to be cautious of what we are putting into CI.

Are we getting an accurate measurement of the bundle size by doing this step with the sample apps?
I want to know are we making our bundles bigger because of the lib or because of what we are doing with the sample app itself?

What is the big thing we are trying to figure out here and what is the best way to check it? Perhaps this is the best but this PR leaves me with a lot of questions about "Why" rather than implementation.

.github/workflows/ci.yml
```diff
@@ -510,6 +528,80 @@ jobs:
               body: 'Failed to pass the component examples UI Test. If this PR is for UI change and the error is snapshot mismatch, please add "ui change" label to the PR for updating the snapshot.'
             })
 
+  build_upload_base_bundle_stats:
```
mgamis-msft, line 531: Just curious if we can store the base stats in our repo and compare to the current stats to those stats instead of having to build each sample on every CI run. Then when merging, we just replace the base stats with the current stats.  But we do need to ensure that the base stats are protected from modification. Just concerned since the obtaining the base bundle stats takes a total of 11 min per CI run.
image

PorterNan: That was what I started with, but those json file take 13mb for each to store, 3x13MB will be around 60MB, which doesn't make a lot of senses to store such a big file 

BTW, building the base will be always faster than other CI tasks since they are running in parallel - building base will only take time to build the base, but all the other CI needs do more steps, so this wouldn't be a bottleneck here

PorterNan:
> I want to be cautious of what we are putting into CI.
> 
> Are we getting an accurate measurement of the bundle size by doing this step with the sample apps? I want to know are we making our bundles bigger because of the lib or because of what we are doing with the sample app itself?
> 
> What is the big thing we are trying to figure out here and what is the best way to check it? Perhaps this is the best but this PR leaves me with a lot of questions about "Why" rather than implementation.

The idea here is to raise the attention to code owner and PR reviewers that "something has been changed here which is impacting the size", so it is code owner and PR reviewers will try to find out what is really changed when the alert sounds, we don't hard block people from checking in PR, it is just trying to provide more information here. So even it is app itself size changes

Another concern to do this for sample app not our npm package, our npm package only contains some src code without the real dependencies, we cannot measure how big the size is when our contoso pack everything together in their app, which only the find webpack app bundle.js will find out

.github/workflows/ci.yml
JamesBurnside, line 531: To avoid pollutng the PR page with a lot of comments this would be a great candidate for the new github job summaries: https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/

Then only if the diff is great than x% do we actually post on the PR to say - wow you added a lot was this intended??

prprabhu-ms: I would also like to discourage us from build base each time just to get stats diffs.

Let's remove the base builds from this PR before merging. We'll still get the per-PR size artifacts that we can compare against PRs. We can then figure out how to track the diff.

JamesBurnside, line 531: One hack I've seen is to store this kind of info as a gist. So everytime main CI runs it would update the gist containing our latest bundle sizes .. might be too hacky and also only works if the base ref is main but just throwing out an idea

PorterNan:
Close this because we open a new PR for the updated version

Issue #1862, opened 2022-05-05
Change of date format in every chat messages in chat composite


sangeethav123:
By using cross-platform - Chat composite js bundle file (Azure communication services the UI Library) for chat, we are looking for more friendlier date time strings per message like MM/DD/YY, h:mm A (Ex: 04/27/22, 9:40 AM) instead of YYYY/MM/DD h:mm A(2022/05/05 6.30 p.m)

alkwa-msft:
Thanks for logging this issue. Something we can start to look into. Do you have any suggestions on how you would envision this API to customize the date time format?
 
 An additional question.
 
 I am curious would you prefer this type of date format for every message or is it preferred to use terms like "Yesterday" to refer to 5/4/22 (in the case that today is 5/5/22 and the message was sent on the previous day?)

carocao-msft:
> 

Hi，
Thanks for the additional information！ We will be posting a draft PR with some usage documentation to see if it matches your expectations :)

carocao-msft:
Hi, we have created a draft PR for this issue, basically we added a new prop for MessageThread so you can pass in a function to format the timestamps in anyway you like. We have added an example about how to customize timestamp formats in this PR as well, located in storybook/message thread  https://github.com/Azure/communication-ui-library/pull/1863. Let me know if that resolves your issue :) Thank you!

sangeethav123:
Hi,

Thanks for the quick response and progress on #1862 issue!

I have gone through the #1863 PR.
Please correct me if I'm wrong, in that PR, time format is only customized but we need date format to be customized like MM/DD/YY and adding a comma between date and time (04/27/22, 9:40 AM). Is it possible to do that using sample code?

And, one more question.

Can I use that CustomizedFormatTimeStampForChatMessage like below sample code for onRenderMessage?
Because, we are using Chat composite js bundle file (Azure communication services the UI Library) for chat.
***@***.***
***@***.***

carocao-msft:
Hi,
For the first question, yes we can customize both date and time. I will update the PR to include a sample of that.
For the second question, sorry I can't see the sample code you attached.

sangeethav123:
> 

Hi, updated sample code in last comment. Please take a look on it. 
Thanks!

carocao-msft:
Hi, thanks for your snippet! We are going to take a closer look into it

sangeethav123:
Hi,

I have gone through your PR for date time formatting and this is what we expected.
[...]

PR #1863, opened 2022-05-05
Add a new prop in message thread to allow customization of date time format

carocao-msft:
# What
Add a new prop in message thread to allow customization of date time format
Add a new prop in localization provider to allow contoso to pass in date time formatting function on the composite level
Add examples in storybook describing how to customize date time format
Add test for message thread to test both composite and component level customization

# Why

https://github.com/Azure/communication-ui-library/issues/1862

Screen Shot 2022-05-16 at 8 36 34 PM

Screen Shot 2022-05-06 at 9 33 25 AM

Screen Shot 2022-05-06 at 9 33 32 AM


packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
```diff
@@ -52,6 +52,11 @@ type ChatMessageComponentProps = {
    * @param userId - user Id
    */
   onRenderAvatar?: OnRenderAvatarCallback;
+  /**
+   * Optional function to provide customized date format.
+   *
+   */
+  customizedFormatTimestampForChatMessage?: (messageDate: Date) => string;
```
alkwa-msft, line 59: should we just have 

dateTimeFormat as the name of the prop that is optional?
If its set we use it and if we don't have it or its invalid we use the default behavior?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -601,6 +601,12 @@ export type MessageThreadProps = {
    * this function will be called with the data inside `fileSharingMetadata` key.
    */
   fileDownloadHandler?: FileDownloadHandler;
+
+  /**
+   * Optional function to provide customized date format.
+   *
+   */
+  customizedFormatTimestampForChatMessage?: (messageDate: Date) => string;
```
alkwa-msft, line 609: can just call this prop "dateTimeFormat"

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -966,6 +973,7 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
             showMessageStatus={showMessageStatus}
             messageStatus={messageProps.message.status}
             onActionButtonClick={onActionButtonClickMemo}
+            customizedFormatTimestampForChatMessage={customizedFormatTimestampForChatMessage}
```
alkwa-msft, line 976: should it be an optionalProp called

dateTimeFormat? would we use it for anything other than chat messages?

carocao-msft: updated

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -966,6 +973,7 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
             showMessageStatus={showMessageStatus}
             messageStatus={messageProps.message.status}
             onActionButtonClick={onActionButtonClickMemo}
+            dateTimeFormat={dateTimeFormat}
```
prabhjot-msft, line 976: As `dateTimeFormat` is an optional prop, don't we need to add a check if prop is available or not.

prabhjot-msft (approved):
minor comment, otherwise looks good to me!

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
```diff
@@ -52,6 +52,11 @@ type ChatMessageComponentProps = {
    * @param userId - user Id
    */
   onRenderAvatar?: OnRenderAvatarCallback;
```
PorterNan, line 54: I am a little bit curious about what layer customers are asking for the dateTimeFormat? Is it component layer or composite layer?

packages/communication-react/stable-review/communication-react.api.md
```diff
@@ -1805,6 +1805,7 @@ export type MessageThreadProps = {
     onSendMessage?: (messageId: string) => Promise<void>;
     disableEditing?: boolean;
     strings?: Partial<MessageThreadStrings>;
+    dateTimeFormat?: (messageDate: Date) => string;
```
PorterNan, line 1808: I am thinking of a more generic way of this function - what if customer wants to further customized their timeStamp? What if they want to hover over time stamp and show some special popup? what if they want to print a img in timestamp?
There could be another option like onRenderTimeStamp?: (messageDate: Date) => React.Component to open all these possibilities (especially in component layer, then we might be able to set more restriction in composite level)

I would like to involve @anjulgarg to this discussion, cause he's the only expert on custom data injection in our team

packages/communication-react/review/communication-react.api.md
```diff
@@ -1942,6 +1942,7 @@ export type MessageThreadProps = {
     disableEditing?: boolean;
     strings?: Partial<MessageThreadStrings>;
     fileDownloadHandler?: FileDownloadHandler;
+    dateTimeFormat?: (messageDate: Date) => string;
```
anjulgarg, line 1945: This prop name is misleading to me.
When I think of `date time format` I think of the string template used for formatting a date.
For example, `YY-MM-DD` etc.
However, this callback is supposed to return the formatted datetime string which is confusing to me.

anjulgarg, line 1945: The callback is named `dateTimeFormat` but the input parameter is `messageDate`
A better name would be `messageDateTime` as it is inline with the callback name.

anjulgarg: OR changing the callback name to `dateFormat` is another option

anjulgarg, line 1945: How about we go 1 step further?
Since Javascript doesn't have native date formatting using `YYYY, MM, DD` etc.
We can use a popular library like momentjs, datejs behind the scenes and accept a date time format making the implementation standardized and inline with other programming languages

anjulgarg: OR use the native platform supported formatting features like
image

anjulgarg: Here's an article to help you out
https://dockyard.com/blog/2020/02/14/you-probably-don-t-need-moment-js-anymore

packages/react-components/src/components/MessageThread.tsx
alkwa-msft: this is really a callback that gives a dateTime and wants to return a string

I am thinking something like on__________ would be a better naming convention? 
onRenderDateTime doesn't seem quite right since it kind of implies returning a JSX.Element based on our other onRenders?

I think it might be good to have the same named prop in ComponentLocale and everywhere else in the library?
@PorterNan  thoughts?

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
alkwa-msft: It should really be in both. The components need to handle it and we want to expose it at the composite level.

@PorterNan  can we confirm this will be exposed through the JS bundle as well?

packages/communication-react/stable-review/communication-react.api.md
alkwa-msft: Lets just only focus on strings for now and we can look into a modification at the component level if devs want to extremely customize this area

alkwa-msft:
I don't think we want to to expose this API under we have reviewed this with ARB. Lets only expose this at the beta level for now if we want to.

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -145,11 +155,18 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
...
           timestamp={
             <Text data-ui-id={ids.messageTimestamp}>
-              {message.createdOn
-                ? showDate
-                  ? formatTimestampForChatMessage(message.createdOn, new Date(), strings)
-                  : formatTimeForChatMessage(message.createdOn)
-                : undefined}
+              {
+                // messageDateTime overwrites messageDateTimeLocale overwrites default
+                message.createdOn
```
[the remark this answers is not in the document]
carocao-msft: Updated the code and made the logic more clear, also added conditional compile. messageDateTimeLocale here comes from const { messageDateTimeLocale } = useLocale(); and the messageDateTime comes from prop passed to message thread component

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -38,3 +61,54 @@ describe('Message date should be formatted correctly', () => {
     expect(component.text()).toContain(testLocale.strings.messageThread.yesterday);
   });
 });
+
+describe('Message date should be customized by function passed through locale', () => {
+  test('Message date should be localized to "24 hours ago"', async () => {
```
prprabhu-ms, line 66: `/* @conditional-compile-remove fancy-feature-name */`

prprabhu-ms:
Removed a couple assignees to create clearer responsibility for moving the review along. Feel free to modify again. As a thumb rule, set 2-3 assignees. Any more and it dilutes responsibility for who needs to review proactively.

change/@internal-react-components-25d9c1f2-2a56-460a-9b8b-7a11ae28996f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Added new prop to chat message to allow custom date/time format",
```
anjulgarg, line 3: Adding prop name here would be helpful when changelog is created :)

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -76,9 +87,28 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     onRenderAvatar,
     showMessageStatus,
     messageStatus,
-    fileDownloadHandler
+    fileDownloadHandler,
+    /* @conditional-compile-remove(dateTimeCustomization) */
+    onDisplayDateTimeString
   } = props;
 
+  // onDisplayDateTimeString overwrites onDisplayDateTimeString overwrites default
```
anjulgarg, line 95: onDisplayDateTimeString overwrites default

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -76,9 +87,28 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
...
+  let formattedTimestamp: string | undefined = undefined;
+  if (message.createdOn) {
+    // default
+    formattedTimestamp = showDate
+      ? formatTimestampForChatMessage(message.createdOn, new Date(), strings)
+      : formatTimeForChatMessage(message.createdOn);
+
+    /* @conditional-compile-remove(dateTimeCustomization) */
+    if (onDisplayDateTimeString) {
```
anjulgarg, line 104: Should we use onDisplayDateTimeString if showDate is false?

carocao-msft: when ShowDate is false, timestamps are still displayed but displayed without dates, only time. So show date is a way to customize default date time format. onDisplayDateTimeString should overwrite showdate

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
carocao-msft: Updated

change/@internal-react-components-25d9c1f2-2a56-460a-9b8b-7a11ae28996f.json
carocao-msft: updated

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -76,9 +87,28 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
...
+      formattedTimestamp = onDisplayDateTimeString(message.createdOn);
+    } else if (onDisplayDateTimeStringLocale) {
+      /* @conditional-compile-remove(dateTimeCustomization) */
```
anjulgarg, line 107: Is this conditional compilation flag unnecessary here?
the one above the if block should remove the entire conditional block

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -76,9 +87,28 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     onRenderAvatar,
     showMessageStatus,
     messageStatus,
-    fileDownloadHandler
+    fileDownloadHandler,
+    /* @conditional-compile-remove(dateTimeCustomization) */
+    onDisplayDateTimeString
   } = props;
 
+  // onDisplayDateTimeString from messagethread overwrites onDisplayDateTimeString from locale overwrites default
```
anjulgarg, line 95: This logic is hard to follow. Can we simplify it by doing something like this?
image

anjulgarg (approved):

anjulgarg:
Although I have approved, I would recommend getting another review from @prprabhu-ms

common/config/babel/.babelrc.js
```diff
@@ -23,6 +23,8 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         'file-sharing',
         // Adhoc calls to a Teams user.
         'teams-adhoc-call',
+        // props to allow Contoso to overwrite timestamp format for chat messages, one in locale and one in message thread component
+        'dateTimeCustomization',
```
prprabhu-ms, line 27: Can we rename this to match the pattern we've used in other features: `date-time-customization`.

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -79,6 +102,16 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     fileDownloadHandler
   } = props;
 
+  const defaultTimeStamp = message.createdOn
+    ? generateDefaultTimestamp(message.createdOn, showDate, strings)
+    : undefined;
+
+  // onDisplayDateTimeString from props overwrite onDisplayDateTimeString from locale
```
prprabhu-ms, line 109: nit: Move comment to the implementatino of `generatCustomizedTimestamp`

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -79,6 +102,16 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     fileDownloadHandler
   } = props;
 
+  const defaultTimeStamp = message.createdOn
+    ? generateDefaultTimestamp(message.createdOn, showDate, strings)
+    : undefined;
+
```
prprabhu-ms, line 108: nitty, nitty, nit: I'd drop the newlines between these variables (they're all related)

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
```diff
@@ -59,6 +59,13 @@ type ChatMessageComponentProps = {
...
+
+  /* @conditional-compile-remove(dateTimeCustomization) */
```
prprabhu-ms, line 63: `ChatMessageComponentProps` is not exported from this module (and from the public API).

So there is no need for any conditional compilation here. Please add it unconditionally (this also simplifies your implementation).

Only public API needs to be added conditionally.

packages/react-components/src/components/utils/Datetime.ts
```diff
@@ -114,3 +114,18 @@ const dayToDayName = (day: number, dateStrings: MessageThreadStrings): string =>
       throw new Error(`Invalid day [${day}] passed`);
   }
 };
+
+/**
+ * @private
+ */
+export const generateDefaultTimestamp = (
```
prprabhu-ms, line 121: nit: I'd keep this helper function in `ChatMessageComponent` related files. This is not generally applicable to anything beyond a message (The `showDate` prop etc are very specific), so no value in moving this to a generic utils file. Instead, moving it here just increases the cross-references in the code-base.

General rule: Keep the modules as self contained as possible. Move generic functionality out to common modules / other modules when it is truly generic (has at least 2 (potential) use-cases in the codebase)

packages/storybook/stories/ChatComposite/snippets/CustomDateTimeFormat.snippet.tsx
```diff
@@ -0,0 +1,115 @@
...
+    {
+      endpoint: props.endpointUrl,
+      userId,
+      // Data model injection: The display name for the local user comes from Contoso's data model.
+      displayName: props.displayName,
+      credential,
+      threadId: props.threadId
+    },
+    // Custom behavior: Capitalize all messages before sending them.
+    capitalizeMessageBeforeSend
```
prprabhu-ms, line 42: Copy-pasta example. Please remove this custom behavior from the example.

packages/storybook/stories/ChatComposite/CustomDateTimeFormat.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+      } else {
+        setContainerProps(undefined);
+      }
+    };
+    fetchToken();
+  }, [args.userId, args.token, args.botId, args.botToken, args.endpointUrl, args.displayName]);
+
+  const strings = compositeLocale(locale)?.component.strings ?? COMPONENT_LOCALE_EN_US.strings;
+  const compositeStrings = compositeLocale(locale)?.strings ?? COMPOSITE_LOCALE_EN_US.strings;
+  const onDisplayDateTimeStringLocale =
```
prprabhu-ms, line 94: Why this fallback? We don't set the datetime format via the storybook locale at all. So let's directly assign the local function.

packages/storybook/stories/ChatComposite/CustomDateTimeFormat.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+  'Have fun!'
+];
+
+const CustomDateTimeFormatStory = (args, context): JSX.Element => {
+  const {
+    globals: { locale }
+  } = context;
+  const [containerProps, setContainerProps] = useState<ChatCompositeSetupProps>();
+
+  const onDisplayDateTimeString = (messageDate: Date): string => {
```
prprabhu-ms, line 33: Please move this function to file-level private function. We should avoid creating functions nested inside other functions because they reduce readability. You need to do this if you are capturing some local variable, even there it's better to create a helper function with most of the logic and call that from your closure defined here.

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -57,20 +57,33 @@ type ChatMessageComponentAsMessageBubbleProps = {
    */
   onRenderAvatar?: OnRenderAvatarCallback;
 
-  /* @conditional-compile-remove(dateTimeCustomization) */
+  /* @conditional-compile-remove(date-time-customization) */
```
prprabhu-ms, line 60: This can still be added unconditionally (not part of public API).

I'll make your code simpler by needing less conditional compilation.

prprabhu-ms (approved):

PR #1865, opened 2022-05-05
Fix beta version number bump 

dmceachernmsft:
# What

Fixed string logic for changing the beta numbers after a stable release
# Why
 fixed issue where `1.2.0` => `1.2.1-beta.1` so instead it becomes what we want where its `1.2.0` => `1.2.1-beta.0` after a stable release. As well `1.2.1-beta.0` => `1.2.2-beta.1` becomes `1.2.1-beta.0` => `1.2.1-beta.1` 

https://skype.visualstudio.com/SPOOL/_workitems/edit/2833878
# How Tested

Ran locally to see package.json file updates.


carocao-msft (approved):

common/scripts/bump-beta-version.js
```diff
@@ -13,11 +13,12 @@ const main = () => {
 
 const bumpBetaVersion = (currentVersion) => {
   const versionStrs = currentVersion.split('-beta.');
+  const wasStable = versionStrs[1] === undefined ? true : false;
   const currentBeta = versionStrs[1] ? Number.parseInt(versionStrs[1]) : 0;
   const [major, minor, patch] = versionStrs[0].split('.');
 
   // We will bump the patch version when create a beta on top of a stable one
-  return [[major, minor, currentBeta === 0? Number.parseInt(patch) + 1 : patch].join('.'), `${currentBeta + 1}`].join('-beta.');
+  return [[major, minor, currentBeta === 0 && wasStable? Number.parseInt(patch) + 1 : patch].join('.'), `${wasStable ? 0 : currentBeta + 1}`].join('-beta.');
```
prprabhu-ms, line 21: This condition should simply be `wasStable`.

Also, this one-liner is becoming hard to understand. Can you extract two variables `newPatch` and `newBeta` above so that the logic is easier to follow?

dmceachernmsft: will do!

prprabhu-ms (approved):

PR #1867, opened 2022-05-09
Fix disposing local stream views when inside a call

JamesBurnside:
# What
Correctly dispose of local video streams when they are attached to a callId

Previous PR in stack: #1866

# Why
Currently we only dispose of unparented local video streams, which are the preview streams created when not in a call yet.

# How Tested
If I change the LocalVideoViewOptions I now correctly see the video being destroyed and recreated. Before nothing would happen and it was impossible to update the LocalVideoViewOptions once the stream had been created.

Demo showing LocalVideoViewOptions scaling mode being slipped every 800ms:

https://user-images.githubusercontent.com/2684369/167450336-a1347136-40cf-4520-88a1-87d503e44a22.mp4

Note: Observe the stream is fully destroyed and recreated updating the scaling mode, this is not ideal and we need to support `updateScalingMode` at the statefulcallclient level (discussions offline regarding this)

Manual regression tests performed:
* Can turn off and on local video
* Can switch camera
* Can switch camera using mobile cycle-camera button
* Can join a call with camera off then turn on
* Can join a call with camera on then switch camera
* Can join a call with camera on then turn off camera
* All changes correctly observed by other participants


prprabhu-ms (approved):

JamesBurnside:
note: this is blocked on this PR stack: #1889 (and a subsequent PR that updates the PiPiP to use the LocalVideoTile)

prprabhu-ms:
Unassigning myself to remove from attention set. Please re-assign if you need another look.

PorterNan (approved):

PR #1871, opened 2022-05-09
Make filesharing responsive: prevent file cards from getting trimmed in case of smaller frame

prabhjot-msft:
# What
Prevent file cards from getting trimmed in case of smaller frame
Screen Shot 2022-05-09 at 1 47 01 PM

Expand chat thread on focus:
Screen Shot 2022-05-09 at 1 47 13 PM
# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2828331

# How Tested

# Process & policy 

checklist


anjulgarg:
@prabhjot-msft possible to make a video recording showcasing this feature?
I am having a hard time understanding it looking at the screenshots attached.

packages/react-composites/src/composites/common/styles/ParticipantContainer.styles.ts
```diff
@@ -98,7 +98,8 @@ export const participantListWrapper = mergeStyles({
   width: '20rem',
   // max width at 50% of view so the People Pane is not squeezing the Message Pane to almost nothing when on small screen or high zoom in
   maxWidth: '50vw',
-  height: '100%'
+  height: '100%',
+  boxShadow: '0px 1.2px 3.6px rgba(0, 0, 0, 0.1), 0px 6.4px 14.4px rgba(0, 0, 0, 0.13)'
```
anjulgarg, line 102: Can we use `theme.effects` here?

prabhjot-msft: Done

prabhjot-msft:
> @prabhjot-msft possible to make a video recording showcasing this feature?
> I am having a hard time understanding it looking at the screenshots attached.

https://user-images.githubusercontent.com/97124699/167514583-9280c827-e834-4b4f-a8cd-1fde3a6bf46d.mov

packages/react-composites/src/composites/common/styles/ParticipantContainer.styles.ts
```diff
@@ -94,12 +94,14 @@ export const peoplePaneContainerTokens: IStackTokens = {
...
-  width: '20rem',
-  // max width at 50% of view so the People Pane is not squeezing the Message Pane to almost nothing when on small screen or high zoom in
-  maxWidth: '50vw',
-  height: '100%'
-});
+export const participantListWrapper = (theme: Theme): string =>
+  mergeStyles({
+    width: '20rem',
+    // max width at 50% of view so the People Pane is not squeezing the Message Pane to almost nothing when on small screen or high zoom in
+    maxWidth: '50vw',
```
JamesBurnside, line 101: Unfortunately we can never use `vw` for styling in our composites, vw is the viewport width, however Contoso doesn't have to make our composite span the entire width of the viewport. If, for example, the composite was sized to smaller of the viewport vh/vw css won't apply correctly

packages/react-composites/src/composites/common/styles/ParticipantContainer.styles.ts
```diff
@@ -94,12 +94,14 @@ export const peoplePaneContainerTokens: IStackTokens = {
...
+    height: '100%',
+    boxShadow: theme.effects.elevation16
```
JamesBurnside, line 103: This looks like a significant change to make file sharing more responsive, on chat on desktop view does it now have a drop shadow?
The drop shadow appears to be applied the whole way around the participant container, if the composite doesn't span the whole screen can the shadow be seen out the top, bottom and right of the container?

change/@internal-react-composites-e27cd3a3-4cde-4372-b24e-aa332aee12b2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Make filesharing responsive: prevent file cards from getting trimmed in case of smaller frame",
```
JamesBurnside, line 3: Current implementation isn't actually preventing cards from getting trimmed in a smaller frame, it is simply circumventing the issue for our chat sample when the participant pane is open.
If you run the chat sample in the chrome emulator, with participant pane closed, on a samsung galaxy fold device, are the file cards still getting trimmed?

anjulgarg:
> > @prabhjot-msft possible to make a video recording showcasing this feature?
> > I am having a hard time understanding it looking at the screenshots attached.
> 
>  Screen.Recording.2022-05-09.at.4.21.17.PM.mov

This looks different from the mockup AlexP provided here
https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?node-id=3263%3A19683

Is this the correct implementation?

PR #1878, opened 2022-05-11
Add delay in overwriting `ComplianceBanner`

prprabhu-ms:
# What

* The state transitions are unchanged.
* If a message is displayed and new message needs to be displayed, it is delayed by 3 seconds.
  * If multiple messages arrive within that 3 second delay, the last one is shown (i.e., intervening messages during delay are skipped). This is consistent with the purpose of this banner -- the final state is the important one.
* If the user dismisses the banner and a new message arrives, it is also delayed by 3 seconds as above.

https://user-images.githubusercontent.com/82062616/167776990-34efa0d2-8441-4009-bde3-765a8641dc64.mp4

# Why

Potential mitigation for the problem that call recording and transcription events are often triggered in quick succession. This leads to a banner being shown and then quickly overwritten before it can be read.

# How Tested

Storybook.


dmceachernmsft (approved):

packages/react-components/src/components/ComplianceBanner/ComplianceBanner.tsx
```diff
@@ -147,6 +152,88 @@ function determineStates(previous: ComplianceState, current: boolean | undefined
...
+    variant,
+    lastUpdated: Date.now()
+  });
+  const pendingUpdateHandle = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  if (variant !== visible.variant && variantLastUpdated > visible.lastUpdated) {
+    // Always clear pending updates.
+    // We'll either update now, or schedule an update for later.
+    if (pendingUpdateHandle.current) {
+      clearTimeout(pendingUpdateHandle.current);
```
prprabhu-ms, line 192: Must set `pendingUpdateHandle.current` to `null` again.

`clearTimeout` is forgiving so this doesn't cause errors in practice, but still.

packages/react-components/src/components/ComplianceBanner/ComplianceBanner.tsx
```diff
@@ -71,15 +71,20 @@ export const _ComplianceBanner = (props: _ComplianceBannerProps): JSX.Element =>
...
       callRecordState: 'off'
-    }
+    },
+    lastUpdated: Date.now()
   });
 
   // Only update cached props and variant if there is _some_ change in the latest props.
   // This ensures that state machine is only updated if there is an actual change in the props.
-  if (
+  const shouldUpdatedCached =
```
edwardlee-msft, line 80: nit: Name is a bit confusing and difficult to interpret.
Possibly rename to **shouldCacheBeUpdated** or **doesCacheNeedUpdate** 

(Another possible one is **isCacheUpdated** but this one adds it's own confusion as well when reading it's use case)

edwardlee-msft (approved):

PR #1879, opened 2022-05-11
UI test for Chat composite localization now using fake adapter.

mgamis-msft:
# What

Switched localization UI test for Chat to use fake chat adapter.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2845218/?workitem=2845215

# How Tested

Tested locally and verify test duration:
https://skype.visualstudio.com/SPOOL/_workitems/edit/2845218
CI should also pass.

# Process & policy checklist


packages/react-composites/tests/browser/chat/fake-adapter/fixture.ts
```diff
@@ -36,15 +36,17 @@ const buildUrlForChatAppUsingFakeAdapter = (
   return url;
 };
 
+/**
+ * Base chat adapter model
+ */
+export const BASE_CHAT_ADAPTER_MODEL = {
```
prprabhu-ms, line 42: nit: This isn't really a model for the adapter. The fake adapter uses realistic state internally.

These are just `FAKE_CHAT_ADAPTER_ARGS` that tell us how to construct the fake.

prprabhu-ms (commented):
The PR LGTM, but I see some unrelated snapshot updates:
image

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -27,6 +32,9 @@ export const FakeAdapterApp = (): JSX.Element => {
     verifyParamExists(params.fakeChatAdapterModel, 'fakeChatAdapterModel')
   ) as ChatAdapterModel;
 
+  // Optional params
+  const useFrLocale = Boolean(params.useFrLocale);
```
dmceachernmsft, line 36: So this is just telling the fake adapter to use the french locale for the tests? Do we need to run the tests against the other locales?

mgamis-msft: Yes, for now. I just wanted to convert the old Localization UI test which used French for this PR.

mgamis-msft:
> The PR LGTM, but I see some unrelated snapshot updates: image

Reverted the real adapter UI test snapshots back after removing the 'ui-change' label now that the fake adapter snapshots have been generated.

prprabhu-ms (approved):

PR #1880, opened 2022-05-11
Disable ae-incompatible-tags warning in beta

anjulgarg:
# What
Disable ae-incompatible-tags warning in beta

# Why
In beta builds, ae-incompatible-tags warning creates friction since it is possible to have unmatching tags in beta build but we take care of those in stable builds by conditionally removing them.

# How Tested

# Process & policy checklist

### DIFF
--- a/common/config/api-extractor/api-extractor.json
+++ b/common/config/api-extractor/api-extractor.json
@@ -26,6 +26,10 @@
       "ae-forgotten-export": {
         "logLevel": "error",
         "addToApiReportFile": false
+      },
+      "ae-incompatible-release-tags": {
+        "logLevel": "none",
+        "addToApiReportFile": false
       }
     },
     "tsdocMessageReporting": {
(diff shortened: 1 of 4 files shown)


common/config/api-extractor/api-extractor.json
```diff
@@ -26,6 +26,10 @@
       "ae-forgotten-export": {
         "logLevel": "error",
         "addToApiReportFile": false
+      },
+      "ae-incompatible-release-tags": {
+        "logLevel": "none",
```
prprabhu-ms, line 31: Do we still want to warn on the command line? Or will that cause build failures as well (because we disallow warnings)?

prprabhu-ms (approved):
This warning is still useful in catching problems if we depend on @beta features from *external* packages in our @public API. But warnings that are "ok to ignore" are worse than not having the warnings, so LGTM.

common/config/api-extractor/api-extractor.json
anjulgarg: it's currently set to warning and causes build failure.

JamesBurnside (approved):

PR #1883, opened 2022-05-11
[Calling][CallWithChat][Bug] Change Microphone header to Audio Device when there are no Speakers detected

edwardlee-msft:
# What
Change Microphone header to Audio Device in the MoreDrawer and ConfigurationScreen -> Local Preview pane.
Icon was changed from Microphone icon to Speaker Icon in the MoreDrawer.
(Not sure how to configure the Icons in the DevicesButton Contextual Menu)

When speakers are not detected under the speakers category and when on mobile view, the "Microphone" header is changed to "Audio Device"

# Why
On Android, all audio devices are being categorized as a microphone. Speakers are not categorized as their own category.
There is an OS limitation on mobile devices that limit how we view microphones and speakers.
https://skype.visualstudio.com/SPOOL/_workitems/edit/2713322

# How Tested
Tested on Calling and CallWithChat samples on my android device

# Process & policy checklist


packages/react-composites/src/composites/CallWithChatComposite/Strings.tsx
```diff
@@ -37,6 +37,12 @@ export interface CallWithChatCompositeStrings {
    * {@link CallWithChatComposite} control bar Chat button enhanced tooltip string.
    */
   chatButtonTooltipClosedWithMessageCount: string;
+  /**
+   * Title for the audio device selection sub-menu in more buttton drawer.
```
JamesBurnside, line 41: nit: remove `in more buttton drawer` as this is currently used in the local preview context menu (which really _should_ be a more drawer but currently isn't)

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
```diff
@@ -116,23 +117,43 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
...
-            ? 'MoreDrawerSelectedMicrophone'
-            : 'MoreDrawerMicrophones'
-        },
-        text: mic.name,
-        onItemClick: onMicrophoneItemClick,
-        secondaryIconProps: isDeviceSelected(mic, props.selectedMicrophone) ? { iconName: 'Accept' } : undefined
-      })),
-      secondaryText: props.selectedMicrophone?.name
-    });
+    if (props.speakers && props.speakers.length > 0) {
```
JamesBurnside, line 120: nit: a lot of the props are the same for the drawer menu item, consider pulling out just the props that changed based on when speakers aren't present. Something like:
```
// Add comment also when and why we use the generic audioDevices over Microphone also
const speakersAvailable = props.speakers && props.speakers.length > 0;
const itemKey = speakersAvailable ? 'microphones' : 'audioDevices';
const text = speakersAvailable ? etc.

```

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/LocalPreview.tsx
```diff
@@ -120,6 +122,11 @@ export const LocalPreview = (props: LocalPreviewProps): JSX.Element => {
               disabled={!microphonePermissionGranted || !cameraPermissionGranted}
               showLabel={true}
               styles={devicesButtonStyles}
+              strings={
```
JamesBurnside, line 125: Consider updating the icon here too to be the speaker instead of the microphone like you did in the more drawer

prprabhu-ms:
Very elegant solution!

dmceachernmsft (approved):
image
I have seen this a couple times lately in PR's with snapshot changes, do we have any idea whats going on here?

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
edwardlee-msft: Made some props for those cases.

packages/react-composites/src/composites/CallComposite/components/LocalPreview.tsx
edwardlee-msft: Had to dive into the DevicesButton and make changes there. There is no prop that can passed down to change the icons unfortunately (maybe this is something we can add into the component for the future).

edwardlee-msft:
@JamesBurnside  and @dmceachernmsft 
Do you happen to know why the snapshot changes are missing the other users in my pr?
Seems like something is in disarray.

PR #1881, opened 2022-05-11
 File Sharing components are tab navigable

prabhjot-msft:
# What
Added Icon button for upload button, cancel file upload icon and download icon to make it tab navigable

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2806326
https://skype.visualstudio.com/SPOOL/_workitems/edit/2782206

# How Tested

# Process & policy checklist

- [ ] I have updated the project documentation to reflect my changes if necessary.
- [ ] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.

**Is this a breaking change?**

- [ ] This change causes current functionality to break.

PR #1886, opened 2022-05-12
Made filecard clickable and also disabled more button in mobile view when waiting in lobby

carocao-msft:
# What
Made filecard clickable in chat
Disabled more button in mobile view when waiting in lobby in callwithchat

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2850128
https://skype.visualstudio.com/SPOOL/_workitems/edit/2841478

# How Tested
Filecard: 

https://user-images.githubusercontent.com/96077406/168145129-a8a0679b-5a94-4870-9b43-8983e3a58853.mov

CallwithChat:
Join teams meeting using mobile, and make sure the more button is disabled when waiting to be admitted 

# Process & policy checklist


change/@internal-react-components-1d5161d0-2ec1-49de-8cf1-fccb26706dfb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "filecard should be clickable, also in mobile view when waiting in lobby, make more button disabled",
```
JamesBurnside, line 3: note: these should be two separate PRs

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -166,6 +166,7 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
                     data-ui-id="call-with-chat-composite-more-button"
                     strings={moreButtonStrings}
                     onClick={props.onMoreButtonClicked}
+                    disabled={props.disableButtonsForLobbyPage}
```
JamesBurnside, line 169: I'm concerned there's still an underlying issue that the chat and people pane could be opened when not in a call yet from somewhere else (in the future) that we won't be protecting against.
Might be an over optimization but we could consider when we call `setActivePane(chat | people)` first ensuring that we are actually in a call (checking the `hasJoinedCall` variable) and setting the active pane unless we have joined a call, open to other suggestions or simply not doing this for now!

This fix should definitely should be kept though - nice simply fix too!

packages/react-components/src/components/FileCard.tsx
```diff
@@ -98,7 +99,12 @@ export const _FileCard = (props: _FileCardProps): JSX.Element => {
   };
 
   return (
-    <Stack className={containerClassName}>
+    <Stack
+      className={containerClassName}
+      onClick={() => {
```
JamesBurnside, line 104: Is this focusable by keyboard nav/screen reader? What does the screen reader read out when focused?

carocao-msft: Good catch! Turns out the whole file card is not tab accessible, let me look into a solution to this or should this be a separate bug?

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
carocao-msft: disableButtonsForLobbyPage={isInLobbyOrConnecting} ,const isInLobbyOrConnecting = currentPage === 'lobby'; this prop should be already detecting if we are in a call?

JamesBurnside: It will, and this will definitely solve the bug today!

My thinking is we still have a hole where a user can hit a totally blank page that might come up in the future. E.g. if we add a `CallWithChatComposite.openPane('people')` function then this could be called at a bad time causing the user to hit a blank white page. Or maybe we add the ability to click on your own tile and it opens the people pane from some option on your tile, we could uncover this issue again accidentally because its hard to spot and very missable in manual testing.
So might be worth putting in extra safeguarding today to avoid a regression that causes this to happen in the future, but also feel free to leave this be - totally up to you!

packages/react-components/src/components/FileCard.tsx
JamesBurnside: Probably best as a seperate bug and PR as I imagine it will add new strings etc and sometimes a11y fixes turn out to be quite large

JamesBurnside (approved):

dmceachernmsft (approved):

packages/react-components/src/components/FileCard.tsx
prabhjot-msft: Yeah, this is being fixed. I have PR up for the accessibility issue

prabhjot-msft: @carocao-msft I think we should wait for the PR(https://github.com/Azure/communication-ui-library/pull/1881) to be merged. It is using icon buttons which might resolve the issue. My concern is if we make the outer stack clickable and the file card is not tab navigable. User might not be able to trigger onclick using keyboard.

JamesBurnside (approved):

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
JamesBurnside: Nice this has been fixed with the latest hasJoinedCall checks! Thanks!

packages/react-components/src/components/FileCard.tsx
carocao-msft: Merging this PR for now. Will resolve accessibility issue if it occurs once https://github.com/Azure/communication-ui-library/pull/1881 is merged in

PR #1887, opened 2022-05-12
Make stable and beta API files peers

prprabhu-ms:
# What

Move generated API files around.

build flavor | before PR | after PR
-- | -- | --
beta | `packages/**/review` | `packages/**/review/beta`
stable | `packages/**/stable-review` | `packages/**/review/stable`

# Why

Housekeeping. There is no need to treat the two build flavors differently.


PorterNan (approved):

.github/workflows/ci.yml
```diff
@@ -74,7 +74,7 @@ jobs:
       - name: API Extractor Check
         run: |
```
JamesBurnside, line 75: When CI succeeds can you confirm the `Upload azure-communication api.md files for easy access` (though should say api.json files) was successful and the api files got uploaded correctly?

JamesBurnside (approved):

.github/workflows/ci.yml
prprabhu-ms: Will do. I did check that we're picking the files off the `/temp/` folder which shouldn't change.

PR #1888, opened 2022-05-12
Move UI test snapshots for beta and stable build flavors to peer directories

prprabhu-ms:
# What

Move UI snapshots around.

build flavor | before PR | after PR
-- | -- | --
beta | Alongside test files | `<test-root>/snapshots/beta/`
stable | `<packlet-root>/stable-snapshots` | `<test-root>/snapshots/stable/`

# Why

Housekeeping. There is no need to treat the two build flavors differently.

Similar in spirit to #1887


prprabhu-ms:
The mother of all `ui change` PRs :D

PorterNan (approved):

packages/react-composites/playwright.config.ts
```diff
@@ -9,6 +9,10 @@ const DESKTOP_VIEWPORT = {
   height: 768
 };
 
+const TEST_ROOT = './tests/browser';
+
+const buildFlavor: 'beta' | 'stable' = process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' ? 'stable' : 'beta';
```
JamesBurnside, line 14: I hope we don't add more flavors or we may miss some update spots 👀

JamesBurnside (approved):

packages/react-composites/playwright.config.ts
prprabhu-ms: Yeah, it's not super crisp how we've implemented this.

But then, even if we do add another flavor (`alpha`) it's a one-time pain. Not a routine task.

PR #1890, opened 2022-05-12
Support `updateScalingMode` for the local video stream in VideoGallery

JamesBurnside:
# What
Call updateScalingMode instead of recreating the stream in the local video tile

Prev PR in stack:
Next PR in stack:

# Why
This means we do not have to destroy and recreate the stream when the scaling mode has changed.

# How Tested
See next PR in stack for video

### DIFF
--- a/packages/react-components/src/components/LocalVideoTile.tsx
+++ b/packages/react-components/src/components/LocalVideoTile.tsx
@@ -3,8 +3,8 @@
 
 import { Stack } from '@fluentui/react';
 import { _formatString } from '@internal/acs-ui-common';
-import React, { useEffect, useMemo } from 'react';
-import { OnRenderAvatarCallback, VideoStreamOptions } from '../types';
+import React, { useEffect, useMemo, useRef } from 'react';
+import { OnRenderAvatarCallback, VideoStreamOptions, CreateVideoStreamViewResult } from '../types';
 import { LocalVideoCameraCycleButton, LocalVideoCameraCycleButtonProps } from './LocalVideoCameraButton';
 import { StreamMedia } from './StreamMedia';
 import { VideoTile, VideoTileStylesProps } from './VideoTile';
@@ -17,7 +17,7 @@ import { VideoTile, VideoTileStylesProps } from './VideoTile';
 export const LocalVideoTile = React.memo(
   (props: {
     userId: string;
-    onCreateLocalStreamView?: (options?: VideoStreamOptions) => Promise<void>;
+    onCreateLocalStreamView?: (options?: VideoStreamOptions) => Promise<void | CreateVideoStreamViewResult>;
     onDisposeLocalStreamView?: () => void;
     isAvailable?: boolean;
     isMuted?: boolean;
@@ -54,17 +54,79 @@ export const LocalVideoTile = React.memo(
       localVideoSelectedDescription
     } = props;
 
+    const [localStreamRendererResult, setLocalStreamRendererResult] = React.useState<CreateVideoStreamViewResult>();
+
+    // When localVideoViewOptions change two things may happen:
+    //
+    // 1. Just the scaling mode has changed and we have access to the original LocalStreamRenderView
+    //   - In this case we just need to call updateScalingMode on the view and do not need to recreate the stream view
+    // 2. isMirrored has changed, or scalingMode change and we don't have access to the original LocalStreamRenderView
+    //   - In this case we need to dispose the old view and create a new one
+    //
+    // For scenario 1, we hold onto a ref to the scalingMode that persists across renders. When localVideoViewOptions.scalingMode
+    // differs from the persistent ref, we know to call updateScalingMode. We then subsequently update the ref to the new value for
+    // future renders.
+    //
+    // For scenario 2, we extract the localVideoViewOptions.isMirrored to ensure any change to the isMirrorred prop triggers the useEffect
+    // that recreates the local stream view. We must also here add an extra check for when the scaling mode changes and we do not have
+    // access to the original LocalStreamView (and hence cannot call updateScalingMode). When this happens we must also trigger the useEffect
+    // that recreates the local stream view.
+
+    const scalingModeRef = useRef(localVideoViewOptions?.scalingMode);
+    const newIsMirrored = localVideoViewOptions?.isMirrored;
+    const newScalingMode = localVideoViewOptions?.scalingMode;
+    const hasScalingModeChanged = scalingModeRef.current !== newScalingMode;
+    const updatingScalingModeDirectly = hasScalingModeChanged && !!localStreamRendererResult;
+
+    if (isAvailable && renderElement && newScalingMode && updatingScalingModeDirectly) {
+      localStreamRendererResult && localStreamRendererResult.view.updateScalingMode(newScalingMode);
+      scalingModeRef.current = localVideoViewOptions?.scalingMode;
+    }
+
+    // scalingModeForUseEffect will trigger the useEffect to recreate the local stream view only if the scaling mode has changed and
+    // we cannot call updateScalingMode on the view object directly. Otherwise it will be null to ensure the useEffect does not trigger
+    // when scaling mode changes.
+    const scalingModeForUseEffect = updatingScalingModeDirectly ? null : newScalingMode;
+
     useEffect(() => {
+      // Avoid race condition where onDisposeLocalStreamView is called before onCreateLocalStreamView
+      // and setLocalStreamRendererResult have completed
+      let wasLocalStreamDisposed = false;
+
       if (isAvailable && !renderElement) {
-        onCreateLocalStreamView?.(localVideoViewOptions);
+        (async (): Promise<void> => {
+          console.log('calling onCreateLocalStreamView');
+          const streamRendererResult = await onCreateLocalStreamView?.({
+            isMirrored: newIsMirrored,
+            scalingMode: scalingModeForUseEffect === null ? scalingModeRef.current : scalingModeForUseEffect
+          });
+
+          console.log(
+            `completed onCreateLocalStreamView, wasLocalStreamDisposed: ${wasLocalStreamDisposed} streamRendererResult: `,
+            streamRendererResult
+          );
+          if (!wasLocalStreamDisposed) {
+            streamRendererResult && setLocalStreamRendererResult(streamRendererResult);
+          }
+        })();
       }
       // Always clean up element to make tile up to date and be able to dispose correctly
       return () => {
         if (renderElement) {
+          wasLocalStreamDisposed = true;
           onDisposeLocalStreamView?.();
         }
       };
-    }, [isAvailable, onCreateLocalStreamView, onDisposeLocalStreamView, localVideoViewOptions, renderElement]);
+    }, [
+      isAvailable,
+      onCreateLocalStreamView,
+      onDisposeLocalStreamView,
+      localVideoViewOptions,
+      renderElement,
+      localStreamRendererResult,
+      newIsMirrored,
+      scalingModeForUseEffect
+    ]);
 
     // The execution order for above useEffect is onCreateRemoteStreamView =>(async time gap) RenderElement generated => element disposed => onDisposeRemoteStreamView
     // Element disposed could happen during async time gap, which still cause leaks for unused renderElement.
--- a/change/@internal-react-components-51d61f5d-5bc8-478a-a806-411ec47dc32f.json
+++ b/change/@internal-react-components-51d61f5d-5bc8-478a-a806-411ec47dc32f.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Support updateScalingMode in the VideoGallery to avoid having to destroy and recreate the stream when the localVideoViewOption scaling mode is changed",
+  "packageName": "@internal/react-components",
+  "email": "2684369+JamesBurnside@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/react-components/review/react-components.api.md
+++ b/packages/react-components/review/react-components.api.md
@@ -262,6 +262,14 @@ export interface ControlBarProps {
 }
 
 // @public
+export interface CreateVideoStreamViewResult {
+    // (undocumented)
+    view: {
+        updateScalingMode: (scalingMode: ScalingMode) => Promise<void>;
+    };
+}
+
+// @public
 export type CustomAvatarOptions = {
     coinSize?: number;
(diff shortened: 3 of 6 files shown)


packages/communication-react/review/communication-react.api.md
```diff
@@ -1324,6 +1324,14 @@ export const createStatefulCallClient: (args: StatefulCallClientArgs, options?:
 // @public
 export const createStatefulChatClient: (args: StatefulChatClientArgs, options?: StatefulChatClientOptions | undefined) => StatefulChatClient;
 
+// @public
+export interface CreateVideoStreamViewResult {
+    // (undocumented)
```
prprabhu-ms, line 1329: Please document new public type.

packages/communication-react/review/communication-react.api.md
```diff
@@ -2263,6 +2271,9 @@ export interface RemoteVideoStreamState {
     view?: VideoStreamRendererViewState;
 }
 
+// @public
+export type ScalingMode = 'Stretch' | 'Crop' | 'Fit';
```
prprabhu-ms, line 2275: Is there a reason we're redefining this type instead of using @azure/communication-calling#ScalingMode. The values are identical, and presumably the only way we use this type is to forward it to the headless SDK.

packages/react-components/src/components/LocalVideoTile.tsx
```diff
@@ -54,17 +54,79 @@ export const LocalVideoTile = React.memo(
...
+
+    const scalingModeRef = useRef(localVideoViewOptions?.scalingMode);
+    const newIsMirrored = localVideoViewOptions?.isMirrored;
+    const newScalingMode = localVideoViewOptions?.scalingMode;
+    const hasScalingModeChanged = scalingModeRef.current !== newScalingMode;
+    const updatingScalingModeDirectly = hasScalingModeChanged && !!localStreamRendererResult;
+
+    if (isAvailable && renderElement && newScalingMode && updatingScalingModeDirectly) {
+      localStreamRendererResult && localStreamRendererResult.view.updateScalingMode(newScalingMode);
+      scalingModeRef.current = localVideoViewOptions?.scalingMode;
```
prprabhu-ms, line 83: This sets the new scalingMode before the view is actually updated (because `updateScalingMode` is async).

Is this intentional? I worry that this desync can cause hard to debug bugs should `updateScalingMode` fail in some scenarios.

packages/react-components/src/components/LocalVideoTile.tsx
```diff
@@ -54,17 +54,79 @@ export const LocalVideoTile = React.memo(
...
+
+    // scalingModeForUseEffect will trigger the useEffect to recreate the local stream view only if the scaling mode has changed and
+    // we cannot call updateScalingMode on the view object directly. Otherwise it will be null to ensure the useEffect does not trigger
+    // when scaling mode changes.
+    const scalingModeForUseEffect = updatingScalingModeDirectly ? null : newScalingMode;
+
     useEffect(() => {
+      // Avoid race condition where onDisposeLocalStreamView is called before onCreateLocalStreamView
+      // and setLocalStreamRendererResult have completed
+      let wasLocalStreamDisposed = false;
```
prprabhu-ms, line 94: Ugh. Is there *any* way you can extract a (wrapper) component that takes care of just updating the scaling mode?

The logic in this component has been out of hands for a while, and this PR adds more complexity :(

prprabhu-ms: Honestly, I almost want to encapsulate all the logic around how we create / update / dispose views outside of React into a class that exposes a simpler API that the react component then uses. React component lifecycle just doesn't seem like a good place to manage the state of the view.

I'd have to give it a spin to really know if such an extraction is feasible. Just a thought: Not everything must be done as part of the React lifecycle.

packages/react-components/src/components/LocalVideoTile.tsx
```diff
@@ -54,17 +54,79 @@ export const LocalVideoTile = React.memo(
...
+
       if (isAvailable && !renderElement) {
-        onCreateLocalStreamView?.(localVideoViewOptions);
+        (async (): Promise<void> => {
+          console.log('calling onCreateLocalStreamView');
```
prprabhu-ms, line 98: nit: remove console.log

packages/react-components/src/components/LocalVideoTile.tsx
```diff
@@ -54,17 +54,79 @@ export const LocalVideoTile = React.memo(
...
+          const streamRendererResult = await onCreateLocalStreamView?.({
+            isMirrored: newIsMirrored,
+            scalingMode: scalingModeForUseEffect === null ? scalingModeRef.current : scalingModeForUseEffect
```
prprabhu-ms, line 101: Oh cool (?), so eslint allows you to use `scalingModeRef` in the `useEffect` without adding it to dependency array?

That _is_ what you want here, but I expected you to need to add an eslint exception for it.

packages/react-components/src/components/LocalVideoTile.tsx
```diff
@@ -54,17 +54,79 @@ export const LocalVideoTile = React.memo(
...
+          });
+
+          console.log(
```
prprabhu-ms, line 104: nit: remove console log

packages/react-components/src/components/LocalVideoTile.tsx
```diff
@@ -54,17 +54,79 @@ export const LocalVideoTile = React.memo(
...
     useEffect(() => {
+      // Avoid race condition where onDisposeLocalStreamView is called before onCreateLocalStreamView
+      // and setLocalStreamRendererResult have completed
+      let wasLocalStreamDisposed = false;
+
       if (isAvailable && !renderElement) {
-        onCreateLocalStreamView?.(localVideoViewOptions);
+        (async (): Promise<void> => {
+          console.log('calling onCreateLocalStreamView');
+          const streamRendererResult = await onCreateLocalStreamView?.({
```
prprabhu-ms, line 99: TIL: JavaScript guarantees that an `async` function will run to the first `await` expression in its body synchronously: 
image

This is why you don't need to check for `wasLocalStreamDisposed` before calling `onCreateLocalStreamView`.

This need not be this way by the way. Other languages don't require concurrent stuff to being immediately (e.g. Go's goroutines can be scheduled at any time by the runtime, even if running on a single thread).

prprabhu-ms:
Thanks for taking a stab at this. Hairy bit of code there...

packages/react-components/src/components/LocalVideoTile.tsx
JamesBurnside: My guess is react has some internal knowledge that a React.MutableRef shouldn't be included in the useEffect array as it should be constant across all renders

packages/react-components/src/components/LocalVideoTile.tsx
JamesBurnside: Good point, updating to be inside an async block inside a useEffect... the complexity grows

JamesBurnside: Although doesn't require being in a useEffect, just feels weird to have a bare async call on a render method

JamesBurnside: updated

packages/react-components/src/components/LocalVideoTile.tsx
JamesBurnside: I've moved to a custom hook for now to atleast put the complex logic in one, isolated place, and a place reusable by the remote video stream (will be updating in a seperate PR).
In theory you could have what you're actually wanting, perhaps a manager class stored in a useRef or upgrade the functional component to a class component - but don't have the time to look into this now

prprabhu-ms (approved):
Custom hook is a good compromise here.

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const [streamRendererResult, setStreamRendererResult] = React.useState<CreateVideoStreamViewResult>();
+  const scalingModeRef = useRef(scalingMode);
+  const hasScalingModeChanged = scalingModeRef.current !== scalingMode;
+  const updatingScalingModeDirectly = hasScalingModeChanged && !!streamRendererResult;
+
+  if (isStreamAvailable && renderElementExists && scalingMode && updatingScalingModeDirectly) {
+    (async () => {
+      streamRendererResult && (await streamRendererResult.view.updateScalingMode(scalingMode));
+      scalingModeRef.current = scalingMode;
```
prprabhu-ms, line 51: Pushing on this a bit more. Thinking through various scenarios that can happen while `updateScalingMode` is blocked.

- We rerender without any changes to props at all (for this hook at least) -- we'll end up calling `updateScalingMode` again (because line 45 will result in `true`)
- stream is disposed (is this possible?): we might end up setting `scalingModeRef.current` to an incorrect value and then have the wrong `scalingMode` going forward. - I'm not a 100% sure of this, but would like to talk this through and document (in comments) why it is not possible.

JamesBurnside: For number 1 - nice catch -- may need an isUpdatingRef check (gross), will look into this.

For number 2:
a. stream is disposed through onComponentDismount -- not concerned about this as component has dismounted
b. stream is disposed because isMirrored or one of the other useEffect dependencies on L85 changes -- another nice catch -- this needs the below useEffect to signal this code block to not update the scalingModeRef... which sounds messy.. I'll have a look into this

JamesBurnside: I could set `scalingModeRef.current = scalingMode` first, but then on failure of updateScalingMode it would be incorrect

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  // scalingModeForUseEffect will trigger the useEffect to recreate the stream view only if the scaling mode has changed and
+  // we cannot call updateScalingMode on the view object directly. Otherwise it will be null to ensure the useEffect does not trigger
+  // when scaling mode changes.
+  const scalingModeForUseEffect = updatingScalingModeDirectly ? null : scalingMode;
+
+  useEffect(() => {
+    // Avoid race condition where onDisposeStreamView is called before onCreateStreamView
+    // and setStreamRendererResult have completed
+    let wasStreamDisposed = false;
```
prprabhu-ms, line 63: This won't work. The value is captured in the closure.

prprabhu-ms:
OK, ok... I'm really done here.

We should get two other approvers, now that I'm a co-author.

packages/communication-react/review/communication-react.api.md
prprabhu-ms: Ugh... looks like I somehow pushed back an old change?

prprabhu-ms: This looks undocumented again.

packages/react-components/src/components/utils/useCancellableTask.ts
```diff
@@ -0,0 +1,125 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
+
+  const response: [TriggerFunc, CancelFunc] = useMemo(() => {
+    const trigger = (action: (cancellable: Cancellable) => Promise<void>): void => {
+      const cancellable = cancelMarker.reset();
+      action(cancellable);
+    };
+    const cancel = (): void => {
```
prprabhu-ms, line 104: nit: this can just be `return [trigger, cancelMarker.cancel]`

packages/react-components/src/components/utils/useCancellableTask.ts
```diff
@@ -0,0 +1,125 @@
...
+  // Memoize all internal objects and returned values so that
+  // this hook returns referentially stable values on each render.
+  const cancelMarker = useMemo(() => {
+    return new CancelMarker();
+  }, []);
+
+  // Cancel any pending task when component unmounts.
+  useEffect(
+    () => {
+      return () => {
```
prprabhu-ms, line 91: nit: this can just be `return cancelMarker.cancel`

packages/react-components/src/components/utils/useCancellableTask.ts
JamesBurnside: updated

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  const [streamRendererResult, setStreamRendererResult] = React.useState<CreateVideoStreamViewResult>();
+  const scalingModeRef = useRef(scalingMode);
+  const hasScalingModeChanged = scalingModeRef.current !== scalingMode;
+  const updatingScalingModeDirectly = hasScalingModeChanged && !!streamRendererResult;
+
+  const [triggerRescale, cancelRescale] = useCancellableTask();
+  const [triggerCreateStreamView, cancelCreateStreamView] = useCancellableTask();
+
+  if (isStreamAvailable && renderElementExists && scalingMode && updatingScalingModeDirectly) {
```
PorterNan, line 52: Do we need to do useeffect here?

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -59,7 +59,7 @@ import { PhoneNumberKind } from '@azure/communication-common';
 import { default as React_2 } from 'react';
 import type { RemoteParticipant } from '@azure/communication-calling';
 import { RemoteParticipantState as RemoteParticipantState_2 } from '@azure/communication-calling';
-import { ScalingMode } from '@azure/communication-calling';
+import { ScalingMode as ScalingMode_2 } from '@azure/communication-calling';
```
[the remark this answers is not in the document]
JamesBurnside: I like ViewScalingMode as it applies directly to a "view" and not some other scaling -- and then avoids the conflict nicely as you mention

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
JamesBurnside: For this it's simply redundant if we do because we have the `updatingScalingModeDirectly` check - so we could have it be we left it out because it just adds a few more lines of nesting and little bit extra complexity

PR #1889, opened 2022-05-12
Return the renderer and view created when statefulCallClient.createView is called

JamesBurnside:
# What
* Return the renderer and view created when statefulCallClient.createView is called

Next PRs in stack:
* #1890
* #1891

# Why
Currently we throw away the view so the likes of view.updateScalingMode cannot be called.
See next PR in stack for our stateful and components making use of this.

# How Tested
See final PR in stack: #1891

PR #1891, opened 2022-05-12
Pipe the created view through the composite interfaces and component-bindings for supporting updateScalingMode

JamesBurnside:
# What
Piping through data from the previous two PRs in the stack:
* PR1: #1889
* PR2: #1890

One further PR todo: make the localAndRemotePiP use the new LocalVideoTile component to get this new behavior

# How Tested
before (notice flash where stream is destroyed and recreated): 

https://user-images.githubusercontent.com/2684369/168173159-c01f6ecf-6d54-413c-9c50-c77e5159d3e7.mp4

after (no flash!):

https://user-images.githubusercontent.com/2684369/168173196-5e899632-4935-41b7-b8a9-cf80784e0ee1.mp4


packages/calling-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -218,7 +220,8 @@ export const createDefaultCallingHandlers = memoizeOne(
         return;
       }
 
-      await callClient.createView(call.id, undefined, localStream, options);
+      const { view } = (await callClient.createView(call.id, undefined, localStream, options)) ?? {};
```
prprabhu-ms, line 223: I want Rust's `Option<T>.map(|value| Ok(value)` (Look inside this optional thing, if you find it modify the value using this closure and return back an option) :)

prprabhu-ms (approved):
Love how this PR "binds" together prior PRs in the stack!

packages/react-components/src/components/LocalVideoTile.tsx
```diff
@@ -17,7 +17,7 @@ import { VideoTile, VideoTileStylesProps } from './VideoTile';
 export const LocalVideoTile = React.memo(
   (props: {
     userId: string;
-    onCreateLocalStreamView?: (options?: VideoStreamOptions) => Promise<void>;
+    onCreateLocalStreamView?: (options?: VideoStreamOptions) => Promise<void | CreateVideoStreamViewResult>;
```
PorterNan, line 20: Just curious, will this still work when we usePropsFor, wondering if auto complete is still working fine for this

JamesBurnside: It should JustWork, here's what the prop looks like:
image

PorterNan (approved):

PR #1895, opened 2022-05-13
Remove dependency on `azure/communication-signaling` beta package in our npm package

JamesBurnside:
# What
Remove dependency on `azure/communication-signaling` beta package in our npm package.

Still required as a dev dependency in some packages due to types required for mocks (chat package still uses this beta package under the hood, but at least our released npm package no longer needs an explicit dependency on it)

# Why
Chat package now exports these types we were using from `@azure/communication-signaling` so we can remove this beta package from our list of dependencies for good cleanliness.

# How Tested
Todo - quickly check removing chat participant in chat sample.
Hopefully nothing should be impacted as changes are (almost entirely) just types not runtime code.


samples/Chat/src/app/ChatScreen.tsx
```diff
@@ -38,13 +39,9 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
     async (adapter: ChatAdapter): Promise<ChatAdapter> => {
       adapter.on('participantsRemoved', (listener) => {
-        // Note: We are receiving ChatParticipant.id from communication-signaling, so of type 'CommunicationIdentifierKind'
-        // while it's supposed to be of type 'CommunicationIdentifier' as defined in communication-chat
-        const removedParticipantIds = listener.participantsRemoved.map(
-          (p) => (p.id as CommunicationUserIdentifier).communicationUserId
-        );
+        const removedParticipantIds = listener.participantsRemoved.map((p) => toFlatCommunicationIdentifier(p.id));
         if (removedParticipantIds.includes(userId)) {
-          const removedBy = (listener.removedBy.id as CommunicationUserIdentifier).communicationUserId;
```
prprabhu-ms, line 47: sneaky.

prprabhu-ms (approved):
<3

carocao-msft (approved):

PR #1896, opened 2022-05-14
Only generate json report

PorterNan:
# What

Only generate json report instead of stats.json file, which is around 500kb instead of 10+Mb size (so we are good to keep it in our repo)
# Why

# How Tested

# Process & policy checklist


JamesBurnside:
Is it intentional to check in the report.json files?

JamesBurnside:
Holding off on approval until #1897 is approved as these .json files should only be checked in if we go ahead with updating them in every CI action

PR #1897, opened 2022-05-14
Compare bundle size in CI comment

PorterNan:
# What

Compare bundle size in CI comment, it pulls base ref and current ref, then read json file to see if bundle size increase or not
# Why

# How Tested

# Process & policy checklist


anjulgarg:
I assigned it to myself since I want to go through it on monday. :)

.github/workflows/ci.yml
```diff
@@ -345,6 +345,37 @@ jobs:
...
+        run: |
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config user.name "github-actions[bot]"
+      - name: Commit bundle report
+        if: ${{ steps.changescheck.outputs.hasChanged == 'true' }}
+        env:
+          # use GITHUB_TOKEN so it won't trigger recursive PR validation
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        run: |
+          git checkout ${{ github.event.pull_request.head.ref }}
```
JamesBurnside, line 374: note: is this isn't a pull request presumably this ref won't exist - e.g. Ci on main branch, or if action run directly from workflow_dispatch

JamesBurnside: never mind the if checks for pull requets!

JamesBurnside:
Can you add to the PR description the runtime length impact to the github actions being affected? Am I correct in saying if the bundle does change it checks in a new file and the CI has to restart?

PorterNan:
> Can you add to the PR description the runtime length impact to the github actions being affected? Am I correct in saying if the bundle does change it checks in a new file and the CI has to restart?

That's a nice concern, I was thinking of this before - remember we got the problem that github account wouldn't trigger new build when they check in new changes? This is actually a feature for this one:D
image

PorterNan:
> > Can you add to the PR description the runtime length impact to the github actions being affected? Am I correct in saying if the bundle does change it checks in a new file and the CI has to restart?
> 
> That's a nice concern, I was thinking of this before - remember we got the problem that github account wouldn't trigger new build when they check in new changes? This is actually a feature for this one:D image

BTW, I realize if we can't make this work, we could only check in the json file after PR in main CI build - which is definitely less rebuild for validation

.github/workflows/ci.yml
```diff
@@ -345,6 +345,37 @@ jobs:
         run: rush build -o calling
       - name: Tests
         run: rush test -o calling
+      # Not check in the bundles stats if size doesn't change
+      # Stats size in the file is a bit random, so we can't only depend on file itself
+      - name: Check size change
+        id: changescheck
+        if: ${{ matrix.flavor == 'beta' && github.event_name == 'pull_request' }}
+        run: |
+          base_size=$(jq -r ".[0].parsedSize" samples/Calling/bundle-report/report.json)
```
anjulgarg, line 354: base_size -> new_size
might be a more indicative variable?

.github/workflows/ci.yml
```diff
@@ -345,6 +345,37 @@ jobs:
         run: rush build -o calling
       - name: Tests
         run: rush test -o calling
+      # Not check in the bundles stats if size doesn't change
+      # Stats size in the file is a bit random, so we can't only depend on file itself
+      - name: Check size change
+        id: changescheck
```
anjulgarg, line 351: nit:
changescheck -> compareBundleSize
might be a more indicative name?

.github/workflows/ci.yml
```diff
@@ -345,6 +345,37 @@ jobs:
...
+        if: ${{ matrix.flavor == 'beta' && github.event_name == 'pull_request' }}
+        run: |
+          base_size=$(jq -r ".[0].parsedSize" samples/Calling/bundle-report/report.json)
+          current_size=$(jq -r ".[0].parsedSize" samples/Calling/dist/build/report.json)
+          change=$([ $current_size = $base_size ] && echo "true" || echo "false")
+          echo "::set-output name=hasChanged::${change}"
+      - name: Copy current bundle report json
+        if: ${{ steps.changescheck.outputs.hasChanged == 'true' }}
+        run: cp samples/Calling/dist/build/report.json samples/Calling/bundle-report/
```
anjulgarg, line 360: Why are we replacing the new report with the old one here?

PorterNan: Because we want to check the new report in, so that new report will be in the current branch, and ready to be checked in when merged

mgamis-msft (approved):

mgamis-msft:
Only takes seconds to compare. Nice.
image

PorterNan:
> 

Thanks @mgamis-msft for pointing those numbers, I didn't look at those time and I thought I would be like 10-20s, sounds like it is even better! This "gist" way was suggested by @JamesBurnside in last PR, I thought I was hacky at the very beginning, but realized we already got a machine account running so it's not a problem to have gist files hosted there

.github/workflows/ci.yml
PorterNan: Remove this step because I refactor the way we do this

PR #1898, opened 2022-05-16
Cancel previous CI runs when newer changes are pushed in a PR

JamesBurnside:
# What
Cancel previous CI runs when newer changes are pushed in a PR.

## How does this work

https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre/67939898#67939898

Code snippet being added:
```
concurrency: 
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

This creates a "concurrency grouping" with `cancel-in-progress: true` where if another action is triggered in that concurrency group, it will cancel any in progress ones.

To ensure the correct action is cancelled, it creates a group name using:
* `github.workflow` id to ensure it only cancels runs of the same github action
* `github.ref` to ensure only runs targeting the same branch/PR are cancelled.

Put those together and we cancel runs of the same action on the same branch.

# Why
Currently if you push 5 changes one after another to your PR, our github actions will run all actions for all those changes creating a ton of waste. This change will ensure old actions will be cancelled if newer changes are pushed (or newer actions triggered on the same branch).

This will also cancel old runs if you manually run an action twice on the same branch - for better or worse -- **I added this to all our github actions, please review if any should omit this** (e.g. if we would want to runs of the same action on the same branch)

# How Tested
Action run was successfully cancelled when new `add comments` change was pushed: https://github.com/Azure/communication-ui-library/runs/6456075631.
And successfully cancelled that run when "merge new changes from main" was clicked:
image

### DIFF
--- a/.github/workflows/alpha-release.yml
+++ b/.github/workflows/alpha-release.yml
@@ -4,6 +4,7 @@ on:
   # Allows running workflow manually from the GitHub Actions tab
   workflow_dispatch:
 
+# cancel workflow when a newer version of the workflow is triggered on the same github ref
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/alpha-release.yml
+++ b/.github/workflows/alpha-release.yml
@@ -4,6 +4,10 @@ on:
   # Allows running workflow manually from the GitHub Actions tab
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build_and_test:
     name: Build and Test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
   # Allows you to run this workflow manually from the Actions tab.
   workflow_dispatch:
 
+# cancel workflow when a newer version of the workflow is triggered on the same github ref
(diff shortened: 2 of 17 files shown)


.github/workflows/alpha-release.yml
```diff
@@ -4,6 +4,11 @@ on:
   # Allows running workflow manually from the GitHub Actions tab
   workflow_dispatch:
 
+# cancel workflow when a newer version of the workflow is triggered on the same github ref
```
dmceachernmsft, line 7: This should only run once a day? as well should we merge changes into main wouldn't that cancel the action as well? or does this only get cancelled if someone manually kicks off alpha?

prprabhu-ms (approved):

.github/workflows/alpha-release.yml
prprabhu-ms: After reviewing your PR, I stopped myself after spending 10 minutes looking at ways to generate the github actions YAML files from starlark / JS and checking the generated files in.

I really want to dedupe some of our GH actions files, but I'm unsure if using non-standard tooling is worth the deduplication (there's learning cost for folks on the team).

Perhaps we should explore composition by writing (local) actions for common stuff: https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action

dmceachernmsft (approved):

PR #1899, opened 2022-05-17
Add custom data model to the chat thread display names

dmceachernmsft:
# What

Adds to the chat thread overrides for the displayNames associated with messages
image

# Why

Allows more complete coverage of our custom data model feature

https://skype.visualstudio.com/SPOOL/_workitems/edit/2833447
# How Tested

Ran locally with custom data

### DIFF
--- a/packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
+++ b/packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
@@ -3,19 +3,23 @@
 
 import { mergeStyles, Stack } from '@fluentui/react';
 import {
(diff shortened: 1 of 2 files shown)


dmceachernmsft:
Making this as a draft since I am unsure about the implementation on this one. I feel like the functionality here is supposed to use our new hook that we added for the MediaGallery for the calling side of this problem. The way that messages store user data, we can build the object to make that list but mapping back the new displayNames to the messages is proving to be a challenge.

We could do it that way for just the displayName on the message but the read receipts, system messages, and typing indicators would have to be done seperately

PR #1901, opened 2022-05-17
Log structural telemetry event for better parse

PorterNan:
# What

Log structural telemetry event for better parse

Currently we use plain text to log telemetry event, which is simple but create a lot of hardness in dashboard event grouping and more, this is the first step to log structural log event data into app insight. As chat and calling SDK also mentioned they would like to follow us to provide a better syntax for our contoso app to consume, let's see how it works.

### This is also a pre-required PR change for perf counter PR (PR will be coming soon)

# Why

# How Tested

# Process & policy checklist


edwardlee-msft (approved):
Love this! Hopefully, this will make logs easier to query :)

packages/acs-ui-common/src/logEvent.ts
```diff
@@ -0,0 +1,27 @@
...
+  level: EventLevel;
+  data?: Record<string, unknown>;
+};
+
+/**
+ * @internal
+ * This is a log function to log structural data for easier parse in telemetry
+ */
+export const logEvent = (logger: AzureLogger, event: TelemetryEvent): void => {
+  logger[event.level](JSON.stringify(event));
```
JamesBurnside, line 26: use `_safeJSONStringify` here

packages/acs-ui-common/src/logEvent.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { AzureLogger } from '@azure/logger';
+
+/**
+ * @internal
+ */
+export type EventLevel = 'verbose' | 'info' | 'warning' | 'error';
```
JamesBurnside, line 9: Can you import `AzureLogLevel` from `'@azure/logger'` instead of defining these ourselves

PR #1904, opened 2022-05-17
[DRAFT] Moved FileSharing.test.ts to use fake adapter

mgamis-msft:
# What

# Why

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/chat/fake-adapter/FileSharing.test.ts
```diff
@@ -254,48 +225,47 @@ test.describe('Filesharing Message Thread', async () => {
...
 
-    await stubMessageTimestamps(page0);
-    expect(await page0.screenshot()).toMatchSnapshot('filesharing-file-download-card-in-sent-messages.png');
+    await stubMessageTimestamps(page);
+    expect(await page.screenshot()).toMatchSnapshot('filesharing-file-download-card-in-sent-messages.png');
+  });
 
-    const page1 = pages[1];
-    await waitForTypingIndicatorHidden(page1);
+  test('contains File Download Card in remote message', async ({ serverUrl, page }) => {
```
mgamis-msft, line 241: I had to create a separate test here with qArg `hasRemoteFileSharingMessage` to get filesharing message from a remote participant to get snapshot 'filesharing-file-download-card-in-received-messages.png'

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -27,17 +39,56 @@ export const FakeAdapterApp = (): JSX.Element => {
     verifyParamExists(params.fakeChatAdapterModel, 'fakeChatAdapterModel')
   ) as ChatAdapterModel;
 
+  // Optional params
+  const useFileSharing = Boolean(params.useFileSharing);
```
JamesBurnside, line 43: avoid `use*` prefix unless this is a react hook
```suggestion
const fileSharingEnabled = Boolean(params.fileSharingEnabled);
```
or if this is doing something more specific e.g. attaching a message to the send box, name this to have that specificity:
```suggestion
const attachFileToInitialSendMessage = Boolean(params.attachFileToInitialSendMessage);
```

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -2,18 +2,30 @@
...
 import { nanoid } from 'nanoid';
 import React, { useEffect, useState } from 'react';
-import { ChatAdapter, ChatComposite, createAzureCommunicationChatAdapterFromClient } from '../../../../src';
+import {
+  ChatAdapter,
+  ChatComposite,
+  createAzureCommunicationChatAdapterFromClient,
+  FileDownloadError,
+  FileDownloadHandler
+} from '../../../../src';
```
JamesBurnside, line 21: nit: Try avoid imports direct from the src's index.ts as it can cause circular import chains

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -55,29 +115,50 @@ export const FakeAdapterApp = (): JSX.Element => {
   );
 };
 
-async function createFakeChatAdapter(model: ChatAdapterModel): Promise<ChatAdapter> {
-  const chatService = new FakeChatService();
+async function createFakeChatAdapter(
+  model: ChatAdapterModel,
+  hasRemoteFileSharingMessage: boolean
+): Promise<ChatAdapter> {
+  const m = new Model({ asyncDelivery: false });
```
JamesBurnside, line 122: this creates a new Model called `m` but also takes in `model` as a function parameter - what is the difference between these two? (why do we create a new one here?0

JamesBurnside: actually I don't see m used, was it left in from testing things?

packages/react-composites/tests/browser/chat/app/fake-back-end/Model.ts
```diff
@@ -48,6 +48,23 @@ export class Model {
       });
     }
   }
+
+  public initializeMessages(threadId: string, messages: BasicChatMessage[]) {
+    messages.forEach((message) => {
+      this.checkedGetThread(message.sender, threadId);
```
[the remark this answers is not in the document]
JamesBurnside: Can this be used instead of `this.threadMap[threadId].thread`?

packages/react-composites/tests/browser/chat/app/fake-back-end/Model.ts
```diff
@@ -48,6 +48,23 @@ export class Model {
...
+      this.threadMap[threadId].thread.messages = [
+        ...this.threadMap[threadId].thread.messages,
+        {
+          id: `${this.threadMap[threadId].thread.messages}`,
+          sequenceId: `${this.threadMap[threadId].thread.messages}`,
```
JamesBurnside, line 59: this is a little confusing to read here, can you add a fn to the thread interface that is pushMessages(messages) or similar that handles updating the id, sequenceId and other default values.

JamesBurnside (approved):
Looks great to me! Make sure to get sign off from someone working on file sharing also

packages/react-composites/tests/browser/chat/app/fake-back-end/types.ts
```diff
@@ -42,3 +42,5 @@ export interface NetworkEventModel {
    */
   maxDelayMilliseconds?: number;
 }
+
+export type BasicChatMessage = Pick<ChatMessage, 'sender' | 'senderDisplayName' | 'content' | 'metadata'>;
```
prprabhu-ms, line 46: Let's keep the `FakeChatClient` totally isolated from e2e testing concerns.

This `BasicChatMessage` is related to how the tests want to initialize the fake. So move this to the test helpers etc.

See also my comment in `FakeChatClient.ts`

packages/react-composites/tests/browser/chat/app/fake-back-end/FakeChatClient.ts
```diff
@@ -85,6 +85,11 @@ export class FakeChatClient implements IChatClient {
     });
   }
 
+  initializeMessages(threadId: string, messages: BasicChatMessage[]): Promise<void> {
```
prprabhu-ms, line 88: Let's keep the fake completely isolated from test logic.

The constructor for `FakeChatClient` already accepts a preconstructed `Model` to initialize with. The test should construct the appropriate model and initialize the `FakeChatClient` with the `Model`.

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -28,24 +36,61 @@ const params = Object.fromEntries(urlSearchParams.entries());
...
-  const fakeChatAdapterModel = JSON.parse(
-    verifyParamExists(params.fakeChatAdapterModel, 'fakeChatAdapterModel')
-  ) as ChatAdapterModel;
+  const fakeChatAdapterArgs = JSON.parse(
+    verifyParamExists(params.fakeChatAdapterArgs, 'fakeChatAdapterArgs')
+  ) as FakeChatAdapterArgs;
 
   // Optional params
   const useFrLocale = Boolean(params.useFrLocale);
+  const fileSharingEnabled = Boolean(params.fileSharingEnabled);
```
prprabhu-ms, line 45: Instead of adding query args for all these cases, can the test create the appropriate `Model` in some of these cases and pass the serialized `Model` as a single query argument?

The fewer the layers between the test and the eventual model / adapter, the easier it is to understand and add tests.

I know we had some trouble earlier importing the types in the test code. Can we give that another look. We seem to be bending over backwards here to pass all this test data, and it makes working with the e2e tests much harder.

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -28,24 +36,61 @@ const params = Object.fromEntries(urlSearchParams.entries());
...
-      setAdapter(await createFakeChatAdapter(fakeChatAdapterModel));
+      setAdapter(await createFakeChatAdapter(fakeChatAdapterArgs, hasRemoteFileSharingMessage));
     };
 
     initialize();
     return () => adapter && adapter.dispose();
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  React.useEffect(() => {
```
prprabhu-ms, line 61: Our live test e2e apps are already quite complex and in my opinion not easy to maintain.

Since we're doing new work here, let's learn from the last attempt. I'd recommend splitting off a separate test app for fake file sharing so that the logic there is all related to file sharing only.

Then you can use a single query arg to select that app from index.tsx (or even better, use react-router to pick the right app instead of query arguments).

packages/react-composites/tests/browser/chat/app/fake-back-end/FakeChatClient.ts
prprabhu-ms: UPDATE:

I see that `Model` does not have to initialize the thread and messages.

You have two options:

1. Add a constructor for `Model` that lets you initialize the data. But use the data types from the headless SDK. The fake uses the headless DTOs. No need to invent another one.
2. Simply use the `ChatClient` that you have created to create a thread, add users and send messages as if you're using the real service.

I recommend (2). It needs not new API in the fake, and is easier to extend (you can get the fake in any state you want by using the public headless API, at least until we want error injection).

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -64,31 +118,52 @@ export const FakeAdapterApp = (): JSX.Element => {
...
       displayName: user
     };
   });
-  const localUser = { id: { communicationUserId: nanoid() }, displayName: model.localParticipant };
-  const firstChatClient = chatService.newClient(localUser.id);
-  const thread = await firstChatClient.createChatThread(
+  const localUser = { id: { communicationUserId: nanoid() }, displayName: fakeAdapterArgs.localParticipant };
+  const chatClient = new FakeChatClient(chatClientModel, localUser.id);
+  const participants = remoteParticipants;
+  const splicePosition =
```
prprabhu-ms, line 135: What is this dark magic?

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -64,31 +118,52 @@ export const FakeAdapterApp = (): JSX.Element => {
   );
 };
 
-async function createFakeChatAdapter(model: ChatAdapterModel): Promise<ChatAdapter> {
-  const chatService = new FakeChatService();
-  const remoteParticipants: ChatParticipant[] = model.remoteParticipants.map((user) => {
+async function createFakeChatAdapter(
+  fakeAdapterArgs: FakeChatAdapterArgs,
+  attachFileToInitialSendMessage: boolean
```
prprabhu-ms, line 123: This flag argument is adding a lot of complexity to this function. We should be able to send the first message etc *after* the adapter creation.

prprabhu-ms (changes requested):
I have substantive comments on this one. Feel free to schedule some time to talk IRL.

mgamis-msft:
> I have substantive comments on this one. Feel free to schedule some time to talk IRL.

Hey @prprabhu-ms, I addressed most of the comments and am now able import types in the test code by upgrading playwright. Please take another look.

change/@internal-react-composites-4beccd45-629a-4c8a-9905-daf1e36ccee0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
prprabhu-ms, line 2: nit: type should be `none`. Playwright is a `devDependency` that doesn't affect the final package in any way.

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -31,34 +31,82 @@ const params = Object.fromEntries(urlSearchParams.entries());
...
+      const chatClient = new FakeChatClient(chatClientModel, fakeChatAdapterArgs.localParticipant.id);
+      const thread = await chatClient.createChatThread(
+        {
+          topic: 'Cowabunga'
+        },
+        {
+          participants
+        }
+      );
+      const adapterInfo = {
```
prprabhu-ms, line 56: nitty nit. Reads cleaner: 
```
const newAdapter = await initializeAdapter({ ... });
setAdapter(newAdapter);
```

packages/react-composites/tests/browser/chat/fake-adapter/FakeChatAdapterArgs.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { ChatParticipant } from '@azure/communication-chat';
+import { FileMetadata } from '@internal/react-components';
+
+export type FileUpload = FileMetadata & { uploadComplete?: boolean; error?: string; progress?: number };
+
+export type FakeChatAdapterArgs = {
```
prprabhu-ms, line 9: Let's add docstrings to any non-trivial fields here. For example, looking at this I don't understand what `localParticipantPosition` does. When writing a new test, it would be helpful to be able to understand all these args quickly.

prprabhu-ms (approved):
I see some extra space in a few screenshots: 
image

Other than that, LGTM. Worth having Anjul take a look as well.

Kudos for getting these tests converted!

PR #1906, opened 2022-05-18
Update remote video tile to use the `useVideoStreamLifecycleMaintainer`

JamesBurnside:
# What
Update remote video tile to use the `useVideoStreamLifecycleMaintainer` created in #1890
Update `useVideoStreamLifecycleMaintainer` to consider remote streams.

# Why
Reuse logic now contained in a custom hook. Follow up PR will enabled remote tile to make use of `updateScalingMode`

# How Tested
Ran locally and remote videos stopped and started correctly with multiple users in a call, also tested one user turning off and on screensharing

### DIFF
--- a/packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
+++ b/packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
@@ -6,22 +6,45 @@ import { VideoStreamOptions, CreateVideoStreamViewResult, ScalingMode } from '..
 import { Cancellable, useCancellableTask } from '../utils/useCancellableTask';
 
 /**
+ * Props needed for {@link useVideoStreamLifecycleMaintainer}
+ * @private
+ */
+export interface VideoStreamLifecycleMaintainerProps {
+  isStreamAvailable?: boolean;
+  renderElementExists?: boolean;
+  isMirrored?: boolean;
+  scalingMode?: ScalingMode;
+  remoteParticipantId?: string;
+  onCreateLocalStreamView?: (options?: VideoStreamOptions) => Promise<void | CreateVideoStreamViewResult>;
+  onDisposeLocalStreamView?: () => void;
+  onCreateRemoteStreamView?: (
+    userId: string,
+    options?: VideoStreamOptions
+  ) => Promise<void | CreateVideoStreamViewResult>;
+  onDisposeRemoteStreamView?: (userId: string, options?: VideoStreamOptions) => Promise<void>;
+  isScreenSharingOn?: boolean;
+}
+
+/**
  * Helper hook to maintain the video stream lifecycle. This calls onCreateStreamView and onDisposeStreamView
  * appropriately based on react lifecycle events and prop changes.
  * This also handles calls to view.update* appropriately such as view.updateScalingMode().
  *
  * @private
  */
-export const useVideoStreamLifecycleMaintainer = (props: {
-  isStreamAvailable?: boolean;
-  renderElementExists?: boolean;
-  isMirrored?: boolean;
-  scalingMode?: ScalingMode;
-  onCreateStreamView?: (options?: VideoStreamOptions) => Promise<void | CreateVideoStreamViewResult>;
-  onDisposeStreamView?: () => void;
-}): void => {
-  const { onCreateStreamView, onDisposeStreamView, isStreamAvailable, renderElementExists, isMirrored, scalingMode } =
-    props;
+export const useVideoStreamLifecycleMaintainer = (props: VideoStreamLifecycleMaintainerProps): void => {
+  const {
+    isMirrored,
+    isScreenSharingOn,
+    isStreamAvailable,
+    onCreateLocalStreamView,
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
```diff
@@ -6,22 +6,45 @@ import { VideoStreamOptions, CreateVideoStreamViewResult, ScalingMode } from '..
...
+
+/**
+ * Helper hook to maintain the video stream lifecycle. This calls onCreateStreamView and onDisposeStreamView
+ * appropriately based on react lifecycle events and prop changes.
+ * This also handles calls to view.update* appropriately such as view.updateScalingMode().
+ *
+ * @private
+ */
+export const useVideoStreamLifecycleMaintainer = (props: VideoStreamLifecycleMaintainerProps): void => {
+  const {
```
prprabhu-ms, line 36: tangent: I don't favor this practice of destructuring large `props` structs. 

- It just adds lines of code that all readers ignore. The moment we ignore something while we read code, there's an opportunity for bugs to lurk.
- I think that writing `props.fieldName` isn't all that long, and in a complex component preferable because then you know that the value comes directly from the props (and not from a local variable).

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
```diff
@@ -6,22 +6,45 @@ import { VideoStreamOptions, CreateVideoStreamViewResult, ScalingMode } from '..
...
   isStreamAvailable?: boolean;
   renderElementExists?: boolean;
   isMirrored?: boolean;
   scalingMode?: ScalingMode;
-  onCreateStreamView?: (options?: VideoStreamOptions) => Promise<void | CreateVideoStreamViewResult>;
-  onDisposeStreamView?: () => void;
-}): void => {
-  const { onCreateStreamView, onDisposeStreamView, isStreamAvailable, renderElementExists, isMirrored, scalingMode } =
-    props;
+  remoteParticipantId?: string;
```
prprabhu-ms, line 17: Please add a comment here noting that local stream is created if this is undefined.

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
```diff
@@ -6,22 +6,45 @@ import { VideoStreamOptions, CreateVideoStreamViewResult, ScalingMode } from '..
...
+}
+
+/**
+ * Helper hook to maintain the video stream lifecycle. This calls onCreateStreamView and onDisposeStreamView
+ * appropriately based on react lifecycle events and prop changes.
+ * This also handles calls to view.update* appropriately such as view.updateScalingMode().
+ *
+ * @private
+ */
+export const useVideoStreamLifecycleMaintainer = (props: VideoStreamLifecycleMaintainerProps): void => {
```
prprabhu-ms, line 35: I'd recommend avoiding polymorphic behavior based on args.

So here, I'd recommend:
* Unexport `useVideoStreamLifecycleMaintainer`. We'll use it for the impls below.
* Add `useLocalVideoStreamLifecycleMaintainer`. This will call `useVideoStreamLifecycleMaintainer` with `onCreateStreamView: onCreateLocalStreamView` and `onDisposeStreamView: onDisposeLocalStreamView`.
* Add `useRemoteVideoStreamLifecyleMaintainer. This will also call `useVideoStreamLifecylceMaintainer`, but will need to use `useCallback` to bind the `remoteParticipantId` to the create/dispose view functions.

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
JamesBurnside: The issue I've found is with the useEffect, if you have a lot of props.<prop> in the dependency array react's eslint auto fix changes your dependency to simply be `props` instead of `props.*`

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
JamesBurnside: Refactor no longer requires this

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
JamesBurnside: Nice idea, refactored into useLocalVideoStreamLifecycleMaintainer and useRemoteVideoStreamLifecyleMaintainer

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
prprabhu-ms: TIL

prprabhu-ms (approved):

edwardlee-msft (approved):

PR #1911, opened 2022-05-19
Add perf counter to calling tests

PorterNan:
# What

Add perf counter to calling tests

- Move selector inner function outside to give it a function name so it could be visited as function.name in runtime
- Add a perf counter in UI Test
- Add a CI action to generate perf counter snapshot

# Why

# How Tested

# Process & policy checklist


PorterNan:
Close this since flakiness in calling is still not resolved, will revisit when we solve the flakiness problem in calling, could be valuable to enable this for chat first

PR #1914, opened 2022-05-19
Add PSTN calling handlers for hold, resume, and addParticipant

dmceachernmsft:
# What

Adds handlers to the calling component bindings and adapters for:
- holding a call
- resuming a call
- adding a participant
# Why

- starts to setup PSTN functionality in the repo

https://skype.visualstudio.com/SPOOL/_workitems/edit/2865531
# How Tested

Built to make sure stable API hasn't changed and ran samples to make sure all is well.


edwardlee-msft (approved):

common/scripts/babel-conditional-preprocess.js
```diff
@@ -81,6 +81,10 @@ exports.default = babelHelper.declare((_api, opts) => {
         Handle(path, featureSet, stabilizedFeatureSet);
       },
 
+      TSMethodSignature(path) {
```
anjulgarg, line 84: @dmceachernmsft Can we do this as a separate PR?
With explanation to why we need to do this?
And possibly an example in the conditional compilation examples doc?

packages/calling-component-bindings/review/beta/calling-component-bindings.api.md
```diff
@@ -80,9 +82,12 @@ export type CallingHandlers = {
     onStopScreenShare: () => Promise<void>;
     onToggleScreenShare: () => Promise<void>;
     onHangUp: () => Promise<void>;
+    onHold: () => Promise<void>;
```
anjulgarg, line 85: `onToggleHold` might be a better name since the same method will toggle between the hold states.

packages/calling-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -302,8 +318,31 @@ export const createDefaultCallingHandlers = memoizeOne(
...
+    const onAddParticipant = async (
+      participant:
+        | CommunicationUserIdentifier
+        | PhoneNumberIdentifier
+        | UnknownIdentifier
+        | MicrosoftTeamsUserIdentifier,
+      options?: AddPhoneNumberOptions
+    ): Promise<void> => {
+      const newACSorTeamsUser = participant as CommunicationUserIdentifier | MicrosoftTeamsUserIdentifier;
+      const newPSTNUser = participant as PhoneNumberIdentifier;
```
anjulgarg, line 331: Wouldn't `if (newPSTNUser)` always be true?

dmceachernmsft: if the casting to that type fails it will be undefined since the properties don't align for those types, so when passing in a participant only one of the casts will work.

packages/calling-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -302,8 +318,31 @@ export const createDefaultCallingHandlers = memoizeOne(
...
+      if (newPSTNUser) {
```
anjulgarg, line 332: I would simplify all this logic by doing this:
```tsx
if (participant.phoneNumber) {
  await call?.addParticipant(participant as PhoneNumberIdentifier, options);    
} else {
  await call?.addParticipant(participant as CommunicationUserIdentifier | MicrosoftTeamsUserIdentifier);
}
```

packages/calling-component-bindings/src/handlers/createHandlers.ts
anjulgarg: The casting is compile time only and won't fail during run time.

packages/calling-component-bindings/src/handlers/createHandlers.ts
anjulgarg: OR use the `isPhoneNumberIdentifier` function exported by the communication-common package.
```tsx
import { isPhoneNumberIdentifier } from "@azure/communication-common";

if (isPhoneNumberIdentifier(participant)) {
  await call?.addParticipant(participant as PhoneNumberIdentifier, options);
} else {
  await call?.addParticipant(
    participant as CommunicationUserIdentifier | MicrosoftTeamsUserIdentifier
  );
}
```

common/scripts/babel-conditional-preprocess.js
dmceachernmsft: @prprabhu-ms has a seperate PR for this as well I just included it so mine will build. its in the same spot so it wont conflict. (hopefully). https://github.com/Azure/communication-ui-library/pull/1913

packages/calling-component-bindings/src/handlers/createHandlers.ts
anjulgarg: Second option is better as it used the intended methods.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -415,6 +425,24 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
     this.handlers.onRemoveParticipant(userId);
   }
 
+  /* @conditional-compile-remove(PSTN-calls) */
+  public async addParticipant(
+    participant: PhoneNumberIdentifier | CommunicationUserIdentifier | UnknownIdentifier,
```
anjulgarg, line 430: Should we have the `MicrosoftTeamsUserIdentifier` here as well?

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -232,6 +240,29 @@ export interface CallAdapterCallManagement {
...
+  /* @conditional-compile-remove(PSTN-calls) */
+  /**
+   * Holds the call for the local user
+   * @beta
+   */
+  holdCall(): Promise<void>;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /**
+   * resumes the call for the local user.
+   * @beta
```
anjulgarg, line 263: Nit: r -> R

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -232,6 +240,29 @@ export interface CallAdapterCallManagement {
...
   removeParticipant(userId: string): Promise<void>;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /**
+   * Adds a participant to the call by dialing them in
+   * @param participant - Phone number or Communication userId for participant being added
+   * @param options - Call options containing ACS number for PSTN calls
+   * @beta
+   */
+  addParticipant(
+    participant: PhoneNumberIdentifier | CommunicationUserIdentifier | UnknownIdentifier | MicrosoftTeamsUserIdentifier,
```
anjulgarg, line 251: Instead of creating a union of this type everywhere. Can we use the `CommunicationIdentifier` exported by communication-common?

packages/calling-component-bindings/src/handlers/createHandlers.ts
dmceachernmsft: oh cool! will update this

common/scripts/babel-conditional-preprocess.js
prprabhu-ms: Merged mine. You can merge from `main` and get rid of this diff now.

Thanks for helping make conditional compilation exhaustive... slowly... incrementally.... some day. ;-)

PR #1917, opened 2022-05-20
Add declaritive pstn handlers

dmceachernmsft:
# What

Adds the following handlers to the Declarative Calling API 
# Why

Allows for these functions to be used with the calling stateful client.

Follows up on this issue https://skype.visualstudio.com/SPOOL/_workitems/edit/2865531
# How Tested

Built and ran samples to validate changes didn't break existing functionality, will test with PSTN POC to see if new functionality works.


change/@internal-calling-stateful-client-59b5cdca-4f3e-445e-b0be-ed4b1fcc3623.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Update Calling declarative so to handle new PSTN functions",
```
anjulgarg, line 3: Nit: extra 'so'

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -160,6 +160,28 @@ export class CallContext {
     });
   }
 
+  /* @conditional-compile-remove(PSTN-calls) */
+  public setCallHold(callId: string): void {
+    const latestCallId = this._callIdHistory.latestCallId(callId);
+    this.modifyState((draft: CallClientState) => {
+      const call = draft.calls[latestCallId];
+      if (call && call.state !== 'RemoteHold') {
```
anjulgarg, line 168: Why can't I put a call on `LocalHold` if it is on `RemoteHold`?

anjulgarg: Also, I would assume a call can't be put on hold if it is in transient states like `Ringing` `Connecting` etc.

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -160,6 +160,28 @@ export class CallContext {
...
+        this.setCallState(latestCallId, 'LocalHold');
+      }
+    });
+  }
+
+  /* @conditional-compile-remove(PSTN-calls) */
+  public setCallResume(callId: string): void {
```
anjulgarg, line 175: Where will these functions be consumed?

anjulgarg:
@dmceachernmsft I would recommend pointing this branch to `main` and pointing `CallingStatefulHandlersUpdatePSTN` to it.
Since this is an independent branch but `CallingStatefulHandlersUpdatePSTN` is dependent on it. :)

That allows us to merge it in to main and keep `CallingStatefulHandlersUpdatePSTN` minimal.

Also, I would recommend adding unit tests for the new context functions you have added in this branch.

packages/calling-stateful-client/src/CallContext.ts
alkwa-msft: does it seem to assume when we resume the call we are on "Connected"? Is it possible to somehow read the last known state of the call instead and to use that value?

packages/calling-stateful-client/src/CallContext.ts
dmceachernmsft: I will do some more digging into this but this I think is a behavior that would mirror when on a phone call between 2 phones. If someone puts me on hold I don't know that I should be able to do anything in the call other than hang up. 

Will update for the transient states. Was just thinking that the button wouldn't be present when the call is connecting but Contoso could have the control bar up when they are connecting to a call in their component app.

anjulgarg (approved):

carocao-msft (approved):

PR #1918, opened 2022-05-20
Added max height and scroll to file uploads as it was covering entire screen

prabhjot-msft:
# What
Added max height and scroll to file uploads as it was covering entire screen

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2818454

Before:
Screen Shot 2022-05-20 at 3 13 08 PM

After:
Screen Shot 2022-05-20 at 3 23 37 PM

# How Tested

# Process & policy checklist


packages/react-components/src/components/SendBox.tsx
```diff
@@ -264,6 +264,12 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
     [styles?.sendMessageIconContainer]
   );
 
+  const fileUploadCardsStyles = mergeStyles({
```
anjulgarg, line 267: Let's move this out of the component.

carocao-msft (approved):
LGTM after addressing Anjul's comment :)

packages/react-components/src/components/SendBox.tsx
prabhjot-msft: done

PR #1919, opened 2022-05-21
Add Hold button

dmceachernmsft:
# What

Adds hold button for PSTN Calls 
# Why

Adds A hold button to the UI Library for PSTN and 1:N Calling


packages/react-components/src/components/HoldButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const onRenderHoldIcon = (): JSX.Element => {
+    return <HighContrastAwareIcon disabled={props.disabled} iconName="HoldCall" />;
+  };
+  const onRenderResumeIcon = (): JSX.Element => {
+    return <HighContrastAwareIcon disabled={props.disabled} iconName="ResumeCall" />;
+  };
+
+  return (
+    <ControlBarButton
+      {...props}
```
alkwa-msft, line 70: do we also want to expose any styles here as well? 

ScreenShareButton has this

image

alkwa-msft (approved):

packages/calling-component-bindings/src/callControlSelectors.ts
```diff
@@ -148,3 +149,28 @@ export const devicesButtonSelector: DevicesButtonSelector = reselect.createSelec
...
+
+/* @conditional-compile-remove(PSTN-calls) */
+/**
+ * Selector for the {@link HoldButton} component.
+ * @public
+ */
+export const holdButtonSelector: HoldButtonSelector = reselect.createSelector([getCallState], (callState) => {
+  return {
+    checked: callState === 'LocalHold' ? true : false,
+    disabled: callState === 'RemoteHold' ? true : false
```
anjulgarg, line 174: Why can't I hold myself as well if the call is in RemoteHold?
Might be worth adding some documentation here

dmceachernmsft: is this something I should omit until we have a answer from calling what the expected behavior is? this isn't something that pertains to PSTN since the local user cannot tell when a PSTN user has them on hold, we never go to this state from what I've seen in my testing.

anjulgarg: If this doesn't pertain to PSTN, then basis on my interactions with the calling team so far.
Hold is basically a shortcut for disabling audio, video and screenshare stream of a participant.
And should be allowed even if the call is on remote hold. 
I am going to ask calling team for clarification around this anyway.

packages/calling-component-bindings/src/callControlSelectors.ts
```diff
@@ -148,3 +150,26 @@ export const devicesButtonSelector: DevicesButtonSelector = reselect.createSelec
...
+};
+
+/* @conditional-compile-remove(PSTN-calls) */
+/**
+ * Selector for the {@link HoldButton} component.
+ * @public
+ */
+export const holdButtonSelector: HoldButtonSelector = reselect.createSelector([getCallState], (callState) => {
+  return {
+    checked: callState === 'LocalHold' ? true : false
```
anjulgarg, line 173: `checked: callState === 'LocalHold'` is sufficient :)

packages/calling-component-bindings/src/hooks/usePropsFor.ts
```diff
@@ -115,6 +121,10 @@ export type GetSelector<Component extends (props: any) => JSX.Element | undefine
 export const getSelector = <Component extends (props: any) => JSX.Element | undefined>(
   component: Component
 ): GetSelector<Component> => {
+  /* @conditional-compile-remove(PSTN-calls) */
+  if (component === HoldButton) {
```
anjulgarg, line 125: we have an if condition here as well as a switch statement below it, both checking for HoldButton
Is one of these redundant?

packages/calling-component-bindings/src/hooks/usePropsFor.ts
```diff
@@ -136,8 +146,18 @@ const findSelector = (component: (props: any) => JSX.Element | undefined): any =
       return participantsButtonSelector;
     case EndCallButton:
       return emptySelector;
+    case HoldButton:
```
anjulgarg, line 149: Possible to have a conditionally remove return statement instead of using an explicit if block above?

dmceachernmsft: we can't conditionally compile out the use of the HoldButton import if we do so we need the extra check

anjulgarg (approved):

prabhjot-msft (approved):
Looks good to me!

PR #1921, opened 2022-05-24
Fix stringify circular ref

PorterNan:
# What

This is a bug fix for https://skype.visualstudio.com/SPOOL/_workitems/edit/2822810

The idea is using [stringyfy replacer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter) to fix circular refs.
# Why

# How Tested

# Process & policy checklist


JamesBurnside (approved):

packages/acs-ui-common/src/safeStringify.ts
```diff
@@ -20,3 +20,20 @@ export const _safeJSONStringify = (
...
+// Log all visited refs to avoid circular ref
+const createSafeReplacer = (): (<T>(key: string, value: T) => T | string) => {
+  const visited = new Set();
+  return function replacer<T>(key: string, value: T): T | string {
+    if (typeof value !== 'object') {
+      return value;
+    }
+
+    if (visited.has(value)) {
+      return 'Visited-Ref';
```
prprabhu-ms, line 33: Would be nice if we remembered the path to each object in `visited: Record<Object: string>` and then returned something like `Already visited at ${visited.get(value)}` so that you can still know what the value was.

prprabhu-ms (approved):

packages/acs-ui-common/src/safeStringify.ts
PorterNan: Good idea, will try to get it work tmr

PorterNan: Just realize we don't have a good way to access current path... We might need to find a way to manipulate some keys+sub-keys+more to get it work, which might be tricky since we don't know which order stringify is visiting the node(Dfs or BFS)
I believe there are solutions, but a little overkill to this, since peple can guess the value from current key(value must be in ancestors)

PR #1923, opened 2022-05-25
Fix chat container 

dmceachernmsft:
# What

Fix type mismatch in the chat container storybook props
# Why

image

# How Tested

Ran tests locally


carocao-msft (approved):

JamesBurnside:
Does JoinExistingChatThread need this also?
image

dmceachernmsft:
> Does JoinExistingChatThread need this also? image

It needed the opposite action, I must have gotten the two confused when I initially worked on this bug 😅

PR #1925, opened 2022-05-25
updating beta @azure/communication-calling to 1.5.4-beta.1

alkwa-msft:
updating to latest beta calling version.

~We still have an issue where callClient.getEnvironmentInfo is being exposed in the stable API :(~


packages/calling-stateful-client/src/TypeAssertions.ts
```diff
@@ -28,6 +28,8 @@ type InferredCallErrorTargets =
    * stabilized in @azure/commmunication-calling.
    * TODO: Remove this hack once 'CallAgent.feature' becomes part of stable @azure/communication-calling.
    */
+  /* @conditional-compile-remove(calling-1.5.4-beta.1) */
+  | 'CallClient.getEnvironmentInfo'
```
alkwa-msft, line 32: I can't seem to remove this addition to the union type. Could use some help here. @PorterNan

alkwa-msft: able to remove it from stable but now the comment exists in beta :(

mgamis-msft (approved):

common/config/babel/.babelrc.js
```diff
@@ -10,7 +10,7 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
       features: [
         // Flag to add API only available in beta calling SDK to mocks and internal types.
         // This feature should be stabilized whenever calling SDK is stabilized.
-        'calling-1.4.3-beta.1',
+        'calling-1.5.4-beta.1',
```
PorterNan, line 13: It might be an outdate question, but @prprabhu-ms might help with answering this:
Why don't we call it 'calling-beta-sdk'? Are we intentionally need to modify all the comments each time when we update the sdk version?

packages/calling-stateful-client/src/TypeAssertions.ts
PorterNan: Did you resolve the issue in the end? Is it just comment left in the .md file or something else?

common/config/babel/.babelrc.js
PorterNan: And all the api files need to be updated just because of those comments are included - which makes reviewing update beta much more difficult

PorterNan (approved):

common/config/babel/.babelrc.js
alkwa-msft: I could change it to that. I think at the time we wanted to be as specific as possible.

packages/calling-stateful-client/src/TypeAssertions.ts
alkwa-msft: Its only the comment left in the .md file. I am not sure how to remove this one.

JamesBurnside (approved):

PR #1926, opened 2022-05-25
Rework `GetSelector` type inference

prprabhu-ms:
# What

Rework how type inference works for `GetSelector`. The final inferred type is unchanged, just how the type inference helpers arrive at the type has changed.

# Why

This will allow us to add a new selector to `GetSelector` via conditional compilation.

See https://github.com/Azure/communication-ui-library/pull/1927 for an example of how a new selector can be added.

# How Tested

Added new type assertions to get build time coverage.

Confirmed that this is not considered a breaking API change.


packages/calling-component-bindings/review/beta/calling-component-bindings.api.md
```diff
@@ -197,7 +207,7 @@ export const useCallClient: () => StatefulCallClient;
...
...ector : undefined)> extends (props: any) => any ? ReturnType<FilterUndefined<(AreEqual<Component, (props: VideoGalleryProps) => JSX.Element> extends true ? VideoGallerySelector : undefined) | (AreEqual<Component, (props: DevicesButtonProps) => JSX.Element> extends true ? DevicesButtonSelector : undefined) | (AreEqual<Component, (props: MicrophoneButtonProps) => JSX.Element> extends true ? MicrophoneButtonSelector : undefined) | (AreEqual<Component, (props: CameraButtonProps) => JSX.Element> extends true ? CameraButtonSelector : undefined) | (AreEqual<Component, (props: ScreenShareButtonProps) => JSX.Element> extends true ? ScreenShareButtonSelector : undefined) | (AreEqual<Component, (props: ParticipantListProps) => JSX.Element> extends true ? ParticipantListSelector : undefined) | (AreEqual<Component, (props: ParticipantsButtonProps) => JSX.Element> extends true ? ParticipantsButtonSelector : undefined) | (AreEqual<Component, (props: EndCallButtonProps) => JSX.Element> extends true ? EmptySelector : undefined) | (AreEqual<Component, (props: ErrorBarProps) => JSX.Element> extends true ? ErrorBarSelector : undefined)>> & Common<CallingHandlers, Parameters<Component>[0]> : undefined;
```
prprabhu-ms, line 210: This seems to be a quirk of api-extractor that inlines type definitions that are simple just aliases, but not those that contain more complex logic.

So in this case, because `GetSelector` is defined as `FilterUndefined<....>`, api-extractor inlines the definition in several places.

packages/calling-component-bindings/src/hooks/TypeAssertions.tsx
```diff
@@ -0,0 +1,64 @@
...
+} from '../callControlSelectors';
+import { ParticipantListSelector } from '../participantListSelector';
+import { ParticipantsButtonSelector } from '../participantsButtonSelector';
+import { ErrorBarSelector } from '../errorBarSelector';
+
+/**
+ * This function is a compile type check that {@link GetSelector} returns
+ * values of the correct type.
+ *
+ * @private
```
prprabhu-ms, line 32: These type assertions were added before my changes for how `GetSelector` works and they passed through the refactor. So we're confident that the final types are unchanged.

dmceachernmsft: how are these functions to be used? is this just for testing?

prprabhu-ms: These are internal type assertions. Making sure that we get the expected concrete types in the end.

packages/calling-component-bindings/src/hooks/usePropsFor.ts
```diff
@@ -80,28 +80,17 @@ const emptySelector: EmptySelector = (): Record<string, never> => ({});
...
-  ? ParticipantListSelector
-  : AreEqual<Component, typeof ParticipantsButton> extends true
-  ? ParticipantsButtonSelector
-  : AreEqual<Component, typeof EndCallButton> extends true
-  ? EmptySelector
-  : AreEqual<Component, typeof ErrorBar> extends true
-  ? ErrorBarSelector
-  : undefined;
+export type GetSelector<Component extends (props: any) => JSX.Element | undefined> = FilterUndefined<
+  | (AreEqual<Component, typeof VideoGallery> extends true ? VideoGallerySelector : undefined)
```
PorterNan, line 84: There could be a very small risk here that, if one selector is the super set of another (I guess EndCallButton could be one of them), there will be 2 selectors chosen in the end and this will be selectorA | selectorB

packages/chat-component-bindings/src/hooks/TypeAssertions.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { getSelector } from './usePropsFor';
+import React from 'react';
+
+/**
+ * This function is a compile type check that {@link GetSelector} returns
+ * values of the correct type.
+ *
+ * @private
+ */
+export function assertGetSelectorTypes(): unknown {
```
PorterNan, line 19: I love this one!!
Maybe we could type the return value a little bit more like
export function assertGetSelectorTypes(): [chatParticipantListProps, errorBarProps, ....] to see if props and selectors get perfect matched

packages/chat-component-bindings/src/hooks/TypeAssertions.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const chatParticipantListSelector: ChatParticipantListSelector = getSelector(ParticipantList);
+  const errorBarSelector: ErrorBarSelector = getSelector(ErrorBar);
+  const messageThreadSelector: MessageThreadSelector = getSelector(MessageThread);
+  const sendBoxSelector: SendBoxSelector = getSelector(SendBox);
+  const typingIndicatorSelector: TypingIndicatorSelector = getSelector(TypingIndicator);
```
PorterNan, line 24: Do we need to manually assign types here? Like
const typingIndicatorSelector: TypingIndicatorSelector = getSelector(TypingIndicator);

Can we do 
const typingIndicatorSelector= getSelector(TypingIndicator);

the return value should be inferenced correctly, then we do the final comparation in return value?

PorterNan: The concern here is sometimes, if getSelector return 'any' type by some mistakes, it will be still inferenced correctly if we do the type assignment at the very beginning

packages/calling-component-bindings/src/hooks/usePropsFor.ts
prprabhu-ms: Hence the TypeAssertions. This isn't happening in practice... yet.
If this happens, we might have to find a way to "pick the first one".

packages/chat-component-bindings/src/hooks/TypeAssertions.tsx
prprabhu-ms: Type inference via return type and via local variable should be identical, no?

Note that I'm _not_ casting the return type, this is just checking the inferred type.

packages/chat-component-bindings/src/hooks/TypeAssertions.tsx
prprabhu-ms: I think checking via return types and via local variable is identical ( see below ).

Using local variables keeps the type and the selector call close together.

packages/chat-component-bindings/src/hooks/TypeAssertions.tsx
PorterNan: If the type is 'any' it will pass the type check after you cast it to anything like
const foo: any = ...;
const bar: AType = foo;

This will still work

dmceachernmsft: So the issue then is even though we are doing the type assertions with the big union type above, getSelector can still return a 'any' type?

packages/calling-component-bindings/src/hooks/usePropsFor.ts
prprabhu-ms: Talked IRL and double checked that our type assertions cover this case.

packages/chat-component-bindings/src/hooks/TypeAssertions.tsx
prprabhu-ms: talked IRL and determined that the existing type assertions are enough.

packages/calling-component-bindings/src/hooks/TypeAssertions.tsx
```diff
@@ -0,0 +1,70 @@
...
+export function assertGetSelectorTypes(): unknown {
+  // In case one the component's selector is a strict sub-type of another,
+  // `getSelector` might return a type union with two selector types.
+  // The following reverse type assertions ensure that we catch that case at build time.
+  //
+  // If the following assertions fail, we need to find a way to disambiguate between the
+  // selectors of those two components.
+  const cameraButtonSelector: CameraButtonSelector = getSelector(CameraButton);
+  const devicesButtonSelector: DevicesButtonSelector = getSelector(DevicesButton);
+  let endCallButtonSelector: EmptySelector = getSelector(EndCallButton);
```
dmceachernmsft, line 43: why were these changed to `let` statements?

prprabhu-ms: Err leaked from a since removed change.

Thanks for catching that.

prprabhu-ms:
Not needed right now because we're unblocked by https://github.com/Azure/communication-ui-library/pull/1942

#1942 is a workaround for now. We _should_ solve this in a way that we don't require beta-only types.

But I prefer the approach in #1940 over this  one.

prprabhu-ms:
Decided to keep the current types and support conditional compilation for another AST node: https://github.com/Azure/communication-ui-library/pull/1967

PR #1927, opened 2022-05-25
[DO NOT REVIEW]: PoC: A conditionally added selector

prprabhu-ms:
This shows how a selector can be added conditionally to `GetSelector`.

This PR *removes* an existing selector, so it'd be a breaking change. Just showing what's possible...


prprabhu-ms:
Just for demonstration.

PR #1928, opened 2022-05-25
Media Gallery onFetchAvatarPersonaData perf fix

dmceachernmsft:
# What

Updates the `MediaGallery` to change the `displayName` for the remote participants by replacing the `displayName` instead of returning a new array of participants each time we update the `displayName` based on Contoso's provided callBack.
# Why

Fixes a performance issue raised with returning a new array of participants each time the callback is invoked.

# How Tested

Ran locally


PorterNan:
As we discussed offline, memoizeFnAll will be used in this PR to resolve the problem between perf and immutability

dmceachernmsft:
closing this until we reapproach the issue. comment saved in user story.

PR #1929, opened 2022-05-25
Participant list selector and getRemoteParticipants selector PSTN updates

dmceachernmsft:
# What

Updates Participant selector
# Why

These updates will allow the participant list to show the phone number of a PSTN user as their displayName. As well these changes allow the participant list to remove a participant from the call if they are a PSTN user. 

https://skype.visualstudio.com/SPOOL/_workitems/edit/2865533
# How Tested

Built locally to make sure the PSTN changes didn't break any of the tests. will package and run in the PSTN POC app for validation.


packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -31,14 +31,17 @@ const convertRemoteParticipantsToParticipantListParticipants = (
...
+          displayName:
+            participant.identifier.kind !== 'phoneNumber'
+              ? participant.displayName
+              : parsePhoneNumberForDisplayName(participant.identifier.phoneNumber),
           state: participant.state,
           isMuted: participant.isMuted,
           isScreenSharing: isScreenSharing,
           isSpeaking: participant.isSpeaking,
           // ACS users can not remove Teams users.
           // Removing phone numbers or unknown types of users is undefined.
```
JamesBurnside, line 43: update this comment is we can now remove phone numbers

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -31,14 +31,17 @@ const convertRemoteParticipantsToParticipantListParticipants = (
 
         return {
           userId: toFlatCommunicationIdentifier(participant.identifier),
-          displayName: participant.displayName,
+          displayName:
```
JamesBurnside, line 34: displayName is used in lots of selectors, is it only the participant list we will want this? Otherwise I would suggest doing this in the baseSelector's getDisplayName

dmceachernmsft: will do!

carocao-msft (approved):

packages/calling-component-bindings/src/callUtils.ts
```diff
@@ -44,3 +46,38 @@ export const disposeAllLocalPreviewViews = async (callClient: StatefulCallClient
...
+  let newParticipants: RemoteParticipantState[] = [];
+  if (participants) {
+    newParticipants = memoizedUpdateDisplayName((memoizedFn) => {
+      return Object.values(participants).map((p) => {
+        const pid = toFlatCommunicationIdentifier(p.identifier);
+        return memoizedFn(pid, p);
+      });
+    });
+  }
+  return newParticipants;
```
JamesBurnside, line 71: This is going to return a new array each time (the contents of the array has been memoized but not the array itself)

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -45,11 +46,9 @@ export const getDominantSpeakers = (
 export const getRemoteParticipants = (
   state: CallClientState,
   props: CallingBaseSelectorProps
-):
-  | undefined
-  | {
-      [keys: string]: RemoteParticipantState;
-    } => state.calls[props.callId]?.remoteParticipants;
+): undefined | RemoteParticipantState[] => {
+  return updateUserDisplayNames(state.calls[props.callId]?.remoteParticipants);
```
JamesBurnside, line 50: cc. @prprabhu-ms - you mentioned not wanting to augment the base selectors -- should this be moved to a secondary selector `getRemoteParticipantsWithUpdatedDisplayNames`?

JamesBurnside: (reason for it to be a base selector/its own selector - we want everyone getting a displayName to get a consistent display name across all components)

dmceachernmsft: I mean it would be a bigger diff but we could just have this pipe into a new base selector if the concern interacting with whats coming out of the call state at this point?

prprabhu-ms: Base selectors should stay simple. They _only_ figure out what part of `state` to return and then return it. No modifications. Reasons.

* Clarity in what a base selector is.
* A base selector runs on *every* React render pass. This means that any functions you call here are called on every render pass (even if they memoize their returned values, they will be called each time)
* The most insidious: You have to be very careful to memoize the returned values yourself. Other selectors don't run if their inputs don't change. So if you fail to memoize return values, it means that they have a snowball effect on what gets updated upstream, *but only when their inputs change*. For base selectors that is on every single render pass. So the cost of failure to memoize (James pointed out an example) is much higher.


As an aside, business logic "use phone number as display name" is likely better kept closer to the UI layer (i.e., each components' top-level-ish selector should make this determination.

packages/calling-component-bindings/src/videoGallerySelector.ts
```diff
@@ -94,8 +95,16 @@ export const videoGallerySelector: VideoGallerySelector = createSelector(
           renderElement: localVideoStream?.view?.target
         }
       },
-      remoteParticipants: _videoGalleryRemoteParticipantsMemo(remoteParticipants),
+      remoteParticipants: _videoGalleryRemoteParticipantsMemo(
+        updateUserDisplayNamesTrampoline(remoteParticipants ? Object.values(remoteParticipants) : [])
```
JamesBurnside, line 99: nit: if this returns `[]` it will return a new array instance each time so isn't memozie - consider making a global const `noRemoteParticipants = []` to have a static ref object

JamesBurnside (approved):

JamesBurnside:
There's been a lot of changes since @carocao-msft approved, @carocao-msft can you give this another look?

packages/calling-component-bindings/src/videoGallerySelector.ts
dmceachernmsft: right will update this!

carocao-msft (approved):

PR #1930, opened 2022-05-25
Use `<LocalVideoTile/>` and `<RemoteVideoTile/>` in the CallWithChat PiPiP

JamesBurnside:
# What
* Update _PictureInPictureInPicture to take in ReactChildren instead of VideoTileProps to support reusing tiles inside this component
* Export LocalVideoTile and RemoteVideoTile to use in the LocalAndRemotePiP
* Use LocalVideoTile and RemoteVideoTile in the LocalAndRemotePiP
* Add some props like showLabel and minPersonaSize to the LocalVideoTile and RemoteVideoTile to allow these to be set in the underlying VideoTile

# Why
Remove custom logic and custom handling of streams and instead reuse the logic from a unified place. This also ensures LocalAndRemotePiP video tiles support the new updateScaleMode.

# How Tested
Lots: joining calls with different folks videos on and off and changing dominant speaker while one user was in the callwithchat chat screen to see the pipip,


packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
```diff
@@ -6,7 +6,8 @@ import { submitWithKeyboard } from '../utils/keyboardNavigation';
 import {
   PictureInPictureInPicturePrimaryTile,
   PictureInPictureInPictureSecondaryTile,
-  _PictureInPictureInPictureTileProps
+  _PictureInPictureInPictureTileProps,
+  _TileOrientation
```
prprabhu-ms, line 10: Unused import? Doesn't `eslint` complain about this?

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPictureTile.tsx
```diff
@@ -1,40 +1,38 @@
...
 export type _TileOrientation = 'portrait' | 'landscape';
 
 /** @internal */
-export interface _PictureInPictureInPictureTileProps
-  extends Pick<
-    VideoTileProps,
-    'styles' | 'displayName' | 'renderElement' | 'isMirrored' | 'noVideoAvailableAriaLabel'
-  > {
+export interface _PictureInPictureInPictureTileProps {
+  content: ReactChild;
```
prprabhu-ms, line 13: This can just be `children` right? We expect exactly one child and we simply forward it to the `Stack` in the implementation.

packages/react-composites/src/composites/CallComposite/components/LocalAndRemotePIP.tsx
```diff
@@ -93,14 +101,39 @@ export const LocalAndRemotePIP = (props: LocalAndRemotePIPProps): JSX.Element =>
...
+
   return (
     <_PictureInPictureInPicture
       onClick={props.onClick}
       strings={strings}
-      // If there are no remote participants, show the local participant as the primary tile
-      primaryTile={remoteVideoTile ?? localVideoTile}
-      // If we are showing the local participant as the primary tile, show nothing for the secondary tile
-      secondaryTile={remoteVideoTile ? localVideoTile : undefined}
+      primaryTile={primaryTileProps}
```
prprabhu-ms, line 135: Consider an alternative API:

```
<_PictureInPictureInPicture
  onClick={...}
  strings={...}
  onRenderPrimaryTile={remoteVideoTileProps ? <_RemoteVideoTile {...remoteVideoTileProps} /> : <_LocalVideoTile {...localVideoTileProps} />}
  primaryTileOrientation='portrait'
  onRenderSecondaryTileProps={remoteVideoTileProps ? <_LocalVideoTile {...localVideoTileProps} /> : undefined}
  secondaryTileOrientation='portrait'
>
```

I _think_ this avoids the rendering issues we observed yesterday because the `onRender*` functions directly use a component.

Benefits of this API: No need to memoize `localVideoTileProps` and `remoteVideoTileProps` because they are all destructured

Drawbacks of this API: Namespacing of other props like `primaryTileOrientation` etc. This is a common React problem. You need to flatten objects out to avoid the need to memoize.

Just a soft suggestion.

prprabhu-ms (approved):

packages/react-composites/src/composites/CallComposite/components/LocalAndRemotePIP.tsx
JamesBurnside: I had this at one point but reverted because `secondaryTileOrientation` becomes moot if `onRenderSecondaryTileProps` is not supplied. Which I realize we do a lot in other places but it feels poor. If we add more tile props then they continue needing to be flat and we have a growing number of potentially redundant props
The way it is currently avoids the rendering issue - but does move the `<RemoteVideoTile>` outside of the return which isn't ideal - gonna leave for now but its all internal so can be easily updated in future

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPicture.tsx
JamesBurnside: `eslint` likes me and decided to not pull me up on that one (I have no idea why its not complaining about it haha)

packages/react-components/src/components/PictureInPictureInPicture/PictureInPictureInPictureTile.tsx
JamesBurnside: I haven't really understood the 'children' naming when it has to be a single child. There is a react.children type but we don't use it anywhere (we typically use react.reactnode for children)

prprabhu-ms: `props.children` is perfect for cases where all your component does is forward the children as children to another component (in this case a `Stack`).

https://reactjs.org/docs/glossary.html#propschildren

It does mean that your Component can not specify there should be only one child. But does that matter for this component? If multiple children are provided, the `Stack` will lay them out.

The big benefit in using `props.children` is how React handles children during reconciliation. A change in a prop for `_PictureInPictureInPictureTileProps` does not cause the childrens' `render()` method getting called. If you pass it in the child as a prop and render it inside, then it will.

dmceachernmsft (approved):

PR #1931, opened 2022-05-25
[DraftPR][Do not review] Custom Button Injection into Call and CallWithChat composite

edwardlee-msft:
# What
A Draft PR for Custom Button Injection API using Proposal 4 from [Link](https://microsoft-my.sharepoint.com/:w:/p/edwardlee/EQtMhJj_7dVAm7zOtUwzI7cB37l1kmrKhUVYKZnJ7vYxYA?e=DqLWSw)
Feature definition can be found here [Link](https://microsoft.sharepoint-df.com/:w:/t/IC3SDK/EQDJoDZ4_lhBo6ze9KGzV9ABN41uzbQhIZmmYNfd-2HlLg?e=wHDw5B)

# Why
Allow Customers to inject Custom buttons into Call and callWithChat composite. 

Create an array of custom buttons:

Pass in custom buttons as options into the CallScreen:

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/components/buttons/Custom.tsx
+++ b/packages/react-composites/src/composites/CallComposite/components/buttons/Custom.tsx
@@ -18,14 +18,9 @@ export const generateCustomButtons = (
   displayType?: CallControlDisplayType
 ): CustomButtons => {
   const response = {
-    first: undefined,
-    afterCameraButton: undefined,
-    afterEndCallButton: undefined,
-    afterMicrophoneButton: undefined,
-    afterDevicesButton: undefined,
-    afterParticipantsButton: undefined,
-    afterScreenShareButton: undefined,
-    last: undefined
+    sideBar: undefined,
+    mainBar: undefined,
+    overflowBar: undefined
   };
   if (!onFetchCustomButtonProps) {
     return response;
@@ -40,6 +35,33 @@ export const generateCustomButtons = (
           .map((buttonProps, i) => (
             <ControlBarButton {...buttonProps} key={`${buttonProps.placement}_${i}`} />
           ))}
+      </>
+    );
+  }
+  return response;
+};
+
+/** @private */
+export const generateCustomDrawerButtons = (
+  onFetchCustomButtonProps?: CustomCallControlButtonCallback[],
+  displayType?: CallControlDisplayType
+): CustomButtons => {
+  const response = {
+    sideBar: undefined,
+    mainBar: undefined,
+    overflowBar: undefined
+  };
+  if (!onFetchCustomButtonProps) {
+    return response;
+  }
+
+  const allButtonProps = onFetchCustomButtonProps.map((f) => f({ displayType }));
+  for (const key in response) {
+    response[key] = (
+      <>
+        {allButtonProps
+          .filter((buttonProps) => buttonProps.placement === key)
+          .map((buttonProps, i) => ({ ...buttonProps, key: `${buttonProps.placement}_${i}` }))}
       </>
     );
   }
--- a/packages/react-composites/src/composites/CallComposite/types/CallControlOptions.ts
+++ b/packages/react-composites/src/composites/CallComposite/types/CallControlOptions.ts
@@ -81,15 +81,7 @@ export type CallControlOptions = {
  *
  * @beta
  */
-export type CustomCallControlButtonPlacement =
-  | 'first'
-  | 'last'
-  | 'afterCameraButton'
-  | 'afterEndCallButton'
-  | 'afterMicrophoneButton'
-  | 'afterDevicesButton'
-  | 'afterParticipantsButton'
-  | 'afterScreenShareButton';
+export type CustomCallControlButtonPlacement = 'sideBar' | 'mainBar' | 'overflowBar';
 
 /**
  * A callback that returns the props to render a custom {@link ControlBarButton}.
--- a/packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
+++ b/packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
@@ -4,6 +4,8 @@
 import React, { useMemo } from 'react';
 import { CallAdapterProvider } from '../CallComposite/adapter/CallAdapterProvider';
 import { CallAdapter } from '../CallComposite';
+/* @conditional-compile-remove(control-bar-button-injection) */
+import { CustomCallControlButtonCallback } from '../CallComposite';
 import { PeopleButton } from './PeopleButton';
 import { concatStyleSets, IStyle, ITheme, mergeStyles, Stack, useTheme } from '@fluentui/react';
 import { controlBarContainerStyles } from '../CallComposite/styles/CallControls.styles';
@@ -20,6 +22,8 @@ import { EndCall } from '../CallComposite/components/buttons/EndCall';
 import { MoreButton } from './MoreButton';
 import { CallWithChatControlOptions } from './CallWithChatComposite';
 import { ContainerRectProps } from '../common/ContainerRectProps';
+/* @conditional-compile-remove(control-bar-button-injection) */
+import { generateCustomButtons } from '../CallComposite/components/buttons/Custom';
 
 /**
  * @private
@@ -105,6 +109,15 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
     () => (!props.mobileView ? getDesktopEndCallButtonStyles(theme) : undefined),
     [props.mobileView, theme]
   );
+  /* @conditional-compile-remove(control-bar-button-injection) */
+  const customButtons = useMemo(
+    () =>
+      generateCustomButtons(
+        onFetchCustomButtonPropsTrampoline(options != false ? options : undefined),
+        options !== false ? options?.displayType : undefined
+      ),
+    [options]
+  );
 
   // when options is false then we want to hide the whole control bar.
   if (options === false) {
@@ -161,6 +174,20 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
                     styles={screenShareButtonStyles}
                   />
                 )}
+                {
+                  /* @conditional-compile-remove(control-bar-button-injection) */
+                  props.mobileView
+                    ? customButtons['mainBar']?.props.children.slice(0, 1)
+                    : customButtons['mainBar']?.props.children.slice(0, 3).map((element) => {
+                        return (
+                          <element.type
+                            {...element.props}
+                            styles={commonButtonStyles}
+                            displayType={options.displayType}
+                          />
+                        );
+                      })
+                }
                 {props.mobileView && (
                   <MoreButton
                     data-ui-id="call-with-chat-composite-more-button"
@@ -176,6 +203,12 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
       </Stack.Item>
       {!props.mobileView && (
         <Stack horizontal className={!props.mobileView ? mergeStyles(desktopButtonContainerStyle) : undefined}>
+          {
+            /* @conditional-compile-remove(control-bar-button-injection) */
+            customButtons['sideBar']?.props.children.slice(0, 2).map((element) => {
+              return <element.type {...element.props} styles={commonButtonStyles} />;
+            })
+          }
           {isEnabled(options?.peopleButton) && (
             <PeopleButton
               checked={props.peopleButtonChecked}
@@ -275,3 +308,13 @@ const getDesktopEndCallButtonStyles = (theme: ITheme): ControlBarButtonStyles =>
 };
 
 const isEnabled = (option: unknown): boolean => option !== false;
+
+/* @conditional-compile-remove(control-bar-button-injection) */
+const onFetchCustomButtonPropsTrampoline = (
+  options?: CallWithChatControlOptions
+): CustomCallControlButtonCallback[] | undefined => {
+  let response: CustomCallControlButtonCallback[] | undefined = undefined;
+  /* @conditional-compile-remove(control-bar-button-injection) */
+  response = options?.onFetchCustomButtonProps;
+  return response;
+};
--- a/packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
+++ b/packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
@@ -2,13 +2,20 @@
 // Licensed under the MIT license.
 
 import React, { useCallback } from 'react';
+/* @conditional-compile-remove(control-bar-button-injection) */
+import { useMemo } from 'react';
 import {
   OptionsDevice,
   _DrawerMenu as DrawerMenu,
-  _DrawerMenuItemProps as DrawerMenuItemProps
+  _DrawerMenuItemProps as DrawerMenuItemProps,
+  _DrawerMenuItemProps
 } from '@internal/react-components';
 import { AudioDeviceInfo } from '@azure/communication-calling';
 import { CallWithChatControlOptions } from '../CallWithChatComposite';
+/* @conditional-compile-remove(control-bar-button-injection) */
+import { generateCustomDrawerButtons } from '../../CallComposite/components/buttons/Custom';
+/* @conditional-compile-remove(control-bar-button-injection) */
+import { CustomCallControlButtonCallback } from '../../CallComposite';
 
 /** @private */
 export interface MoreDrawerStrings {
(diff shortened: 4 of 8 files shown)


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1363,7 +1365,7 @@ export interface CustomCallControlButtonCallbackArgs {
 }
 
 // @beta
-export type CustomCallControlButtonPlacement = 'first' | 'last' | 'afterCameraButton' | 'afterEndCallButton' | 'afterMicrophoneButton' | 'afterDevicesButton' | 'afterParticipantsButton' | 'afterScreenShareButton';
+export type CustomCallControlButtonPlacement = 'mainBar' | 'sideBar' | 'overflowMenu';
```
JamesBurnside, line 1368: I think `Bar` is confusing here, `sideBar` would imply inside the panel that appears from the side but buttons placed here could be independant of the side bar.
Perhaps `primary | secondary | primaryOverflow` ? (maybe primaryControls or primarySpace .. very open to suggestions)

packages/react-composites/src/composites/CallComposite/types/CallControlOptions.ts
```diff
@@ -66,30 +66,19 @@ export type CallControlOptions = {
...
 /**
- * Placement for a custom button injected in the {@link CallControls}.
+ * Placement for a custom button injected in the {@link CallControls} and {@link CallWithChatControlBar}.
  *
- * 'first': Place the button on the left end (right end in rtl mode).
- * 'afterCameraButton': Place the button on the right (left in rtl mode) of the camera button.
- * ... and so on.
+ * 'mainBar': Place the button(s) on the right end of the center control bar but before the EndCallButton (left end in rtl mode).
+ * 'sideBar': Place the button(s) on the left end of the side control bar (right in rtl mode).
+ * 'overflowMenu': Place the buttons(s) on the end of the overflow Menu {@link MoreDrawer}
```
JamesBurnside, line 73: Note `MoreDrawer` isn't public here so this `@link` won't work for external devs

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -602,6 +602,8 @@ export interface CallWithChatControlOptions {
     displayType?: CallControlDisplayType;
     endCallButton?: boolean;
     microphoneButton?: boolean;
+    // @beta
+    onFetchCustomButtonProps?: CustomCallControlButtonCallback[];
```
JamesBurnside, line 606: I see `CustomCallControlButtonCallback` exposes the whole `ControlBarButtonProps` - I just want to confirm if we want to expose this whole interface or something more restrictive

JamesBurnside: Actually `ControlBarButtonProps`'s interface doesn't quite work for the overflowMenu item - we may need to restrict it or have something like `(CustomCallControlButtonCallback | CustomCallControlOverflowButtonCallback)[]`

packages/communication-react/review/beta/communication-react.api.md
edwardlee-msft: We could even go with:
`primaryControlBar | 'secondaryControlBar | 'overFlowControlBar`
or something a bit more basic
`primary | 'secondary | 'overflow`

edwardlee-msft: Decided to go with 
`primary | secondary | overflow`

Please comment if there is an issue with what I used :)

packages/react-composites/review/beta/react-composites.api.md
edwardlee-msft: This is a good point and would ultimately put the responsibility on Contoso to provide correct props for an overflow item. 
However, I've deliberately kept it as just customCallControlButtonCallback because I am automagically moving buttons into the overflow if certain conditions are met (max # of buttons reached, desktop to mobile view switched).
ControlBarButtonProps and DrawerMenuItemProps are similar enough that a DrawerMenuItem can be drawn from ControlBarButtonProps..

I am open to discuss how we would then automagically move buttons from the control bar to the overflow if we did not want to lean on converting ControlBarButtonProps and utilizing them as a DrawerMenuItemProps

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1363,7 +1365,7 @@ export interface CustomCallControlButtonCallbackArgs {
 }
 
 // @beta
-export type CustomCallControlButtonPlacement = 'first' | 'last' | 'afterCameraButton' | 'afterEndCallButton' | 'afterMicrophoneButton' | 'afterDevicesButton' | 'afterParticipantsButton' | 'afterScreenShareButton';
+export type CustomCallControlButtonPlacement = 'primary' | 'overflow' | 'secondary';
```
JamesBurnside, line 1368: `CallComposite` doesn't have a secondary at the moment, will using secondary do nothing or do we plan to allow call composite to have a secondary section (that is empty currently?)
Otherwise we should have two types: one for callcomposite and one for callwithchatcomposite

packages/react-composites/review/beta/react-composites.api.md
edwardlee-msft: Thats a great point! With that we would still be able to extract and render a ControlBarButton and a DrawerMenuItem. 
Thanks James!

packages/communication-react/review/beta/communication-react.api.md
edwardlee-msft: My thinking was to provide CallComposite with a secondary (but leave empty/do nothing for now), which readies us the with the option to add a secondary section in the future for CallComposite.

I am fine with splitting it into two types, but thought it would scale better in this case.

prprabhu-ms:
Quick review latency note. I'm a bit overloaded today, so won't be able to do this one justice. Happy to take a look tomorrow or feel free to reassign.

packages/communication-react/review/beta/communication-react.api.md
dmceachernmsft: I wonder if there is value in exposing the secondary area now for the calling composite? allow Contoso to place their own buttons there with their own functionality even though we might not have anything there?

packages/react-composites/src/composites/CallComposite/components/buttons/Custom.tsx
```diff
@@ -9,6 +9,17 @@ import {
   CustomCallControlButtonPlacement
 } from '../../types/CallControlOptions';
 
+/**
+ * Max number of Custom Buttons in primary and secondary ControlBar
+ *
+ * @private
+ */
+export const CUSTOM_BUTTON_OPTIONS = {
+  MAX_PRIMARY_DESKTOP_CUSTOM_BUTTONS: 3,
```
dmceachernmsft, line 18: These values exclude the existing buttons in the control bar?

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -105,6 +109,15 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
...
+  const customButtons = useMemo(
+    () =>
+      generateCustomButtons(
+        onFetchCustomButtonPropsTrampoline(options !== false ? options : undefined),
+        options !== false ? options?.displayType : undefined
+      ),
+    [options]
+  );
 
   // when options is false then we want to hide the whole control bar.
```
dmceachernmsft, line 122: Do we want to maintain this logic for the custom buttons? Say for example Contoso wants to put their own entire control bar in replacing the one we provide. If they set options to `false` then neither set wont be visible even if they provide their own custom buttons.

packages/react-composites/src/composites/CallComposite/components/buttons/Custom.tsx
edwardlee-msft: Yes, they exclude the existing buttons in the control bar. 
I will make that explicit in the description.

packages/communication-react/review/beta/communication-react.api.md
edwardlee-msft: I am open to having separate areas for Calling and CallWithChat. However, it would involve more maintenance and would require adding separate custom button files for the two.

packages/react-composites/src/composites/CallComposite/components/buttons/Custom.tsx
```diff
@@ -1,47 +1,100 @@
...
           .filter((buttonProps) => buttonProps.placement === key)
           .map((buttonProps, i) => (
-            <ControlBarButton {...buttonProps} key={`${buttonProps.placement}_${i}`} />
+            <ControlBarButton
+              onClick={buttonProps.onItemClick}
+              showLabel={buttonProps.showLabel}
+              strings={generateCustomControlBarButtonStrings(buttonProps.text)}
+              styles={buttonProps.styles}
+              key={`${buttonProps.placement}_${i}`}
+              onRenderIcon={() => <Icon iconName={buttonProps.icon} />}
```
dmceachernmsft, line 59: should this be? `<CallCompositeIcon iconName={buttonProps.icon} />`

edwardlee-msft: Good catch, yes. This would be a good argument for separating custom.tsx into two files, one for Call Composite and one for CallWithChat Composite.

packages/react-composites/src/composites/CallComposite/types/CallControlOptions.ts
```diff
@@ -121,13 +110,37 @@ export interface CustomCallControlButtonCallbackArgs {
...
    */
   placement: CustomCallControlButtonPlacement;
+  /**
+   * Whether the button is disabled
+   */
+  disabled?: boolean;
+  /**
+   * Icon to render
+   */
+  icon?: string;
```
prprabhu-ms, line 131: Should be `iconName`. Also should document how we use this name (looking at just this interface it is not clear how we'll get an icon from this string)

edwardlee-msft: Changed to iconName. Also changed how iconNames are provided. Icon must be of type CallWIthChatCompositeIcons or CallCompositeIcons.

packages/communication-react/review/beta/communication-react.api.md
edwardlee-msft: I separated the implementation between how Calling and CallWithChat custom buttons are injected.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
edwardlee-msft: Would love to have a deeper conversation about this with you through a short call.
If options is provided as false, how would they provide custom buttons?

edwardlee-msft: After discussing with Don. 
It makes sense to leave it as is. If Contoso choses to hide the ControlBar then it will also hide custom buttons. 
If Contoso would like to build out a ControlBar using only the CustomButtons then they will need to hide all other default buttons individually through the options.

dmceachernmsft (approved):

packages/react-composites/src/composites/CallWithChatComposite/CustomButton.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { ControlBarButton } from '@internal/react-components';
+import React from 'react';
+import { generateCustomControlBarButtonStrings } from '../CallComposite/components/buttons/Custom';
+import {
+  CallControlDisplayType,
+  CustomCallControlButtonCallbackArgs,
+  CustomControlButtonProps
+} from '../CallComposite/types/CallControlOptions';
+import { CallWithChatCompositeIcon, CallWithChatCompositeIcons } from '../common/icons';
+/* @conditional-compile-remove(control-bar-button-injection) */
```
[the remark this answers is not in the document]
edwardlee-msft: Yes, cause it's only used below in line 155
`export const onFetchCustomButtonPropsTrampoline`
with onFetchCustomButtonpropsTrampoline being conditionally compiled.

If line 12 is not conditionally compiled then it causes a lint error as it's not being used otherwise.

change/@internal-react-composites-2bfa76c6-79a1-4af1-ba29-dfa1a0e23c01.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
JamesBurnside, line 2: should be type prerelease

change/@internal-react-composites-2bfa76c6-79a1-4af1-ba29-dfa1a0e23c01.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Add Custom button injection for CallWithChat and Call Composite",
```
JamesBurnside, line 3: Technically this already exists in beta for call composite. Can you mention the change we're making here to it

packages/react-composites/src/composites/CallWithChatComposite/CustomButton.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Max number of Custom Buttons in primary and secondary ControlBar
+ * Does not include existing buttons in the controlBar.
+ *
+ * @private
+ */
+export const CUSTOM_BUTTON_OPTIONS = {
+  MAX_PRIMARY_DESKTOP_CUSTOM_BUTTONS: 3,
+  MAX_PRIMARY_MOBILE_CUSTOM_BUTTONS: 1,
```
JamesBurnside, line 23: How do we decide on the cut off here? Only a very narrow device I don't think we have any room for a custom button

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
```diff
@@ -170,6 +179,30 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
...
+    [drawerSelectionOptions]
+  );
+
+  /* @conditional-compile-remove(control-bar-button-injection) */
+  customDrawerButtons['overflow']?.props.children.forEach((element) => {
+    drawerMenuItems.push(element);
+  });
+  /* @conditional-compile-remove(control-bar-button-injection) */
+  customDrawerButtons['primary']?.props.children
+    .slice(CUSTOM_BUTTON_OPTIONS.MAX_PRIMARY_MOBILE_CUSTOM_BUTTONS)
```
JamesBurnside, line 198: should this slice less if a user has hidden some buttons?

packages/react-composites/src/composites/CallWithChatComposite/CustomButton.tsx
JamesBurnside: I think I'm more in favor of not trying to be smart and letting contoso decide where they put the buttons - they can detect what device size they are running on and put buttons in the primary or overflow accordingly

change/@internal-react-composites-2bfa76c6-79a1-4af1-ba29-dfa1a0e23c01.json
edwardlee-msft: Good point! Revised to:
Rework how Custom buttons are injected in Call Composite and add Custom Button Injection into CallWithChat Composite

A bit long but explains the changes.

packages/react-composites/src/composites/CallWithChatComposite/CustomButton.tsx
edwardlee-msft: The MAX custom buttons for each platform/screen was provided by Alex P. 
But good point on "What happens when Contoso decides to hide buttons" in your other comment. 

I am okay with not limiting how many buttons get injected. However, this would complicate how we "automagically" move buttons into the overflow menu.

JamesBurnside: Yeah the automagically part is tough... e.g. Do we do it on container size or based on a fixed number? If we do a fixed number what if the user knows their site will only be on larger devices and know they have more screen real estate to make use of.. 
I won't block this PR on the automagical moving of buttons, we can always update later, but I think it's complicated enough with enough edge cases to warrant its own section in your proprosed solution 4 to call out how this should be implemented

edwardlee-msft: Container size vs fixed number, I can see the argument for both.

The idea with max custom buttons was "Guaranteeing to Contoso that X # of Custom Buttons would be injected depending on desktop or mobile". 

However, implementing based on container size would be ideal but would require quite a bit of change. I like this approach and will explore this avenue. Will also discuss further with Alex P. on design if this is to be the case. 
I will also add onto Proposal #4, automagically moving buttons warrants clarity and further outlining.

For now, should we remove max buttons? (this would remove automagically moving buttons into the Overflow from this PR)

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
edwardlee-msft: For now, the goal is to provide clear and simplistic boundaries on how many custom buttons are injectable. 
We can then further explore the responsiveness of this design once we receive customer feedback.

JamesBurnside (approved):

PR #1932, opened 2022-05-26
VideoGallery selector and remote participant type updates

dmceachernmsft:
# What

Update `VideoGalleryRemoteParticipant` ,`VideoGallerySelector` and supporting functions to return remote participant state and type.
# Why

This is so that the `VideoGallery` knows how to handle a PSTN user when they are joining and in either a 1:1 or group call with a ACS user.
image

https://skype.visualstudio.com/SPOOL/_workitems/edit/2865526
https://skype.visualstudio.com/SPOOL/_workitems/edit/2865527
# How Tested

Built locally, will validate the changes in the POC app.


change/@internal-react-components-3b1b621a-3d43-4490-9374-b067034d374d.json
```diff
@@ -0,0 +1,7 @@
+{
```
anjulgarg, line 1: There should be a changelog for calling-component-bindings as well

packages/react-components/src/types/VideoGalleryParticipant.ts
```diff
@@ -90,4 +90,10 @@ export interface VideoGalleryRemoteParticipant extends VideoGalleryParticipant {
   isSpeaking?: boolean;
   /** Video stream of shared screen */
   screenShareStream?: VideoGalleryStream;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** State that the participant is in */
```
anjulgarg, line 94: Let's add example states to the doc here just to make it more clear

packages/react-components/src/types/VideoGalleryParticipant.ts
```diff
@@ -90,4 +90,10 @@ export interface VideoGalleryRemoteParticipant extends VideoGalleryParticipant {
...
+  state?: string;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** the type of participant that it is*/
+  kind?: 'communicationUser' | 'phoneNumber' | 'microsoftTeamsUser' | 'unknown';
```
anjulgarg, line 98: Where are we planning to use this prop?

change/@internal-react-components-3b1b621a-3d43-4490-9374-b067034d374d.json
dmceachernmsft: I thought so as well but the tool didn't ask for one?

packages/react-components/src/types/VideoGalleryParticipant.ts
dmceachernmsft: the kind? this would be used either in the `VideoGallery` itself to give a PSTN participant tile or in the `RemoteVideoTile` to change the UI based on what kind of user it is.

change/@internal-react-components-3b1b621a-3d43-4490-9374-b067034d374d.json
JamesBurnside: Stacked PR (where the base PR also has one). The rush changelog compares against main and not an arbitrary base branch, so once the base PR goes in it will complain you don't have one

packages/react-components/src/types/VideoGalleryParticipant.ts
JamesBurnside: +1 to adding more doc comments here. `state` is a really vague term, state of what? e.g. it could randomly mean there video in a black and white state, or isSpeaking variable above could easily be considered a piece of "state" pertaining to this VideoGalleryRemoteParticipant. I can't think of a better name - perhaps connectedState? Otherwise provide a good doc comment on what state refers to here

packages/react-components/src/types/VideoGalleryParticipant.ts
```diff
@@ -90,4 +90,10 @@ export interface VideoGalleryRemoteParticipant extends VideoGalleryParticipant {
   isSpeaking?: boolean;
   /** Video stream of shared screen */
   screenShareStream?: VideoGalleryStream;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** State that the participant is in */
+  state?: string;
```
JamesBurnside, line 95: Avoid `string` as the type here, contoso won't know how to use it, can we use a string literal instead?
e.g. `state: 'connected' | 'onHold' | etc..
The literal values should be only what we use inside our react-components code and not every possible option

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2564,7 +2564,9 @@ export interface VideoGalleryProps {
 // @public
 export interface VideoGalleryRemoteParticipant extends VideoGalleryParticipant {
     isSpeaking?: boolean;
+    kind?: 'communicationUser' | 'phoneNumber' | 'microsoftTeamsUser' | 'unknown';
```
mgamis-msft, line 2567: I think you just need to build again to update the removal of the `kind` property.

packages/react-components/src/types/VideoGalleryParticipant.ts
dmceachernmsft: would something appropriate to do would maybe be to link the comment to the type our samples and composites would use? something @anjulgarg and I discussed was keeping the ACS headless types out of the components. If we restrict the string here to be the ones from calling we run into this issue here.

JamesBurnside: Don't restrict it to the ones from the calling SDK, but instead restrict it to ones that are used inside our react-components codebase. What are we currently using this state for? What are the different possible values it could be with respect to the video gallery?

packages/react-components/src/types/VideoGalleryParticipant.ts
```diff
@@ -90,4 +90,10 @@ export interface VideoGalleryRemoteParticipant extends VideoGalleryParticipant {
   isSpeaking?: boolean;
   /** Video stream of shared screen */
   screenShareStream?: VideoGalleryStream;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** Call state that the participant is in.
+   *
+   * If nothing is provided here the participant will appear as if in Connected state.
+   */
+  state?: 'Connecting' | 'Connected' | 'Ringing' | 'Hold' | 'Idle' | 'InLobby';
```
JamesBurnside, line 98: is videogallery using all these different participant states? If not scope down to only what is being used by the video gallery

PR #1933, opened 2022-05-26
Stop inverting container in picture-in-picture mode

prprabhu-ms:
# Why

Stops inverting the text displayed for local video tile when using the brand new picture-in-picture web API.

Note that the picture-in-picture video stops being inverted and there is no widely supported way to invert it right now.
Also, this PR is a stop-gap. Ideally the logic here should be moved to the headless SDK (and removed from the UI library).

# How Tested

`CallWithChat` sample.


packages/react-components/src/components/StreamMedia.tsx
```diff
@@ -39,6 +39,7 @@ export const StreamMedia = (props: StreamMediaProps): JSX.Element => {
   const theme = useTheme();
 
   const { isMirrored, videoStreamElement, styles } = props;
+  const [pipEnabled, setPipEnabled] = useState(false);
```
JamesBurnside, line 42: Dam, this can _almost_ be done in just CSS, sadly `:has` is in the CSS4 spec and not implemented yet: https://developer.mozilla.org/en-US/docs/Web/CSS/:has

packages/react-components/src/components/styles/StreamMedia.styles.ts
```diff
@@ -21,7 +21,18 @@ export const mediaContainer = (theme: Theme): string =>
...
-export const invertedVideoStyle = (theme: Theme): string =>
+export const invertedVideoInPipStyle = (theme: Theme): string =>
   mergeStyles(mediaContainer(theme), {
-    transform: 'rotateY(180deg)'
+    // The HTMLElement returned by the headless SDK is already inverted.
+    // But in picture-in-picture mode, we do not want to invert the host HTMLElement.
+    // Instead, we need to target the :picture-in-picture pseudoclass.
+    //
+    // First reset the host HTMLElement.
+    transform: 'rotateY(180deg)',
```
JamesBurnside, line 31: Today we apply `transform: 'rotateY(180deg)'` - is this un-rotating it? - Or how does this 'reset' the host element?

packages/react-components/src/components/styles/StreamMedia.styles.ts
```diff
@@ -21,7 +21,18 @@ export const mediaContainer = (theme: Theme): string =>
 /**
  * @private
  */
-export const invertedVideoStyle = (theme: Theme): string =>
+export const invertedVideoInPipStyle = (theme: Theme): string =>
   mergeStyles(mediaContainer(theme), {
-    transform: 'rotateY(180deg)'
+    // The HTMLElement returned by the headless SDK is already inverted.
```
JamesBurnside, line 26: This relies on the headless SDK not changing their implementation.. could we have the calling SDK fix this on their render element instead? (I can't see a way with just css though)

packages/react-components/src/components/StreamMedia.tsx
JamesBurnside: We should probably file a bug on them regardless as everyone using their render element will hit this bug

packages/react-components/src/components/styles/StreamMedia.styles.ts
JamesBurnside: We should probably file a bug on them regardless as everyone using their render element will hit this bug

packages/react-components/src/components/styles/StreamMedia.styles.ts
```diff
@@ -21,7 +21,18 @@ export const mediaContainer = (theme: Theme): string =>
...
+    // But in picture-in-picture mode, we do not want to invert the host HTMLElement.
+    // Instead, we need to target the :picture-in-picture pseudoclass.
+    //
+    // First reset the host HTMLElement.
+    transform: 'rotateY(180deg)',
+    // This doesn't work yet on latest Edge.
```
JamesBurnside, line 32: Testing this also and I can't find where this would exist in the DOM tree to ensure the pseudoclass is applied, the video seems to exist outside of the dom tree somewhere - can you find it in the inspect element tab?

packages/react-components/src/components/styles/StreamMedia.styles.ts
JamesBurnside: Ooo I see we weren't passing in `isMirrored=true` to StreamMedia lol so this was never being applied, instead of flipping another element, can we set `transform: 'rotateY(0deg)'` to whatever the CallingSDK is flipping? (though maybe a hacky div selector might be even more reliant on the calling sdk render element so would be a bad idea)

prprabhu-ms: Yeah I decided double flipping is better than poking into whatever headless returns.

For example, I didn't know that they return a the `<video>` inside a container `<div>`. It's the `<div>` that is transformed, but it could also be the `<video>` itself.

packages/react-components/src/components/styles/StreamMedia.styles.ts
prprabhu-ms: I'm already talking to the headless folks. This PR is a stop-gap.

packages/react-components/src/components/styles/StreamMedia.styles.ts
prprabhu-ms: Maybe the `:picture-in-picture` selector needs to be applied to root then... The `<video>` element doesn't move in the DOM. The picture in picture frame itself is outside of the DOM. It's a snowflake :(

prprabhu-ms: https://developer.mozilla.org/en-US/docs/Web/API/PictureInPictureWindow doesn't even inherit "common" properties... I'm hoping this API will evolve a bit

dmceachernmsft (approved):

PR #1935, opened 2022-05-27
Fix missing memoize in MediaGallery's onRenderAvatar

JamesBurnside:
# What
Wrap onRenderAvatar in a memoize (useCallback)

# Why
It changed every render pass - spotted entirely by react-render-tracker!
image

"But James - from that screenshot I also see remoteParticipants changing every render?". Indeed, another time another PR.

# How Tested
Ran locally and rendering avatar still a-OK for remote participants


packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -66,6 +66,15 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
     props.onFetchAvatarPersonaData
   );
 
+  const onRenderAvatar = useCallback(
```
JamesBurnside, line 69: @prprabhu-ms for review just to check this doesn't cause the anonymous issue we saw in the previous bug. if it does we do this a lot in out codebase .... 👀

prprabhu-ms (approved):

edwardlee-msft (approved):

PorterNan (approved):
This new perf tool is fantastic! Love the immediate result we get after using this!

dmceachernmsft (approved):

PR #1936, opened 2022-05-27
[DO NOT REVIEW] Transient Video Tile for PSTN when adding a participant.

dmceachernmsft:
# What

Adds a transient video tile to the library to handle remote participants joining calls
# Why

Allows us to display in the `VideoGallery` when someone is being connected to a call.

This PR is to serve as a basic example of what using the `VideoGallerySelector` Changes  from this PR https://github.com/Azure/communication-ui-library/pull/1932 might be used if we wanted to have the VideoGallery know about participants joining a call.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2865528 
Would be to address above if the `VideoGallery` is to support this logic.


packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -66,6 +67,27 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
     props.onFetchAvatarPersonaData
   );
 
+  /** we would want to use the onRenderRemoteVideoTile prop in the videoGallery to do something like this only
+   * rendering the transient tile when the remote user is in those states, otherwise we would want to use the defaule behavior
+   */
+  const onRenderTransientVideoTile = (remoteParticipant: VideoGalleryRemoteParticipant): JSX.Element | undefined => {
+    if (remoteParticipant.state !== ('connecting' || 'Ringing')) {
```
prprabhu-ms, line 74: This is where I think it is preferable to have two separate selectors, one each for connected and transient state participants. So that you don't have this logic about what states are transient etc here.

dmceachernmsft: thinking about this I think I agree, because we can provide the other tile based on whether they are a transient user or a connected user

PR #1939, opened 2022-05-30
Fixed bug when voice over does not announce chat status on iphone

carocao-msft:
# What
Fixed bug when voice over does not announce chat status on iphone by adding livemessage to chat status

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2791691

# How Tested
Tested with voiceover on iphone


packages/react-components/src/components/MessageStatusIndicator.tsx
```diff
@@ -96,6 +97,7 @@ export const MessageStatusIndicator = (props: MessageStatusIndicatorProps): JSX.
           data-ui-id="chat-composite-message-tooltip"
           calloutProps={{ ...calloutProps }}
         >
+          <LiveMessage message={strings.failedToSendAriaLabel ?? ''} aria-live="polite" />
```
dmceachernmsft, line 100: Is this just a announcer?

carocao-msft: Yeah basically, we use this in other places in the code for annoucing information

prprabhu-ms, line 100: nit: Should we only add the `LiveMessage` if we know `!!strings.failedToSendAriaLabel` is `true` to avoid an empty announcer entry?

packages/react-components/src/components/MessageStatusIndicator.tsx
```diff
@@ -116,6 +118,7 @@ export const MessageStatusIndicator = (props: MessageStatusIndicatorProps): JSX.
           data-ui-id="chat-composite-message-tooltip"
           calloutProps={{ ...calloutProps }}
         >
+          <LiveMessage message={strings.sendingAriaLabel ?? ''} aria-live="polite" />
           <Icon
             role="status"
             data-ui-id="chat-composite-message-status-icon"
```
prprabhu-ms, line 124: A few concerns with this approach:

* We already set `aria-label` on the `Icon` here. Why is this not enough?
* Adding announcer to just the `MessageStatusIndicator` is likely to lead to end-user confusion: Isn't it the case that this will simply announces "sending message" without saying which message? This could be particularly problematic for "message seen" because an older message might be seen and then the user won't know which message was seen.

The alternative would be to add `area-describedby` to the `MessageStatusIndicator` and use the message status into account when announcing the message (and announce the message again when status changes). So the announcement would be something like:

"Message "Hello world" sent by you at "xxx" was sent"
"Message "Hello world" sent by you at "xxx" was seen by "YY"
etc.

Not sure that's great UX (The "Hello world" part could be long and would become repetitive). My questions are:

* What is the UX we're aiming for here. Are we sure we want the UX we'll get by the current PR?
* Even for this UX, I don't understand why we need `LiveMessage` when we aleady have `aria-label` defined for the `Icon`.

prprabhu-ms: Please add a comment in the code with this explanation, so that we know in the future that we needed this for mobile.

prprabhu-ms (approved):

packages/react-components/src/components/MessageStatusIndicator.tsx
```diff
@@ -4,6 +4,7 @@
 import { ICalloutContentStyles, Icon, mergeStyles, TooltipHost } from '@fluentui/react';
 import { MessageStatus, _formatString } from '@internal/acs-ui-common';
 import React, { useState } from 'react';
+import { LiveMessage } from 'react-aria-live';
```
edwardlee-msft, line 7: Just a note for future reference:
Noting that LiveMessage needs to be wrapped in a LiveAnnouncer for it to work, according to the docs: 
https://www.npmjs.com/package/react-aria-live

Followed the trail and found that MessageThread.tsx has LiveAnnouncer which wraps MessageStatusIndicator like so:
\<LiveAnnouncer\> 
**->** \<Chat\> 
**->** item=MessageToDisplay 
**->** itemOnRenderMessageStatus 
**->** memoizeAllMessages 
**->** {message: content: ... defaultStatusRenderer} 
**->** \<MessageStatusIndicator\> 
**->** \<LiveMessage\>

(As of June 2, 2022, things may have changed in between the time you read this comment)

edwardlee-msft (approved):
Just a note for future reference:
Noting that LiveMessage needs to be wrapped in a LiveAnnouncer for it to work, according to the docs: 
https://www.npmjs.com/package/react-aria-live

Followed the trail and found that MessageThread.tsx has LiveAnnouncer which wraps MessageStatusIndicator like so:
\<LiveAnnouncer\> 
**->** \<Chat\> 
**->** item=MessageToDisplay 
**->** itemOnRenderMessageStatus 
**->** memoizeAllMessages 
**->** {message: content: ... defaultStatusRenderer} 
**->** \<MessageStatusIndicator\> 
**->** \<LiveMessage\>

(As of June 2, 2022, things may have changed in between the time you read this comment)

packages/react-components/src/components/MessageStatusIndicator.tsx
edwardlee-msft: Would like to note that **LiveMessage** needs to be wrapped in **LiveAnnouncer** in some shape or form.
So if using this in other places, make sure to double check that **LiveAnnouncer** is living as a parent, if not, make it a parent.

Pasting my other comment here for reference:


Just a note for future reference:
Noting that LiveMessage needs to be wrapped in a LiveAnnouncer for it to work, according to the docs: 
https://www.npmjs.com/package/react-aria-live

Followed the trail and found that MessageThread.tsx has LiveAnnouncer which wraps MessageStatusIndicator like so:
\<LiveAnnouncer\> 
**->** \<Chat\> 
**->** item=MessageToDisplay 
**->** itemOnRenderMessageStatus 
**->** memoizeAllMessages 
**->** {message: content: ... defaultStatusRenderer} 
**->** \<MessageStatusIndicator\> 
**->** \<LiveMessage\>

(As of June 2, 2022, things may have changed in between the time you read this comment)

PR #1940, opened 2022-05-30
[DO NOT REVIEW] PoC for replacing `GetSelector` with function overloads

prprabhu-ms:
with love for @dmceachernmsft

### DIFF
--- a/packages/calling-component-bindings/src/hooks/usePropsFor.ts
+++ b/packages/calling-component-bindings/src/hooks/usePropsFor.ts
@@ -104,21 +104,78 @@ export type GetSelector<Component extends (props: any) => JSX.Element | undefine
   : undefined;
 
 /**
- * Get the selector for a specified component.
- *
- * Useful when implementing a custom component that utilizes the providers
- * exported from this library.
+ * Get the selector for {@link VideoGallery}.
  *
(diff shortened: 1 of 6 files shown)


packages/calling-component-bindings/src/hooks/usePropsFor.ts
```diff
@@ -104,21 +104,78 @@ export type GetSelector<Component extends (props: any) => JSX.Element | undefine
...
 
 /**
- * Get the selector for a specified component.
+ * Get the selector for {@link VideoGallery}.
  *
- * Useful when implementing a custom component that utilizes the providers
- * exported from this library.
+ * @public
+ */
+export function getSelector<T>(component: typeof VideoGallery): VideoGallerySelector;
```
prprabhu-ms, line 111: Should be `<T extends typeof VideoGallery>` ensures that this function overload is not picked for any other component.

JamesBurnside:
> ## Calling bundle size is _**increased❗**_.
> * Current size: 4548224
> * Base size: 4548223
> * Diff size: 1

Rejecting PR, you added a whole byte to the bundle size

prprabhu-ms:
Only for demonstration (for now).

PR #1941, opened 2022-05-30
Prevent CallWithChat reload when toggling side panes

anjulgarg:
# What
Bugfix that prevents CallWithChat Composite from reloading whenever Chat/Participant pane is toggled.
Should greatly improve perf on mobile devices since it doesn't recreate the entire call arrangement when side pane is toggled

# Why
Fixes the following bug
https://skype.visualstudio.com/SPOOL/_workitems/edit/2835161

Improves performance of callwithchat

# How Tested
Manually

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
+++ b/packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
@@ -248,6 +248,10 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
     togglePeople();
   }, [togglePeople]);
 
+  const callCompositeContainerCSS = useMemo(() => {
+    return { display: isMobileWithActivePane ? 'none' : 'flex' };
+  }, [isMobileWithActivePane]);
+
   return (
     <div ref={containerRef} className={mergeStyles(containerDivStyles)}>
       <Stack verticalFill grow styles={compositeOuterContainerStyles} id={compositeParentDivId}>
@@ -256,7 +260,7 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
             grow
             styles={callCompositeContainerStyles}
             // Perf: Instead of removing the video gallery from DOM, we hide it to prevent re-renders.
-            style={{ display: isMobileWithActivePane ? 'none' : 'flex' }}
+            style={callCompositeContainerCSS}
           >
             <CallComposite
               {...props}
--- a/packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
+++ b/packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
@@ -252,7 +252,12 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
     <div ref={containerRef} className={mergeStyles(containerDivStyles)}>
       <Stack verticalFill grow styles={compositeOuterContainerStyles} id={compositeParentDivId}>
         <Stack horizontal grow>
-          <Stack.Item grow styles={callCompositeContainerStyles}>
+          <Stack.Item
+            grow
+            styles={callCompositeContainerStyles}
+            // Perf: Instead of removing the video gallery from DOM, we hide it to prevent re-renders.
+            style={{ display: isMobileWithActivePane ? 'none' : 'flex' }}
+          >
             <CallComposite
               {...props}
               formFactor={formFactor}
--- a/packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
+++ b/packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
@@ -252,17 +252,16 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
     <div ref={containerRef} className={mergeStyles(containerDivStyles)}>
       <Stack verticalFill grow styles={compositeOuterContainerStyles} id={compositeParentDivId}>
         <Stack horizontal grow>
-          {!isMobileWithActivePane && (
-            <Stack.Item grow styles={callCompositeContainerStyles}>
-              <CallComposite
-                {...props}
-                formFactor={formFactor}
-                options={{ callControls: false }}
-                adapter={callAdapter}
-                fluentTheme={fluentTheme}
-              />
-            </Stack.Item>
-          )}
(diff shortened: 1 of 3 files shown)


packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -252,17 +252,16 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
     <div ref={containerRef} className={mergeStyles(containerDivStyles)}>
       <Stack verticalFill grow styles={compositeOuterContainerStyles} id={compositeParentDivId}>
         <Stack horizontal grow>
-          {!isMobileWithActivePane && (
```
JamesBurnside, line 255: Can you test the incoming and outgoing video tiles in the PiPiP works as expected on mobile? i.e. start a call with another participant, turn on video for both, on one that is mobile open the pane and check the videos show correctly, then close the pane and check video gallery tiles are correct (Want to test in case the call video gallery might still have a hold to the video streams if it is not unmounted)

JamesBurnside: (we just unified all the logic surrounding creating and disposing streams so hopefully this won't be an issue anymore)

anjulgarg: Good Point. Let me test that

anjulgarg: Confirmed. Working as expected. :)

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatPane.tsx
```diff
@@ -138,8 +138,12 @@ export const CallWithChatPane = (props: {
       <Stack.Item verticalFill grow styles={paneBodyContainer}>
         <Stack horizontal styles={scrollableContainer}>
           <Stack.Item verticalFill styles={scrollableContainerContents}>
-            <Stack styles={props.activePane === 'chat' ? availableSpaceStyles : hiddenStyles}>{chatContent}</Stack>
-            <Stack styles={props.activePane === 'people' ? availableSpaceStyles : hiddenStyles}>{peopleContent}</Stack>
+            <Stack styles={props.activePane === 'chat' ? availableSpaceStyles(theme) : hiddenStyles}>
```
JamesBurnside, line 141: memoize styles now they are no longer a constant

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatPane.tsx
```diff
@@ -182,7 +186,9 @@ const sidePaneStyles: IStackStyles = {
   }
 };
 
-const availableSpaceStyles: IStackStyles = { root: { width: '100%', height: '100%' } };
+const availableSpaceStyles = (theme: Theme): IStackStyles => ({
+  root: { width: '100%', height: '100%', position: 'absolute', zIndex: 1, background: theme.palette.white }
```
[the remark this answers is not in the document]
anjulgarg: Any suggestions for handling the background color?
In any case, we will have to specify a color to the chat pane Or it will be transparent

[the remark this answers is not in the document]
JamesBurnside: my thinking was applying it to the root of the callwithchat or chat composite, but it doesn't really make a difference (and thinking more it would be wrong to apply to the root of the chat composite as it might want to be transparent) so fine to leave here!

JamesBurnside (approved):

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatPane.tsx
```diff
@@ -182,7 +192,15 @@ const sidePaneStyles: IStackStyles = {
   }
 };
 
-const availableSpaceStyles: IStackStyles = { root: { width: '100%', height: '100%' } };
+const availableSpaceStyles = (theme: Theme): IStackStyles => ({
+  root: {
+    width: '100%',
+    height: '100%',
+    position: 'absolute',
```
JamesBurnside, line 199: Just curious on the reason to overlay the chat on top of the call composite instead of setting the call composite visbility to hidden?

anjulgarg: Never came to my mind :D
Tested it and it works gr8 Haha

JamesBurnside: Oo good to know!

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -252,17 +252,21 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
...
-                options={{ callControls: false }}
-                adapter={callAdapter}
-                fluentTheme={fluentTheme}
-              />
-            </Stack.Item>
-          )}
+          <Stack.Item
+            grow
+            styles={callCompositeContainerStyles}
+            // Perf: Instead of removing the video gallery from DOM, we hide it to prevent re-renders.
```
JamesBurnside, line 258: <3

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -252,17 +252,21 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
...
+            style={{ display: isMobileWithActivePane ? 'none' : 'flex' }}
```
JamesBurnside, line 259: I don't know with the style prop if it has to be memoized... this will be a new object each render so maybe?

JamesBurnside (approved):

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
anjulgarg: We have had this discussion before. I think the verdict was to memoize `styles` and small inline `style` were fine.
Since `styles` are a prop for underlying fluent components and cause re-renders while `style` is just css.

anjulgarg: Actually, I am still gonna memoize it and move it outside. Good to follow a standard practice I think.
Thanks for pointing it out

PR #1945, opened 2022-06-01
Typing indicator UI test using fake adapter

mgamis-msft:
# What

Convert typing indicator UI test to use a fake adapter

❗Feel free to let me know  if you are not able to review or want to have walkthrough of this PR

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2845220

# How Tested

Ran UI test locally

# Process & policy checklist


mgamis-msft:
I will stabilize the flaky filesharing icon causing this snapshot [change](packages/react-composites/tests/browser/snapshots/beta/tests/browser/chat/fake-adapter/FileSharing.test.ts-snapshots/filesharing-file-download-card-in-sent-messages-Mobile-Android-Landscape-linux.png) in another PR.

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
```diff
@@ -180,3 +193,51 @@ const sendRemoteFileSharingMessage = (chatClientModel: Model, remoteParticipant:
...
+            }}
+          />
+        </_IdentifierProvider>
+      </div>
+    );
+  });
+};
+
+/**
+ * Helper function to create IDs for {@link _IdentifierProvider} using a user id
```
JamesBurnside, line 235: This is a little worrysome, its not obvious that if I add a data-ui-id somewhere in the DOM, it'll suddenly be appended by the participants userId - meaning I might write a UI test and have it fail not because the UI test couldn't find my element, but because the data-ui-id I specified was changed to be appended with the userID.
This also only applies when we use the IdentifierProvider which we've moved away from and are just specifying data-ui-ids typically as hardcoded strings directly in the DOM.

Consider having the hidden composites having the id of `hidden-composite-{userId}` instead, such that the query selector can then do something like:
`#hidden-composite-{userId} data-ui-id=[typing-indicator]` meaning its a totally transparent change for anyone not making use of the hidden-composites.

Either that or change _all_ data-ui-id selectors used throughout all tests to be appended with the userId

packages/react-composites/tests/browser/chat/fake-adapter/TypingIndicator.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    await waitForSelector(page, dataUiId(IDS.typingIndicator));
+    const indicator = await page.$(dataUiId(IDS.typingIndicator));
+
+    expect(await indicator?.innerHTML()).toContain(typingParticipant.displayName);
+    expect(await stableScreenshot(page, {})).toMatchSnapshot('typing-indicator.png');
+
+    // Advance time by 10 seconds to make typingindicator go away
+    await page.evaluate(() => {
+      const currentDate = new Date();
+      currentDate.setSeconds(currentDate.getSeconds() + 10);
```
JamesBurnside, line 35: any way the mocked chat client can mean we don't need to wait 10 seconds here ? That would be a huge improvement

packages/react-composites/tests/browser/chat/fake-adapter/TypingIndicator.test.ts
```diff
@@ -0,0 +1,43 @@
...
+      Date.now = () => currentDate.getTime();
+    });
+    await page.waitForTimeout(1000);
```
JamesBurnside, line 38: any idea why the extra second wait is here? (I realize it was already here but would be great to remove it)

packages/react-composites/tests/browser/chat/app/FakeAdapterApp.tsx
mgamis-msft: Great idea! Removed `IdentifierProvider` from the hidden composites and using the id as a qualifier for the selector. Thanks.

JamesBurnside (approved):

packages/react-composites/tests/browser/chat/fake-adapter/TypingIndicator.test.ts
mgamis-msft: Not sure how this works in terms of moving time forward but it seems like this UI tests takes less than 10 seconds to run.
https://skype.visualstudio.com/SPOOL/_workitems/edit/2845222

packages/react-composites/tests/browser/chat/fake-adapter/TypingIndicator.test.ts
mgamis-msft: It seems this is necessary for the date advancement in the previous to be realized.

PR #1954, opened 2022-06-02
RemoteVideoStream.isReceiving loading spinner

chriswhilar:
# What

Display a loading spinner over remote video streams when remotevideostream's isReceiving flag goes to false.
When isReceiving goes to false it means the video data is 'loading'.
Documentation about RemoteVideoStream.isReceiving flag and what it means: https://docs.microsoft.com/en-us/azure/communication-services/how-tos/calling-sdk/manage-video?pivots=platform-web#remote-video-stream-properties

Some scenarios that make this flag go to false:
Scenario 1: I am viewing the video of a remote participant who is on mobile browser. The remote participant bring the mobile browser app to the background. I now see a black/frozen video with a loading spinner over it (Its better UX to see the loading spinner and not just black/frozen video). When remote participant bring the mobile browser back to the foreground, the loading spinner is removed from UI and video continues to flow normally(no frozen/black video)

Scneario 2: I am viewing the video of a remote participant who is on whatever platforms. There are network issues from either side, his video start to look pretty laggy, bad quality, probbaly because of network issues, so i see a loading spinner on top of the video becuase it is basically "loading".

Scenario 3: On MacOS/iOS Safari, from the address bar, you can choose to "Pause" / "Resume" camera, in which case a cameraStoppedUnexpctedly true ufd will be triggered. Here we do not remove the local video stream from collection, because technically video is still on, just paused. Call state will say video is on. Remote side will see your video is on and they'll see RemoteVideoStream.isReceiving flag is false and see a loading spinner. You can click on resume and video will start flowing normally again and remote side will see RemotevideoStream.isReceivin flag go back to true and remove loading spinner.

Scenario 4: I am viewing the video of a remote participant who in on whatever platform. And for whatever reason their network disconnects. This will actually leave the remote participant in the call for a little while and ill see his video frozen/black with a loading spinner over it. The remote participant can get network back and reconnect and his audio/video should start flowing normally and i should not see a loading spinner over his video anymore.

Scenario 5: I am viewing the video of a remote participant who is on mobile browser. The remote participant terminates/kills the mobile browser. Since that remote participant was on mobile, this will actually leave the participant in the call for a little while and I will still see him in the call and his video will be frozen, I'll see a loading spinner over his frozen video. At some point, service will kick participant out of the call and I would just see that the participant disconnected from the call.

Scenario 6: Remote mobile user locks/unlocks device.

Scenario 7: Remote desktop user locks/sleeps device
 
There also other scenairos miscellanious scenario

- Example of disconnecting/reconnecting from network: 

  https://user-images.githubusercontent.com/65419124/171912914-8b8e36d3-3dee-43e7-9833-56eca65aa30a.mp4

- Example of bad network and then good network:

  https://user-images.githubusercontent.com/65419124/171913988-d1bd42d5-49ff-41ae-aa2c-a31af58a55c7.mp4

- Example of remote mobile user putting browser to background:
 
  https://user-images.githubusercontent.com/65419124/171911324-65a694ac-a31a-46c2-b8b0-ddaea4e52098.mp4

# Why

It provides a better user experience when remote video freezes/lags. So instead of just seeing a frozen video, youll see a loaing spinner. This is basically what streaming services do.

# How Tested

Will add e2e tests if we have it. We may not be able to do e2e automation for this though. Its not easy to make this flag go to false in automation.

# Process & policy checklist

No i dont think so


JamesBurnside:
This is awesome! For the UI change in this PR, can you post a screenshot or video in your PR description. Reviewing pure DOM&CSS is difficult to spot issues.

Agree e2e will be tough, it will be tough to capture the spinner as I imagine it doesn't hang around for a long time - once we complete the infra for mocked service e2e tests this would be a great candidate - @mgamis-msft is this possible today?

chriswhilar:
> 

@JamesBurnside yep sounds good i can add a screen recordgin so you can see it in action.

Yeah im thinkin e2e may be hard, we'd have to able to drive one of those scenarios in the description mentioned.

We the acs calling web sdk team have been able to do it but in CIFX which is a full blow lab with real devices lol. We do the scenario for mobile browser going to background in cifx.

I see in the CI, the Build storybook part fails, what is this and what do i have to do for that?

JamesBurnside:
> I see in the CI, the Build storybook part fails, what is this and what do i have to do for that?

Updated the pnpm lock files and removed the package-lock file (and merged in main after completing #1925). Unfortunately in our repo we can't use `npm install` for new packages, instead we have to use `cd <dir with package.json>; rush add -p <package name>` (or update the package.json manually then run `rush update`)

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -105,6 +105,10 @@ export interface RemoteVideoStreamState {
    * Proxy of {@link @azure/communication-calling#RemoteVideoStream.isAvailable}.
    */
   isAvailable: boolean;
+  /**
+   * Proxy of {@link @azure/communication-calling#RemoteVideoStream.isReceiving}.
+   */
+  isReceiving: boolean;
```
JamesBurnside, line 111: note: we will need to put this behind a beta conditional-compilation flag

packages/react-components/package.json
```diff
@@ -65,10 +65,11 @@
     "@types/jest": "~26.0.22",
     "@types/json-stringify-safe": "^5.0.0",
     "@types/node": "^14.14.10",
+    "@types/react": "^16.9.49",
     "@types/react-aria-live": "^2.0.0",
     "@types/react-dom": "^16.9.8",
     "@types/react-linkify": "^1.0.0",
-    "@types/react": "^16.9.49",
+    "@types/styled-components": "^5.1.25",
```
JamesBurnside, line 72: Is this needed?

packages/react-components/src/components/LoadingSpinner.tsx
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT license.
+import { Spinner } from '@fluentui/react';
+import React from 'react';
+import { loadingSpinnerContainerStyle } from './styles/LoadingSpinner.style';
+
+/**
+ * @public
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface LoadingSpinnerProps {}
```
JamesBurnside, line 11: remove empty interface

packages/react-components/src/components/LoadingSpinner.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const spinnerStyles = {
+  circle: {
+    height: '100px',
+    width: '100px',
+    borderWidth: '5px'
+  },
+  label: {
```
JamesBurnside, line 19: label here has an explicit value, but I don't see a label in the spinner, can this be removed?

packages/react-components/src/components/LoadingSpinner.tsx
```diff
@@ -0,0 +1,33 @@
...
+  }
+};
+
+/**
+ * @public
+ */
+export const LoadingSpinner = (): JSX.Element => {
+  return (
+    <div className={loadingSpinnerContainerStyle()}>
+      <Spinner styles={spinnerStyles} label={''} ariaLive="assertive" labelPosition="top" />
```
JamesBurnside, line 30: Open question to reviewers: do we need any accessibility aria label here? Tiles aren't currently focusable so perhaps not?

JamesBurnside, line 30: `ariaLive="assertive"` will be bad here as assertive will immediately announce any change (and thus could interrupt hearing someone on a call). Would suggest `polite` instead if required at all

JamesBurnside, line 30: Is `labelPosition="top"` needed if label is `''`?

packages/react-components/src/components/styles/LoadingSpinner.style.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { mergeStyles } from '@fluentui/react';
+
+/**
+ * @private
+ */
+export const loadingSpinnerContainerStyle = (): string =>
```
JamesBurnside, line 9: This positions the spinner centrally on top of anything its contained in, does this mean if there's a freeze frame of a video we will see the spinner on top of the freeze frame?
(My mental image of this would be to replace the video tile contents with a spinner)

packages/react-components/package.json
chriswhilar: styled-components? Nope doesnt look like its being used anywhere

chriswhilar:
@JamesBurnside @alkwa-msft @anjulgarg Hey folks i added to the pr description some screen recording examples

packages/react-components/src/components/styles/LoadingSpinner.style.ts
chriswhilar: Yep, iv added some screen recordings in the PR description so you can see how it works

chriswhilar: ill prob refactor this a bit and get rid of LoadingSpinner.tsx, i dont really think i need it

packages/react-components/src/components/LoadingSpinner.tsx
chriswhilar: Mmmm nope not needed. I can remove it, unless we wanna show a label like "Loading...", wdyt?

packages/calling-stateful-client/src/CallClientState.ts
chriswhilar: Ok ill add @beta. Do i just add the @beta tag here or everywhere?

packages/react-components/src/components/LoadingSpinner.tsx
JamesBurnside: I like the spinner on its own the way you have it!

packages/react-components/src/components/LoadingSpinner.tsx
chriswhilar: Mmmm nope not needed. I can remove it, unless we wanna show a label like "Loading...", wdyt?

packages/calling-stateful-client/src/CallClientState.ts
JamesBurnside: It's actually pretty complicated. We compile our codebase at all times for both stable and beta releases.
For stable we conditional-compile out all beta code.
We use compilation flags to achieve this, can read more here: https://github.com/Azure/communication-ui-library/blob/main/docs/references/beta-only-features.md
Have a stab if you like but otherwise I'm happy to add these flags once I get some time!

packages/react-components/src/components/styles/LoadingSpinner.style.ts
```diff
@@ -0,0 +1,19 @@
...
+  mergeStyles({
+    width: '100px',
```
JamesBurnside, line 11: use `rem` instead of pixles throughout (can manually convert (divide by 16) or use the _pxToRem fn we have)

packages/calling-stateful-client/src/CallClientState.ts
chriswhilar: I see, yep we also do this in web calling sdk.

Yes if you can pls add the beta flags, im not sure sure where to add and where not to add

JamesBurnside: Actually looks like 1.5.4-stable is released! Will make more sense to simply update to that and not require the beta flags!
I'll throw up a PR to update our stable build to use the latest stable calling sdk package then circle back to this comment once that's completed

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -65,6 +66,12 @@ export type CallCompositeOptions = {
    * @defaultValue true
    */
   callControls?: boolean | CallControlOptions;
+  /**
+   * Customize video rendering options
+   * Can be customized by providing an object of type {@link @azure/communication-react#VideoRenderingControlOptions}.
+   * @beta
+   */
+  videoRenderingControls?: VideoRenderingControlOptions;
```
JamesBurnside, line 74: for composites we dictate with strong opinion what UI is shown to the users, so this customization should be removed at the composite level and just always be enabled

packages/react-components/src/components/StreamMedia.tsx
```diff
@@ -69,12 +81,19 @@ export const StreamMedia = (props: StreamMediaProps): JSX.Element => {
...
-        styles?.root
+    <div className={container()}>
+      <div
+        className={mergeStyles(
+          isMirrored && pipEnabled ? invertedVideoInPipStyle(theme) : mediaContainer(theme),
+          styles?.root
+        )}
+        ref={containerEl}
+      />
+      {isRemoteVideoStream && videoRenderingControls?.remoteVideoStreamLoadingSpinner && !isReceiving && (
```
JamesBurnside, line 92: We'd want base components to be as non-opinionated as possible and avoid having complex internal logic that makes it hard for contoso to use -- e.g. if contoso sets `remoteVideoStreamLoadingSpinner = true` without looking at the source code they don't know when a spinner will and won't show as it relies on the exact state of three different props. The goal is for contoso to set explicit props that each impact one and only one thing with high clarity what will happen (for god-like components such as VideoGallery that badly need broken into sub components  this simply isn't possible).

Instead consider a a single prop `showLoadingSpinner` and the parent has to control when this is set to true or false.

We normally also want to avoid booleans in our API as they can't be extended once set (imagine if we want a different kind of loading spinner). So perhaps `showLoadingState?: 'spinner' -- but open to suggestions on naming here (cc @anjulgarg)

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -45,6 +45,8 @@ import { localVideoTileWithControlsContainerStyle, LOCAL_VIDEO_TILE_ZINDEX } fro
 import { _ICoordinates, _ModalClone } from './ModalClone/ModalClone';
 import { _formatString } from '@internal/acs-ui-common';
 import { _LocalVideoTile } from './LocalVideoTile';
+// eslint-disable-next-line no-restricted-imports
+import { VideoRenderingControlOptions } from '../../../react-composites/src/composites/CallComposite/types/VideoRenderingControlOptions';
```
JamesBurnside, line 49: `react-composites` dir has a dependency on `react-components` but react-components cannot use code from the composites dir.
Instead move any interfaces defined into the `react-components` dir and reference those when used in the react-composites dir. But `VideoRenderingControlOptions` should be able to be simply removed from the composites dir per https://github.com/Azure/communication-ui-library/pull/1954/files#r889292086

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -153,6 +155,10 @@ export interface VideoGalleryProps {
    * Camera control information for button to switch cameras.
    */
   localVideoCameraCycleButtonProps?: LocalVideoCameraCycleButtonProps;
+  /**
+   * Customize video rendering options
+   */
+  videoRenderingControls: VideoRenderingControlOptions;
```
JamesBurnside, line 161: Any new props on something exported publicly must be optional so as to not be a breaking API change
```suggestion
  videoRenderingControls?: VideoRenderingControlOptions;
```

packages/react-composites/src/composites/CallComposite/types/VideoRenderingControlOptions.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Customization options video rendering.
+ *
+ * @beta
+ */
+export type VideoRenderingControlOptions = {
+  /**
+   * Show a loading spinner over a remote video when video data is loading
+   * @defaultValue true
+   */
+  remoteVideoStreamLoadingSpinner: boolean;
```
JamesBurnside, line 14: We want to try avoid booleans as they can't be extended well. Consider
```suggestion
  videoStreamLoadingOverlay: 'none' | 'spinner';
```

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
chriswhilar: So should we just remove videoRenderingControls for now and just always use loading spinner for remote video streams? Tbh it should always be enabled anyways

chriswhilar: We have decided to just remove videoRenderingControlOptions

packages/react-components/src/components/StreamMedia.tsx
chriswhilar: We have decided to just remove videoRenderingControlOptions

packages/react-components/src/components/VideoGallery.tsx
chriswhilar: We have decided to just remove videoRenderingControlOptions

packages/react-components/src/components/VideoGallery.tsx
chriswhilar: We have decided to just remove videoRenderingControlOptions

packages/react-composites/src/composites/CallComposite/types/VideoRenderingControlOptions.ts
chriswhilar: We have decided to just remove videoRenderingControlOptions

packages/calling-stateful-client/src/RemoteVideoStreamSubscriber.ts
```diff
@@ -30,11 +30,13 @@ export class RemoteVideoStreamSubscriber {
 
   private subscribe = (): void => {
     this._remoteVideoStream.on('isAvailableChanged', this.isAvailableChanged);
+    this._remoteVideoStream.on('isReceivingChanged', this.isReceivingChanged);
     this.checkAndUpdateScreenShareState();
   };
 
   public unsubscribe = (): void => {
     this._remoteVideoStream.off('isAvailableChanged', this.isAvailableChanged);
+    this._remoteVideoStream.on('isReceivingChanged', this.isReceivingChanged);
```
JamesBurnside, line 39: `off` instead of `on` here

change/@internal-calling-component-bindings-73ffbe45-e3f7-47f2-a940-808f72633b07.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
JamesBurnside, line 2: this should be of type `prerelease` as this packlet code is only available in beta releases,
same as chage files for `@internal-calling-stateful-client` and `change/@internal-react-composites`.

`@internal/react-components` should stay as minor as this is code in the stable build

packages/calling-stateful-client/src/RemoteVideoStreamSubscriber.ts
JamesBurnside: update this

packages/calling-stateful-client/src/CallClientState.ts
JamesBurnside: Added beta flags 👍

change/@internal-calling-component-bindings-73ffbe45-e3f7-47f2-a940-808f72633b07.json
JamesBurnside: (easiest way to update this is to manually change the text and not through any `rush changelog` command)

packages/react-components/src/components/StreamMedia.tsx
```diff
@@ -17,6 +23,10 @@ export interface StreamMediaProps {
   videoStreamElement: HTMLElement | null;
   /** Decides whether to mirror the video or not. */
   isMirrored?: boolean;
+  /** Weather the remote stream is receiving data */
+  isReceiving?: boolean;
```
JamesBurnside, line 27: to prevent breaking changes this needs to default to `true` (another reason I would recommend renaming to something like `loadingState`)

packages/react-components/src/components/StreamMedia.tsx
```diff
@@ -1,12 +1,20 @@
...
+  invertedVideoInPipStyle,
+  mediaContainer,
+  container,
+  loadingSpinnerContainer,
+  loadSpinnerStyles
+} from './styles/StreamMedia.styles';
 import { useTheme } from '../theming';
 import { BaseCustomStyles } from '../types';
 
+type LoadingState = 'loading' | 'none';
```
JamesBurnside, line 16: note: this will need exported publicly
Then update both the stable and beta api files with `rush build-api:all-flavors`

JamesBurnside (approved):

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -379,6 +381,27 @@ export class CallContext {
     });
   }
 
+  /* @conditional-compile-remove(video-stream-is-receiving-flag) */
+  public setRemoteVideoStreamIsReceiving(
```
anjulgarg, line 385: Where is this function being used?

packages/react-components/src/components/StreamMedia.tsx
```diff
@@ -69,12 +83,19 @@ export const StreamMedia = (props: StreamMediaProps): JSX.Element => {
...
+    <div className={container()}>
+      <div
+        className={mergeStyles(
+          isMirrored && pipEnabled ? invertedVideoInPipStyle(theme) : mediaContainer(theme),
+          styles?.root
+        )}
+        ref={containerEl}
+      />
+      {loadingState === 'loading' && (
+        <div className={loadingSpinnerContainer()}>
```
anjulgarg, line 95: Can we memoize this className `loadingSpinnerContainer()` otherwise it will return a new object everytime the component re-renders.

anjulgarg: Ideally initialize these styles outside the component at file level.

anjulgarg (approved):

packages/react-components/src/components/StreamMedia.tsx
JamesBurnside: note: this will be somewhat memoized as it returns a string and should always return the same string (so not as important to memoize as when we use the `styles` prop that always is a new object)

JamesBurnside:
The loading spinner should be made a little more responsive to different widths, in the mobile preview window it is very large: 
image

JamesBurnside:
From the failed UI tests looks like the live service UI tests are picking up the loading spinner... I'll push a change to hide them via css in the ui tests as we can't guarantee if they will show or not

JamesBurnside:
Note: this is just blocked on being checked in on #1962

PR #1955, opened 2022-06-02
Fix logic for triggering updateScalingMode

JamesBurnside:
# What
* Fix logic for triggering updateScalingMode
* Add keys to the different video tiles so they are easily distinguishable when debugging
* Add unit tests for useVideoStreamLifecycleMaintainer
  * Update react testing library to latest (not much changed except has types now)

# Why
Logic was wrong, see comment in PR

# How Tested
* CI for new Calling SDK beta package now working (streams take slightly longer to render with the new package so flakiness for this issue amplified
* Added unit tests
* A lot locally adding and removing remote participants and sharing&stopsharing screen

### DIFF
--- a/packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
+++ b/packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
@@ -5,7 +5,8 @@ import React, { useEffect, useMemo, useRef } from 'react';
 import { VideoStreamOptions, CreateVideoStreamViewResult, ViewScalingMode } from '../../types';
 import { Cancellable, useCancellableTask } from '../utils/useCancellableTask';
 
-interface VideoStreamLifecycleMaintainerExtendableProps {
+/** @private */
+export interface VideoStreamLifecycleMaintainerExtendableProps {
   isStreamAvailable?: boolean;
   renderElementExists?: boolean;
   isMirrored?: boolean;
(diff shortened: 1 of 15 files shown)


packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
```diff
@@ -53,36 +57,47 @@ const useVideoStreamLifecycleMaintainer = (props: VideoStreamLifecycleMaintainer
...
       if (cancellable.cancelled) {
         return;
       }
-      scalingModeRef.current = scalingMode;
+      scalingModeForDirectUpdateRef.current = scalingMode;
     });
   }
 
   // scalingModeForUseEffect will trigger the useEffect to recreate the stream view only if the scaling mode has changed and
   // we cannot call updateScalingMode on the view object directly. Otherwise it will be null to ensure the useEffect does not trigger
```
JamesBurnside, line 84: update this comment

packages/react-components/src/components/VideoGallery/useVideoStreamLifecycleMaintainer.ts
```diff
@@ -53,36 +57,47 @@ const useVideoStreamLifecycleMaintainer = (props: VideoStreamLifecycleMaintainer
...
   // when scaling mode changes.
-  const scalingModeForUseEffect = updatingScalingModeDirectly ? null : scalingMode;
```
JamesBurnside, line 76: This was ultimately what what was wrong:
* On first render, scaling mode is "Crop", therefore the dep in the useEffect below is Crop
* When we finish `onCreateStreamView` and this has another render pass with RenderElement === true, `updatingScalingModeDirectly` can also be true now
* When `updatingScalingModeDirectly` the scalingModeForUseEffect is now `null`
  * _This now differs from "Crop" it originally was causing the below useEffect to destroy and recreate the stream_

JamesBurnside:
Found a bug, closing until can be resolved

PR #1956, opened 2022-06-03
Add support for Teams identity

prprabhu-ms:
Add minimal support for the in-preview feature in the calling SDK to use teams identity backed tokens for joining Teams calls.

After this PR, teams endpoint tokens can be used with the Teams user ID to join Teams meetings.
There is no support for 1:1 / 1:N calling in this PR.

Added behind a new feature flag.

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
+++ b/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
@@ -44,7 +44,11 @@ import {
 import { getCallCompositePage, isCameraOn } from '../utils';
 import { CreateVideoStreamViewResult, VideoStreamOptions } from '@internal/react-components';
 import { fromFlatCommunicationIdentifier, toFlatCommunicationIdentifier } from '@internal/acs-ui-common';
-import { CommunicationTokenCredential, CommunicationUserIdentifier } from '@azure/communication-common';
+import {
+  CommunicationTokenCredential,
+  CommunicationUserIdentifier,
+  MicrosoftTeamsUserIdentifier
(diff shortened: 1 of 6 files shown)


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -591,7 +595,7 @@ export type CallAdapterLocator =
  * @public
  */
 export type AzureCommunicationCallAdapterArgs = {
-  userId: CommunicationUserIdentifier;
+  userId: CommunicationUserIdentifier | /* @conditional-compile-remove(teams-identity) */ MicrosoftTeamsUserIdentifier;
```
JamesBurnside, line 598: can you add this to the callwithchat adapter args also?

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
prprabhu-ms: Didn't add it there intentionally.

We can't use CTE token for access ACS chat service, so it doesn't make sense to support this in `CallWithChatComposite` at this point.

alkwa-msft:
Do we want to say Communicationuseridentifier | teams Identifier  

or should we just use CommunicationIdentifier and simplify everything?

dmceachernmsft (approved):

prprabhu-ms:
Teams identity API is going to be a bit more complicated at the headless SDK layer. Abandoning this teeny tiny PR until the headless API is available to build upon.

Issue #1900, opened 2022-05-17
Support React 18


juandgarin:
When i try to install the library, it returns me an error indicating that there are dependencies that cant resolve.

Anyone with the same problem?

I think it is a React 18 issue

![Screenshot 2022-05-17 at 06 09 36](https://user-images.githubusercontent.com/62432378/168774913-9f4cbe9d-1507-4bb8-9d50-2826dcb40938.png)

mgamis-msft:
Hi @juandgarin, thank you for filing this issue. I will look into this today.

mgamis-msft:
Hi @juandgarin, our dependency on [fluent ui library](https://www.npmjs.com/package/@fluentui/react) restricts our package from using React 18 or higher. What is your use case for `backoffice_next@0.1.0`? Is it possible to use another version of backoffice that does not rely on React 18?

prprabhu-ms:
Hi @juandgarin I don't expect us to be able to lift the < React18 requirement easily at the moment (Fluentv8 recently moved to React 17, and I'm not aware of any plans to move beyond React18+)

Closing this issue as it is not currently actionable, but please feel free re-open with any more information / insight you have for the need for React 18+

prprabhu-ms:
Actually, let's keep this open to track when FluentUI goes to React 18 and then this issue will be actionable: https://github.com/microsoft/fluentui/issues/22592

prprabhu-ms:
It is still unclear if Fluentv8 will gain React18 support at all: https://github.com/microsoft/fluentui/issues/22592#issuecomment-1113633465

sai0909:
I am also facing the same issue when I followed these steps. https://azure.github.io/communication-ui-library/?path=/docs/quickstarts-uicomponents--page

prprabhu-ms:
@sai0909 Can you try downgrading to React v17 to see if it fixes your issue?

prprabhu-ms:
It is a big problem that anybody using our quickstart will currently hit the problem posted in OP.

I confirmed that downgrading the app to React 17 with

```bash
npm install react@17 react-dom@17 @testing-library/react@12 @types/react@17 @types/react-dom@17
```

almost fixes the issue.
The problem is that `react-dom` had a breaking change from v17 -> v18 so that the `create-react-app` generated boilerplate doesn't work with React 17.

If you modify the generated `index.tsx` to be 

```ts
import React from 'react';
import { render }  from 'react-dom';
import './index.css';
import App from './App';
[...]

PR #1959, opened 2022-06-03
Update quickstarts to avoid React18 pitfall

prprabhu-ms:
The default instructions in our storybook currently lead developers to a React v18 application. This is broken because Fluent UI does not support React18 yet.

Pin the `react-scripts` version in the instructions to provide a hassle-free first devX.

Tiny mitigation for #1900


prprabhu-ms:
This won't work.

`create-react-app` simply picks up the latest versions of `react` and `react-dom`.

There is absolutely no way to specify an older version - not on the command line, nor via a custom template.

:(

PR #1962, opened 2022-06-06
Remove tsc in app build and only do transipleOnly when rushx start

PorterNan:
# What

1. Remove tsc in app build
2. Only do transipleOnly when rushx start
3. Add conditional compilation to CallWithChat Sample app
# Why

# How Tested

# Process & policy checklist


JamesBurnside (approved):
Approved x1000

prprabhu-ms (approved):
What about StaticHTMLComposites, and ComponentExamples apps?

prprabhu-ms:
Sending this PR on its way to unblock #1954

PorterNan:
> > Sending this PR on its way to unblock #1954
> 
> Change looks to be causing out-of-memory exceptions /: image

That's weird, we don't add anything additional to the build, but only remove tsc from it

JamesBurnside:
> That's weird, we don't add anything additional to the build, but only remove tsc from it

We are updating the build to have webpack perform type checking, must be tipping it over the edge, is there a way to increase the heap allocation?

PorterNan:
> > That's weird, we don't add anything additional to the build, but only remove tsc from it
> 
> We are updating the build to have webpack perform type checking, must be tipping it over the edge, is there a way to increase the heap allocation for a node process? (or have we literally eaten all the RAM the build agent allows)

Looks like there is a logic problem, we are not enabling it, we are disabling type check in CI build but enable that in local development (Cause tsc was running in other pipeline already)

JamesBurnside:
> Looks like there is a logic problem, we are not enabling it, we are disabling type check in CI build but enable that in local development (Cause tsc was running in other pipeline already)

Oh I thought it was the opposite way around, what pipeline type checks the sample code files?

PorterNan:
> > Looks like there is a logic problem, we are not enabling it, we are disabling type check in CI build but enable that in local development (Cause tsc was running in other pipeline already)
> 
> I thought it was the opposite way around, what pipeline type checks the sample code files? The reason it was disabled for local development was it was taking so long so we decided to only typecheck at build&CI time

The original idea is - we need to disable typecheck in pipeline because it is not following conditional compilation

But I realized we might still need some typeguards for our ci build - so I add a conditional running tsc here - which will only run when it is beta

PR #1965, opened 2022-06-06
Stub MORE css classnames in storybook snapshot tests

JamesBurnside:
# What
* Overriding getClassName was running in the beforeEach and not at global scope.
* Adding resetIds also (typically its just the classnames that are causing issues, we can monitor this one but it doesn't have a good way to override it)

# Why
Previous PR #1963 completed a little early - found some more when looking at the snapshots in the PR.

### DIFF
--- a/packages/storybook/stories/storybook.test.ts
+++ b/packages/storybook/stories/storybook.test.ts
@@ -15,13 +15,13 @@ jest.mock('@azure/communication-calling', () => {
 
 ReactDom.createPortal = (node: any) => node;
 
-// Classnames in fluent are of the format <display name>-#, where # is an integer that is incremented by a global singleton
-// when each new classname is generated.
+// Classnames in fluent are of the format <classname-prefix>-# (e.g. css-42), where # is an integer that is incremented by a
+// global singleton when each new classname is generated.
 // Here we mock the classname generation for two reasons:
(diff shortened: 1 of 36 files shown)


packages/storybook/stories/storybook.test.ts
```diff
@@ -15,15 +15,18 @@ jest.mock('@azure/communication-calling', () => {
 
 ReactDom.createPortal = (node: any) => node;
 
-// Classnames in fluent are of the format css-#, where # is an integer that is incremented by a global singleton
+// Classnames in fluent are of the format <display name>-#, where # is an integer that is incremented by a global singleton
```
mgamis-msft, line 18: Sorry, I don't understand this. The display name of what?

JamesBurnside: Just what fluent calls it: https://github.com/microsoft/fluentui/blob/66dc6e9684baf35e7930622e1f3accdb51ba8516/packages/merge-styles/src/Stylesheet.ts#L44
I'll update this to `<classname-prefix>-#` and add an example for clarity!

JamesBurnside: updated

JamesBurnside:
Still seeing some but very few:
image

Lets check this in and see how much these still existing ones impact PRs

mgamis-msft (approved):

prprabhu-ms (approved):

PR #1966, opened 2022-06-06
Dispose of adapters in the `beforeUnload` in our Sample apps

JamesBurnside:
# What
Dispose of adapters in the `beforeUnload` event of the browser

# Why
Follow Calling's Best Practices guidence: https://docs.microsoft.com/en-us/azure/communication-services/concepts/best-practices#hang-up-the-call-on-onbeforeunload-event

# How Tested
I didn't due to small change so please scrutinize code


samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -83,6 +83,15 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     afterCreate
   );
 
+  // Dispose of the adapter in the window's before unload event.
+  // This ensures the service knows the user intentionally left the call if the user
+  // closed the browser tab during an active call.
+  useEffect(() => {
```
prprabhu-ms, line 89: We should probably use the `hidden` event instead of `beforeunload`. `beforeunload` is a legacy API that is not recommended: https://developer.chrome.com/blog/page-lifecycle-api/#the-beforeunload-event

Aside: I also find the name `beforeunload` unsettling. I always first assume that the page is about to be torn down... so React state is all gone and the components are all gone and this `useEffect` has already called `removeEventListener`. That's not true though -- nothing has *yet* been unloaded at all.

prprabhu-ms: Actually scratch that. We _don't_ wan to dispose the adapter every time the page goes to `hidden` state. The page could very well return to `active` state.

So, `beforeunload` is our only choice here. That said, perhaps we should follow the guideline in the blog above and only subscribe to `beforeunload` when we join a call, so that we can `leaveCall` before unloading. Being in a call is our "unsaved state" that we want to clean up before unload.

JamesBurnside: My feeling was the adapter's dispose method should handle if its in a call and hangup accordingly (see #1968 for calling call.hangup in the dispose method). We're gonna have to tell contoso to do this in a best practices page of our own so wanted to keep the snippet as small as possible so either `adapter.dispose()` or ~ `if (adapter.getState().inCall()) { adapter.getState().call.hangUp()}`

anjulgarg (approved):

prabhjot-msft (approved):

PR #1967, opened 2022-06-06
Support ternary type definitions in conditional compilation

prprabhu-ms:
# Why

`GetSelector` needs it.

# How Tested

Added examples to live documentation.

### DIFF
--- a/common/scripts/babel-conditional-preprocess.js
+++ b/common/scripts/babel-conditional-preprocess.js
@@ -104,12 +104,17 @@ exports.default = babelHelper.declare((_api, opts) => {
             	Handle(identifier_path, featureSet, stabilizedFeatureSet);
             }
         }});
+      },
+
+      TSConditionalType(path) {
+        HandleConditionalType(path, featureSet, stabilizedFeatureSet);
       }
     }
   };
 });
 
-function Handle(path, featureSet, stabilizedFeatureSet) {
+
+function Handle(path, featureSet, stabilizedFeatureSet, relaceWith=undefined) {
   let { node } = path;
 
   if (!node.leadingComments) {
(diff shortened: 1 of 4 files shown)


common/scripts/babel-conditional-preprocess.js
```diff
@@ -134,13 +139,37 @@ function Handle(path, featureSet, stabilizedFeatureSet) {
...
-    path.remove();
+
+  switch(path?.container?.type) {
+    case 'JSXExpressionContainer':
+      // We cannot remove Expression in JSXExpressionContainer cause it is not correct for AST
+      // Replacing it with jSXEmptyExpression will get us the same result
+      // There will always be only one expression under JSXExpressionContainer
+      path.replaceWith(t.jSXEmptyExpression());
+      break;
+    case 'TSConditionalType':
```
prprabhu-ms, line 150: Err... remove this, please?

dmceachernmsft (approved):

prprabhu-ms:
Manually verified the diff of the conditionally compiled sample:

```
@prprabhu-ms ➜ /workspaces/communication-ui-library/packages/acs-ui-common (prprabhu/support-ternary-type-if ✗) $ diff ./index.backup.tsx preprocessed/conditional-compilation-sample/index.tsx 
79a80,93
>  * Conditionally compile a ternary type if.
>  *
>  * Used in selector {@link GetSelector} definitions etc.
>  *
>  * See expanded example below.
>  *
>  * WARNING: Conditional compilation directives must be added in the conditional branch to be removed (not in the condition itself).
>  * When a branch is conditionally compiled out, both the condition and the other branch are removed.
>  */
> 
> export type MyFalseySelector<T> = false;
> export type MyTrutySelector<J> = true;
> export type MyNestedSelector<Q, R> = false;
> /**
209a224,232
> /**
>  * When adding a new selector to beta-only API, you need to extend the relevant {@link GetSelector} type.
>  * This example shows how to conditionally add your selector type.
>  *
>  * WARNING: The conditional compilation directive is in the branch to be removed, NOT the condition. (So not before {@link AreEqual}).
>  */
> 
> import { AreEqual } from '../areEqual';
> export type GetSelector<Component extends (props: unknown) => JSX.Element | undefined> = AreEqual<Component, typeof utils.StableComponentWithSelector> extends true ? utils.StableSelector : undefined;
```

carocao-msft (approved):

packages/acs-ui-common/src/conditional-compilation-sample/index.tsx
```diff
@@ -270,6 +287,23 @@ export const myExtensibleSelector: MyExtensibleSelector = utils.dummyCreateSelec
   }
 );
 
+/**
```
prprabhu-ms, line 290: @carocao-msft here's how you'd add a beta-only selector.

PR #1968, opened 2022-06-06
Ensure resources are cleaned up when disposing of adapter

JamesBurnside:
# What
Call leaveCall in the dispose method to ensure call.hangup is triggered as well as resource cleanup

# Why

* We want to make sure `call.hangup()` is correctly called anytime we dispose of our adapter per https://docs.microsoft.com/en-us/azure/communication-services/concepts/best-practices#hang-up-the-call-on-onbeforeunload-event.
* Gives a chances for any onHangUp code to execute

Related PR: #1966


mgamis-msft (approved):

JamesBurnside:
This will cause an extra call ended event to emit. Since callclient.dispose is already called this is not necessary. Closing.

PR #1973, opened 2022-06-07
[PSTN] Update call Adapter to use new PSTN handlers (Hold, addParticipant)

dmceachernmsft:
# What

Update Call and CallWithChat Adapters to have `onToggleHold` and `addParticipant` handlers
# Why

Prepares the CallComposite and CallWithChatComposite to perform PSTN actions

https://skype.visualstudio.com/SPOOL/_workitems/edit/2876718
# How Tested

Built locally, checked to see if exposed properly in sample apps.

### DIFF
--- a/packages/communication-react/review/beta/communication-react.api.md
+++ b/packages/communication-react/review/beta/communication-react.api.md
@@ -183,6 +183,8 @@ export interface CallAdapter extends AdapterState<CallAdapterState>, Disposable,
 
 // @public
 export interface CallAdapterCallManagement {
+    // @beta
+    addParticipant(participant: CommunicationIdentifier, options?: AddPhoneNumberOptions): Promise<void>;
     createStreamView(remoteUserId?: string, options?: VideoStreamOptions): Promise<void | CreateVideoStreamViewResult>;
     disposeStreamView(remoteUserId?: string, options?: VideoStreamOptions): Promise<void>;
     joinCall(microphoneOn?: boolean): Call | undefined;
@@ -194,6 +196,8 @@ export interface CallAdapterCallManagement {
     startScreenShare(): Promise<void>;
     stopCamera(): Promise<void>;
     stopScreenShare(): Promise<void>;
+    // @beta
+    toggleHold(): Promise<void>;
     unmute(): Promise<void>;
 }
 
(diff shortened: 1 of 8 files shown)


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -194,6 +196,8 @@ export interface CallAdapterCallManagement {
     startScreenShare(): Promise<void>;
     stopCamera(): Promise<void>;
     stopScreenShare(): Promise<void>;
+    // @beta
+    toggleHold(): Promise<void>;
```
JamesBurnside, line 200: is there a strong case for `toggle` here? Are other handlers are in pairs of stop/start so for consistency might be better to have `hold()` and `endHold()` (not convinced `endhold` is the right terminology lol)

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatAdapter.ts
```diff
@@ -276,6 +280,12 @@ export interface CallWithChatAdapterManagement {
   /* @conditional-compile-remove(file-sharing) */
   /** @beta */
   updateFileUploadMetadata: (id: string, metadata: FileMetadata) => void;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** @beta */
+  toggleHold: () => Promise<void>;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** @beta */
```
JamesBurnside, line 287: nit: these should have a doc comments with some explaination (although pretty obvious)

JamesBurnside (approved):
Just one comment on toggle vs start/stop but otherwise looks great!

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: Resume() perhaps?

dmceachernmsft: Will do!

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -418,6 +428,21 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
   }
 
+  /* @conditional-compile-remove(PSTN-calls) */
+  public async addParticipant(participant: CommunicationIdentifier, options?: AddPhoneNumberOptions): Promise<void> {
+    this.handlers.onAddParticipant(participant, options);
+  }
+
+  /* @conditional-compile-remove(PSTN-calls) */
+  public async holdCall(): Promise<void> {
+    this.handlers.onToggleHold();
```
JamesBurnside, line 438: note: this can't blindly call `onToggleHold`, it will need to check if the call handlers.holdCall() or check if the call isn't already on hold

carocao-msft (approved):

PR #1974, opened 2022-06-08
Convert read receipt UI tests to use fake adapter

mgamis-msft:
# What

Convert read receipt UI tests to use fake adapter

Note: the 3 snapshots removed in folder packages/react-composites/tests/browser/snapshots/beta/tests/browser/chat/ChatComposite.test.ts-snapshots are snapshots that take a snapshot of the remote participant receiving a message. Getting this snapshot of a hidden composite could make the fake adapter app more complicated that it already is and this snapshot can be verified in a different test.
 
# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2845224

# How Tested

Ran UI tests locally and compared run duration in comment of this ticket: https://skype.visualstudio.com/SPOOL/_workitems/edit/2845226

# Process & policy checklist


dmceachernmsft (commented):
Scroll bar bug is present here as well.

packages/react-composites/tests/browser/chat/fake-adapter/ReadReceipt.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    const testMessageText = 'How the turn tables';
+    await sendMessage(page, testMessageText);
+    await waitForMessageDelivered(page);
+    expect(await stableScreenshot(page, { stubMessageTimestamps: true })).toMatchSnapshot('sent-messages.png');
+
+    await setParticipantAbleToSeeMessages(page, messageReader, true);
+    await waitForMessageSeen(page);
+
+    await page.locator(dataUiId('chat-composite-message-status-icon')).click();
+    await page.waitForSelector(dataUiId('chat-composite-message-tooltip'));
```
JamesBurnside, line 32: use the `waitForSelector` utils wrapper method we have in the utils instead of the page method

mgamis-msft:
> Scroll bar bug is present here as well.

Fixed in #1992 :)

dmceachernmsft (approved):
Make a comment in the E2E doc about the weird snapshot renames on a failed test. Otherwise looks great!

packages/react-composites/tests/browser/chat/fake-adapter/ReadReceipt.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    await waitForSelector(page, '[id="chat-composite-message-contextual-menu-read-name-list"]');
+    expect(await page.screenshot()).toMatchSnapshot('read-message-contextualMenu.png');
+  });
+});
+
+async function getHiddenCompositeHandleOfParticipant(
+  page: Page,
+  participantIdentifier: CommunicationIdentifier
+): Promise<ElementHandle<SVGElement | HTMLElement>> {
+  return await page.waitForSelector(`[id="hidden-composite-${toFlatCommunicationIdentifier(participantIdentifier)}"]`, {
```
JamesBurnside, line 75: use `waitForSelector(page` instead

JamesBurnside (approved):
Where these intentionally not needed in the updated tests?
image

mgamis-msft:
> are snapshots that take a snapshot of the remote participant receiving a message. Getting this snapshot of a hidden composite could make the fake adapter app more complicated that it already is and this snapshot can be verified in a different test.



> Where these intentionally not needed in the updated tests? image

Yes, they were. These where live test snapshots taken when switching to page a remote participant receiving a message. For fake test, getting this snapshot of a hidden composite could make the fake adapter app more complicated that it already is and this snapshot can be verified in a different test. So I did not think it was worth it. The important thing in this test is the remote participant emitting a read receipt event.

PR #1975, opened 2022-06-08
Declarative Incoming Call

anjulgarg:
# What
- Adds DeclarativeIncomingCall class
- Manage declarative incoming call objects inside declarative callagent
- Add a new method `getIncomingCall` in declarative call agent to allow contoso to obtain a declarative incoming call object using incoming call id.

# Why
If Contoso is observing stateful clients state changes, they will get access to the incoming call id. To get the incoming call object, we need an additional method on the call agent.

# How Tested
TBD

# Process & policy checklist


packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -1,26 +1,51 @@
...
  * unsubscribe from all state updates.
  */
-class ProxyCallAgent implements ProxyHandler<CallAgent> {
+class ProxyCallAgent implements ProxyHandler<DeclarativeCallAgent> {
   private _callAgent: CallAgent;
   private _context: CallContext;
   private _internalContext: InternalCallContext;
   private _callSubscribers: Map<Call, CallSubscriber>;
   private _incomingCallSubscribers: Map<string, IncomingCallSubscriber>;
+  private _declarativeIncomingCalls: Map<string, DeclarativeIncomingCall>;
```
dmceachernmsft, line 48: Should this not be conditionally compiled?

packages/calling-stateful-client/src/IncomingCallDeclarative.ts
```diff
@@ -0,0 +1,49 @@
...
+          return await target.accept(...args);
+        }, 'IncomingCall.accept');
+      }
+      case 'reject': {
+        return this._context.withAsyncErrorTeedToState(async function (...args: Parameters<IncomingCall['reject']>) {
+          return await target.reject(...args);
+        }, 'IncomingCall.reject');
+      }
+      default:
+        return Reflect.get(target, prop);
```
dmceachernmsft, line 36: What does Reflect do here?

packages/calling-stateful-client/src/CallAgentDeclarative.ts
anjulgarg: It's private. Doesn't change public API :)

packages/calling-stateful-client/src/IncomingCallDeclarative.ts
anjulgarg: If we don't proxy a function on incomingCallObject, it exposes the default implementation

common/config/babel/.babelrc.js
```diff
@@ -26,7 +26,9 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         // PSTN calls 
         'PSTN-calls',
         // props to allow Contoso to overwrite timestamp format for chat messages, one in locale and one in message thread component
-        'date-time-customization'
+        'date-time-customization',
+        // 1 to N Calling feature.
+        '1-n-calling'
```
prprabhu-ms, line 31: Heh. Even though I thoroghly know how the conditional compilation is implemented and know that it is just string matching... something primal in me recoils at a flag starting with a numeral. :)

Make it `one-to-n-calling` and I'll get you an extra coffee.

(This comment is about 98% a joke)

packages/calling-stateful-client/review/stable/calling-stateful-client.api.md
```diff
@@ -70,7 +71,7 @@ export type CallErrors = {
...
+export type CallErrorTarget = 'Call.addParticipant' | 'Call.feature' | 'Call.hangUp' | 'Call.hold' | 'Call.mute' | 'Call.off' | 'Call.on' | 'Call.removeParticipant' | 'Call.resume' | 'Call.sendDtmf' | 'Call.startScreenSharing' | 'Call.startVideo' | 'Call.stopScreenSharing' | 'Call.stopVideo' | 'Call.unmute' | 'CallAgent.dispose' | 'CallAgent.feature' | 'CallAgent.join' | 'CallAgent.off' | 'CallAgent.on' | 'CallAgent.startCall' | 'CallClient.createCallAgent' | 'CallClient.feature' | 'CallClient.getDeviceManager' | 'DeviceManager.askDevicePermission' | 'DeviceManager.getCameras' | 'DeviceManager.getMicrophones' | 'DeviceManager.getSpeakers' | 'DeviceManager.off' | 'DeviceManager.on' | 'DeviceManager.selectMicrophone' | 'DeviceManager.selectSpeaker' | 'IncomingCall.accept' | 'IncomingCall.reject';
```
prprabhu-ms, line 74: +1 to adding these to the stable API directly.

packages/calling-stateful-client/review/stable/calling-stateful-client.api.md
```diff
@@ -106,6 +107,9 @@ export type CreateViewResult = {
     view: VideoStreamRendererView;
 };
 
+// @public
+export type DeclarativeCallAgent = CallAgent & {};
```
prprabhu-ms, line 111: Do you _have_ to add this to stable API right now?

You should be able to add the function overload to @beta only instead, right?

This makes for an awkward stable API.

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -1,26 +1,51 @@
...
+export type DeclarativeCallAgent = CallAgent & {
+  /**
+   * @beta
+   * Used to return a declarative incoming call object for an incoming call id.
+   *
+   * @Remark This function doesn't exist on the {@link @azure/communication-calling#CallAgent} interface.
+   * @param incomingCallId - Incoming Call ID
+   * @returns proxied IncomingCall
+   */
+  /* @conditional-compile-remove(1-n-calling) */
```
prprabhu-ms, line 33: nitty nit: Move the conditional compilation directive above the docstring so that `tsdoc` isn't confused.

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -104,18 +131,20 @@ class ProxyCallAgent implements ProxyHandler<CallAgent> {
...
+    if (!this._incomingCallSubscribers.has(incomingCall.id)) {
       this._incomingCallSubscribers.set(
-        event.incomingCall.id,
-        new IncomingCallSubscriber(event.incomingCall, this.setIncomingCallEnded)
+        incomingCall.id,
+        new IncomingCallSubscriber(incomingCall, this.setIncomingCallEnded)
       );
     }
-    this._context.setIncomingCall(convertSdkIncomingCallToDeclarativeIncomingCall(event.incomingCall));
+    this._declarativeIncomingCalls.set(incomingCall.id, incomingCallDeclaratify(incomingCall, this._context));
```
prprabhu-ms, line 146: @PorterNan Can te `Call.id` change for an incoming call (before it is accepted / rejected)? If yes, we might need to handle the `idChanged` event similar to ongoing call.

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -187,6 +216,20 @@ class ProxyCallAgent implements ProxyHandler<CallAgent> {
           });
         };
       }
+      /**
```
prprabhu-ms, line 219: Great comment. Thank you!

prprabhu-ms (approved):
<3 Very clean PR.

prprabhu-ms:
> ## CallWithChat bundle size is ***increased❗***.
- Current size: 5093018
- Base size: 5092004
- Diff size: 1014 

LGTM.

packages/calling-stateful-client/src/CallAgentDeclarative.ts
anjulgarg: Haha, confirmed with Artur. It doesn't change :)

common/config/babel/.babelrc.js
anjulgarg: As you desire :P

packages/calling-stateful-client/review/stable/calling-stateful-client.api.md
anjulgarg: Made it cleaner as suggested

dmceachernmsft (approved):

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -2207,6 +2210,7 @@ export interface SendBoxStylesProps extends BaseCustomStyles {
 
 // @public
 export interface StatefulCallClient extends CallClient {
+    createCallAgent(...args: Parameters<CallClient['createCallAgent']>): Promise<DeclarativeCallAgent>;
```
PorterNan, line 2213: We don't need a new api here - it is declarative behavior so it should be provided by CallClient(SDK one), and you also have type DeclarativeCallAgent = CallAgent, so just leave the createCallAgent one, and it should be coming from CallClient directly

PorterNan: Noticed things are slightly different on beta version... Feel free to close this one

packages/calling-stateful-client/src/IncomingCallDeclarative.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  constructor(context: CallContext) {
+    this._context = context;
+  }
+
+  public get<P extends keyof IncomingCall>(target: IncomingCall, prop: P): any {
+    switch (prop) {
+      case 'accept': {
+        return this._context.withAsyncErrorTeedToState(async function (...args: Parameters<IncomingCall['accept']>) {
+          return await target.accept(...args);
```
PorterNan, line 27: Do we do anything more than withAsyncErrorTeedToState? Looks like we are not making any state changes when accept or reject

PorterNan:
I like the overall implementations! I would like to bring one point up to discussion:
1. Do have incoming call plain object in our centralized state, and does that contains all information like call id and group id - is it possible to use callAgent.join function to join the imcoming call? I am thinking of a way that doesn't require an api change, it would be nice if we can treat incoming calls like normal call object
2. If we can't do #1, which means an api change is definitely required - let's do it on the top level, which means on callClient level - the principle here is we only add additional apis on the top level so we will never need to manipulate the returning type of its children - for example, here we don't have to add a different type for callAgent because we have everything in callClient - createView is a good example, that is something in call object, but we don't proxy that too deep inside of the api trees

PorterNan:
> 1. Do have incoming call plain object in our centralized state, and does that contains all information like call id and group id - is it possible to use callAgent.join function to join the imcoming call? I am thinking of a way that doesn't require an api change, it would be nice if we can treat incoming calls like normal call object

For point 1, it's more like a sdk api problem - we now have 3+ way to join a call - callAgent.start, callAgent.join, incomingCall.accept... I hope there is a consistent way to do this(for this PR and all the future api from our SDK), but if we can't make this happen, this is the sdk's issue

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1397,6 +1398,14 @@ export interface CustomMessage extends MessageCommon {
 // @public
 export const darkTheme: PartialTheme & CallingTheme;
 
+// @public
+export type DeclarativeCallAgent = CallAgent & /* @conditional-compile-remove(one-to-n-calling) */ {
+    getIncomingCall: (incomingCallId: string) => DeclarativeIncomingCall;
```
PorterNan, line 1403: We can move this one to the top StatefulCallClient api, so DeclarativeCallAgent can just share the same set of api callAgent has - which is the principal of centralize all states and all additional apis in one place

anjulgarg:
> I like the overall implementations! I would like to bring one point up to discussion:
> 
> 1. Do have incoming call plain object in our centralized state, and does that contains all information like call id and group id - is it possible to use callAgent.join function to join the imcoming call? I am thinking of a way that doesn't require an api change, it would be nice if we can treat incoming calls like normal call object
> 2. If we can't do point 1, which means an api change is definitely required - let's do it on the top level, which means on callClient level - the principle here is we only add additional apis on the top level so we will never need to manipulate the returning type of its children - for example, here we don't have to add a different type for callAgent because we have everything in callClient - createView is a good example, that is something in call object, but we don't proxy that too deep inside of the api trees

Thanks for your feedback @PorterNan 

1. Is not doable unfortunately, 1:1 calls are peer to peer calls and can not be joined like regular calls that have a groupId. 
2. The concern I have with this approach is that it allows Contoso to use `getIncomingCall` even before initializing the `callAgent`. We would have to throw an error in that case which makes for a non-intuitive experience.

prprabhu-ms:
> > I like the overall implementations! I would like to bring one point up to discussion:
> > 
> > 1. Do have incoming call plain object in our centralized state, and does that contains all information like call id and group id - is it possible to use callAgent.join function to join the imcoming call? I am thinking of a way that doesn't require an api change, it would be nice if we can treat incoming calls like normal call object
> > 2. If we can't do point 1, which means an api change is definitely required - let's do it on the top level, which means on callClient level - the principle here is we only add additional apis on the top level so we will never need to manipulate the returning type of its children - for example, here we don't have to add a different type for callAgent because we have everything in callClient - createView is a good example, that is something in call object, but we don't proxy that too deep inside of the api trees
> 
> Thanks for your feedback @PorterNan
> 
> 1. Is not doable unfortunately, 1:1 calls are peer to peer calls and can not be joined like regular calls that have a groupId.
> 2. The concern I have with this approach is that it allows Contoso to use `getIncomingCall` even before initializing the `callAgent`. We would have to throw an error in that case which makes for a non-intuitive experience.

Let's bring this up IRL with the team / or as a team thread for higher bandwidth discussion.

I am in favor of the approach in this PR. It doesn't not introduce a new way to join a call that departs from the headless layer. And it keeps the methods in the same layer as the headless API (does not hoist the method from call agent to client).

PR #1978, opened 2022-06-08
Fixed bug when join without display name

carocao-msft:
# What
When participant join call without displayname, set displayname to unnamed participant

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2835967

# How Tested
Tested with Donald's PTSN sample


packages/calling-component-bindings/src/callUtils.ts
```diff
@@ -71,6 +71,11 @@ const memoizedUpdateDisplayName = memoizeFnAll((participantId: string, participa
       ...participant,
       displayName: participant.identifier.phoneNumber
     };
+  } else if (participant.displayName === undefined || participant.displayName === '') {
+    return {
+      ...participant,
+      displayName: 'Unnamed Participant'
```
alkwa-msft, line 77: WE should be using a locale string instead of a hard coded string. Consider french do we wan to show "Unnamed Participant"?

JamesBurnside: I like this logic better in the reacct-components layer also as it allows Contoso to decide on what behavior they want if someone doesn't have a display name. If we set a display name here they won't know if someone is actually called Unnamed Participant or if the display name was unset. But it will mean adding to the participant list and to the video gallery (I see the video gallery shows no display name also)

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -59,6 +59,8 @@ export interface ParticipantItemStrings {
   sharingIconLabel: string;
   /** Label for the muted icon in participant state stack  */
   mutedIconLabel: string;
+  /** placeholder text for participants who does not have a display name*/
+  displayNamePlaceholder: string
```
dmceachernmsft, line 63: Shouldn't this be on the `ParticipantListStrings` type?

dmceachernmsft: I see why disregard this comment.

dmceachernmsft (approved):

packages/calling-component-bindings/src/callUtils.ts
carocao-msft: Updated :)

packages/react-components/src/components/VideoGallery/LocalScreenShare.tsx
```diff
@@ -48,12 +48,13 @@ export const LocalScreenShare = React.memo((props: { localParticipant: VideoGall
     </Stack>
   );
 
+  const displayName =
+    localParticipant?.displayName === '' || !localParticipant?.displayName
```
JamesBurnside, line 52: nit, this can just be `!localParticipant?.displayName` (no need for empty string check)
```suggestion
  !localParticipant?.displayName
```

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -94,7 +94,7 @@ export const participantListSelector: ParticipantListSelector = createSelector(
       : [];
     participants.push({
       userId: userId,
-      displayName: displayName,
+      displayName: displayName ?? '',
```
JamesBurnside, line 97: Was this needed?

JamesBurnside: Ideally stateful can propagate undefined safely to signify that there was no display name set

JamesBurnside (approved):

packages/calling-component-bindings/src/participantListSelector.ts
carocao-msft: It is needed so that unnamed participants are actually pushed into the participants array

PR #1979, opened 2022-06-08
Update fluentui/react version to 8.62.4

carocao-msft:
# What
Update fluentui/react version to [8.62.4](https://github.com/microsoft/fluentui/tree/@fluentui/react_v8.62.4)

# Why
We want to disable the tooltip on Persona component and the option to disable the tooltip is only introduced in a newer version


alkwa-msft:
Is there a specific reason why we are updating this now?

JamesBurnside (approved):

PR #1980, opened 2022-06-09
Participant list selector Memoization.

dmceachernmsft:
# What

Memoizes the participant list selector
# Why

Helps with performance in the participant list selector

https://skype.visualstudio.com/SPOOL/_workitems/edit/2875585
# How Tested

Ran locally CallWithChat sample to verify that the participant list works.

### DIFF
--- a/packages/calling-component-bindings/src/participantListSelectorUtils.ts
+++ b/packages/calling-component-bindings/src/participantListSelectorUtils.ts
@@ -1,0 +1,52 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { CommunicationIdentifier, getIdentifierKind } from '@azure/communication-common';
+import { memoizeFnAll, toFlatCommunicationIdentifier } from '@internal/acs-ui-common';
+import { CallParticipantListParticipant } from '@internal/react-components';
+
+/**
+ * @private
+ */
+export const memoizedConvertAllremoteParticipants = memoizeFnAll(
+  (
+    userIdentifier: CommunicationIdentifier,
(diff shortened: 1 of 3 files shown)


packages/calling-component-bindings/src/participantListSelectorUtils.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation.
```
PorterNan, line 1: Since we have multiple utils files(another is callUtils), let's move all of these into utils folder, to make the top folder cleaner

packages/calling-component-bindings/src/participantListSelectorUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import { CommunicationIdentifier, getIdentifierKind } from '@azure/communication-common';
+import { memoizeFnAll, toFlatCommunicationIdentifier } from '@internal/acs-ui-common';
+import { CallParticipantListParticipant } from '@internal/react-components';
+
+/**
+ * @private
+ */
+export const memoizedConvertAllremoteParticipants = memoizeFnAll(
+  (
```
PorterNan, line 12: To make it safe, the first param of the memoizeFunction should be key(the key to find the right memoized instance), and the key is flatten parameter like string or number (Object is okay, but we will never know when it updates)

Let's flatten the identifier first then use it as the first parameter outside of the function

PorterNan (approved):
Nice, I can see we are getting into a better situation of perf optimization

carocao-msft (approved):

PR #1981, opened 2022-06-09
Add detection to only permit a single instance of calling on mobile browsers

JamesBurnside:
# What
* Add to CallWithChatSample only allow one browser instance open at a time
* Add same snippet to our best practices in our storybook

# Why
Calling best practice: https://docs.microsoft.com/en-us/azure/communication-services/concepts/best-practices#handling-multiple-calls-on-multiple-tabs-on-mobile

# How Tested
image

Alos updated the unsupported browser page:
|before | after|
| --| --|
| image | image |


alkwa-msft (approved):

samples/CallWithChat/src/app/utils/useIsMobile.ts
```diff
@@ -0,0 +1,28 @@
...
+import { useEffect, useState } from 'react';
+import MobileDetect from 'mobile-detect';
+
+const detectMobileSession = (): boolean => !!new MobileDetect(window.navigator.userAgent).mobile();
+
+/**
+ * Detect if the app is running in a mobile browser and updated if that changes (e.g. when switching to a mobile browser in a browser emulator)
+ * @returns true if this detects the app is running in a mobile browser
+ */
+export const useIsMobile = (): boolean => {
```
edwardlee-msft, line 13: Very nice! Makes sense to have this as its own util.
Moving forward, it would still make sense to carry down the prop "isMobile" rather than utilizing this util in children elements right?

JamesBurnside: We definitely shouldn't use this in the composites - so from that perspective it should be a passed down prop.
On the app side its up to you - there's pros and cons. I like "prop blasting" because then there is an interface where if you supply these props you get an exact UI. However prop blasting leads to hard-to-refactor code and overly verbose code... so up to you!

PR #1982, opened 2022-06-10
Expose Device change events in the Calling and CallWithChat adapters

dmceachernmsft:
# What

Update device change events in Calling and CallWithChat Adapters
- 'selectedCameraChanged'
- 'selectedMicrophoneChanged'
- 'selectedSpeakerChanged'
# Why

Exposes these events so that contoso can perform actions on them

https://skype.visualstudio.com/SPOOL/_workitems/edit/2842399
# How Tested

Ran locally with these outputs in CallWithChat sample 
image


anjulgarg:
Do we need to expose new events?
I thought `onStateChange` event listener would suffice
image

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -327,18 +330,21 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
     return await this.asyncTeeErrorToEventEmitter(async () => {
       await this.handlers.onSelectCamera(device, options);
+      this.selectedCameraChanged();
     });
   }
 
   public async setMicrophone(device: AudioDeviceInfo): Promise<void> {
     return await this.asyncTeeErrorToEventEmitter(async () => {
       await this.handlers.onSelectMicrophone(device);
+      this.selectedMicrophoneChanged();
```
anjulgarg, line 340: Looks like this would be called even when the device hasn't actually changed?
For example, if someone calls this function with the old device again?

edwardlee-msft (approved):
nit: should we remove selectedCameraChanged if it is not available at the moment by the Calling team?

anjulgarg (approved):

PR #1984, opened 2022-06-13
Fix flaky scrollbar in FileSharing UI tests

mgamis-msft:
# What

Added html, body, and .Root styles that apply width and height of 100% to UI test app for Chat just like in our sample apps. This avoid scrollbars outside of composites that cause transient scroll bars when new messages don't fit the browser screen.

I got a build error on storybook shown in this log https://github.com/Azure/communication-ui-library/runs/6856612900?check_suite_focus=true. So I ran `rush add -p browserslist@latest on packages/storybook`

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2883029

# How Tested

Tested locally
https://user-images.githubusercontent.com/79475487/173290869-d38714ad-a4cc-48eb-8fe9-4d9bdbb2f081.mp4

# Process & policy checklist


packages/storybook/package.json
```diff
@@ -37,7 +37,8 @@
     "copy-to-clipboard": "~3.3.1",
     "resize-observer-polyfill": "~1.5.1",
     "tslib": "~2.3.1",
-    "uuid": "^8.1.0"
+    "uuid": "^8.1.0",
+    "browserslist": "~4.20.4"
```
JamesBurnside, line 41: nit: alphabetize

packages/react-composites/tests/browser/chat/app/index.css
```diff
@@ -0,0 +1,13 @@
+/* Copyright (c) Microsoft Corporation. */
+/* Licensed under the MIT license. */
+
+body,
+html,
+.Root {
+  max-width: 100%;
```
JamesBurnside, line 7: Still seeing the chat composite going beyond screen height:
image

Instead of `max-height` try just `height`? My guess is because the chat composite sizes itself to 100% of the avilable space of the parent, however html, body and root don't actually have a height so they are considered height: 0.
If that doesn't work feel free to wrap the chat composite in a container div sized to `100vh`

dmceachernmsft:
image
there is a interesting timing issue with these messages being sent. is this something to do with the fake adapter?

PR #1985, opened 2022-06-13
Add ui tests for isReceiving video loading spinners

JamesBurnside:
# What
* Add UI tests for #1954 
  * Added ability to show loading spinner in the stable screenshots util (as it should be hidden for live service tests)
  * Added mock service ui tests that set isReceiving to false to test loading spinner
* Needed to update a bunch of screenshots to use stableScreenshot util

# Why
Regression tests

# How Tested
See PR for UI snapshot updates


chriswhilar:
image

So i dont think i tested this out but i was thinking that this loading spinner should be smaller.

JamesBurnside:
> So i dont think i tested this out but i was thinking that this loading spinner should be smaller.

Yeah noticed that also, we'll need to update to be max-width: 5rem and width: 100% and then appropriate padding so it will try to size to 5rem, but be smaller when places in a smaller container size (lets keep that change to a seperate PR though)

chriswhilar (approved):

anjulgarg:
Approving this. Hoping there is a fast follow up PR for reducing the size of loader.

PR #1990, opened 2022-06-14
Disable tooltip for persona component

carocao-msft:
# What
Disable tooltip for persona component
For participant lists only remove tooltip when on mobile

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2844028

# How Tested
checked 3 samples and storybook


JamesBurnside (approved):

edwardlee-msft (commented):
If we remove a persona tooltip from being accessible, how would a user be able to see the full name of a user if the name is super long?

packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -81,6 +81,8 @@ export type ParticipantListProps = {
   onParticipantClick?: (participant?: ParticipantListParticipant) => void;
   /** Styles for the {@link ParticipantList} */
   styles?: ParticipantListStyles;
+  /** True if using mobile */
+  isMobile?: boolean;
```
JamesBurnside, line 85: We should always avoid isMobile flag on react-component (ideally it doesn't exist anywhere but we don't allow fine-grained customizations at the composite level).
Instead make this prop `showOverflowTooltip` (same with participant button only it will be slightly different and something like `showParticipantOverflowTooltip`)
The composite level can then do `showOverflowTooltip={props.isMobile}`

JamesBurnside (changes requested):

packages/react-components/src/components/ParticipantList.tsx
JamesBurnside: For more context, the reason we avoid it is we want props to be extremely explicit about their intent and to only control one thing. If Contoso sets isMobile to true anything could happen inside the component

carocao-msft: Updated :)

JamesBurnside (approved):

edwardlee-msft (approved):

PR #1991, opened 2022-06-14
Add new UFDs for camera stopped unexpectedly and camera recovered

JamesBurnside:
# What
Add two new UFDs based on cameraStoppedUnexpectedly call diagnostic:
* 'callVideoStoppedBySystem',
* 'callVideoRecoveredBySystem'

Also updated the types in the TestCallingState to use the types from the calling sdk package instead of redeclaring subsets

# Why
https://techcommunity.microsoft.com/t5/azure-communication-services/checklist-for-advanced-calling-experiences-in-mobile-web/ba-p/3266312

image

# How Tested
I could not get this to happen locally, but verified with e2e tests


dmceachernmsft (approved):

prprabhu-ms (commented):
Please verify with the headless calling team that this is the intended use of the UFD. The name of the UFD is misleading if that is the case (what's unexpected if the system chose to stop the camera)?

prprabhu-ms (approved):

PR #1993, opened 2022-06-14
Disable microphone and camera buttons when there are no microphones or cameras

JamesBurnside:
# What
* Update camera and mic button selectors to make the buttons disabled when there are no cameras or mic available

# Why
They shouldn't be pressable in those cases - causes odd cannot start microphone errors - ideally we should also provide a tooltip showing why they are disabled but we show an error UFD there are no microphones so that should suffice and avoid a lot of extra logic

# How Tested
See UI tests added.


packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -48,6 +49,8 @@ export class MockCallAdapter implements CallAdapter {
       initialState.latestErrors = testState.latestErrors;
     }
 
+    initialState.devices = merge(initialState.devices, testState.devices);
```
mgamis-msft, line 52: Cool. I did not know fluent/react has a deep merge method.

mgamis-msft (approved):

dmceachernmsft (approved):

packages/calling-component-bindings/src/callControlSelectors.ts
```diff
@@ -43,7 +43,7 @@ export const microphoneButtonSelector: MicrophoneButtonSelector = reselect.creat
   (callExists, isMuted, deviceManager) => {
     const permission = deviceManager.deviceAccess ? deviceManager.deviceAccess.audio : true;
     return {
-      disabled: !callExists || !permission,
+      disabled: !callExists || !permission || !deviceManager.microphones.length,
```
anjulgarg, line 46: I am assuming the drop down still works allowing users to change their speakers?

anjulgarg (approved):

packages/calling-component-bindings/src/callControlSelectors.ts
anjulgarg, line 46: does this still allow the speakers to be changes using the microphone dropdown? (callwithchat)

JamesBurnside: Great catch! Updated: 
image

PR #1997, opened 2022-06-15
Try see if createAdapter retries help e2e live test flakiness

JamesBurnside:
# What
Quick PR to keep trying to create an adapter, instead of trying once and failing the test suite.

# Why
Our e2e tests fail frequently and are stuck in the `initializing adapter` page

## Why might this change be bad
Putting as draft as not sure we want this.
We don't actually know why the create adapter failed, is it network on the build machine? is the service throttling us?
If its the former this change won't fix anything. If its the latter we should really have a discussion with the service layer to see how we could not be throttled.
This masks the problem instead of fixing it.

# How Tested
CI will test.


JamesBurnside:
Sadly won't work:
image

PR #2000, opened 2022-06-15
Fix UFD messages incorrect persisting when issue was resolved

JamesBurnside:
# What
Ensure we are checking the value is `true` when creating UFD messages

# Why
If I unplugged my mic the UFD would correctly show, if I plugged my mic back in the UFD would persist when it should have vanished. This is because the diagnostics object looked like this:
image

i.e. the noMicrophoneDevicesEnumerated property does exist in the latest media diagnostics - but it is now set to `false`. We were only checking if the property existed and not if it was explicitly raised to `true`

This matches the documentation for the properties:
image

# How Tested
Plugging mic back in now removes the UFD


prprabhu-ms (approved):
Mea culpa...

edwardlee-msft (approved):

PR #2001, opened 2022-06-16
Error bar UI tests converted to use fake adapter

mgamis-msft:
# What

 ErrorBar UI tests converted to use fake adapter instead of real adapters

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2845227

# How Tested

Tested locally. Local run time comparison: 

# Process & policy checklist


packages/react-composites/tests/browser/chat/app/FakeChatAdapterArgs.ts
```diff
@@ -58,4 +58,10 @@ export type FakeChatAdapterArgs = {
    * Determines if chat composite will be using a custom data model
    */
   customDataModelEnabled?: boolean;
+  /*
+   * Record of rest errors to throw when functions of ChatThreadClient are called
+   */
+  chatThreadClientFunctionErrors?: Partial<Record<keyof ChatThreadClient, ChatThreadRestError>>;
```
prprabhu-ms, line 64: Naming nit: `chatThreadClientMethodErrors`. An interface / class has methods, not functions.

packages/react-composites/tests/browser/chat/fake-adapter/ErrorBar.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    );
+    await waitForChatCompositeToLoad(page);
+    await stubMessageTimestamps(page);
+    expect(await page.screenshot()).toMatchSnapshot('error-bar-wrong-thread-id.png');
+
+    await sendMessage(page, TEST_MESSAGE);
+    await waitForSendMessageFailure(page);
+    await stubMessageTimestamps(page);
+    expect(await page.screenshot()).toMatchSnapshot('error-bar-send-message-with-wrong-thread-id.png');
+    // test resend button in contextual menu
```
prprabhu-ms, line 31: Tangent: You didn't add this check in this PR, so I agree you should carry it over.

But we should split off tests for resend button from tests for showing the correct error bar. Especially with the speed up we'll get with the fake services, we can push to make our tests be more specific.

packages/react-composites/tests/browser/chat/fake-adapter/ErrorBar.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const TEST_MESSAGE = 'No, sir, this will not do.';
+
+test.describe('ErrorBar is shown correctly', async () => {
+  test('with wrong thread ID', async ({ page, serverUrl }) => {
+    await page.goto(
+      buildUrlForChatAppUsingFakeAdapter(serverUrl, {
+        ...DEFAULT_FAKE_CHAT_ADAPTER_ARGS,
+        chatThreadClientFunctionErrors: {
+          getProperties: { message: 'Could not get properties', statusCode: 400 },
```
prprabhu-ms, line 17: Did you check these return codes by running the real test?

prprabhu-ms (approved):

packages/react-composites/tests/browser/chat/fake-adapter/ErrorBar.test.ts
mgamis-msft: Yes I did.

dmceachernmsft (approved):

PR #2004, opened 2022-06-19
Fix SendBox button when using VoiceOver on iOS

mgamis-msft:
# What

Fix SendBox button when using VoiceOver on iOS

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2791680

# How Tested

Tested while using VoiceOver on IPad by accessing app through ngrok tunnel. 
Also tested that SendBox button works as before on: 
- IPad without VoiceOver
- Desktop

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/InputBoxComponent.tsx
+++ b/packages/react-components/src/components/InputBoxComponent.tsx
@@ -23,7 +23,8 @@ import {
   inlineButtonsContainerStyle,
   newLineButtonsContainerStyle,
   inputBoxNewLineSpaceAffordance,
-  inputButtonTooltipStyle
+  inputButtonTooltipStyle,
+  iconWrapperStyle
 } from './styles/InputBoxComponent.style';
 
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -192,8 +193,9 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
         onMouseLeave={() => {
           setIsHover(false);
         }}
-        onRenderIcon={() => onRenderIcon(isHover)}
       />
```
dmceachernmsft, line 196: Maybe update this to a div since it isn't behaving like a IconButton (possibly try a default button as well)

mgamis-msft: Thanks. Went with DefaultButton because it has the ariaLabel prop.

dmceachernmsft (approved):

edwardlee-msft (approved):

PR #2006, opened 2022-06-21
More workers for fake adapter UI tests

mgamis-msft:
# What

Moved live tests using real adapters to live-tests folder

# Why

Lower duration of Chat UI tests which reduces total CI time

# How Tested

Testing on CI

# Process & policy checklist


packages/react-composites/tests/browser/chat/live-tests/ChatComposite.test.ts
```diff
@@ -1,10 +1,15 @@
...
 import { expect } from '@playwright/test';
-import { chatTestSetup, sendMessage, waitForMessageSeen, waitForMessageWithContent } from '../common/chatTestHelpers';
+import {
+  chatTestSetup,
+  sendMessage,
+  waitForMessageSeen,
+  waitForMessageWithContent
+} from '../../common/chatTestHelpers';
+import { stubMessageTimestamps, waitForChatCompositeToLoad } from '../../common/utils';
+import { test } from '../fixture';
```
mgamis-msft, line 13: The tests below: 
- on line 14 is just to load the composite with a live adapter to make sure our composite works with the real adapter
- on line 18 cannot be converted to fake adapters as refreshing the page to rejoin the chat would lose the state of the chat because the state is maintained in browser memory

Please let me know if either these tests should removed from our CI

packages/react-composites/tests/server/server.ts
```diff
@@ -10,20 +10,21 @@ const app = express();
 export const createTestServer =
   (props: { appDir: string; serverUrl: string }) =>
   // eslint-disable-next-line no-empty-pattern, @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/explicit-function-return-type
-  async ({}, use: (r: string) => Promise<void>) => {
-    await startServer(props.appDir);
+  async ({}, use: (r: string) => Promise<void>, workerInfo) => {
+    const port = 3000 + workerInfo.workerIndex;
```
JamesBurnside, line 14: This is a neat fix for this issue where each worker spawns their own server! Can you file a bug however to have a single server that runs (playwright has support for this: https://playwright.dev/docs/test-advanced#launching-a-development-web-server-during-the-tests but its complicated for us as we have three different apps we would need to route traffic between)

packages/react-composites/tests/browser/chat/live-tests/ChatComposite.test.ts
JamesBurnside: Lets remove in a seperate PR if we find them causing issues

JamesBurnside (approved):

packages/react-composites/tests/server/server.ts
mgamis-msft: Good point. Created this bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2914372.

prprabhu-ms (approved):

PR #2013, opened 2022-06-22
Dmceachernmsft/custom data modelfix

dmceachernmsft:
# What

Update the participant list to reflect the composite displayName passed in. 
# Why

Brings composite onFetchAvatarPersonaData and displayName custom data model points in line with our documentation

https://skype.visualstudio.com/SPOOL/_workitems/edit/2833447
# How Tested

image
image
Ran locally with onFetchAvatarPersonaData that sets the `text` field and passed in a display name to make sure the displayNames reflected in the right places.


packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -84,3 +88,9 @@ export const ParticipantListWithHeading = (props: {
     </Stack>
   );
 };
+
+const displayNameStyle: ITextStyles = {
```
PorterNan, line 92: Move this to participantContainer.style.ts

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -67,11 +66,17 @@ export interface AvatarPersonaProps extends IPersonaProps {
...
-  }, [userId]);
 
-  const [data] = useCustomAvatarPersonaData(userIds, dataProvider);
+  const [data, setData] = useState<AvatarPersonaData | undefined>();
+
+  useEffect(() => {
+    (async () => {
+      if (dataProvider && userId) {
+        const data = await dataProvider(userId);
+        setData(data);
```
PorterNan, line 76: Would be nice to do a deep compare of data object, if every properties in the data object is the same, then skip setData() function

PorterNan (approved):
Thanks for making this change! 
I can see all those efforts we made when we were trying to fix the displayName bug in another way, I do love the perf improvement learnings we had there and some beautiful hooks we wrote, and I am glad that we open a branch to save them!
And I am glad to see we are brave enough to take a step back, considering the purpose of some changes we made and choose the right way that aligns with our initial concept, excellent work!

mgamis-msft:
Not part of this PR but the chat participant pane in some of the snapshots look strange:
image
image
Logged this bug: https://skype.visualstudio.com/SPOOL/_workitems/edit/2903915

mgamis-msft (approved):

PR #2015, opened 2022-06-23
Dmceachernmsft/1.3.0 arb cherry pick

dmceachernmsft:
# What

Fix comments from ARB
# Why

Allows for releasing 1.3.0

PR #2016, opened 2022-06-23
1.3.0 localization strings

dmceachernmsft:
# What

Updates localization strings for 1.3.0
# Why

Update locale strings added in 1.3.0

### DIFF
--- a/packages/react-composites/src/composites/localization/locales/de-DE/strings.json
+++ b/packages/react-composites/src/composites/localization/locales/de-DE/strings.json
@@ -44,11 +44,13 @@
   },
   "callWithChat": {
     "peopleButtonLabel": "Personen",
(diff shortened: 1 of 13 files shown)


prprabhu-ms (approved):
Note that you have changed the name of some of the String fields in https://github.com/Azure/communication-ui-library/pull/2015/files#diff-0724208fc79925e04489dc1226e47e5918655ca40cb7dfc0e893a9e06eb56e05

This means that you'll need to merge that PR and *then* fetch the strings again, making sure you get the correctly translated strings.

I'd encourage merging this PR, and then fetching after the field name change PR is merged. You should then get a diff that just moves the translated values around (if you see the translated string values being changed, it'd likely mean that you're not getting the latest translations...)

PR #2017, opened 2022-06-23
Update Icons in Custom buttons to use iconName strings from registered Icons.

edwardlee-msft:
# What
Custom buttons now take in strings that represent an iconName from a registered Icon.

# Why
Previously implementation was limiting Contoso to composite only icons with preset iconNames. Now Contoso can inject and name any icons to be used with a custom button.

For example: Contoso can now use any icons provided by fluentui and assign it to a name for a Custom Button.

```typescript
import { DEFAULT_COMPONENT_ICONS } from '@azure/communication-react';
import { Laptop20Filled } from '@fluentui/react-icons';
const customIcons = {...DEFAULT_COMPONENT_ICONS, 
  CustomFeedbackIcon: <Laptop20Filled />,
  CustomSecondaryIcon: <Laptop20Filled />,
  CustomInfoIcon: <Laptop20Filled />
};
registerIcons({ icons: customIcons });
```
  

# How Tested
Tested locally on chrome MacOS

# Process & policy checklist


JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/buttons/Custom.tsx
```diff
@@ -48,7 +48,7 @@ export const generateCustomControlBarButtons = (
               styles={buttonProps.styles}
               key={`${buttonProps.placement}_${i}`}
               // set default icon if not provided
-              onRenderIcon={() => <CallCompositeIcon iconName={buttonProps.iconName ?? 'ControlButtonOptions'} />}
+              onRenderIcon={() => <CallCompositeIcon iconName={buttonProps.iconName ?? ''} />}
```
dmceachernmsft, line 51: Why is this a empty string here now? shouldn't it default to something? or is this so that it doesn't render a icon with a empty prop?

packages/react-composites/src/composites/common/icons.tsx
```diff
@@ -83,7 +83,7 @@ export const DEFAULT_COMPOSITE_ICONS = {
 };
 
 /** @private */
-export type CompositeIconProps<Icons> = IIconProps & { iconName: keyof Icons };
+export type CompositeIconProps<Icons> = IIconProps | { iconName: keyof Icons };
```
dmceachernmsft, line 86: Changing this to a OR union is to make it so contoso can provide any string here correct?

JamesBurnside (approved):

dmceachernmsft (approved):

PR #2023, opened 2022-06-27
Fix for flaky file sharing upload cards UI tests 

prabhjot-msft:
# What
Added wait function for rendering file type icons

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2876692

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -402,3 +407,21 @@ const enableTooltips = async (page: Page): Promise<void> => {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (args: any) => {
+      const iconNodes = Array.from(document.querySelectorAll(args.fileTypeIconId));
+      return iconNodes.every((node) => node?.querySelector('img').height > 0);
+    },
+    {
+      fileTypeIconId: fileTypeIconId
+    }
+  );
+};
```
mgamis-msft, line 427: This PR looks good to me. Just wondering if we should go one step further and query all icon tags and wait for all height is > 0 in the `stableScreenshot` function. Thoughts @JamesBurnside @dmceachernmsft? I know we register other icons so they load faster but the file type icon is determined dynamically (using function `getFileTypeIconProps` from '@fluentui/react-file-type-icons') and can't be registered.

mgamis-msft (approved):

change/@internal-react-components-cbda6e4f-8622-4f48-bcc9-42d5fe413676.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "patch",
+  "comment": {
```
dmceachernmsft, line 3: I think you missed adding a comment here?

packages/react-composites/tests/browser/common/utils.ts
dmceachernmsft: this to me seems tricky, what if a Icon isn't in a snapshot but we are waiting for it in the stable screenshot method? seems to me that this might be the way to do it with a series of helper functions when we run into a icon like this we can't register?

mgamis-msft: It should be fine for snapshots without icons since we can return true if iconNodes is empty after running:
`const iconNodes = Array.from(document.querySelectorAll('i');`
But I'll just resolve this thought as it is outside the scope of this ticket. I'll just keep this in mind the next time we fail UI tests because icons don't load in time.

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -368,6 +370,9 @@ export async function stableScreenshot(
   if (stubOptions?.hideVideoLoadingSpinner !== false) {
     await hideVideoLoadingSpinner(page);
   }
+  if (stubOptions?.awaitFileTypeIcon !== false) {
```
JamesBurnside, line 373: avoid the `!== false` here and just have `if (stubOptions?.awaitFileTypeIcon)`, the reason the hideVideoLoadingSpinner has it is because by default all tests should hide it. Whereas here we want to be "opt in" and only await if a test explicitly wants to

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -402,3 +407,21 @@ const enableTooltips = async (page: Page): Promise<void> => {
...
+ * Helper function for waiting for file type icons.
+ */
+const awaitFileTypeIcon = async (page: Page): Promise<void> => {
+  const fileTypeIconId: string = dataUiId(IDS.fileTypeIcon);
+  await waitForFunction(
+    page,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (args: any) => {
+      const iconNodes = Array.from(document.querySelectorAll(args.fileTypeIconId));
+      return iconNodes.every((node) => node?.querySelector('img').height > 0);
```
JamesBurnside, line 421: I'm not totally convinced by `height > 0`, it's possible the icon exists in the DOM and has a height, but the image has not yet finished being downloaded and rendered. consider `complete` instead: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/complete

packages/react-composites/tests/browser/common/utils.ts
JamesBurnside: Another option would be to mock the `@fluentui/react-file-type-icons` package at build time.. but don't love that...

packages/react-composites/tests/browser/common/utils.ts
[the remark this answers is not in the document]
prabhjot-msft: Node is the wrapper icon div and image is contained within that, so shouldn't we select image from the node?

change/@internal-react-components-cbda6e4f-8622-4f48-bcc9-42d5fe413676.json
prabhjot-msft: done

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -402,3 +407,21 @@ const enableTooltips = async (page: Page): Promise<void> => {
 const hideVideoLoadingSpinner = async (page: Page): Promise<void> => {
   await page.addStyleTag({ content: '[data-ui-id="stream-media-loading-spinner"] {display: none}' });
 };
+
+/**
+ * Helper function for waiting for file type icons.
```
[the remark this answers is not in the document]
prabhjot-msft: done

change/@internal-react-components-cbda6e4f-8622-4f48-bcc9-42d5fe413676.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
prprabhu-ms, line 2: This should have been `none` as the PR has no user-visible changes (nothing changes for Contoso using our library).

PR #2025, opened 2022-06-27
Generate test results artifact for `playwright` tests in CI

prprabhu-ms:
# What

Start generating a JSON test results artifact for playwright tests in CI.

# Why

First step towards tracking runtime / flakiness of these tests.

### DIFF
--- a/packages/react-composites/playwright.config.ts
+++ b/packages/react-composites/playwright.config.ts
@@ -33,7 +33,7 @@ const chromeLaunchOptions = {
   ]
 };
 
-const CI_REPORTERS: ReporterDescription[] = [['github'], ['json', { outputFile: `${OUTPUT_DIR}/e2e-results.json` }]];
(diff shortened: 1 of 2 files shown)


JamesBurnside (approved):
Oh nice!

packages/react-composites/playwright.config.ts
```diff
@@ -32,7 +33,11 @@ const chromeLaunchOptions = {
   ]
 };
 
+const CI_REPORTERS: ReporterDescription[] = [['dot'], ['json', { outputFile: `${OUTPUT_DIR}/e2e-results.json` }]];
```
edwardlee-msft, line 36: For reference for others:
"Dot reporter is very concise - it only produces a single character per successful test run. It is the default on CI and useful where you don't want a lot of output."
https://playwright.dev/docs/test-reporters

edwardlee-msft (approved):

PR #2027, opened 2022-06-28
onFetchAvatarPersonaData doc update

dmceachernmsft:
# What

Updates docs for onFetchAvatarPersonData
# Why

Adds clarity to what the onFetchAvatarPersonaData prop will effect in the composites
# How Tested

is a doc change

### DIFF
--- a/packages/react-composites/src/composites/common/BaseComposite.tsx
+++ b/packages/react-composites/src/composites/common/BaseComposite.tsx
@@ -42,6 +42,9 @@ export interface BaseCompositeProps<TIcons extends Record<string, JSX.Element>> 
   /**
    * A callback function that can be used to provide custom data to Avatars rendered
    * in Composite.
+   *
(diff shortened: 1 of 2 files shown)


packages/react-composites/src/composites/common/BaseComposite.tsx
```diff
@@ -42,6 +42,9 @@ export interface BaseCompositeProps<TIcons extends Record<string, JSX.Element>>
   /**
    * A callback function that can be used to provide custom data to Avatars rendered
    * in Composite.
+   *
+   * This will not affect the displayName prop provided to the composite's adapter. The displayName throughout the composite
```
PorterNan, line 46: "...will not affect the displayName shown in the composite. The displayName throughout the composite will be what is provided to the adapter when the adapter is created."

PorterNan (approved):

prabhjot-msft (approved):

PR #2034, opened 2022-06-29
Alkwa/fix contextmenu with lots of participants

alkwa-msft:
# What
When there are too many participants it can mess up how things scroll when there is a divider. The divider and other elements will overlap the participants in the context menu

# Why
The max-height css property is incorrectly set on the wrong component. We should be setting it on the context menu as opposed to the stack that contains all of the participant elements

Before: 
withoutfix

After:
withFix

# How Tested
Loaded up a UI component in storybook and added some HTML elements 

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/ParticipantsButton.tsx
+++ b/packages/react-components/src/components/ParticipantsButton.tsx
@@ -18,7 +18,7 @@ import {
   ParticipantListStyles,
   ParticipantMenuItemsCallback
 } from './ParticipantList';
-import { defaultParticipantListContainerStyle, participantsButtonMenuPropsStyle } from './styles/ControlBar.styles';
+import { participantsButtonMenuPropsStyle } from './styles/ControlBar.styles';
 import { useLocale } from '../localization';
 import { ControlBarButton, ControlBarButtonProps, ControlBarButtonStyles } from './ControlBarButton';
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/styles/ControlBar.styles.ts
```diff
@@ -159,15 +158,6 @@ export const participantsButtonMenuPropsStyle: Partial<IContextualMenuStyles> =
   }
 };
 
-/**
- * @private
- */
-export const defaultParticipantListContainerStyle: ParticipantListStyles = {
```
alkwa-msft, line 165: This style was only ever used for this one element and it was messing us up.

packages/react-components/src/components/ParticipantsButton.tsx
```diff
@@ -300,6 +300,9 @@ export const ParticipantsButton = (props: ParticipantsButtonProps): JSX.Element
                 maxWidth: '100%'
               }
             },
+            style: {
```
alkwa-msft, line 303: This properly sets the maxHeight of the people context menu to 20rem and when we scroll we won't get odd visual artifacts.

JamesBurnside (approved):

PR #2038, opened 2022-06-29
Bugfix - Incorrect scrollbar in chat composite landscape mobile view

anjulgarg:
# What
This was being caused because the participant pane is now an absolute component and it's parent component was not positioned relatively. Participant pane had a height of 100% which caused it to expand more than it's parents height. Thus showing a scrollbar.

### Before
image

### After
image

# Why
Fix for https://skype.visualstudio.com/SPOOL/_workitems/edit/2903915

# How Tested

# Process & policy checklist


prprabhu-ms (approved):
Is it intentional that the participant pane is entirely gone from some of the screenshots. The prior snapshot looks really weird with the `SendBox` moved up the screen:

image

mgamis-msft (approved):

PR #2043, opened 2022-07-05
[Draft] Overlaying renderElement in VideoTile instead of replacing it with Placeholder

anjulgarg:
# What
Overlaying renderElement in VideoTile instead of replacing it with Placeholder
- Allows us to render elements overlayed on top of the placeholder. For example, `ringing...` text along with the Persona Icon.
image

# Why
This will come in handy for displaying various connection states of a participant during 1:N or PSTN calls without changing any public API

# How Tested

# Process & policy checklist


packages/react-components/src/components/VideoTile.tsx
```diff
@@ -219,24 +228,25 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
-        {isVideoRendered ? (
+        <Stack className={mergeStyles(videoContainerStyles, { zIndex: _VIDEO_TILE_Z_INDEX })}>
+          {onRenderPlaceholder ? (
+            onRenderPlaceholder(userId ?? '', placeholderOptions, DefaultPlaceholder)
+          ) : (
+            <DefaultPlaceholder {...placeholderOptions} />
+          )}
+        </Stack>
+
+        {isVideoRendered && (
```
dmceachernmsft, line 239: Will this allow for us to provide a custom tile overlay for when the user is being called in? the logic of needing the video rendered is confusing me a little

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -219,24 +228,25 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
-        {isVideoRendered ? (
+        <Stack className={mergeStyles(videoContainerStyles, { zIndex: _VIDEO_TILE_Z_INDEX })}>
+          {onRenderPlaceholder ? (
+            onRenderPlaceholder(userId ?? '', placeholderOptions, DefaultPlaceholder)
+          ) : (
+            <DefaultPlaceholder {...placeholderOptions} />
+          )}
+        </Stack>
+
+        {isRenderElementAvailable && (
```
edwardlee-msft, line 239: A render element is a "Component with the video stream." 

I'm voicing my thoughts here, but my understanding is:
In this case, for ringing a PSTN user, no video stream would be passed but a component with some state indicating if it is ringing would be rendered. 

While for calling an ACS user, it would be ringing, then when they pick up the call with their camera on a video stream would be rendered.

If this is the case, should we update the comment on line 64 for renderElement as a video stream is not always guaranteed? And the fact that you renamed isVideoRendered to isRenderElementAvailable leads me to believe this is the case.

edwardlee-msft (commented):
A render element is a "Component with the video stream."

I'm voicing my thoughts here, but my understanding is:
In this case, for ringing a PSTN user, no video stream would be passed but a component with some state indicating if it is ringing would be rendered.

While for calling an ACS user, it would be ringing, then when they pick up the call with their camera on a video stream would be rendered.

If this is the case, should we update the comment on line 64 for renderElement as a video stream is not always guaranteed? And the fact that you renamed isVideoRendered to isRenderElementAvailable leads me to believe this is the case.

packages/react-components/src/components/VideoTile.tsx
dmceachernmsft: yup! so its actually based on whether there is a video there or our new state tile. is that right?

anjulgarg: So it shouldn't matter,
as long as the renderElement prop has been set with a JSX element, this piece should render

anjulgarg:
> A render element is a "Component with the video stream."
> 
> I'm voicing my thoughts here, but my understanding is: In this case, for ringing a PSTN user, no video stream would be passed but a component with some state indicating if it is ringing would be rendered.
> 
> While for calling an ACS user, it would be ringing, then when they pick up the call with their camera on a video stream would be rendered.
> 
> If this is the case, should we update the comment on line 64 for renderElement as a video stream is not always guaranteed? And the fact that you renamed isVideoRendered to isRenderElementAvailable leads me to believe this is the case.

Good point. Updated the doc string. LMK if it works

packages/react-components/src/components/VideoTile.tsx
anjulgarg: Correct, as long as the prop has a JSX element.

edwardlee-msft (approved):

dmceachernmsft (approved):

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -52,7 +73,12 @@ export interface VideoTileProps {
   styles?: VideoTileStylesProps;
   /** user id for the VideoTile placeholder. */
   userId?: string;
-  /** Component with the video stream. */
+  /**
```
prprabhu-ms, line 76: I don't think we should do this, for a few reasons -

* This is a breaking change with respect to behavior (though not the TypeScript API). If Contoso was using a transparent injected tile before this change, their app will break.
* It is surprising that this `render*` method behaves differently from all the other `render` methods. Granted this is named `renderElement` as opposed to `onRender*` and is intended to be *always* used as opposed to overriding interanlly rendered UI. Still, this is surprising behavior in the context of our library.
* Most importantly though, the same effect should be achieavable by Contoso without the changed behavior: `renderElement` is just a `JSX.Element`, so as long as we provide the building blocks for rendering the placeholder, Contoso can overlay the video and provide that here. We already support `onRenderPlaceholder`, so we should anyway include a way for Contoso to easily render placeholder (i.e., another public component in the UI library). This case is easy as the placeholder is essentially just the `Persona` component.

prprabhu-ms:
I'm not sure this PR is needed for the intended purpose: 

- For the transitional PSTN / call connecting states, there is never a video element. So we just need to provide an apporpriate `onRenderPlaceholder` prop that adds the overlayed text.
- We should support the transitional states in the component layer properly. Adding overlayed strings from the composite increases the cliff from composite -> component.

So I'd push us to properly add support for the transitional call states in the component layer.

prprabhu-ms (changes requested):
Blocking the PR because I would like to clarify the need for this PR, and also because it is a breaking change.

prprabhu-ms:
Removed from my attention set as there is no action to be taken right now

PR #2056, opened 2022-07-08
Fix PiPiP storybook page with slim wrapper around PiPiP component

mgamis-msft:
# What

Since upgrading storybook to version 6.5.7 in #1972, the PiPiP component storybook page does not load and times out.

I explored redefining the `primaryTile.children` ReactNode sub property to `primaryTile` direct ReactNode property and similarly `secondaryTile.children` ReactNode sub property to `secondaryTile` direct ReactNode property.
This fixes PiPiP storybook page. It seems like ReactNode/JSX.Element sub properties don't work in storybook version 6.5.7

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2918197

# How Tested

See generated chromatic link.

# Process & policy checklist

### DIFF
--- a/change/@internal-storybook-25132f7a-8366-4fe3-9b0a-8dd72b96d04e.json
+++ b/change/@internal-storybook-25132f7a-8366-4fe3-9b0a-8dd72b96d04e.json
@@ -1,7 +1,0 @@
-{
-  "type": "patch",
-  "comment": "Updated code in PictureInPictureInPicture story based on api updates in internal PictureInPictureInPicture component.",
-  "packageName": "@internal/storybook",
-  "email": "79475487+mgamis-msft@users.noreply.github.com",
-  "dependentChangeType": "patch"
-}
--- a/change/@internal-storybook-25132f7a-8366-4fe3-9b0a-8dd72b96d04e.json
+++ b/change/@internal-storybook-25132f7a-8366-4fe3-9b0a-8dd72b96d04e.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Updated code in PictureInPictureInPicture story.",
+  "comment": "Updated code in PictureInPictureInPicture story based on api updates in internal PictureInPictureInPicture component.",
   "packageName": "@internal/storybook",
   "email": "79475487+mgamis-msft@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@internal-storybook-25132f7a-8366-4fe3-9b0a-8dd72b96d04e.json
+++ b/change/@internal-storybook-25132f7a-8366-4fe3-9b0a-8dd72b96d04e.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Updated code in PictureInPictureInPicture story.",
+  "packageName": "@internal/storybook",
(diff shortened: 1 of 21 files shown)


change/@internal-storybook-25132f7a-8366-4fe3-9b0a-8dd72b96d04e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Updated code in PictureInPictureInPicture story.",
```
prprabhu-ms, line 3: nit: The changelogs should be a single line that has enough context to understand what the change did. Imagine being the release train engineer and seeing this description with a 100 other changelog entries. Would you understand enough to know how to summarize the change for CHANGELOG.md

prprabhu-ms:
I'm afraid I don't quite understand why the changed props fix the original issue. I also think that this PR changes the props for the worse. So, I'd like to understand why it is needed better, and document the reason in the code for posterity.

(I am still hopeful that we don't need to change the props, and that we can instead fix some bug in the story itself)

mgamis-msft:
> I'm afraid I don't quite understand why the changed props fix the original issue. I also think that this PR changes the props for the worse. So, I'd like to understand why it is needed better, and document the reason in the code for posterity.
> 
> (I am still hopeful that we don't need to change the props, and that we can instead fix some bug in the story itself)

It seems Object properties with React.ReactNode/JSX.Element property do not work in storybook v6.5,7. Would  the prop change in [commit ](https://github.com/Azure/communication-ui-library/commit/dc4b5735d7bb1f866d99800db2907eeb3b73a32a) be better?

prprabhu-ms:
> > I'm afraid I don't quite understand why the changed props fix the original issue. I also think that this PR changes the props for the worse. So, I'd like to understand why it is needed better, and document the reason in the code for posterity.
> > (I am still hopeful that we don't need to change the props, and that we can instead fix some bug in the story itself)
> 
> It seems Object properties with React.ReactNode/JSX.Element properties like `primaryTile` and `secondaryTile` of PiPiP do not work in storybook v6.5,7. Would the prop change in [commit ](https://github.com/Azure/communication-ui-library/commit/dc4b5735d7bb1f866d99800db2907eeb3b73a32a) be better?

Instead of modifying the component to fit storybook, can we create a slim wrapper component within the story (perhaps with the getTile() hack) and display that?

mgamis-msft:
> > > I'm afraid I don't quite understand why the changed props fix the original issue. I also think that this PR changes the props for the worse. So, I'd like to understand why it is needed better, and document the reason in the code for posterity.
> > > (I am still hopeful that we don't need to change the props, and that we can instead fix some bug in the story itself)
> > 
> > 
> > It seems Object properties with React.ReactNode/JSX.Element properties like `primaryTile` and `secondaryTile` of PiPiP do not work in storybook v6.5,7. Would the prop change in [commit ](https://github.com/Azure/communication-ui-library/commit/dc4b5735d7bb1f866d99800db2907eeb3b73a32a) be better?
> 
> Instead of modifying the component to fit storybook, can we create a slim wrapper component within the story (perhaps with the getTile() hack) and display that?

Yep, that worked. No internal api changes needed. Thanks!

dmceachernmsft (approved):

PR #2057, opened 2022-07-08
Fix compilation error in composite quickstart guide

mgamis-msft:
# What

Added installation of uuid package in composite quickstart guide to avoid customer compilation errors when running `npm run start`

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2918664

# How Tested

Local deploy. See 'Get Started' page under Composites in storybook in publish chromatic site.

# Process & policy checklist


packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
```diff
@@ -70,7 +70,7 @@ export const getDocs: () => JSX.Element = () => {
       </Description>
       <Source code={formFactorSnippet} />
       <MessageBar>
-        Note: Only Protrait mode is supported when the `formFactor` is set to "mobile". Landscape mode is not supported.
+        Note: Only Portrait mode is supported when the `formFactor` is set to "mobile". Landscape mode is not supported.
```
prprabhu-ms, line 73: Thanks for catching this.

I think it's better for mental flow to keep all the package installations in one place. So I'd move this up to the "install the package" section, noting that we will use `uuid` for generating a random groupId later in the quickstart.

That's only one opinion though. Worth asking around in the team.

alkwa-msft (approved):

prabhjot-msft (approved):

PR #2060, opened 2022-07-08
Fixed bug where voice over is not announcing menu item checked 

carocao-msft:
# What
Fixed bug where voice over is not announcing menu item checked 

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2908125

# How Tested

Screen Shot 2022-07-08 at 12 41 57 PM
est your change.


dmceachernmsft (approved):

dmceachernmsft:
Before this is checked in lets setup a time to test this on windows and android. just want to make sure there isn't a regression on the other platforms here

edwardlee-msft:
Can we double check that IOS that it is still voicing out menuItem correctly and not Group?

edwardlee-msft (commented):
Can we double check that IOS that it is still voicing out menuItem correctly and not Group?

edwardlee-msft (approved):

PR #2063, opened 2022-07-12
Call adapter changes to join room

mgamis-msft:
# What

Changed call adapter to be able to join a room with a RoomLocator.
Updated Call composite preview in storybook to be able to parse a Room ID. 

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2907180

# How Tested

Tested on Call composite preview in storybook to be able to join a room as well as a group and Teams meeting like before.

# Process & policy checklist


prabhjot-msft (approved):
looks good!

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -297,15 +299,30 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
+      const isRoomsCall = !('roomId' in this.locator);
+
+      let call;
+
+      if (isTeamsMeeting) {
+        call = this.callAgent.join(this.locator as TeamsMeetingLinkLocator, {
+          audioOptions,
+          videoOptions
+        });
+      } else {
```
PorterNan, line 312: This else will be running even it is Rooms locator, the behavior might be right (cause in the run time callAgent.join points to the same js function), but let's fix this in case that it creates hidden regressions

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -297,15 +299,30 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
-            videoOptions
-          })
-        : this.callAgent.join(this.locator as GroupCallLocator, {
-            audioOptions,
-            videoOptions
-          });
+      /* @conditional-compile-remove(rooms) */
+      const isRoomsCall = !('roomId' in this.locator);
+
+      let call;
```
PorterNan, line 305: Code lines from 302-325 could be wrapped into a function, (maybe name it joinCall?)

Then we can do conditional compilation in the function then:
const call = joinCall(locator, {audioOptions, videoOptions});

mgamis-msft: Sure. Moved to private function.

packages/storybook/stories/CallComposite/snippets/Container.snippet.tsx
```diff
@@ -22,6 +23,16 @@ export type ContainerProps = {
 };
 
 const isTeamsMeetingLink = (link: string): boolean => link.startsWith('https://teams.microsoft.com/l/meetup-join');
+const isGroupLink = (link: string): boolean => link.indexOf('-') !== -1;
```
PorterNan, line 26: link.indexOf('-') !== -1; seems like a little hacky for our customer to check (this snippet would be a public one exposed to our customer, they will have to know the syntax difference from groupId, roomId), is there a way to check it like a function isGuid()?

PorterNan (approved):

packages/storybook/stories/CallComposite/snippets/Container.snippet.tsx
mgamis-msft: Good point. Will polish that in this bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2925571. Thanks.

PR #2067, opened 2022-07-12
Introduce fixture for hermetic CallComposite tests

prprabhu-ms:
# What

Introduce a separate test fixture for the hermetic `CallComposite` tests.

# Why

The dedicated fixture avoids the need to create real Azure Communication Services users. It does not interact with the backend services at all.

# How Tested

Ran hermetic tests locally!


edwardlee-msft:
This was a really fun PR to look over! 
Thanks for assigning me to this one Prathmesh!

edwardlee-msft (approved):

mgamis-msft:
Thanks for splitting the fixtures and tests into hermetic and live folders. Just for my knowledge, is hermetic the right word? Wouldn't changes in Calling SDK interfaces/class (though rare) would cause a mock adapter to change too?

prprabhu-ms:
> Thanks for splitting the fixtures and tests into hermetic and live folders. Just for my knowledge, is hermetic the right word? Wouldn't changes in Calling SDK interfaces/class (though rare) would cause a mock adapter to change too?

These tests are hermetic in that they do not access any environment resources (network / persistent storage etc.) at runtime. So running them in different environments without any credentials etc will give the same results. Contrast with live tests that depend on the backend services to be accessible and stable.

The term "hermetic" Is commonly used in this case: https://blog.testproject.io/2021/09/13/my-recent-obsession-with-hermetic-tests/

PR #2068, opened 2022-07-12
Run CallComposite hermetic e2e tests with specialized config

prprabhu-ms:
Similar to #2064 but for `CallComposite` hermetic tests.

# Why

Run the hermetic tests faster by using more workers.


packages/react-composites/package.json
```diff
@@ -23,11 +23,11 @@
     "test": "rushx _by-flavor \"rushx _test:by-flavor\"",
     "test:coverage": "rushx _by-flavor \"rushx test:_by-flavor -- --coverage\"",
     "test:e2e": "rushx _by-flavor \"rushx test:e2e:call && rushx test:e2e:chat && rushx test:e2e:callwithchat\"",
-    "test:e2e:call": "rushx _by-flavor \"playwright test -c ./playwright.config.live.ts tests/browser/call/\"",
+    "test:e2e:call": "rushx _by-flavor \"playwright test -c ./playwright.config.live.ts tests/browser/call/live/ && playwright test -c ./playwright.config.hemertic.ts tests/browser/call/hermetic/\"",
```
edwardlee-msft, line 26: nit: Should we keep naming consistent?
Chat uses tests/browser/call/**live-tests**/
while 
Call uses tests/browser/call/**live**/

edwardlee-msft (approved):

packages/react-composites/package.json
prprabhu-ms: Yep, that's point#3 on my TODO sticky note :D

prprabhu-ms: As in, change chat to use the "live" vs "hermetic" naming.

Currently, we call hermetic tests in chat "fake-adapter" tests. The naming "hermetic" is better. It tells us *what* the tests are (they are isolated / hermetic tests) as opposed to how we made it that way (by using a fake-adapter). Imagine we replace the fake-adapter with a fake stateful client -- the "hermetic" test stays unchanged as it should.

mgamis-msft (approved):

PR #2069, opened 2022-07-12
Fix menu dismiss when scrolling

PorterNan:
# What

This is the fix for dismissed menu when chat pane is scrolling
https://skype.visualstudio.com/SPOOL/_workitems/edit/2915497/

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/DevicesButton.tsx
+++ b/packages/react-components/src/components/DevicesButton.tsx
@@ -9,7 +9,7 @@ import {
   IContextualMenuStyles,
   merge
 } from '@fluentui/react';
-import React, { useCallback } from 'react';
+import React from 'react';
 import { useLocale } from '../localization';
 import { ControlBarButton, ControlBarButtonProps, ControlBarButtonStyles } from './ControlBarButton';
 import { HighContrastAwareIcon } from './HighContrastAwareIcon';
--- a/packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
+++ b/packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
@@ -22,7 +22,6 @@ import {
   menuItemIncreasedSizeStyles,
   menuSubIconStyleSet
 } from '../styles/ChatMessageComponent.styles';
-import { preventDismissOnEvent } from '../utils/common';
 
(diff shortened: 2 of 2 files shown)


edwardlee-msft (approved):

packages/react-components/src/components/DevicesButton.tsx
```diff
@@ -268,7 +268,8 @@ export const generateDefaultDeviceMenuProps = (
       // A side effect of this workaround is that the context menu stays open when window is resized, and may
       // get detached from original target visually. That bug is preferable to the bug when this value is not set -
       // The Callout (frequently) gets dismissed automatically.
-      preventDismissOnResize: true
+      preventDismissOnResize: true,
+      preventDismissOnScroll: true
```
prprabhu-ms, line 272: Looks like this API is now deprecated: https://developer.microsoft.com/en-us/fluentui#/controls/web/callout#ICalloutProps

image

But also note that using `preventDismissOnEvent` will override `preventDismissOnResize` as well, so you'll have to test to make sure that resize behavior remains correct (e.g., add a remote participant while the participant pane flyout is open).

packages/react-components/src/components/DevicesButton.tsx
```diff
@@ -268,7 +268,8 @@ export const generateDefaultDeviceMenuProps = (
       // A side effect of this workaround is that the context menu stays open when window is resized, and may
       // get detached from original target visually. That bug is preferable to the bug when this value is not set -
       // The Callout (frequently) gets dismissed automatically.
-      preventDismissOnResize: true
+      preventDismissOnResize: true,
```
prprabhu-ms, line 271: Please add a comment explaining why we need to prevent menu dismissal on scroll. Essentially include the context from the bug you're fixing in the comment so that somebody reading the code understands why we set this flag.

prprabhu-ms (approved):
The suggestion for using non-deprecated API is a soft suggestion since we're already using a related deprecated API.

Do give it a try and if it works please use the newer method.

prprabhu-ms:
Also can you check other places where we use `preventDismissOnResize` and also update those if necessary?

prprabhu-ms (approved):
<3

PR #2070, opened 2022-07-12
Add alternativeCallerId to Adpaters and state for outbound PSTN

dmceachernmsft:
# What

Introduce `alternativeCallerId` to the CallAdapter and Stateful layer.
# Why

Adding this property to these areas allows for Contoso to provide the `alternativeCallerId` at adapter creation so that they can leverage it from the adapter state provider. For us this allows us to use this number in our composites to introduce adding participants in the composite using this phone number. 

Note: The phone number is stored in the callAgentState because the phone number is dependent on the callAgent being created to be used. When Contoso asks for the state of the adapter these properties get surfaced as well. 
image

https://skype.visualstudio.com/SPOOL/_workitems/edit/2920717
# How Tested

Built locally and ran Calling and CallWithChat sample applications to make sure that the new property didn't break anything.


carocao-msft (commented):
Wondering if we should have the alternativeCallerId in the CallAgentState instead since it's similar to the display name?

dmceachernmsft:
> Wondering if we should have the alternativeCallerId in the CallAgentState instead since it's similar to the display name?

im torn on this. I like the idea of it living in this state since the number should also be dependent on the callAgent existing, however, this callerId does behave like a identifier for the ACS user calling the PSTN user. Any thoughts here on your side?

carocao-msft (approved):

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
```diff
@@ -193,7 +196,7 @@ export interface RemoteVideoStreamState {
 
 // @public
 export interface StatefulCallClient extends CallClient {
-    createCallAgent(...args: Parameters<CallClient['createCallAgent']>): Promise<DeclarativeCallAgent>;
+    createCallAgent(tokenCredential: CommunicationTokenCredential, options?: CallAgentOptions, alternativeCallerId?: string): Promise<DeclarativeCallAgent>;
```
anjulgarg, line 199: Why are we completely changing the function signature here?

packages/calling-stateful-client/review/stable/calling-stateful-client.api.md
```diff
@@ -185,7 +187,7 @@ export interface RemoteVideoStreamState {
 
 // @public
 export interface StatefulCallClient extends CallClient {
-    createCallAgent(...args: Parameters<CallClient['createCallAgent']>): Promise<DeclarativeCallAgent>;
+    createCallAgent(tokenCredential: CommunicationTokenCredential, options?: CallAgentOptions): Promise<DeclarativeCallAgent>;
```
anjulgarg, line 190: This change shouldn't be in the stable API

packages/calling-stateful-client/src/StatefulCallClient.ts
```diff
@@ -139,7 +150,11 @@ export interface StatefulCallClient extends CallClient {
    * @param options - The CallAgentOptions for additional options like display name.
    * @public
    */
-  createCallAgent(...args: Parameters<CallClient['createCallAgent']>): Promise<DeclarativeCallAgent>;
+  createCallAgent(
```
anjulgarg, line 153: Why are we completely changing the function signature here?
Would `createCallAgent(...args: Parameters<CallClient['createCallAgent']>, alternativeCallerId?: string)` not work?

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -2246,7 +2247,7 @@ export interface SendBoxStylesProps extends BaseCustomStyles {
 
 // @public
 export interface StatefulCallClient extends CallClient {
-    createCallAgent(...args: Parameters<CallClient['createCallAgent']>): Promise<DeclarativeCallAgent>;
+    createCallAgent(tokenCredential: CommunicationTokenCredential, options?: CallAgentOptions): Promise<DeclarativeCallAgent>;
```
anjulgarg, line 2250: We shouldn't be changing the stable API

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2528,7 +2532,7 @@ export interface SendBoxStylesProps extends BaseCustomStyles {
 
 // @public
 export interface StatefulCallClient extends CallClient {
-    createCallAgent(...args: Parameters<CallClient['createCallAgent']>): Promise<DeclarativeCallAgent>;
+    createCallAgent(tokenCredential: CommunicationTokenCredential, options?: CallAgentOptions, alternativeCallerId?: string): Promise<DeclarativeCallAgent>;
```
anjulgarg, line 2535: Wondering if we can add `alternativeCallerId` to `options`?

anjulgarg:
@dmceachernmsft Possible to add Unit test to this PR to ensure that the callerId you are setting as a function argument is correctly reflected in the state?

packages/calling-stateful-client/src/StatefulCallClient.ts
dmceachernmsft: this unfortunately didn't work tsc hated the idea of a optional prop before a required one, and as well it hated the spread of `...args` happening before the `alternativeCallerId`

packages/communication-react/review/beta/communication-react.api.md
anjulgarg: So adding an optional prop shouldn't be a breaking change since it doesn't break an end-user's implementation.
Also, we are not proposing that this function matches the headless function.

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
anjulgarg: @dmceachernmsft could you possibly reach out to someone from ARB and see if they consider this a change?

packages/communication-react/review/beta/communication-react.api.md
dmceachernmsft: okay sounds good, after writing the test I do like it better in the options than its own field. I will update this and the test then reach out to the ARB about this change. 

it will be needed if we are to store this prop in the callAgentState. otherwise it will have to go back to the original PR where it was in the callClientState

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
prprabhu-ms: I've asked a similar question before. As long as the final concrete types are unchanged, it is not considered a breaking change.

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
```diff
@@ -191,9 +194,14 @@ export interface RemoteVideoStreamState {
...
 
+// @public
+export interface StatefulCallAgentOptions extends CallAgentOptions {
+    alternativeCallerId?: string;
+}
+
 // @public
 export interface StatefulCallClient extends CallClient {
-    createCallAgent(...args: Parameters<CallClient['createCallAgent']>): Promise<DeclarativeCallAgent>;
+    createCallAgent(tokenCredential: CommunicationTokenCredential, options?: StatefulCallAgentOptions): Promise<DeclarativeCallAgent>;
```
prprabhu-ms, line 204: It makes sense to me that Contoso needs to provide `alternativeCallerId` when creating the adapter -- that is the only place Contoso has to provide this information.

It also makes sense to me that we need to save `alternativeCallerId` *somewhere* in the `CallClientState`, because it is needed by the handlers for the `StartCallButton`.. The alternative would be for Contoso to provide the `alternativeCallerId` directly to the props for `StartCallButton`, but that would break the nice property that they simply need to pass in `usePropsFor(StartCallButton)` to get the button working.


That said, I'm uncomfortable with this extension of `CallAgentOptions` and `CallAgentState`.

The important distinction between `alternativeCallerId` and most of the fields in `CallClientState` is that we never recieve `alternativeCallerId` from the ACS services -- this is not a cached value we get from the backed. In that way, this is similar to `CallClientState.userId`.
I'd also expect `CallClientState.userId` to only be used from handlers coresponding to `CallAgent` methods, similar to the handlers that use `alternativeCallerId`.

I'd prefer to keep our departure from headless SDK APIs in one place -- while creating the `StatefulCallClient`, and store `alternativeCallerId` alongside `userId`.

I think it's worth bringing this up in the team channel to see what folks think.

dmceachernmsft: So this phone number that is going in `alternativeCallerId` is from the ACS backend. it is required that your phone number comes from the same connection string and resource that you are running your app with. That said I have been going back and forth with where to store this phone number, both places (callClientState & callAgentState) I think have good reasons to have it. I will post something about this in the teams chat.

prprabhu-ms (approved):

prprabhu-ms:
Thanks for getting consensus on where to place the field.

PR #2073, opened 2022-07-13
Add startCall Functionality to composites

dmceachernmsft:
# What

Update handlers to use alternativeCallerId
# Why

Allows for composites to start a call for PSTN or ACS users

https://skype.visualstudio.com/SPOOL/_workitems/edit/2920719
# How Tested

built locally and ran samples to make sure new functionality didn't break anything. will be tested properly when samples are updated.


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -82,6 +82,7 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
 
   const adapter = useAdapter();
   const locale = useLocale();
+
```
prabhjot-msft, line 85: Looks like extra space got added here

prabhjot-msft (approved):

anjulgarg (approved):

PR #2075, opened 2022-07-14
pstn 1:n calling storybook doc update

dmceachernmsft:
# What

Introduce docs for PSTN and 1:N in the composites
# Why

starts to introduce how to use PSTN and 1:N calling with the composites

https://skype.visualstudio.com/SPOOL/_workitems/edit/2924137
# How Tested

inspected storybook locally.


dmceachernmsft:
Going to close this with the updated comments off main, there is some code holding over from before thats not correct.

PR #2076, opened 2022-07-14
Dropdown in Participant pane to open dialpad and add new participants

carocao-msft:
# What
Created add people dropdown in people pane
Created dialpad modal

Coming soon in other PRs: 
dialpad delete button
UI tests 

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2920627
https://skype.visualstudio.com/SPOOL/_workitems/edit/2920628

# How Tested

https://user-images.githubusercontent.com/96077406/179122620-716af982-7e37-49f7-88bb-529e85235f92.mov


carocao-msft (commented):
:)

prprabhu-ms:
A quick pass for today. Will continue reviewing the `DialpadModel.tsx` tomorrow.

packages/react-composites/src/composites/CallWithChatComposite/PeoplePaneContent.tsx
```diff
@@ -109,39 +122,154 @@ export const PeoplePaneContent = (props: {
...
-      {inviteLink && (
+    <>
+      <DialpadModal isMobile={false} strings={dialpadModalStrings} isModalOpen={isModalOpen} hideModal={hideModal} />
+      <Stack tokens={peoplePaneContainerTokens}>
         <Stack styles={copyLinkButtonStackStyles}>
           <DefaultButton
-            text={strings.copyInviteLinkButtonLabel}
-            onRenderIcon={() => <CallWithChatCompositeIcon iconName="Link" style={linkIconStyles} />}
-            onClick={() => copy(inviteLink)}
+            onRenderIcon={() => <PersonAdd20Regular />}
```
dmceachernmsft, line 263: Should we be registering this icon?

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const onChange = (input: string): void => {
+    setTextfieldInput(input);
+  };
+
+  const onClickCall = (): void => {
+    //place holder for adding calling functionality
+    console.log(textfieldInput);
+  };
+
+  const dialpadModelStyle: Partial<IModalStyles> = useMemo(
```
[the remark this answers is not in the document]
carocao-msft: Using useMemo here to improve perf so leaving them in the tsx file similar to what we did with copyLinkButtonStylesThemed, not sure if they should be in a separate style file.

prprabhu-ms (commented):
Will do another pass shortly.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -372,6 +372,7 @@ export type CallCompositeIcons = {
     ParticipantItemScreenShareStart?: JSX.Element;
     VideoTileMicOff?: JSX.Element;
     LocalCameraSwitch?: JSX.Element;
+    AddPerson?: JSX.Element;
```
prprabhu-ms, line 375: This name needs more context given that we have a flat list of icons.

How about `ParticipantPaneAddPerson`?

packages/react-composites/src/composites/CallWithChatComposite/Strings.tsx
```diff
@@ -107,4 +112,30 @@ export interface CallWithChatCompositeStrings {
...
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  /**
+   * Label for add people dropdown
+   */
+  addPeopleButtonLabel: string;
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  /**
+   * Label for button to start a call
+   */
+  callButtonLabel: string;
```
prprabhu-ms, line 125: This name is too generic. Should this be `dialpadStartButtonLabel`?

packages/react-composites/src/composites/CallWithChatComposite/PeoplePaneContent.tsx
```diff
@@ -127,22 +267,57 @@ export const PeoplePaneContent = (props: {
...
-            onClick={() => copy(inviteLink)}
-            styles={copyLinkButtonStylesThemed}
-          />
+  };
+
+  const desktopViewAddPeopleButton = (inviteLink?: string): JSX.Element => {
+    /* @conditional-compile-remove(PeoplePaneDropdown) */
+    return (
+      <Stack>
+        <DialpadModal isMobile={false} strings={dialpadModalStrings} isModalOpen={isModalOpen} hideModal={hideModal} />
```
prprabhu-ms, line 276: I have requeseted in ealier comments to extract helper components here. That is also necessary to minimize the conditional compilation directives you need.

Beyond the API (which in this case simply means the new strings and icon name), you might get away with just 1 conditional compilation directive: To add the "add people" component that internally contains the modal and tracks the state for modal being shown or not etc.

Beyond that all helper components, and the Dialpad component itself should be implemented without any conditional compilation (it's just not used for stable builds).

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+export interface DialpadModalProps {
+  isMobile: boolean;
+  isModalOpen: boolean;
+  strings: DialpadModalStrings;
+  hideModal: () => void;
+}
+
+/** @private */
+export const DialpadModal = (props: DialpadModalProps): JSX.Element => {
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
```
prprabhu-ms, line 35: This file needs no conditional compilation at all.

The live documentation for conditional compilation explains how to keep conditional compilation to a minimum: https://github.com/Azure/communication-ui-library/blob/main/packages/acs-ui-common/src/conditional-compilation-sample/index.tsx#L30

Since `DialpadModal` is a `@private` component, we should define it unconditionally. Only the `PeoplePaneContent` should have (minimal) conditional compilation necessary to avoid rendering the `DialpadModal` in stable builds.

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+  closeModalButtonAriaLabel: string;
+  callButtonLabel: string;
+}
+
+/** @private */
+export interface DialpadModalProps {
+  isMobile: boolean;
+  isModalOpen: boolean;
+  strings: DialpadModalStrings;
+  hideModal: () => void;
```
prprabhu-ms, line 30: Rename to match our convention of naming handlers with `on` prefix: `onDismiss` (`Modal` is redudant).

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+export const DialpadModal = (props: DialpadModalProps): JSX.Element => {
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  const { strings, isMobile, isModalOpen, hideModal } = props;
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  const [textfieldInput, setTextfieldInput] = useState('');
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  const theme = useTheme();
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  const onLightDismissTriggered = (): void => hideModal();
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
```
prprabhu-ms, line 43: nit: `const onLightDismissTriggered = hideModal`. That way your avoid creating a new closure on each render (and causing child components to render again).

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+  strings: DialpadModalStrings;
+  hideModal: () => void;
+}
+
+/** @private */
+export const DialpadModal = (props: DialpadModalProps): JSX.Element => {
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  const { strings, isMobile, isModalOpen, hideModal } = props;
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  const [textfieldInput, setTextfieldInput] = useState('');
```
prprabhu-ms, line 38: nitty nit: Is `Textfield` one word? I always think of it as two words `TextField` (applies to various names here)

A quick search seems to suggest that it is two words:

https://developer.microsoft.com/en-us/fabric-js/components/textfield/textfield

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  const theme = useTheme();
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  const onLightDismissTriggered = (): void => hideModal();
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  const onChange = (input: string): void => {
```
prprabhu-ms, line 44: nit: again this can just be `onChange = setTextFieldInput` (or better still, simply pass `setTextFieldInput` as the callback below.

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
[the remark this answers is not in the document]
prprabhu-ms: +1 to Donald's comment. Even if you want to use `useMemo` (yay!), extract a helper function into `DialpadModal.styles.ts` and use it here.

```
const modalStyle = useMemo(() => themedModalStyle(theme), [theme]);
```

Same comment for other styles below. The intent of extracting these functions is to make this component more readable.

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+    },
+    textContainer: {
+      display: 'contents'
+    }
+  };
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  if (isMobile) {
+    return (
+      <Stack>
+        {isModalOpen && (
```
prprabhu-ms, line 107: The `isModalOpen` field clearly controls non-modal surfaces on mobile.

Overall, we should probably drop the word `modal` from naming in most places. So this should be `showDialpad`, `onDismissDialpad` etc.

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
prprabhu-ms: nit: again this can just be `onChange = setTextFieldInput` (or )

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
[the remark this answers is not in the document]
prprabhu-ms: +1 to Donald's comment. Even if you want to use `useMemo` (yay!), extract a helper function into `DialpadModal.styles.ts` and use it here.

```
const modalStyle = useMemo(() => themedModalStyle(theme), [theme]);
```

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { Dialpad, DialpadStyles, _DrawerMenu, _DrawerMenuItemProps, _DrawerSurface } from '@internal/react-components';
+/* @conditional-compile-remove(PeoplePaneDropdown) */
+import { IButtonStyles, IModalStyles, Modal, Stack, useTheme, Text, IconButton, PrimaryButton } from '@fluentui/react';
+/* @conditional-compile-remove(PeoplePaneDropdown) */
+import { Call20Regular } from '@fluentui/react-icons';
+/* @conditional-compile-remove(PeoplePaneDropdown) */
+import { drawerContainerStyles } from '../styles/CallWithChatCompositeStyles';
+
+/** @private */
+export interface DialpadModalStrings {
```
prprabhu-ms, line 18: See another comment below.

This component shows a modal or a drawer based on whether we're on mobile. As such, perhaps just name it `DialpadContainer`, or better still `PreparedDialpad` (that's what it really is -- it brings together styles, hooks together things for dialpad). I've done something similar for buttons in the control bar.

prprabhu-ms:
Left somewhat more detailed comments than usual because of the timezone difference. Let me know if you would like to go over some of the comments together.

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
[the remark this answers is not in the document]
carocao-msft: Updated

packages/react-composites/src/composites/CallWithChatComposite/components/DialpadModal.tsx
carocao-msft: Updated, however still need to conditional compile part of this because the dialpad component is conditionally compiled

common/config/babel/.babelrc.js
```diff
@@ -34,7 +34,9 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         // dialpad
         'dialpad',
         // rooms
-        'rooms'
+        'rooms',
+        // Add people dropdown and dialpad modal in people pane
+        'PeoplePaneDropdown',
```
prprabhu-ms, line 39: nit: Can we rename this to `people-pane-dropdown` to be consistent with other feature flag names?

packages/react-composites/src/composites/common/icons.tsx
```diff
@@ -69,7 +71,13 @@ export const COMPOSITE_ONLY_ICONS: CompositeIcons = {
   NoticePageLeftCall: <CallEnd20Filled />,
   NoticePageRemovedFromCall: <Info20Filled />,
   /* @conditional-compile-remove(file-sharing) */
-  SendBoxAttachFile: <Attach20Regular />
+  SendBoxAttachFile: <Attach20Regular />,
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  PeoplePaneAddPerson: <PersonAdd20Regular />,
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  Dialpad: <Dialpad20Regular />,
```
prprabhu-ms, line 78: This name is too generic. How about `PeoplePaneOpenDialpad`

packages/react-composites/src/composites/common/icons.tsx
```diff
@@ -69,7 +71,13 @@ export const COMPOSITE_ONLY_ICONS: CompositeIcons = {
...
+  /* @conditional-compile-remove(PeoplePaneDropdown) */
+  Call: <Call20Regular />
```
prprabhu-ms, line 80: This name is too generic. How about `DialpadStartCall`

packages/react-composites/src/composites/CallWithChatComposite/PeoplePaneContent.tsx
```diff
@@ -109,39 +122,154 @@ export const PeoplePaneContent = (props: {
...
+    menuStyleThemed
+  ]);
+
+  const hideModal = (): void => {
+    setIsModalOpen(false);
+  };
+
+  //mobile add people button with botten sheet drawers
+
+  const [addPeopleDrawerMenuItems, setAddPeopleDrawerMenuItems] = useState<_DrawerMenuItemProps[]>([]);
```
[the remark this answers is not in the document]
prprabhu-ms: What I mean here is to create a new component that holds the show/hide dialpad logic along with the people button dropdown.

This way, you can minimize the conditional compilation directives and also keep the `PeoplePaneContent` component from growing too large / doing too many things.

packages/react-composites/src/composites/CallWithChatComposite/components/AddPeopleDropdown.styles.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { IButtonStyles, IContextualMenuStyles, Theme } from '@fluentui/react';
+
+/**
+ * @private
+ */
+export const themedCopyLinkButtonStyles = (mobileView: boolean | undefined, theme: Theme): Partial<IButtonStyles> => ({
```
prprabhu-ms, line 9: nitty nit: Switch the order of arguments so you don't have to explicitly pass `undefined`: `themedCopyLinkButtonStyles(theme)` vs `themedCopyLinkButtonStyles(undefined, theme)`.

packages/react-composites/src/composites/CallWithChatComposite/components/AddPeopleDropdown.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+  root: {
+    borderRadius: theme.effects.roundedCorner6
+  }
+});
+
+/**
+ * @private
+ */
+export const iconStyles = {
+  lineHeight: '0',
```
prprabhu-ms, line 33: Might be good to add a comment why thi `lineHeight` needed to be set to 0 (if we know).

packages/react-composites/src/composites/CallWithChatComposite/components/AddPeopleDropdown.tsx
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/** @private */
+export const AddPeopleDropdown = (props: AddPeopleDropdownProps): JSX.Element => {
+  const theme = useTheme();
+
+  const { inviteLink, strings, mobileView } = props;
+  const [showDialpad, setShowDialpad] = useState(false);
+
+  const menuStyleThemed: Partial<IContextualMenuStyles> = useMemo(() => themedMenuStyle(theme), [theme]);
```
prprabhu-ms, line 44: nitty nit: You should not need to explicitly specify the type here (`useMemo` should figure it out).

packages/react-composites/src/composites/CallWithChatComposite/components/AddPeopleDropdown.tsx
```diff
@@ -0,0 +1,159 @@
...
+        //   happens when we change the VideoGallery layout, or even when the video stream element is internally resized
+        //   by the headless SDK.
+        // - There is a `preventDismissOnEvent` prop that we could theoretically use to only dismiss when the target of
+        //   of the 'resize' event is the window itself. But experimentation shows that setting that prop doesn't
+        //   deterministically avoid dismissal.
+        //
+        // A side effect of this workaround is that the context menu stays open when window is resized, and may
+        // get detached from original target visually. That bug is preferable to the bug when this value is not set -
+        // The Callout (frequently) gets dismissed automatically.
+        preventDismissOnResize: true
```
prprabhu-ms, line 68: See https://github.com/Azure/communication-ui-library/pull/2069 that recently introduced a common constant for `calloutProps`. Test around to see if you need to prevent dismiss on scroll as well.

packages/react-composites/src/composites/CallWithChatComposite/components/AddPeopleDropdown.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  const onDismissDialpad = (): void => {
+    setShowDialpad(false);
+  };
+
+  const [addPeopleDrawerMenuItems, setAddPeopleDrawerMenuItems] = useState<_DrawerMenuItemProps[]>([]);
+
+  const setDrawerMenuItemsForAddPeople: () => void = useMemo(() => {
+    return () => {
+      const drawerMenuItems = defaultMenuProps.items.map((contextualMenu: IContextualMenuItem) =>
```
prprabhu-ms, line 108: nit: Should this check `mobileView` and return just an empty list if it is unset (On desktop, avoid creating a new list that is never used).

packages/react-composites/src/composites/CallWithChatComposite/components/PreparedDialpad.tsx
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import React from 'react';
+/* @conditional-compile-remove(people-pane-dropdown) */
```
prprabhu-ms, line 4: You should be able to remove *all* conditional compilation from this file.

packages/react-composites/src/composites/common/PeoplePaneContent.tsx
```diff
@@ -95,26 +107,21 @@ export const PeoplePaneContent = (props: {
...
-        }
-      }),
-    [props.mobileView, theme.effects.roundedCorner6, theme.effects.roundedCorner4]
+    (): IButtonStyles => concatStyleSets(copyLinkButtonStyles, themedCopyLinkButtonStyles(props.mobileView, theme)),
+    [props.mobileView, theme]
   );
 
-  if (props.mobileView) {
+  // for conditional compile
+  const mobileViewAddPeopleButton = (inviteLink?: string): JSX.Element => {
```
prprabhu-ms, line 115: nit: I'm very scared of local variables that are really potentially React components.

Can we extract these into file-private component `MobileViewAddPeopleButton` and use it as `<MobileViewAddPeopleButton .../>` below?

prprabhu-ms (commented):
Mostly there. `PreparedDialpad.tsx` needs another look, but otherwise looking good.

Let's make sure another reviewer takes a good look.

packages/react-composites/src/composites/CallWithChatComposite/components/PreparedDialpad.tsx
carocao-msft: Need to conditional compile this and addpeopledropdown because both is using conditionally compiled components, if not will have an error in stable build

packages/react-composites/src/composites/CallWithChatComposite/components/AddPeopleDropdown.tsx
carocao-msft: From my understanding this should not be called when !mobileView?

packages/react-composites/src/composites/CallWithChatComposite/components/AddPeopleDropdown.styles.ts
carocao-msft: updated

packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -311,17 +302,14 @@ const DialpadContainer = (props: {
...
-      }
-    </>
-  );
+  return <DialpadContainer strings={strings} {...props} />;
 };
+
+function dialpadLocaleStringsTrampoline(): DialpadStrings {
+  /* @conditional-compile-remove(dialpad) */
+  return useLocale().strings.dialpad;
+  return {} as DialpadStrings;
```
prprabhu-ms, line 314: nitty nit: Funilly `useLocale().strings.dialpad` actually exists and has the right values (only the type doesn't include those fields in the stable build).
So I would `return useLocale().strings.dialpad as unknown as DialpadStrings`

Just a nit, we don't anticipate ever using that code path

packages/react-composites/src/composites/CallWithChatComposite/components/AddPeopleButton.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const { inviteLink, mobileView, strings, participantList } = props;
+
+  const theme = useTheme();
+
+  const copyLinkButtonStylesThemed = useMemo(
+    (): IButtonStyles => concatStyleSets(copyLinkButtonStyles, themedCopyLinkButtonStyles(mobileView, theme)),
+    [mobileView, theme]
+  );
+
+  if (mobileView) {
```
prprabhu-ms, line 40: nit: I'd keep the conditional compilation in one place

```TypeScript

if (mobileView) {
    return <AddPeopleDropdown ... />
} else {
    return (
        <Stack>
          <AddPeopleDropdown ... />
        </Stack>
    );
}

if (mobileView) {
   ...
}
...
```

packages/react-composites/src/composites/CallWithChatComposite/PreventDismissOnEvent.tsx
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation.
```
prprabhu-ms, line 1: Can we move https://github.com/Azure/communication-ui-library/blob/main/packages/react-components/src/components/utils/common.ts to `packages/acs-ui-common` and reuse it.

prprabhu-ms (changes requested):

prprabhu-ms (commented):
Oops: Did not intend to "Request changes"

prprabhu-ms (approved):
Updated `Dialpad` tests to run in stable build (mostly).

change/@internal-acs-ui-common-cef2f404-69a8-4f4a-a136-4789be03810c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
anjulgarg, line 2: should be none
It is not changing any public API

packages/acs-ui-common/src/common.ts
```diff
@@ -2,17 +2,18 @@
...
  * @returns units of pixels
  */
-export const convertRemToPx = (rem: number): number => {
+export const _convertRemToPx = (rem: number): number => {
   return rem * parseFloat(getComputedStyle(document.documentElement).fontSize);
 };
 
 /**
+ * @internal
  * Converts units of rem to units of pixels
```
anjulgarg, line 16: Invalid comment here.

packages/react-components/review/stable/react-components.api.md
```diff
@@ -414,6 +415,53 @@ export interface DevicesButtonStyles extends ControlBarButtonStyles {
     menuStyles?: Partial<DevicesButtonContextualMenuStyles>;
 }
 
+// @beta
+export const Dialpad: (props: DialpadProps) => JSX.Element;
+
+// @beta
+export interface DialpadButtonProps {
+    // (undocumented)
+    primaryContent: string;
```
anjulgarg, line 424: Let's document these props.

packages/react-composites/src/composites/CallWithChatComposite/components/PreparedDialpad.styles.ts
```diff
@@ -0,0 +1,60 @@
...
+export const themedDialpadStyle = (isMobile: boolean, theme: Theme): Partial<DialpadStyles> => ({
+  root: {
+    padding: 0,
+    marginLeft: 0,
+    marginRight: 0,
+    maxWidth: '100%'
+  },
+  textField: {
+    root: {
+      borderBottom: '1px solid lightgrey'
```
anjulgarg, line 30: Can we use a palette color here? If not what is the reason behind using a static color?

packages/react-composites/src/composites/CallWithChatComposite/components/PreparedDialpad.styles.ts
```diff
@@ -0,0 +1,60 @@
...
+    },
+    field: {
+      backgroundColor: 'white',
```
anjulgarg, line 33: Can we use a palette color here? If not what is the reason behind using a static color?

packages/react-composites/src/composites/CallWithChatComposite/components/PreparedDialpad.styles.ts
```diff
@@ -0,0 +1,60 @@
...
+    color: theme.palette.themeDarkAlt
+  }
+});
+
+/**
+ * @private
+ */
+export const themedCallButtonStyle = (theme: Theme): Partial<IButtonStyles> => ({
+  root: {
+    fontWeight: 600,
```
anjulgarg, line 50: Why are we enforcing a static fontWeight here?
Can we use `theme.fonts.medium.fontWeight`?

packages/react-composites/src/composites/CallWithChatComposite/components/PreparedDialpad.tsx
```diff
@@ -0,0 +1,118 @@
...
+/** @private */
+export interface PreparedDialpadProps {
+  isMobile: boolean;
+  showDialpad: boolean;
+  strings: PreparedDialpadStrings;
+  onDismissDialpad: () => void;
+}
+
+/** @private */
+export const PreparedDialpad = (props: PreparedDialpadProps): JSX.Element => {
```
anjulgarg, line 33: Why is it called a PreparedDialpad?

carocao-msft: It's a dialpad with call button implemented and all styling required for being in the composite :)

anjulgarg:
Given that the all the new components introduced here will be used by Call Composite and CallWithChat composite.
We need to place them in the `composites/common` folder.

packages/react-composites/src/composites/CallWithChatComposite/components/PreparedDialpad.tsx
anjulgarg: Given that we will have (most probably) another dialpad in the composites for just sending DTMF tones without the start call button, any idea what that dialpad would be called?

Since in this context that dialpad would also be a "prepared" dialpad for sending dtmf tones.

packages/react-components/review/stable/react-components.api.md
```diff
@@ -414,6 +415,53 @@ export interface DevicesButtonStyles extends ControlBarButtonStyles {
     menuStyles?: Partial<DevicesButtonContextualMenuStyles>;
 }
 
+// @beta
+export const Dialpad: (props: DialpadProps) => JSX.Element;
```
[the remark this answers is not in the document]
anjulgarg: Oh this is react-components.api.md, when we say stable api, we only talk of `communication-react.api.md`

packages/react-composites/src/composites/CallWithChatComposite/components/PreparedDialpad.tsx
carocao-msft: PreparedDialpadForDTMF? and this one we can call it PreparedDialpadForCalling

change/@internal-acs-ui-common-cef2f404-69a8-4f4a-a136-4789be03810c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "moved common.ts file from react components to acs-ui-common to be reused by other packages",
+  "packageName": "@internal/acs-ui-common",
+  "email": "carolinecao@microsoft.com",
+  "dependentChangeType": "patch"
```
anjulgarg, line 6: patch -> none

packages/react-composites/src/composites/CallWithChatComposite/components/PreparedDialpad.tsx
anjulgarg: I would shorten it to `CallingDialpad` and `DTMFDialpad` to keep the naming crisp.

anjulgarg: Personally `Prepared` as a prefix is not adding value here since these components exists in the composites package already.

anjulgarg (approved):

packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -143,7 +137,6 @@ const DialpadButton = (props: {
   );
 };
 
-/* @conditional-compile-remove(dialpad) */
```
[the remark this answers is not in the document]
prprabhu-ms: They aren't required because we don't export `Dialpad` from the public API.

prprabhu-ms (approved):

PR #2078, opened 2022-07-14
Convert `ErrorBar` e2e tests to use `MockCallAdapterState`

prprabhu-ms:
# What

Replaces a test specific `TestCallingState` data structure with `MockCallAdapterState`. The new type closely follows `CallAdapterState`. The only difference is that video element views `HTMLElement` is left unset, to be populated by the test app.

# Why

Makes it easier to write more e2e test cases by avoiding having to extend `TestCallingState` repeatedly.


packages/react-composites/tests/browser/call/MockCallAdapterState.ts
```diff
@@ -0,0 +1,92 @@
...
+} from '@internal/calling-stateful-client';
+import type { CallAdapterState } from '../../../src';
+
+/**
+ * A slight modification of {@link CallAdapterState} for initializing the
+ * {@link MockCallAdapter} for hermetic e2e tests.
+ *
+ * - HTMLElement views are replaced with placeholders to be populated
+ *   by the test app.
+ */
```
mgamis-msft, line 19: Thanks for creating these Mock interfaces. Now it is tightly coupled with the state interfaces of in 'calling-stateful-client'

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -2,56 +2,33 @@
...
-          media: { latest: { ...initialState.call.diagnostics.media.latest, ...testState.diagnostics?.media } }
-        };
-      }
-    }
-    if (testState.page) {
-      initialState.page = testState.page;
+  /**
+   * `testState` is deprecated. Use `initialState` instead.
+   */
+  constructor(testState?: TestCallingState, initialState?: MockCallAdapterState) {
```
mgamis-msft, line 23: Thanks for adding the state as a constuctor argument. Looks much cleaner now :)

mgamis-msft (approved):

carocao-msft (approved):

packages/react-composites/tests/browser/call/app/HermeticApp.tsx
```diff
@@ -6,30 +6,23 @@ import React, { useState, useEffect } from 'react';
 import { CallAdapter } from '../../../../src';
 import { MockCallAdapter } from './mocks/MockCallAdapter';
 import { QueryArgs } from './QueryArgs';
-import { TestCallingState } from '../TestCallingState';
 import { BaseApp } from './BaseApp';
 
-/** @internal */
-export function HermeticApp(props: { queryArgs: HermeticAppQueryArgs }): JSX.Element {
+export function HermeticApp(props: { queryArgs: QueryArgs }): JSX.Element {
```
dmceachernmsft, line 11: Maybe a silly question, but why is it called a `HermeticApp`?

packages/react-composites/tests/browser/call/app/mocks/MockCallAdapter.ts
```diff
@@ -2,56 +2,33 @@
...
-import { CallAdapter, CallAdapterState } from '../../../../../src';
-import { TestCallingState, TestRemoteParticipant } from '../../TestCallingState';
+import type { CallAdapter, CallAdapterState } from '../../../../../src';
+import type { MockCallAdapterState } from '../../MockCallAdapterState';
+import type { TestCallingState, TestRemoteParticipant } from '../../TestCallingState';
 
 /**
  * Mock class that implements CallAdapter interface for UI snapshot tests. The handler implementation is currently limited so
  * some composite behaviour will not work like clicking the 'Start Call' button in the Configuration page. The usage of
  * MockCallAdapter is intended to talk snapshot based only on the state of a CallAdapter.
```
dmceachernmsft, line 17: nit: take

packages/react-composites/tests/browser/call/hermetic/ErrorBar.test.ts
```diff
@@ -1,53 +1,50 @@
...
-            target: 'Call.stopVideo',
-            innerError: new Error('Inner error of failure to stop video')
-          }
-        }
-      })
-    );
+    const initialState = defaultMockCallAdapterState();
+    initialState.latestErrors = {
+      'Call.unmute': {
+        timestamp: new Date(),
```
dmceachernmsft, line 33: Do these time stamps need to be real times? or would just giving them a dummy value be better?

dmceachernmsft (approved):

packages/react-composites/tests/browser/call/hermetic/ErrorBar.test.ts
prprabhu-ms: I guess not necessary, but I'm avoiding any semantic changes in this PR.

PR #2079, opened 2022-07-14
Merge Permissions provider to main

PorterNan:
Add permission provider to the library and roles

More component capabilities support will be in the following PR
need https://github.com/Azure/communication-ui-library/pull/2063 to get conditional compile running
# What

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/index.ts
```diff
@@ -6,9 +6,21 @@ export { _IdentifierProvider } from './identifiers';
 export type { _Identifiers, _IdentifierProviderProps } from './identifiers';
 export * from './localization/locales';
 export { LocalizationProvider } from './localization';
+
```
alkwa-msft, line 9: can we remove this newline?

alkwa-msft (approved):

mgamis-msft (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -45,6 +48,12 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
    * Flags to enable/disable or customize UI elements of the {@link CallComposite}.
    */
   options?: CallCompositeOptions;
+
+  /* @conditional-compile-remove(rooms) */
+  /**
+   * Set this to enable/disable capacities for different roles
```
JamesBurnside, line 54: If this is optional, mark what the `@defaultValue` is in the doc comment

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -45,6 +48,12 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
...
+   */
+  role?: Role;
```
JamesBurnside, line 56: Is this just for testing and it will usually be obtained through a useSelector? This seems like it should be an adapter level property instead of a direct Composite prop

JamesBurnside (changes requested):
I'm a little confused on the purpose of the provider -- is there a document to say why this is a UI level feature and not something we integrate into the adapter/stateful layer? (soft blocking comment)

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
PorterNan: We've discussed this for a while - In this beta version, we will merge provider into main and keep it in a provider way(a localization-like solution) and we will decide whether make it selector level once call sdk and rooms sdk decided to provide permission object in stateful layer

PorterNan:
> I'm a little confused on the purpose of the provider -- is there a document to say why this is a UI level feature and not something we integrate into the adapter/stateful layer? (soft blocking comment)

Currently permissions are not coming from either calling sdk or rooms api, it is a UI layer object created by us, we are merging this for beta only because we don't want to leave everything in our feature branch, and we will make decision for stable version later once we get more determined answer from call sdk and rooms sdk

JamesBurnside (approved):
Chatted offline on what the current situation is on the rooms SDK, unblocking

PR #2080, opened 2022-07-14
Disable two very flakey e2e tests and disable internal retries

prprabhu-ms:
# 🚧 Work In Progress  🚧 

* Ferret out all the flakey hermetic tests.
* Disable those tests, file bugs.
* Fix flakey tests, then reenable them.

# Why

See https://skype.visualstudio.com/SPOOL/_workitems/edit/2925017 for context. That bug also tracks stabilizing and re-enabling the disabled tests.

Internal retries for hermetic tests have two very real costs:

* Longer CI times due to retries.
* Lower e2e test quality because we paper over flakey tests.

In the future, we will aggressively disable any flakey tests instead of retrying internally and then re-enable them once they are stabilized.

### DIFF
--- a/change/@internal-react-composites-67e017a7-8fcf-4b2c-8ea4-5f152b1634fd.json
+++ b/change/@internal-react-composites-67e017a7-8fcf-4b2c-8ea4-5f152b1634fd.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Disable flakey e2e tests and drop retries",
+  "packageName": "@internal/react-composites",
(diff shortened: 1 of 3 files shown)


prprabhu-ms:
Flakiness is death by a thousand cuts:

image

The current state of flakiness makes it very hard to work on e2e tests (or on anything for that matter) and also flakiness + retries is likely a large contributing factor to the e2e job runtimes.

I'm going to prioritize flakiness reduction work next sprint and have more than just this one-off PR for it.

PR #2083, opened 2022-07-15
Convert PeoplePaneContent to a common file for Composite use

edwardlee-msft:
# What
Convert PeoplePaneContent.tsx to a common file to be used in both Call Composite and CallWithChatComposite

# Why
In order to bring participant list view in Call Composite in sync with CallWithChat Composite.
Initial steps requiring converting CallWIthChat composite specific files to common files to be used in Call Composite.
https://skype.visualstudio.com/SPOOL/_workitems/edit/2919194

# How Tested
Tested locally on MacOs on mobile and desktop view. 

# Process & policy checklist


alkwa-msft (approved):

change/@internal-react-composites-64003277-fb49-4fbc-bc6e-1d122ee02dc6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
dmceachernmsft, line 2: Should this be tagged as pre-release since this work is related to PSTN and 1:N?

dmceachernmsft (approved):

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -391,6 +391,7 @@ export interface CallCompositeStrings {
     complianceBannerTranscriptionStopped: string;
     configurationPageCallDetails?: string;
     configurationPageTitle: string;
+    copyInviteLinkButtonLabel: string;
```
anjulgarg, line 394: Should this be in stable?
I am assuming this is new feature specific update

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -37,7 +37,10 @@
     "removedFromCallTitle": "You were removed",
     "soundLabel": "Sound",
     "startCallButtonLabel": "Start call",
-    "microphoneToggleInLobbyNotAllowed": "Cannot mute or unmute while in lobby."
+    "microphoneToggleInLobbyNotAllowed": "Cannot mute or unmute while in lobby.",
```
anjulgarg, line 40: Ideally, this change should not be a part of this PR.
The purpose of this PR is to convert people pane into a common component.
Doing so would have kept this change as a `none` change (doesn't affect the public API in any way)

The strings can then be introduced in the PR that actually adds this people pane to call composite.

anjulgarg: And those strings would be conditionally compiled.

packages/communication-react/review/stable/communication-react.api.md
anjulgarg, line 394: Shouldn't be part of the stable api change.
These strings are tied to the call composite PSTN/1:N Calling work

packages/react-composites/src/composites/localization/locales/en-US/strings.json
edwardlee-msft: That's a good point! Thanks Anjul. Will remove these strings and only focus on moving the PeoplePaneContent out as a common with no Api change for this PR.

packages/communication-react/review/stable/communication-react.api.md
edwardlee-msft: Good point! As per the other review comment below, I have made changes to remove any api related code out. 
Those will come in a later PR which will target beta :)! Thanks Anjul!

change/@internal-react-composites-64003277-fb49-4fbc-bc6e-1d122ee02dc6.json
prprabhu-ms: This should have been `patch`, or even `none` as it has no user-visible impact.

`minor` means that there is new API / large new UX for the customer to use.

PR #2089, opened 2022-07-16
Ask device permissions based on role

mgamis-msft:
# What

If role is Consumer, then only request for audio permissions. 

- Need to do access audio devices to show speaker options in dropdown. 
- Also we disable the 'Start Call' button in the configuration page if microphonePermissionGranted is false.

Request for audio and video permissions for other roles or no role at all.

Let me know if there are changes needed.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2919171
https://skype.visualstudio.com/SPOOL/_workitems/edit/2919172

# How Tested

Tested on local rooms sample:
https://user-images.githubusercontent.com/79475487/179336360-790989e5-52b0-4dd6-8c63-68a690aa8088.mp4
Note: There is a ComplianceBanner showing camera device permissions are not granted for the Consumer role. This probably is confusing for the user. Let me know if I should remove this in this PR or another PR.

Also added unit tests

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/CallComposite.tsx
+++ b/packages/react-composites/src/composites/CallComposite/CallComposite.tsx
@@ -194,12 +194,17 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
   } = props;
   useEffect(() => {
     (async () => {
-      await adapter.askDevicePermission({ video: true, audio: true });
-      adapter.queryCameras();
-      adapter.queryMicrophones();
+      if (role === 'Consumer') {
+        // Need to ask for audio devices to get access to speakers
+        await adapter.askDevicePermission({ video: false, audio: true });
+      } else {
+        await adapter.askDevicePermission({ video: true, audio: true });
+        adapter.queryCameras();
+        adapter.queryMicrophones();
+      }
       adapter.querySpeakers();
     })();
-  }, [adapter]);
+  }, [adapter, role]);
 
   const mobileView = formFactor === 'mobile';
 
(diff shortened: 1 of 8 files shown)


alkwa-msft (approved):

prabhjot-msft (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -194,12 +194,17 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
   } = props;
   useEffect(() => {
     (async () => {
-      await adapter.askDevicePermission({ video: true, audio: true });
-      adapter.queryCameras();
-      adapter.queryMicrophones();
+      if (role === 'Consumer') {
+        // Need to ask for audio devices to get access to speakers
```
PorterNan, line 198: Do we need this? I don't think access to speaker need any permissions? I don't remember youtube is asking me any permissions to play a video

samples/Calling/src/app/utils/AppUtils.ts
```diff
@@ -59,7 +60,7 @@ export const createRoomId = async (): Promise<string> => {
  * Joins an ACS room with a given roomId and role
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const joinRoom = async (userId: string, roomId: string, role): Promise<void> => {
+export const joinRoom = async (userId: string, roomId: string, role: Role): Promise<void> => {
```
alkwa-msft, line 63: do we need to conditional hide this as well?

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
mgamis-msft: I will look into how to get speakers access without permissions from the Calling SDK.

samples/Calling/src/app/utils/AppUtils.ts
mgamis-msft: Yes when we merge it into main.

PorterNan (approved):

PR #2090, opened 2022-07-18
Refresh PNPM dependency lockfiles

prprabhu-ms:
# Why

We have seen failures in CI that look similar to https://github.com/microsoft/rushstack/issues/2542

Refreshing PNPM lockfiles as it is one of the stated workarounds. Refreshing pinned dependencies is anyway a good thing to do regularly.

### DIFF
--- a/common/config/package.json
+++ b/common/config/package.json
@@ -15,7 +15,7 @@
     "prettier:check": ""
   },
   "devDependencies": {
-    "@octokit/rest": "~18.0.6",
+    "@octokit/rest": "~19.0.3",
     "@rollup/plugin-commonjs": "~17.1.0",
     "@types/node": "^14.14.10",
     "beachball": "2.20.0",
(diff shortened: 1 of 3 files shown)


JamesBurnside (approved):

common/config/rush/command-line.json
```diff
@@ -57,6 +57,14 @@
       "shellCommand": "rush update --variant stable",
       "safeForSimultaneousRushProcesses": true
     },
+    {
+      "commandKind": "global",
+      "name": "update:full",
+      "summary": "Update dependencies for all build flavours to latest SemVers.",
```
anjulgarg, line 63: When would this command be used?

prprabhu-ms: `rush` recommends running this "regularly".

Azure SDK for JS actually has automation to run this weekly: https://github.com/Azure/azure-sdk-for-js/blob/1af9322b9a13e7b3dd758e30d55c2cff71f9fb03/eng/pipelines/weekly_automation.yml#L5

PR #2091, opened 2022-07-18
Enable PNPM workspaces

prprabhu-ms:
# What

* Enabled workspaces in `rush.json`
* Migrate packlets: `rush update:full`

# Why

This is an experimental `rush` feature that is quite stable at this point, and is recommended to get better dependency management than without workspaces.

In particular, this avoids certain edge cases where the PNPM lockfile can get into a weird state and fail future `rush update` attempts. See https://github.com/microsoft/rushstack/issues/2542

# How Tested

- Locally ran all the affected / updated `common/scripts/*`
- CI


packages/calling-component-bindings/package.json
```diff
@@ -29,9 +29,9 @@
   },
   "dependencies": {
     "@azure/communication-common": "2.0.0",
-    "@internal/acs-ui-common": "1.3.1-beta.1",
-    "@internal/calling-stateful-client": "1.3.1-beta.1",
-    "@internal/react-components": "1.3.1-beta.1",
+    "@internal/acs-ui-common": "workspace:*",
```
JamesBurnside, line 32: oh I love this if it works!

prprabhu-ms: Yep, less custom stuff!

JamesBurnside (approved):

edwardlee-msft (approved):

prprabhu-ms:
Looks like `rush` `variant` and `useWorkspaces` might not be playing well together.

This isn't really necessary for the bug I was trying to fix, so going to put this on the icebox.

PR #2092, opened 2022-07-18
Delete an unused rush sub-command

prprabhu-ms:
# Why

Every subcommand listed by `rush --help` adds cognitive load.

`rush install:stable` is neither used by scripts, nor necessary for devs. Devs always use `rush switch-flavor:stable` instead.

### DIFF
--- a/common/config/rush/command-line.json
+++ b/common/config/rush/command-line.json
@@ -59,14 +59,6 @@
     },
     {
       "commandKind": "global",
-      "name": "install:stable",
-      "summary": "Install dependencies for stable version.",
-      "description": "Install dependencies for stable version.",
-      "shellCommand": "rush install --variant stable",
-      "safeForSimultaneousRushProcesses": true
-    },
-    {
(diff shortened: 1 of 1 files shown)


PorterNan (approved):
I use this sometimes, but I am good to remove it

I usually open 2 codespaces, one for beta, one for stable, and I never do those switch(at lease not much), so everytime when I pull something, I run 'rush install:stable'

prprabhu-ms:
> I use this sometimes, but I am good to remove it
> 
> I usually open 2 codespaces, one for beta, one for stable, and I never do those switch(at lease not much), so everytime when I pull something, I run 'rush install:stable'

It is preferred to run `rush update` (and rush `update:stable`) locally. `rush install` is best used in CI, where we explicitly don't want to update the lock file.

prprabhu-ms:
Welcome back @JamesBurnside

JamesBurnside (approved):

PR #2095, opened 2022-07-18
update name for alternateCallerId

dmceachernmsft:
# What

Fix naming for `alternativeCallerId`
# Why

When programming this prop out I deviated from the name that the CallingSDK uses. this is to bring us back into parity (OOPS!)

# How Tested

Ran the updated Calling sample to verify this prop now is correct across the board.

### DIFF
--- a/packages/calling-stateful-client/src/CallClientState.ts
+++ b/packages/calling-stateful-client/src/CallClientState.ts
@@ -415,7 +415,7 @@ export interface CallClientState {
   /**
    * Stores an ACS aquired phone number for making PSTN calls.
    */
-  alternativeCallerId?: string;
+  alternateCallerId?: string;
 }
 
 /**
--- a/samples/Calling/src/app/views/CallScreen.tsx
+++ b/samples/Calling/src/app/views/CallScreen.tsx
@@ -22,16 +22,16 @@ export interface CallScreenProps {
   userId: CommunicationUserIdentifier;
   callLocator: CallAdapterLocator;
   displayName: string;
-  alternativeCallerId?: string;
+  alternateCallerId?: string;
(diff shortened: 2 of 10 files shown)


samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -22,16 +22,16 @@ export interface CallScreenProps {
...
 }
 
 export const CallScreen = (props: CallScreenProps): JSX.Element => {
-  const { token, userId, callLocator, displayName, onCallEnded, alternativeCallerId } = props;
+  const { token, userId, callLocator, displayName, onCallEnded, alternateCallerId } = props;
   const callIdRef = useRef<string>();
   const { currentTheme, currentRtl } = useSwitchableFluentTheme();
   const isMobileSession = useIsMobile();
-  console.log(alternativeCallerId);
+  console.log(alternateCallerId);
```
JamesBurnside, line 34: remove

JamesBurnside (approved):

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -415,7 +415,7 @@ export interface CallClientState {
   /**
    * Stores an ACS aquired phone number for making PSTN calls.
```
anjulgarg, line 416: Let's avoid using abbreviations like ACS.
Also I would recommend using a more verbose doc copied from the headless sdk
/**
     * A phone number in E.164 format that will be used to represent callers identity.
     * For example, using the alternateCallerId to add a participant using PSTN, this number will
     * be used as the caller id in the PSTN call.
     */

anjulgarg (approved):

PR #2103, opened 2022-07-20
CallPane created for Call Composite to render the side Participant Pane

edwardlee-msft:
# What
Created the CallPane.tsx which is the Call Composite equivalent of the CallWithChatPane.tsx from CallWithChat Composite without any chat related details.

Only People pane TabHeader. Modal is working correctly, participants are able to be removed, copy invite link is working.

People pane opens, chat does not open using the new CallPane.tsx. This is expected behavior :).
 |||| 

# Why
In order to bring Call Composite and CallWIthChat Composite to parity, Call Composite needs to have a side participant pane in desktop, and a full pane in mobile, without chat proponents.

# How Tested
Tested locally on MacOS chrome mobile and desktop view.
Tested by porting into CallWithChat and making sure UI renders and all actions are working correctly.

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/components/CallPane.tsx
```diff
@@ -0,0 +1,193 @@
...
+  mobileView?: boolean;
+  inviteLink?: string;
+  rtl?: boolean;
+}): JSX.Element => {
+  const [drawerMenuItems, setDrawerMenuItems] = useState<_DrawerMenuItemProps[]>([]);
+
+  const hidden = props.activePane === 'none';
+  const paneStyles = hidden ? hiddenStyles : props.mobileView ? availableSpaceStyles : sidePaneStyles;
+
+  const useStrings = () => {
```
edwardlee-msft, line 52: This is used to resolve issues with conditional compilation complaining that CallCompositeStrings do not have the necessary strings. Once out of beta, this will be changed to only use CallCompositeStrings.

packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -173,4 +173,49 @@ export interface CallCompositeStrings {
...
+  /* @conditional-compile-remove(call-composite-participant-pane) */
+  /**
+   * control bar People button label
+   */
+  peopleButtonLabel: string;
+  /* @conditional-compile-remove(call-composite-participant-pane) */
+  /**
+   * control bar Chat button label.
+   */
+  chatButtonLabel: string;
```
edwardlee-msft, line 200: chatButtonlabel is used in TabHeader.tsx which was made into a common file. This could be resolved if we have a common set of strings, otherwise, this will be needed in CallCompositeStrings as well. 

For now, as this is in beta, I have placed it in CallCompositeStrings. This will need to be discussed further at a later date.

carocao-msft (approved):

common/config/babel/.babelrc.js
```diff
@@ -34,7 +34,9 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         // dialpad
         'dialpad',
         // rooms
-        'rooms'
+        'rooms',
+        // Participant pane in the `CallComposite`.
+        'call-composite-participant-pane'
```
anjulgarg, line 39: This work is part of the 1-to-n-calling and has a tag already defined in babelrc :)

edwardlee-msft: Ah ha, I will retag everything to use  "one-to-n-calling"!

packages/react-composites/src/composites/CallComposite/components/CallPane.tsx
```diff
@@ -0,0 +1,195 @@
...
+const sidePaneTokens: IStackTokens = {
+  childrenGap: '0.5rem'
+};
+
+const modalPipRightPositionPx = 16;
+const modalPipTopPositionPx = 52;
+const modalPipWidthPx = 88;
+const modalPipHeightPx = 128;
+
+const getPipStyles = (theme: ITheme): ModalLocalAndRemotePIPStyles => ({
```
anjulgarg, line 184: We have `common/styles` directory for holding styles common across the composites.
Possible to keep a single copy of Pip styles there and share between call and call with chat?

packages/react-composites/src/composites/CallComposite/components/CallPane.tsx
```diff
@@ -0,0 +1,195 @@
...
+/** @beta */
+export type CallPaneOption = 'none' | 'people';
+
+const hiddenStyles: IStackStyles = {
+  root: {
+    display: 'none'
+  }
+};
+
+const sidePaneStyles: IStackStyles = {
```
anjulgarg, line 165: Possible to re-use these styles by moving them to `common/styles`?

packages/react-composites/src/composites/CallComposite/components/CallPane.tsx
```diff
@@ -0,0 +1,195 @@
...
+      />
+    </CallAdapterProvider>
+  );
+
+  // Use document.getElementById until Fluent's Stack supports componentRef property: https://github.com/microsoft/fluentui/issues/20410
+  const modalLayerHostElement = document.getElementById(props.modalLayerHostId);
+  const modalHostRef = useRef<HTMLElement>(modalLayerHostElement);
+  const modalHostWidth = _useContainerWidth(modalHostRef);
+  const modalHostHeight = _useContainerHeight(modalHostRef);
+  const minDragPosition: _ICoordinates | undefined = useMemo(
```
anjulgarg, line 98: Possible to move these util functions to `common/utils.tsx` and re-use them in both call and callwithchat?
I feel uneasy about duplicating such complex logic

packages/react-composites/src/composites/CallComposite/components/CallPane.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  const dataUiId = props.activePane === 'people' ? 'call-composite-people-pane' : '';
+
+  return (
+    <Stack verticalFill grow styles={paneStyles} data-ui-id={dataUiId} tokens={props.mobileView ? {} : sidePaneTokens}>
+      {header}
+      <Stack.Item verticalFill grow styles={paneBodyContainer}>
+        <Stack horizontal styles={scrollableContainer}>
+          <Stack.Item verticalFill styles={scrollableContainerContents}>
+            <Stack styles={props.activePane === 'people' ? availableSpaceStyles : hiddenStyles}>{peopleContent}</Stack>
```
anjulgarg, line 129: assigning a JSX element to `peopleContent` doesn't provide much value here since it is the only jsx being rendered here.
I would recommend moving the JSX inline.

edwardlee-msft: Great point! Moved it to a JSX inline.

packages/react-composites/src/composites/CallComposite/components/CallPane.tsx
edwardlee-msft: Moved to `common/styles/Pane.styles.ts`

packages/react-composites/src/composites/CallComposite/components/CallPane.tsx
edwardlee-msft: Moved to common/styles/Pane.styles.ts

packages/react-composites/src/composites/CallComposite/components/CallPane.tsx
edwardlee-msft: Good point, I moved it out into a util file so that it can be used in both CallPane and CallWithChatPane.
`common/utils.tsx`

edwardlee-msft: Functioning correctly in CallWithChat after moving to `common/utils.tsx`
https://user-images.githubusercontent.com/97130533/180077645-17778263-9bb6-4e1d-b3dc-2448e007e68a.mov

packages/react-composites/src/composites/common/utils.tsx
```diff
@@ -0,0 +1,53 @@
...
+  maxDragPosition: _ICoordinates | undefined;
+}
+
+/**
+ * @private
+ */
+// Use document.getElementById until Fluent's Stack supports componentRef property: https://github.com/microsoft/fluentui/issues/20410
+export const useMinMaxDragPosition = (
+  rtl: boolean | undefined,
+  modalLayerHostElement: HTMLElement | null
```
anjulgarg, line 22: Should both rtl and modalLayerHostElement be required params?
I am not sure if this hook offers value if used without these params?

anjulgarg: If yes, I would modify the signature to be 

`useMinMaxDragPosition = (modalLayerHostElement: HTMLElement, rtl = false)`

edwardlee-msft: Good point, rtl is not needed and can be optional. 
modaLayerHostElement will be needed though. Will make that change! Good catch, thanks Anjul!

anjulgarg (approved):

PR #2104, opened 2022-07-20
Add component behaviors according to permissions

PorterNan:
# What

- Add support for localPreview
- Disable mic and camera selection when not permitted
- Disable buttons when there is no permissions

# Why

# How Tested

# Process & policy checklist


alkwa-msft (approved):

alkwa-msft:
This works for what we need to do. Are we potentially putting in more logic into each UI component especially about how the UI component is being used which is changing how we determine if this should be disabled?

mgamis-msft (approved):

prabhjot-msft (approved):

PorterNan:
> This works for what we need to do. Are we potentially putting in more logic into each UI component especially about how the UI component is being used which is changing how we determine if this should be disabled?

Yep, all these logic currently should live in a component if that is only a component behavior - and we can also do something in composite level to hide a component directly, I would say we might need to keep all these logics within 2-3 lines of code for disable/enable so that it would be easy to migrate when we have all state from calling sdk

PR #2105, opened 2022-07-20
Refresh end-to-end documentation

prprabhu-ms:
How we run, write and organize these tests has changed. Existing documentation wasn't very ergonomic for new engineers and had bitrotted.


packages/react-composites/tests/README.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+### Mobile Only tests
+
+If you are writing a test for only on Mobile make sure to add it to a test suite that is just for mobile. They will be marked with `[Mobile Only]` in the suite title. If there is not a suite for the page you are testing add one for that page with the `[Mobile only]` in the title.
+
+Once you have added your test to the appropriate suite use the following call to make sure it is not run on the desktop project:
+
+```Typescript
+test.only('Your test name here', async ({ pages }, testInfo) => {
```
prprabhu-ms, line 125: replace `test.only` with `test`.

docs/references/automated-tests.md
```diff
@@ -0,0 +1,46 @@
...
+The following packlets contain browswer tests:
+
+* [packages/react-composites](../../packages/react-composites/tests/README.md)
+* [samples/tests](../../samples/tests/README.md)
+
+#### Hermetic tests vs Live Tests
+
+Broswer tests come in two varieties:
+
+* *Live tests* load up a test application that works with actual Azure Communication Services backend services. These are slow to run, must be run serially (for throttling reasons),a nd tend to be flaky. But they test the UI against live services, giving us high confidence when they pass.
```
prprabhu-ms, line 26: typo: a nd -> and

packages/react-composites/tests/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+This folder contains some hermetic and live browser tests for the composites. See the [top-level testing documentation](../../../docs/references/automated-tests.md) to understand how these tests fit into our overall testing strategy.
+
+## Folder structure
+
+Tests for each composite are contained in their own folder. Further, hermetic and live tests for each composite are contained in separate sub-folders.
+
+* [./browser/call](./browser/call) - Tests for `CallComposite`
+  * [./browser/call/app](./browser/call/app) - Test application used for all `CallComposite` tests
+  * [./browser/call/hermetic](./browser/call/hermetic) - Hermetic tests for `CallComposite`
```
anjulgarg, line 11: What's the difference b/w hermetic and fake-adapter?
Would call not have fake-adapter tests as well?

prprabhu-ms: fake-adapter is *how* we make sure the test are hermetic. _Hermetic_ better describes what's special about these tests.

For example, the tests could also be made hermetic via a different kind of fake - a full in-memory service implementation, or a record-replay type implementation etc.

anjulgarg (approved):

dmceachernmsft (approved):

PR #2106, opened 2022-07-20
Display chat message bubble border in high contrast mode

anjulgarg:
# What
Add border around messages in high contrast mode

image
image

# Why
To resolve the bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2864691

# How Tested
Manually on a windows device

# Process & policy checklist


packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -91,7 +91,10 @@ export const defaultChatItemMessageContainer = (overlapAvatarAndMessage: boolean
     marginRight: '0rem',
     marginLeft: `${messageAvatarGap}rem`,
     width: `calc(100% - ${AVATAR_WIDTH_REM + MESSAGE_AMOUNT_OUT_FROM_EDGE_REM + messageAvatarGap}rem)`,
-    zIndex: CHAT_MESSAGE_ZINDEX
+    zIndex: CHAT_MESSAGE_ZINDEX,
+    '@media (forced-colors: active)': {
+      forcedColorAdjust: 'none'
```
dmceachernmsft, line 96: does stopping forced colours maintain the required colour ratio for the different modes? @edwardlee-msft I believe its like 3:1 or something for items like this? or was it 5:1?

anjulgarg: How can we check that?
Teams app also shows the same colors in high contrast modes.

dmceachernmsft: I use the accessibility insights tool for windows, you can download it here: https://accessibilityinsights.io/docs/windows/overview/

anjulgarg: Updated the description with screenshots from the tool. LMKWYT

JamesBurnside: For me Teams has the same bug where messages have no border:
image

JamesBurnside (commented):
Wihout seeing any concrete documentation for how to style for high contrast modes, I think we should be leaving the background to be set by the high contrast settings - the background can be set to any color in the system settings:
image (so our colors might be broken then if we force them to not use the active colors)

Is it possible to leave the colors as is but provide a border around the message in forced-colors:active?

Otherwise leave as is and let the a11y team decide if the fix is sufficient!

JamesBurnside (commented):
Wihout seeing any concrete documentation for how to style for high contrast modes, I think we should be leaving the background to be set by the high contrast settings - the background can be set to any color in the system settings:
image (so our colors might be broken then if we force them to not use the active colors)

Is it possible to leave the colors as is but provide a border around the message in forced-colors:active?

Otherwise leave as you have it and let the a11y team decide if the fix is sufficient!

packages/react-components/src/components/styles/MessageThread.styles.ts
anjulgarg: Correct.

anjulgarg:
@JamesBurnside Great recommendations. Added borders during high contrast mode instead of forcing bg colors. This is the best fix IMO

packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -101,7 +101,9 @@ export const defaultChatItemMessageContainer = (overlapAvatarAndMessage: boolean
 export const defaultMyChatMessageContainer: ComponentSlotStyle = {
   maxWidth: '100%',
   minWidth: `${CHAT_MESSAGE_CONTAINER_MIN_WIDTH_REM}rem`,
-  marginLeft: '0rem'
+  marginLeft: '0rem',
+  // This makes message bubble show border in high contrast mode making each message distinguishable
+  border: '1px solid transparent'
```
edwardlee-msft, line 106: note: This would introduce "thicker" messages by default but it is an elegant and probably unnoticeable change by users :)

edwardlee-msft (approved):

packages/react-components/src/components/styles/MessageThread.styles.ts
edwardlee-msft: On another note, can someone test this on a 4k/high res display? I'm curious if the 1px will cause issues

dmceachernmsft (approved):

PR #2108, opened 2022-07-20
Create common composite strings for CallWithChat and Call Composite

edwardlee-msft:
# What
Create a common set of strings to be used by both CallWithChat and Call Composite.

# Why
Brought about when converting multiple files to Common files related to participant pane from CallWithChat.
API Changes are minimal, but are present, hence I would love some feedback and your expertise @JamesBurnside, @prprabhu-ms, and @anjulgarg. 

# How Tested
Tested Locally on MacOS on both Desktop and Mobile. No issues seen.

# Process & policy checklist

### DIFF
--- a/packages/communication-react/review/stable/communication-react.api.md
+++ b/packages/communication-react/review/stable/communication-react.api.md
@@ -372,7 +372,7 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
 }
 
 // @public
-export interface CallCompositeStrings {
+export interface CallCompositeStrings extends CommonCompositeStrings {
     cameraLabel: string;
     cameraPermissionDenied: string;
     cameraTurnedOff: string;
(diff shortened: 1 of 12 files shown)


packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -372,7 +372,7 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
 }
 
 // @public
-export interface CallCompositeStrings {
+export interface CallCompositeStrings extends CommonCompositeStrings {
```
JamesBurnside, line 375: Ordinarily this is exactly what we should do! But the AzureReviewBoard have asked us to keep interfaces as flat as possible so for intellisense and when a developer looks at the type they get the whole interface and not have to find the extended interface and look at that as well

edwardlee-msft: Ah interesting! That's good to know, thanks James! 
It would have definitely made common composite files easier to work with if we had a common set of strings to access, that is unfortunate.

PR #2110, opened 2022-07-21
Stop building `@internal/storybook` on stable flavor builds

prprabhu-ms:
After this PR:

* `rush build`, `rush test` and `rush lint` are silently skipped for `stable` build of storybook
* `rush deploy-storybook` and `rush start` fail immediately for `stable` build of storybook.

# Why

We build storybook only against the `beta` build flavor. This creates problems with stable release branches:

* CI builds start failing for storybook
* We risk deploying the incorrect build of storybook if it somehow compiles correctly.

This PR addresses both those concerns, and ensures that local development experience for `stable` build is consistent with expecations - it is not possible to develop storybook for `stable` builds.


common/scripts/lib/exec.mjs
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @param cmd: string - command to run.
+ * @param env: Optional environment variables for the child process. By default, process.env is forwarded.
+ *
+ * - stdout, stderr are piped to the current process' stdout and stderr respectively.
+ * - the returned Promise is rejected if the child process exits with a non-zero exit code.
+ */
+export async function exec(cmd, env) {
+    console.log(`Running ${cmd}`);
+    // Inheriting the stdio (and implied stderr and stdin) ensures that colorized output is preserved.
```
JamesBurnside, line 17: neat!

JamesBurnside (approved):

alkwa-msft (approved):

PR #2114, opened 2022-07-21
Add Dialpad to Calling and CallWithChat sample apps

dmceachernmsft:
# What

Update application home pages to have dial pad and create a different selection to do 1:N calling
# Why

Allows us to demonstrate how to use the Dialpad component to start calls with our composites. 

Splitting the screens allows us to drive simplicity to starting a call so there is less confusion on what information you need to provide to start what call

https://skype.visualstudio.com/SPOOL/_workitems/edit/2920799
# How Tested

Ran locally on the sample apps to verify dialpad works. will test sample apps with 1:N POC to verify ACS calling.


samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -145,15 +180,29 @@ export const HomeScreen = (props: HomeScreenProps): JSX.Element => {
...
+
+/* @conditional-compile-remove(PSTN-calls) */
+/**
+ * Parse the unformatted string that is given from the dialpad to something that the
+ * Calling adapter will accept as a PSTN participant.
+ *
+ * @param participant - string stored from dialpad inputs.
+ * @returns formatted phone number to be added to call locator.
+ */
+const parseDialPadParticipant = (participant?: string): string[] | undefined => {
```
anjulgarg, line 203: This should ideally be happening in the dialpad itself.
I can't think of a scenario where a formatted string returned by the dialpad is useful.

carocao-msft (approved):

samples/Calling/src/app/styles/HomeScreen.styles.ts
```diff
@@ -59,6 +59,17 @@ export const buttonStyle = mergeStyles({
...
   padding: '0.625rem'
 });
+/* @conditional-compile-remove(PSTN-calls) */
 export const outboundtextField = mergeStyles({
   paddingTop: '0.5rem'
 });
+/* @conditional-compile-remove(PSTN-calls) */
+export const addDialparticipantStyles = mergeStyles({
+  fontWeight: 600,
+  fontSize: '0.875rem', // 14px
```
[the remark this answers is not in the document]
anjulgarg: Not sure why we need to provide font size here?
Isn't the default font size working?

samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -100,20 +119,40 @@ export const HomeScreen = (props: HomeScreenProps): JSX.Element => {
...
+                      placeholder={"Comma seperated ACS user ID's"}
+                      onChange={(_, newValue) => setOutboundParticipants(newValue)}
+                    />
+                  </Stack>
+                </Stack>
+              )
+            }
+            {
+              /* @conditional-compile-remove(PSTN-calls) */ pstnCallChosen && (
+                <Stack>
```
anjulgarg, line 137: Why the 2 stacks? Seem redundant

anjulgarg (approved):

samples/Calling/src/app/views/HomeScreen.tsx
dmceachernmsft: Hold over from previous version, will remove

PR #2115, opened 2022-07-21
Callwithchat mobile: dismissed drawer after device selection

carocao-msft:
# What
Callwithchat mobile: make drawer dismissed after device selection

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2896714

# Test

https://user-images.githubusercontent.com/96077406/180506794-dc3039f5-ba78-44ec-8a08-b4016bdcce53.mov

https://user-images.githubusercontent.com/96077406/180510270-94f6d812-78d5-4773-a0ce-3409ed8e2db1.mov


JamesBurnside:
Two things to consider:
* The mobile drawer is also used when a participant is clicked in callwithchat - can either add the same manual close logic to when they are clicked too, or consider instead changing the `_DrawerMenu` itself to trigger close when a menu item (that doesn't have a sub menu) is clicked
* We have some UI tests that may need updating if the drawer dismisses on click

carocao-msft:
> Two things to consider:
> 
> * The mobile drawer is also used when a participant is clicked in callwithchat - can either add the same manual close logic to when they are clicked too, or consider instead changing the `_DrawerMenu` itself to trigger close when a menu item (that doesn't have a sub menu) is clicked
> * We have some UI tests that I think will need the test checks updated to reopen the drawer if the drawer dismisses on click

Updated PR with screen recording

JamesBurnside (approved):

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
```diff
@@ -107,8 +107,9 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
         // But this is a known issue in our state.
         onSelectSpeaker(selected as AudioDeviceInfo);
       }
+      props.onLightDismiss();
     },
-    [speakers, onSelectSpeaker]
+    [speakers, onSelectSpeaker, props]
```
JamesBurnside, line 112: avoid `props` as this means any prop change will cause this to be recreated.
Instead update the above `const { speakers, onSelectSpeaker } = props;`
to have the onLightDismiss:
`const { speakers, onSelectSpeaker, onLightDismiss } = props;`
and put `onLightDismiss` into the dependency array

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
```diff
@@ -142,8 +143,9 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
         // But this is a known issue in our state.
         onSelectMicrophone(selected as AudioDeviceInfo);
       }
+      props.onLightDismiss();
     },
-    [microphones, onSelectMicrophone]
+    [microphones, onSelectMicrophone, props]
```
JamesBurnside, line 148: (nit: same here on avoiding props in the dep array)

edwardlee-msft: ^^ I second this as well

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
edwardlee-msft: ^^ I second this as well!

edwardlee-msft (approved):

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
carocao-msft: updated :)

packages/react-composites/src/composites/CallWithChatComposite/components/MoreDrawer.tsx
carocao-msft: updated :)

PR #2116, opened 2022-07-21
DO NOT REVIEW: How to conditionally compile less for fun and profit

prprabhu-ms:
# What

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/file-sharing/FileUploadButton.tsx
```diff
@@ -2,6 +2,9 @@
 // Licensed under the MIT license.
 import { IconButton, mergeStyles, Stack, useTheme } from '@fluentui/react';
 import React from 'react';
+import { CallCompositeStrings } from '../../CallComposite';
```
prprabhu-ms, line 5: Heh, I have nooooo idea why I was mucking around in `FileUploadButton.tsx`  ;)

PR #2117, opened 2022-07-21
Add a background color to all composites

anjulgarg:
# What
Add a background color to all composites using the `theme.semanticColors.backgroundColor` property.

# Why
To resolve the bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2903659

# How Tested
Manually

# Process & policy checklist


JamesBurnside:
We need another PR/extra change here to fully resolve this bug.
Currently the call control bar and the message thread both have a background that is different from semanticColors.background:
image

image

packages/react-components/src/components/ControlBar.tsx
```diff
@@ -66,10 +66,9 @@ export const ControlBar = (props: ControlBarProps): JSX.Element => {
   const theme = useTheme();
   const controlBarClassName = useMemo(() => {
     const controlBarStyle = controlBarStyles[layout ?? 'horizontal'];
-    // if theme is dark and layout is floating then use palette.neutralQuaternaryAlt as background, otherwise use palette.white
+    // if theme is dark and layout is floating then use palette.neutralQuaternaryAlt as background, otherwise use the composite background color
```
JamesBurnside, line 69: nit: this is a UI component so we shouldn't mention composite here

packages/react-composites/src/composites/common/BaseComposite.tsx
```diff
@@ -93,7 +102,9 @@ export const BaseProvider = (
   const CompositeElement = (
     <FluentThemeProvider fluentTheme={fluentTheme} rtl={rtl}>
       <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=0" />
-      <Customizer scopedSettings={{ Layer: { hostId: globalLayerHostId } }}>{props.children}</Customizer>
+      <Customizer scopedSettings={{ Layer: { hostId: globalLayerHostId } }}>
```
prprabhu-ms, line 105: tangential: What is `Customizer`? That's a very enigmatic and not very useful name.

packages/react-composites/src/composites/common/BaseComposite.tsx
```diff
@@ -110,3 +121,25 @@ const BaseContext = createContext<boolean>(false);
...
+/**
+ * @private
+ * Provides a wrapper with a background color to ensure that composites always have a background color.
+ * This is necessary to ensure that composites are not transparent,
+ * and the background color of it's parent elements doesn't show through the composite.
+ */
+const WithBackgroundColor = (props: { children: React.ReactNode }): JSX.Element => {
+  const { children } = props;
+  const theme = useTheme();
+  const className = React.useMemo(
```
prprabhu-ms, line 134: `mergeStyles()` returns a string, and the computation is not costly. I would advise against memoization in this case.

prprabhu-ms (approved):
We should add a hermetic snapshot test with non-white background before closing the associated bug.

packages/react-composites/src/composites/common/BaseComposite.tsx
anjulgarg: @PorterNan Might know more about this. Customizer was already here.

alkwa-msft (approved):

PR #2120, opened 2022-07-22
Update babelrc to maintain file lines in preprocess for sample apps

dmceachernmsft:
# What

Update `package.json` and `babelrc` in all samples
# Why

generates a nice preprocessed project to help maintain hero apps and allows for conditional compilation of new features in apps

https://skype.visualstudio.com/SPOOL/_workitems/edit/2930429
# How Tested

Tested on branch


samples/CallWithChat/.babelrc.js
```diff
@@ -1,5 +1,15 @@
...
-  ...commonConfig
+  ...commonConfig,
+  /**
+   * retainLines is needed to maintain the lines of the comments in the output to preprocess. This will cause in the output
+   * the comments assigned to their original position in the file so any linter options will be maintained on their respective node.
+   *
+   * this is caused by the comments being both trailing and leading to the nodes either above or below it, regardless
+   * of which node it is touching on the next line.
+   *
+   * link to babel docs on this: https://babeljs.io/docs/en/options#retainlines
```
prprabhu-ms, line 12: <3

The documentation says it is best effort, but that is good enough for us as we're using it for source-to-source transformation that we can (to some degree) eyeball as necessary.

samples/Chat/package.json
```diff
@@ -18,7 +19,9 @@
     "webpack": "rushx _by-flavor \"webpack --mode=production --env production\"",
     "_current-flavor": "echo You are running under COMMUNICATION_REACT_FLAVOR: && env-cmd -f ../../common/config/env/.env node -p process.env.COMMUNICATION_REACT_FLAVOR",
     "_by-flavor": "rushx _current-flavor && env-cmd -f ../../common/config/env/.env --use-shell",
-    "_typecheck": "rushx _by-flavor \"if-env COMMUNICATION_REACT_FLAVOR=stable && echo skip || (if-env COMMUNICATION_REACT_FLAVOR=beta && tsc)\""
+    "_stable-flavor": "rushx _current-flavor && env-cmd -f ../../common/config/env/stable.env --use-shell",
+    "_typecheck": "rushx _by-flavor \"if-env COMMUNICATION_REACT_FLAVOR=stable && echo skip || (if-env COMMUNICATION_REACT_FLAVOR=beta && tsc)\"",
+    "_preprocess:stable-flavor": "rushx _stable-flavor \"rushx preprocess && rushx lint:fix && rushx prettier\""
```
prprabhu-ms, line 24: These `package.json` files are getting out of hand with the number of scripts. The problem is:

* Helper scripts like "_typecheck" create noise
* There is no documentation, help message -- nothing ergonomic.

I'd like us to start adding `scripts/*mjs` instead that are invoked from `package.json`. See https://github.com/Azure/communication-ui-library/blob/main/packages/react-composites/scripts/runBrowserTests.mjs for example.

I'd recommend merging this PR as is, and then iterating on top to extract some helper scripts with better help / documentation.

samples/StaticHtmlComposites/package.json
```diff
@@ -12,10 +12,13 @@
...
     "lint": "echo skip",
     "lint:fix": "echo skip",
+    "preprocess": "cpx \"./src/**\" ./preprocessed && babel ./src --out-dir ../preprocessed --extensions \".ts,.tsx\" --keep-file-extension --config-file ./.babelrc.js --relative",
     "prettier": "",
     "prettier:check": "",
     "_current-flavor": "echo You are running under COMMUNICATION_REACT_FLAVOR: && env-cmd -f ../../common/config/env/.env node -p process.env.COMMUNICATION_REACT_FLAVOR",
-    "_by-flavor": "rushx _current-flavor && env-cmd -f ../../common/config/env/.env --use-shell"
+    "_by-flavor": "rushx _current-flavor && env-cmd -f ../../common/config/env/.env --use-shell",
+    "_stable-flavor": "rushx _current-flavor && env-cmd -f ../../common/config/env/stable.env --use-shell",
+    "_preprocess:stable-flavor": "rushx _stable-flavor \"rushx preprocess && rushx lint:fix && rushx prettier\""
```
prprabhu-ms, line 21: Why is this script prefixed with `_`? If I understand correctly, this is intended to be used by developers to generate the hero sample code. The `_` prefix is for "helper scripts" that are not intended to be directly invoked.

prprabhu-ms (approved):

samples/StaticHtmlComposites/package.json
dmceachernmsft: Will update the naming here for preprocess 😊

anjulgarg (approved):

samples/Chat/package.json
dmceachernmsft: I'm not quite sure what is happening in the file you are linking to. Could we set up a meeting about this either tuesday or when I get back on the 3rd? is this a script file that depending on the flags you give as args runs different command line functions like you would see in package.json?

PR #2121, opened 2022-07-22
Add event logs for disposing local video stream

JamesBurnside:
# What
Add event trace logs for disposing local video stream

# Why
We have them for remote video streams and for creating local video streams, but not for disposing local video streams.

# How Tested
I did not as this is just adding logs - please review thoroughly


carocao-msft (approved):

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -446,24 +446,58 @@ function disposeViewRemoteVideo(
...
     return;
   }
 
   // Sets the status and also renderer. I think we need to always set renderer to undefined since in all status when
   // cleaned up should have renderer as undefined. If the status is 'Rendered' and renderer is not defined it should
   // be cleaned up below so we can set it to undefined here.
   if (renderInfo.status === 'Rendering') {
+    _logEvent(callingStatefulLogger, {
+      name: EventNames.LOCAL_STREAM_STOPPING,
+      level: 'warning',
```
prprabhu-ms, line 477: Why is this a warning?

Stopping a stream that is in the process of being started is normal, right? It just means that the user clicked the video toggle button too soon.

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -446,24 +446,58 @@ function disposeViewRemoteVideo(
...
+      name: EventNames.DISPOSING_LOCAL_RENDERER,
+      level: 'info',
+      message: 'Disposing local view renderer.',
+      data: streamLogInfo
+    });
     renderInfo.renderer.dispose();
+  } else {
+    _logEvent(callingStatefulLogger, {
+      name: EventNames.LOCAL_RENDERER_NOT_FOUND,
+      level: 'warning',
```
prprabhu-ms, line 497: Shouldn't this also be an `error`, similar to line 464?

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -472,21 +506,48 @@ function disposeViewUnparentedVideo(
...
+      level: 'error',
+      message: 'Cannot find render info when disposing unparented local stream.'
+    });
     return;
   }
 
   if (renderInfo.status === 'Rendering') {
+    _logEvent(callingStatefulLogger, {
+      name: EventNames.LOCAL_STREAM_STOPPING,
+      level: 'warning',
```
prprabhu-ms, line 530: Same comment as above.

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -472,21 +506,48 @@ function disposeViewUnparentedVideo(
...
+    _logEvent(callingStatefulLogger, {
+      name: EventNames.DISPOSING_LOCAL_RENDERER,
+      level: 'info',
+      message: 'Disposing unparented local view renderer.'
+    });
     renderInfo.renderer.dispose();
+  } else {
+    _logEvent(callingStatefulLogger, {
+      name: EventNames.LOCAL_RENDERER_NOT_FOUND,
+      level: 'warning',
```
prprabhu-ms, line 548: Same comment as above.

It's unfortunate we have duplicate code-paths for such closely related non-trivial operations. It's an invitation to bugs...

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -511,7 +572,11 @@ export function createView(
     // Render LocalVideoStream that is not part of a Call
     return createViewUnparentedVideo(context, internalContext, stream, options);
   } else {
-    console.warn('Invalid combination of parameters');
+    _logEvent(callingStatefulLogger, {
+      name: EventNames.CREATE_STREAM_INVALID_PARAMS,
+      level: 'warning',
```
prprabhu-ms, line 577: I agree with this being a `warning`. Client used the API incorrectly, not an internal error.

prprabhu-ms (approved):

packages/calling-stateful-client/src/StreamUtils.ts
JamesBurnside: Good note, updating here and remote rendering ops to be info

packages/calling-stateful-client/src/StreamUtils.ts
JamesBurnside: Yep I agree, updating here and remote rendering ops to be error

packages/calling-stateful-client/src/StreamUtils.ts
JamesBurnside: updated

packages/calling-stateful-client/src/StreamUtils.ts
JamesBurnside: updated, yeah I agree on the redundancy but also can't commit time to making it better, the code is so similar we should be able to have a local | remote type and keep it somewhat agnostic

PR #2125, opened 2022-07-25
Check if locator in CallComposite preview is a UUID before joining group call

mgamis-msft:
# What

Check if locator in CallComposite preview is a UUID before joining group call.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2925571

# How Tested

Tested joining teams call, group call, and room on locally deployed storybook.

# Process & policy checklist


edwardlee-msft (approved):

packages/storybook/stories/CallComposite/snippets/Container.snippet.tsx
```diff
@@ -23,15 +23,16 @@ export type ContainerProps = {
 };
 
 const isTeamsMeetingLink = (link: string): boolean => link.startsWith('https://teams.microsoft.com/l/meetup-join');
-const isGroupLink = (link: string): boolean => link.indexOf('-') !== -1;
+const uuidRegexExp = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/gi;
```
PorterNan, line 26: Looks like we have "uuid" dependency in this folder, can we do 

import uuid from 'uuid';
...
uuid.validate(str);

PorterNan (approved):
Left a small comment so that we don't need a regex string here, feel free to check this in

packages/storybook/stories/CallComposite/snippets/Container.snippet.tsx
```diff
@@ -23,15 +24,15 @@ export type ContainerProps = {
...
-  } else if (isGroupLink(link)) {
-    return { groupId: link };
+const createCallAdapterLocator = (locator: string): CallAdapterLocator => {
+  if (isTeamsMeetingLink(locator)) {
+    return { meetingLink: locator };
+  } else if (isGroupID(locator)) {
+    return { groupId: locator };
   }
-  return { roomId: link };
+  return { roomId: locator };
```
JamesBurnside, line 35: (non-blocking comment)
is there a way to test if it is a roomId?
Would be nice to do:
```
if (isTeamsLink()) return { meetingLink: loc }
else if (isGroupId()) return { groupId: loc }
else if (isRoomsId()) return { roomId: loc }
else {
  throw new Error(`Unrecognized locator: ${locator}`);
}
```

JamesBurnside: To ensure if we start supporting a new locator this can't assume its a roomsId

PR #2128, opened 2022-07-26
Create GitHub issue if tdbuild trigger fails

prprabhu-ms:
# Why

We didn't notice that the action was failing for 8 days: 
image

Also, this action is very robust and almost never fails except when the PAT expires. So we don't anticipate false positives.

# Validation

Injected a failure in the action and ran https://github.com/Azure/communication-ui-library/actions/runs/2737864358
which failed and created the issue https://github.com/Azure/communication-ui-library/issues/2129


prprabhu-ms:
Assigned the previous and next OCE ;)

.github/workflows/run-td-build.yml
```diff
@@ -43,3 +43,25 @@ jobs:
...
+      - name: Create GitHub issue
+        id: create-issue
+        run: |
+          # check for an issue that is already open
+          curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/search/issues?q=org:Azure+repo:communication-ui-library+label:TDBUILD_FAILURE+state:open > $HOME/tdbuild_issues.json
+          issue_count=$(jq -r '.total_count' $HOME/tdbuild_issues.json)
+          if [ $issue_count -gt 0 ]; then
+            echo "No need to create new issue, one already exists"
+          else
+            curl -X POST -H 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' https://api.github.com/repos/Azure/communication-ui-library/issues -d '{"title":"Action to trigger TD Build failed", "body": "Please investigate failures of https://github.com/Azure/communication-ui-library/actions/workflows/run-td-build.yml", "labels":["TDBUILD_FAILURE"]}'
```
JamesBurnside, line 64: Any way we can link to this build that failed? I know for the main CI fail we don't and it would be really really helpful to put a link to the failed build

JamesBurnside (approved):

.github/workflows/run-td-build.yml
prprabhu-ms: I got lazy and just linked to the actions page (you see all the latest builds, hopefully including the red ones)

prabhjot-msft (approved):

PR #2131, opened 2022-07-26
fix: joinCall set correct mute state based on microphoneOn parameter

bfwg:
# What
Fixed a bug where the initial muted state is not being set correctly. 

# Why
When `microphoneOn` is true, `muted` will be true which is backward of what the code is intended to do. 

# How Tested
I tested calling `joinCall` with `microphoneOn` parameter `true`  and the call will start muted which is incorrect.

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -299,7 +299,7 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
     }
 
     return this.teeErrorToEventEmitter(() => {
-      const audioOptions: AudioOptions = { muted: microphoneOn ?? !this.getState().isLocalPreviewMicrophoneEnabled };
+      const audioOptions: AudioOptions = { muted: !microphoneOn ?? !this.getState().isLocalPreviewMicrophoneEnabled };
```
prprabhu-ms, line 302: `microphoneOn` is optional. If we check for falsiness, calling `joinCall()` without any arguments will cause `muted` to be set to `true`.

So, fallback on `undefined`, and *then* check for falsiness...

```suggestion
      const audioOptions: AudioOptions = { muted: !(microphoneOn ?? this.getState().isLocalPreviewMicrophoneEnabled) };
```

prprabhu-ms (changes requested):
Thank you for your contribution!

I checked in the rest of the codebase and turns out we don't ever call `joinCall` with `microphoneOn` set (to either `true` or `false`). That is why things were not on fire already.

Thanks for catching this.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
bfwg: Great suggestion. I've fixed my code.

JamesBurnside (approved):

prprabhu-ms (approved):

prprabhu-ms:
Do you mind running `rush changelog` and providing a changelog for this PR?

The type should be "patch" (as this is a bugfix). The PR title is a good change description.

prprabhu-ms:
I'm trying to figure out why chromatic is unhappy.

bfwg:
Done.

PR #2136, opened 2022-07-27
Resolve ModalPip boundary issue in callpane in Call Composite

edwardlee-msft:
# What
Move LayerHost for ModalLocalAndRemotePIP outside of CallArrangement into CallComposite file

Boundaries properly set 

https://user-images.githubusercontent.com/97130533/181358663-73aef1ce-3b27-4d06-86dd-edb785091f26.mov

# Why
Resolve boundary issue with pip when in mobile view with people pane open. 
LayerHost needs to be rendered before useMinMaxDragPosition in common/utils is called. 
Cannot be in CallArrangement as previously done as CallArrangement is rendered when callStatus is connected, causing useRef reference to the ModalLayerHost to be initialized as null.

Clipping and boundaries not properly set

https://user-images.githubusercontent.com/97130533/181358591-c6cbf8d0-f824-494d-985c-d43ce2506467.mov

# How Tested
Test locally on MacOS on Chrome mobile view.

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/CallComposite.tsx
+++ b/packages/react-composites/src/composites/CallComposite/CallComposite.tsx
@@ -252,7 +252,7 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
             // the Modal because the draggable bounds thinks it has no space and will always return to its initial position after dragging.
             // Additionally, this layer host cannot be in the Call Arrangement as it needs to be rendered before useMinMaxDragPosition() in
             // common/utils useRef is called.
-            // This modal needs to be called before a call gets connected.
+            // Warning: this is fragile and works because the call arrangement page is only rendered after the call has connected and thus this LayerHost will be guaranteed to have rendered (and subsequently mounted in the DOM). This ensures the DOM element will be available before the call to `document.getElementById(modalLayerHostId)` is made.
             /* @conditional-compile-remove(one-to-n-calling) */
             mobileView && <LayerHost id={modalLayerHostId} className={mergeStyles(modalLayerHostStyle)} />
           }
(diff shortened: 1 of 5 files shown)


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -220,10 +240,22 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
...
             /* @conditional-compile-remove(rooms) */
             role={role}
           />
+          {
+            // This layer host is for ModalLocalAndRemotePIP in CallPane. This LayerHost cannot be inside the CallPane
+            // because when the CallPane is hidden, ie. style property display is 'none', it takes up no space. This causes problems when dragging
+            // the Modal because the draggable bounds thinks it has no space and will always return to its initial position after dragging.
+            // Additionally, this layer host cannot be in the Call Arrangement as it needs to be rendered before useMinMaxDragPosition() in
+            // common/utils useRef is called.
+            // This modal needs to be called before a call gets connected.
```
JamesBurnside, line 255: I would reword this - it reads as if it is being impacted by the connected state of the call explicitly instead of relying on the call state to ensure rendering order. Maybe something like:
```suggestion
            // Warning: this is fragile and works because the call arrangement page is only rendered after the call has connected and thus this LayerHost will be guaranteed to have rendered (and subsequently mounted in the DOM). This ensures the DOM element will be available before the call to `document.getElementById(modalLayerHostId)` is made.
```

JamesBurnside (approved):

carocao-msft (approved):

anjulgarg (approved):

PR #2141, opened 2022-07-29
CI: Build packlets used by browser tests

prprabhu-ms:
# Why

The browser tests use dependencies of `@internal/react-composites`  in the *test code*. Before this PR, the CI actions did not build these dependencies. When a [recent PR](https://github.com/Azure/communication-ui-library/pull/2076) introduced an import that needed to import from `@internal/acs-ui-common`, the tests started failing at runtime because of missing dependencies.

FixIt.

### DIFF
--- a/samples/tests/package.json
+++ b/samples/tests/package.json
@@ -19,7 +19,6 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@azure/communication-react": "1.3.1-beta.1",
     "@azure/communication-calling": "1.6.1-beta.1 || 1.4.4",
     "@azure/communication-chat": "1.2.0",
     "@azure/communication-common": "2.0.0",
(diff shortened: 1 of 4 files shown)


samples/tests/package.json
```diff
@@ -19,7 +19,6 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@azure/communication-react": "1.3.1-beta.1",
```
prprabhu-ms, line 22: These tests currently don't use this dependency at all, so dropping it (instead of adding to CI time by building all of `@azure/communication-react` before running these tests.

Let's see what CI has to say about this ;)

prprabhu-ms:
Added `ui change` label to add the update snapshot action.

carocao-msft (approved):

JamesBurnside:
(Disabled auto merge until CI checks are re-enabled)

JamesBurnside:
Renabled auto merge

JamesBurnside (approved):

PR #2142, opened 2022-08-02
1. [VideoTile Participant States] - Add a new prop VideoTile.participantState

anjulgarg:
# What
As part of PSTN and 1:N features, we need to show a remote participants call connection state.
Adding a new prop to video tile for making this possible.

image

Localization coming soon.
Tracked via https://skype.visualstudio.com/SPOOL/_workitems/edit/2940251

# Why

# How Tested

# Process & policy checklist


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2758,8 +2758,13 @@ export interface VideoGalleryProps {
 export interface VideoGalleryRemoteParticipant extends VideoGalleryParticipant {
     isSpeaking?: boolean;
     screenShareStream?: VideoGalleryStream;
+    // @beta
+    state?: VideoGalleryRemoteParticipantState;
 }
 
+// @beta
+export type VideoGalleryRemoteParticipantState = 'Connecting' | 'Ringing' | 'Hold';
```
JamesBurnside, line 2766: Should something like 'connected' or 'inCall' be a state?

packages/react-components/review/beta/react-components.api.md
```diff
@@ -1395,8 +1395,13 @@ export interface VideoGalleryProps {
 export interface VideoGalleryRemoteParticipant extends VideoGalleryParticipant {
     isSpeaking?: boolean;
     screenShareStream?: VideoGalleryStream;
+    // @beta
+    state?: VideoGalleryRemoteParticipantState;
 }
 
+// @beta
+export type VideoGalleryRemoteParticipantState = 'Connecting' | 'Ringing' | 'Hold';
```
JamesBurnside, line 1403: does calling SDK use `hold` or `onHold` when it referes to something being on hold?

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -100,27 +106,40 @@ export interface VideoTileProps {
...
+      <Stack styles={defaultPersonaStyles}>
+        <Persona
+          coinSize={coinSize}
+          hidePersonaDetails={hidePersonaDetails}
+          text={text ?? ''}
+          initialsTextColor="white"
+          aria-label={noVideoAvailableAriaLabel ?? ''}
+          showOverflowTooltip={false}
+        />
+        {participantState && <Text className={mergeStyles(participantStateStyle)}>{participantState}</Text>}
```
JamesBurnside, line 141: This text will need to be localized

packages/communication-react/review/beta/communication-react.api.md
anjulgarg: Didn't include those states because we don't show the state in our video tile visually when the participant is incall or connected.

packages/react-components/src/components/VideoTile.tsx
anjulgarg: It's part of a follow up PR
Tracked by this item https://skype.visualstudio.com/SPOOL/_workitems/edit/2940251

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: My thinking is that it does impact the UI - by explicitly not showing a state in the video tile. Contoso should be always able to set all props and an enum/type like `VideoGalleryRemoteParticipantState` should show all the possibilities - if I tried to use that prop its not clear that I can set it as undefined and get a different behavior than if I set it to one of the string litreral values.

anjulgarg: The concern here is by adding `Connected` to the enum, we are suggesting that similar to `Ringing...` and `Connecting...` text under the Persona, `Connected` will also be shown. Which is misleading.

carocao-msft (approved):
Looks good to me aside from James's comment, should we discuss this in parking lot? I agree with Anjul on the misleading part however are we limiting options for Contoso here?

packages/communication-react/review/beta/communication-react.api.md
prprabhu-ms: I agree that `Connected` should be included.
We would expect the state to be `Connected` for participants actually in the call (once our selectors have been updated) so we should leave a note that `undefined` state is treated the same as `Connected` for backward compatibility.

prprabhu-ms (approved):

packages/communication-react/review/beta/communication-react.api.md
prprabhu-ms: I don't think the value of `Connected` is misleading. The field isn't called `connectionOverlayLabel` or some such, in which case it would be misleading to show nothing for some explicit value.

An alternative implementation of the `VideoTile` might not even attempt to show the video until state enters `Connected`...

PR #2143, opened 2022-08-02
Return participant state from videoGallerySelector

anjulgarg:
# What
Return participant state from videoGallerySelector and wire it in video gallery -> remote video tiles

# Why
To show remote participant states such as ringing, connecting in videogallery during pstn / 1:N calls

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
+++ b/packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
@@ -1,10 +1,17 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { DominantSpeakersInfo } from '@azure/communication-calling';
+import {
+  DominantSpeakersInfo,
+  RemoteParticipantState as RemoteParticipantConnectionState
+} from '@azure/communication-calling';
 import { memoizeFnAll, toFlatCommunicationIdentifier } from '@internal/acs-ui-common';
 import { RemoteParticipantState, RemoteVideoStreamState } from '@internal/calling-stateful-client';
-import { VideoGalleryRemoteParticipant, VideoGalleryStream } from '@internal/react-components';
+import {
+  VideoGalleryRemoteParticipant,
+  VideoGalleryRemoteParticipantState,
+  VideoGalleryStream
+} from '@internal/react-components';
 import { checkIsSpeaking } from './SelectorUtils';
 
 /** @internal */
@@ -30,6 +37,7 @@ export const _videoGalleryRemoteParticipantsMemo = (
           participant.isMuted,
           checkIsSpeaking(participant),
           participant.videoStreams,
+          participant.state,
           participant.displayName
         );
       });
@@ -42,6 +50,7 @@ const memoizedAllConvertRemoteParticipant = memoizeFnAll(
     isMuted: boolean,
     isSpeaking: boolean,
     videoStreams: { [key: number]: RemoteVideoStreamState },
+    state: RemoteParticipantConnectionState,
     displayName?: string
   ): VideoGalleryRemoteParticipant => {
     return convertRemoteParticipantToVideoGalleryRemoteParticipant(
@@ -49,6 +58,7 @@ const memoizedAllConvertRemoteParticipant = memoizeFnAll(
       isMuted,
       isSpeaking,
       videoStreams,
+      state,
       displayName
     );
   }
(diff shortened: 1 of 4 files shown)


packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -82,14 +93,24 @@ export const convertRemoteParticipantToVideoGalleryRemoteParticipant = (
     }
   }
 
+  /**
+   * Since VideoTile only cares about the following states,
+   * we don't observe the other states like 'EarlyMedia'.
+   */
+  let participantState: VideoGalleryRemoteParticipantState | undefined;
+  if (state === 'Connecting' || state === 'Ringing' || state === 'Hold') {
```
JamesBurnside, line 101: Do we filter out participants with other states anywhere? Or do we show a participant with state "Disconnected" or "Idle" in the video gallery currently?

JamesBurnside: Hmm we just filter out `inLobby`:
image

(not part of this PR) Should we be filtering out every state we don't explicitly handle?

JamesBurnside, line 101: Per comment on other PR, I think we should support 'Connected' as a valid state and pass that through

JamesBurnside (approved):

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
anjulgarg: Any particular reason for supporting the Connected state?
There is no visual feedback to the customers when the state is connected

anjulgarg: I think you are right.
Pinged the calling sdk team to understand what the states like `Idle` and `EarlyMedia` mean.
Will modify the filter logic once I have answers.

anjulgarg: Updated the code, we didn't need to filter out states but convert them to video tile states.

carocao-msft (approved):

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -103,3 +115,18 @@ const convertRemoteVideoStreamToVideoGalleryStream = (stream: RemoteVideoStreamS
...
   };
 };
+
+const convertRemoteParticipantStateToVideoGalleryRemoteParticipantState = (
+  state: RemoteParticipantConnectionState
+): VideoGalleryRemoteParticipantState | undefined => {
+  if (state === 'Idle' || state === 'Connecting') {
+    return 'Connecting';
+  }
+  if (state === 'EarlyMedia' || state === 'Ringing') {
```
prprabhu-ms, line 125: What is `EarlyMedia`, and how does it differ from `Ringing`?

Should the logic for how to deal with `EarlyMedia` be in the UI component instead of here?
Justification: We avoid diff from headless base types unless we have a reason for it (in that case add a comment to the code above about why we are mapping these states together)

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -103,3 +115,18 @@ const convertRemoteVideoStreamToVideoGalleryStream = (stream: RemoteVideoStreamS
     renderElement: stream.view?.target
   };
 };
+
+const convertRemoteParticipantStateToVideoGalleryRemoteParticipantState = (
+  state: RemoteParticipantConnectionState
+): VideoGalleryRemoteParticipantState | undefined => {
+  if (state === 'Idle' || state === 'Connecting') {
```
prprabhu-ms, line 122: Same as below: Add a comment explaining why we need to map these two together.

change/@internal-calling-component-bindings-d97d9a4b-3ddf-40e4-85a8-a59cfa2c246a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "videoGallerySelector returned remote participants contain a state attribute reflecting their call connection state",
```
prprabhu-ms, line 3: nit: weird grammar

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -42,13 +50,15 @@ const memoizedAllConvertRemoteParticipant = memoizeFnAll(
     isMuted: boolean,
     isSpeaking: boolean,
     videoStreams: { [key: number]: RemoteVideoStreamState },
+    state: RemoteParticipantConnectionState,
```
prprabhu-ms, line 53: Let's call this `connectionState` everywhere. `state` is too generic, and already has at least two distinct meanings in our package.

prprabhu-ms: Ugh. I see that this is called `RemoteParticipantState` and the field is called `state` in the headless API.

So... let's stick to `state` to be consistent with that (bad) naming.

PR #2145, opened 2022-08-03
Separate out output directories for multiple playwright runs

prprabhu-ms:
Starting with https://github.com/Azure/communication-ui-library/pull/2072, browser tests involve multiple invocations of playwright. Before this PR, the multiple invocations end up overwriting the output directory from previous invocations.

* Use a new output directory for each invocation. This PR simply adds a sub-directory with the current Unix timestamp for each invocation.
* Update workflow to upload all playwright test stats from all the invocations.
* Add a new flag `runBrowserTests.mjs -x` to skip further invocations on failure (builds atop `playwright -x`)

## Fixes

* Missing snapshots from some test failures in CI.
* Missing per-test metrics for analysis.

Does *not* affect the results of CI runs.

## Validation

https://github.com/prprabhu-ms/acr-e2e-analysis/commit/dd0b0e0b0e3265dab170e8eef0051c021deef3b8 updates the analysis pipeline and verifies that the data from CI runs for this PR is correctly processed.

### DIFF
--- a/packages/react-composites/playwright.config.common.ts
+++ b/packages/react-composites/playwright.config.common.ts
@@ -10,7 +10,11 @@ const DESKTOP_VIEWPORT = {
 };
 
 const TEST_ROOT = './tests/browser';
-const OUTPUT_DIR = './test-results';
+
+const outputDir = process.env.PLAYWRIGHT_OUTPUT_DIR;
+if (!outputDir) {
+  throw new Error('Environment variable PLAYWRIGHT_OUTPUT_DIR not set');
+}
 
 const buildFlavor: 'beta' | 'stable' = process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' ? 'stable' : 'beta';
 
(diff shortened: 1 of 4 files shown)


prprabhu-ms:
https://github.com/prprabhu-ms/acr-e2e-analysis/commit/dd0b0e0b0e3265dab170e8eef0051c021deef3b8 updates the analysis pipeline and verifies that the data from CI runs for this PR is correctly processed.

packages/react-composites/scripts/runBrowserTests.mjs
```diff
@@ -61,14 +63,34 @@ async function runStress(args) {
 }
 
 async function runAll(args) {
+  let success = true;
   for (const composite of args.composites) {
-    await runOne(args, composite, 'hermetic');
+    try {
```
alkwa-msft, line 68: I realize this was in before but we are saying we always run all of the hermetic tests and then if we set !hermeticOnly we run the live tests?

is there no chance we would ever just want to run live and not hermetic?

alkwa-msft (approved):

packages/react-composites/playwright.config.common.ts
```diff
@@ -10,7 +10,11 @@ const DESKTOP_VIEWPORT = {
 };
 
 const TEST_ROOT = './tests/browser';
-const OUTPUT_DIR = './test-results';
+
+const outputDir = process.env.PLAYWRIGHT_OUTPUT_DIR;
+if (!outputDir) {
```
JamesBurnside, line 15: Can this be a default if I don't want to set this for local runs?

JamesBurnside (approved):
Just a comment on not throwing errors on local runs if output dir is not set

packages/react-composites/playwright.config.common.ts
prprabhu-ms: Nope. All runs should go via `runBrowserTests.mjs` (there is so much specific configuration in our repo....)

packages/react-composites/scripts/runBrowserTests.mjs
prprabhu-ms: At least right now, I didn't see a need for this. We can definitely tweak that in the future if we want.

prprabhu-ms: Note that you can always run a specific test / test file etc by passing arguments to playwright (after `--`)

PR #2149, opened 2022-08-03
3. [VideoTile Participant States] - Localization for video tile participant state prop

anjulgarg:
# What
Localization for video tile participant state prop

# Why

# How Tested

# Process & policy checklist


carocao-msft (approved):

JamesBurnside (approved):

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -26,6 +27,17 @@ import {
 } from './styles/VideoTile.styles';
 import { getVideoTileOverrideColor } from './utils/videoTileStylesUtils';
 
+/* @conditional-compile-remove(PSTN-calls) */
```
dmceachernmsft, line 30: should this also have the CC tag for `one-to-n-calling`?

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -138,7 +162,7 @@ const DefaultPlaceholder = (props: DefaultPlaceholderProps): JSX.Element => {
           aria-label={noVideoAvailableAriaLabel ?? ''}
           showOverflowTooltip={false}
         />
-        {participantState && <Text className={mergeStyles(participantStateStyle)}>{participantState}</Text>}
+        {participantState && <Text className={mergeStyles(participantStateStyle)}>{participantStateString}</Text>}
```
dmceachernmsft, line 165: Do we also maybe want to embed a Announcer using these strings here as well? or does the narrator catch these string changes when its just in a `Text` tag?

anjulgarg: Not sure. Lets postpone a11y work as complex as announcer till public preview.

dmceachernmsft (approved):

PR #2153, opened 2022-08-04
Fix a11y of FileSharing download button

mgamis-msft:
# What

- The onFocus callback to focus on the message content in order to skip the author when tab navigating on iOS VoiceOver was removed. It prevented tab navigating
- the aria label text that is read aloud by iOS VoiceOver for the local user's message was changed to 'You said {message}' instead of '{author} said message'

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2864669

# How Tested

Tab navigating on Windows 10 on Chrome:
https://user-images.githubusercontent.com/79475487/182792712-ae2f92d3-3170-482c-812e-44d0242b9b34.mp4

Tab navigating on IPad on Safari with VoiceOver (note it is reading out the author of remote messages because fluentui n* is focusing on the author first even if the author is not shown for subsequent messages):
https://user-images.githubusercontent.com/79475487/182792514-93ff39c1-ce64-4cc7-821e-7faf6f17faeb.mp4

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -179,27 +179,30 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
...
-        }}
-      >
+      <div ref={messageRef} tabIndex={0}>
         <Chat.Message
           data-ui-id="chat-composite-message"
           className={mergeStyles(messageContainerStyle as IStyle)}
           styles={messageContainerStyle}
           content={
-            <div ref={messageContentRef} tabIndex={0}>
+            <div tabIndex={0}>
```
anjulgarg, line 201: Is this working?
During my testing I noticed that if we add a tabIndex here, it gets overriden and set to -1

mgamis-msft: Yes. That is fluentUI n* Chat behaviour. Arrow keys need to be uses to navigate from message to message.

alkwa-msft: I am having a hard time understanding do we need to do anything here and have a chat with fluentUI n* or are things working as expected?

mgamis-msft: Things are working as expected. tabIndex=-1 means that message to message navigation in the n* chat cannot be done by pressing Tab. Arrow key navigation to navigate through messages. This can be verified from the n* Chat documentation sample here https://fluentsite.z22.web.core.windows.net/0.63.0/components/chat/definition.

anjulgarg (approved):

prabhjot-msft (approved):

PR #2155, opened 2022-08-04
Speculative fix for a common source of flakiness

prprabhu-ms:
# What

In the browser tests, bump the timeout for 

* Enumerating devices
* Creating `CallAgent` and connecting to call.
* Create remote video streams

These steps often take more than the current timeout of 5 seconds.
The cost of this flaky test failure is high: Sometimes the entire test, and even the whole CI suite has to be rerun.

Trade some potential CI individual job runtime for overall CI stability.

Once this merges, I will observe the top-level metrics for job runtimes and flakiness to see if there is any movement.

## Expectations

Based on multiple stress tests:
| Attempt | # of passed suites | remote video timeout | # of times remote video timed out |
|--|--|--|--|
| [1](https://github.com/Azure/communication-ui-library/actions/runs/2796180601) | 3/7 | 5s | 51 |
| [2](https://github.com/Azure/communication-ui-library/actions/runs/2799007899) |  5/7 | 10s | 60 |
| [3](https://github.com/Azure/communication-ui-library/actions/runs/2801324256) |  6/7 | 20s | 53 |
| [4](https://github.com/Azure/communication-ui-library/actions/runs/2802919070) | TBD | 50s | TBD |

I expect this PR to help reduce flakiness, but not quite eliminate it.
We are observing e2e latency for enabling video >60 seconds in our test setup, so the tests might still fail in this way, but bumping the timeout will make the failure less likely.

On the balance, this PR seems worthwhile as a stop-gap while we slowly migrate tests away towards hermetic tests.


packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -114,7 +115,13 @@ export const waitForCallCompositeToLoad = async (page: Page): Promise<void> => {
   await page.bringToFront();
   await page.waitForLoadState('load');
   await waitForPageFontsLoaded(page);
-  const startCallButton = await waitForSelector(page, dataUiId('call-composite-start-call-button'));
+
+  // Tests often timeout at this point because device enumeration can take > 5 seconds.
```
JamesBurnside, line 119: I'm OK bumping the total overall timeout we wait for all selectors to 10 seconds, the assumption being that most are immediate so only failed tests will trigger the 10s timeout

prprabhu-ms: We should start with targeted bumps.

Besides being a more pointed hammer, it gives us obvious targets to bring the timeouts back down. A blanket bump means we have no idea where the timeouts are adding more latency.

In my experience, timeouts and retry config should be as specific as possible.

dmceachernmsft (approved):

JamesBurnside (approved):

prprabhu-ms:
Disabling automerge.

Triggered a stress test: https://github.com/Azure/communication-ui-library/actions/runs/2799007899 to know if I got all the bad 'uns.

prprabhu-ms:
image

10 seconds probably ain't good enough :(

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -193,7 +193,9 @@ export const loadCallPageWithParticipantVideos = async (pages: Page[]): Promise<
       // The tests often timeout at this step because loading remote video streams can take > 5 seconds.
       // Extend the timeout here to trade flakiness for potentially longer test runtime.
       // Test flake has a much larger impact on overall CI time than individual test runtime.
-      { timeout: 2 * perStepLocalTimeout() }
+      //
+      // The extended timeout was determined by stress testing on CI.
```
prprabhu-ms, line 197: https://github.com/Azure/communication-ui-library/runs/7679141230

This timeout was the only one that was hit. It was hit *60 times* in 7 runs of the entire suite :(

prprabhu-ms:
https://github.com/Azure/communication-ui-library/runs/7685332349

The 20 second timeout was exceeded 53 times in 7 runs of the entire test suite.

This is not encouraging. I definitely don't want to bump up timeouts if I don't see a reduction in flakiness with it.

A hail mary: 1 minute timeout to see what happens.

PR #2159, opened 2022-08-04
Poll to see if the npm package published successfully

JamesBurnside:
# What
* Polling script that checks if the npm package is on the npm registry
  * Checks every 5 seconds for 5 minutes (typically takes 1-1.5mins for the package to publish)
* Added script to run in the nightlycd to see if it works

# Why
We currently get no feedback that the package successfully published after we trigger the ado pipeline

# How Tested
Just the script locally:
image


common/scripts/poll-npm-package-published.mjs
```diff
@@ -0,0 +1,42 @@
+import { get } from "https";
+import process from "process";
+
+const POLL_INTERVAL_MS = 5 * 1000; // 5 seconds in millis
+const TOTAL_TIME_TO_POLL_MS = 5 * 60 * 1000; // 5 minutes in millis
+const MAX_RETRIES = TOTAL_TIME_TO_POLL_MS / POLL_INTERVAL_MS;
+
+const makeRequest = (url) => {
+  return new Promise((resolve, reject) => {
+    get(url, (response) => {
```
prprabhu-ms, line 10: Should implement a timeout here: https://stackoverflow.com/a/42357125/16098202

common/scripts/poll-npm-package-published.mjs
```diff
@@ -0,0 +1,42 @@
...
+    }).end();
+  });
+}
+
+const sleep = (ms) => new Promise( res => setTimeout(res, ms));
+
+const main = async () => {
+  const url = process.argv[2];
+  let retries = 0;
+  while(true) {
```
prprabhu-ms, line 21: Timeout should be enforced directly instead of by computing number of retries. Something like

```typescript

startTime = new Date();
while (new Date() - startTime < TIMEOUT_MS) {
}
```

Why? Because otherwise you don't account for non-trivial time spent each time in the loop. The `await makeRequest(url)` call below could take 30 seconds. At that point, you're really going to poll for 35 minutes instead of 5 minutes.

common/scripts/poll-npm-package-published.mjs
```diff
@@ -0,0 +1,42 @@
...
+    get(url, (response) => {
+      resolve(response);
+    }).end();
+  });
+}
+
+const sleep = (ms) => new Promise( res => setTimeout(res, ms));
+
+const main = async () => {
+  const url = process.argv[2];
```
prprabhu-ms, line 19: Huh, I expected the arguments to be `packageName` and `packageVersion`.

The script says it polls NPM, but the arguments allow polling anything ;)

.github/workflows/nightly-cd.yml
```diff
@@ -145,3 +145,6 @@ jobs:
           azure-pipeline-name: 'azuresdkpartnerdrops to npm'
           azure-devops-token: '${{ secrets.AZURE_SDK_RELEASE_PIPELINE_DEVOPS_TOKEN }}'
           azure-pipeline-variables: '{"accessLevel": "public", "BlobPath": "azure-communication-services/react/npm/${{ steps.version.outputs.version }}", "registry": "https://registry.npmjs.org/", "skipDiff": "False", "tag": "dev"}'
+
+      - name: Check if npm package published successfully
+        run: node common/scripts/poll-npm-package-published.mjs
```
JamesBurnside, line 150: oop  need to pass the url with the version here

prprabhu-ms, line 150: Don't you need to pass in the URL here?

Or better still, pass in the package name and version to poll for.

JamesBurnside: updated

JamesBurnside: yeah, updated lol

common/scripts/poll-npm-package-published.mjs
JamesBurnside: Added

common/scripts/poll-npm-package-published.mjs
JamesBurnside: Great point! Updated to use this!

common/scripts/poll-npm-package-published.mjs
JamesBurnside: I never like faffing with argument names haha, but I have updated it to assume `@azure/communication-react` and to take in a version as the arg

prprabhu-ms: Unfortunately, we don't have `yargs` as a dependency here.

Look at the beauty of a fully fledged CLI args parser: https://github.com/Azure/communication-ui-library/blob/a1cc29e9a751b4a1a5f4b9039abd96b447363ff9/packages/react-composites/scripts/runBrowserTests.mjs#L139

Taking in just the version is fine. That's what I was going to do as well.

common/scripts/poll-npm-package-published.mjs
```diff
@@ -0,0 +1,57 @@
...
+const GET_REQUEST_TIMEOUT_MS = 5000;
+
+const GET_REQUEST_OPTIONS = {
+  hostname: 'registry.npmjs.org',
+  timeout: GET_REQUEST_TIMEOUT_MS
+}
+
+const checkNpm = (packageName, packageVersion) => {
+  return new Promise((resolve, reject) => {
+    const request = get({...GET_REQUEST_OPTIONS, path: `/${packageName}/${packageVersion}`}, (response) => {
```
prprabhu-ms, line 24: Do you need to URL encode the package name and/or version?

common/scripts/poll-npm-package-published.mjs
```diff
@@ -0,0 +1,57 @@
...
+      resolve(response);
+    });
+    console.log(`Pinging: https://${GET_REQUEST_OPTIONS.hostname}/${packageName}/${packageVersion}`);
+    request.end();
+  });
+}
+
+const sleep = (ms) => new Promise( res => setTimeout(res, ms));
```
prprabhu-ms, line 32: nifty. I've written this in more verbose forms before :)

common/scripts/poll-npm-package-published.mjs
```diff
@@ -0,0 +1,57 @@
...
+  while(true) {
+    const response = await checkNpm(packageName, packageVersion);
+    console.log('responseCode: ', response.statusCode);
+
+    if (response.statusCode === 200) {
+      console.log("Successfully found npm package")
+      process.exitCode = 0;
+      return;
+    } else if ((new Date() - startTime) > TIMEOUT_TIME_MS) {
+      console.log('Failed to find package on the npm registry');
```
prprabhu-ms, line 47: I'd just `throw` here.

common/scripts/poll-npm-package-published.mjs
```diff
@@ -0,0 +1,57 @@
...
+      process.exitCode = 1;
+      return;
+    } else {
+      console.log('sleeping for 5 seconds...');
```
prprabhu-ms, line 51: nit: I'd just say `Sleeping for a bit...` so that you don't have to update this message if you update `POLL_INTERVAL_MS` (or more likely, we'll forget to update this comment and then it will be wrong logging (which is worse than no logging at all)

prprabhu-ms (approved):

common/scripts/poll-npm-package-published.mjs
JamesBurnside: Seemingly not (works on the command line), but will know once the nightly cd runs with this

dmceachernmsft (approved):

PR #2160, opened 2022-08-04
PSTN storybook updates

dmceachernmsft:
# What

Update Storybook docs to reflect new props needed for making PSTN calls.
# Why

Informs contoso on what new props are needed to make a PSTN call

https://skype.visualstudio.com/SPOOL/_workitems/edit/2924137
# How Tested

ran story book locally.


packages/storybook/stories/CallWithChatComposite/CallWithChatCompositeDocs.tsx
```diff
@@ -108,6 +109,17 @@ export const getDocs: () => JSX.Element = () => {
...
       </Description>
       <Source code={addFileSharingSnippet} />
+
+      <Heading>PSTN and 1:N Calling</Heading>
+      <SingleLineBetaBanner />
+      <MessageBar>Note: see CallComposite for detailed implementation description</MessageBar>
+      <Description>
+        The CallWithChatComposite supports the PSTN Calling modality as well. With CallWithChat you will need to still
+        provide a chat thread Id for instances where you will add a Azure Communications user to the call. If you are
+        looking to use the CallWithChatComposite and are looking to hide the chat because you are only calling PSTN
```
JamesBurnside, line 119: I would leave this out with the assumption that if someone is doing a call only they should be using the call composite

JamesBurnside: Oh actually I guess it could be dynamically hidden if suddenly an ACS user is there or not

packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
```diff
@@ -123,6 +124,22 @@ export const getDocs: () => JSX.Element = () => {
...
+      <Description>
+        The CallComposite supports making outbound PSTN and 1:N calls. 1:N is a mix between Azure Communication Users
+        and PSTN users. To make these out bound calls you need to provide a `locator` that contains participantID's that
+        you are looking to call to the [CallAdapter](./?path=/docs/composite-adapters--page). For PSTN these ID's are
+        the phone numbers that you are looking to call. For Azure Communication Users you will need to provide their
+        unique ACS aquired `userId`.
+      </Description>
+      <Description>
+        As well as these participantID's you are required to provide a [phone
+        number](https://docs.microsoft.com/en-us/azure/communication-services/quickstarts/telephony/get-phone-number?tabs=windows&pivots=platform-azcli)
```
JamesBurnside, line 138: is a phone number needed in a 1:N call where all participants are an ACS participant?

packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
```diff
@@ -123,6 +124,22 @@ export const getDocs: () => JSX.Element = () => {
         useful if you want to explore the composite with multiple users.
       </Description>
 
+      <Heading>PSTN and 1:N Calling</Heading>
+      <SingleLineBetaBanner />
+      <Description>
+        The CallComposite supports making outbound PSTN and 1:N calls. 1:N is a mix between Azure Communication Users
+        and PSTN users. To make these out bound calls you need to provide a `locator` that contains participantID's that
```
JamesBurnside, line 131: nit: `out bound` -> `outbound`

JamesBurnside, line 131: nit: `participantID's` -> `participantIDs`

JamesBurnside: (and line below `ID's` -> `IDs`)

JamesBurnside (approved):

packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
dmceachernmsft: no its only needed for PSTN. I wonder though if we should have a note about how the calling modalities can mix so its encouraged so they can add a PSTN user to a 1:N call?

JamesBurnside: Good point - phone numbers just add a whole nother level of complexity to get set up hoewever.
Regarding this doc line, can it be updated to something like `If you are looking to make a have PSTN attendees in the 1:N call then alongside the participantIDs you must also provide a [phone number ....`

anjulgarg (approved):

PR #2130, opened 2022-07-26
Refactored File sharing code to improve the use of conditional compilation

prabhjot-msft:
# What
Refactored File sharing code to improve the use of conditional compilation

# Why

# How Tested

# Process & policy checklist

- [ ] I have updated the project documentation to reflect my changes if necessary.
- [ ] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.

**Is this a breaking change?**

- [ ] This change causes current functionality to break.

PR #2161, opened 2022-08-04
Participant Item to show Participant States

anjulgarg:
# What
Uses consistent types b/w VideoTile and ParticipantItem participantState
Displays Participants state in participant item

FIGMA - https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?node-id=6180%3A48173

# Why

# How Tested

# Process & policy checklist


dmceachernmsft (approved):

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -185,14 +211,47 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
...
-    [itemHovered, menuTitle, ids.participantItemMenuButton]
+    [itemHovered, strings.menuTitle, ids.participantItemMenuButton]
   );
 
   const onDismissMenu = (): void => {
     setItemHovered(false);
     setMenuHidden(true);
   };
 
+  /* @conditional-compile-remove(one-to-n-calling) */
```
prprabhu-ms, line 222: Conditional compilation note: Please reduce the amount of conditional compilation in this file by following the pattern from https://github.com/Azure/communication-ui-library/pull/2130

* Add a function to encpasulate string related conditional code

```
function partcipantStateStringTrampoline(state: ParticipantState, strings: ParticipantItemStrings): string { 
  if (participantState === 'Idle' || participantState === 'Connecting') {
    return strings?.participantStateConnecting;
  } else if (participantState === 'EarlyMedia' || participantState === 'Ringing') {
    return strings?.participantStateRinging;
  } else if (participantState === 'Hold') {
    return strings?.participantStateHold;
  }
   // Always return the default for stable builds.
  return '';
};
```

* Add a function to encapsulate the difference in API for state:
```

// This assumes that `ParticipantState` is defined unconditionally, but exported only for beta builds.

function participantStateTrampoline(props: ParticipantItemProps): ParticipantState {
  return props.state;
  // On stable builds, where this prop does not exist, the default behavior matches that of 'Connecting' state.
  return 'Connected';
}
```

Then no more conditional compilation anywhere.

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -185,14 +211,47 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
...
+  /* @conditional-compile-remove(PSTN-calls) */
+  const participantStateString = React.useMemo((): string | undefined => {
```
prprabhu-ms, line 224: This is a simple `switch` statement that returns a `string`, so I wouldn't `useMemo` here.

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -185,14 +211,47 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
...
+    strings?.participantStateHold,
+    strings?.participantStateRinging
+  ]);
+
+  const menuItemsWrapperStyle = useMemo(() => {
+    /* @conditional-compile-remove(one-to-n-calling) */
+    /* @conditional-compile-remove(PSTN-calls) */
+    return {
+      /* If participant state is being displayed, hide the menu icon until it is hovered. */
+      display: itemHovered ? 'block' : participantState ? 'none' : 'block'
```
prprabhu-ms, line 245: I think it's clearer to state this as `display: (!participantState || itemHovered) ? 'block' : 'none'`.

The current conditional is hard to read because it's `'block'` is set in the true and then false nested branch.

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -62,6 +65,18 @@ export interface ParticipantItemStrings {
   mutedIconLabel: string;
   /** placeholder text for participants who does not have a display name*/
   displayNamePlaceholder?: string;
+  /* @conditional-compile-remove(one-to-n-calling) */
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** String shown when `participantState` is `Connecting` */
+  participantStateConnecting: string;
```
JamesBurnside, line 71: to prevent breaking change these need to be optional

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -216,28 +275,41 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
...
-            }}
-          />
-        </>
-      )}
+      {/* Show the participant state for a remote participant only. Hide it on hover. */}
+      {
+        /* @conditional-compile-remove(one-to-n-calling) */
+        /* @conditional-compile-remove(PSTN-calls) */
+        !me && participantStateString && (
+          <div onMouseEnter={() => setItemHovered(true)} style={{ display: itemHovered ? 'none' : 'block' }}>
```
JamesBurnside, line 288: should also be hidden when focused on by tab naviation? Or when tapped on mobile?

JamesBurnside: Actually if it's hidden on focus, that could be a a11y issue

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2450,6 +2454,9 @@ export type ParticipantsRemovedListener = (event: {
     removedBy: ChatParticipant;
 }) => void;
 
+// @public
+export type ParticipantState = 'Idle' | 'Connecting' | 'Ringing' | 'Connected' | 'Hold' | 'InLobby' | 'EarlyMedia' | 'Disconnected';
```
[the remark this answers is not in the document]
anjulgarg: We don't import headless sdk types into react composites package.
We have defined new types there matching the headless types to avoid this.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -2146,6 +2146,9 @@ export type ParticipantsRemovedListener = (event: {
     removedBy: ChatParticipant;
 }) => void;
 
+// @public
+export type ParticipantState = 'Idle' | 'Connecting' | 'Ringing' | 'Connected' | 'Hold' | 'InLobby' | 'EarlyMedia' | 'Disconnected';
```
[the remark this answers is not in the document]
anjulgarg: We don't import headless sdk types into react composites package.
We have defined new types there matching the headless types to avoid this.

packages/react-components/src/components/ParticipantItem.tsx
anjulgarg: Because, this is used inside another useMemo, we need to memoize this function too.

anjulgarg: Converted it to shorthand if else statement.

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -102,7 +99,7 @@ export const convertRemoteParticipantToVideoGalleryRemoteParticipant = (
     isScreenSharingOn: screenShareStream !== undefined && screenShareStream.isAvailable,
     /* @conditional-compile-remove(one-to-n-calling) */
     /* @conditional-compile-remove(PSTN-calls) */
-    state: convertRemoteParticipantStateToVideoGalleryRemoteParticipantState(state)
+    state
```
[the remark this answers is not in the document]
anjulgarg: We haven't.
So we are good here :)

packages/react-components/src/components/ParticipantItem.tsx
prprabhu-ms: Spelling out the issues here:

* For keyboard nav users, tabbing onto the "Connecting..." state will hide the string and so the screen reader will not be able to announce the state. A solution is to update the aria string for the participant name to include the connection state (so they tab onto the name and *that* announces the state)

* Verify behavior on mobile. I think this is not an issue because on mobile, there is no overflow button. User needs to tap on the entire menuitem to open the sub-drawer:

image
vs
image

prprabhu-ms:
LGTM except for @JamesBurnside point about a11y on desktop for keyboard users.

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -62,6 +65,18 @@ export interface ParticipantItemStrings {
...
   /** placeholder text for participants who does not have a display name*/
   displayNamePlaceholder?: string;
+  /* @conditional-compile-remove(one-to-n-calling) */
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** String shown when `participantState` is `Connecting` */
+  participantStateConnecting?: string;
+  /* @conditional-compile-remove(one-to-n-calling) */
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** String shown when `participantState` is `Ringing` */
+  participantStateRinging?: string;
```
JamesBurnside, line 75: ideally this component is agnostic of being chat or calling...
Thinking out loud so feel free to shut this down - Instead of passing participantState into here, would it be possible to pass any string that is simply written to the right hand side of the participant item, `participantItemSecondaryText?: string` - then the VideoGalleryParticipant does the mapping of state->string and passes in the appropriate string

anjulgarg: I understand your point of view. And it's a good idea.
However, that removes the localization support we offer for these connection states.
There's another concern of customer injected strings being too long and messing up the UI. They would then also have to deal with truncation.
By adding a fixed set of connection states, we keep this component communication specific.

JamesBurnside: To keep the locale strings, would the localization not live in the videoGallery component locale?

JamesBurnside: I think we need to handle truncation regardless and show an ellipsis or something for long text as we don't know what our translated strings will, or what width Contoso will set the participantItem to be

anjulgarg: Good point.

anjulgarg: But does that mean that now we have to introduce these strings to VideoGallery props as well?

JamesBurnside: They would be introduced there *instead (they should only need to be in one place, but I haven't dug into the code)

packages/react-components/src/components/ParticipantItem.tsx
anjulgarg: Resolving by disabling the menu when a participant is in connecting state

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -553,7 +553,7 @@ export interface CallingTheme {
 
 // @public
 export type CallParticipantListParticipant = ParticipantListParticipant & {
-    state: 'Idle' | 'Connecting' | 'Ringing' | 'Connected' | 'Hold' | 'InLobby' | 'EarlyMedia' | 'Disconnected';
+    state: ParticipantState;
```
JamesBurnside, line 556: (non-blocking comment, and I think this is mentioned in another PR somewhere so might be a duplicate comment)
but I have reservervations around the word `state` here - just because its used in so many contexts throughout our codebase and API, and things like isMuted and isScreenSharing are also the state of a participant. Alternate suggestions: `callConnectionState`, `connectionState` (grr still having to use state as a suffix lol - struggling to think of anything different)

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -205,9 +231,11 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
       onMouseEnter={() => setItemHovered(true)}
       onMouseLeave={() => setItemHovered(false)}
       onClick={() => {
-        setItemHovered(true);
-        setMenuHidden(false);
-        onClick?.(props);
+        if (!participantStateString) {
```
JamesBurnside, line 234: this scares me, as a developer its not clear that if I set the state suddenly my menu item button is gone

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -216,28 +244,53 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
...
-            className={contextualMenuStyle}
-            calloutProps={{
-              preventDismissOnEvent
-            }}
-          />
-        </>
+
+      {/* When the participantStateString has a value, we don't show the menu  */}
+      {!me && participantStateString ? (
+        <Text>{participantStateString}</Text>
```
JamesBurnside, line 255: do we have styles to handle truncation of long strings of text here?

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -62,6 +65,18 @@ export interface ParticipantItemStrings {
   mutedIconLabel: string;
   /** placeholder text for participants who does not have a display name*/
   displayNamePlaceholder?: string;
+  /* @conditional-compile-remove(one-to-n-calling) */
+  /* @conditional-compile-remove(PSTN-calls) */
+  /** String shown when `participantState` is `Connecting` */
+  participantStateConnecting?: string;
+  /* @conditional-compile-remove(one-to-n-calling) */
+  /* @conditional-compile-remove(PSTN-calls) */
```
JamesBurnside, line 73: how do double conditional compiled work? Is this an AND or an OR?
My assumption is the implementation will be an AND (both of these features must be stabilized for this feature to be public), but we would want an OR (if either of these features are stabilized we want this feature public)

JamesBurnside (approved):

packages/react-components/src/components/ParticipantItem.tsx
anjulgarg: Not currently.

packages/react-components/src/components/ParticipantItem.tsx
anjulgarg: Noted!
Have this as a todo item.

packages/react-components/src/components/ParticipantItem.tsx
JamesBurnside: We chatted offline and making this non-blocking for cost to implement this change.. but I still strongly feel this should happen, for posterity in case someone does decide to make this change reiterating why:
* Not pollute participantItem with calling-only properties
* Keep the property as simple as possible to ensure ease of use and understanding by Contoso devs (nothing surprising can happen like hiding the menu button)
* Keeps it flexible - e.g. if Contoso wanted "sharing screen" text here or something (which we have an icon for but just an example)

packages/react-components/src/components/ParticipantItem.tsx
anjulgarg: So if either of these feature is stabilized, we remove both the flags

Because the assumption is that if this is stable for one feature, it is stable for another as well

JamesBurnside: We chatted offline, we don't have support for this so whoever stabilizes this feature first needs to ensure they delete both

packages/communication-react/review/beta/communication-react.api.md
anjulgarg: The `state` attribute here is part of our public api. Can't change it now.
I didn't add it.

packages/react-components/src/components/ParticipantItem.tsx
JamesBurnside: Filed a bug for tracking this [Bug 2946816](https://skype.visualstudio.com/SPOOL/_workitems/edit/2946816): [Web] We do not support OR conditional compilation flags

packages/react-components/src/components/ParticipantItem.tsx
anjulgarg: Yes, going to bring this up in front of AlexP tomorrow.
Creating an item for this.
https://skype.visualstudio.com/SPOOL/_workitems/edit/2946830

PR #2162, opened 2022-08-05
browser tests: Rename projects to remove spaces

prprabhu-ms:
I tried and gave up on properly escaping whitespace on bash, cmd, powershell... your favourite POSIX-like shell.

Just rename the projects to match the chioces in `runBrowserTests.mjs`.

**All snapshot updates on this PR are just file rename**


JamesBurnside (approved):

mgamis-msft (approved):

prprabhu-ms:
There is no free lunch.

Fixing this still doesn't allow me to do `node scripts/runBrowserTests.mjs -c call -- -g 'some fancy call test'`.

Must quote spaces correctly if we want to support windows dev env properly.

PR #2164, opened 2022-08-05
Bypass Lobby Screen for outbound PSTN/1:N Calls

anjulgarg:
# What
Bypass Lobby Screen for outbound PSTN/1:N Calls

# Why
When PSTN or ACS users are called, we need to change the composite to show the call screen and bypass the lobby screen.

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -96,9 +96,21 @@ export const getCallCompositePage = (
...
+  call?.remoteParticipants;
   if (call) {
+    // For PSTN/ACS Calls, the composite should display the call screen immediately.
+    // Currently there is no way to determine the type of a call (group, 1:1, PSTN etc)
+    // For PSTN and ACS calls, when you call a single user, the call state differs from when you call multiple users.
+    // For example, When you call a single PSTN user, the call state will be `Connecting | Ringing` until user picks up.
+    // When you call multiple users, the call state will be `Connected` and individual participant states will be `Connecting | Ringing`.
+    if (
+      (call?.state === 'Connecting' || call?.state === 'Ringing') &&
+      Object.keys(call?.remoteParticipants).length > 0
```
alkwa-msft, line 108: I am curious if we know the number of remoteparticipants at this point. Is it a better check to read the locator to see if we are expecting to call a lot of people or just one?

alkwa-msft:
I had a chat with Donald about this. I think instead of saying we want to bypass lobby when it is being shown for the connecting state. 

We should show the lobby screen if we are a teams meeting interop call (known via locator) and then when we are connecting || lobby then show the lobby screen

This now means that we shouldn't see the "lobby" screen for when other types of calls are connecting. However showing a blank screen isn't preferable either.

While we call it the lobby screen I am curious if perhaps we can inject in different strings based on different kinds of calls if we really do need to show something for a first seconds while we are "connecting"?

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -96,9 +96,21 @@ export const getCallCompositePage = (
   // Must check for ongoing call *before* looking at any previous calls.
   // If the composite completes one call and joins another, the previous calls
   // will be populated, but not relevant for determining the page.
+  call?.remoteParticipants;
   if (call) {
+    // For PSTN/ACS Calls, the composite should display the call screen immediately.
+    // Currently there is no way to determine the type of a call (group, 1:1, PSTN etc)
+    // For PSTN and ACS calls, when you call a single user, the call state differs from when you call multiple users.
+    // For example, When you call a single PSTN user, the call state will be `Connecting | Ringing` until user picks up.
```
JamesBurnside, line 104: When is it `connecting` and and when is it `ringing`? Does it go:
`Idle -> Connecting -> Ringing -> (someone picks up) -> Connected`
or
`Idle  -> Ringing -> (someone picks up) -> Connecting -> Connected`

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -96,9 +96,21 @@ export const getCallCompositePage = (
...
+    // When you call multiple users, the call state will be `Connected` and individual participant states will be `Connecting | Ringing`.
+    if (
+      (call?.state === 'Connecting' || call?.state === 'Ringing') &&
+      Object.keys(call?.remoteParticipants).length > 0
+    ) {
+      return 'call';
```
JamesBurnside, line 110: Why is it we don't want to show a new `ringing` page here instead?

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
dmceachernmsft: its the first one, Connecting state is part of when we are joining the call, you can follow it in this flow chart found here: https://docs.microsoft.com/en-us/azure/communication-services/how-tos/cte-calling-sdk/manage-calls
image

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -96,9 +96,21 @@ export const getCallCompositePage = (
   // Must check for ongoing call *before* looking at any previous calls.
   // If the composite completes one call and joins another, the previous calls
   // will be populated, but not relevant for determining the page.
+  call?.remoteParticipants;
```
prprabhu-ms, line 99: This expression is unused.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
prprabhu-ms: Please add a link to this docs page in the docstring for this function, or as a comment. I'm sure I'll need to refer to it the next time I'm trying to modify this function.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -96,9 +96,21 @@ export const getCallCompositePage = (
...
+    // When you call multiple users, the call state will be `Connected` and individual participant states will be `Connecting | Ringing`.
+    if (
+      (call?.state === 'Connecting' || call?.state === 'Ringing') &&
+      Object.keys(call?.remoteParticipants).length > 0
+    ) {
+      return 'call';
+    }
     // `_isInLobbyOrConnecting` needs to be checked first because `_isInCall` also returns true when call is in lobby.
-    if (_isInLobbyOrConnecting(call?.state)) {
+    else if (_isInLobbyOrConnecting(call?.state)) {
```
prprabhu-ms, line 113: This function is has a lot of business logic at this point. Please add unit-tests for all the different cases it handles.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -96,9 +96,21 @@ export const getCallCompositePage = (
   // Must check for ongoing call *before* looking at any previous calls.
   // If the composite completes one call and joins another, the previous calls
   // will be populated, but not relevant for determining the page.
+  call?.remoteParticipants;
   if (call) {
+    // For PSTN/ACS Calls, the composite should display the call screen immediately.
```
prprabhu-ms, line 101: Please add a comment *why* we should skip the call screen. It is not obvious to me.

Is it because we don't want to transition the UX through two transitory states: Connecting screen -> Ringing tile? Or is there another reason?

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
prprabhu-ms: I agree with alex that the intent to skip the connecting screen is already there in the locator used to join the call. So it'd be more direct and robust to key off of the locator.

prprabhu-ms (commented):
All comments inline

PR #2169, opened 2022-08-08
Merge two live tests for `CallComposite`

prprabhu-ms:
This PR merges two live tests for `CallComposite`.

Currently, estimated success rate of each test on `main` is 93% and 85% respectively. The flakiness stems almost entirely from a step in test setup. By merging the tests the PR reduces the overall flakiness of CI builds.

The merged test is one of two `CallComposite` tests that will be left behind as live smoke tests.


packages/react-composites/tests/browser/call/live/CallComposite.test.ts
```diff
@@ -148,15 +141,44 @@ test.describe('Call Composite E2E CallPage Tests', () => {
...
-      `video-gallery-page-camera-toggled.png`
+    // Starting / stopping video streams is known to take up to 15 seconds.
+    // Bump this timeout to trade potential longer runtimes for less flakiness.
+    await waitForFunction(
+      page,
+      () => {
+        return document.querySelectorAll('video').length === 1;
+      },
+      undefined,
+      { timeout: 4 * perStepLocalTimeout() }
```
JamesBurnside, line 173: nit: can you add a comment here why `4 * perStepLocalTimeout()`

mgamis-msft (approved):

prprabhu-ms:
The snapshot updates show that this test is now flip-flopping between the camera being off or not in the second snapshot.

PR #2170, opened 2022-08-08
Remove incomplete but flaky test

prprabhu-ms:
Remove an incomplete test that is currently very flaky. The success rate of this test on `main` is 88%

This test seems incomplete -- it does not take any snapshots, nor assert any conditions. The test will not fail even if clicking on the camera cycle button throws an error...


prprabhu-ms:
Primarily a question for @dmceachernmsft 

We should make should this test is stable before adding it back next time.

dmceachernmsft:
I'm fine with removing this. the thing about this test was there wasn't a good way to mimic the camera changing when I wrote this. is it still the case that we can't mimic this? if we can then its worth having a test to this, otherwise I am okay with removing it until we figure that out @prprabhu-ms

dmceachernmsft (approved):

prprabhu-ms:
> I'm fine with removing this. the thing about this test was there wasn't a good way to mimic the camera changing when I wrote this. is it still the case that we can't mimic this? if we can then its worth having a test to this, otherwise I am okay with removing it until we figure that out @prprabhu-ms

@dmceachernmsft There still isn't an easy way to test this hermetically as our mocks are static.

That said, my main point is that this _live_ test isn't testing anything either (because it takes not snapshots, and iiuc even if the application throws an error in the browser the test will not notice).

So currently this is adding flake for not value.

Let's reconsider adding this back as a live test, but making sure via stress tests that it is stable and actually tests what we want.

anjulgarg (approved):

PR #2171, opened 2022-08-08
Make localization test hermetic

prprabhu-ms:
'cause there's no reason to load the whole composite for these.


JamesBurnside:
Is this meant to be (deletion of the beta ones and new stable ones):
image

prprabhu-ms:
> Update: I see the update action failed, rerunning....
> 
> Is this meant to be (deletion of the beta ones and new stable ones): image

The deleted files are `.../live/...` and the added ones are `.../hermetic/...`

I've been trying to manually delete the old snapshots as I move tests around, but I suspect that some will be left behind to be cleaned up later because there are so many conflicting PRs in flight.

The fact that there are only hermetic snapshots for one build flavor is just because our snapshot update automation is terrible at the moment -- we kick off three of them and one of them wins... forcing the other two to restart. And each of these takes hours...

JamesBurnside (approved):

dmceachernmsft (approved):

PR #2174, opened 2022-08-08
Fix poll script timing and exit code

JamesBurnside:
# What
* Update npm script to wait for 5 minutes instead of 12 seconds (which was just for testing and not meant to be committed to code)
* Added `await main()` to ensure the exit code is propagated and ensure the github action actually fails

# Why
Currently, despite not finding the package, the action still passed:
image
https://github.com/Azure/communication-ui-library/runs/7700649145?check_suite_focus=true

# How Tested
Will be tested in nightly CD

### DIFF
--- a/common/scripts/poll-npm-package-published.mjs
+++ b/common/scripts/poll-npm-package-published.mjs
@@ -44,9 +44,7 @@ const main = async () => {
       process.exitCode = 0;
       return;
     } else if ((new Date() - startTime) > TIMEOUT_TIME_MS) {
-      console.error("Failed to find package on the npm registry");
-      process.exitCode = 1;
-      return;
+      throw new Error('Failed to find package on the npm registry');
     } else {
       console.log('Sleeping for a bit...');
       await sleep(POLL_INTERVAL_MS);
@@ -54,4 +52,4 @@ const main = async () => {
   }
 };
 
-main();
+await main();
--- a/common/scripts/poll-npm-package-published.mjs
+++ b/common/scripts/poll-npm-package-published.mjs
@@ -11,7 +11,7 @@ import process from "process";
  */
 
 const POLL_INTERVAL_MS = 5000; // 5 seconds in millis
-const TIMEOUT_TIME_MS = 12 * 1000; // 5 minutes in millis
+const TIMEOUT_TIME_MS = 5 * 60 * 1000; // 5 minutes in millis
 const GET_REQUEST_TIMEOUT_MS = 5000;
 
 const GET_REQUEST_OPTIONS = {
@@ -44,7 +44,9 @@ const main = async () => {
       process.exitCode = 0;
       return;
     } else if ((new Date() - startTime) > TIMEOUT_TIME_MS) {
-      throw new Error('Failed to find package on the npm registry');
+      console.error("Failed to find package on the npm registry");
+      process.exitCode = 1;
+      return;
     } else {
       console.log('Sleeping for a bit...');
       await sleep(POLL_INTERVAL_MS);


prprabhu-ms (approved):

common/scripts/poll-npm-package-published.mjs
```diff
@@ -44,7 +44,9 @@ const main = async () => {
       process.exitCode = 0;
       return;
     } else if ((new Date() - startTime) > TIMEOUT_TIME_MS) {
-      throw new Error('Failed to find package on the npm registry');
+      console.error("Failed to find package on the npm registry");
```
prprabhu-ms, line 47: I recommend adding a comment that you had to set exit code explicit for the GH action to detect failure.

Though... I'm curious why that is the case.

prprabhu-ms: If you're taking requests, I'd rather launch the script with `--unhandled-rejections=strict` ;)

image

JamesBurnside: We could add a flag, but seems for backwards compat they don't have a non-zero error code when it fails:
image

Perhaps doing
```
process.existCode = 1;
throw new Error(...)
```
would work also?

prprabhu-ms: `--unhandled-rejections=strict` seems to have the intended behavior:

For,
```typescript
async function badUn() {
    throw new Error('do not handle me');
}

badUn();
```

```
@prprabhu-ms ➜ /workspaces/communication-ui-library/packages/react-composites (prprabhu/deflake-config-page ✗) $ node --unhandled-rejections=strict scripts/why-wont-you-die.mjs; echo $?
file:///workspaces/communication-ui-library/packages/react-composites/scripts/why-wont-you-die.mjs:2
    throw new Error('do not handle me');
          ^

Error: do not handle me
    at badUn (file:///workspaces/communication-ui-library/packages/react-composites/scripts/why-wont-you-die.mjs:2:11)
    at file:///workspaces/communication-ui-library/packages/react-composites/scripts/why-wont-you-die.mjs:5:1
    at ModuleJob.run (internal/modules/esm/module_job.js:170:25)
    at async Loader.import (internal/modules/esm/loader.js:178:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
1
```

But more importantly, all this is only needed if you don't `await` the async function. If you modify that toy example to:

```typescript
async function badUn() {
    throw new Error('do not handle me');
}

await badUn();
```

Then you get

```
@prprabhu-ms ➜ /workspaces/communication-ui-library/packages/react-composites (prprabhu/deflake-config-page ✗) $ node scripts/why-wont-you-die.mjs; echo $?
file:///workspaces/communication-ui-library/packages/react-composites/scripts/why-wont-you-die.mjs:2
    throw new Error('do not handle me');
          ^

Error: do not handle me
    at badUn (file:///workspaces/communication-ui-library/packages/react-composites/scripts/why-wont-you-die.mjs:2:11)
    at file:///workspaces/communication-ui-library/packages/react-composites/scripts/why-wont-you-die.mjs:5:7
    at ModuleJob.run (internal/modules/esm/module_job.js:170:25)
    at async Loader.import (internal/modules/esm/loader.js:178:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
1
```

Without any extra flags.

prprabhu-ms: So you simply need to `await main()` on the last line.

JamesBurnside: Updated !

dmceachernmsft (approved):

PR #2175, opened 2022-08-08
browser tests: Use higher bitrate test video

prprabhu-ms:
Thanks to the amazing folks from the headless calling team, we found out that a bunch of the test flakiness with symptoms

- `isReceiving` being false intermittently
- remote video render timing out

is likely because the test video we use to fake the local camera has very low bitrate. Beyond some configurable / implementation depenedent bitrate, the calling infrastructure may determine that the stream is unavailable / poor quality.

This PR bumps up the bitrate of the test video by increasing the effective resolution of the static image in the test video.

Stress tests with this PR are encouraging: There were only 2 timeouts waiting for videos to load in https://github.com/Azure/communication-ui-library/actions/runs/2816968835

Compare with 50-60 timeouts in stress tests earlier (#2155 )


JamesBurnside (approved):

PorterNan (approved):

prprabhu-ms:
This PR introduces a different kind of flakiness: The UI snapshots now contain video compression artifacts...

PR #2176, opened 2022-08-08
VideoGallery Calling Participant tests

dmceachernmsft:
# What

Add tests for hermetic testing the video gallery
# Why

Allows us to run new tests for the video gallery to verify that the participant state is reflected properly in their video tile

https://skype.visualstudio.com/SPOOL/_workitems/edit/2919178
# How Tested

Ran locally
# Process & policy checklist


packages/react-components/src/types/OnRender.ts
```diff
@@ -36,7 +36,7 @@ export type CustomAvatarOptions = {
   text?: string;
   /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
   /** State for the participant to be displayed in the defaultPlaceHolder */
-  state?: ParticipantState;
+  participantState?: ParticipantState;
```
prprabhu-ms, line 39: Why the name change in this PR? Let's keep this PR focused on tests. Also, this is, unfortunately, called `state` in headless SDK, so perhaps we want to keep the naming aligned even in the components API?

packages/react-composites/tests/browser/call/hermetic/HorizontalGallery.test.ts
```diff
@@ -73,4 +74,117 @@ test.describe('HorizontalGallery tests', async () => {
...
+    );
+  });
+
+  test('HorizontalGallery should have multiple audio participants and 1 calling participant', async ({
+    page,
+    serverUrl
+  }) => {
+    test.skip(isTestProfileStableFlavor());
+
+    const paul = defaultMockRemoteParticipant('Paul Bridges');
```
prprabhu-ms, line 103: Let's see the snapshots for this, but it looks like there is only 1 audio participant here (contrary to the test title)

packages/react-composites/tests/browser/call/hermetic/HorizontalGallery.test.ts
```diff
@@ -73,4 +74,117 @@ test.describe('HorizontalGallery tests', async () => {
...
+
+    await page.goto(buildUrlWithMockAdapter(serverUrl, initialState));
+
+    await waitForSelector(page, dataUiId(IDS.videoGallery));
+    expect(await stableScreenshot(page, { dismissTooltips: true })).toMatchSnapshot(
+      'horizontal-gallery-with-joining-participant.png'
+    );
+  });
+
+  test('HorizontalGallery should have multiple audio participants and 1 calling participant', async ({
```
prprabhu-ms, line 97: I'm confused by the terminology here: audio vs calling. What is a calling participant? Do you mean video participant?

prprabhu-ms: Oh, you mean PSTN participant. Please let's use the word PSTN uniformly for this. "Calling" is so overloaded in our context.

prprabhu-ms: I'd recommend making these PSTN participants more realistic. They should have the right `userId` type, a phone number for their `displayName` etc. I recommend adding a helper `defaultMockPSTNParticipant` and then setting the `state` to `Connecting` for all the tests here.

prprabhu-ms:
Let's get those updated snapshots. I have nitty comments, overall looks good.

prprabhu-ms:
Please also stress test these new tests by running `node scripts/runBrowserTests.mjs -s 10` before merging.

It is much harder to fix flaky tests once they're in and they have an outsized (negative) impact on eng productivity. So I'm pushing for all of us to work harder to deflake before we add tests.

packages/react-components/src/types/OnRender.ts
anjulgarg: Recommend against making it `state`.
It is too generic. Moreover, we have used the term `participantState` in other components as well.

packages/react-composites/tests/browser/call/hermetic/HorizontalGallery.test.ts
dmceachernmsft: So the tests here are to mimic for 1:N and PSTN, should I make a helper for both types?

packages/react-components/src/types/OnRender.ts
prprabhu-ms: I also like `participantState` better, especially if we have precedent.

packages/react-composites/tests/browser/call/hermetic/HorizontalGallery.test.ts
prprabhu-ms: You can start with one. Add another only if the snapshots will differ.

packages/react-composites/tests/browser/call/hermetic/HorizontalGallery.test.ts
```diff
@@ -134,14 +133,12 @@ test.describe('HorizontalGallery tests', async () => {
     addVideoStream(fiona, true);
     const reina = defaultMockRemoteParticipant('Reina Takizawa');
     reina.isSpeaking = true;
-    const vasily = defaultMockRemoteParticipant('Vasily Podkolzin');
-    vasily.isMuted = true;
-    vasily.state = 'Connecting';
+    const phoneUser = defaultMockRemoteParticipant('+15555555555');
```
prprabhu-ms, line 136: Did you mean `defaultMockRemotePSTNParticipant`?

packages/react-composites/tests/browser/call/hermetic/fixture.ts
```diff
@@ -127,6 +127,51 @@ export function defaultMockRemoteParticipant(displayName: string): MockRemotePar
...
 
+/**
+ * Create the default {@link MockRemoteParticipantState} for a PSTN participant in a hermetic e2e test
+ *
+ * use to add PSTN participants to the {@link defaultCallAdapterState}
+ */
+export function defaultMockRemotePSTNParticipant(phoneNumber: string): MockRemoteParticipantState {
+  return {
+    identifier: { kind: 'phoneNumber', phoneNumber: `${phoneNumber}` },
+    state: 'Connecting',
```
prprabhu-ms, line 138: I'd set the "default" state to `Connected` and explicitly set the state in the test.

Why?

The test should be readable without jumping into the definition of any helper functions. The `state` is essential to understanding the test, so it should be set explicitly.

`default*` helper methods just provide a sensible value to use. Then explicitly set what matters so that the test code is both simple, and tells you exactly what is happening.

packages/react-composites/tests/browser/call/hermetic/fixture.ts
```diff
@@ -127,6 +127,51 @@ export function defaultMockRemoteParticipant(displayName: string): MockRemotePar
...
+
+/**
+ * Create the default {@link MockRemoteParticipantState} for a 1-N participant in a hermetic e2e test
+ *
+ * use to add 1-N participants to the {@link defaultCallAdapterState}
+ */
+export function defaultMockRemoteOneToNParicipant(displayName: string): MockRemoteParticipantState {
+  return {
+    identifier: { kind: 'phoneNumber', phoneNumber: `${displayName}` },
+    state: 'Ringing',
```
prprabhu-ms, line 154: Same comment as above.

prprabhu-ms (approved):

carocao-msft (approved):

packages/react-composites/tests/browser/call/hermetic/fixture.ts
dmceachernmsft: will update this!

PR #2177, opened 2022-08-08
Mock Adapter test for people pane dialpad 

carocao-msft:
# What
Mock Adapter test for people pane dialpad


anjulgarg:
The landscape mobile view for DialPad is clipping the 0 and call button.
Do we support landscape views?

carocao-msft:
> The landscape mobile view for DialPad is clipping the 0 and call button. Do we support landscape views?

I think for landscape view we need to scroll down to see the buttons?

JamesBurnside:
Not for this PR but just a couple styling issues from the screenshots:
| figma | current |
|--|--|
| image | image |

* icon should be black instead of themePrimary
* When active the button shouldn't show the gray background
* `People` -> `people`
* (hard to tell from the snapshot so this might not be an issue) Font weight of "Add People" looks like it is semibold instead of regular
* Down chevron isn't centered (this might be snapshots only and not repro in regular app in which case not an issue)

JamesBurnside:
> > The landscape mobile view for DialPad is clipping the 0 and call button. Do we support landscape views?
> 
> I think for landscape view we need to scroll down to see the buttons?

Can we make it fit in landscape while still retaining minimum touch targets? If so we should definitely try make it fit

anjulgarg (approved):

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) Microsoft Corporation.
```
prprabhu-ms, line 1: Beautiful test file. I could scan the whole file in 1 minute and understand what was happening!

prprabhu-ms (approved):

PR #2178, opened 2022-08-08
[draft] getting the callEnded event before the screen changes

alkwa-msft:
Updating when we send the event that the call has ended.

It turns out when you say call.hangup it will change the state and change up the UI in the composite.
We want to send the event that the call has ended beforehand so we can fix up our UI as necessary.

Since we are only potentially killing the DOM the adapter is still active and can continue to hangout, turn off mic + camera.

Thoughts?


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -97,13 +98,25 @@ type MainScreenProps = {
...
 
+  if (props.options?.endCallScreen === false) {
+    // keep the user on the call screen knowing that the developer will change up the UI at their earliest convenience
+    if (
+      page === 'leftCall' ||
+      page === 'joinCallFailedDueToNoNetwork' ||
+      page === 'accessDeniedTeamsMeeting' ||
+      page === 'removedFromCall'
+    ) {
+      page = 'call';
```
JamesBurnside, line 116: quick note: if going down this route we should use page's last render page. e.g. if I denied access via teams lobby it shouldn't try and render a call page
```
const page = useRef<Page>();
let previousPage = page.current;
page.current = useSelector(getPage);
```

JamesBurnside: I worry about this route as the props and selectors may rely on the call state being correctly in a call - we may get UI changing on the call page for split second when props are changed (e.g. I presume we don't have any remote participants any more when we render the call page after onCallEnded occurs)

JamesBurnside: Alternative way to handle emitting callEnded when the state is transitioning: #2201

alkwa-msft: I admit I primarily tested out this feature with leftCall which worked out well. I think we should definitely validate some of these other scenarios.

I like this last page concept.

JamesBurnside:
Changes encapsulated in #2201

PR #2179, opened 2022-08-08
[draft] another possibility for handling callEnded event

JamesBurnside:
ignore - just for discussions

related: #2178

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
+++ b/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
@@ -189,6 +189,12 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -189,6 +189,12 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
         return;
       }
 
+      // State is going from connected -> disconnected, emit the callEnded event.
+      const activeCall = this._call;
+      if (activeCall && activeCall.state === 'Connected' && clientState.calls[activeCall.id].state === 'Disconnected') {
```
JamesBurnside, line 194: not final code in any way, this would instead use the same function that is used to check for end call page and do:
```
if (getPage(currentState) !== 'endcallpage' && getPage(newState) === 'endcallpage')
```

PR #2180, opened 2022-08-08
Add rooms to Calling sample

mgamis-msft:
# What

Add rooms to Calling sample

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2940134

# How Tested

Tested sample on beta and stable

# Process & policy checklist


samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -165,6 +221,8 @@ export const HomeScreen = (props: HomeScreenProps): JSX.Element => {
                 props.startCallHandler({
                   displayName,
                   teamsLink,
+                  /* @conditional-compile-remove(rooms) */
```
alkwa-msft, line 224: this startCallhandler. How I am reading this. It means we can supply multiple different locators?I think we only want to ever have one locator?

samples/Calling/src/app/utils/AppUtils.ts
```diff
@@ -37,6 +41,43 @@ export const getGroupIdFromUrl = (): GroupLocator | undefined => {
...
+/* @conditional-compile-remove(rooms) */
+export const joinRoom = async (userId: string, roomId: string, role: Role): Promise<void> => {
+  const requestOptions = {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify({ acsUserId: userId, roomId: roomId, role: role })
+  };
+  const response = await fetch('http://localhost:7071/api/Rooms-AddParticipants', requestOptions);
```
alkwa-msft, line 75: can we export this localhost:7071 url to a common variable? I think in the future we are going to ask all backend functionality coming from a target url?

samples/Calling/src/app/App.tsx
```diff
@@ -1,8 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { GroupCallLocator, TeamsMeetingLinkLocator } from '@azure/communication-calling';
```
prabhjot-msft, line 4: Don't we need `GroupCallLocator` & `TeamsMeetingLinkLocator` in the app?

mgamis-msft: Yes, we need for the app. But not for this file. GroupCallLocator & TeamsMeetingLinkLocator are in the union type CallAdapterLocator.

alkwa-msft (approved):

prabhjot-msft (approved):

PR #2187, opened 2022-08-10
live browser tests: Mask videos before taking snapshot

prprabhu-ms:
# What

In browser tests,
* Mask video streams before taking UI snapshots
* Use a high bitrate video to avoid test flake caused by the Calling infrastructure's detection of low quality video stream.

# How tested

Stress test: https://github.com/Azure/communication-ui-library/runs/7768990193
succeeded, with 0 timeouts related to video loading delays, compared 50 - 60 in previous experiments in #2155


packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -437,6 +467,61 @@ const hideVideoLoadingSpinner = async (page: Page): Promise<void> => {
...
+const MASK_ATTRIB_KEY = 'data-ui-id';
+const MASK_ATTRIB_VALUE = 'stream-media-video-mask';
+const MASK_SELECTOR = `[${MASK_ATTRIB_KEY}="${MASK_ATTRIB_VALUE}"]`;
+/**
+ * Masks the video element in {@link VideoTile} by overlaying an opaque <div/>.
+ *
+ * This function is destructive: It changes the video tile's DOM irreversibly.
+ * This function is idempotent: It is safe to call this function multiple times.
+ */
+async function maskVideos(page: Page): Promise<void> {
```
JamesBurnside, line 479: note: we should only need to do this for live tests, the hermetic tests (I think) all have fake video DOM already

prprabhu-ms: Yeah, I noticed the unnecessary snapshot updates for the hermetic tests.

I'm undecided on whether I prefer to have uniform snapshots and code (one less difference to learn between live vs hermetic) or different snapshots.

prprabhu-ms:
Stress test running: https://github.com/Azure/communication-ui-library/runs/7768990193

Partial results encouraging: 
image

packages/react-composites/tests/browser/common/utils.ts
prprabhu-ms: On the balance, I think the consistency between hermetic / live tests beats an extra DOM node during testing.

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -437,6 +467,61 @@ const hideVideoLoadingSpinner = async (page: Page): Promise<void> => {
   await page.addStyleTag({ content: '[data-ui-id="stream-media-loading-spinner"] {display: none}' });
 };
 
+const MASK_ATTRIB_KEY = 'data-ui-id';
+const MASK_ATTRIB_VALUE = 'stream-media-video-mask';
+const MASK_SELECTOR = `[${MASK_ATTRIB_KEY}="${MASK_ATTRIB_VALUE}"]`;
+/**
+ * Masks the video element in {@link VideoTile} by overlaying an opaque <div/>.
```
JamesBurnside, line 474: Can you add a comment here why someone would want to mask the videos in their test/why we are masking videos? (Just hey wow artifacts killed our UI test stability so applying a mask here!)

JamesBurnside (approved):

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -437,6 +467,61 @@ const hideVideoLoadingSpinner = async (page: Page): Promise<void> => {
...
+        const children = Array.from(node.children ?? []) as HTMLElement[];
+        const masks = children.filter((child) => child.getAttribute(mask_key) === mask_value);
+        if (masks.length > 0) {
+          // Already have a mask. Nothing to do.
+          return;
+        }
+
+        const mask = document.createElement('div');
+        mask.setAttribute(mask_key, mask_value);
+        mask.setAttribute('style', 'position: absolute; height: 100%; width: 100%; background: green;');
```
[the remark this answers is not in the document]
prprabhu-ms: I did have some fun deciding what video to run behind the opaque div.

It's hidden, so everything is fair game, right? right?

carocao-msft (approved):

PR #2188, opened 2022-08-10
E2E tests for configuration screen for different roles

prabhjot-msft:
# What
E2E tests for configuration screen for different roles

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2919227

# How Tested

# Process & policy checklist


prprabhu-ms:
This looks great. Let's have those snapshots so we can review the golden files as well.

packages/react-composites/tests/browser/call/app/QueryArgs.ts
```diff
@@ -31,6 +33,7 @@ export function parseQueryArgs(): QueryArgs {
     userId: params.userId ?? '',
     groupId: params.groupId ?? '',
     token: params.token ?? '',
-    displayName: params.displayName ?? ''
+    displayName: params.displayName ?? '',
+    role: params.role as Role
```
alkwa-msft, line 37: If they don't pass in a role should we assume they are a presenter?

alkwa-msft (approved):

packages/react-composites/tests/browser/call/app/QueryArgs.ts
mgamis-msft: I would say no. We don't want to pass the role Presenter to the composite for non-rooms calls.

carocao-msft (approved):

packages/react-composites/tests/browser/call/app/QueryArgs.ts
prabhjot-msft: It is an optional prop, so I don't think we need to set default value

dmceachernmsft:
image
is this update intentional?

prabhjot-msft:
> image
> is this update intentional?

No, got updated automatically

prprabhu-ms (approved):

packages/react-composites/tests/browser/call/app/QueryArgs.ts
JamesBurnside, line 37: nit: `as role | undefined;`

PR #2197, opened 2022-08-10
Add hold button to Call composite control bar

dmceachernmsft:
# What

Add more button with hold button to Calling composite control bar
# Why

Allows end user to hold a call that they are in

https://skype.visualstudio.com/SPOOL/_workitems/edit/2943681
# How Tested

image
inspected and ran in sample application


carocao-msft (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -61,6 +65,37 @@ export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX
...
+      tooltipOffContent: localeStrings.strings.callWithChat.moreDrawerButtonTooltip
+    }),
+    [localeStrings]
+  );
+
+  /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
+  const holdButtonProps = usePropsFor(HoldButton);
+
+  /* @conditional-compile-remove(one-to-n-calling) */ /* @conditional-compile-remove(one-to-n-calling) */
+  const moreButtonContextualMenuItems = (): IContextualMenuItem[] => {
```
anjulgarg, line 81: Might be worth memoizing this.

anjulgarg (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -61,6 +65,37 @@ export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX
...
+    const items: IContextualMenuItem[] = [];
+
+    items.push({
```
JamesBurnside, line 84: Just to confirm: this will be added for group calls and teams interop calls also

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -113,6 +148,17 @@ export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX
           {isEnabled(options?.devicesButton) && (
             <Devices displayType={options?.displayType} increaseFlyoutItemSize={props.increaseFlyoutItemSize} />
           )}
+          {
+            /* @conditional-compile-remove(one-to-n-calling) */ /* @conditional-compile-remove(one-to-n-calling) */
+            isEnabled(options?.moreButton) && (
```
JamesBurnside, line 153: IIRC we can only hold 5 buttons in the width of the call page on small mobile devices so 6 would be too many:
image

Can you confirm if on the likes of iphone SE or the galaxy fold narrow screen does this fit?
If not we may need to move people into the overflow (is that breaking change?)

JamesBurnside (changes requested):
re: 6 buttons fitting

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
anjulgarg: Screen Shot 2022-08-10 at 4 33 56 PM

Works :)

mgamis-msft (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
dmceachernmsft: I was going to bring this up in parking lot, my testing revealed to me this works in those kinds of calls to

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
dmceachernmsft: image
People button is now in more button overflow on mobile since it opens a people pane now and isn't a context menu

JamesBurnside: We should discuss as a team if we want to release this as a 2.0.0 release or just a minor release (we are at the point where we have a long list of breaking changes we want to make so maybe its time)

JamesBurnside (approved):

PR #2198, opened 2022-08-10
Fix react hook console errors

mgamis-msft:
# What

Fix _usePermissions usage. Not calling it after a logical AND/OR operator which causes React hook console errors

# Why

There are React hook console errors as shown in this video for any type of call.
https://user-images.githubusercontent.com/79475487/184026374-f849fc09-b6bf-4e2c-bc82-4c91dad8dfe4.mp4

# How Tested

Local sample app test

# Process & policy checklist


packages/react-components/src/components/CameraButton.tsx
```diff
@@ -150,7 +150,9 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
 
   let disabled = props.disabled || waitForCamera;
   /* @conditional-compile-remove(rooms) */
-  disabled = disabled || !_usePermissions().cameraButton;
+  const permissions = _usePermissions();
+  /* @conditional-compile-remove(rooms) */
+  disabled = disabled || !permissions.cameraButton;
```
alkwa-msft, line 155: are these the same but we just want to give them more space?

alkwa-msft (approved):

packages/react-components/src/components/CameraButton.tsx
mgamis-msft: I believe the logical OR here is conditionally calling a react hook.

prabhjot-msft (approved):

JamesBurnside (approved):

PR #2200, opened 2022-08-10
Call Composite Hermetic Tests for Participant List

anjulgarg:
# What
Call Composite Hermetic Tests for Participant List

# Why

# How Tested

# Process & policy checklist


prprabhu-ms:
Replacing myself with @dmceachernmsft as assignee to spread the knowledge. Both @carocao-msft  and @dmceachernmsft have recently written hermetic tests and they have the mojo ;)

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -86,4 +87,41 @@ test.describe('Participant pane tests', async () => {
...
+    const paul = defaultMockRemoteParticipant('Paul Bridges');
+    paul.state = 'Connecting';
+    const participants = [paul];
+    const initialState = defaultMockCallAdapterState(participants);
+    await page.goto(buildUrlWithMockAdapter(serverUrl, initialState));
+    await waitForSelector(page, dataUiId(IDS.videoGallery));
+    await pageClick(page, dataUiId('call-composite-participants-button'));
+    await waitForSelector(page, dataUiId('call-composite-people-pane'));
+    await hidePiPiP(page);
+    expect(await stableScreenshot(page)).toMatchSnapshot('PSTN-participant-pane-callee-name.png');
```
prprabhu-ms, line 102: Should this be called `PSTN-participant-pane-connecting-participant`?

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -86,4 +87,41 @@ test.describe('Participant pane tests', async () => {
 
     expect(await stableScreenshot(page)).toMatchSnapshot(`PSTN-call-screen-with-dialpad.png`);
   });
+
+  test.only('callee participant is displayed with connection state', async ({ page, serverUrl }) => {
+    test.skip(isTestProfileStableFlavor());
+    const paul = defaultMockRemoteParticipant('Paul Bridges');
```
prprabhu-ms, line 93: @dmceachernmsft recently added a better helper to create 1:N and PSTN remote participants. Please use one of those so that the state is more realistic.

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -86,4 +87,41 @@ test.describe('Participant pane tests', async () => {
...
+    await waitForSelector(page, dataUiId(IDS.videoGallery));
+    await pageClick(page, dataUiId('call-composite-participants-button'));
+    await waitForSelector(page, dataUiId('call-composite-people-pane'));
+    await hidePiPiP(page);
+    expect(await stableScreenshot(page)).toMatchSnapshot('PSTN-participant-pane-callee-name.png');
+  });
+
+  test.only('callee participant name and connection state are truncated', async ({ page, serverUrl }) => {
+    test.skip(isTestProfileStableFlavor());
+    const longPaul = defaultMockRemoteParticipant(
```
prprabhu-ms, line 107: :D

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -86,4 +87,41 @@ test.describe('Participant pane tests', async () => {
...
+    const initialState = defaultMockCallAdapterState(participants);
+    await page.goto(buildUrlWithMockAdapter(serverUrl, initialState));
+    await waitForSelector(page, dataUiId(IDS.videoGallery));
+    await pageClick(page, dataUiId('call-composite-participants-button'));
+    await waitForSelector(page, dataUiId('call-composite-people-pane'));
+    await hidePiPiP(page);
+    const participantStringId = dataUiId('participant-item-state-string');
+    await page.evaluate((participantStringId) => {
+      const el = document.querySelector(participantStringId);
+      if (el) {
```
prprabhu-ms, line 121: This is a bit hacky. Is there no way we can test actually need to truncate the connection state string?

How about zooming in so all text is larger and hence needs to be truncated?

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -565,3 +565,15 @@ export async function blockForMinutes(m: number): Promise<void> {
     setTimeout(() => resolve(), 1000 * 60 * m);
   });
 }
+
+/**
+ * Hides the PiPiP video tile.
+ * Useful when testing participant list on mobile devices
+ * where the content being tested is hidden behind the PiPiP video tile.
```
prprabhu-ms, line 572: Thanks for adding this very useful comment.

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -86,4 +87,41 @@ test.describe('Participant pane tests', async () => {
 
     expect(await stableScreenshot(page)).toMatchSnapshot(`PSTN-call-screen-with-dialpad.png`);
   });
+
+  test.only('callee participant is displayed with connection state', async ({ page, serverUrl }) => {
+    test.skip(isTestProfileStableFlavor());
```
prprabhu-ms, line 92: Any chance we can run these tests for both flavors (on stable we don't expect the connection string to be shown).

This is something I learnt from https://github.com/Azure/communication-ui-library/pull/2191

When writing beta-only features, we should think about browser tests that validate that the feature is *not* in the stable build.

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
anjulgarg: Doesn't matter, we are just testing the names here.

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
anjulgarg: Zooming in changes the entire UI and might not reveal inconsistencies we end up introducing in the future.
Not sure how this is hacky, we do something similar to stub out timestamps as well.
Ultimately, what we are testing here is the css

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
anjulgarg: Good idea.

anjulgarg: Unfortunately, we can't do this for these tests as we are introducing a new participant pane all together and the same tests won't work for stable flavour

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
dmceachernmsft: there isn't much of a risk but I think the idea here is for readability. I think we should be able to read the tests and know exactly what is happening and using the kind of participant as PSTN could help us determine what kind of call it is.

dmceachernmsft (approved):

carocao-msft (approved):

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
callpraths: Realized later we should write localization tests for this instead

PR #2201, opened 2022-08-10
Call `onCallEnded` at the time the state is changing and add `callEndReason`

JamesBurnside:
# What
* Ensure we call onCallEnded before the state transaction occurs preventing a flash of rendering the composite page before the Contoso Page
* Add `callEndReason` to the onCallEnded event so Contoso can decide an action to take

# Why
Currently it's called too late to act on - if Contoso wants to show their own end call page they would have a flash where our call ended page shows, then theirs would render. This ensures the onCallEnded gets triggered before the render pass that renders our call ended screen occurs.

# How Tested
subscribed to onCallEnded, received this payload:
image

And internal leftCall break point that renders the composite's left call page was successfully not triggered:
image

Tested following end call events:
* Hanging up
* Being removed from a call
* Being denied access to a call
* Network loss (not working - but unrelated to this change, bug filed)


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -188,6 +188,12 @@ export interface BaseCustomStyles {
 export interface CallAdapter extends AdapterState<CallAdapterState>, Disposable, CallAdapterCallManagement, CallAdapterDeviceManagement, CallAdapterSubscribers {
 }
 
+// @public
+export type CallAdapterCallEndedEvent = {
+    callId?: string;
+    callEndReason?: CallEndReason;
```
JamesBurnside, line 194: @alkwa-msft fyi this contains both errorCode and errorSubCode

anjulgarg (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -112,19 +113,41 @@ class CallContext {
...
-      page: getCallCompositePage(call, latestEndedCall),
-      endedCall: latestEndedCall,
-      devices: clientState.deviceManager,
-      latestErrors: clientState.latestErrors
-    });
+
+    // As the state is transitioning to a new state, trigger appropriate callback events.
+    const oldPage = this.state.page;
+    const newPage = getCallCompositePage(call, latestEndedCall);
+    if (!IsCallEndedPage(oldPage) && IsCallEndedPage(newPage)) {
```
anjulgarg, line 131: Let's say customer was subscribed to the `callEnded` event.
They have custom logic for showing an end call screen when this event is triggered, but they have a timeout that renders that page after 5 seconds.
They would see our end call page for 5 seconds before seeing their own right?


Just trying to make sure I understand how this fix is working...

anjulgarg (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
JamesBurnside: Yep that's correct

edwardlee-msft (approved):

PR #2202, opened 2022-08-11
Add resume screen to Calling composite

dmceachernmsft:
# What

Adds a Hold Pane component to Calling Composite
# Why

When call goes to 'localHold' state we need to show another display to show the end user how long they have been on hold. also provides a way for them to resume the call.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2945167
# How Tested

image
Ran locally in sample application


packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -136,6 +145,8 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
           inviteLink={props.callControlProps.callInvitationURL}
         />
       );
+    } else if (adapter && callStatus === 'LocalHold') {
```
JamesBurnside, line 148: not saying its the right thing but just asking if it was evaluated whether the OnHold page should be a CompositePage state?

JamesBurnside (changes requested):
re: time calculation

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const { onToggleHold } = props;
+  const locale = useLocale();
+  const theme = useTheme();
+
+  const [rawTime, setRawTime] = useState<number>(0);
+  const [time, setTime] = useState<string>('00:00');
+
+  React.useEffect(() => {
+    const interval = setInterval(() => {
+      setRawTime((time) => time + 10);
```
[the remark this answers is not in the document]
anjulgarg: OR maybe we can use an open source library for this?

packages/react-composites/src/composites/CallComposite/styles/HoldPane.styles.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * styles for hold pane main text
+ *
+ * @private
+ */
+export const holdPaneLabelStyles = {
+  fontWeight: 600,
+  fontHeight: _pxToRem(22),
+  fontSize: _pxToRem(16),
+  fontFamily: 'Segoe UI',
```
anjulgarg, line 36: Let's use the theme default font here and avoid defining a font family

packages/react-composites/src/composites/CallComposite/styles/HoldPane.styles.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * styles for hold pane timer
+ *
+ * @private
+ */
+export const holdPaneTimerStyles = {
+  fontWeight: 600,
+  fontHeight: _pxToRem(22),
+  fontSize: _pxToRem(16),
+  fontFamily: 'Segoe UI',
```
anjulgarg, line 50: Same here

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
anjulgarg: thanks for pointing this out @JamesBurnside 
We didn't evaluate this approach.
My recommendation would also be to make this a composite state instead of doing it in Call Arrangement.
Feel free to reach out in person and discuss

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @beta
+ */
+export const HoldPane = (props: HoldPaneProps): JSX.Element => {
+  const { onToggleHold } = props;
+  const locale = useLocale();
+  const theme = useTheme();
+
+  const [rawTime, setRawTime] = useState<number>(0);
+  const [time, setTime] = useState<string>('00:00');
```
JamesBurnside, line 34: we don't need two states here, one for time as a number and one for time as a string, simply store the duration number in state and for the string you can do that calculation in the regular body:
```
const [timeElapsed, setTimeElapsed] = useState<number>(0);
const elapsedTimeString = `${getMinutes(timeElapsed)}:${getSeconds(timeElapsed)}`;
```

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  React.useEffect(() => {
+    const interval = setInterval(() => {
+      setRawTime((time) => time + 10);
+
+      const mins: string =
```
JamesBurnside, line 40: nit: to make it easy to read this block, i'd recommend pulling out these helpers to const functions to the bottom of this file:
```
const getMinutes = (duration: number): string =>  Math.floor....
const getSeconds = (duration: number): string =>  Math.floor....
```

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,73 @@
...
+        Math.floor((rawTime / 60000) % 60) < 10
+          ? '0' + Math.floor((rawTime / 60000) % 60)
+          : `${Math.floor((rawTime / 60000) % 60)}`;
+
+      const seconds: string =
+        Math.floor((rawTime / 1000) % 60) < 10
+          ? '0' + Math.floor((rawTime / 1000) % 60)
+          : `${Math.floor((rawTime / 1000) % 60)}`;
+
+      setTime(mins + ':' + seconds);
```
JamesBurnside, line 50: What about if I've been on hold for more than an hour?

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
JamesBurnside: Difficulty here is I've realized we've worked ourselves into a corner where this is kinda a breaking change to add a new page in the CompositePage state?
Like if someone is doing `if (state.page !== 'Call') { ... }` they'll be broken and in a way that's hard to know... Not sure the best thing here

packages/react-composites/src/composites/CallComposite/styles/HoldPane.styles.ts
```diff
@@ -0,0 +1,79 @@
...
+ * styles for hold pane main text
+ *
+ * @private
+ */
+export const holdPaneLabelStyles = {
+  fontWeight: 600,
+  fontHeight: _pxToRem(22),
+  fontSize: _pxToRem(16),
+  fontFamily: 'Segoe UI',
+  fontStyle: 'normal',
```
JamesBurnside, line 37: Is this necessary to specify?

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const getMinutes = (time: number): number => {
+  return Math.floor((time / 60000) % 60);
+};
+
+const getSeconds = (time: number): number => {
+  return Math.floor((time / 1000) % 60);
+};
+
+const getHours = (minutes: number): number => {
```
JamesBurnside, line 73: nit: for consistency instead of passing `minutes` in here, pass the same as the other functions and instead have:
```
const getHours = (time: number): number => Math.floor(getMinutes(time) / 60)
```

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,81 @@
...
+  return (minutes / 60) % 60;
```
JamesBurnside, line 74: this seems incorrect here my guess is it should be `Math.floor(minutes / 60)` instead of modulo 60

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+const getReadableTime = (hours: number, minutes: number, seconds: number): string => {
```
JamesBurnside, line 77: nit: for ease of use consider
```
const durationToReadableString = (duration: number): string => {
  const hours = getHours(duration);
  const minutes = getMinutes(duration);
  const seconds = getSeconds(duration);
  ...
}

File: packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx (hunk 10 above)
Comment (dmceachernmsft) [reply packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx:74]: this will display a decimal version of a hour if we don't do the modulus

File: packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx [hunk 12]
```diff
@@ -0,0 +1,81 @@
...
+  const readableMinutes = ('0' + minutes).slice(-2);
+  const readableSeconds = ('0' + seconds).slice(-2);
+  return hours >= 1 ? hours + ':' + readableMinutes + ':' + readableSeconds : readableMinutes + ':' + readableSeconds;
```
JamesBurnside, line 80: so many colons 😵‍💫😄
(again very optional nit) for readability consider wrapping the hours check to only affect the hours part:
```
const optionalHoursString = hours > 0 ? hours + ':' ? '';
return optionalHoursString + ':' + readableMinutes + ':' + readableSeconds;
```
or consider using string template:
```
return `${hours > 0 ? hours + ':' : ''}:${readableMinutes}:${readableSeconds}`
```

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,81 @@
...
+const getSeconds = (time: number): number => {
+  return Math.floor((time / 1000) % 60);
+};
+
+const getHours = (minutes: number): number => {
+  return (minutes / 60) % 60;
+};
+
+const getReadableTime = (hours: number, minutes: number, seconds: number): string => {
+  const readableMinutes = ('0' + minutes).slice(-2);
```
JamesBurnside, line 78: fun way of doing this! 😍

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
dmceachernmsft: I like the second option 😊

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -103,7 +105,7 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
   const locale = useLocale();
 
   let pageElement: JSX.Element;
-
+  console.log(page);
```
JamesBurnside, line 108: remove

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const startTime = useRef(performance.now());
+
+  React.useEffect(() => {
+    const interval = setInterval(() => {
+      setTime(performance.now() - startTime.current);
+    }, 10);
+    return () => {
+      clearInterval(interval);
+    };
+  });
```
JamesBurnside, line 45: no dep array here so will currently execute on every render pass

JamesBurnside, line 45: checking if expected: this has no dependency array so will be executed every render pass

packages/react-composites/src/composites/CallComposite/styles/HoldPane.styles.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import { IButtonStyles, IStackStyles, ITextStyles } from '@fluentui/react';
+import { _pxToRem } from '@internal/acs-ui-common';
+
+/**
+ * styles for hold pane resume button
+ *
+ * @private
+ */
+export const resumeButtonStyles = (): IButtonStyles => {
```
JamesBurnside, line 12: nit: don't use functions for these as the result would need to be memoized, instead just use a static object

packages/react-composites/src/composites/CallComposite/pages/HoldPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * @beta
+ */
+export const HoldPage = (props: HoldPageProps): JSX.Element => {
+  const errorBarProps = usePropsFor(ErrorBar);
+  const strings = useLocale().strings.call;
+
+  const callControlOptions = props.mobileView
+    ? reduceCallControlsForMobile(props.options?.callControls)
+    : props.options?.callControls;
+  console.log('hold page');
```
JamesBurnside, line 33: remove

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const getSeconds = (time: number): number => {
+  return Math.floor((time / 1000) % 60);
+};
+
+const getHours = (time: number): number => {
+  return Math.floor(getMinutes(time) / 60);
+};
+
+const getReadableTime = (time: number): string => {
```
JamesBurnside, line 78: add unit test for this fn

JamesBurnside (approved):

anjulgarg (approved):

edwardlee-msft (approved):

PR #2204, opened 2022-08-11
[WIP] Fetch translated strings for `1.3.1` release

prprabhu-ms:
TODO: Fetch strings once `dismissSidePaneButtonLabel` has been translated.

* `1.3.0` stable release shipped with this string untranslated.
  * Compare: [en-US](https://github.com/Azure/communication-ui-library/blob/1.3.0/packages/react-composites/src/composites/localization/locales/en-US/strings.json#L65) vs [es-ES](https://github.com/Azure/communication-ui-library/blob/1.3.0/packages/react-composites/src/composites/localization/locales/es-ES/strings.json#L65)
  * This bug was fixed in [1.3.1-beta.1](https://github.com/Azure/communication-ui-library/blob/1.3.1-beta.1/packages/react-composites/src/composites/localization/locales/es-ES/strings.json#L66)

Fix this bug for the `1.3.1` stable release.


prprabhu-ms:
Decided to keep the hotfix release targeted and safe. This fix will (organically) be included in the next stable release.

PR #2205, opened 2022-08-11
Enable inlineSource, remove d.ts.map, and fix source not found warnings

PorterNan:
# What

https://skype.visualstudio.com/SPOOL/_workitems/edit/2946975

This is a bug that we pack .map file into the package but not .ts file, the expectation here is we don't provide any source and source maps in our npm packages (because they are for debug purposes)

# Why

# How Tested

# Process & policy checklist


JamesBurnside:
We do intentional ship the source maps per azure policy guidelines:
image
https://azure.github.io/azure-sdk/typescript_design.html

As a comparison, communication-chat so ship the .map file in their package but not the ts file - will theirs also be broken in the same way?
image

JamesBurnside (changes requested):
re: source map inclusion check

PorterNan:
> > re: source map inclusion check
> 
> Wow, just double checked what chat-client is doing, they packed source into the .map file: image Maybe that's what we wanna do since it wouldn't increase our final bundle size

Interesting that I found the template that azure js sdk is using:
https://github.com/Azure/azure-sdk-for-js/blob/main/tsconfig.json

They are doing embed source into sourcemap, let's align with them

PorterNan:
> > > re: source map inclusion check
> > 
> > 
> > Wow, just double checked what chat-client is doing, they packed source into the .map file: image Maybe that's what we wanna do since it wouldn't increase our final bundle size
> 
> Interesting that I found the template that azure js sdk is using: https://github.com/Azure/azure-sdk-for-js/blob/main/tsconfig.json
> 
> They are doing embed source into sourcemap, let's align with them

Update - d.ts.map is for developers to navigate in IDE, which requires running under our repo since we don't include real source in our npm package - so it makes no sense to include them and it is causing lots of warnings (also this is what chat sdk is doing)

we add inline source into js.map so people are able to debug our code a bit in browser with no warnings (also that's what chat sdk does):
image

JamesBurnside (approved):

dmceachernmsft (approved):

PR #2208, opened 2022-08-11
Fix `snapshot:update` command

JamesBurnside:
# What
Broke recently when the scripts were added

# Why
`rushx snapshot:update` was not working, failed with not finding any snapshots:
image

# How Tested
Locally, `rushx snapshot:update` works:
image


prabhjot-msft (approved):

packages/storybook/package.json
```diff
@@ -8,9 +8,9 @@
     "start": "node ./scripts/start.mjs",
     "deploy-storybook": "node ./scripts/deploy.mjs",
     "test": "node ./scripts/test.mjs",
-    "test:coverage": "rushx test -- --coverage",
+    "test:coverage": "rushx test --coverage",
     "test:snippets": "tsc --p tsconfig.snippets.json",
-    "snapshot:update": "rushx test -- --update-snapshot",
+    "snapshot:update": "rushx test --update-snapshot",
```
callpraths, line 13: This should stay unchanged I think. There is a -u flag in runBeowserScripts. Perhaps the long name is called just --uodate

prprabhu-ms: https://github.com/Azure/communication-ui-library/blob/main/packages/react-composites/scripts/runBrowserTests.mjs#L109

(I'm also OK with renaming the flag name in `runBrowserScripts` to `--update-snapshots`)

prprabhu-ms: Oh wait... I'm soo confused about what packge this is.

prprabhu-ms (approved):

PR #2210, opened 2022-08-11
Video Tile displays Hold and Ringing states of participants in responsive manner

anjulgarg:
# What
Video Tile displays Hold and Ringing states of participants in responsive manner

https://user-images.githubusercontent.com/9782747/184734626-d69dfd6e-4e2c-4df5-b696-5dde2c55fdea.mov

https://user-images.githubusercontent.com/9782747/184734639-4d5d503f-cd48-4450-8f94-21ce83561411.mov

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -109,12 +111,24 @@ export const _RemoteVideoTile = React.memo(
         isMuted={isMuted}
         isSpeaking={isSpeaking}
         showMuteIndicator={showMuteIndicator}
-        showLabel={props.showLabel}
         personaMinSize={props.personaMinSize}
         /* @conditional-compile-remove(one-to-n-calling) */
         /* @conditional-compile-remove(PSTN-calls) */
+        // When the remote video tile is rendered as a small tile in horizontal gallery,
+        // we hide the participants name if they are in hold/connecting states.
+        showLabel={isNarrow && isCallingOrHold(participantState) ? false : props.showLabel}
```
dmceachernmsft, line 119: this is sneaky I love it

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -109,12 +111,24 @@ export const _RemoteVideoTile = React.memo(
...
     );
   }
 );
+
+/**
+ * Checks if a participant state is calling or hold.
+ * These states match the states used to render the participantStateString in VideoTile and ParticipantItem.
+ */
+const isCallingOrHold = (participantState?: ParticipantState): boolean => {
+  return !!participantState && ['Idle', 'Connecting', 'EarlyMedia', 'Ringing', 'Hold'].includes(participantState);
```
dmceachernmsft, line 133: Do all of these states fit for this? I don't know how `idle` kicks off but are we wanting to display something in the videoGallery for it?

dmceachernmsft (approved):

packages/react-components/src/components/RemoteVideoTile.tsx
anjulgarg: So Idle is the starting state of a participant, we have logic in participant item and video tile to show this state as Connecting

carocao-msft (approved):

JamesBurnside (changes requested):
I see two approvals but also a [DO NOT REVIEW] in the title so just doing a super quick pass here
Couple things from the snapshots:

Regression showing name in the PiPiP:
image

Feedback for design, on a 1-1 this state will be hidden by the floating video tile by default:
image

anjulgarg:
> I see two approvals but also a [DO NOT REVIEW] in the title so just doing a super quick pass here Couple things from the snapshots:
> 
> Regression showing name in the PiPiP: image
> 
> Feedback for design, on a 1-1 this state will be hidden by the floating video tile by default: image

Correct. That's why it's set to DNR. 
Need to fix this. :)

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -99,22 +113,80 @@ export const _RemoteVideoTile = React.memo(
...
+};
+
+/* @conditional-compile-remove(one-to-n-calling) */
+/* @conditional-compile-remove(PSTN-calls) */
+/**
+ * Determines if a label should be shown for a remote video tile.
+ * When the remote video tile is rendered as a small tile in horizontal gallery,
+ * we hide the participants name if they are in hold/connecting states.
+ */
+const canShowLabel = (
```
dmceachernmsft, line 148: is it possible to move this logic into the VideoGallery? this is a layer up from the VideoTile component but its in a interesting spot living in the internal wrapper (_RemoteVideoTile) for the VideoTile that the VideoGallery uses.

dmceachernmsft: Actually never mind, doing this here is very clever since Contoso has the ability to override this in the VideoGallery level.

anjulgarg: Plus video gallery is already doing too much

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -52,12 +57,13 @@ export const _RemoteVideoTile = React.memo(
...
-      showMuteIndicator,
-      /* @conditional-compile-remove(one-to-n-calling) */
-      /* @conditional-compile-remove(PSTN-calls) */
-      participantState
+      showMuteIndicator
     } = props;
 
+    const containerRef = React.useRef<HTMLDivElement>(null);
+    /* @conditional-compile-remove(one-to-n-calling) */ /* @conditional-compile-remove(PSTN-calls) */
+    const containerWidth = _useContainerWidth(containerRef);
```
JamesBurnside, line 65: This gets the container width, but also requires an `isNarrow` prop be passed in - can just one of these be required?

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -99,22 +105,67 @@ export const _RemoteVideoTile = React.memo(
...
+ * When the remote video tile is rendered as a small tile in horizontal gallery,
+ * we hide the participants name if they are in hold/connecting states.
+ */
+const canShowLabel = (
+  participantState?: ParticipantState,
+  isNarrow?: boolean,
+  showLabel?: boolean,
+  containerWidth?: number
+): boolean | undefined => {
+  const isCallingOrHold = (participantState?: ParticipantState): boolean => {
```
JamesBurnside, line 154: nit: this can be defined outside this function
(this type of function is a good candidate for a unit test that can catch an error when a new participant state is added and this function may or may not need updating)

packages/react-components/src/components/RemoteVideoTile.tsx
JamesBurnside: (is isNarrow necessary or can we just use the container width here?)

JamesBurnside: (better would be if the parent could just use the showLabel prop but I'm guessing the parent won't know this tile's width well enough)

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -290,32 +264,28 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
-                strings={strings}
-              />
+              <DefaultPlaceholder {...placeholderOptions} />
             )}
           </Stack>
         )}
 
-        {showLabel && (displayName || (showMuteIndicator && isMuted)) && (
-          <Stack horizontal className={tileInfoContainerStyle}>
+        <Stack horizontal className={tileInfoContainerStyle} tokens={tileInfoContainerTokens}>
```
JamesBurnside, line 272: If there is `showLabel && (displayName || (showMuteIndicator && isMuted)` and `participantStateString` are both false, we probably shouldn't have this empty stack

JamesBurnside (commented):
Two snapshot things:
PIPIP seems to not be showing the remote video:
image

Might be too fiddly to fix, one a very small tile the tile content looks quite off-centered vertically:
image

packages/react-components/src/components/RemoteVideoTile.tsx
anjulgarg: Good idea @JamesBurnside 
isNarrow does seem redundant now. 
Will remove it.
And yes, doing this inside video gallery requires quite a bit of refactor.

JamesBurnside (approved):

PR #2211, opened 2022-08-11
People button with role permissions

mgamis-msft:
# What

Add role permissions to People button

# Why

People button needs to be disabled for the 'Consumer' role when room calls is released

# How Tested

Local sample testing.
image

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/components/buttons/People.tsx
```diff
@@ -15,6 +17,14 @@ const icon = (): JSX.Element => <CallCompositeIcon iconName={'ControlButtonParti
 /** @beta */
 export const People = (props: ControlBarButtonProps): JSX.Element => {
   const { strings, onRenderOnIcon, onRenderOffIcon, onClick } = props;
+
+  /* @conditional-compile-remove(rooms) */
```
PorterNan, line 21: This line shouldn't be conditionally complied?

mgamis-msft: We are spreading props directly to ControlButton (on line 45) in stable. So isn't this line is not needed in stable?

prabhjot-msft: I think we only need to conditionally compile the permissions statement `disabled = disabled || !_usePermissions().participantList`

mgamis-msft: But the line: 
```let disabled = props.disabled;```
is not needed in stable.

PorterNan: Got it, so it's all beta feature even disable prop

PorterNan (approved):

prabhjot-msft (approved):

PR #2215, opened 2022-08-12
Disable control bar for CallWithChat and Calling Composite when on Hold

dmceachernmsft:
# What

Disables buttons when in hold 
# Why

Disables the control bar buttons that cannot be actioned when on hold. 

https://skype.visualstudio.com/SPOOL/_workitems/edit/2950563
# How Tested

image
image
ran locally and inspected behaviors


packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -81,6 +81,9 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
   /* @conditional-compile-remove(one-to-n-calling) */
   const { callStatus } = useSelector(callStatusSelector);
 
+  /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
+  const disableControlsForHoldScreen = adapter.getState().page === 'hold';
```
JamesBurnside, line 85: can this be done at the selector level instead of plumbing through the composite? i.e. the usePropsFor(CameraButton) will return a `disabled: true` when on hold

JamesBurnside, line 85: can this be done at the selector level instead of plumbing through the composite? i.e. the usePropsFor(CameraButton) will return a `disabled: true` when on hold

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -209,6 +237,9 @@ export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX
                 menuIconProps={{ hidden: true }}
                 menuProps={{ items: moreButtonContextualMenuItems() }}
                 showLabel={!props.isMobile}
+                /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */ disabled={
+                  props.disableForHoldScreen
```
JamesBurnside, line 241: i think the more button should only be disabled iff all child items are disabled (this will allow support for custom button injection into the more button to still function correctly)

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -172,10 +174,30 @@ export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX
             occluding some of its content.
          */}
         <ControlBar layout="horizontal" styles={controlBarStyles(theme.semanticColors.bodyBackground)}>
-          {isEnabled(options?.microphoneButton) && <Microphone displayType={options?.displayType} />}
-          {isEnabled(options?.cameraButton) && <Camera displayType={options?.displayType} />}
+          {isEnabled(options?.microphoneButton) && (
+            <Microphone
+              displayType={options?.displayType}
+              /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */ disabled={
+                props.disableForHoldScreen
```
JamesBurnside, line 181: Do these show appropriate tooltips when disabled due to being on hold? Do we need to add a new string for these?

JamesBurnside (changes requested):

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
JamesBurnside: (note: if we need different strings for tooltips while being on hold that could be tough)

dmceachernmsft: will we want to add it at the selector level for when Contoso makes a component app? doing it here will allow Contoso to still use the selectors and as well give them the choice of what their behavior would be.

I will talk to Alex P about the strings, I think we should be able to set the strings dynamically in a similar fashion to the announcer component and how we have used that before.

JamesBurnside: Hmm good question for parking lot, the reason I argue for it is the composite should have less internal logic of its own and just display whatever the state gives us, but you're right that it might be hard for contoso to understand why a button is disabled. Otherwise perhaps this should live in a composite-level selector

anjulgarg: Do we have an existing component being hidden by selectors?

dmceachernmsft: I do like the idea of it being in a composite level selector, get the best of both worlds. Is there a situation we already do this?

JamesBurnside: > Do we have an existing component being disabled by selectors?
Yeah the camera and mic buttons depending on certain state from the calling sdk:
image

> I do like the idea of it being in a composite level selector, get the best of both worlds. Is there a situation we already do this?
Yeah, check out the selectors folder in the the individual composites dir and there should be some examples of applying composite-level selectors

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
JamesBurnside: Discussed offline, just a nice-to-have, can revisit if necessary

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -134,3 +134,36 @@ export const getCallCompositePage = (
...
+  callControlOptions: CallControlOptions | boolean | undefined,
+  disabledControls: (keyof CallControlOptions)[]
+): CallControlOptions | boolean | undefined => {
+  if (callControlOptions === false) {
+    return false;
+  }
+  // Ensure we clone the prop if it is an object to ensure we do not mutate the original prop.
+  let newOptions = (callControlOptions instanceof Object ? { ...callControlOptions } : callControlOptions) ?? {};
+  if (newOptions === true || newOptions === undefined) {
+    newOptions = disabledControls.reduce((acc, key) => {
```
JamesBurnside, line 157: wow nice

JamesBurnside (approved):

anjulgarg (approved):

carocao-msft (approved):

dmceachernmsft:
image
we have some interesting snapshots here

anjulgarg (changes requested):
UI Snapshots indicate we have introduced a regression

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
edwardlee-msft: wow nice!

anjulgarg (approved):

PR #2216, opened 2022-08-12
Integrate the rooms api into the nodejs server 

mgamis-msft:
# What

- Created routes for `createRoom` and `addUserToRoom` in node server
- Changed the url used in the fetch calls in Calling sample to point to the new routes on the node server

# Why

So that the Rooms API will be deployed with the Calling sample.

# How Tested

Local sample testing

# Process & policy checklist


samples/Calling/src/app/utils/AppUtils.ts
```diff
@@ -41,20 +41,12 @@ export const getGroupIdFromUrl = (): GroupLocator | undefined => {
 
 export const createGroupId = (): GroupLocator => ({ groupId: generateGUID() });
 
-/* @conditional-compile-remove(rooms) */
-export const ROOMS_API_BASE_URL = 'http://localhost:7071';
-
 /* @conditional-compile-remove(rooms) */
 export const createRoomId = async (): Promise<string> => {
```
alkwa-msft, line 45: I think we should just rename the function to createRoom and then it just turns out we return an id to represent it has been created.

samples/Calling/src/app/utils/AppUtils.ts
```diff
@@ -41,20 +41,12 @@ export const getGroupIdFromUrl = (): GroupLocator | undefined => {
...
-    headers: {
-      Host: 'alkwa-fn-test.azurewebsites.net',
-      'Content-Type': 'application/json',
-      'Referrer-Policy': 'no-referrer'
-    }
+    method: 'POST'
   };
-  const response = await fetch(ROOMS_API_BASE_URL + '/api/Rooms-CreateRoom', requestOptions);
+  const response = await fetch(`/createRoom`, requestOptions);
   if (!response.ok) {
```
alkwa-msft, line 50: invalid token response doesn't seem appropriate for creating a room

"Unable to create room" might be better

samples/Server/src/routes/createRoom.ts
```diff
@@ -0,0 +1,45 @@
...
+  const user = await createUser();
+
+  // Options payload to create a room
+  const createRoomOptions: CreateRoomOptions = {
+    validFrom: validFrom,
+    validUntil: validUntil,
+    participants: [
+      {
+        id: user,
+        role: 'Attendee'
```
alkwa-msft, line 35: can we pass the role up as a post  body parameter so we can set our role correctly and not assume everyone is an attendee? I think we can definitely default to attendee if no role is provided.

samples/Server/src/routes/createRoom.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @returns The new roomId as string
+ */
+
+router.post('/', async function (req, res, next) {
+  // create RoomsClient
+  const roomsClient: RoomsClient = new RoomsClient(getResourceConnectionString());
+
+  const validFrom = new Date();
+  const validUntil = new Date(validFrom.getTime() + 5 * 60 * 1000);
```
alkwa-msft, line 24: can we have a comment for what this time length is representing? I am guessing it means 

the room is available from now till now + 5 hrs?

samples/Calling/src/app/utils/AppUtils.ts
mgamis-msft: Renamed.

samples/Calling/src/app/utils/AppUtils.ts
mgamis-msft: Changed. Good call.

samples/Server/src/routes/createRoom.ts
mgamis-msft: Actually we don't need to pass in participants to create a room. Removed the participant option. Thanks.

alkwa-msft (approved):

samples/Calling/src/app/utils/AppUtils.ts
```diff
@@ -42,21 +42,16 @@ export const getGroupIdFromUrl = (): GroupLocator | undefined => {
 export const createGroupId = (): GroupLocator => ({ groupId: generateGUID() });
 
 /* @conditional-compile-remove(rooms) */
-export const ROOMS_API_BASE_URL = 'http://localhost:7071';
```
PorterNan, line 45: Nice, no localhost anymore

PorterNan (approved):

samples/Calling/src/app/utils/AppUtils.ts
```diff
@@ -42,21 +42,16 @@ export const getGroupIdFromUrl = (): GroupLocator | undefined => {
...
-      'Content-Type': 'application/json',
-      'Referrer-Policy': 'no-referrer'
-    }
+    method: 'POST'
   };
-  const response = await fetch(ROOMS_API_BASE_URL + '/api/Rooms-CreateRoom', requestOptions);
+  const response = await fetch(`/createRoom`, requestOptions);
   if (!response.ok) {
-    throw 'Invalid token response';
+    throw 'Unable to create room';
```
JamesBurnside, line 54: note: always use `throw new Error(msg)` instead of just `throw msg`: https://stackoverflow.com/a/46295341/3047190

samples/Server/src/routes/createRoom.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const TWENTY_FOUR_HOURS = 24 * 60 * 60 * 1000;
+/**
+ * route: /createRoom/
+ *
+ * purpose: Create a new room.
+ *
+ * @returns The new roomId as string
+ */
+
```
JamesBurnside, line 19: remove newline here

samples/Server/src/routes/createRoom.ts
```diff
@@ -0,0 +1,38 @@
...
+router.post('/', async function (req, res, next) {
+  // create RoomsClient
+  const roomsClient: RoomsClient = new RoomsClient(getResourceConnectionString());
+
+  const validFrom = new Date();
+  const validUntil = new Date(validFrom.getTime() + TWENTY_FOUR_HOURS);
```
JamesBurnside, line 25: Add a comment in the function doc comment why this is valid specifically for 24 hours

samples/Server/src/routes/createRoom.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  // Options payload to create a room
+  const createRoomOptions: CreateRoomOptions = {
+    validFrom: validFrom,
+    validUntil: validUntil
+  };
+
+  // create a room with the request payload
+  const createRoom = await roomsClient.createRoom(createRoomOptions);
```
JamesBurnside, line 34: nit: variable name here is slightly off, createRoom would imply a verb and that this creates a room, instead consider `createdRoom`

samples/Server/src/routes/addUserToRoom.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * purpose: Add the user to the room with given roomId and role.
+ *
+ * @param userId: id of the user as string
+ * @param roomId: roomId to add user as string
+ * @param role: role of user as string
+ */
+
+router.post('/', async function (req, res, next) {
+  // create RoomsClient
```
JamesBurnside, line 26: nit: this comment doesn't add value as it is just stating the code below it (can remove it)

samples/Server/src/routes/addUserToRoom.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  // request payload to add participants
+  const addParticipantsList: RoomParticipant[] = [
+    {
+      id: { communicationUserId: addUserToRoomParam.userId },
+      role: addUserToRoomParam.role as Role
+    }
+  ];
+
+  const addParticipants = await roomsClient.addParticipants(addUserToRoomParam.roomId, addParticipantsList);
```
JamesBurnside, line 39: similar to my other comment on `createRoom`, `addParticipants` is what has happened, but not what this variable is

JamesBurnside: Actually what does this return?

mgamis-msft: It returns void. No need a variable and just sending a 201. Thanks.

PR #2218, opened 2022-08-12
Role based device permissions request in CallComposite

mgamis-msft:
# What

Porting over code from feature branch to ask device permissions based on role for a Rooms call.
Temporarily using a simplified copy of MockCallAdapter for enzyme test. Will move MockCallAdapter from hermetic tests to fake-backends package in this bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2951482 so that it can be used in the enzyme test.

# Why

To have public Calling sample ready to be deployed for testing

# How Tested

Tested in local Calling sample

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -215,12 +215,17 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
   } = props;
   useEffect(() => {
     (async () => {
-      await adapter.askDevicePermission({ video: true, audio: true });
-      adapter.queryCameras();
-      adapter.queryMicrophones();
-      adapter.querySpeakers();
+      if (role === 'Consumer') {
```
alkwa-msft, line 218: when we stop explicitly passing in our role. How are we going to know our role is consumer so we only ask for audio permissions?

mgamis-msft: I hope I understand the question correctly. The CallComposite can only know the role is Consumer when it is passed as a prop because that role will be used for the PermissionProvider.

PorterNan, line 218: I guess we need to conditionally compile this part?

mgamis-msft: Good catch. Yes we do!

prabhjot-msft (approved):

PorterNan (approved):

PR #2222, opened 2022-08-15
Update Hermetic docs

dmceachernmsft:
# What

Updates Hermetic testing docs
# Why

Adds a section on how to properly run `test.only()` in the hermetic tests.

# How Tested

Doc Change


packages/react-composites/tests/README.md
```diff
@@ -145,3 +145,6 @@ To debug a particular test,
   ```
 * Run the relevant test suite, usually selecting a single project: `node ./scripts/runBrowserTests.mjs -c chat -p desktop -d`
 * Single-step through the test, record a video etc through the Inspector.
+
+### Running just specific hermetic tests
+When working on a specific hermetic test be sure to use `node scripts/runBrowserTests.mjs -l`. This should be used because all of the live tests will be run after even though you are using `test.only(...)`. This happens because the wrapper script calls playwright multiple times depending on which flags are given.
```
prprabhu-ms, line 150: Unfortunately, you also need to pick the composite for the same reason:

So `runBrowserTests.mjs -l -c chat` or `runBrowserTests.mjs -l -c call` etc. :/

prprabhu-ms (approved):

prabhjot-msft (approved):

PR #2223, opened 2022-08-15
Hermetic tests for call screen for different roles

prabhjot-msft:
# What
Hermetic tests for call screen for different roles

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2945144

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
```
JamesBurnside, line 1: Question looking at the snapshots - are we (or are we planning to) update the tooltips to say why certain controls are disabled?
image

e.g. say something like `You do not have permission to start your camera in this call`

JamesBurnside: If we do, we should add snapshot tests that capture the tooltip

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT license.
+
+import { buildUrlWithMockAdapter, defaultMockCallAdapterState, test } from './fixture';
+import { expect } from '@playwright/test';
+import { dataUiId, stableScreenshot, waitForSelector } from '../../common/utils';
+import { IDS } from '../../common/constants';
+
+test.describe('Rooms CallScreen tests for different roles', async () => {
+  test('All CallControls are enabled for Presenter', async ({ page, serverUrl }) => {
+    const initialState = defaultMockCallAdapterState();
```
dmceachernmsft, line 11: should we be adding any remote participants in these tests? what does this look like if you are a consumer and one of the remote participants is the organizer for example?

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
prabhjot-msft: We had the discussion around that and we are not going to have specific tooltips for rooms as of now. If we add the tooltips in future, will add the tests for same :)

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
prabhjot-msft: I have added the remote participants for the tests for consumer, but we cannot control the role for remote participant as it is passed as a prop to call composite as of now. Let me know if you have any thoughts on that.

dmceachernmsft (approved):

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    await page.goto(buildUrlWithMockAdapter(serverUrl, initialState, { role: 'Presenter' }));
+    await waitForSelector(page, dataUiId(IDS.videoGallery));
+    expect(await stableScreenshot(page, { dismissTooltips: true })).toMatchSnapshot('rooms-call-screen-presenter.png');
+  });
+
+  test('All CallControls are enabled for Attendee', async ({ page, serverUrl }) => {
```
JamesBurnside, line 17: Screenshare should not be enabled for an attendee (correct in the snapshots but this test text is incorrect)

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    await page.goto(buildUrlWithMockAdapter(serverUrl, initialState, { role: 'Attendee' }));
+    await waitForSelector(page, dataUiId(IDS.videoGallery));
+    expect(await stableScreenshot(page, { dismissTooltips: true })).toMatchSnapshot('rooms-call-screen-attendee.png');
+  });
+
+  test('Only few CallControls are enabled for Consumer', async ({ page, serverUrl }) => {
+    const initialState = defaultMockCallAdapterState();
+    await page.goto(buildUrlWithMockAdapter(serverUrl, initialState, { role: 'Consumer' }));
+    await waitForSelector(page, dataUiId(IDS.videoGallery));
+    expect(await stableScreenshot(page, { dismissTooltips: true })).toMatchSnapshot('rooms-call-screen-consumer.png');
```
JamesBurnside, line 28: I know this is just a hermetic fake-data artifact, but is odd that the participant is stuck in the unmuted state:
image

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  test('Only few CallControls are enabled for Consumer with remote participants', async ({ page, serverUrl }) => {
```
JamesBurnside, line 31: Is there value to having two tests, one with remote participants and one without?

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
JamesBurnside: (I don't think we can do anything about this here, the call state should just always be muted when in this role)

JamesBurnside (approved):

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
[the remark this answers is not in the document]
prabhjot-msft: Yeah, I am writing tests for that. There will be tests covering remove button as well checking device lists.

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
prabhjot-msft: Yeah, we are using same default state in all mock tests and user is unmuted by default in mock state, that is not the case in actual call though :)

packages/react-composites/tests/browser/call/hermetic/RoomsCallScreen.test.ts
prabhjot-msft: Not really, added this test later will only keep the one with remote participants, thanks for noticing

carocao-msft (approved):

PR #2224, opened 2022-08-16
Add landmarks to chat

carocao-msft:
# What
Add landmarks to chat

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2908209

# How Tested
Screen Shot 2022-08-15 at 10 31 04 PM


carocao-msft:
> I think we have two options to fix this bug:
> 
> 1. (easier) we only apply landmarks to the sample app (only code inside `samples/`)
> 2. (better but this would be a feature request) we expose the ability to set roles through our composite API
> 
> I would stick with 1 for now if it sufficiently solves the bug

Updated to only modify the sample :)

JamesBurnside (approved):

edwardlee-msft (approved):

PR #2225, opened 2022-08-16
Removed participant list button permissions and added enzyme tests for room

prabhjot-msft:
# What
Removed participant list button permissions and added enzyme tests for room

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/permissions/PermissionsProvider.test.tsx
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import React from 'react';
```
mgamis-msft, line 4: I think you need to add a cond-comp on every line on this file. The stable build CI is getting this error.
image
@PorterNan is there a way to easy cond-comp a whole file?

prabhjot-msft: done

PorterNan: @mgamis-msft, there was, but we removed that feature for some concerns, but maybe we can do this for test only case - because for test there are no entry point, each test.ts is an entry

PorterNan: Lemme open a bug for this case, but I don't think this is something we can resolve in this PR

mgamis-msft: Sorry, my mistake. You can remove all cond-comp comments in each line in PermisionProvider.test.tsx. It is okay for PermissionProvider to be part of stable build but just wont be exposed. Also doing this resolves this stable build 
image

packages/react-components/src/components/ParticipantList.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+};
+
+describe('ParticipantList tests for different roles', () => {
+  beforeEach(() => {
+    registerIcons({
+      icons: {
+        participantitemoptions: <></>
+      }
+    });
+  });
```
mgamis-msft, line 34: To resolve this stable build error:
image
you can just add a test that tests that ParticipantList should have remove item (with no PermissionProvider as a wrapping component). 
```
test('ParticipantList should have remove item', async () => {
    mount( <ParticipantList
        participants={[{ displayName: 'User1', userId: '1', isRemovable: true }]}
        onRemoveParticipant={dummyOnRemoveParticipantCallback}
      />);
      ...
```

alkwa-msft (approved):

packages/react-composites/src/composites/CallComposite/components/buttons/People.tsx
```diff
@@ -50,8 +43,6 @@ export const People = (props: ControlBarButtonProps): JSX.Element => {
       onRenderOffIcon={onRenderOffIcon ?? icon}
       onClick={onClick}
       styles={styles}
-      /* @conditional-compile-remove(rooms) */
-      disabled={disabled}
```
PorterNan, line 54: Do we have a chance to test this, can consumer show the partcipant list without problem?

PorterNan (approved):
LGTM, just need to double check if consumer role can successfully show all the participant wihtout problem when they click the people button

packages/react-composites/src/composites/CallComposite/components/buttons/People.tsx
prabhjot-msft: Consumer can see the participants now

change/@internal-react-components-1aaac0fc-16c0-4dcb-bfab-cbb1d469c733.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
prprabhu-ms, line 2: This should have been `prerelease` as rooms is a beta-only feature.

change/@internal-react-composites-95bfe8dd-8766-4464-be3b-004be9d1ea17.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
prprabhu-ms, line 2: This should have been `prerelease` as rooms is a beta-only feature.

PR #2227, opened 2022-08-17
Add DTMF dialpad to call and callwithchat more button 

carocao-msft:
# What
Add DTMF dialpad to call and callwithchat more button 

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2919258

# How Tested
Screen Shot 2022-08-17 at 12 21 49 AM

Screen Shot 2022-08-17 at 12 21 43 AM

Screen Shot 2022-08-17 at 12 17 08 AM


packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -268,4 +268,14 @@ export interface CallCompositeStrings {
...
+  /* @conditional-compile-remove(PSTN-calls) */
+  /**
+   * Placeholder text for dtmf dialpad
+   */
+  dtmfDialpadPlaceHolderText: string;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /**
+   * Label for the button to open dtmf dialpad
+   */
+  openDtmfDialpad: string;
```
JamesBurnside, line 280: Nomenclature in this file is to suffix with `Label` for button labels

JamesBurnside (approved):
(not in this work item) We should explore what the API looks like if contoso doesn't want this button - can it be hidden through the callControls Composite Options?

anjulgarg (approved):

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/Strings.tsx
carocao-msft: Sorry missed this comment! Will change this in the next PR

PR #2228, opened 2022-08-17
Change Calling sample room creation to an open policy

mgamis-msft:
# What

Change roomJoinPolicy of room when starting a new room in Calling sample.

# Why

Solve issue of not being able to join room calls that have already started

# How Tested

Local calling sample

# Process & policy checklist

### DIFF
--- a/samples/Calling/src/app/App.tsx
+++ b/samples/Calling/src/app/App.tsx
@@ -23,7 +23,7 @@ import {
   WEB_APP_TITLE
 } from './utils/AppUtils';
 /* @conditional-compile-remove(rooms) */
-import { createRoom, getRoomIdFromUrl, addUserToRoom } from './utils/AppUtils';
+import { createRoom, getRoomIdFromUrl } from './utils/AppUtils';
 import { useIsMobile } from './utils/useIsMobile';
 import { useSecondaryInstanceCheck } from './utils/useSecondaryInstanceCheck';
 import { CallError } from './views/CallError';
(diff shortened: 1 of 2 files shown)


samples/Calling/src/app/App.tsx
```diff
@@ -134,12 +138,10 @@ const App = (): JSX.Element => {
...
-              if (userId) {
-                setRole(callDetails.role as Role);
-                await addUserToRoom(userId.communicationUserId, callLocator.roomId, callDetails.role as Role);
-              } else {
-                throw 'Invalid userId!';
-              }
+              setRole(callDetails.role as Role);
+            } else {
+              // unset role when call is not a Rooms call
+              setRole(undefined);
```
alkwa-msft, line 144: We can be silly and just set this as Presenter instead and just add a comment that a Presenter is the same as if you hadn't set a role.

alkwa-msft (approved):

PorterNan (commented):
Is there a documentation about open room policy?
Sounds like open room policy doesn't require participant to join the room - so how do you determine your role from service side - this looks a pure UI side behavior after your change

PorterNan (commented):
The worry I have here is this workflow looks like not align with the RBAC workflow, it looks like just a pure room without permissions limitations - we just add some UI things up to it.
Once we need to get role from calling sdk this sample app would not work

prabhjot-msft (approved):

mgamis-msft:
> The worry I have here is this workflow looks like not align with the RBAC workflow, it looks like just a pure room without permissions limitations - we just add some UI things up to it. Once we need to get role from calling sdk this sample app would not work

That is a good point. RBAC no longer exists for an open room policy. Closing this PR without merging. Will revisit in the future.

PR #2232, opened 2022-08-17
Update _isInCall for PSTN and 1:N scenarios

edwardlee-msft:
# What
Update _isInCall for PSTN and 1:N scenarios and include it's usage in Call Composite in place of checking "callStatus === 'connected'".

# Why
Scalable solution for when new CallStatus's are added.
RemoteHold is technically still a connected a call, this should also be accounted for when rendering the Call Composite.

# How Tested
Test locally on MacOS on chrome browser

# Process & policy checklist


packages/calling-component-bindings/src/utils/callUtils.ts
```diff
@@ -11,8 +11,16 @@ import { memoizeFnAll, toFlatCommunicationIdentifier } from '@internal/acs-ui-co
...
-export const _isInCall = (callStatus?: CallStatus): boolean =>
-  !!callStatus && !['None', 'Disconnected', 'Connecting', 'LocalHold', 'RemoteHold', 'Ringing'].includes(callStatus);
+export const _isInCall = (callStatus?: CallStatus): boolean => {
+  /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
+  return (
+    !!callStatus && !['None', 'Disconnected', 'Connecting', 'LocalHold', 'Ringing', 'EarlyMedia'].includes(callStatus)
+  );
+  return (
+    !!callStatus &&
+    !['None', 'Disconnected', 'Connecting', 'LocalHold', 'RemoteHold', 'Ringing'].includes(callStatus as string)
```
JamesBurnside, line 21: intentional that now `EarlyMedia` will mean isInCall is true? ( don't have a good grasp on what early media means)

dmceachernmsft: so the hold states and ringing shouldn't be part of the stable version of this function since they were originally not there. 

Isn't the logic of this function @JamesBurnside that if the state is in the array the user is not in the call? just the negation of that check?

JamesBurnside: Oop thanks, I read the conditional compile the wrong way around!

JamesBurnside: I would probably wouldn't even conditionally compile this and just update the isInCall - looks like its simply fixing a bug where we aren't checking the states correctly!

JamesBurnside (approved):

packages/calling-component-bindings/src/utils/callUtils.ts
edwardlee-msft: Hi @anjulgarg and @dmceachernmsft , 
Can you please clarify on EarlyMedia details?

anjulgarg: EarlyMedia will only occur during PSTN calls.
It basically means that media is being received by the caller.
For example, you call a number and their phone starts ringing.
Media starts flowing from their device to yours telling you that their phone is ringing.
That's EarlyMedia. You can think of it as a 'Ringing' state.

anjulgarg (approved):

PR #2235, opened 2022-08-18
Hide Dialpad when it's not 1:1 PSTN call

carocao-msft:
# What
Hide Dialpad when it's not 1:1 PSTN call

# Why
dtmf tones does not work for non 1:1 call

# How Tested

Hermantic tests for non 1:1 call , non PSTN call and 1:1 PSTN call

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
+++ b/packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
@@ -35,6 +35,7 @@ import { usePropsFor } from '../hooks/usePropsFor';
 import { buttonFlyoutIncreasedSizeStyles } from '../styles/Buttons.styles';
 /* @conditional-compile-remove(PSTN-calls) */
 import { SendDtmfDialpad } from '../../common/SendDtmfDialpad';
+/* @conditional-compile-remove(PSTN-calls) */
 import { useAdapter } from '../adapter/CallAdapterProvider';
 
 /**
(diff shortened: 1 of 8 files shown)


packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -97,6 +101,11 @@ export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX
   /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
   const holdButtonProps = usePropsFor(HoldButton);
 
+  /* @conditional-compile-remove(PSTN-calls) */
+  const alternateCallerId = useAdapter().getState().alternateCallerId;
+  /* @conditional-compile-remove(PSTN-calls) */
+  const participantNumber = usePropsFor(ParticipantList).participants.length;
```
anjulgarg, line 107: ambiguous attribute name `participantNumber`
I read it and think this is a participants phone number.
Something like `participantCount` would make more sense here.

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -130,17 +139,20 @@ export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX
...
-      onClick: () => {
-        setShowDialpad(true);
-      },
-      iconProps: { iconName: 'Dialpad', styles: { root: { lineHeight: 0 } } },
-      itemProps: {
-        styles: buttonFlyoutIncreasedSizeStyles
-      }
-    });
+    // dtmf tone sending only works for 1:1 PSTN call
+    if (alternateCallerId && participantNumber <= 2) {
```
anjulgarg, line 143: This is incorrect. Even if the participant length is 1, we need to ensure that the participant is a PSTN number.
One could provide an alternate caller id to the composite and still call an acs user id.

Plus, I recommend using a function outside the component to do this calculation instead of doing it inline.
Something like `canShowDialpad()`

carocao-msft:
Decided to show dialpad for all calls

PR #2236, opened 2022-08-18
Video Tile using pseudoelement inset border instead of floating div with border

mgamis-msft:
# What

Using pseudoelement border on VideoTile instead of using another floating div with border. Pseudoelements are "fake" elements made by CSS which is created within any HTML tag but does not affect its spacing. They also allow you to set a zIndex so that children of the HTML tag don't cover the border. Source: https://fjolt.com/article/css-easy-inset-borders#option-3:-pseudo-elements

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2931746
Note: Could not actually reproduce this bug on my Surface, my Mac, or my IPad.

# How Tested

Tested on local storybook:
https://user-images.githubusercontent.com/79475487/185558856-b8f6c771-9fa6-40af-853e-7fe8c66e0b06.mp4
(first participant of VideoGallery story in the video is hardcoded to be isSpeaking to show what it will look like)

# Process & policy checklist


packages/react-components/src/components/VideoTile.tsx
```diff
@@ -240,16 +239,15 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
             background: theme.palette.neutralLighter,
             borderRadius: theme.effects.roundedCorner4
           },
+          isSpeaking
+            ? {
+                outlineOffset: '-0.25rem',
```
JamesBurnside, line 244: didn't know about this property - nice one!

JamesBurnside (commented):
This looks like a brilliant fix! Especially when we can't repro the problem!

From the snapshot failure looks like we can no longer see the isSpeaking in the snapshot tests:
image

Note, this is likely be because we overlay the video with a green block to prevent flakey video artifacts (in which case the green block would need to be 0.25rem smaller in each direction and margin of 0.25rem), here is where we do that: https://github.com/Azure/communication-ui-library/blob/655e457b8260d492ccffd927f25a16e59e95ea0f/packages/react-composites/tests/browser/common/utils.ts#L523

packages/react-components/src/components/VideoTile.tsx
anjulgarg: I didn't even know this was possible

anjulgarg (approved):

anjulgarg (changes requested):
Missing the `ui change` label

mgamis-msft:
> mask.setAttribute('style', 'position: absolute; height: 100%; width: 100%; background: green;');

Darn not such a good solution after all. Children will block the a parent's outline since outlines it do not take up space. So when the video is on the outline will not show.
https://user-images.githubusercontent.com/79475487/185454322-dab609ec-17df-4dbf-b910-f6569e927f54.mp4

mgamis-msft:
> Missing the `ui change` label

I meant to use the UI automation to verify that no visual changes occur with this PR.

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -240,16 +239,20 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
             background: theme.palette.neutralLighter,
             borderRadius: theme.effects.roundedCorner4
           },
+          isSpeaking && {
+            '&::before': {
```
JamesBurnside, line 243: wow even cooler magic! How does this work?

JamesBurnside (approved):

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -240,16 +239,20 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+              content: `''`,
+              position: 'absolute',
+              zIndex: 1,
```
dmceachernmsft, line 246: should we make this a constant somewhere so we know why the z-index is being set to 1?

dmceachernmsft (approved):

anjulgarg (approved):

packages/react-components/src/components/VideoTile.tsx
mgamis-msft: Oops. I missed this comment. But I think it should be okay with a constant since the value is 1. If it was a number like 999 then I would.

PR #2239, opened 2022-08-19
Fix custom menu items being triggered twice on call composite (and not at all on call with chat composite) 

JamesBurnside:
# What
* Ensure onFetchParticipantMenuItems is being passed down as a prop in the call with chat composite
* Ensure on mobile we only trigger the onFetchParticipantMenuItems for drawer menu items and not the ordinary menu

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2949057
Plus found a latent bug that they do not show at all on call with chat composite currently! (Made debugging very fun)
And found a bug where we are always trying to render the callcomposite's participant pane but that will be a separate PR (#2240) (made debugging even *more* fun.. and by fun I mean pain, just pain)

# How Tested
Works locally, snapshots should also be updated:
image


JamesBurnside:
UI test checks need updated with fix - closing PR until they are complete, will reopen once fixed

edwardlee-msft (approved):

prprabhu-ms (approved):
<3

PR #2240, opened 2022-08-19
Fix: Prevent running render passes on CallComposite people pane on every render pass (even when not shown)

JamesBurnside:
# What
* **MAIN CHANGE HERE TO REVIEW:** Added `activePane === 'people'` check
* Also added some useCallbacks to functions

# Why
We were running render passes it over every render, made debugging #2239 incredibly confusing

# How Tested
`react-render-tracker` no longer shows rendering updates for the people pane correctly


packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -197,34 +197,47 @@ export const ParticipantList = (props: ParticipantListProps): JSX.Element => {
...
+          disabled: disabled,
+          'data-ui-id': ids.participantListRemoveParticipantButton
+        });
+      }
 
-    if (onFetchParticipantMenuItems) {
-      menuItems = onFetchParticipantMenuItems(participant.userId, myUserId, menuItems);
-    }
+      if (onFetchParticipantMenuItems) {
+        menuItems = [...onFetchParticipantMenuItems(participant.userId, myUserId, menuItems)];
```
prabhjot-msft, line 222: Why did we enforce the type of menu items to array, is it not array already?

JamesBurnside: Oops great catch! That was from me testing, I will revert this line!

JamesBurnside: reverted

edwardlee-msft (approved):

prabhjot-msft (approved):

PR #2241, opened 2022-08-19
Add react render tracker to samples

JamesBurnside:
# What
Add [react render tracker](https://github.com/lahmatiy/react-render-tracker) to samples.

# Why
Ensure this is set up for everyone to use.
It just saved my life fixing https://skype.visualstudio.com/SPOOL/_workitems/edit/2785576. I had to get everything spun up locally, plus I put off doing this for like an hour because I had to find the cdn script to inject and thought I could fix it without it. Having this by default locally, and in our deployed samples, would be incredibly useful.

# How Tested
n/a


samples/CallWithChat/public/index.html
```diff
@@ -8,6 +8,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="description" content="Azure Communication Services - UI Library: Call With Chat Sample app" />
     <title>UI Library Call With Chat Sample</title>
+
+    <!-- For debugging purposes only. Remove for production builds. -->
```
PorterNan, line 12: Is there a way to only inject this using webpack plugin - when it is production build, like --mode=production --env production, we remove this

PorterNan (approved):

samples/CallWithChat/public/index.html
JamesBurnside: Technically would be possible, but I didn't walk down that road as I really want to have this in our deployed samples (and its probably relatively complicated). This comment is more for if Contoso copies this file they have a hit to remove this line

anjulgarg (approved):

prprabhu-ms:
I worry about the memory implications of this PR for the samples.

react render tracker keeps a (somewhat filtered) history of all React component updates. As the app runs continuously, I'd expect the memory requirement of this tracking to balloon up.

That is why I had thought we should only enable react render tracker when actually debugging.

JamesBurnside:
> I worry about the memory implications of this PR for the samples.
> 
> react render tracker keeps a (somewhat filtered) history of all React component updates. As the app runs continuously, I'd expect the memory requirement of this tracking to balloon up.
> 
> That is why I had thought we should only enable react render tracker when actually debugging.

Very fair concerns - my (very biased) vote is to not optimize early but instead, given this could be an issue, we should keep an eye on memory usage when running the samples for a long time. Having this in allows us instant debugging when we hit an issue instead of having to spin up locally with these changes, or restart a local instance when we hit a repro (useful in beta validation builds, quickly debugging an issue without opening the codebase, or when we hit a repro that scarcely happens - and just lowers the barrier to getting this debugging tool set up and working)

PR #2249, opened 2022-08-22
register icons in ui-components snippet

dmceachernmsft:
# What

Register icons
# Why

Icons on completed snippet doesn't load without registering icons

# How Tested

Ran snippet locally


edwardlee-msft (approved):

packages/storybook/stories/QuickStarts/snippets/CompletedComponentsApp.snippet.tsx
```diff
@@ -1,10 +1,13 @@
...
+import { DEFAULT_COMPONENT_ICONS, FluentThemeProvider } from '@azure/communication-react';
+import { Stack, registerIcons } from '@fluentui/react';
 import React from 'react';
 import { CallingComponents } from './CallingComponents';
 import { ChatComponents } from './ChatComponents';
 
 function CompletedComponentsApp(): JSX.Element {
+  // If you don't want to provide custom icons, you can register the default ones included with the library.
+  // This will ensure that all the icons are rendered correctly.
+  registerIcons({ icons: DEFAULT_COMPONENT_ICONS });
```
prprabhu-ms, line 10: tl;dr -> `registerIcon` should be called only once. There are two options: (1) call it outside the Component scope, at import time (on line 6), or (2) call it inside a `useEffect` block with `[]` dependency list.

Can you try to run this snippet with an added counter that introduces a `useState` in the `CompletedComponentsApp`? My understanding is that when the app rerenders due to the state change, it will try to register icons again, and get a warning on console from fluent about duplicate icon registration.

This PR is consistent with the other snippets in storybook, so if I'm right we'll need to update snippets across the board. But the `samples/*` apps all use strategy (1) stated above, so we should make the snippets and samples consistent regardless.

dmceachernmsft: Upon testing this you were correct, I will make this PR a little larger by moving that statement in our snippets. there shouldnt be too many since its the component snippets only I think

prabhjot-msft (approved):

PR #2251, opened 2022-08-23
Fix disabled start call button for consumer

mgamis-msft:
# What

Fixed logic to disabling start call button. If role permissions do not give access to microphone button then do *NOT* disable start call button.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2960638

# How Tested

Local calling sample

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -52,6 +54,13 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
   const errorBarProps = usePropsFor(ErrorBar);
   const adapter = useAdapter();
   const deviceState = adapter.getState().devices;
+
+  let disableStartCallButton = !microphonePermissionGranted || deviceState.microphones?.length === 0;
+  /* @conditional-compile-remove(rooms) */
+  const permissions = _usePermissions();
+  /* @conditional-compile-remove(rooms) */
```
PorterNan, line 61: this logic needs some time to understand for reader, would be nice to add some comments to explain permissions.microphoneButton will impact the logic of startButton

PorterNan (approved):

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -52,6 +54,17 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
   const errorBarProps = usePropsFor(ErrorBar);
   const adapter = useAdapter();
   const deviceState = adapter.getState().devices;
+
+  let disableStartCallButton = !microphonePermissionGranted || deviceState.microphones?.length === 0;
+  /* @conditional-compile-remove(rooms) */
```
prabhjot-msft, line 59: Should we create trampoline function to minimize use of conditional-compile statements

mgamis-msft: I think even if I use a trampoline function. I would still have 2 conditional-compile statements. 1 for the `usePermissions` hook and 1 for the the if statement . I cannot have the hook in the function :(.

prabhjot-msft (approved):

change/@internal-react-composites-6c4b2b00-8f23-460d-baf9-712e7e7614a6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
prprabhu-ms, line 2: This should have been `prerelease` as rooms is a beta-only feature.

PR #2220, opened 2022-08-15
Dismiss tooltips in new hermetic test

prprabhu-ms:
CONSTANT VIGILANCE!

PR #2252, opened 2022-08-23
Deflake a class of browser test failures

prprabhu-ms:
# What

Dismiss tooltips by default before taking screenshots in browser tests.

# Why

This is the most common source of flakes in (hermetic) tests. It is especially easy to forget to dismiss tooltips in new tests.

Example:
* I wrote a new test that I had to fix immediately: https://github.com/Azure/communication-ui-library/pull/2220
* A newly added test this week is very flaky for this reason: https://github.com/Azure/communication-ui-library/actions/runs/2909685706

# How Tested

Stress test: https://github.com/Azure/communication-ui-library/actions/runs/2916433530


packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -451,7 +455,7 @@ export async function stableScreenshot(
   try {
     return await page.screenshot(screenshotOptions);
   } finally {
-    if (stubOptions?.dismissTooltips) {
+    if (stubOptions?.dismissTooltips !== false) {
```
JamesBurnside, line 458: Should 
```
  if (stubOptions?.dismissTooltips) {
    await disableTooltips(page);
  }
```

Also have the updated `!== false` check

prprabhu-ms: Eh right. I didn't actually fix the bug.

dmceachernmsft (approved):

PR #2253, opened 2022-08-23
Fix join teams meeting logic

mgamis-msft:
# What

Detecting the correct key 'meetingLink' to use a Teams meeting locator.
No bug is actually observable because Calling SDK is smart enough to join a teams Call even if a group locator is used with a Teams meeting link

# Why

For correctness.

# How Tested

Tested in local sample

# Process & policy checklist


PorterNan (approved):
It's interesting to see this wasn't caught by any of our e2e test, do we have a snapshot/test for this one for e2e test @prabhjot-msft @mgamis-msft ?

PorterNan:
> It's interesting to see this wasn't caught by any of our e2e test, do we have a snapshot/test for this one for e2e test @prabhjot-msft @mgamis-msft ?

Just read miguel's description, no wonder we can't catch this - sounds like call sdk is kinda too smart to deal with groupIds

dmceachernmsft (approved):

PR #2255, opened 2022-08-24
browser tests: rename common folder to lib

prprabhu-ms:
This PR is simply a folder rename: `packages/react-composites/tests/browser/common` -> `packages/react-composites/tests/browser/lib`

# Why

This folder now contains helpers common to the different composite tests along with composite specific code. It will soon also contain code common to the test apps and test app specific code.

The final folder structure will look so:

packages/react-composite/tests/browser/lib/* -- common logic used by all tests
  - app/
    - common/* -- common code for test apps
    - call/* -- `CallComposite` related code used across test apps (e.g. by both the call and callwithchat test app)
    - chat/* -- `ChatComposite` related code ...
  - call/*  -- common logic used by calling related tests etc.


prprabhu-ms:
Decided it was better to merge the test apps together and then move the common utility code for the test applications to that merged app folder.

PR #2258, opened 2022-08-24
Actually typecheck browser test code

prprabhu-ms:
* Fix typo in `packages/react-composites/tests/tsconfig.json` that resulted in the test code never getting type checked.
* Fix uncovered type errors

In particular, fix the problems caused by `@internal/fake-backends` packlet not being setup properly:

* Drop source maps in `react-composites/tests/tsconfig.json` -- source maps don't work with stable flavor builds (the only other `tsconfig.json` that has source maps is `storybook`, but that is never built in stable flavor)
* Fix the build setup of `@internal/fake-backends` so that it correctly generates the rolled up types file for both beta and stable flavor builds.

### DIFF
--- a/packages/react-composites/tests/browser/chat/app/CustomDataModel.tsx
+++ b/packages/react-composites/tests/browser/chat/app/CustomDataModel.tsx
@@ -1,16 +1,15 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { ChatParticipant } from '@azure/communication-signaling';
 import { Stack } from '@fluentui/react';
-import { MessageProps } from '@internal/react-components';
+import { CommunicationParticipant, MessageProps } from '@internal/react-components';
 import React from 'react';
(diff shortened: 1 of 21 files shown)


prprabhu-ms:
Apologies that this PR is a bit diffused. It could be at least two PRs -- one that only fixes the `fake-backends` packlet and one for the rest. But it was a can of worms that kept spewing out gunk...

packages/react-composites/tests/browser/chat/app/CustomDataModel.tsx
```diff
@@ -54,6 +53,7 @@ const getMessageContentInUppercase = (messageProps: MessageProps): string => {
     default:
       'CUSTOM MESSAGE';
   }
+  throw new Error('unrechange code');
```
JamesBurnside, line 56: `unreachable`?

JamesBurnside (approved):
💪

mgamis-msft (approved):

PR #2259, opened 2022-08-24
Extract `CallComposite` browser test app to external folder

prprabhu-ms:
This is the first step towards merging the test applications used by the `CallComposite`, `ChatComposite` and `CallWithChatComposite` tests.

These test applications already share some logic and will share a lot more as `CallWithChatComposite` test app gains the ability to run hermetically, so it makes sense to merge these apps.

This PR duplicates some helper files for apps (e.g. the `fonts`, `index.tsx` etc.) because they're used from different locations by the migrated app vs to-be-migrated apps. #2263 removes the duplicate files later.


PorterNan:
Nice! One further step we could do is to merge these 3 composite test folders into app/test folder (+ some code sharing)

PorterNan (approved):

JamesBurnside (approved):

PR #2260, opened 2022-08-24
Long press on dialpad 0 button should return +

carocao-msft:
# What
Long press on dialpad 0 button should return +

# Why
This is standard dialpad behaviour 

# How Tested

https://user-images.githubusercontent.com/96077406/186470655-0af22bf1-57ab-4bf8-b3d0-4aa5c2939258.mov


packages/react-components/src/components/utils/useLongPress.ts
```diff
@@ -0,0 +1,55 @@
...
+  const [action, setAction] = useState('');
+  const timerRef = useRef<NodeJS.Timeout>();
+  const [isLongPress, setIsLongPress] = useState(false);
+
+  function startPressTimer() {
+    setIsLongPress(false);
+    timerRef.current = setTimeout(() => {
+      setIsLongPress(true);
+      setAction('longpress');
+    }, 500);
```
dmceachernmsft, line 16: is half a second long enough?

packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -175,9 +139,44 @@ const DtmfTones: DtmfTone[] = [
...
+}): JSX.Element => {
+  const theme = useTheme();
+  const clickFunction = () => {
+    props.onClick(props.primaryContent, props.index);
+  };
+  const longPressFunction = () => {
+    props.onLongPress(props.primaryContent, props.index);
+  };
+
+  const { handlers } = useLongPress(clickFunction, longPressFunction);
```
JamesBurnside, line 158: both clickFunction and longPressFunction should be memoized with useCallback if being passed into a hook

packages/react-components/src/components/utils/useLongPress.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { useRef, useState } from 'react';
+
+export default function useLongPress(onClick: () => void, onLongPress: () => void) {
```
JamesBurnside, line 6: Does this work for keyboard and ally users?

packages/react-components/src/components/utils/useLongPress.ts
carocao-msft: I played around with it, 1 second feels too long. feel free to try it out and let me know what you think :)

packages/react-components/src/components/Dialpad/Dialpad.tsx
carocao-msft: Updated

packages/react-components/src/components/utils/useLongPress.ts
carocao-msft: Updated longpress hook to include keyboard event

anjulgarg (approved):

JamesBurnside:
Trying out the keyboard I found two points of friction:
* It didn't always work, here's a repro video (you can see how long I hold spacebar for to try get it to trigger) sometimes it creates a `+` and sometimes it's creates a `0`:

   https://user-images.githubusercontent.com/2684369/186729170-8dc9db44-a51a-47e4-ac0a-1c495515a4cc.mp4

* Holding enter key doesn't work

carocao-msft:
> Trying out the keyboard change in the PR storybook link I found two points of friction:
> 
> * It didn't always work, here's a repro video (you can see how long I hold spacebar for to try get it to trigger) sometimes it creates a `+` and sometimes it's creates a `0`:
>   
>     
>       
>       
>   
>       keyboard-0-dialpad-repro.mp4
> * Holding enter key doesn't work

Fixed that weird bug! Thanks to stackoverflow :) enter key don't work since holding it will just be continuely inputting 0s instead of being a long press event

dmceachernmsft (approved):

JamesBurnside:
> enter key don't work since holding it will just be continuely inputting 0s instead of being a long press event

Just a note that after the spacebar fix the enter key behavior has changed and now enters a singgle `0` followed by `+`s when held:


https://user-images.githubusercontent.com/2684369/186760344-dc869cc5-53fb-4115-8c36-7854c15fde19.mp4

PR #2264, opened 2022-08-24
Extract common library for browser test apps

prprabhu-ms:
Preparatory step for adding a hermetic app for `CallWithChatComposite` tests.

* Create a separate `.../tests/common` folder to house types that are used by both the test apps and the tests.
  * Before this PR, there were dependencies in both directions between the test app library code and the test library code.
* Extract `MockCallAdapter` to be reused from `CallWithChatComposite` app.
* Extract `useFakeChatAdapters` to be reused from `CallWithChatComposite` app.


packages/react-composites/tests/app/lib/useFakeChatAdapters.ts
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) Microsoft Corporation.
```
JamesBurnside, line 1: thanks for breaking all this out to its own file!

packages/react-composites/tests/app/lib/useFakeChatAdapters.ts
```diff
@@ -0,0 +1,161 @@
...
+
+const initializeAdapters = async (
+  participants: ChatParticipant[],
+  chatClientModel: Model,
+  thread
+): Promise<ChatAdapter[]> => {
+  const remoteAdapters: ChatAdapter[] = [];
+  for (const participant of participants) {
+    if (!participant.displayName) {
+      throw new Error(`All participants must have displayName defined, got ${JSON.stringify(participant)}`);
```
JamesBurnside, line 91: nit: we should use a safe json stringingify to prevent the error not being displayed if JSON stringify throws an error

packages/react-composites/tests/app/lib/useFakeChatAdapters.ts
```diff
@@ -0,0 +1,161 @@
...
+    userId: adapterInfo.userId as CommunicationUserIdentifier,
+    displayName: adapterInfo.displayName,
+    endpoint: 'FAKE_ENDPOINT',
+    credential: fakeToken
+  });
+  statefulChatClient.startRealtimeNotifications();
+  const chatThreadClient: ChatThreadClient = await statefulChatClient.getChatThreadClient(
+    adapterInfo.chatThreadClient.threadId
+  );
+  registerchatThreadClientMethodErrors(chatThreadClient, chatThreadClientMethodErrors);
```
JamesBurnside, line 119: nit: camelCase 
```suggestion
  registerChatThreadClientMethodErrors(chatThreadClient, chatThreadClientMethodErrors);
```

JamesBurnside (approved):

carocao-msft (approved):

packages/react-composites/tests/app/lib/useFakeChatAdapters.ts
prprabhu-ms: While I agree, I don't want to introduce a dependency from `@internal/acs-ui-common` in this PR.

prprabhu-ms: Will try to remember and follow up on top of the stack.

PR #2266, opened 2022-08-24
update holdPane to have resuming state

dmceachernmsft:
# What

Adds resuming state to hold pane
# Why

Stops user from clicking the resume button over and over again

https://skype.visualstudio.com/SPOOL/_workitems/edit/2958030
# How Tested

Ran locally 
image
image

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
+++ b/packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { PrimaryButton, Stack, Text } from '@fluentui/react';
+import { PrimaryButton, Spinner, Stack, Text } from '@fluentui/react';
 import { _pxToRem } from '@internal/acs-ui-common';
 import React, { useRef, useState } from 'react';
 import { CompositeLocale, useLocale } from '../../localization';
@@ -21,6 +21,7 @@ interface HoldPaneStrings {
   holdScreenLabel: string;
   resumeCallButtonLabel: string;
   resumeCallButtonAriaLabel: string;
+  resumingCallButtonLabel: string;
 }
 
 /**
@@ -40,6 +41,7 @@ export const HoldPane = (): JSX.Element => {
   const elapsedTime = getReadableTime(time);
 
   const startTime = useRef(performance.now());
+  const [resumingCall, setResumingCall] = useState<boolean>(false);
 
   React.useEffect(() => {
     const interval = setInterval(() => {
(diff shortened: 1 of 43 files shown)


packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -50,21 +53,29 @@ export const HoldPane = (): JSX.Element => {
     };
   }, [startTime]);
 
+  const showSpinner = (): JSX.Element => {
```
JamesBurnside, line 56: nit: semantically this is the spinner itself as opposed to a function that triggers a spinner to show
```suggestion
  const ResumeSpinner = (): JSX.Element => {
```

then where its used:
```
{resumingCall && <ResumeSpinner  />}
```

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -50,21 +53,29 @@ export const HoldPane = (): JSX.Element => {
...
         <PrimaryButton
-          text={strings.resumeCallButtonLabel}
-          ariaLabel={strings.resumeCallButtonAriaLabel}
+          text={!resumingCall ? strings.resumeCallButtonLabel : undefined}
+          ariaLabel={!resumingCall ? strings.resumeCallButtonAriaLabel : strings.resumingCallButtonAriaLabel}
           styles={resumeButtonStyles}
+          disabled={resumingCall}
           onClick={() => {
             /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
             holdButtonProps.onToggleHold();
```
JamesBurnside, line 72: can this be awaited to catch an error, e.g. it would be great if this could be:
```
try {
  setResumingCall(true);
  await holdButtonProps.hold();
} finally {
  setResumingCall(false);
}
```

The key part there being the finally that resets the state so resume could be clicked again if it fails

anjulgarg (approved):

packages/react-composites/src/composites/CallComposite/components/HoldPane.tsx
```diff
@@ -50,21 +53,34 @@ export const HoldPane = (): JSX.Element => {
...
+              await holdButtonProps.onToggleHold();
+            } catch (e) {
+              setResumingCall(false);
+              throw e;
+            }
           }}
           data-ui-id="hold-page-resume-call-button"
-        />
+        >
+          {resumingCall && resumeSpinner()}
```
JamesBurnside, line 82: Is there a reason to choose a function expression instead of JSX syntax?
i.e. `resumeSpinner()` instead of `<ResumeSpinner />`

JamesBurnside (approved):

PR #2277, opened 2022-08-26
Hide dialpad when it is not a PSTN call

carocao-msft:
# What
Hide dialpad when it is not a PSTN call

# Why
Dialpad should not show up in non  PSTN call

# How Tested
start a call and click on more button, should not see dialpad

start a PSTN call and click on more button, should see dialpad

# Process & policy checklist


packages/react-composites/tests/browser/call/hermetic/DtmfDialpad.test.ts
```diff
@@ -6,20 +6,34 @@ import { dataUiId, isTestProfileStableFlavor, pageClick, stableScreenshot, waitF
...
     const paul = defaultMockRemoteParticipant('Paul Bridges');
-    const vasily = defaultMockRemoteParticipant('Vasily Podkolzin');
 
-    const participants = [paul, vasily];
-    const initialState = defaultMockCallAdapterState(participants);
+    const participant = [paul];
+    const initialState = defaultMockCallAdapterState(participant);
 
+    //PSTN call has alternate caller id
+    initialState.alternateCallerId = '+1676568678999';
```
edwardlee-msft, line 18: Is this a "random" number not associated with a real phone number?
I am unsure whether or not we have numbers we can use for our testing.

packages/react-composites/tests/app/callwithchat/LiveApp.tsx
```diff
@@ -34,7 +34,9 @@ export function LiveApp(props: { queryArgs: QueryArgs }): JSX.Element {
       displayName: args.displayName,
       credential,
       endpoint: args.endpoint,
-      locator
+      locator,
+      /* @conditional-compile-remove(PSTN-calls) */
+      alternateCallerId: '+182927203720'
```
edwardlee-msft, line 39: Same as my other question. Is this a number that is permitted to be used for testing?

change/@internal-react-composites-b482f6b7-6a84-44ef-b79f-e7cd47c9570c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "hide dialpad when it's not PSTN",
```
edwardlee-msft, line 3: nit:
Change to "Dialpad hidden in non PSTN 1:1 Calls."

carocao-msft: updated

packages/react-composites/tests/browser/call/hermetic/DtmfDialpad.test.ts
edwardlee-msft: As long as we ensure that the number will not be a valid number ever, I am okay with this.

anjulgarg (approved):

PR #2278, opened 2022-08-26
Add first hermetic test for `CallWithChatComposite`

prprabhu-ms:
- Add a new `@internal` factory function for `CallWithChatAdapter` that takes the underlying `CallAdapter` and `ChatAdapter` as inputs.
- Extend the hermetic test setup for `CallComposite` and `ChatComposite` to be usable together for `CallWithChatComposite` tests.
- Migrate the simplest `CallWithChatComposite` test to be hermetic.

Aside:

- Fix a minor bug discovered by the new test: `CallWithChatComposite` initially shows only the `CallComposite` until _some_ state change is triggered.


packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -191,6 +191,7 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
       setCurrentPage(newState.page);
       setCurrentCallState(newState.call?.state);
     };
+    updateCallWithChatPage(callWithChatAdapter.getState());
```
prprabhu-ms, line 194: This is a fix for a bug I discovered by the new hermetic test -- we depended on _some_ state change to correctly show the `ChatComposite` and `PeoplePane` that are part of the `CallWithChatComposite`. It is probably not noticed in practice because state changes occur frequently.

packages/react-composites/tests/browser/callwithchat/hermetic/fixture.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @returns URL string
+ */
+export const buildUrlForApp = (
+  serverUrl: string,
+  mockCallAdapterState: MockCallAdapterState,
+  fakeChatAdapterArgs: FakeChatAdapterArgs,
+  qArgs?: { [key: string]: string }
+): string => {
+  return `${serverUrl}?${encodeQueryData({
```
edwardlee-msft, line 30: Nothing significant to note here, a random observation: 
URLs have a max length of 2048 characters. So loading a URL might experience odd behaviour if there are multiple users with really long usernames in our tests. Which we can easily avoid and constrain if we wanted to.

packages/react-composites/tests/app/callwithchat/HermeticApp.tsx
```diff
@@ -11,15 +17,48 @@ import { QueryArgs } from './QueryArgs';
...
+}
+
+function HermeticAppImpl(props: { queryArgs: HermeticQueryArgs & CommonQueryArgs }): JSX.Element {
+  const { queryArgs: args } = props;
+
+  const [callAdapter, setCallAdapter] = useState<CallAdapter | undefined>(undefined);
+  useEffect(() => {
+    (async (): Promise<void> => {
+      console.log('Creating mock adapter with args', args.mockCallAdapterState);
+      setCallAdapter(new MockCallAdapter(args.mockCallAdapterState));
```
edwardlee-msft, line 45: Correct me if I am mistaken, but I don't see an arg for alternateCallerId. I suppose for now the HermeticApp will not work with PSTN calls?

anjulgarg (approved):

packages/react-composites/tests/browser/callwithchat/hermetic/fixture.ts
prprabhu-ms: Yeah, we'll cross that bridge when we have to.

gzip ftw ;)

packages/react-composites/tests/app/callwithchat/HermeticApp.tsx
prprabhu-ms: As long as `alternateCallerId` is in the adapter state, tests can initialize it to required value.

In fact, there are already hermetic tests for CallComposite that set `alternateCallerId`.

edwardlee-msft (approved):

dmceachernmsft (approved):

PR #2280, opened 2022-08-26
Update addParticipant adapter API to match remove participant

dmceachernmsft:
# What

adds overloads to `addParticipant` to work with a flat identifier
# Why

Brings `addParticipant` into parity with `removeParticipant`

https://skype.visualstudio.com/SPOOL/_workitems/edit/2964569
# How Tested


packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -144,6 +144,10 @@ export type CallAdapterCallEndedEvent = {
 export interface CallAdapterCallManagement {
     // @beta
     addParticipant(participant: CommunicationIdentifier, options?: AddPhoneNumberOptions): Promise<void>;
+    // (undocumented)
+    addParticipant(participant: string): Promise<void>;
+    // (undocumented)
+    addParticipant(participant: CommunicationIdentifier | string, options?: AddPhoneNumberOptions): Promise<void>;
```
prprabhu-ms, line 150: Did we finally agree on something like:

```
addParticipant(participant: string): Promise<void>;
addParticipant(participant: CommunicationUserIdentifier): Promise<void>;
addParticipant(participant: PhoneNumberIdentifier, options? AddPhoneNumberOptions): Promise<void>;
```

Or did

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -423,7 +427,11 @@ export interface CallWithChatAdapter extends CallWithChatAdapterManagement, Adap
 // @public
 export interface CallWithChatAdapterManagement {
     // @beta
-    addParticipant: (participant: CommunicationIdentifier, options?: AddPhoneNumberOptions) => Promise<void>;
+    addParticipant(participant: CommunicationIdentifier, options?: AddPhoneNumberOptions): Promise<void>;
+    // (undocumented)
+    addParticipant(participant: string): Promise<void>;
+    // (undocumented)
+    addParticipant(participant: CommunicationIdentifier | string, options?: AddPhoneNumberOptions): Promise<void>;
```
prprabhu-ms, line 434: This last one should only be an implementation signature, not part of the API.

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatBackedCallAdapter.ts
```diff
@@ -116,12 +116,20 @@ export class CallWithChatBackedCallAdapter implements CallAdapter {
...
+  public async addParticipant(participant: CommunicationUserIdentifier): Promise<void>;
+  /* @conditional-compile-remove(PSTN-calls) */
+  public async addParticipant(
+    participant: PhoneNumberIdentifier | CommunicationUserIdentifier,
     options?: AddPhoneNumberOptions
-  ): Promise<void> => {
-    await this.callWithChatAdapter.addParticipant(participant, options);
-  };
+  ): Promise<void> {
+    if (typeof participant === 'string') {
```
anjulgarg, line 127: This doesn't seem right.
The type here will never be string.

anjulgarg: Additionally, wouldn't just calling `this.callWithChatAdapter.addParticipant(participant, options);` work irrespective of what the participant type is?

dmceachernmsft: yes, and yes, will update! missed this string usage 😅

anjulgarg (approved):

packages/calling-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -340,14 +347,19 @@ export const createDefaultCallingHandlers = memoizeOne(
...
-      if (isPhoneNumberIdentifier(participant)) {
-        await call?.addParticipant(participant, options);
-      } else if (isCommunicationUserIdentifier(participant) || isMicrosoftTeamsUserIdentifier(participant)) {
-        await call?.addParticipant(participant);
+    const onAddParticipant: AddParticipantHandler = async (participant, options?) => {
+      const participantType = participantTypeHelper(participant);
+      switch (participantType) {
+        case 'PSTN':
+          if (options) {
+            await call?.addParticipant(participant as PhoneNumberIdentifier, options);
```
prprabhu-ms, line 355: Aren't `options` optional in the headless API as well? We shouldn't add another layer of validation. Instead, the headless SDK should throw if it really needs the options to be defined.

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -423,7 +426,9 @@ export interface CallWithChatAdapter extends CallWithChatAdapterManagement, Adap
 // @public
 export interface CallWithChatAdapterManagement {
     // @beta
-    addParticipant: (participant: CommunicationIdentifier, options?: AddPhoneNumberOptions) => Promise<void>;
+    addParticipant(participant: CommunicationIdentifier, options?: AddPhoneNumberOptions): Promise<void>;
```
prprabhu-ms, line 429: This should be `PhoneNumberIdentifier`.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -505,8 +503,19 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
+  public async addParticipant(participant: PhoneNumberIdentifier, options?: AddPhoneNumberOptions): Promise<void>;
+  /* @conditional-compile-remove(PSTN-calls) */
+  public async addParticipant(participant: CommunicationUserIdentifier): Promise<void>;
+  /* @conditional-compile-remove(PSTN-calls) */
+  public async addParticipant(
+    participant: PhoneNumberIdentifier | CommunicationUserIdentifier,
+    options?: AddPhoneNumberOptions
+  ): Promise<void> {
+    if (isPhoneNumberIdentifier(participant) && options) {
+      this.handlers.onAddParticipant(participant as PhoneNumberIdentifier, options);
```
prprabhu-ms, line 515: Do you need `as PhoneNumberIdentifier` here? Doesn't `isPhoneNumberIdentifier` constrain the type already? Try removing that cast.

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatBackedCallAdapter.ts
```diff
@@ -116,12 +116,16 @@ export class CallWithChatBackedCallAdapter implements CallAdapter {
...
+  public async addParticipant(participant: CommunicationUserIdentifier): Promise<void>;
+  /* @conditional-compile-remove(PSTN-calls) */
+  public async addParticipant(
+    participant: PhoneNumberIdentifier | CommunicationUserIdentifier,
     options?: AddPhoneNumberOptions
-  ): Promise<void> => {
-    await this.callWithChatAdapter.addParticipant(participant, options);
-  };
+  ): Promise<void> {
+    return await this.callWithChatAdapter.addParticipant(participant, options);
```
prprabhu-ms, line 127: nitty nit: You can just `return this.callWithChatAdapter.addParticipant(...)`.

No need to `await` the Promise here since you don't do anything with the result. Caller can `await` when the want.

prprabhu-ms (approved):
Mostly there. Let me know if any of the comments are unclear.

packages/calling-component-bindings/src/handlers/createHandlers.ts
anjulgarg: +1
@dmceachernmsft let's keep this optional in our api as well.
And let the headless sdk throw :)

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -143,7 +144,9 @@ export type CallAdapterCallEndedEvent = {
 // @public
 export interface CallAdapterCallManagement {
     // @beta
-    addParticipant(participant: CommunicationIdentifier, options?: AddPhoneNumberOptions): Promise<void>;
+    addParticipant(participant: PhoneNumberIdentifier, options?: AddPhoneNumberOptions): Promise<void>;
+    // (undocumented)
+    addParticipant(participant: CommunicationUserIdentifier): Promise<void>;
```
JamesBurnside, line 149: out of interest what happens in a group call here? does it call the participant and show them as calling in the video gallery?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
dmceachernmsft: the function unfortunately doesn't. it just returns a boolean so typescript gets no type information about `participant` here.

packages/react-composites/review/beta/react-composites.api.md
dmceachernmsft: I believe if Contoso were to call `adapter.addParticipant(...)` it would fail since the type of call is incorrect for these handlers.

PR #2281, opened 2022-08-26
Allow contoso to directly set the value inside textfield 

carocao-msft:
# What
removed ondisplaydialpadinput, add textFieldValue for contoso to directly set the value inside textfield

# Why
Change made according to internal review recommendations 

# How Tested
storybook example showing how to use those 2 props to customized textfield format

### DIFF
--- a/packages/react-components/src/components/Dialpad/Dialpad.tsx
+++ b/packages/react-components/src/components/Dialpad/Dialpad.tsx
@@ -1,11 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-<<<<<<< HEAD
-import React, { useEffect } from 'react';
-=======
-import React, { useCallback } from 'react';
->>>>>>> b4ff2cfcd37a8d13143e560fae5a775760623137
+import React, { useEffect, useCallback } from 'react';
 import { IStyle, IButtonStyles, ITextFieldStyles } from '@fluentui/react';
 
 import { IconButton } from '@fluentui/react';
@@ -102,7 +98,6 @@ export type DtmfTone =
   | 'Pound'
   | 'Star';
 
-<<<<<<< HEAD
 /**
  * Props for {@link Dialpad} component.
  *
@@ -126,11 +121,8 @@ export interface DialpadProps {
   styles?: DialpadStyles;
 }
 
-const DialpadButton = (props: {
-=======
 type DialpadButtonContent = {
   /** Number displayed on each dialpad button */
->>>>>>> b4ff2cfcd37a8d13143e560fae5a775760623137
   primaryContent: string;
   /** Letters displayed on each dialpad button */
   secondaryContent?: string;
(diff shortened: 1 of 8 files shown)


packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -184,73 +207,64 @@ const DialpadContainer = (props: {
   onSendDtmfTone?: (dtmfTone: DtmfTone) => Promise<void>;
   /**  Callback for dialpad button behavior */
   onClickDialpadButton?: (buttonValue: string, buttonIndex: number) => void;
-  /**  customize dialpad input formatting */
-  onDisplayDialpadInput?: (input: string) => string;
+  /** Pass in custom content to dialpad textfield */
```
dmceachernmsft, line 210: should this comment maybe be something like :
`/** Access to the text field within dialpad */`

packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -184,73 +207,64 @@ const DialpadContainer = (props: {
...
-    }
   };
 
   const onLongPressDialpad = (input: string, index: number): void => {
-    let value;
     if (input === '0' && index === 10) {
-      // remove non-valid characters from input: letters,special characters excluding +, *,#
-      value = sanitizeInput(textValue + '+');
-      setTextValue(value);
+      setText(plainTextValue + '+');
```
dmceachernmsft, line 246: I might be forgetting something from this morning but is this supposed to be: 
`'+' + plainTextValue`? 

I feel like this would result in a phone number like `14443332222+`

carocao-msft: This is code for the longpress functionality, long pressing 0 will append + instead of 0

packages/react-components/src/components/Dialpad/Dialpad.tsx
carocao-msft: value passed in here will directly set the content in the textfield, so it should not be called 'access'?

dmceachernmsft: oh sorry I was thinking about accessing that value, they can still do that through onChange right?

dmceachernmsft (approved):

packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -98,6 +78,26 @@ export type DtmfTone =
...
+export interface DialpadProps {
+  strings?: DialpadStrings;
+  /**  function to send dtmf tones on button click */
+  onSendDtmfTone?: (dtmfTone: DtmfTone) => Promise<void>;
+  /**  Callback for dialpad button behavior*/
+  onClickDialpadButton?: (buttonValue: string, buttonIndex: number) => void;
+  /** set dialpad textfield content */
+  textFieldValue?: string;
+  /**  on change function for text field */
+  onChange?: (input: string) => void;
```
anjulgarg, line 95: Can we call it out here that this callback provides an unformatted plain text?

anjulgarg (approved):

packages/react-components/src/components/Dialpad/Dialpad.tsx
carocao-msft: updated

PR #2282, opened 2022-08-26
Add new troubleshooting API for Call Readiness

dmceachernmsft:
# What

Introduces new troubleshooting API's for networking and devicePermission advice pages
# Why

Allows contoso to pass in a on click function to allow our errorBars to redirect the end user to a supplied help page for devices or networking.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2961426
# How Tested

will be introduced to errorBar in following PR for these conditions.


mgamis-msft (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -82,6 +82,39 @@ export type CallCompositeOptions = {
...
+    microphone: PermissionState;
+  }) => void;
+  /* @conditional-compile-remove(call-readiness) */
+  /**
+   * Callback you may provide to supplu users with further steps to troubleshoot why they have been
+   * having network issues when connecting to the call.
+   *
+   * @example
+   * ```ts
+   * onNetworkingTroubleShootingClick?: () =>
```
prprabhu-ms, line 110: Since these methods return `void`, the only UX that Contoso can provide here is via:

* navigating away from the call entirely
* Overlaying something on top of the call UI

The first one seems fairly disruptive, and not something I'd advice Contoso to do.

Should we instead accept a function that returns a `JSX.Element` that we overlay on the composite ourselves?

dmceachernmsft: So the reason for this is Contoso is to provide their own 'extra' help page. due to the possibilities of issues we cannot provide a solution every single possible reason that the end user is having issues with their camera, mic or network. these API's are to be used in our error bars to send the end user to a help page that is provided by Contoso. This page might talk about common network issues, or a more detailed guide on how to enable camera/mic permissions on system and browser.

JamesBurnside: It's similar to custom button injection, sure they can inject a button - but there's such limited space for Contoso to present UI based on that button click.
I think we should keep this as void - Contoso can then present a modal dialog or open a new tab page as desired.
We can then continue the conversation on customization points - I strongly feel we should be exposing the side pane a custom content point

common/config/babel/.babelrc.js
```diff
@@ -34,7 +34,9 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         // dialpad
         'dialpad',
         // rooms
-        'rooms'
+        'rooms',
+        // Call Readiness
```
JamesBurnside, line 38: remove comment, the name is enough here

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -82,6 +82,39 @@ export type CallCompositeOptions = {
...
+   * @remarks
+   * if this is not supplied, the composite will not show a 'further troubleshooting' link.
+   */
+  onPermissionsTroubleshootingClick?: (permissionsState: {
+    camera: PermissionState;
+    microphone: PermissionState;
+  }) => void;
+  /* @conditional-compile-remove(call-readiness) */
+  /**
+   * Callback you may provide to supplu users with further steps to troubleshoot why they have been
```
edwardlee-msft, line 105: supplu to supply

edwardlee-msft (approved):

Issue #2257, opened 2022-08-24
Unable to change the style of myChatMessageContainer for the MessageThread component


jboinembalome:
**Describe the bug; what happened?**
First of all, thank you very much for your work.
The `MessageThread `component has a "styles" prop allowing us to change the style of the following elements:
- `myChatMessageContainer`
- `chatMessageContainer`

If we add style for both elements we can see that the style is applied for `chatMessageContainer `and not for `myChatMessageContainer`.

**What are the steps to reproduce the issue?**

1. Create a constant containing the style to be applied:
```
const messageThreadStyle : MessageThreadStyles = {
  chatMessageContainer: {
    background: '#F5F5F5',
    borderRadius: '8px'
  },
  myChatMessageContainer: {
    background: 'cornflowerblue',
    borderRadius: '8px'
  }
};
```

2. Add the style to the MessageThread component:
<MessageThread
        userId={currentUserId}
        messages={messages}
        showMessageStatus={true}
        showMessageDate={true}
        styles={messageThreadStyle}
 />

3. If you start your project, you will get the following result:
![image](https://user-images.githubusercontent.com/70820612/186355102-9b217bf9-820a-443a-a0cd-85f5aa1ffdeb.png)

As we can see, the background of `myChatMessageContainer `is in red.

**What behavior did you expect?**
If possible, I would like to have the possibility to add styling for `myChatMessageContainer `in order to be consistent with `chatMessageContainer `.

**If applicable, provide screenshots:**
Instead of this:
![image](https://user-images.githubusercontent.com/70820612/186355102-9b217bf9-820a-443a-a0cd-85f5aa1ffdeb.png)

I would like this:
![image](https://user-images.githubusercontent.com/70820612/186356638-57368ccf-bbef-42f6-b95a-dd62f75b33a1.png)

**In what environment did you see the issue?**

- `@azure/communication-react` npm package version (if applicable): 1.3.1-beta.1

- OS & Device: Windows
- Browser: Google Chrome
- Browser Version: 104.0.5112.102 (Official Build) (64-bit)

**Is there any additional information?**
After analyzing the source code, it seems that the style is not applied because of the condition line 361 of the **MessageThread.tsx** file:  
```
const myChatMessageStyle =
          styles?.myChatMessageContainer || message.status === 'failed'
            ? FailedMyChatMessageContainer
            : defaultMyChatMessageContainer;
```

[...]

PR #2285, opened 2022-08-27
Bugfix - Unable to update myChatMessageContainer style in MessageThread (#2257)

jboinembalome:
# What

Fixed the inability to change the style of myChatMessageContainer for the MessageThread component.

- Added failedMyChatMessageContainer to the MessageThreadStyles interface.
- Updated the myChatMessageStyle constant according to failedMyChatMessageContainer.

# Why

The `MessageThread `component has a "styles" prop allowing us to change the style of the following elements:
- myChatMessageContainer
- chatMessageContainer

If we add style for both elements we can see that the style is applied for `chatMessageContainer `and not for `myChatMessageContainer`.

Link of the issue: #2257 

# How Tested

We can check the behaviour in the storybook:
- Go to the storybook
- Go to the "Messages with Customized Message Container" section of the `MessageThread `component.

The `MessageThread `will apply the style of `myChatMessageContainer `(fontStyle and boxShadow) instead of applying the style of `FailedMyChatMessageContainer `(red background).

# Process & policy checklist


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -359,9 +361,9 @@ const memoizeAllMessages = memoizeFnAll(
     switch (message.messageType) {
       case 'chat': {
         const myChatMessageStyle =
-          styles?.myChatMessageContainer || message.status === 'failed'
-            ? FailedMyChatMessageContainer
-            : defaultMyChatMessageContainer;
+          message.status === 'failed'
+            ? styles?.failedMyChatMessageContainer || FailedMyChatMessageContainer
```
JamesBurnside, line 365: In the event `failedMyChatMessageContainer` this should fallback to `myChatMessageContainer` (i.e. we consider failedMyChatMessageContainer to be a more specific override).

```suggestion
            ? styles?.failedMyChatMessageContainer ?? styles?.myChatMessageContainer ?? FailedMyChatMessageContainer
```

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -152,6 +152,8 @@ export interface MessageThreadStyles extends BaseCustomStyles {
   chatItemMessageContainer?: ComponentSlotStyle;
   /** Styles for my chat message container. */
   myChatMessageContainer?: ComponentSlotStyle;
+  /** Styles for my chat message container in case of failure.  */
+  failedMyChatMessageContainer?: ComponentSlotStyle;
```
JamesBurnside, line 156: This changes the public API for this component, can you run `rush build:all-flavors` to update the api.md files

JamesBurnside (changes requested):
Thanks for making this contribution! Left some comments.
One extra step along with updating the API files you will need to make is to run `rush changelog`.
This generates a change file that will be included in our changelog when we release a new version.
This change is a `minor` change as we are adding a new field to the message thread component.
And for a changelog entry your PR title is good to use 👍

packages/react-components/src/components/MessageThread.tsx
jboinembalome: Updated according to your suggestion 👍

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -359,9 +361,9 @@ const memoizeAllMessages = memoizeFnAll(
     switch (message.messageType) {
       case 'chat': {
         const myChatMessageStyle =
-          styles?.myChatMessageContainer || message.status === 'failed'
-            ? FailedMyChatMessageContainer
-            : defaultMyChatMessageContainer;
+          message.status === 'failed'
+            ? styles?.failedMyChatMessageContainer || FailedMyChatMessageContainer
+            : styles?.myChatMessageContainer || defaultMyChatMessageContainer;
```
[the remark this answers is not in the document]
jboinembalome: Updated according to your suggestion 👍

packages/react-components/src/components/MessageThread.tsx
jboinembalome: I ran the `rush build:all-flavors` command and pushed the api.md files

jboinembalome:
Thank you @JamesBurnside for checking the PR.
For information, I did not get any changes after running `rush changelog`.

Below is the result:
image

Could it be because I have synchronized the `Azure:main` branch with my branch?

JamesBurnside (approved):

JamesBurnside:
> Thank you @JamesBurnside James Burnside FTE for checking the PR. For information, I did not get any changes after running `rush changelog`.

It appears we have a bug in our changefile configuration when accepting contributions from a forked repo - we will look into fixing this asap and I will update this thread once we have a fix in

anjulgarg (approved):

JamesBurnside:
Thanks for your pateience, the `rush changelog` command has now been fixed so you should be good to run and push the changefile to this PR.
Can you also run `rush build:all-flavors` one more time? I would expect updates to `review/stable/react-components.api.md` and `review/stable/communication-react.api.md` files also.

JamesBurnside:
> 
> @JamesBurnside Thanks for fixing the `rush changelog` command, I ran the command and pushed the changefile. Next, I ran the `rush build:all-flavors` command and unfortunately I got no change.

We're unsure why this is happening for you locally. I tried cloning your fork and building locally and it is updating the stable API file. As we are doing a new npm release very soon I'd like to include this change in the release so we've created a clone of this PR #2449 and included you as coauthor.

PR #2286, opened 2022-08-29
remove the rejoin button from end call screen in calling composite

dmceachernmsft:
# What

Remove the end call screen's rejoin call button
# Why

Better supports the locator method of starting a PSTN or 1:N call, should the users change in the call the locator would not update. the issue here being then that when they action the `rejoin call` button they will redial the original participants of the call. instead of the composite having to update the locator on each new participant change (leave or join) we rely on Contoso to make this button. 

This is probably a breaking UI change so please leave your thoughts about this. 

# How Tested

image


PorterNan:
Thanks for bringing this up in standup - I guess the current problem is we can't 100% determine if that is a PSTN call or not only by locator - since this could be a future problem as well: Is it possible to add a new property in adapter state named isPSTNCall?

If someone call the PSTN number - even it is a group call, the flag will be permanently changed - then we can always leverage this flag for multiple behaviors in the future (including dialog three dots button or so?)

dmceachernmsft:
This work was completed in a better way.

PR #2287, opened 2022-08-29
Composite end call screen when room call join fails

mgamis-msft:
# What

Using subcodes from Calling SDK  to show a different end call screen in the composite when call is ended because:
- room does not exist
- user was not invited to room

The end call screens were added to composite instead of the sample as discussed with @alkwa-msft. Also, composite users will be much more informed when a room call join fails.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2966741
https://skype.visualstudio.com/SPOOL/_workitems/edit/2966748

# How Tested

Tested with call composite preview in storybook.

1) When room does not exist
https://user-images.githubusercontent.com/79475487/187261071-2ace3bc3-26a8-4f07-a3b9-0073ff2efb5b.mp4

2) When user was not invited to the room
https://user-images.githubusercontent.com/79475487/187261180-64b5f24b-7083-45eb-a63f-569d04140eab.mp4

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -156,6 +156,34 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
         />
       );
       break;
+    case roomNotFoundPageTrampoline():
```
alkwa-msft, line 159: Lets be explicit about the two cases rather than putting them inside one function?

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -147,6 +165,12 @@ export const getCallCompositePage = (
           return 'joinCallFailedDueToNoNetwork';
         }
         return 'leftCall';
+      /* @conditional-compile-remove(rooms) */
+      case CallEndReasons.ROOM_NOT_FOUND:
+        return roomNotFoundPageTrampoline();
+      /* @conditional-compile-remove(rooms) */
+      case CallEndReasons.NOT_INVITED_TO_ROOM:
+        return notInvitedToRoomPageTrampoline();
```
PorterNan, line 173: Do we need a trampoline here? The entire case statement will be removed (if not just remove return statement as well)

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -156,6 +156,34 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
...
+              iconName="InvalidRoom"
+              title={locale.strings.call.roomNotFoundTitle}
+              moreDetails={locale.strings.call.roomNotFoundDetails}
+              dataUiId={'room-not-found-page'}
+            />
+          }
+        </>
+      );
+      break;
+    case notInvitedToRoomPageTrampoline():
```
PorterNan, line 173: Possible to remove whole case block in conditional compilation? This becomes 'endCall' when it is stable, wondering if this will block the condition for following sentences. Adding less code for stable will make it safer

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -156,6 +156,34 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
         />
       );
       break;
+    case roomNotFoundPageTrampoline():
+      pageElement = (
+        <>
+          {
+            /* @conditional-compile-remove(rooms) */ <NoticePage
```
prprabhu-ms, line 163: Soft suggestion:
Instead of conditionally compiling these JSX, add a single trampoline that returns all the strings necessary for beta even on stable build, by providing empty default values.

There are examples of this strategy in the codebase.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -10,6 +10,10 @@ import { isPhoneNumberIdentifier } from '@azure/communication-common';
 const ACCESS_DENIED_TEAMS_MEETING_SUB_CODE = 5854;
 const REMOTE_PSTN_USER_HUNG_UP = 560000;
 const REMOVED_FROM_CALL_SUB_CODES = [5000, 5300, REMOTE_PSTN_USER_HUNG_UP];
+/* @conditional-compile-remove(rooms) */
```
prprabhu-ms, line 13: This constant is not exported in the API. Remove conditional compilation?

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -46,7 +48,9 @@ export const END_CALL_PAGES: CallCompositePage[] = [
   'accessDeniedTeamsMeeting',
   'joinCallFailedDueToNoNetwork',
   'leftCall',
-  'removedFromCall'
```
prprabhu-ms, line 49: This is an `@private` symbol. Add the new values without conditional compilation. It might help you reduce some conditional compilation elsewhere in the logic as well.

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
prprabhu-ms, line 159: Why do you need a trampoline function here at all? What happens if you simply say `case 'roomNotFound'` here?

Your trampoline function returns a `string` which TypeScript seems to be happy about...

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -36,7 +36,9 @@ export type CallCompositePage =
   | 'joinCallFailedDueToNoNetwork'
   | 'leftCall'
   | 'lobby'
-  | 'removedFromCall';
```
prprabhu-ms, line 39: I'm regretting that we added separate pages for all the end call reasons. Why do we expose to contoso exactly what the reason for ending a call was via the page? :(
This is something there should be an adapter event for. There should have been a single `callEnded` page. We could always fill in the right content based on the reason for the call being ended.

How do we feel about cutting our losses and adding a single `callEnded` page now and lumping all future call ended notices to that single page?

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -58,7 +62,11 @@ export const reduceCallControlsForMobile = (
 enum CallEndReasons {
   LEFT_CALL,
   ACCESS_DENIED,
-  REMOVED_FROM_CALL
+  REMOVED_FROM_CALL,
+  /* @conditional-compile-remove(rooms) */
```
prprabhu-ms, line 66: This enum is not exported. Remove conditional compilation directives.

This may allow you to remove conditional compilation from logic that uses this enum as well.

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -46,6 +48,8 @@
     "removeMenuLabel": "Remove",
     "returnToCallButtonAriaDescription": "Return to Call",
     "returnToCallButtonAriaLabel": "Back",
+    "roomNotFoundDetails": "Room ID provided in call locator is not valid.",
```
prprabhu-ms, line 51: The end user doesn't know anything about "call locator". Let's double check on these strings with UX.

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
alkwa-msft: https://skype.visualstudio.com/SPOOL/_workitems/edit/2976124

prabhjot-msft (approved):

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -358,6 +358,7 @@ export type CallCompositeIcons = {
     ParticipantItemScreenShareStart?: JSX.Element;
     VideoTileMicOff?: JSX.Element;
     LocalCameraSwitch?: JSX.Element;
+    InvalidRoom?: JSX.Element;
```
prprabhu-ms, line 361: Looks like you went a bit too far with removal of conditional compilation.

This icon shouldn't be there in the stable API.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -367,7 +368,7 @@ export type CallCompositeOptions = {
 };
 
 // @public
-export type CallCompositePage = 'accessDeniedTeamsMeeting' | 'call' | 'configuration' | 'joinCallFailedDueToNoNetwork' | 'leftCall' | 'lobby' | 'removedFromCall';
+export type CallCompositePage = 'accessDeniedTeamsMeeting' | 'call' | 'configuration' | 'joinCallFailedDueToNoNetwork' | 'leftCall' | 'lobby' | 'notInvitedToRoom' | 'removedFromCall' | 'roomNotFound';
```
prprabhu-ms, line 371: Ditto, these new page types shouldn't be there in the stable API.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -415,10 +416,14 @@ export interface CallCompositeStrings {
     mutedMessage: string;
     networkReconnectMoreDetails: string;
     networkReconnectTitle: string;
+    notInvitedToRoomDetails?: string;
```
prprabhu-ms, line 419: Ditto. These new strings shouldn't be in the stable API.

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -105,6 +105,27 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
...
+  switch (page) {
+    case 'roomNotFound':
+      pageElement = (
+        <NoticePage
+          iconName="InvalidRoom"
+          title={locale.strings.call.roomNotFoundTitle}
+          moreDetails={locale.strings.call.roomNotFoundDetails}
+          dataUiId={'room-not-found-page'}
+        />
+      );
```
PorterNan, line 118: break;

just as what warning says

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -10,6 +10,8 @@ import { isPhoneNumberIdentifier } from '@azure/communication-common';
 const ACCESS_DENIED_TEAMS_MEETING_SUB_CODE = 5854;
 const REMOTE_PSTN_USER_HUNG_UP = 560000;
 const REMOVED_FROM_CALL_SUB_CODES = [5000, 5300, REMOTE_PSTN_USER_HUNG_UP];
+const ROOM_NOT_FOUND_SUB_CODE = 5751;
```
JamesBurnside, line 13: do you know if calling will support these as end call reasons is their new end call reason api?

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -86,6 +90,14 @@ const getCallEndReason = (call: CallState): CallEndReasons => {
     return CallEndReasons.REMOVED_FROM_CALL;
   }
 
+  if (call.callEndReason?.subCode && call.callEndReason.subCode === ROOM_NOT_FOUND_SUB_CODE) {
```
JamesBurnside, line 93: we should conditionally compile these - unless we are *certain* we can't get these subcodes unintentionally

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -36,6 +36,8 @@
     "mutedMessage": "You're muted",
     "networkReconnectMoreDetails": "Looks like something went wrong. We're trying to get back into the call.",
     "networkReconnectTitle": "Hold on",
+    "notInvitedToRoomDetails": "You were not invited to this room.",
```
JamesBurnside, line 39: nit: normally we talk about not having permission to join instead of being 'invited' (e.g. you could be unauthorized for reasons other than an invitation (e.g. if you were invited but your invite subsequently revoked).
Just want to check we shouldn't go with something like `deniedPermissionToRoom`

PorterNan (approved):

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
mgamis-msft: Good question. I don't know at this time. I will ask their Teams thread.

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -334,6 +335,8 @@ export interface CallCompositeStrings {
     copyInviteLinkActionedAriaLabel: string;
     copyInviteLinkButtonLabel: string;
     defaultPlaceHolder: string;
+    deniedPermissionToRoomDetails?: string;
```
prprabhu-ms, line 338: Are these strings added to the stable API as well? 

I don't see the stable API file at all. Did you build for stable flavor as well?

packages/react-composites/src/composites/localization/locales/en-US/strings.json
prprabhu-ms: I'd reconsider the string values as well. The message "not invited" was a bit confusing to me in the UI snapshots.

"You do not have access" would have sounded more assertive.

But, if we can, let's get UX input for the string values.

prprabhu-ms:
LGTM, but letting @JamesBurnside approve because he has valid outstanding comments.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -387,6 +387,7 @@ export type CallCompositeIcons = {
     PeoplePaneAddPerson?: JSX.Element;
     PeoplePaneOpenDialpad?: JSX.Element;
     DialpadStartCall?: JSX.Element;
+    InvalidRoom?: JSX.Element;
```
JamesBurnside, line 390: For consistency, can we use `NoticePageInvalidRoom` -- don't love this naming but just to keep it consistent with the other end call page icons

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -36,6 +36,8 @@
     "mutedMessage": "You're muted",
     "networkReconnectMoreDetails": "Looks like something went wrong. We're trying to get back into the call.",
     "networkReconnectTitle": "Hold on",
+    "deniedPermissionToRoomDetails": "You were not invited to this room.",
```
JamesBurnside, line 39: I would update the strings here (and the line below) too, we can't know exactly that they weren't "invited" to the room and instead use a more generic "You do not have permission to join this room"

JamesBurnside (approved):
Some outstanding nit comments RE strings and conditional-compilation but otherwise looks great!

PR #2288, opened 2022-08-29
left side ellipsis on dtmf dialpad

dmceachernmsft:
# What

Introduces a hide formatting prop to the dialpad. 
# Why

this allows the dtmf dialpad and Contoso to disable the autoformatting in the dialpad that we provide. also moves the ellipsis for the dtmf dialpad overflow to the left side the end user can see what they are typing.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2958039
# How Tested

Ran locally in the apps to verify props and styles work appropriately


anjulgarg:
Not sure how hiding the formatting starts showing the ellipses on the left?

anjulgarg:
Wouldn't `rtl` cause the bug @carocao-msft just fixed again?

carocao-msft:
> Wouldn't `rtl` cause the bug @carocao-msft just fixed again?

We chat about this, the new prop is added so contoso can use that prop to disable the default formatting we have on dialpad, rtl styling to show ... on the left. The bug I fixed is caused by rtl and default formatting exisiting together, so for dtmf dialpad we apply rtl styling and diable default format, in this way ... is shown on the left and the bug I fixed ealier should not happen again :)

PR #2289, opened 2022-08-29
add announcer to composites copyinvite link

dmceachernmsft:
# What

Adds announcer component in people pane of Calling and CallWIthChat composites
# Why

Causes narrator to announce the link was copied when they action the button

https://skype.visualstudio.com/SPOOL/_workitems/edit/2790797
# How Tested

Ran locally and validated with Narrator the string was being announced.


packages/react-composites/src/composites/common/AddPeopleDropdown.tsx
```diff
@@ -60,6 +63,12 @@ export const AddPeopleDropdown = (props: AddPeopleDropdownProps): JSX.Element =>
       useTargetWidth: true,
       calloutProps: {
         preventDismissOnEvent
+      },
+      onMenuOpened(contextualMenu?) {
+        console.log('actioned' + announcerStrings);
```
edwardlee-msft, line 68: Should there be a console log here?

packages/react-composites/src/composites/common/AddPeopleDropdown.tsx
```diff
@@ -60,6 +63,12 @@ export const AddPeopleDropdown = (props: AddPeopleDropdownProps): JSX.Element =>
...
+        if (announcerStrings !== undefined) {
+          setAnnouncerStrings(undefined);
```
edwardlee-msft, line 70: nit: simplify if statement to
`setAnnouncerStrings(announcerStrings ?? undefined)`

edwardlee-msft: For reference:
https://dev.to/sunflower/reactjs-if-it-is-setting-a-state-with-the-same-value-will-the-component-be-re-rendered-5g24 
undefined is needed here as the function component would not be re-rendered if it was set to the same value down in line 82. Meaning, the announcer would not announce again if the copy invite link was actioned on more than once unless the announcerString is reset like in this case.

edwardlee-msft (approved):

packages/react-composites/src/composites/common/AddPeopleDropdown.tsx
dmceachernmsft: from testing this does work this way, but I will change it since yours is more readable

edwardlee-msft (approved):

prabhjot-msft (approved):

PR #2295, opened 2022-08-31
Show correct end call screen when PSTN user is removed from the call

anjulgarg:
# What
- Sow "You Left the Call" screen when you remove the last PSTN user.
- Show "You were removed" screen when a PSTN user hangs up the call.

# Why

# How Tested

# Process & policy checklist


dmceachernmsft (approved):

carocao-msft (approved):

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -60,6 +62,21 @@ enum CallEndReasons {
 }
 
 const getCallEndReason = (call: CallState): CallEndReasons => {
+  const remoteParticipantsEndedArray = Array.from(Object.values(call.remoteParticipantsEnded));
+  /**
+   * Handle the special case in a PSTN call where removing the last user kicks the caller out of the call.
```
PorterNan, line 67: So PSTN call can't hold the call for just one participant? Every PSTN call needs at least 2 people?

PorterNan (approved):

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
anjulgarg: Correct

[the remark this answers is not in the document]
edwardlee-msft: Tested with Anjul. Yes, that is the case. This will only work in an initialized 1:1 call scenario without any other participant being added to the call.

Issue #2099, opened 2022-07-19
Message Timestamp not showing for every message in (Azure communication services the UI Library) for chat 


sangeethav123:
We were using cross-platform - Chat composite js bundle file (Azure communication services the UI Library) for chat.
(https://azure.github.io/communication-ui-library/?path=/docs/use-composite-in-non-react-environment--page)

Currently its only showing the timestamp of the first message when a group of messages come from the same sender. Because of this we may not able to know the exact timestamp of the last received message. 
Example: In the screenshot attached below,
The first **hi** message sent **yesterday** and the last **hi** message sent **today** but, the receiver cannot able to know the exact timestamp of last message.
![image](https://user-images.githubusercontent.com/97170459/179698599-6f717e4c-ef34-46e7-9e41-4feb2b0382e0.png)

prabhjot-msft:
Hi @sangeethav123, I am looking into the issue. Will post here once I have some update.

carocao-msft:
Hi @sangeethav123 , thanks for bringing this up! 
I am wondering if you prefer to have timestamps on all messages or only show timestamps when there is a certain time gap between each message?

sangeethav123:
Hi,
I prefer to have timestamps when there is a certain time gap between each messages

carocao-msft:
Sounds good thanks! We will create a work item for this new feature and update you once it's available :)

carocao-msft:
Hi, we have made the changes and got this PR https://github.com/Azure/communication-ui-library/pull/2299 checked in for the problem, it will be in our next beta version very soon, please keep your eyes on next beta release 😉

PR #2299, opened 2022-09-01
Show datetime when there are >= 5 mins between each message 

carocao-msft:
# What
Show datetime when there are >= 5 mins between each message 

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2931330
https://github.com/Azure/communication-ui-library/issues/2099

# How Tested
Set the time gap to be 2 mins and test the chat sample

Screen Shot 2022-08-31 at 11 00 34 PM
# Process & policy checklist


anjulgarg:
Screen Shot 2022-09-01 at 2 08 01 PM
@carocao-msft Are we adding this gap between 2 message groups?

packages/chat-component-bindings/src/utils/updateMessagesWithAttached.ts
```diff
@@ -26,8 +26,16 @@ export const updateMessagesWithAttached = (chatMessagesWithStatus: Message[]): v
     const previousSenderId = previousMessage?.messageType === 'chat' ? previousMessage.senderId : undefined;
     const nextSenderId = nextMessage?.messageType === 'chat' ? nextMessage.senderId : undefined;
 
+    const timediff =
+      new Date(message?.createdOn ?? '').getTime() - new Date(previousMessage?.createdOn ?? '').getTime();
```
anjulgarg, line 30: I believe `new Date('')` will end up creating a new date time object with the latest date.
If `createdOn` properties are ever blank, this would fail.
Can you try taking this into account in the tests you wrote?

Ideally, this logic shouldn't run if createdOn date is not present for the previous message or the current message.

carocao-msft: Yes I tested this scenario, if createdOn is empty, new Date(' ') returns an invalid date object and in this case timediff will be NaN, then it wont meet the logic if (timediff && timediff >=5)

carocao-msft: Updated test :)

anjulgarg (approved):

packages/chat-component-bindings/src/utils/updateMessagesWithAttached.ts
```diff
@@ -26,8 +26,16 @@ export const updateMessagesWithAttached = (chatMessagesWithStatus: Message[]): v
...
+
+    const diffMins = Math.round(timediff / 1000 / 60); // minutes
```
PorterNan, line 32: Add a value to constants.ts 
like 
const MINUTE = 1000*60;

PorterNan: or MINUTE_IN_MS

packages/chat-component-bindings/src/utils/updateMessagesWithAttached.ts
```diff
@@ -26,8 +26,16 @@ export const updateMessagesWithAttached = (chatMessagesWithStatus: Message[]): v
...
+
     if (previousSenderId !== message.senderId) {
       attached = message.senderId === nextSenderId ? 'top' : false;
+    } // if there are more than or equal to 5 mins time gap between messages do not attach and show time stamp
```
PorterNan, line 36: move this comment into the code block so we don't need an extra new line for "else if"

packages/chat-component-bindings/src/utils/updateMessagesWithAttached.test.ts
```diff
@@ -56,4 +64,25 @@ describe('update message with attached status', () => {
     updateMessagesWithAttached(messagesArrayWithOtherMessage);
     expect(getAttachedStatusArray(messagesArrayWithOtherMessage)).toEqual([false, undefined, false]);
   });
+
+  test('Set right status for attached property for messages with more than or equal to 5 mins time gap', () => {
+    const currentDateTime = new Date();
+    const messagesArrayWithOtherMessage = [
+      cannedCustomDateTimeChatMessage('1', currentDateTime),
+      cannedCustomDateTimeChatMessage('1', new Date(currentDateTime.getDate() + 5 * 60000)),
```
PorterNan, line 72: Could also test mins like 2 or 4 to see if that works well

PorterNan (approved):

packages/chat-component-bindings/src/utils/updateMessagesWithAttached.ts
carocao-msft: updated

packages/chat-component-bindings/src/utils/updateMessagesWithAttached.test.ts
carocao-msft: updated

packages/chat-component-bindings/src/utils/updateMessagesWithAttached.ts
carocao-msft: updated

PR #2300, opened 2022-09-01
add isPSTNorACS call to adapter state to hide rejoin call button

dmceachernmsft:
# What

Adds new state to the call adapter to track if is ACS call or PSTN, set in `startCall` handler 
# Why

allows the composite to adapt to the type of call that it is when the call has ended.

# How Tested

image
Group call end call screen
image
PSTN end call screen


packages/react-composites/src/composites/CallComposite/pages/NoticePage.tsx
```diff
@@ -31,6 +31,7 @@ export interface NoticePageProps {
  */
 export function NoticePage(props: NoticePageProps): JSX.Element {
   const adapter = useAdapter();
+  const isPSTNorACSCall = adapter.getState().isPSTNorACSCall;
```
anjulgarg, line 34: Wouldn't recommend adding this attribute to adapter state since it will become publicly available
Could we track it internally in composites somehow?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -481,6 +481,11 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
       return backendId as UnknownIdentifier;
     });
 
+    if (idsToAdd !== []) {
+      // if we have id's to call then we want to set this state in the adapter to true.
+      this.getState().isPSTNorACSCall = true;
```
anjulgarg, line 486: When are we clearing this state variable?

packages/react-composites/src/composites/CallComposite/pages/NoticePage.tsx
dmceachernmsft: wouldn't we want to show this publicly in the adapter? this state could be useful to contoso to have their app do something special based on the type of call. we do also do the same thing with the `isTeamsCall` prop, its just being set in the constructor though.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
dmceachernmsft: as long as the composite isn't going back to the home page this should be set so that the rejoin button does not appear. setting this state back to false would be to contoso's implementation of their app.

packages/react-composites/src/composites/CallComposite/pages/NoticePage.tsx
anjulgarg: Ideally, such flags should be coming directly from the headless SDK as they are the ultimate source of truth

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -71,6 +71,7 @@ export type CallAdapterClientState = {
   devices: DeviceManagerState;
   endedCall?: CallState;
   isTeamsCall: boolean;
+  isPSTNorACSCall?: boolean;
```
PorterNan, line 74: Maybe we can set it as callType - so it could be 'PSTN', 'ACS' or more, so we can add more call types for future compatibilities

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -47,7 +47,7 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
   const afterCreate = useCallback(
     async (adapter: CallAdapter): Promise<CallAdapter> => {
       adapter.on('callEnded', () => {
-        onCallEnded();
+        // onCallEnded();
```
PorterNan, line 50: Is there specific reason to add comment here?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -481,6 +481,11 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
       return backendId as UnknownIdentifier;
     });
 
+    if (idsToAdd !== []) {
+      // if we have id's to call then we want to set this state in the adapter to true.
```
PorterNan, line 485: Question here: I guess idsToAdd includes normal acs id, so we should only set this to true when there is true phoneNumber in it

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: If we have a PSTN participant in the call, but that participant leave in the middle of the call, do we still consider this call as a PSTN call?

samples/Calling/src/app/views/CallScreen.tsx
JamesBurnside: (this I assume is just for testing the ensure the composite's end call page is shown and not the sample's custom end call page)

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -71,6 +71,7 @@ export type CallAdapterClientState = {
   devices: DeviceManagerState;
   endedCall?: CallState;
   isTeamsCall: boolean;
```
JamesBurnside, line 73: this is why we need to avoid booleans *sigh*

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
JamesBurnside: Is a growing list a breaking change? e.g. if the callType in a new version added 'pizzaParty' then contoso would hit a breaking change as they need to account for a new callType

packages/react-composites/src/composites/CallComposite/pages/NoticePage.tsx
JamesBurnside: I'd also like to avoid it being a public property - ideally contoso either would know as they pass in the locator, or they shouldn't care...
But our design means that the composite should only be using public adapter props so not sure we can avoid that...

JamesBurnside: (my vote would be to go with the other PR that removes the rejoin button and we do a major version bump)

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
dmceachernmsft: we want to be checking to see if we are calling ACS users as well, both kinds of id's can show in this array

PR #2301, opened 2022-09-02
Migrate chat-specific `CallWithChatComposite` browser tests to be hermetic

prprabhu-ms:
Migrate 3 chat-specific browser tests for `CallWithChatComposite` to the hermetic app.

While there:

* Add an id to the top-level test app container and use that to better target elements of the test app, as opposed to elements of the hidden chat composites.
* Fix some bugs in the migrated tests where the tests weren't actually waiting for UI updates in some cases.
* Fix a source of flakiness where tests weren't actually waiting for fonts to load before taking screenshot.
* Improve `waitForAndHideTypingIndicator` so that it doesn't introduce a delay in tests.


packages/react-composites/tests/browser/common/chatTestHelpers.ts
```diff
@@ -86,25 +87,57 @@ export const waitForMessageWithContent = async (page: Page, messageContent: stri
...
+  });
+  // Now wait for the indicator to disappear.
+  await indicator.waitFor({
+    state: 'hidden',
+    timeout: perStepLocalTimeout()
+  });
 };
 
 export const waitForNSeenMessages = async (page: Page, n: number): Promise<void> => {
+  return waitForNOf(page, n, '[data-ui-status="seen"]', 'body');
```
prprabhu-ms, line 119: Note that before this PR, `waitForNSeenMessages` was simply broken: It would always return a `Promise`, which is truthy.

So this async function never really waited for the N messages to be seen.
In fact, its use was also broken -- it was never possible for the N messages to be seen with the chat pane closed.

So... two wrongs make a (flaky!) right. We just kept getting lucky that executing this selector in the browser took long enough that the messages were actually delivered.

packages/react-composites/tests/browser/callwithchat/hermetic/Chat.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    await waitForSelector(page, dataUiId('call-with-chat-composite-chat-pane'));
+    await sendMessage(page, 'Call with Chat composite is awesome!');
+
+    await temporarilyShowHiddenChatComposite(page, chatRemoteParticipant);
+    await waitForMessageSeen(page);
+
+    await sendMessageFromHiddenChatComposite(page, chatRemoteParticipant, 'I agree!');
+
+    // Local participant has both a sent message and a received message.
+    await waitForNMessages(page, 2, '#test-app-root');
```
prprabhu-ms, line 45: Introduce a constant.

packages/react-composites/tests/browser/common/utils.ts
```diff
@@ -112,7 +112,7 @@ export const waitForPageFontsLoaded = async (page: Page): Promise<void> => {
     //     interface Document {}
     // this breaks `tsc`, even though it works correctly in the browser.
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    await (document as any).fonts.ready;
+    return await (document as any).fonts.ready;
```
prprabhu-ms, line 115: Before this PR, this always returned a fulfilled promise that was always truthy.
So the tests did not actually wait for fonts to load.

dmceachernmsft (approved):

packages/react-composites/tests/app/callwithchat/BaseApp.tsx
```diff
@@ -17,7 +17,7 @@ export function BaseApp(props: { queryArgs: CommonQueryArgs; adapter?: CallWithC
     return <h3>Initializing call-with-chat adapter...</h3>;
   }
   return (
-    <div style={{ position: 'fixed', width: '100%', height: '100%' }}>
+    <div id="test-app-root" style={{ position: 'fixed', width: '100%', height: '100%' }}>
```
JamesBurnside, line 20: _slight_ misnomer as the hermetic app root actually has hidden composites as well as this base app
```suggestion
    <div id="base-app-root" style={{ position: 'fixed', width: '100%', height: '100%' }}>
```
?

packages/react-composites/tests/app/callwithchat/index.tsx
```diff
@@ -8,8 +8,9 @@ import { HermeticApp, shouldLoadHermeticApp } from './HermeticApp';
 import { LiveApp } from './LiveApp';
 import { parseQueryArgs } from './QueryArgs';
 
-const queryArgs = parseQueryArgs();
 initializeIconsForUITests();
+const queryArgs = parseQueryArgs();
+console.log(`Parsed args for test app: ${JSON.stringify(queryArgs)}`);
```
JamesBurnside, line 13: Are we logging anything we shouldn't here? OK if we log a token?

prprabhu-ms: Thanks for catching this. No I don't think we should log a token. Will remove.

packages/react-composites/tests/app/callwithchat/BaseApp.tsx
prprabhu-ms: Still a misnomer then. What I want to say is that this be the app you're trying to test.

The HermeticApp / BaseApp split is implementation detail.

Perhaps `app-under-test-root` captures it best?

JamesBurnside: Ahh I think just leave it as is then

JamesBurnside (approved):
Hard to see the delta between the deleted and added but looks good to me!

PR #2302, opened 2022-09-02
Migrate `CallWithChatComposite` browser tests for More Drawer to be hermetic

prprabhu-ms:
This PR introduces the first instance of some non-static behavior in `MockCallAdapter`.
Device selection now works "as expected", when a known device is selected, the internal state is updated and a state change event is fired.

Note that other events that may be fired by the headless calling SDK when device selection changes are not fired by `MockCallAdapter`, but our UI does not depend on those events.

---

Beyond migration to hermetic tests:

* Merge some tests of device selection to balance test runtime vs how focussed the tests are.
  * Opening the device selection menu, selecting a new device and then validating the selection are now all one test instead of three.


dmceachernmsft (approved):

JamesBurnside:
lgtm, waiting to approve based on ui snapshots succeeding

prprabhu-ms:
verification: eyeballed the new snapshots

anjulgarg (approved):

PR #2303, opened 2022-09-06
Hide control bar buttons based on Rooms role permissions

mgamis-msft:
# What

Hide control bar buttons based on Rooms role permissions instead of disabling them. These were changes requested by Alex Pereira before public preview.

Will add ControlBar unit tests after. It is surprisingly challenging.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2972689
https://skype.visualstudio.com/SPOOL/_workitems/edit/2972694
https://skype.visualstudio.com/SPOOL/_workitems/edit/2972697
https://skype.visualstudio.com/SPOOL/_workitems/edit/2972703
https://skype.visualstudio.com/SPOOL/_workitems/edit/2974319

# How Tested

Ran local sample

# Process & policy checklist


PorterNan (approved):
Thanks for making this change in such a velocity! There are a concern around here I wanna discuss:
1. Since we are moving rendering responsibility from component itself to composite, we removed a bunch of "disable" code - since we don't provide support for component level, this is fine. When we support component level things - what if a developer still render a component under an unsupported role? 

My answer is still, we need to gray the component out since we are going to have role info from stateful client and selector, it should not be a concern for a long run(we will disable these component in selector) - but maybe we can consider reserve these gray code out code until then?

Would like to hear voices from @alkwa-msft @prabhjot-msft as well

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -60,7 +62,7 @@ export type CallControlsProps = {
 const controlBarStyles = memoizeFunction((background: string) => ({ root: { background: background } }));
 
 /**
- * @private
+ * @public
```
PorterNan, line 65: CallControl isn't an exposed api, why do we have a jsdoc change here?

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -38,7 +40,7 @@ import { useAdapter } from '../adapter/CallAdapterProvider';
 import { isDisabled } from '../utils';
 
 /**
- * @private
+ * @public
```
PorterNan, line 43: Same here, revert back to private

PorterNan (commented):
Looks like we still need remove ellipsis button from people pane - is this covered in another work item?

mgamis-msft:
> Looks like we still need remove ellipsis button from people pane - is this covered in another work item?



> Thanks for making this change in such a velocity! There are a concern around here I wanna discuss:
> 
> 1. Since we are moving rendering responsibility from component itself to composite, we removed a bunch of "disable" code - since we don't provide support for component level, this is fine. When we support component level things - what if a developer still render a component under an unsupported role?
> 
> My answer is still, we need to gray the component out since we are going to have role info from stateful client and selector, it should not be a concern for a long run(we will disable these component in selector) - but maybe we can consider reserve these gray code out code until then?
> 
> Would like to hear voices from @alkwa-msft @prabhjot-msft as well

It's easy enough to keep the components grayed out by keeping them hooked to the Permissions context. Interested in what @alkwa-msft @prabhjot-msft think as well.

prabhjot-msft (approved):
Looks good to me! I agree to Nan's point even if we are hiding the components, we should still considering to disable components based on Permissions if user decides to use components instead of composites.

PR #2304, opened 2022-09-06
[Call Readiness] Add permissions api for call readiness

edwardlee-msft:
# What
Add permissions api for Call Composite and CallWithChat Composite with its corresponding unit tests.
Depending on permission level ('required', 'optional', 'doNotPrompt'), user will be prompted for permissions and will be allowed to join/start a call depending on permission level.

# Why
In preparation for Call Readiness
https://skype.visualstudio.com/SPOOL/_workitems/edit/2966640
https://skype.visualstudio.com/SPOOL/_workitems/edit/2974875

# How Tested
Tested using units and on locally ran samples. 
Example:
Setting permissions for camera and microphone to "doNotPrompt" allows user to start call.
image

# Process & policy checklist


common/config/babel/.babelrc.js
```diff
@@ -34,7 +34,9 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         // dialpad
         'dialpad',
         // rooms
-        'rooms'
+        'rooms',
+        // Call Readiness
```
prprabhu-ms, line 38: nit: delete the comment. It doesn't add anything that the well named flag doesn't say already.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -393,6 +393,7 @@ export type CallCompositeIcons = {
 export type CallCompositeOptions = {
     errorBar?: boolean;
     callControls?: boolean | CallControlOptions;
+    permissions?: CallPermissionOptions;
```
prprabhu-ms, line 396: Naming: this should be more specific. There are many permissions when making a call: ACS scopes, file upload/download if that is supported etc.

See naming discussion for the type below. The field name should follow closely.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -589,6 +590,12 @@ export type CallParticipantsLocator = {
     participantIDs: string[];
 };
 
+// @beta
+export type CallPermissionOptions = {
```
prprabhu-ms, line 594: Two naming comments:

* Similar to above, `CallPermission` is too generic. There are multiple types of permissions in the context of a call.
  * Also, avoid the suffix `Options`. Many optional structs are... options. And this one sits inside a `CallCompositeOptions` already.
* It might also be confusing that there are multiple layers that require / check for device permissions. Headless calling already has two related types: `PermissionsConstraint` and `DeviceAccess`.

How about

```typescript
interface DevicePermissionPrompts {
  camera: 'promptAndBlock' | 'prompt' | 'doNotPrompt';
  microphone: 'promptAndBlock' | 'prompt' | 'doNotPrompt';
};
```

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -64,6 +64,29 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
...
+ * Permission options
+ *
+ * @beta
+ */
+export interface DevicePermissionPrompts {
+  /**
+   * Camera Permission prompts for your call.
+   * 'required' - requires the permission to be allowed before permitting the user join the call.
+   * 'optional' - permission can be disallowed and the user is still permitted to join the call.
+   * 'doNotPrompt' - permission is not required and the user is not prompted to allow the permission.
```
dmceachernmsft, line 78: Question about `doNotPrompt`, what happens with the Calling SDK here? it will still ask for permissions so the browser window will still show if not blocked by the OS. does this just hide the devices in the drop downs?

edwardlee-msft: The device permissions here are tied with browser prompts for Microphone and Camera.
The idea is to either have prompts show up when 'required' or 'optional' is set and to hide/do not show prompts when 'doNotPrompt' is set.

You are correct, that the Calling sdk does mandate for at least one device permission (camera or microphone). 
I will be following up during Calling Office hours to address this particular issue. 
image

dmceachernmsft: so right now the calling SDK will ask for device permissions on first load of the application, how does this play into that?
this little prompt: 
image

dmceachernmsft (approved):

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1680,6 +1682,12 @@ export type DeviceManagerState = {
     unparentedViews: LocalVideoStreamState[];
 };
 
+// @beta
+export interface DevicePermissionPrompts {
```
JamesBurnside, line 1686: I like having device permission in the name! `Prompts` however isn't quite right - will be block joining a call if the permission is required regardless of any prompt activity.. although they do control if prompts are shown
Perhaps `RequiredUserDevicePermissions` or `DevicePermissionRestrictions`?

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -291,6 +291,7 @@ export type CallCompositeIcons = {
 export type CallCompositeOptions = {
     errorBar?: boolean;
     callControls?: boolean | CallControlOptions;
+    devicePermissions?: DevicePermissionPrompts;
```
JamesBurnside, line 294: add doc comments to either this or the interface declaration

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -64,6 +64,30 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
...
+   */
+  camera: 'required' | 'optional' | 'doNotPrompt';
+  /**
+   * Microphone permission prompts for your call.
+   * 'required' - requires the permission to be allowed before permitting the user join the call.
+   * 'optional' - permission can be disallowed and the user is still permitted to join the call.
+   * 'doNotPrompt' - permission is not required and the user is not prompted to allow the permission.
+   */
+  microphone: 'required' | 'optional' | 'doNotPrompt';
```
JamesBurnside, line 88: I guess remove this based on calling SDK limitation?

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
edwardlee-msft: Technically this microphone device permission still "works" with "askDevicePermission".
askDevicePermission asks for camera and microphone independently of what the call type actually requires.
So you can essentially prompt for devices in any page outside of a call using askDevicePermission. But if microphone was not "allowed" then when starting/joining a call will prompt for Microphone regardless.

PR #2309, opened 2022-09-07
Update automatically filed issues to have a link to the failing run

JamesBurnside:
# What
Update issues filed as part of CI failures to have a link to the failing run

# Why
So it's a one-click link to start the investigation instead of trawling through the github action runs UI

# How Tested
See test here: https://github.com/Azure/communication-ui-library/issues/2308 (forced the issue to always be created, but verified that the link worked)


edwardlee-msft (approved):

prprabhu-ms (approved):

prprabhu-ms:
Removing from attention set.

PR #2315, opened 2022-09-08
Better concurrent snapshot updates via GitHub action

prprabhu-ms:
This PR includes several improvements to the workflow to update browser test snapshots.

- Merge changes from the target branch before pushing new snapshots.
  - This allows multiple jobs from the workflow to effectively update unrelated snapshots concurrently.
- Change how the workflow is triggered: We still use a PR label (used to be _ui change_, now modified to _update\_snapshots_ in order to highlight the changed behavior), but the label is cleared by the workflow at the outset. When a developer sets the label on a PR, it triggers the workflow run exactly once, instead of on each commit to the branch.
- Split `@internal/react-composite` UI snapshot updates for different composites into different jobs.

## Impact

- ✅ Typical time to update snapshots for both beta and stable build reduced from ~1 hour to ~15 minutes 🚀
- ✅ Unrelated changes to the PR while snapshots are being updated no longer cancel (and hence delay) snapshot update workflow run.
- 👀 Developers must request snapshots to be updated each time it is needed on a PR instead of adding the _ui change_ label once which ended up triggering a lot of wasted snapshot update workflow runs.

---

### Validation

After this PR, this [example workflow run](https://github.com/Azure/communication-ui-library/actions/runs/3019687738) ran in under 15 minutes and pushed many snapshot updates concurrently: 

image

Before this PR, a [typical PR](https://github.com/Azure/communication-ui-library/pull/2301) that updated both stable and beta composite snapshots required two workflow runs ([1](https://github.com/Azure/communication-ui-library/actions/runs/2981231101), [2](https://github.com/Azure/communication-ui-library/actions/runs/2981374064)), each taking over 25 minutes for a total of nearly 1 hour before snapshot updates were done. In the meantime, any update to the PR would have canceled an restarted snapshot updates.


.github/workflows/deprecated-update-snapshot.yml
```diff
@@ -1,4 +1,7 @@
-name: Update snapshots
+# This workflow is deprecated. Use ./update-snapshots.yml instead.
```
prprabhu-ms, line 1: Keeping the old workflow around until I've proven that the updated one works properly.

Some aspects of the rewrite are impossible to test before merging this PR. I tried out the label-deletion logic in a throaway repo, so I have reasonable confidence, but still.

.github/workflows/update-snapshots.yml
```diff
@@ -0,0 +1,506 @@
...
+  precondition:
+    name: Determine if this workflow is applicable
+    runs-on: ubuntu-latest
+    outputs:
+      met: ${{ steps.label.outputs.found || steps.workflow_dispatch.outputs.found }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          # Use a machine account when checking out. This is to workaround the issue were GitHub
```
edwardlee-msft, line 23: "were" to "where"

.github/workflows/update-snapshots.yml
```diff
@@ -0,0 +1,506 @@
...
+          # actions, when using the default account, cannot trigger other actions.
+          # This machine account is only for this PAT, pwd was created and thrown away
+          # If any update needed, create a new account, add access to the repo and generate a new PAT
```
edwardlee-msft, line 26: "If any update needed" to "If an update is needed"

packages/react-composites/tests/README.md
```diff
@@ -83,7 +83,11 @@ node scripts/runBrowserTests.mjs -c call -s 10
 
 Snapshots must only be updated as part of the CI pipeline. This is because the build agents may run on a different OS and have different graphics setup that your local machine (so there will be slight pixel differences between locally generated snapshots and ones generated by the CI pipeline).
 
-To update snapshots, add the `ui-change` label to your PR. A GitHub action will generate new snapshots and push commit to your branch.
+To update snapshots,
+
+- Add the `update_snapshots` label to your PR.
+- The [update-snapshots.yml](https://github.com/Azure/communication-ui-library/actions/workflows/update-snapshots.yml) GitHub action will generate new snapshots and push commit to your branch.
+  - The workflow will remove the `update_snapshots` label from your PR. If you need to update snapshots again, you must add the label once again.
```
JamesBurnside, line 90: Oh nice! great solution

.github/workflows/update-snapshots.yml
```diff
@@ -0,0 +1,506 @@
...
+    if: needs.precondition.outputs.met
+    name: Update component examples snapshots
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: ${{ fromJSON(needs.get_matrix.outputs.matrix) }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          # Use a machine account when checking out. This is to workaround the issue were GitHub
```
edwardlee-msft, line 71: Similar comments to the my other comment.

.github/workflows/update-snapshots.yml
```diff
@@ -0,0 +1,506 @@
...
+        # Thus we do not expect any merge conflicts due to the concurrent jobs from this workflows.
+        #
+        # If the UI snapshots are updated in the upstream branch while this job was running
+        # (e.g., due to a merge from the base_ref for a PR), this merge will fail, and the
+        # workflow will have to be triggered again. This is the desired behavior because the workflow
+        # should not silently overwrite updates to the target branch.
+        run: |
+          git add packages/react-composites/*.png
+          git commit -m 'Update packages/react-composites CallWithChatComposite browser test snapshots'
+          git pull origin ${{ needs.get_target_branch.outputs.target }} --no-rebase --no-edit
```
JamesBurnside, line 505: instead of commit-pull-push on each action (triggering PR updates each time), another option would be to commit, then have a final job that depends on all the children that does the push

edwardlee-msft (approved):

.github/workflows/update-snapshots.yml
prprabhu-ms: Each job runs on a separate node. A single push would require us to use GitHub artifacts API to share commits around. I felt that using git directly was a simpler solution -- to implement, understand and maintain in the future.

The more technologies we bring into a workflow, the more the points of failure as well.

prprabhu-ms: To this point though, my approach does introduce an opportunity for flake if multiple jobs finish at the same time.

And at scale, all flake happens frequently enough.

Let me explore the use of artifacts API once as well.

prprabhu-ms: Here's an exploration of using artifacts to synchronize patches: https://github.com/prprabhu-ms/scratch-actions/blob/main/.github/workflows/sync.yml

It works well, but It'll add ~150 lines of code to an already complex workflow. 

I think it's worth merging this PR as is for now. If we find that we actually see conflicts / flaky failures due to the concurrent jobs trying to update changes at the same time, we can add the extra complexity.

wdyt?

.github/workflows/ci.yml
```diff
@@ -168,7 +168,7 @@ jobs:
           name: snapshots
           path: packages/react-composites/test-results
       - name: Comment on PR
-        if: ${{ github.event_name == 'pull_request' && always() && steps.visualregressiontests.outcome == 'failure' && !contains( github.event.pull_request.labels.*.name, 'ui change') }}
+        if: ${{ github.event_name == 'pull_request' && always() && steps.visualregressiontests.outcome == 'failure' && !contains( github.event.pull_request.labels.*.name, 'update_snapshots') }}
```
JamesBurnside, line 171: To reduce re-education on updating snapshots can we keep the old label name? (`ui change`)

JamesBurnside (approved):

.github/workflows/update-snapshots.yml
JamesBurnside: Oh yeah good point on the separate job nodes

JamesBurnside: And yeah I see the artifact method will definitely add even more complexity to an already enormous yml file... gosh yml needs a *proper* way to have re-usable local code

.github/workflows/ci.yml
prprabhu-ms: I intentionally chose a new label for two reasons:

* Allows me to introduce this new workflow without affecting the old one - gentle transition
* I want to reeducate folks. If I keep the same label, folks might not notice that anything changed, and may be confused when the label auto-disappears from their PRs.

PR #2319, opened 2022-09-09
Show/hide menu button on hover/click/focus

carocao-msft:
# What
Show/hide menu button on hover/click/focus

# Why
accessibility users can now see menu button shown when focused on the participant item

# How Tested

https://user-images.githubusercontent.com/96077406/189247652-b1c00803-e326-4651-8045-5b561d79a8e8.mov


JamesBurnside (commented):
Change looks good! (Holding off approving until CI is green)

JamesBurnside:
Question to ask Alex, should the 3 dots still be showing when the context menu is open?
image

carocao-msft:
> Question to ask AlexP, should the 3 dots still be showing when the context menu is open? image

There might be some flakiness here, since when I was testing the 3 dots were present when menu is open. But still going to verify with Alex P if we should keep the 3 dots

JamesBurnside:
> There might be some flakiness here, since when I was testing the 3 dots were present when menu is open. But still going to verify with Alex P if we should keep the 3 dots

When testing, are the 3 dots still showing when you click on a participant (so flyout opens) then move the mouse away (without clicking anywhere, just so it is no longer hovering over the participant or participant menu).
Just assumption is that the 3 dots should always be showing when the participantMenu is open even when the mouse is not hovering over it

carocao-msft:
> > There might be some flakiness here, since when I was testing the 3 dots were present when menu is open. But still going to verify with Alex P if we should keep the 3 dots
> 
> When testing, are the 3 dots still showing when you click on a participant (so flyout opens) then move the mouse away (without clicking anywhere, just so it is no longer hovering over the participant or participant menu). Just assumption is that the 3 dots should always be showing when the participantMenu is open even when the mouse is not hovering over it

Let me verify with Alex P since before this change, the ellipses hover effect disappears (we show bold ellipses when it's hovered/clicked before) when the mouse is no longer hovering over the participant/participant menu

carocao-msft:
> > There might be some flakiness here, since when I was testing the 3 dots were present when menu is open. But still going to verify with Alex P if we should keep the 3 dots
> 
> When testing, are the 3 dots still showing when you click on a participant (so flyout opens) then move the mouse away (without clicking anywhere, just so it is no longer hovering over the participant or participant menu). Just assumption is that the 3 dots should always be showing when the participantMenu is open even when the mouse is not hovering over it

Verified with Alex P and made sure ellipses shown when mouse is hovered away :)

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -206,13 +206,10 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
         title={strings.menuTitle}
         data-ui-id={ids.participantItemMenuButton}
       >
-        <Icon
-          iconName={itemHovered ? 'ParticipantItemOptionsHovered' : 'ParticipantItemOptions'}
-          className={iconStyles}
-        />
+        <Icon iconName={'ParticipantItemOptionsHovered'} className={iconStyles} id="menuButtonIcon" />
```
JamesBurnside, line 209: can we avoid using id? [id](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) should be "unique across the global document" which we can never garuntee.
If we need to use `id` here use fluent's `useId()` hook to generate a more-likely-to-be unique id

carocao-msft: OK let me look into that! I'm using id here because using svg tag will affect the css of other icons in participant item.

carocao-msft: Fixed by changing id to classname :) Thanks for the suggestion!

JamesBurnside (approved):

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -206,13 +206,10 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
         title={strings.menuTitle}
         data-ui-id={ids.participantItemMenuButton}
       >
-        <Icon
-          iconName={itemHovered ? 'ParticipantItemOptionsHovered' : 'ParticipantItemOptions'}
```
anjulgarg, line 210: Do we need to remove the declarations of `ParticipantItemOptions` icon now that it is no longer being used?

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -229,7 +226,11 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
       data-is-focusable={true}
       data-ui-id="participant-item"
       className={mergeStyles(
-        participantItemContainerStyle({ localparticipant: me, clickable: !!menuItems }),
+        participantItemContainerStyle({
+          localparticipant: me,
+          clickable: !!menuItems,
+          showMenuButton: itemHovered || !menuHidden
```
anjulgarg, line 232: Why can't we hide the `menuButton` declared below instead of using class names in styles?

anjulgarg (approved):

packages/react-composites/tests/app/call/BaseApp.tsx
```diff
@@ -31,13 +32,16 @@ export function BaseApp(props: { queryArgs: QueryArgs; callAdapter?: CallAdapter
       'Some details about the call that span more than one line - many, many lines in fact. Who would want fewer lines than many, many lines? Could you even imagine?! 😲';
   }
 
+  const ParticipantItemOptions = queryArgs.showParticipantItemIcon ? <MoreHorizontal20Regular /> : <></>;
```
JamesBurnside, line 35: wow nice love to see tests for obscure, easily-forgetten things!

JamesBurnside (changes requested):
This looks like a regression in these UI tests:
image

packages/react-composites/tests/app/call/BaseApp.tsx
JamesBurnside: Should these new snapshots have the participant menu showing?
image

carocao-msft:
> This looks like a regression in these UI tests: image

Those disappeared because we change the ellipses icon to <></>

packages/react-composites/tests/app/call/BaseApp.tsx
carocao-msft: yes they should! let me update this

JamesBurnside (approved):

PR #2324, opened 2022-09-09
Add UI test for custom composite call control options

JamesBurnside:
# What
* Add a UI test for setting specific call and callwithchat composite call control options
* Update callwithchat hermetic tests to support qArgs

# Why
Missing test area, came across a bug when locally testing: #2325 

# How Tested
n/a - see updated snapshots


packages/react-composites/tests/browser/call/hermetic/CallControls.test.ts
```diff
@@ -31,3 +32,29 @@ test.describe('Call composite custom button injection tests', () => {
...
+    serverUrl
+  }) => {
+    const initialState = defaultMockCallAdapterState([defaultMockRemoteParticipant('Paul Bridges')]);
+    const testOptions: CallCompositeOptions = {
+      callControls: {
+        displayType: 'compact',
+        cameraButton: {
+          disabled: true
+        },
+        microphoneButton: true,
```
prabhjot-msft, line 48: What does `microphoneButton: true` signify as the microphone is disabled as well in the snapshots

packages/react-composites/tests/browser/callwithchat/hermetic/CallControls.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  }) => {
+    const paul = defaultMockRemoteParticipant('Paul Bridges');
+    const testOptions: CallWithChatCompositeOptions = {
+      callControls: {
+        displayType: 'compact',
+        cameraButton: {
+          disabled: true
+        },
+        endCallButton: false,
+        microphoneButton: true,
```
prabhjot-msft, line 23: same as above

packages/react-composites/tests/app/call/BaseApp.tsx
```diff
@@ -46,7 +46,9 @@ export function BaseApp(props: { queryArgs: QueryArgs; callAdapter?: CallAdapter
                 queryArgs.injectParticipantMenuItems ? onFetchParticipantMenuItems : undefined
               }
               options={
-                queryArgs.injectCustomButtons
+                queryArgs.customCallCompositeOptions
+                  ? queryArgs.customCallCompositeOptions
+                  : queryArgs.injectCustomButtons
```
mgamis-msft, line 51: Not part of this PR but can we remove `injectCustomButtons` query arg. It seems like a less flexible version of `customCallCompositeOptions`.

mgamis-msft (approved):

packages/react-composites/tests/browser/call/hermetic/CallControls.test.ts
JamesBurnside: They are simply valid options so including them in the test, we actually have a bug where setting it to true was causing it to become disabled: #2325

prabhjot-msft (approved):

PR #2325, opened 2022-09-09
Fix control bar controls showing as disabled when the button is set to `true`

JamesBurnside:
# What
* Fix `isDisabled` fn logic

# Why
* Logic bug where we are incorrectly disabling a call control bar button when set to true

# How Tested
Locally and snapshots should update once #2324 is completed


packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -210,14 +210,16 @@ export const disableCallControls = (
...
+ * A button is only disabled if is explicitly set to disabled.
  *
  * @param option
  * @returns whether a button is disabled
  * @private
  */
-export const isDisabled = (option?: boolean | { disabled: boolean }): boolean => {
-  if (typeof option !== 'boolean') {
-    return !!option?.disabled;
+export const isDisabled = (option: boolean | { disabled: boolean } | undefined): boolean => {
```
edwardlee-msft, line 219: Ah good find! Thanks for addressing this :)

edwardlee-msft (approved):

dmceachernmsft (approved):

PR #2333, opened 2022-09-13
Update storybook to contain docs on customizing call controls in call composite

anjulgarg:
# What
Update storybook to contain docs on customizing call controls in call composite

Screen Shot 2022-09-13 at 1 56 39 PM

# Why

# How Tested

# Process & policy checklist


packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
```diff
@@ -182,6 +182,26 @@ export const getDocs: () => JSX.Element = () => {
         accepts an object where you can provide a custom JSX element against an icon name key.
       </Description>
 
+      <Heading>Customize Call Controls</Heading>
```
JamesBurnside, line 185: This is great! Should we also add to the callwithchat composite docs also?

JamesBurnside (approved):

packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
dmceachernmsft: maybe just a extension about the two extra buttons linking back to this page?

dmceachernmsft (approved):

packages/storybook/stories/CallComposite/CallCompositeDocs.tsx
anjulgarg: Don e

PR #2334, opened 2022-09-14
[Call Readiness] Unsupported Browser UI

dmceachernmsft:
# What

Introduces new UI to alert the user they are using a unsupported browser.
# Why

Allows Contoso to provide a way to inform the user they are using a unsupported browser and direct them to documentation to be provided by Contoso letting the user know which browsers to use.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2980682
# How Tested

image
image
Validated UI and styles in storybook.


JamesBurnside (approved):

packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -84,6 +86,9 @@ export interface ComponentStrings {
   /* @conditional-compile-remove(one-to-n-calling) */ /* @condtional-compile-remove(one-to-n-calling) */
   /** Strings for HoldButton */
   holdButton: HoldButtonStrings;
+  /* @conditional-compile-remove(call-readiness) */ /* @conditional-compile-remove(unsupported-browser) */
```
JamesBurnside, line 89: Before when two features were above a conditional compile they implied that id _either_ feature1 _or_ feature2 was stabilized we want to enabled this feature. But in this case we only want unsupported-browser to enable this feature so remove the `/* @conditional-compile-remove(call-readiness) */` flag if possible

packages/react-components/src/components/UnsupportedBrowser.tsx
```diff
@@ -0,0 +1,77 @@
...
+  /** Handler to perform a action when the help link is actioned */
+  onTroubleShootingClick: () => void;
+  /** String overrides for the component */
+  strings: UnsupportedBrowserStrings;
+}
+
+/* @conditional-compile-remove(unsupported-browser) */
+const UnsupportedBrowserContainer = (props: UnsupportedBrowserProps): JSX.Element => {
+  const { onTroubleShootingClick, strings } = props;
+  console.log('test2');
```
edwardlee-msft, line 48: Sneaky console log sneaking it's way in

packages/react-components/src/components/UnsupportedBrowser.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <Stack styles={containerStyles}>
+      <Icon iconName="UnsupportedBrowserWarning"></Icon>
```
edwardlee-msft, line 51: Icon and text seems a bit too close. 
When looking at the other figma mobile designs, the icon seems to be further up

edwardlee-msft (approved):

PR #2338, opened 2022-09-14
Remove 'Add people' from all roles in Rooms call

prabhjot-msft:
# What
Remove 'Add people' from all roles in Rooms call

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2981209

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -65,7 +65,9 @@
     "resumingCallButtonAriaLabel": "Resume call",
     "holdScreenLabel": "You're on hold",
     "openDtmfDialpadLabel": "Show dialpad",
-    "dtmfDialpadPlaceHolderText": "Enter number"
+    "dtmfDialpadPlaceHolderText": "Enter number",
+    "roomsConfigurationPageTitle": "Join a call",
```
alkwa-msft, line 69: Should we use "Room" instead of "Call"?

packages/react-composites/src/composites/common/PeoplePaneContent.tsx
```diff
@@ -45,6 +46,7 @@ export const PeoplePaneContent = (props: {
 }): JSX.Element => {
   const { inviteLink, onFetchParticipantMenuItems, setDrawerMenuItems, strings, onRemoveParticipant } = props;
   const participantListDefaultProps = usePropsFor(ParticipantList);
+  const isRoomsCall = 'roomId' in useAdapter().getState()['locator'];
```
alkwa-msft, line 49: This seems like it would be a great util function

alkwa-msft (commented):
I see we are checking to see if we are in a rooms-based call in a lot of places. Is it possible to extract it out as a util function?

mgamis-msft:
Closing this PR. Bug ticket https://skype.visualstudio.com/SPOOL/_workitems/edit/2981209 was closed with PR https://github.com/Azure/communication-ui-library/pull/2370.

PR #2342, opened 2022-09-14
Small Refactor: move call composite side pane state handling to its own hook

JamesBurnside:
# What
Instead of defining useState and all the handlers in the (already big) Call Arrangement file, move to its own useState-style hook file

# Why
Move business logic out of the view code. If we have more complex side pane content this file will grow even more large

# How Tested
Storybook, people pane showed and closed as expected in 1-n call:
image


packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -16,17 +16,14 @@ import {
...
 import { AvatarPersonaDataCallback } from '../../common/AvatarPersona';
 import { containerDivStyles } from '../../common/ContainerRectProps';
 /* @conditional-compile-remove(one-to-n-calling) */
 import { useAdapter } from '../adapter/CallAdapterProvider';
 import { CallControls, CallControlsProps } from '../components/CallControls';
-/* @conditional-compile-remove(one-to-n-calling) */
-import { useSelector } from '../hooks/useSelector';
-/* @conditional-compile-remove(one-to-n-calling) */
-import { callStatusSelector } from '../selectors/callStatusSelector';
+import { useSidePaneState } from '../hooks/useSidePaneState';
```
edwardlee-msft, line 26: Should this be conditionally compiled as well?

JamesBurnside: indeed! thanks

edwardlee-msft: This should be conditionally compiled for one-to-n-calling

JamesBurnside: 🤦 - updated!

JamesBurnside: (my secret plan next PR is to remove the fact we have two people panes and make callwithchat hook into this so would remove the cc line, but also involves doing controversial things like creating an imperative handle to the call composite to control the pane from the callwithchat composite 👀)

edwardlee-msft (approved):
Love the extraction!!

dmceachernmsft (approved):

anjulgarg (approved):

PR #2345, opened 2022-09-15
Device Permission and network connection error bar 

carocao-msft:
# What
Device Permission and network connection error bar 

# Why
[USER STORY 2957783](https://skype.visualstudio.com/SPOOL/_workitems/edit/2957783)
[USER STORY 2957781](https://skype.visualstudio.com/SPOOL/_workitems/edit/2957781)

# How Tested
Storybook internal story 

Screen Shot 2022-09-14 at 11 51 26 PM


packages/react-components/src/components/utils.ts
```diff
@@ -19,3 +22,158 @@ export const truncatedFileName = (fileName: string, length: number): string => {
...
+      type: toDismiss.type,
+      dismissedAt: now,
+      activeSince: toDismiss.timestamp
+    }
+  ];
+};
+
+/**
+ * @private
+ * @param ctiveErrorMessages
```
dmceachernmsft, line 66: nit: ctive => active

packages/storybook/stories/INTERNAL/TroubleShootingGuideErrorBar/TroubleShootingGuideErrorBar.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+        onNetworkingTroubleShootingClick={onNetworkingTroubleShootingClick}
+        permissionsState={permissionsState}
+        troubleShootingGuideStrings={networkTroubleShootingGuideStrings}
+      />
+    </div>
+  );
+};
+
+// This must be the only named export from this module, and must be named to match the storybook path suffix.
+// This ensures that storybook hoists the story instead of creating a folder with a single entry.
```
dmceachernmsft, line 72: Should this story go in the internal CallReadiness folder introduced by the DomainPermissions component? (that PR snuck in yesterday)

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
```diff
@@ -0,0 +1,150 @@
...
+  // Without this behaviour, this `useEffect` block would cause a render loop.
+  useEffect(
+    () => setDismissedErrors(dropDismissalsForInactiveErrors(props.activeErrorMessages, dismissedErrors)),
+    [props.activeErrorMessages, dismissedErrors]
+  );
+
+  const toShow = errorsToShow(props.activeErrorMessages, dismissedErrors);
+
+  return (
+    <Stack data-ui-id="error-bar-stack">
```
JamesBurnside, line 107: Can this use the existing errorBar? And do something like:
```tsx
<ErrorBar onTroubleShootingClick={() => etc...} linkText={troubleShootingGuideStrings.linkText} ...otherProps />
```

These new props could be added as a beta property, or have an <ErrorBarInner /> component such that it has these extra props and errorbar is simply a wrapper around it:
```tsx
const ErrorBar = (props: ErrorBarProps) => <ErrorBarInner {...props} />;
const _TroubleShootingGuideErrorBar = (props) => <ErrorBarInner {...props} onTroubleShootingClick={() => etc...} linkText={troubleShootingGuideStrings.linkText} />;
```

packages/react-components/src/components/styles/TroubleShootingGuideErrorBar.styles.ts
```diff
@@ -0,0 +1,54 @@
...
+    color: theme.palette.black
+  }
+});
+
+/**
+ * @private
+ */
+export const messageBarStyle = (theme: ITheme, errorType: number): IMessageBarStyles => ({
+  innerText: {
+    paddingTop: errorType === 5 ? '0.15rem' : '0.1rem', // to move the inner text of the message bar down to be centered
```
JamesBurnside, line 27: what is errorType 5?

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
carocao-msft: We can't use the existing one because we need to replace the X button with new OK button

packages/react-components/src/components/styles/TroubleShootingGuideErrorBar.styles.ts
carocao-msft: This is some old code solving a bug for the original Error bar

packages/storybook/stories/INTERNAL/TroubleShootingGuideErrorBar/TroubleShootingGuideErrorBar.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+const TroubleShootingGuideErrorBarStory = (args): JSX.Element => {
+  const theme = useTheme();
+  const onPermissionsTroubleshootingClick = (permissionState: {
+    camera: PermissionState;
+    microphone: PermissionState;
+  }): void => {
+    console.log(permissionState);
+  };
+
+  const onNetworkingTroubleShootingClick = (): void => {
```
JamesBurnside, line 24: troubleshooting is one word, so onNetworkingTroubleShootingClick -> onNetworkingTroubleshootingClick

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
```diff
@@ -0,0 +1,150 @@
...
+   * strings related to trouble shooting guidance link and dismiss button
+   */
+  troubleShootingGuideStrings: _TroubleShootingGuideErrorBarStrings;
+}
+
+/**
+ * @internal
+ * A component to show device Permission/network connection related errors, contains link that leads to further trouble shooting guide
+ */
+export const _TroubleShootingGuideErrorBar = (props: _TroubleShootingGuideErrorBarProps): JSX.Element => {
```
JamesBurnside, line 82: troubleshooting is one word, so _TroubleShootingGuideErrorBar-> _TroubleshootingGuideErrorBar (and filename, props names etc.)

packages/react-components/src/components/styles/TroubleShootingGuideErrorBar.styles.ts
JamesBurnside: Ah I see, without the context of the call to `messageBarType` we can't know what this 5 is anymore, can you update this type from number to `MessageBarType` and update 5 to `MessageBarType.warning`

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
JamesBurnside: Ahh gotcha

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
```diff
@@ -0,0 +1,150 @@
...
+                  onClick={() => {
+                    if (onPermissionsTroubleshootingClick) {
+                      onPermissionsTroubleshootingClick(permissionsState);
+                    } else if (onNetworkingTroubleShootingClick) {
+                      onNetworkingTroubleShootingClick();
+                    }
+                  }}
+                  underline
+                >
+                  <span>{troubleShootingGuideStrings.linkText}</span>
```
dmceachernmsft, line 131: Does this need to be in a span?

change/@internal-react-components-60cee620-62c5-430b-b711-33bc1a4fdc34.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "created error bar to display trouble shooting links for network/device permission errors",
```
edwardlee-msft, line 3: nit: Create error bar to display troubleshooting links for network/device permission errors

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
edwardlee-msft: Is it not possible to have some kind of condition that would show the x or the ok button?
Looking at FluentUI, it looks like there is a prop for "actions" that can be used to present the okay button:
https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar
image

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
carocao-msft: It's in a span so I can apply styling targeted to it, otherwise some how styling does not get applied to innertext

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
carocao-msft: I think we are tyring to avoid changing the error bar API

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
```diff
@@ -0,0 +1,150 @@
...
+                  }}
+                  underline
+                >
+                  <span>{troubleshootingGuideStrings.linkText}</span>
+                </Link>
+              </div>
+            ) : (
+              <div>{strings[error.type]} </div>
+            )}
+            <PrimaryButton
```
edwardlee-msft, line 137: Can we move this into the "actions" prop?
https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar

carocao-msft: fixed that!

packages/react-components/src/components/TroubleShootingGuideErrorBar.tsx
JamesBurnside: To avoid changing the error bar API you would introduce a `<ErrorBarInner />` that ErrorBar, NetworkErrorBar etc. would all use

JamesBurnside (approved):

packages/react-components/src/components/TroubleshootingGuideErrorBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+              styles={confirmButtonStyle(theme)}
+              onClick={() => {
+                setDismissedErrors(dismissError(dismissedErrors, error));
+              }}
+              ariaLabel={strings.dismissButtonAriaLabel}
+            />
+          }
+          isMultiline={false}
+        >
+          {onPermissionsTroubleshootingClick || onNetworkingTroubleshootingClick ? (
```
dmceachernmsft, line 127: what happens here if the `onPermissionsTroubleshootingClick` is set but not the `onNetworkingTroubleshootingClick` and the network error fires? won't the network error bar send the user to the permissions help screen?

packages/react-components/src/components/TroubleshootingGuideErrorBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Props for {@link _TroubleshootingGuideErrorBar}.
+ *
+ * @internal
+ */
+export interface _TroubleshootingGuideErrorBarProps extends ErrorBarProps {
+  /**
+   * permissions state for camera/microphone
+   */
+  permissionsState: {
```
dmceachernmsft, line 37: Should this be required when the function its passed to is a optional property?

packages/react-components/src/components/TroubleshootingGuideErrorBar.tsx
dmceachernmsft: the way I'm reading the logic there though is if there is a error set for the error bar and one of those functions are set you will have the error string of the incorrect error with the function.

dmceachernmsft (approved):

PR #2346, opened 2022-09-15
Configuration page text change for rooms

mgamis-msft:
# What

- Configuration page text changed to 'Join Call' when call is a Rooms call.
- Needed add locator prop to CallContext state

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2981217

# How Tested

Local calling sample
https://user-images.githubusercontent.com/79475487/192650280-bf597d10-5e71-4d61-8d42-8ea3830c6ebd.mp4

# Process & policy checklist


prprabhu-ms (changes requested):
This PR breaks the `CallAdapter` abstraction to reach into `AzureCommunicationCallAdapter` from `CallComposite`. We don't want to do that.

Can you please add more context around why we needed to do this, so that we can find an alternative approach?

mgamis-msft:
> This PR breaks the `CallAdapter` abstraction to reach into `AzureCommunicationCallAdapter` from `CallComposite`. We don't want to do that.
> 
> Can you please add more context around why we needed to do this, so that we can find an alternative approach?

I need to know the call type in the CallComposite code to be able to change the 'Start call' button text to 'Join call' when it is a Rooms call. Knowing the call type is also required in this [PR](https://github.com/Azure/communication-ui-library/pull/2338) to know if the 'Copy Invite Link' button should be hidden. 

The `locator` prop is meant to be temporary added to the CallAdapter context to avoid an API change until the CallingSDK will make the call type available. 

Would it be better to just add in the locator/ callType prop in `CallAdapterClientState` like [alternateCallerId](https://github.com/Azure/communication-ui-library/blob/b3fa8567c4817423343143d1f6fb21d15b0ce1d1/packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts#L86)?

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -488,8 +488,10 @@ export interface CallCompositeStrings {
     returnToCallButtonAriaLabel?: string;
     roomNotFoundDetails?: string;
     roomNotFoundTitle: string;
+    roomsConfigurationPageTitle: string;
```
JamesBurnside, line 491: from the meeting yesterday this still concerns me - Contoso now has two places to set the same text, but only one will be used depending on the type of locator they provide. This could also explode to `n` different places to set the text with `n` different ways that text is chosen if we do the same for PSTN text and future call types.

My feeling is we shouldn't change this at a composite level but instead have the sample change the text per the way the call is joined -- if rooms is used we can say 'join room', if a PSTN call we can say something like 'place call'.

Otherwise, if we do want to change the text at a composite level, we need to have a new localization architecture that supports different call types. One that allows different strings in the same field based on augmentations like call type (but we can't assume only call type will do this, perhaps a new feature will require different string changes). One example could be we have EN_US_ROOMS_OVERRIDE_STRINGS instead of just EN_US and contoso has to do
```
const strings = {...EN_US, ...EN_US_ROOMS_OVERRIDE_STRINGS};
```
 (I don't particularly like this either but just using this as a top-of-my-head example)

JamesBurnside (changes requested):

packages/communication-react/review/beta/communication-react.api.md
mgamis-msft: I see. That makes sense to me to not have multiple ways to change the strings and there is less code change. We can leave it up to Contoso to change the Configuration Page strings because they should know what type of call they are using the CallComposite for. 
Here is the simpler PR https://github.com/Azure/communication-ui-library/pull/2368

alkwa-msft: **My feeling is we shouldn't change this at a composite level but instead have the sample change the text per the way the call is joined -- if rooms is used we can say 'join room', if a PSTN call we can say something like 'place call'.**
>> This helps customers using the sample but what about customers just using the Composite? Are they stuck on place call vs join call and having to localize it themselves (e.g Italian)

**Otherwise, if we do want to change the text at a composite level, we need to have a new localization architecture that supports different call types. One that allows different strings in the same field based on augmentations like call type (but we can't assume only call type will do this, perhaps a new feature will require different string changes). One example could be we have EN_US_ROOMS_OVERRIDE_STRINGS instead of just EN_US and contoso has to do**
>> I was thinking of an another example where we may have multiple strings that are injected into one spot. I like to think about a Calendar app. Based on some condition (time of day) it will say "Good morning" vs "Good afternoon". Both strings are valid, the condition is valid, and customers would like us to localize this instead of loading that effort onto them. 

@JamesBurnside  any thoughts?

JamesBurnside: > Are they stuck having to localize it themselves

They would, so you're correct we _do_ need to provide these localized strings, but we need to evaluate if we should be changing our localization structure instead of adding more props which could get out of hand and very difficult to reverse.

> "Good morning" vs "Good afternoon"

This is a good example of a case-by-case which would make sense - this would translate to: `morningGreeting` and `afternoonGreeting` strings - crucially however _these both appear in the same instance lifetime_, i.e. when the time changes the greeting changes, as a developer I expect to see both of these in my app when I override them.

For us `configurationPageTitle` and `roomsConfigurationPageTitle` will never be used at the same time - that is to say at no point does it make sense for Contoso to supply both of these props into the composite - the composite will only ever use one or the other. Why are we adding it here? Purely to workaround a technical limitation, not for Contoso's benefit (and the public API should really only be for Contoso's benefit), here we add redundancy and complexity to how Contoso uses the locale API.

(We may not have a good solution to avoid it, but as I mention above we should be evaluating alternatives instead adding more hard-to-revert API methods)

JamesBurnside: One possible solution (again just top-of-head, I think we should have a document evaluating solutions) to handling this internally: we have `strings-internal.json`, and we send that for localization like we do our regular strings and then we can keep changing strings based on `callType` internal

mgamis-msft: Really great discussion here. Thank you for your input and consideration of our overall design. We have decided to push this task for GA where I will revisit this PR.

PR #2340, opened 2022-09-14
Devices Icon change to speaker button for consumer

prabhjot-msft:
# What
Devices Icon change to speaker button for consumer

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2981211

# How Tested

# Process & policy checklist

- [ ] I have updated the project documentation to reflect my changes if necessary.
- [ ] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.

**Is this a breaking change?**

- [ ] This change causes current functionality to break.

PR #2347, opened 2022-09-15
[Up for discussion] Re-hook buttons to permissions

mgamis-msft:
# What

Restored old code to hook back the CameraButton, MicrophoneButton, and ScreenShareButton components to permissions so that they are disabled based on role.
Hooked back DeviceButton component to not show devices based on role.

# Why

From this PR comment https://github.com/Azure/communication-ui-library/pull/2303#pullrequestreview-1101505030

# How Tested

Locally

# Process & policy checklist

### DIFF
--- a/change/@internal-react-components-3559264c-7cb3-42d7-9724-db311a1e2481.json
+++ b/change/@internal-react-components-3559264c-7cb3-42d7-9724-db311a1e2481.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Hook permissions to CameraButton, DevicesButton, MicrophoneButton, and ScreenShareButton.",
+  "packageName": "@internal/react-components",
+  "email": "miguelgamis@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/react-components/src/components/CameraButton.test.tsx
+++ b/packages/react-components/src/components/CameraButton.test.tsx
@@ -7,6 +7,12 @@ import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 import { createTestLocale, mountWithLocalization } from './utils/testUtils';
 import { registerIcons } from '@fluentui/react';
+/* @conditional-compile-remove(rooms) */
+import { mountWithPermissions } from './utils/testUtils';
+/* @conditional-compile-remove(rooms) */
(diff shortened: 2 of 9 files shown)


mgamis-msft:
@JamesBurnside  From your comment [here](https://github.com/Azure/communication-ui-library/pull/2340#discussion_r972139283), I think your stance is to NOT proceed with this PR which is valid because components should be as pure as possible. Just wanted your thoughts on the scenario of Contoso building their app with pure components, they would then have enforce role based access control on their own.

JamesBurnside:
> @JamesBurnside From your comment [here](https://github.com/Azure/communication-ui-library/pull/2340#discussion_r972139283), I think your stance would be NOT proceed with this PR which is valid because components should be as pure as possible. Just wanted your thoughts on the scenario of Contoso building their app with pure components, they would then have enforce role based access control on their own.

My feeling is the role-based access control should come in the form of props from the usePropsFor and keep the pure components pure. Given we can't do that yet as role is not in the SDK yet, second best would be to do the same mapping of role->props in the Composite layer until that is ready, then pass in the props to the pure component. However, after chatting with Prabhjot about that comment I feel we are OK to use the `usePermissions` inside the component -- but only with the assumption we will remove it when the usePropsFor is ready.

The larger issue on the previous PR which meant we had to move the logic to composites was an incorrect assumption when using the usePermissions inside the devicesbutton component -- it changed the devices button based on the state of sibling component state. Devices button logic should only be concerned with itself. E.g. It could be possible that Contoso still allows the user to set a different camera and mic in their application even when given a certain role that doesn't permit sharing their audio/video. The aim of UI components is to try not dictate policy but instead allow explicit props to drive UI - then we allow usePropsFor to give awesome default behavior.

alkwa-msft:
In the future when we do pull in the role from the SDK (it is available today in a calling build if we want to make that change) then yes we should pull everything in from the selectors and then we can avoid custom logic in the composite and the UI component.

I agree that we definitely do not want to touch the internal logic of a UI component they should be pure and should only be in a specific roles "configuration" based on the incoming props.

This leaves us with two potential places to get props from usePropsFor and from the composite. Since we can't get it from usePropsFor today but we are planning for it in Q4 then similar to the capability mapping  this functionality should temporarily come from the composite. Thoughts @JamesBurnside?

PR #2349, opened 2022-09-15
Hide PiPIP for Consumer role

mgamis-msft:
# What

Hide PiPIP for Consumer role 

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2981208

# How Tested

Local test and snapshot updates on this CI.

# Process & policy checklist

### DIFF
--- a/change/@internal-react-composites-950d55c1-62f5-4248-a46d-bbde257e51ad.json
+++ b/change/@internal-react-composites-950d55c1-62f5-4248-a46d-bbde257e51ad.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Local and remote PIP not showing local video tile if no permissions to camera",
+  "packageName": "@internal/react-composites",
+  "email": "miguelgamis@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/react-components/src/components/VideoGallery.tsx
+++ b/packages/react-components/src/components/VideoGallery.tsx
@@ -45,6 +45,8 @@ import { localVideoTileWithControlsContainerStyle, LOCAL_VIDEO_TILE_ZINDEX } fro
 import { _ICoordinates, _ModalClone } from './ModalClone/ModalClone';
 import { _formatString } from '@internal/acs-ui-common';
 import { _LocalVideoTile } from './LocalVideoTile';
+/* @conditional-compile-remove(rooms) */
+import { _usePermissions } from '../permissions';
 
 // Currently the Calling JS SDK supports up to 4 remote video streams
 const DEFAULT_MAX_REMOTE_VIDEO_STREAMS = 4;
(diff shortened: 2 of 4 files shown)


packages/react-composites/src/composites/CallComposite/components/LocalAndRemotePIP.tsx
```diff
@@ -102,31 +104,37 @@ export const LocalAndRemotePIP = (props: LocalAndRemotePIPProps): JSX.Element =>
...
+  canTurnCameraOn = permissions.cameraButton;
+
   // If there are no remote participants, show the local participant as the primary tile
   const primaryTileProps: _PictureInPictureInPictureTileProps = useMemo(
     () => ({
       children: remoteVideoTileProps ? (
         <_RemoteVideoTile {...remoteVideoTileProps} />
       ) : (
-        <_LocalVideoTile {...localVideoTileProps} />
+        canTurnCameraOn && <_LocalVideoTile {...localVideoTileProps} />
```
dmceachernmsft, line 119: wont this render just a white square in the PIPIP?

mgamis-msft: Good catch! Moved the changes to the parent, ModalLocalAndRemotePIP,

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -262,10 +264,17 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     maxDominantSpeakers: MAX_AUDIO_DOMINANT_SPEAKERS
   });
 
+  /* @conditional-compile-remove(rooms) */
+  const permissions = _usePermissions();
```
mgamis-msft, line 268: I have to add the hook in the VideoGallery component because `localParticipant` is a required prop. So there is no way for me prevent the PiPiP from the composite without a changing `localParticipant to an optional prop in the API.

change/@internal-react-composites-950d55c1-62f5-4248-a46d-bbde257e51ad.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Local and remote PIP not showing local video tile if no permissions to camera",
```
JamesBurnside, line 3: instead of "no permissions" to camera can we use "if the role does not permit camera" - reason being is I read that as if the browser has blocked permissions to the camera

packages/react-components/src/components/VideoGallery.tsx
JamesBurnside: I would actually be for making the local participant an optional prop - no reason Contoso should be forced to provide it. But happy with the solution also

JamesBurnside, line 268: Similar to the comment on the change file, (not in this PR) can we change this hook to be useRolePermissions (or just useRole) - reason being I read this as if the browser has blocked permissions to the camera

JamesBurnside (approved):

packages/react-components/src/components/VideoGallery.tsx
mgamis-msft: Making the local participant optional will be considered when for the new Video Gallery feature in Q4.

alkwa-msft: useRolePermissions will be good until we start getting Teams and Rooms roles in the mix.  (Possible suggestion "useRoomRolePermissions")
 
 I am less concerned about the name since by the time we hit GA this information will be available through calling sdk (and our selectors) and we won't need to bring the hooks from public preview -> GA with us.

alkwa-msft (approved):

PR #2350, opened 2022-09-16
Tooltip strings uncapitilized

mgamis-msft:
# What

Updated en-US tool tip strings to be capitalized but not title case.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2961689

# How Tested

Locally
https://user-images.githubusercontent.com/79475487/190564926-32f47efd-3cc0-44fb-b28f-f7d7cac5042f.mp4

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/localization/locales/en-US/strings.json
+++ b/packages/react-components/src/localization/locales/en-US/strings.json
@@ -39,7 +39,7 @@
   },
   "endCallButton": {
     "label": "Leave",
-    "tooltipContent": "Leave Call"
+    "tooltipContent": "Leave call"
   },
   "cameraButton": {
     "onLabel": "Turn off",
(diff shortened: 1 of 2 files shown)


edwardlee-msft (approved):

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -49,7 +49,7 @@
     "tooltipOffContent": "Turn on camera",
     "tooltipVideoLoadingContent": "Video is loading",
     "cameraMenuTitle": "Camera",
-    "cameraMenuTooltip": "Choose Camera",
+    "cameraMenuTooltip": "Choose camera",
```
dmceachernmsft, line 52: Is this something that also needs to be done in the composite strings as well?

dmceachernmsft (approved):

packages/react-components/src/localization/locales/en-US/strings.json
mgamis-msft: Yes. But I did not find any tooltip string in the composite. So I think we are good :)

PR #2352, opened 2022-09-16
Refactor `CallControls` to disentangle lots of conditionals and conditional compilations

prprabhu-ms:
This component had grown organically to a point where I had to spend over 2 hours breaking it up.

This PR:

* Pushes conditional compilation closer to the API.
* Removes a lot of conditional compilation directives, by doing the above.
* Streamlines the `CallControls` and helper components by splitting it up such that each component does fewer things.
* Adds a few TODOs for issues found while there.


common/config/env/.env
```diff
@@ -1,2 +1,2 @@
 // THIS IS AN AUTO GENERATED FILE BY `RUSH SWITCH <BETA|STABLE>`. DO NOT MODIFY
-COMMUNICATION_REACT_FLAVOR=beta
\ No newline at end of file
+COMMUNICATION_REACT_FLAVOR=stable
```
prprabhu-ms, line 2: Remove. We clearly need to .gitignore this, or at leaset add a git-hook warning.

dmceachernmsft (approved):

packages/react-components/review/stable/react-components.api.md
```diff
@@ -677,6 +685,23 @@ export interface GridLayoutStyles extends BaseCustomStyles {
     children?: IStyle;
 }
 
+// @beta
```
JamesBurnside, line 688: Do we allow beta exports in the stable API files?

JamesBurnside: (I know it's not in the communication-react API file but wanted to check if the API files should be consistent with the public api file while also have internal exports)

packages/react-composites/src/composites/CallComposite/components/buttons/People.tsx
```diff
@@ -27,22 +29,34 @@ export const People = (props: ControlBarButtonProps): JSX.Element => {
...
         },
-        props.styles ?? {},
         controlButtonBaseStyle
       ),
-    [props.styles, theme.palette.neutralLight]
+    [theme.palette.neutralLight]
+  );
+
+  const locale = useLocale();
+  // FIXME (?): Why is this using callWithChat strings?
```
edwardlee-msft, line 38: Thanks for the catch!
Submitted a PR to address this:
https://github.com/Azure/communication-ui-library/pull/2358

packages/calling-component-bindings/review/stable/calling-component-bindings.api.md
```diff
@@ -143,6 +143,14 @@ export type GetCallingSelector<Component extends (props: any) => JSX.Element | u
 // @public
 export const getCallingSelector: <Component extends (props: any) => JSX.Element | undefined>(component: Component) => GetCallingSelector<Component>;
 
+// @public
+export type HoldButtonSelector = (state: CallClientState, props: CallingBaseSelectorProps) => {
```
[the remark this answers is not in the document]
prprabhu-ms: I guess the right thing to do would be to convert all these to `@internal` symbols and then rename them when exporting from `@azure/communication-react`.

We must have done this before. Let me check.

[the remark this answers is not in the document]
prprabhu-ms: Hmm.. This _is_ a first.

These are symbols that are exported by @azure/communication-react in beta builds.
But react-composite wants to use them unconditionally (both stable and beta builds).

Allowing this leads to a smaller diff between beta and stable builds of the composites.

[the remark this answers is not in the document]
JamesBurnside: Should we mark it `@beta` instead of `@public`? (which I also don't like as then to make something stable a lot of signatures need updated)
But then with this implementation the export would need to be added to the `azure/communication-react` at stabilization time so changes like that would need to be made anyway... (_this particular component it doesn't matter because we don't export individual selectors but in terms of this happening again..._)

I'm torn on this, trying to think of a way it could be nicely excluded from the api file...

packages/react-components/review/stable/react-components.api.md
```diff
@@ -1119,6 +1161,11 @@ export type _PictureInPictureInPictureTileProps = PropsWithChildren<{
     orientation: _TileOrientation;
 }>;
 
+// Warning: (ae-internal-missing-underscore) The name "presenterPermissions" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
+export const presenterPermissions: _Permissions;
```
JamesBurnside, line 1167: nit: underscore `presenterPermissions`

JamesBurnside (approved):
Finding it hard to review the code as its being nicely refactored, but changelists aren't smart enough to know that a chunk of code has moved... 
So looks good at a glance! But I will need to trust it has been well tested/regressions will be caught while we selfhost

edwardlee-msft (approved):

prprabhu-ms:
PRs can be reopened you know.

And frankly, I agree with James. I wouldn't merge this PR at this (functional) size.

Issue #2321, opened 2022-09-09
Autoscrolling of messages in Safari and WebKit based browsers


Roy19:
**Describe the bug; what happened?**
On Safari/WebKit based browsers, on scrolling upwards for previous messages, new messages are continuously fetched and are added continuously and the messages list keeps autoscrolling to the top till the first message is received at which point scrolling is stopped. This however doesn't happen in Chrome or Firefox based browsers.

**What are the steps to reproduce the issue?**
Using [this](https://github.com/Azure-Samples/communication-services-javascript-quickstarts/tree/main/ui-library-starting-with-chat-stateful) sample repo, we place our user ids, access tokens and chat thread id here and scroll up in Safari or iOS mobile browsers. You can see that autoscrolling to the top happens continuously until the first message is reached. This however doesn't happen in Chrome or Firefox. Use a chat thread with more than 200 messages. 

**What behavior did you expect?**
Scrolling behaviour should be consistent as with Chrome and Firefox i.e. autoscrolling to the top should be stopped.

**In what environment did you see the issue?**

- `@azure/communication-react` npm package version: 1.2.0
- `@azure/communication-chat` npm package version: 1.1.0

- OS & Device: MacOS, iOS
- Browser: Apple Safari
- Browser Version: Apple Safari on MacOS 14.1.0

JamesBurnside:
Thanks for filing this issue, we'll take a look at this as soon as possible.

edwardlee-msft:
Thank you for filing this issue and bringing it to our attention.
We were able to reproduce the issue on Safari MacOS with our latest communication-react package version of 1.3.1 and are working on a resolving this bug.

Roy19:
@edwardlee-msft @JamesBurnside any idea when a public release might be available with the issue resolved?

edwardlee-msft:
Hi @Roy19, we are still working on the issue and will get back to you with an update as soon as one is available.

edwardlee-msft:
Status update: We are still looking into this issue and are currently working on integrating a fix into our packages. 
I will provide a new update once a fix is available.

Roy19:
@edwardlee-msft I see that the pull request has been merged. Is there a release planned out for this bug?

edwardlee-msft:
Hi @Roy19 ,

We are working on having a release ready that includes this bug fix soon. 
I will update you once the release is available.

Roy19:
Sounds good. Thanks @edwardlee-msft.

[...]

PR #657, opened 2021-08-09
Fix infinite scroll

PorterNan:
- Fix jumpy behavior
- Dedup reusable code
- Remove extra state update to improve perf of fetch messages
- fix scroll down fetching problem by always fetching enough height

scrolling

# What

# Why

# How Tested

# Process & policy checklist

- [ ] I have updated the project documentation to reflect my changes if necessary.
- [ ] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.

**Is this a breaking change?**

- [ ] This change causes current functionality to break.

PR #2355, opened 2022-09-20
[bugfix] Safari specific: Load old messages only when scroll bar is at the top of its container

edwardlee-msft:
# What
Change loading old messages from 500 to 0px scroll bar position (from the top of its container).

https://user-images.githubusercontent.com/97130533/191359858-27af185c-c8ce-4372-a22d-4c61e972efdc.mov

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2984697
https://github.com/Azure/communication-ui-library/issues/2321
On safari browser, messages were constantly being loaded as the scroll bar get's "stuck" between 1-50 px from the top of it's container.

https://user-images.githubusercontent.com/97130533/191359276-3a994a44-657f-4023-aa0f-04ccd8e5c50d.mov

# How Tested
Tested locally on MacOS safari and on chrome.

# Process & policy checklist


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -849,7 +849,7 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
       if (onLoadPreviousChatMessages) {
         isLoadingChatMessagesRef.current = true;
         // Fetch message until scrollTop reach the threshold for fetching new message
-        while (!isAllChatMessagesLoadedRef.current && chatScrollDivRef.current.scrollTop <= 500) {
+        while (!isAllChatMessagesLoadedRef.current && chatScrollDivRef.current.scrollTop <= 0) {
```
JamesBurnside, line 852: I think we'll still need some tolerance value - does reducing this to `50px` instead of `0` still work? I'll play around in storybook and see what the behavior is like with `0px` on windows

edwardlee-msft: 50px is a hit or miss. In safari, the scrollbar can actually stick to the top < 50. 
From my testing, I've had the scrollbar stick as low as 19px from the top. (Hard to repro, but I was able to).

dmceachernmsft (approved):

packages/react-components/src/components/MessageThread.tsx
mgamis-msft: At any value greater than 0 it can theoretically stick to the top given enough messages/ height of the browser window.  0 is the only value that guarantees no subsequent loading of previous messages. @JamesBurnside I was wondering why we need a tolerance value.

mgamis-msft (approved):

packages/react-components/src/components/MessageThread.tsx
JamesBurnside: My gut feeling was scrollTop sometimes doesn't go to 0 when scrolling (it can stick at like 1 or 5) but I cannot repro nor find any historical context to support that.
But maybe get @PorterNan to sign off on this as he added the original 200 check: https://github.com/Azure/communication-ui-library/pull/146/files#diff-a7cad3c035297ccc8b0624d8faffb541631f0c836f4c2bc3383bae7d76581f55
Then upped it to 500: https://github.com/Azure/communication-ui-library/pull/657/files#diff-ab8c8370a350df97f2fd3fb1e1ebaa994f438c4b8f0b569552d92ca1b70087e5
@PorterNan do you recall the reason for the 500? (especially since it was upped from 200?)

PorterNan, line 852: The logic here is for smoothing the infinite scroll - the chat thread will be starting pulling your messages before you actually hit the top of the message (pulling only when scrollTop equals 0 will make it feel super laggy when scrolling up)

So this will actually fix the problem, but it changes the experience of infinite scrolling

PorterNan: I just looked around and found out Safari doesn't support overflow-anchor propery - so there is no way from css side to change its behavior of doing this "sticky on the top" problem.

But we do have the full control from js side - https://kirbysayshi.com/2013/08/19/maintaining-scroll-position-knockoutjs-list.html this sounds like a promising solution with [JSFiddle](https://jsfiddle.net/kirbysayshi/57HbV/)

Maybe we can enable this polyfill when it is under Safari

edwardlee-msft: @PorterNan ,
Good point. I figured that it was for a "smooth" experience.
However, I did notice that even on chrome, if you drag the scrollbar high enough it can also get stuck and keep scrolling. It's less obvious compared to on Safari, but it is reproducible. 

What do you think would be a good alternative to this?

PorterNan: If you look at the code around line 871 - 891, it is targeting the problem for chrome, maybe you can change the threshold a bit (set it as 50px or something more, if 50px is not working try something like percentage of totally height) then test if the bug would be fixed in Safari (also check chrome if you can repro it well)

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -849,8 +849,12 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
       if (onLoadPreviousChatMessages) {
         isLoadingChatMessagesRef.current = true;
         // Fetch message until scrollTop reach the threshold for fetching new message
+        previousTopRef.current = chatScrollDivRef.current.scrollTop;
```
PorterNan, line 852: Additional lines in this PR (852-857) looks a bit duplicate with line 878-892, do we have any concern of having same logic in 2 places?
I guess we can do this:
1. test if the logic could be in separated useEffect closures, if yes, then put them in different use effect (use effect will run after each loop of while, because rendering will be re-triggered), fetch function is responsible for only getting new messages without any dom operation
2. If we can't separate them into different useEffect, I guess we can remove 874-893, cause all the logic has been covered by your new code (your code condition is like a superset)

PorterNan: You've found the core problem of the bug, just want to make the code a bit readable and without duplicate logic, please lemme know if you have anything unsure about my comment, would like to have a quick sync offline

PorterNan (approved):
Sync with Edward offline - looks like we are able to merge all logic into useEffect by removing the condition of (previousTopRef.current === 0)
And it also fixes another bug of chrome infinite fetch :D, excellent job Edward!

The fix perfectly resolves the infinite fetch problem in Safari., there is still minor offset for Safari, we guess it is some wrong height caculation of Safari, but that's not the purpose of this fix(please open another bug for safari once you check this in, @edwardlee-msft), 

Approved here, and please feel free to merge once you push your change @edwardlee-msft

packages/react-components/src/components/MessageThread.tsx
edwardlee-msft: Thanks so much Nan!

Issue #2361, opened 2022-09-23
Upgrade Calling SDK to 1.8.0-beta.1


mgamis-msft:
Available in NPM - https://www.npmjs.com/package/@azure/communication-calling/v/1.8.0-beta.1

PR #2362, opened 2022-09-23
Bump calling beta to 1.8.0 beta.1

mgamis-msft:
# What

Bump calling beta sdk version to 1.8.0-beta.1. Made the appropriate changes from the breaking changes.

# Why

#2361 

# How Tested

Tested 3 participants in a group call with video and audio.
Tested 1 participant joining a Teams call with video and audio.

# Process & policy checklist


alkwa-msft (approved):

edwardlee-msft (approved):

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -209,8 +209,20 @@ class ProxyCallAgent implements ProxyHandler<DeclarativeCallAgent> {
         };
       }
       case 'dispose': {
+        // Wrapping CallAgent.dispose in a callback type (): Promise<void> to accomodate the change of CallAgent.dispose
+        // in calling beta version 1.8.0-beta.1 from callback type (): Promise<void> to (): void
+        const disposeCallbackWrapper = (): Promise<void> => {
+          target.dispose();
```
anjulgarg, line 215: Can `target.dipose()` throw errors?

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -209,8 +209,20 @@ class ProxyCallAgent implements ProxyHandler<DeclarativeCallAgent> {
...
+          return Promise.resolve();
+        };
+        /* @conditional-compile-remove(calling-beta-sdk) */
+        return (): void => {
+          disposeCallbackWrapper().then(() => {
```
anjulgarg, line 220: Why aren't we using await/async here?

mgamis-msft: Good catch. I changed the beta code to be a simple synchronous call of CallAgent.dispose.

packages/calling-stateful-client/src/CallAgentDeclarative.ts
mgamis-msft: I suppose it can. But we never had any catch logic before and an error in the console log will still show.

PR #2365, opened 2022-09-27
[SPIKE] - Proposed API Changes for Pinning Video Tiles in Call Composite

anjulgarg:
# What
Pinned Participants - Call Composite Changes 

The other part of this change
https://github.com/Azure/communication-ui-library/pull/2369

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -61,6 +61,8 @@ export const END_CALL_PAGES: CallCompositePage[] = [
 export type CallAdapterUiState = {
   isLocalPreviewMicrophoneEnabled: boolean;
   page: CallCompositePage;
+  /* @conditional-compile-remove(pinned-participants) */
+  pinnedParticipants: string[];
```
anjulgarg, line 65: This is exploratory code.
Another option is to track pinned participants directly inside the VideoGallery
Which would then require the composite to drill the `pinParticipant(userId: string)` function down to the VideoGallery

JamesBurnside: I like it better at this higher level than pure components; it makes the component do slightly less magic and also there's a chance other components could need this info (e.g. if participant pane shows a pin icon next to whomever is pinned)

The only friction with storing the state in the adapter is that usePropsFor won't supply this feature to the video gallery

PR #2366, opened 2022-09-27
Fix blank camera devices on mac

mgamis-msft:
# What

Filtering out all devices with blank names which are the dummy devices from the Calling SDK.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2963211

# How Tested

Locally on a Mac

# Process & policy checklist


packages/calling-component-bindings/src/callControlSelectors.ts
```diff
@@ -145,16 +145,20 @@ export const devicesButtonSelector: DevicesButtonSelector = reselect.createSelec
...
+      speakers: filterBlankNames(deviceManager.speakers),
+      cameras: filterBlankNames(deviceManager.cameras),
       selectedMicrophone: deviceManager.selectedMicrophone,
       selectedSpeaker: deviceManager.selectedSpeaker,
       selectedCamera: deviceManager.selectedCamera
     };
   }
 );
 
+function filterBlankNames<T extends { name: string }>(devices: T[]): T[] {
```
anjulgarg, line 158: The name `filterBlankNames` doesn't seem right semantically.
Might I suggest something simpler like `removeBlankNameDevices` ?

mgamis-msft: Thanks. Updated.

carocao-msft (approved):

anjulgarg (approved):

PR #2369, opened 2022-09-28
[SPIKE] - Proposed API Changes for Pinning Video Tiles in Video Gallery

anjulgarg:
# What
[SPIKE] - Proposed API Changes for Pinning Video Tiles in Video Gallery

The other part of this change
https://github.com/Azure/communication-ui-library/pull/2365

# Why

# How Tested

# Process & policy checklist


JamesBurnside (commented):
LGTM!

Leaving nit comments here to not comment the changelist for others reviewing:
* click is normally the suffix in our codebase, so `onPinIconClick` or just `onPinClick`
* Typically we have a single function for onX and offX (e.g. [toggle button](https://developer.microsoft.com/en-us/fluentui#/controls/web/toggle) has just onChange for both on and off) so for `onPinParticipant` and `onUnpinParticipant` RemoteVideoTile.tsx layer I would suggest just have `onPinChange: (newPinState: 'on' | 'off') => void`

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -111,6 +114,20 @@ export const _RemoteVideoTile = React.memo(
         /* @conditional-compile-remove(one-to-n-calling) */
         /* @conditional-compile-remove(PSTN-calls) */
         participantState={props.participantState}
+        isPinned={props.isPinned}
+        onClickPinIcon={() => {
+          props.onUnpinParticipant?.();
+        }}
+        menuItems={[
```
JamesBurnside, line 121: if we don't add these conditionally its kinda a breaking change - Contoso will now have a pin button that doesn't work.
Consider only including the menu items if `onPinParticipant` is provided?

anjulgarg, line 121: Open Question:
How do we make this work with the drawer menu on mobile?
Should we refrain from using a Fluent type here and introduce a simplified structure that works with both the contextual menu and the drawer menu?

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -125,6 +125,25 @@ export interface VideoTileProps {
...
+   * The pin icon is only visible when `isPinned` is true.
+   */
+  onClickPinIcon?: () => void;
+  /* @conditional-compile-remove(pinned-participants) */
+  /**
+   * Display custom menu items in the VideoTile's contextual menu.
+   * It uses Fluent UI's ContextualMenu component. Thus the menu items should be of type `IContextualMenuItem`.
+   * For more information see https://developer.microsoft.com/en-us/fluentui#/controls/web/contextualmenu
+   */
+  menuItems?: IContextualMenuItem[];
```
anjulgarg, line 146: Open Question:
How do we make this work with the drawer menu on mobile?
Should we refrain from using a Fluent type here and introduce a simplified structure that works with both the contextual menu and the drawer menu?

PR #2370, opened 2022-09-29
Hiding copy link button for Rooms call

mgamis-msft:
# What

1.) If role is defined in CallComposite then don't pass in the callInvitationUrl.
2.) Hiding Desktop and mobile 'Add People' button if it has no menu items. This just makes sense because if we have no menu items then why have the button at all. It won't have any menu items because of 1.)

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2981209

# How Tested

Local sample testing
https://user-images.githubusercontent.com/79475487/192914993-7f175797-e730-44e1-8385-68d1570de15e.mp4

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -372,7 +372,7 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
       <BaseProvider {...props}>
         <CallAdapterProvider adapter={adapter}>
           <MainScreen
-            callInvitationUrl={callInvitationUrl}
+            callInvitationUrl={role ? undefined : callInvitationUrl}
```
JamesBurnside, line 375: Is this needed? If Contoso wishes to provide a copyInviteLink should we not permit them to do so?

I definitely think we should have the below change where you only show the add people button iff there are actual menu items attached to it!

JamesBurnside: (our samples and selfhost shouldn't pass it in though because it won't function correctly)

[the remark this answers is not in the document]
mgamis-msft: Fixed. Added a comment to why role indicates a Rooms call.

mgamis-msft: Great point. We shouldn't hide a copyInviteLink if Contoso has provided one. Moved this logic to the Calling sample.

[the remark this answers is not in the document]
mgamis-msft: Moved this logic to the sample code and wrote a comment to explain a defined role indicates a Rooms call.

JamesBurnside (approved):

JamesBurnside:
LGTM! reset approval until CI tests are green

prabhjot-msft (approved):

PR #2372, opened 2022-09-29
Carocao/device button composite

carocao-msft:
# What
Add device dropdown to composite 

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2957778

# How Tested

https://user-images.githubusercontent.com/96077406/192966550-9d346217-f5ac-4554-bf1c-1635b0570a7b.mov


packages/react-composites/src/composites/CallComposite/components/LocalDeviceSettings.tsx
```diff
@@ -104,7 +114,146 @@ export const LocalDeviceSettings = (props: LocalDeviceSettingsType): JSX.Element
...
+            event: React.FormEvent<HTMLDivElement>,
+            option?: IDropdownOption | undefined,
+            index?: number | undefined
+          ) => {
+            props.onSelectMicrophone(props.microphones[index ?? 0]);
+          }}
+          onRenderTitle={(props?: IDropdownOption[]) => onRenderTitle('Microphone', props)}
+        />
+      )}
+      <Dropdown
```
JamesBurnside, line 212: Check with Juan on this but I think speakers should always be showing as this isn't dependant on having mic permissions

carocao-msft: Haven't got a reply from Juan yet but I believe you are right! Modify the code to not hide the speaker dropdown for now

packages/react-composites/src/composites/CallComposite/components/LocalDeviceSettings.tsx
```diff
@@ -116,32 +265,7 @@ export const LocalDeviceSettings = (props: LocalDeviceSettingsType): JSX.Element
...
-                  ? props.cameras[0]?.id
-                  : ''
-                : 'deniedOrUnknown'
-            }
-            onChange={(event, option, index) => {
-              props.onSelectCamera(props.cameras[index ?? 0], localVideoViewOptions);
-            }}
-            onRenderTitle={(props?: IDropdownOption[]) => onRenderTitle('Camera', props)}
-          />
+          {conditionalCompileCameraDropdown}
```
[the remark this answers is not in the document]
carocao-msft: Fixed that!

packages/react-composites/src/composites/CallComposite/components/ConfigurationpageMicDropdown.tsx
```diff
@@ -0,0 +1,55 @@
...
+  };
+
+  /* @conditional-compile-remove(call-readiness) */
+  const microphoneBlockedDropdown = (
+    <_DevicePermissionDropdown
+      styles={dropDownStyles(theme)}
+      onClickActionButton={async () => {
+        await adapter.askDevicePermission({ video: false, audio: true });
+        if (props.micPermissionGranted) {
+          adapter.queryMicrophones();
```
JamesBurnside, line 43: Is this needed? I would've assumed this would trigger the queries:
image

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/ConfigurationpageMicDropdown.tsx
carocao-msft: This is for when camera/mic permission is already granted, it will find microphone/camera options and display them on the config screen. Without this when using the calling sample the second time ( already set up permissions the first time so pop up will not be triggered), you wont be able to see any microphone/camera device options.

prabhjot-msft (approved):

PR #2374, opened 2022-09-29
Fix rush changelog not working for forks of the repo

JamesBurnside:
# What
* Change the beachball config to use the default repo origin specified in a root package json
* Add a root level package.json that specifies this repo as the root repo to use

# Why
`rush changelog` is not working for forks of the repo (as origin/main on a fork of the repo is the forked repo and not this repo, we need rush changelog to check new changes against _this_ repo)

More information: https://github.com/Azure/communication-ui-library/pull/2285#issuecomment-1258756974

# How Tested
* Upon pulling the branch first verified running without this change rush changelog did not work, then manually applying this change it now works.
* Locally a branch on this repo still detects changes
   image

### DIFF
--- a/common/config/beachball/changelog-config.ts
+++ b/common/config/beachball/changelog-config.ts
@@ -2,7 +2,6 @@ import { BeachballConfig } from 'beachball';
 import { renderHeader, renderEntry } from './changelog-custom-renders';
 
 export const config: BeachballConfig = {
-  branch: 'origin/main',
   changelog: {
     customRenderers: {
       renderHeader,
--- a/package.json
+++ b/package.json
@@ -1,0 +1,11 @@
+{
+  "homepage": "https://azure.github.io/communication-ui-library/",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Azure/communication-ui-library.git"
+  },
+  "bugs": {
+    "url": "https://github.com/Azure/communication-ui-library/issues/new/choose"
+  }
+}


common/config/beachball/changelog-config.ts
```diff
@@ -2,7 +2,6 @@ import { BeachballConfig } from 'beachball';
 import { renderHeader, renderEntry } from './changelog-custom-renders';
 
 export const config: BeachballConfig = {
-  branch: 'origin/main',
```
anjulgarg, line 5: So this will cause beachball to use `main` but not just from `origin`?

JamesBurnside: Indeed! If there is an upstream that matches the .git of the package .json it will use that instead automagically

screencap it auto-used the `upstream/main` without telling it to:
image

JamesBurnside: If your curious here's how:
image

prprabhu-ms:
What's the new behavior on a release branch?

prprabhu-ms (approved):

JamesBurnside:
> What's the new behavior on a release branch?

Should be no behavior change here, both before and after this change should check against `origin/main` (do we have any custom logic anywhere that changes this for release branches currently?)

prprabhu-ms:
> > What's the new behavior on a release branch?
> 
> Should be no behavior change here, both before and after this change should check against `origin/main` (do we have any custom logic anywhere that changes this for release branches currently?)

Ah.. I added a hack for this. The prerelease branch creation workflow just adds a changelog for all packlets to shut rush changelog for the branch forever more.

So this is an improvement there as well, and we can probably remove the hack now.

mgamis-msft (approved):

PR #2378, opened 2022-09-29
Hide Hold button for Rooms Call

prabhjot-msft:
# What
Hide Hold button for Rooms Call

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2984362

# How Tested

# Process & policy checklist


mgamis-msft:
LGTM! Waiting for CI to pass then will approve.

mgamis-msft:
> LGTM! Waiting for CI to pass then will approve.

Looks like you need to update the PermissionProvider.test.ts enzyme tests. But other than that it looks good.

mgamis-msft (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -279,3 +281,13 @@ export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX
 };
 
 const isEnabled = (option: unknown): boolean => option !== false;
+
+/** @private */
+export const isRoomsCallTrampoline = (): boolean => {
+  /* @conditional-compile-remove(rooms) */
+  const rolePermissions = _usePermissions();
+  /* @conditional-compile-remove(rooms) */
+  return rolePermissions.role ? true : false;
```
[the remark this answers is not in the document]
prabhjot-msft: done

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
```diff
@@ -279,3 +281,13 @@ export const CallControls = (props: CallControlsProps & ContainerRectProps): JSX
 };
 
 const isEnabled = (option: unknown): boolean => option !== false;
+
+/** @private */
```
PorterNan, line 285: This function seems appear multiple time in composite (correct me if I am wrong), can we put it somewhere as a util function?

PorterNan (approved):

packages/react-composites/src/composites/CallComposite/components/CallControls.tsx
prabhjot-msft: Earlier it was supposed to be used for config screen text and hold button. But since we are not doing that change right now, it is only being used here. But we can definitely move this to util function if we think it can be reused

PR #2379, opened 2022-09-29
Center DialPad backspace icon

anjulgarg:
# What
Center DialPad backspace icon

# Why

# How Tested

# Process & policy checklist


dmceachernmsft (approved):

mgamis-msft (approved):

packages/react-components/src/components/styles/Dialpad.styles.ts
```diff
@@ -95,7 +95,11 @@ export const iconButtonStyles = (theme: Theme): IButtonStyles => {
       color: `${theme.palette.black}`
     },
     icon: {
-      height: 'auto'
+      height: 'auto',
+      // Needed to keep the icon vertically centered.
+      '> span': {
+        display: 'flex'
```
edwardlee-msft, line 101: I'm surprised this is working as the span inside the icon button already has a display flex type
image

edwardlee-msft: I did find that that changing the '> i': { fontSize: '20px' } resolves this issue.

edwardlee-msft: I did find doing the following fixed the issue:
image

carocao-msft (approved):

packages/react-components/src/components/styles/Dialpad.styles.ts
anjulgarg: :o
Strange. That explains why the snapshots were not generated

edwardlee-msft: nit: instead of using px, use 1.25rem.

PR #2385, opened 2022-10-03
[Call readiness] Unsupported browser page wired into composite

dmceachernmsft:
# What

Introduces unsupported browser page into Calling Composite
# Why

to show when user is on a unsupported browser and provide troublshooting information provided by contoso. 

https://skype.visualstudio.com/SPOOL/_workitems/edit/2980683
# How Tested

Will have hermetic tests in follow-up PR to verify styling and positioning of component is correct.

Since we do not have access to the browser-check API from calling yet this page will never functionally show.


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -282,6 +298,16 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
         </>
       );
       break;
+    case unsupportedBrowserPageTrampoline():
```
mgamis-msft, line 301: Personally, I would prefer an if statement outside this switch statement without the trampoline. But this works.

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -145,6 +147,20 @@ export type CallCompositeOptions = {
...
+   * @example
+   * ```ts
+   * onBrowserTroubleShootingClick?: () =>
+   *  window.open('https://contoso.com/browser-troubleshooting', '_blank');
+   * ```
+   *
+   * @remarks
+   * if this is not supplied, the composite will not show a unsupported browser page.
+   */
+  onBrowserTroubleShootingClick?: () => void;
```
JamesBurnside, line 163: `TroubleShooting` -> `Troubleshooting`

JamesBurnside: We should update onNetworkingTroubleShootingClick at some point as well (in a seperate PR)

JamesBurnside: Can we update this to `onEnvironmentUnsupportedTroubleshootingClick`? Its a bit verbose but given we are checking the environmentInfo it would be nice to be consistent

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -40,7 +40,8 @@ export type CallCompositePage =
   | 'lobby'
   | /* @conditional-compile-remove(rooms) */ 'deniedPermissionToRoom'
   | 'removedFromCall'
-  | /* @conditional-compile-remove(rooms) */ 'roomNotFound';
+  | /* @conditional-compile-remove(rooms) */ 'roomNotFound'
+  | /* @conditional-compile-remove(unsupported-browser) */ 'unsupportedBrowser';
```
JamesBurnside, line 44: per other comment if we decide to so that, update from unsupportedBrowser to unsupportedEnvironment

packages/react-composites/src/composites/CallComposite/pages/UnsupportedBrowser.tsx
```diff
@@ -1,20 +1,38 @@
...
+import { useLocale } from '../../localization';
 
 /**
  * @internal
  */
-export type UnsupportedBrowserProps = {
-  onTroubleshootingClick: () => void;
-  // strings type from new component...
+export type UnsupportedBrowserPageProps = {
+  onTroubleShootingClick: () => void;
```
JamesBurnside, line 16: onTroubleShootingClick -> onTroubleshootingClick

JamesBurnside (approved):
just some naming nits

change/@internal-react-composites-a837720f-f773-4c44-a998-b3af9d2b928c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "introduces unsupported browser page for the calling composites using new component.",
```
edwardlee-msft, line 3: nit:
introduces -> Introduce

edwardlee-msft (approved):
Approved ahead of changes to 'TroubleShooting' text change to 'Troubleshooting'

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
dmceachernmsft: Which other comment? did you want me to change the feature flag here after our discussion this morning?

JamesBurnside: image
^this one

PR #2391, opened 2022-10-05
[CallReadiness] unsupported browser check handlers

dmceachernmsft:
# What

adds handlers to the CallingStatefulClient and CallAdapter to get environment information.
# Why

Allows us to use the Calling SDK to check whether the end user is using a supported browser

https://skype.visualstudio.com/SPOOL/_workitems/edit/2980686
# How Tested


packages/calling-stateful-client/src/StatefulCallClient.ts
```diff
@@ -211,6 +211,12 @@ class ProxyCallClient implements ProxyHandler<CallClient> {
           return this._deviceManager;
         }, 'CallClient.getDeviceManager');
       }
+      case 'getEnvironmentInfo': {
+        return this._context.withAsyncErrorTeedToState(async () => {
+          const environmentInfo = await target.getEnvironmentInfo();
+          return environmentInfo;
```
JamesBurnside, line 217: instead of just returning here, this result should be stored into the call context then the selector can pluck it out of the context

dmceachernmsft:
blocked on Calling API. 

Current behavior is the constructor for the call client fails so we cannot utilize the handlers in this PR. Will update when path forward

dmceachernmsft:
Closing this, revised in this PR: https://github.com/Azure/communication-ui-library/pull/2479

PR #2392, opened 2022-10-05
Fix composite elements from disappearing when in the disconnecting stage

JamesBurnside:
# What
* Stop call with chat control bar disappearing in the disconnecting stage
* Mitigate video spinners showing in the disconnecting stage

# Why
We recently changed the isInCall behavior which meant we show the video gallery during the transition from incall->disconnecting->disconnected.
This is the correct behavior (to show the video gallery) because during this phase its possible the local user's camera and mic are still active.
However we were accidently hiding the call controls bar and showing video spinners in the remote participant video tiles during this disconnection phase.

# How Tested
Locally - video gallery spinners still show but there is a bug on the calling sdk to resolve that and ensure their state is correct: https://skype.visualstudio.com/SPOOL/_workitems/edit/3009282


packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -257,7 +257,7 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
   /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
   const isInLocalHold = currentPage === 'hold';
   const hasJoinedCall = !!(currentPage && hasJoinedCallFn(currentPage, currentCallState ?? 'None'));
-  const showControlBar = isInLobbyOrConnecting || hasJoinedCall;
+  const showControlBar = isInLobbyOrConnecting || hasJoinedCall || currentCallState === 'Disconnecting';
```
PorterNan, line 260: Should we consider "disconnecting" still a part hasJoinedCall state in line 259?

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -83,6 +83,8 @@ export const _RemoteVideoTile = React.memo(
     // Handle creating, destroying and updating the video stream as necessary
     useRemoteVideoStreamLifecycleMaintainer(remoteVideoStreamProps);
 
+    const showLoadingIndicator = isAvailable && isReceiving === false && props.participantState !== 'Disconnected';
```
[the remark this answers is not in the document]
JamesBurnside: We _shouldn't_ as `renderElement` should be null in that case. But added as an extra check here anyway to remove the dependency that the loading spinner must then be tied to that check

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
JamesBurnside: Yep I like that better, updated

mgamis-msft (approved):

PR #2393, opened 2022-10-07
Call readiness opt in feature 

carocao-msft:
# What
Add call readiness option so contoso can choose to opt in/opt out callreadiness
modified samples to default opt in

# How Tested
Enzyme test added for call composite
removed old logic for call readiness permission promt


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -115,6 +115,12 @@ export type CallCompositeOptions = {
    */
   devicePermissions?: DevicePermissionRestrictions;
   /* @conditional-compile-remove(call-readiness) */
+  /**
+   * Opt in call readiness feature for your call
+   * setting this to True will add call readiness feature in call experience
+   */
+  callReadinessOptedIn?: boolean;
```
edwardlee-msft, line 122: I prefer we do not use the term "optedIn" for when Contoso would like to use Call Readiness features.
Possibly rename to something like:
"enableCallReadiness"
or 
"enableCallReadinessChecks"
etc..

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -111,6 +111,8 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
       rtl={currentRtl}
       callInvitationUrl={callInvitationUrl}
       formFactor={isMobileSession ? 'mobile' : 'desktop'}
+      /* @conditional-compile-remove(call-readiness) */
+      options={{ callReadinessOptedIn: true }}
```
edwardlee-msft, line 115: Should this be set to true? Wouldn't the prop be passed down?
The Calling Sample and CallWithChat sample would pass down the prop to set this value

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -366,7 +366,9 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
               options={{
                 callControls: false,
                 /* @conditional-compile-remove(call-readiness) */
-                devicePermissions: props.devicePermissions
+                devicePermissions: props.devicePermissions,
+                /* @conditional-compile-remove(call-readiness) */
+                callReadinessOptedIn: true
```
edwardlee-msft, line 371: Same as my other comment from the CallScreen.tsx
Should this be set to true? Wouldn't the prop be passed down?
The Calling Sample and CallWithChat sample would pass down the prop to set this value

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
carocao-msft: @JamesBurnside should we name it optedin or enable? When we were syncing up on this we named it callReadinessOptedIn

samples/Calling/src/app/views/CallScreen.tsx
carocao-msft: callReadinessOptedIn is a callcomposite option, I believe it should be set directly on the callcomposite? similar to what we did with rooms and devicepermission options

change/@internal-react-composites-43594bee-c7ad-484e-a7e6-844473bba4f6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "addd callreadiness opt in option for call composite",
```
edwardlee-msft, line 3: "addd callreadiness opt in option for call composite",
to
"Add enable call readiness option for Call and CallWithChat composite"

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
carocao-msft: fixed

samples/Calling/src/app/views/CallScreen.tsx
carocao-msft: fixed

change/@internal-react-composites-43594bee-c7ad-484e-a7e6-844473bba4f6.json
carocao-msft: fixed

samples/CallWithChat/src/app/views/CallScreen.tsx
```diff
@@ -103,6 +106,8 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
       rtl={currentRtl}
       joinInvitationURL={window.location.href}
       formFactor={isMobileSession ? 'mobile' : 'desktop'}
+      /* @conditional-compile-remove(call-readiness) */
+      options={{ callReadinessOptedIn: callReadinessOptedIn }}
```
JamesBurnside, line 110: this will need memoized as its an object

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -346,43 +353,43 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
     /* @conditional-compile-remove(rooms) */
     role
   } = props;
+
+  const getQueryOptions = (options: {
```
JamesBurnside, line 357: move this helper function outside of the render function body (either the bottom of the file or its own file)

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -115,6 +116,12 @@ export type CallCompositeOptions = {
    */
   devicePermissions?: DevicePermissionRestrictions;
   /* @conditional-compile-remove(call-readiness) */
+  /**
+   * Opt in call readiness feature for your call
+   * setting this to True will add call readiness feature in call experience
```
JamesBurnside, line 121: nit
```suggestion
 * Setting this to `true` will add call readiness features to the call experience
```

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
JamesBurnside: I think we should write a one-pager on this for the team to review. The doc should cover naming, as well as discussing if there is a time from which we should enable by default instead of opt-in (e.g. "after a year if we do a major bump we should look to convert all opt-ins to byDefault features"), and if possible contrast this API with Calling's current feature API.
I agree the suffix `OptedIn` isn't quite right, but I think getting the team's opinion is best here as it will need to scale to more features

JamesBurnside: We are going to keep it in its current state then iterate on API name

JamesBurnside (approved):

dmceachernmsft (approved):

PR #2396, opened 2022-10-11
Add TeamsCall and TeamsCallAgent to stateful

PorterNan:
- Add new beta feature flag "teams-call"
- Extract common logic CallAgentCommon and CallCommon
- Add TeamsCall and TeamsCallAgent to stateful

# What

# Why

# How Tested

# Process & policy checklist


packages/calling-component-bindings/package.json
```diff
@@ -36,12 +36,12 @@
     "reselect": "~4.0.0"
   },
   "peerDependencies": {
-    "@azure/communication-calling": "1.6.1-beta.1 || 1.4.4",
```
prprabhu-ms, line 39: Please separate out the calling SDK version bump into its own PR from the PR that introduces the `TeamsCall` related hierarchy?

The `TeamsCall` introduction is a sizable independenty change and I'd like to avoid other routine changes needed for the headless SDK bump from getting mixed into that change.

prprabhu-ms:
Actual PR review pending splitting off of the SDK version bump into its own PR.

packages/calling-component-bindings/package.json
prprabhu-ms: Still waiting for this PR to be split for a proper review.

Please re-assign when ready.

prprabhu-ms (changes requested):
Are we able to splti this PR into two -- one to just bump the calling SDK version and then one to add upport for `TeamsCall`?

This PR is fairly large and complex, so any ways you can think of to take smaller steps would also be good.

PorterNan:
> Are we able to splti this PR into two -- one to just bump the calling SDK version and then one to add upport for `TeamsCall`?
> 
> This PR is fairly large and complex, so any ways you can think of to take smaller steps would also be good.

The bump one has been done by some other PRs (not from me), so currently it is 1.8.0-beta.1, I did merge it into this PR, so there is no bumpping needed anymore

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -1,21 +1,13 @@
...
+import { CallAgent } from '@azure/communication-calling';
+import { ProxyCallAgentCommon } from './CallAgentDeclarativeCommon';
 import { CallContext } from './CallContext';
-import { callDeclaratify, DeclarativeCall } from './CallDeclarative';
-import { CallSubscriber } from './CallSubscriber';
-import { convertSdkCallToDeclarativeCall, convertSdkIncomingCallToDeclarativeIncomingCall } from './Converter';
-import { DeclarativeIncomingCall, incomingCallDeclaratify } from './IncomingCallDeclarative';
-import { IncomingCallSubscriber } from './IncomingCallSubscriber';
+/* @conditional-compile-remove(teams-call) */
+import { DeclarativeIncomingCall } from './IncomingCallDeclarative';
```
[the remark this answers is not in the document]
prprabhu-ms: This should be (one-to-n-calling)

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -38,29 +30,16 @@ export type DeclarativeCallAgent = CallAgent & /* @conditional-compile-remove(on
...
-    this._callSubscribers = new Map<Call, CallSubscriber>();
-    this._incomingCallSubscribers = new Map<string, IncomingCallSubscriber>();
-    this._declarativeIncomingCalls = new Map<string, DeclarativeIncomingCall>();
-    this._declarativeCalls = new Map<Call, DeclarativeCall>();
-    this._externalCallsUpdatedListeners = new Set<CollectionUpdatedEvent<Call>>();
     this.subscribe();
   }
 
-  private subscribe = (): void => {
+  protected subscribe = (): void => {
```
prprabhu-ms, line 42: There is no need for inheritence. This should just be `private` in this class.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+      this._declarativeCalls.set(call, newDeclarativeCall as DeclarativeTeamsCall);
+      return newDeclarativeCall;
+    }
+
+    const newDeclarativeCall = callDeclaratify(call, this._context);
+    this._declarativeCalls.set(call, newDeclarativeCall as DeclarativeCall);
+    return newDeclarativeCall;
+  };
+
+  public getCommon<
```
prprabhu-ms, line 178: `getCommon` should be `protected`.

Can this just be `protected get`?

prprabhu-ms: Answer: can't be `get` because you can't override functions that are being proxied.

Please add a comment to this function describing this.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    for (const [_, declarativeCall] of this._declarativeCalls.entries()) {
+      declarativeCall.unsubscribe();
+    }
+    this._declarativeCalls.clear();
+  };
+
+  protected abstract unsubscribe(): void;
+
+  protected callsUpdated = (event: {
```
prprabhu-ms, line 84: I don't like this.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+      }
+      this._context.setCallEnded(call.id, call.callEndReason);
+      const declarativeCall = this._declarativeCalls.get(call);
+      if (declarativeCall) {
+        declarativeCall.unsubscribe();
+        removedStatefulCall.push(declarativeCall);
+        this._declarativeCalls.delete(call);
+      } else {
+        if (isACSCall(call)) {
+          removedStatefulCall.push(callDeclaratify(call, this._context));
```
prprabhu-ms, line 109: Can we add `protected abstract declaratifyCall(call)` and then pass just `CallCommon` down.
You'll have to cast down in the implementation of `declaratifyCall`, but you can also add a check with `isAcsCall` there.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+} from './Converter';
+import { DeclarativeIncomingCall, incomingCallDeclaratify } from './IncomingCallDeclarative';
+import { IncomingCallSubscriber } from './IncomingCallSubscriber';
+import { InternalCallContext } from './InternalCallContext';
+import { disposeAllViewsFromCall } from './StreamUtils';
+/* @conditional-compile-remove(teams-call) */
+import { DeclarativeTeamsCall, teamsCallDeclaratify } from './TeamsCallDeclarative';
+
+/**
+ * TODO: This should likely be exported?
```
prprabhu-ms, line 25: Tangent: Remove TODO.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+    this._context = context;
+    this._internalContext = internalContext;
+    this._callSubscribers = new Map<Call, CallSubscriber>();
+    this._incomingCallSubscribers = new Map<string, IncomingCallSubscriber>();
+    this._declarativeIncomingCalls = new Map<string, DeclarativeIncomingCall>();
+    this._declarativeCalls = new Map<Call, DeclarativeCall>();
+    this._externalCallsUpdatedListeners = new Set<CollectionUpdatedEvent<CallCommon>>();
+  }
+
+  protected abstract subscribe(): void;
```
prprabhu-ms, line 60: note: remove `protected subscribe`.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+
+  // Unsubscribe is called when CallAgent is disposed. This should mean no more updating of existing call but we don't
+  // remove any existing state.
+  protected unregisterSubscriber = (): void => {
```
prprabhu-ms, line 64: This should be `private`.

And change 

```
protected unsubscribe(): void {
    this.unregisterSubscriber();
}

// CallAgentDeclarative.
protected unsubscribe() {
    super.unsubscribe();
    // ... do my thing.
}
```

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
prprabhu-ms:
```
// TeamsCallAgent

protected declaratifyCall(call: CallCommon): DeclarativeTeamsCallAgent {
    if (isACSCall(call)) {
        raise Error('UNREACHABLE');
    }
    return teamsCallDeclaratify(call as TeamsCall);
}
```

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+    if (incomingCallSubscriber) {
+      incomingCallSubscriber.unsubscribe();
+      this._incomingCallSubscribers.delete(incomingCallId);
+    }
+    this._declarativeIncomingCalls.delete(incomingCallId);
+    this._context.setIncomingCallEnded(incomingCallId, callEndReason);
+  };
+
+  protected incomingCall = ({
+    incomingCall
```
prprabhu-ms, line 133: Let's make this use `CallCommon` as well.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    // For API extentions we need to have the call in the state when we are subscribing as we may want to update the
+    // state during the subscription process in the subscriber so we add the call to state before subscribing.
+    this._context.setCall(convertSdkCallToDeclarativeCall(call));
+    this._callSubscribers.set(call, new CallSubscriber(call, this._context, this._internalContext));
+    return this.getOrCreateDeclarativeCall(call);
+  };
+
+  private getOrCreateDeclarativeCall = (
+    call: Call | /* @conditional-compile-remove(teams-call) */ TeamsCall
```
prprabhu-ms, line 159: change args to `CallCommon`.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+  ): DeclarativeCallCommon => {
+    const declarativeCall = this._declarativeCalls.get(call);
+    if (declarativeCall) {
+      return declarativeCall;
+    }
+
+    /* @conditional-compile-remove(teams-call) */
+    if (!isACSCall(call)) {
+      const newDeclarativeCall = teamsCallDeclaratify(call, this._context);
```
prprabhu-ms, line 168: Use `protected this.declaratifyCall`

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    const newDeclarativeCall = callDeclaratify(call, this._context);
+    this._declarativeCalls.set(call, newDeclarativeCall as DeclarativeCall);
+    return newDeclarativeCall;
+  };
+
+  public getCommon<
+    AgentType extends CallAgent | /* @conditional-compile-remove(teams-call) */ TeamsCallAgent,
+    P extends keyof CallAgent
```
prprabhu-ms, line 180: Two reasons for referring to `CallAgent` and `TeamsCallAgent` here (as opposed to `CallAgentCommon`).

1. Generic implementation for both. You just need type constraints.
2. Is actual behavior difference (you call `isACS*`).

For 1, let's think if we can extract this to a separate file.
For 2, we need to use protected methods.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+  >(target: AgentType, prop: P): any {
+    switch (prop) {
+      case 'startCall': {
```
prprabhu-ms, line 183: Consider if this is not worth having a common implementation for.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+            return this.getOrCreateDeclarativeCall(call);
+          }
+
+          call = (target as CallAgent).startCall(...(args as Parameters<CallAgent['startCall']>));
+
+          this.addCall(call);
+          return this.getOrCreateDeclarativeCall(call);
+        }, 'CallAgent.startCall');
+      }
+      case 'join': {
```
prprabhu-ms, line 200: Same here. Consider pushing down the impl to specific classes.

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+      }
+      case 'on': {
+        return (...args: Parameters<CallAgent['on']> | Parameters<AgentType['on']>): void => {
+          const isCallsUpdated = args[0] === 'callsUpdated';
+          if (isCallsUpdated) {
+            const listener = args[1];
+            this._externalCallsUpdatedListeners.add(listener as CollectionUpdatedEvent<CallCommon>);
+          } else {
+            /* @conditional-compile-remove(teams-call) */
+            if (!isACSCallAgent(target)) {
```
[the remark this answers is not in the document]
prprabhu-ms: `protected forwardOn` `protected forwardOff` etc.

packages/calling-stateful-client/src/TeamsCallAgentDeclarative.ts
```diff
@@ -0,0 +1,99 @@
...
+export type DeclarativeTeamsCallAgent = TeamsCallAgent & /* @conditional-compile-remove(one-to-n-calling) */ {
+  /**
+   * @beta
+   * A readonly array that returns all the active `incomingCalls`.
+   * An active incoming call is a call that has not been answered, declined or disconnected.
+   *
+   * @Remark This attribute doesn't exist on the {@link @azure/communication-calling#CallAgent} interface.
+   * @returns readonly array of {@link DeclarativeIncomingCall}
+   */
+  incomingCalls: ReadonlyArray<DeclarativeIncomingCall>;
```
prprabhu-ms, line 28: This should be moved to `DeclarativeCallAgentCommon`

packages/calling-stateful-client/src/TeamsCallAgentDeclarative.ts
```diff
@@ -0,0 +1,99 @@
...
+/* @conditional-compile-remove(teams-call) */
+import { TeamsCallAgent as TeamsCallAgentBeta } from '@azure/communication-calling';
+import { ProxyCallAgentCommon } from './CallAgentDeclarativeCommon';
+import { CallContext } from './CallContext';
+/* @conditional-compile-remove(teams-call) */
+import { DeclarativeIncomingCall } from './IncomingCallDeclarative';
+import { InternalCallContext } from './InternalCallContext';
+import { disposeAllViews } from './StreamUtils';
+
+type TeamsCallAgent = never | /* @conditional-compile-remove(teams-call) */ TeamsCallAgentBeta;
```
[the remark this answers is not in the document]
prprabhu-ms: Consider `unknown`.

packages/calling-stateful-client/src/TeamsCallAgentDeclarative.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @param callAgent - TeamsCallAgent from SDK
+ * @param context - CallContext from StatefulCallClient
+ * @param internalContext- InternalCallContext from StatefulCallClient
+ */
+export const teamsCallAgentDeclaratify = (
+  callAgent: TeamsCallAgent,
+  context: CallContext,
+  internalContext: InternalCallContext
+): DeclarativeTeamsCallAgent => {
+  // Make sure there are no existing call data if creating a new TeamsCallAgentDeclarative (if creating a new
```
prprabhu-ms, line 89: Consider adding a common helper so this is not duplicated in `callAgentDeclaratify`.

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -500,7 +505,7 @@ export type CallErrorTarget =
   | 'CallAgent.on'
   | 'CallAgent.startCall'
   | 'CallClient.createCallAgent'
-  | /* @conditional-compile-remove(calling-beta-sdk) */ 'CallClient.createTeamsCallAgent'
+  | /* @conditional-compile-remove(teams-call) */ 'CallClient.createTeamsCallAgent'
```
prprabhu-ms, line 508: You should still have both conditional compilations.

packages/calling-stateful-client/src/Converter.ts
```diff
@@ -131,3 +142,23 @@ export function convertFromSDKToDeclarativeVideoStreamRendererView(
     target: view.target
   };
 }
+
+/**
+ * @private
+ */
+export const isACSCall = (call: Call | /* @conditional-compile-remove(teams-call) */ TeamsCall): call is Call => {
```
prprabhu-ms, line 149: This should take a `CallCommon`

packages/calling-stateful-client/src/Converter.ts
```diff
@@ -131,3 +142,23 @@ export function convertFromSDKToDeclarativeVideoStreamRendererView(
...
+  /* @conditional-compile-remove(teams-call) */
+  return call.kind === 'Call';
+  return true;
+};
+
+/**
+ * @private
+ */
+export const isACSCallAgent = (
```
prprabhu-ms, line 158: This should take `CallAgentKind`.

packages/calling-stateful-client/src/CallCommonDeclarative.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/* @conditional-compile-remove(teams-call) */
+import { CallCommon as CallCommonBeta } from '@azure/communication-calling';
+import { Call } from '@azure/communication-calling';
+import { CallContext } from './CallContext';
+
+/**
+ * @private
+ */
+export type CallCommon = Call | /* @conditional-compile-remove(teams-call) */ CallCommonBeta;
```
prprabhu-ms, line 12: Please add a comment describing what this black magic is, and why this is needed.

prprabhu-ms: This is da real black magic.

packages/calling-stateful-client/src/Converter.ts
```diff
@@ -7,8 +7,13 @@ import {
...
   LocalVideoStream as SdkLocalVideoStream,
   IncomingCall as SdkIncomingCall,
-  VideoStreamRendererView
+  VideoStreamRendererView,
+  Call,
+  CallAgent
 } from '@azure/communication-calling';
+
+/* @conditional-compile-remove(teams-call) */
+import { TeamsCall, TeamsIncomingCall, TeamsCallAgent } from '@azure/communication-calling';
```
prprabhu-ms, line 16: Try to get rid of all `Teams*` imports.

packages/calling-stateful-client/src/Converter.ts
```diff
@@ -110,7 +119,9 @@ export function convertSdkCallToDeclarativeCall(call: SdkCall): CallState {
 /**
  * @private
  */
-export function convertSdkIncomingCallToDeclarativeIncomingCall(call: SdkIncomingCall): DeclarativeIncomingCall {
+export function convertSdkIncomingCallToDeclarativeIncomingCall(
+  call: SdkIncomingCall | /* @conditional-compile-remove(teams-call) */ TeamsIncomingCall
```
prprabhu-ms, line 123: Use `IncomingCallCommon` with the black magic to map it to `IncomingCall` for stable.

prprabhu-ms: I'd say add a file StableCommonTypes.ts and have all SDK type mapping in that file.

packages/calling-stateful-client/src/IncomingCallDeclarative.ts
```diff
@@ -43,7 +45,10 @@ export class ProxyIncomingCall implements ProxyHandler<DeclarativeIncomingCall>
  * @param incomingCall - IncomingCall from SDK
  * @returns proxied IncomingCall
  */
-export const incomingCallDeclaratify = (incomingCall: IncomingCall, context: CallContext): DeclarativeIncomingCall => {
+export const incomingCallDeclaratify = (
+  incomingCall: IncomingCall | /* @conditional-compile-remove(teams-call) */ TeamsIncomingCall,
```
prprabhu-ms, line 49: Use `IncomingCallCommon`.

packages/calling-stateful-client/src/TypeAssertions.ts
```diff
@@ -32,6 +32,7 @@ type InferredCallErrorTargets =
    */
   | 'CallAgent.feature'
   | /* @conditional-compile-remove(calling-beta-sdk) */ 'CallClient.getEnvironmentInfo'
+  | /* @conditional-compile-remove(teams-call) */ 'TeamsCall.addParticipant'
```
prprabhu-ms, line 35: Add both conditional compilation features

/* @conditional-compile-remove(teams-call) @conditional-compile-remove(calling-beta-sdk) */

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,287 @@
...
+  convertSdkIncomingCallToDeclarativeIncomingCall,
+  isACSCall,
+  isACSCallAgent
+} from './Converter';
+import { DeclarativeIncomingCall, incomingCallDeclaratify } from './IncomingCallDeclarative';
+import { IncomingCallSubscriber } from './IncomingCallSubscriber';
+import { InternalCallContext } from './InternalCallContext';
+import { disposeAllViewsFromCall } from './StreamUtils';
+/* @conditional-compile-remove(teams-call) */
+import { DeclarativeTeamsCall, teamsCallDeclaratify } from './TeamsCallDeclarative';
```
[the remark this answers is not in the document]
PorterNan: Remove them as much as we can, still one teamsCallAgent is needed for "get" function because we need some type guard there for function names and parameters

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
PorterNan: Me either! Updating these to the new pattern we've discussed!

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
PorterNan: Tried this, looks like private function can't be accessed by child class in TS

PorterNan: *Even use super keyword*

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
PorterNan: Tried some ways to resolve this...

The problem here is we don't have startCall/join function in CallAgentCommon, but they do exist in CallAgent | TeamsCallAgent

I will leave it for this PR, we can do some iterations later once we find a better way to refactor this

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
PorterNan: Tried to push down implementation, ended up writing lots of duplicate code. 

So still followed the protected function pattern for here, which makes the common class much clearer :D

packages/calling-stateful-client/src/TeamsCallAgentDeclarative.ts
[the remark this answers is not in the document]
PorterNan: Moved all these types into a single file and add a common for the whole file :P

packages/calling-stateful-client/src/TeamsCallAgentDeclarative.ts
PorterNan: We don't expose DeclarativeCallAgentCommon as a public type, so we still need it for  DeclarativeCallAgentand  DeclarativeTeamsCallAgent, but I've extract IncomingCallManagement, as a shared type between these two

packages/calling-stateful-client/src/TeamsCallAgentDeclarative.ts
PorterNan: Hard to extract the last line because a lot of types involved, but I extract all the other lines into a common function in callAgentCommon

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
PorterNan: If you look at CallDeclarativeCommon  - things are a lot better, that is because we can almost find every function there in callCommon. For CallAgent.. I think we can't do much with it

packages/react-components/src/localization/locales/de-DE/strings.json
```diff
@@ -216,4 +216,4 @@
     "step3DigitText": "3",
     "step4DigitText": "4"
   }
-}
\ No newline at end of file
+}
```
PorterNan, line 219: This happened magically when I was merging back from main branch, is it prettier or some other things doing their job?

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -2,29 +2,37 @@
...
-   * An active incoming call is a call that has not been answered, declined or disconnected.
-   *
    * @Remark This attribute doesn't exist on the {@link @azure/communication-calling#CallAgent} interface.
    * @returns readonly array of {@link DeclarativeIncomingCall}
    */
   incomingCalls: ReadonlyArray<DeclarativeIncomingCall>;
 };
 
+/**
+ * @public
```
prprabhu-ms, line 30: nit: When you merge from `main` this type should itself become `@beta`.

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -57,6 +65,41 @@ class ProxyCallAgent extends ProxyCallAgentCommon implements ProxyHandler<Declar
     this.unregisterSubscriber();
   };
 
+  protected callDeclaratify(call: CallCommon, context: CallContext): DeclarativeCallCommon {
+    if (isACSCall(call)) {
+      return callDeclaratify(call, context);
+    }
+    throw new Error('Not reachable code for callDeclaratify.');
```
prprabhu-ms, line 72: Make the error string more actionable: `DaclarativeCallAgent.callDeclaratify must be called with an ACS call`.

same for errors thrown below.

packages/calling-stateful-client/src/Converter.ts
```diff
@@ -148,7 +141,7 @@ export function convertFromSDKToDeclarativeVideoStreamRendererView(
 /**
  * @private
  */
-export const isACSCall = (call: Call | /* @conditional-compile-remove(teams-call) */ TeamsCall): call is Call => {
+export const isACSCall = (call: CallCommon): call is Call => {
```
prprabhu-ms, line 144: Consider extracting the `isAcsCall*` type functions to their own file. That way `Converter.ts` can be completely agnostic of Teams/ACS differences. These functions aren't helping the conversion as such, and keeping them here increases the cognitive load reading understanding this file.

packages/react-components/src/localization/locales/de-DE/strings.json
prprabhu-ms: Unfortunate effect of merging string translations into `main` from another PR.
Wait for some other PR to take the brunt of fixing this when you merge from `main` again :P

prprabhu-ms (commented):
Getting there....

Let's wait for another set of eyes on this PR before iterating more with me.

packages/react-composites/src/composites/localization/locales/de-DE/strings.json
```diff
@@ -107,4 +107,4 @@
     "openDtmfDialpadLabel": "Wähltastatur anzeigen",
     "dtmfDialpadPlaceHolderText": "Nummer eingeben"
   }
-}
\ No newline at end of file
+}
```
JamesBurnside, line 110: Assume these were added by a code editor, they should be removed from this changelist - then file a bug to standardize and lint these files

tools/check-treeshaking/importChecker.js
```diff
@@ -4,15 +4,15 @@ const path = require('path');
 const forbiddenDependencies = require('./forbiddenDependencies');
 
 // thanks to https://github.com/webpack/webpack/issues/2090#issuecomment-302643018
-function findEntry(mod) {
-  if (mod.reasons.length > 0 && mod.reasons[0].module.resource) {
-    return findEntry(mod.reasons[0].module);
+function findEntry(mod, compilation) {
```
JamesBurnside, line 7: keep these changes local if they were required for building locally to avoid extra PR reviewing overhead

common/config/babel/.babelrc.js
```diff
@@ -16,27 +17,28 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
...
+        // PSTN calls 
+        'PSTN-calls',
         // rooms
         'rooms',
-        'call-readiness',
-        'unsupported-browser'
+        // Adhoc calls to a Teams user.
+        'teams-adhoc-call',
+        // Joining calls using teams token (CTE)
+        'teams-call',
```
JamesBurnside, line 38: nit: teams-call is ambiguous as a feature name (we can currently make teams calls, just not as a Teams identity)
```suggestion
        'teams-identity-support',
```
(or whatever the official standin name for CTE is)

tools/check-treeshaking/importChecker.js
PorterNan: The change has been merged in another PR, should be good once I merge main again

packages/react-composites/src/composites/localization/locales/de-DE/strings.json
PorterNan: image
Looks like it is running when you are doing a 'git commit' - how do we turn it off?

PorterNan: Found it in precommit config file

change/@internal-calling-stateful-client-eb705db4-7ad2-4024-a1e6-0d49a85188b2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Add TeamsCall and TeamsCallAgent to stateful - Add new beta feature flag \"teams-call\" - Extract common logic CallAgentCommon and CallCommon",
```
JamesBurnside, line 3: nit: these comments are for the public changelogs, `Add new beta feature flag "teams-call"` doesn't apply for external developers of a changelog

JamesBurnside: nit: `stateful` -> `StatefulCallClient`

change/@internal-react-components-3cfd175a-bccd-43d5-b2bd-d9b8e0ee7b29.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Add TeamsCall and TeamsCallAgent to stateful - Add new beta feature flag \"teams-call\" - Extract common logic CallAgentCommon and CallCommon",
+  "packageName": "@internal/react-components",
```
JamesBurnside, line 4: looks like there's no longer changes to this package - remove

change/@internal-react-composites-0d94c318-0522-4ac7-a3b4-dc495fb5e32c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Add TeamsCall and TeamsCallAgent to stateful - Add new beta feature flag \"teams-call\" - Extract common logic CallAgentCommon and CallCommon",
+  "packageName": "@internal/react-composites",
```
JamesBurnside, line 4: looks like there's no longer changes to this package - remove (or change to `none` if the update to the tests folder triggers this)

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
```diff
@@ -100,6 +100,7 @@ export interface CallState {
     startTime: Date;
     state: CallState_2;
     transcription: TranscriptionCallFeature;
+    type: 'Teams' | 'ACS';
```
JamesBurnside, line 103: Can `Teams` be conditional-compiled out?

packages/calling-stateful-client/src/CallAgentDeclarative.ts
```diff
@@ -2,29 +2,37 @@
...
+/* @conditional-compile-remove(one-to-n-calling) */
 /**
- * @public
- * `DeclarativeCallAgent` extends and proxies the {@link @azure/communication-calling#CallAgent}
+ * @beta
+ * This contains a readonly array that returns all the active `incomingCalls`.
+ * An active incoming call is a call that has not been answered, declined or disconnected.
  */
-export type DeclarativeCallAgent = CallAgent & /* @conditional-compile-remove(one-to-n-calling) */ {
+export type IncomingCallManagement = {
```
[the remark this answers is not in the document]
PorterNan: The sad story is interface does not support 'da black magic' of conditional compilation :(

packages/calling-stateful-client/src/BetaToStableTypes.ts
```diff
@@ -0,0 +1,47 @@
...
+  CallCommon as CallCommonBeta,
+  CallAgentCommon as CallAgentCommonBeta,
+  TeamsCallAgent as TeamsCallAgentBeta,
+  IncomingCallCommon as IncomingCallCommonBeta
+} from '@azure/communication-calling';
+
+/**
+ * @private
+ */
+export type CallCommon = Call | /* @conditional-compile-remove(teams-identity-support) */ CallCommonBeta;
```
JamesBurnside, line 25: Will this become the calling sdk `CallCommon` type when the beta is removed?

JamesBurnside: (likewise with the other 'Commons' in this file?)

packages/calling-stateful-client/src/CallDeclarative.ts
```diff
@@ -16,69 +17,19 @@ export interface DeclarativeCall extends Call {
...
+          ...args: Parameters<Call['addParticipant']>
+        ) {
           return await target.addParticipant(...args);
-        }, 'Call.addParticipant');
+        },
+        'Call.addParticipant');
       }
       default:
-        return Reflect.get(target, prop);
+        return super.get(target, prop as any);
```
JamesBurnside, line 32: Why is `as any` required here?

packages/calling-stateful-client/src/CallDeclarative.ts
```diff
@@ -99,5 +50,5 @@ export const callDeclaratify = (call: Call, context: CallContext): DeclarativeCa
     value: () => proxyCall.unsubscribe()
   });
 
-  return new Proxy(call, proxyCall) as DeclarativeCall;
+  return new Proxy(call, proxyCall) as any as DeclarativeCall;
```
JamesBurnside, line 53: similar here, why as any?

packages/calling-stateful-client/src/TeamsCallAgentDeclarative.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  protected callDeclaratify(call: CallCommon, context: CallContext): DeclarativeCallCommon {
+    /* @conditional-compile-remove(teams-identity-support) */
+    if (isTeamsCall(call)) {
+      return teamsCallDeclaratify(call, context);
+    }
+    throw new Error('Not reachable code, DeclarativeTeamsCallAgent.callDeclaratify must be called with an TeamsCall.');
+  }
+
+  protected startCall(agent: CallAgentCommon, args: unknown[]): CallCommon {
```
JamesBurnside, line 68: why are all these `unknown`? This will become really difficult in the future for developers to use these args correctly

JamesBurnside (commented):
Did a quick pass, trying to wrap my head around the agent inheritance, I'm concerned about the runtime type checks that are happing (`isACSCallAgent`) coupled with the `any`s and `unknown`s

packages/calling-stateful-client/src/TeamsCallAgentDeclarative.ts
JamesBurnside: (e.g. why aren't these `args: Parameters<TeamsCallAgent['startCall']>`)

JamesBurnside: Oh because of the runtime check `isTeamsCallAgent` I see.. let me think on this more

packages/calling-stateful-client/src/BetaToStableTypes.ts
```diff
@@ -0,0 +1,41 @@
...
+  TeamsCall as TeamsCallBeta,
+  CallCommon as CallCommonBeta,
+  CallAgentCommon as CallAgentCommonBeta,
+  TeamsCallAgent as TeamsCallAgentBeta,
+  IncomingCallCommon as IncomingCallCommonBeta
+} from '@azure/communication-calling';
+
+/**
+ * @private
+ */
```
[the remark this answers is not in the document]
PorterNan: Added an example explanation at the top :D

packages/calling-stateful-client/src/CallDeclarative.ts
PorterNan: Cause props for parent class and this child class are not inferred correctly for typescript(it's hard to make it right from complier side though)

packages/calling-stateful-client/src/TeamsCallAgentDeclarative.ts
PorterNan: These are args coming from parent class override parameters, from parent class, it's hard to determine args(or the type would be super complicated and unreadable) because they are generic one

PorterNan: Update:
New typescript support to do that only in child class, so I can actually type the parameter in child class but still keep parent class the same

packages/calling-stateful-client/src/CallDeclarative.ts
PorterNan: Hmmm... forget why added this(I guess it was because there was some type problems there before I complete the whole type dependencies), removed

packages/calling-stateful-client/src/BetaToStableTypes.ts
PorterNan: In beta build
Call | CallCommon = CallCommon (because Call is child class of CallCommon)

In stable build 
It is just simply Call type

Once everything is in stable, this will be still running fine
But we can remove these types and import all "commons" from sdk directly to make it perfect once we get there (just like what comments say at the top of file)

packages/calling-stateful-client/src/BetaToStableTypes.ts
```diff
@@ -0,0 +1,47 @@
...
+export type CallAgentCommon = CallAgent | /* @conditional-compile-remove(teams-identity-support) */ CallAgentCommonBeta;
+
+/**
+ * @private
+ */
+export type TeamsCall = never | /* @conditional-compile-remove(teams-identity-support) */ TeamsCallBeta;
```
mgamis-msft, line 35: Can't we just conditionally compile this export since this won't be imported from BetaToStableStypes.ts in stable

mgamis-msft: Is it because we can't conditionally compile the file TeamsCallAgentDeclarative.ts where TeamsCallAgent is imported?

PorterNan: If we do that for the whole file, we will have to conditional compile every line of that file

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
JamesBurnside: From what I'm reading a callAgent is always a CallAgent or TeamsCallAgent, why do we use CallAgentCommon and not `CallAgent | TeamsCallAgent` (that would contain the correct overlaps)?

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -0,0 +1,248 @@
...
+          const call = this.joinCall(target, args);
+          this.addCall(call);
+          return this.getOrCreateDeclarativeCall(call);
+        }, 'CallAgent.join');
+      }
+      case 'calls': {
+        return Array.from(this._declarativeCalls.values());
+      }
+      case 'on': {
+        return (...args: Parameters<CallAgent['on']> | Parameters<AgentType['on']>): void => {
```
JamesBurnside, line 180: This is the only function with `Parameters<CallAgent['PROP']>` (the rest only have `Parameters<AgentType['PROP']>` - can this be:
```suggestion
        return (...args: Parameters<AgentType['on']>): void => {
```

packages/calling-stateful-client/src/TypeGuards.ts
```diff
@@ -0,0 +1,41 @@
...
+  return call.kind === 'TeamsCall';
+  return false;
+};
+
+/**
+ * @private
+ */
+export const isTeamsCallAgent = (callAgent: CallAgentCommon): callAgent is TeamsCallAgent => {
+  /* @conditional-compile-remove(teams-identity-support) */
+  return callAgent.kind === 'TeamsCallAgent';
```
JamesBurnside, line 39: My understanding is this technically isn't true. `callAgent` here is of type `CallAgentCommon`, but CallAgentCommon is not a TeamsCallAgent:
image
Although _we_ always are using it as either a (ACS)CallAgent or TeamsCallAgent this function could take any call agent common and it will only check the `kind` property then incorrectly typecast the entire object.

(This is a continuation of this comment: https://github.com/Azure/communication-ui-library/pull/2396/files#r1010585946 where I'm interested to know why we use `CallAgentCommon` and not `CallAgent | TeamsCallAgent` instead)

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
JamesBurnside: This might need a separate comment thread but I want to confirm we _want_ a common abstract class. We're seemingly just using it to dedupe code? But going by the fact we have runtime checks to access properties the CallAgentCommon does not have would suggest there isn't enough overlap in these two classes to define them as two types of the same thing and warrant a base class.
Instead when common code exists between the two classes we could have pure helper functions.

I'm trying to find some good online resources on one vs the other but don't have too much time to invest in it and not finding much of use - but I'd love to hear your guys' opinions on this. (perhaps + @DominikMe)

PorterNan: This is the long standing questions for every engineers - inheritance or composition. 

For inheritance, it adds more constrains and limits your logic all aligned to base class

For composition, it gives you more flexibility to grab those functions and use those in some other classes

The good news this is just the implementation details, for current scale of implementation, I would say both are good choice, composition implementation requires lots of big if else to make things happens, and that's why I chose more constrains from abstract class so I don't miss any part for CTE. 

if we are going to have 3 or 4+ classes or wanna reuse these startCall/join logics inside another class (just like callContext), we can always come back and refactor them into composition again

packages/calling-stateful-client/src/TypeGuards.ts
PorterNan: This is a good point which I was thinking of for a quite while.

I was using CallAgent | TeamsCallAgent as parameter at the very beginning, but found it a pain to implement it in an elegant way. This was actually modified after I had first iteration review with @prprabhu-ms. There are a few reason we use CallAgentCommon here:

1. Using CallAgent | TeamsCallAgent as parameters requires us to pass it from parent type as CallAgent | TeamsCallAgent, which means there must be a type cast somewhere from the caller (cause it should be callAgentCommon at very top level), which means this type guard loses 50% of its functionality. (and type cast will highly depend on .kind property still)

2. This is a fairly common trick to do both type check both in js and ts. Kind property exists in callCommon, and the only reason it exists there is for type cast(or we say runtime type check for js), if the kind value indicates a wrong type, I would say it is a serious bug coming from our calling sdk

In short, any assumption to check one or two properties then do type cast is not safe in the end, unless you check all the properties there and check all their type. So this is just another "better typescript version" of js runtime check

prprabhu-ms (approved):

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
```diff
@@ -111,10 +112,8 @@ export type CreateViewResult = {
     view: VideoStreamRendererView;
 };
 
-// @public
-export type DeclarativeCallAgent = CallAgent & /* @conditional-compile-remove(one-to-n-calling) */ {
-    incomingCalls: ReadonlyArray<DeclarativeIncomingCall>;
-};
+// @beta
+export type DeclarativeCallAgent = CallAgent & IncomingCallManagement;
```
prprabhu-ms, line 116: Rename to `StatefulCallAgent`. Not sure whether this PR added it or just an artifact of needing to merge from `main`.

PorterNan: image
Looks like it is DeclarativeCallAgent name in the main (I only extra the IncomingCallManagement ckass, but not renaming)
I can create a PR to rename this if we want (or we have a PR resolving this but not merged?)

packages/calling-stateful-client/src/TypeGuards.ts
JamesBurnside: Can you point to where the typecast from `CallAgent | TeamsCallAgent` to `CallAgentCommon` would be needed?
`CallAgent` and `TeamsCallAgent` should be a superset of `CallAgentCommon` (they should extend call agent common) so anywhere requiring a CallAgentCommon could take a CallAgent or TeamsCallAgent without issue no?

PorterNan: Sry, my description above isn't clear enough.

It's not casting CallAgent | TeamsAgent to AgentCommon
It's casting AgentCommon to CallAgent | TeamsAgent

Because at some point of ancestor of functions who calls isACSAgent, the type is CallAgentCommon (because everything start with common code then we make it specify later using if condition)

If this type guard uses CallAgent | TeamsAgent, that means we will have to cast CallAgentCommon to CallAgent | TeamsAgent before we call this typeGuard(which is actually what this type guard for), otherwise it wouldn't satisfy type check

PorterNan: One way to make this work is never using CallAgentCommon in ancestor but that makes typing much complicated than this one(you can check my first several commit, which I used CallAgent | TeamsAgent for these functions)

JamesBurnside: Have to step out but will be back in 30mins, Looking at the old commits:
image

Why are we type casting in the Common, and not having the CallAgent or the TeamsCallAgent house their specific implementations?

PorterNan: That's what we are doing for current implementation, you can see the only difference here is those if-elses, so we forward that child class

There will be lots of CallCommon/CallAgentCommon usage in handler and Adapter layer which requires some minor run time checks, the need of converting CallAgentCommon to CallAgent or TeamsAgent is not avoidable, 

Back to typeguard itself, that's the intention of why sdk put kind property in their common object not only in TeamsAgent | CallAgent, so this is just a way to leverage that pattern in typescript (I do think sdk should provide us the type guard function)

JamesBurnside (approved):

PR #2399, opened 2022-10-12
Add missing leaveCall(foreveryone:true) logic to adapter.leavecall

carocao-msft:
# What
Add missing leaveCall(foreveryone:true) logic to adapter.leavecall

# Why
[BUG 3010555](https://skype.visualstudio.com/SPOOL/_workitems/edit/3010555)

# How Tested
Add those lines of code in afterCreate in callscreen

      const leaveCallFunc= adapter.leaveCall.bind(adapter);
      adapter.leaveCall = async () => {
            await leaveCallFunc(true);
      };
Run calling sample/ callwithchat to see if clicking on endcall will endcall for everyone

# Process & policy checklist

### DIFF
--- a/change/@internal-calling-component-bindings-5ea8a8ee-352d-4df0-b056-bb3c982e0fba.json
+++ b/change/@internal-calling-component-bindings-5ea8a8ee-352d-4df0-b056-bb3c982e0fba.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Add missing foreveryone logic for leaveCall",
+  "comment": "BugFix: Add missing foreveryone logic for leaveCall",
   "packageName": "@internal/calling-component-bindings",
   "email": "carolinecao@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@internal-calling-component-bindings-5ea8a8ee-352d-4df0-b056-bb3c982e0fba.json
+++ b/change/@internal-calling-component-bindings-5ea8a8ee-352d-4df0-b056-bb3c982e0fba.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add missing foreveryone logic for leaveCall",
+  "packageName": "@internal/calling-component-bindings",
+  "email": "carolinecao@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@internal-react-composites-96700b47-d167-4a39-abc0-83c26a59d859.json
+++ b/change/@internal-react-composites-96700b47-d167-4a39-abc0-83c26a59d859.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Add missing foreveryone logic for leaveCall",
+  "comment": "BugFix: Add missing foreveryone logic for leaveCall",
   "packageName": "@internal/react-composites",
   "email": "carolinecao@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@internal-react-composites-96700b47-d167-4a39-abc0-83c26a59d859.json
+++ b/change/@internal-react-composites-96700b47-d167-4a39-abc0-83c26a59d859.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add missing foreveryone logic for leaveCall",
+  "packageName": "@internal/react-composites",
+  "email": "carolinecao@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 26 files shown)


packages/react-components/src/components/EndCallButton.tsx
```diff
@@ -63,10 +62,16 @@ export const EndCallButton = (props: EndCallButtonProps): JSX.Element => {
...
+    if (props.onHangUp) {
+      props.onHangUp();
+    }
+  };
+
   return (
     <ControlBarButton
       {...props}
-      onClick={props.onHangUp ?? props.onClick}
+      onClick={props.onHangUp ? onHangUp : props.onClick}
```
JamesBurnside, line 74: we are checking `props.onHangUp` twice, once here and once in `onHangUp`. Can changes to this file be reverted? (is there any change in behavior in this file?)

JamesBurnside (approved):

packages/react-components/src/components/EndCallButton.tsx
carocao-msft: in onHangup is checking and calling onHangup, in here props.onHangUp ? onHangUp : props.onClick is if props.onHangUp is not passed in then don't use onHangup function, use onClick instead

carocao-msft: No changes in behaviour, previously it's onClick = {props.onHangup?? props.onClick}, however now onHangup has an optional prop so we can't assign it straight to onCLick anymore, need to do something like onClick = () =>{ props.onHangup()} instead

JamesBurnside: Something to look at here is currently, using this component, how would someone use onHangUp here to call `forEveryone` as `true`? This currently always calls `props.onHangUp()` with no parameter.

Instead you'll want this pure component to always have onHangUp be a simple function that is just `onHangUp?: () => Promise<void>;`

It is then up to the parent of the pure EndCallButton component to decide how the hang up function is implemented, and if it kicks everyone out of the call or not`

carocao-msft: The challenge here is onhangeup is grabbed by usepropsfor so even if I set it up as onHangUp?: () => Promise<void>, what's passed down to endcallButton is still going to be onHangUp?: (foreveryone) => Promise<void>;

JamesBurnside (approved):

change/@internal-calling-component-bindings-5ea8a8ee-352d-4df0-b056-bb3c982e0fba.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add missing foreveryone logic for leaveCall",
```
anjulgarg, line 3: should ideally be 'none' since it's a bug fix.
Starting the comment with the text "Bugfix" makes it easier to tweak the changeling during release.

JamesBurnside: bug fixes should be of type `patch` (`none` excludes them from changelog generation)

packages/react-components/src/components/EndCallButton.tsx
```diff
@@ -63,10 +62,16 @@ export const EndCallButton = (props: EndCallButtonProps): JSX.Element => {
     styles ?? {}
   );
 
+  const onHangUp = (): void => {
```
anjulgarg, line 65: What is the point of this function?

anjulgarg (approved):

packages/react-components/src/components/EndCallButton.tsx
carocao-msft: This function is there so we can call onHangup() with no params then pass it to onClick
We cannot pass onHangup directly to onClick anymore since the parent component is passing in onHangup(foreveryone: boolean) into endcallbutton

change/@internal-calling-component-bindings-5ea8a8ee-352d-4df0-b056-bb3c982e0fba.json
anjulgarg: Ah yes. My bad.
This is correct.

change/@internal-react-composites-96700b47-d167-4a39-abc0-83c26a59d859.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "BugFix: Add missing foreveryone logic for leaveCall",
```
prprabhu-ms, line 3: This should have been `minor` because it changes the stable API.

PR #2403, opened 2022-10-12
[PSTN 1-N calling] Update lobby page with new strings

dmceachernmsft:
# What

PSTN
image
ACS
image
# Why

Communicates to the end user who they are calling in the Call composite lobby screen

https://skype.visualstudio.com/SPOOL/_workitems/edit/2968937
# How Tested

Ran locally with PSTN and ACS call

### DIFF
--- a/change/@internal-react-composites-7452c589-eca7-482e-b144-0572503ede0e.json
+++ b/change/@internal-react-composites-7452c589-eca7-482e-b144-0572503ede0e.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Update lobby screen to reflect to the user who they are calling.",
+  "packageName": "@internal/react-composites",
+  "email": "94866715+dmceachernmsft@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@internal-react-composites-7452c589-eca7-482e-b144-0572503ede0e.json
+++ b/change/@internal-react-composites-7452c589-eca7-482e-b144-0572503ede0e.json
@@ -1,6 +1,6 @@
 {
   "type": "prerelease",
-  "comment": "Update lobby screen to reflect to the user who they are calling.",
+  "comment": "Update lobby screen to reflect to the user who they are calling in a  1:1 PSTN or ACS call.",
   "packageName": "@internal/react-composites",
   "email": "94866715+dmceachernmsft@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/packages/react-composites/src/composites/CallComposite/Strings.tsx
+++ b/packages/react-composites/src/composites/CallComposite/Strings.tsx
@@ -322,8 +322,7 @@ export interface CallCompositeStrings {
    * Control bar People button ToolTipContent
    */
   peopleButtonTooltipClose: string;
-  /* @conditional-compile-remove(PSTN-calls) */
-  /* @conditional-compile-remove(1-to-n-calling) */
+  /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
   /**
    * label for when making a outbound call
    *
--- a/packages/react-composites/src/composites/CallComposite/Strings.tsx
+++ b/packages/react-composites/src/composites/CallComposite/Strings.tsx
@@ -322,4 +322,12 @@ export interface CallCompositeStrings {
    * Control bar People button ToolTipContent
    */
   peopleButtonTooltipClose: string;
+  /* @conditional-compile-remove(PSTN-calls) */
+  /* @conditional-compile-remove(1-to-n-calling) */
+  /**
+   * label for when making a outbound call
+   *
+   * Will either be the title or the description based on the type of call you are making.
+   */
+  outboundCallingNoticeString: string;
 }
--- a/packages/react-composites/src/composites/CallComposite/Strings.tsx
+++ b/packages/react-composites/src/composites/CallComposite/Strings.tsx
@@ -324,9 +324,7 @@ export interface CallCompositeStrings {
   peopleButtonTooltipClose: string;
   /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
   /**
-   * label for when making a outbound call
-   *
-   * Will either be the title or the description based on the type of call you are making.
+   * Label disaplayed on the lobby screen during a 1:1 outbound call.
    */
   outboundCallingNoticeString: string;
 }
--- a/packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
+++ b/packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
@@ -14,11 +14,11 @@ import { CallCompositeStrings } from '../Strings';
 import { useLocale } from '../../localization';
 import { useLocalVideoStartTrigger } from '../components/MediaGallery';
 import { CallCompositeIcon } from '../../common/icons';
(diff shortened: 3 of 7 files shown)


packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -322,4 +322,11 @@ export interface CallCompositeStrings {
    * Control bar People button ToolTipContent
    */
   peopleButtonTooltipClose: string;
+  /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
+  /**
+   * label for when making a outbound call
```
anjulgarg, line 327:
```suggestion
   * Label displayed on the lobby screen during a 1:1 outbound call.
```

packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -322,4 +322,11 @@ export interface CallCompositeStrings {
...
+   *
+   * Will either be the title or the description based on the type of call you are making.
```
anjulgarg, line 329: I think this comment adds confusion. Maybe it's better to omit this line.
OR change it such that it's easier to understand.

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
```diff
@@ -56,14 +72,31 @@ export const LobbyPage = (props: LobbyPageProps): JSX.Element => {
...
     />
   );
 };
 
-const overlayProps = (strings: CallCompositeStrings, inLobby: boolean): LobbyOverlayProps =>
-  inLobby ? overlayPropsWaitingToBeAdmitted(strings) : overlayPropsConnectingToCall(strings);
+const overlayProps = (
+  strings: CallCompositeStrings,
+  inLobby: boolean,
+  remoteParticipants: RemoteParticipantState[] | undefined
```
anjulgarg, line 86:
```suggestion
  remoteParticipants?: RemoteParticipantState[]
```

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
```diff
@@ -56,14 +72,31 @@ export const LobbyPage = (props: LobbyPageProps): JSX.Element => {
...
+): LobbyOverlayProps => {
+  /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
+  const outboundCallParticipant = remoteParticipants !== undefined ? remoteParticipants[0] : undefined;
```
anjulgarg, line 89: Slightly more terse
`const outboundCallParticipant = remoteParticipants ? remoteParticipants[0] : undefined;`

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
```diff
@@ -36,6 +41,17 @@ export const LobbyPage = (props: LobbyPageProps): JSX.Element => {
   const callState = useSelector(getCallStatus);
   const inLobby = callState === 'InLobby';
 
+  /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
+  const participants = useSelector(getRemoteParticipants) ?? {};
+
+  const remoteParticipantsTrampoline = (): RemoteParticipantState[] | undefined => {
```
prprabhu-ms, line 47: Define this as a file-level free function. Currently, it's an anonymous function object redefined on each render.
Also, all these function definitions inline make the Component code harder to read.

prprabhu-ms: Also, move the `useSelector` call inside the trampoline so you need only one conditional compilation directive.

anjulgarg, line 47: Do we need this trampoline?
I thought remoteParticipants are only available for outbound 1:1 or PSTN calls?

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
```diff
@@ -56,14 +72,31 @@ export const LobbyPage = (props: LobbyPageProps): JSX.Element => {
...
 };
 
-const overlayProps = (strings: CallCompositeStrings, inLobby: boolean): LobbyOverlayProps =>
-  inLobby ? overlayPropsWaitingToBeAdmitted(strings) : overlayPropsConnectingToCall(strings);
+const overlayProps = (
+  strings: CallCompositeStrings,
+  inLobby: boolean,
+  remoteParticipants: RemoteParticipantState[] | undefined
+): LobbyOverlayProps => {
+  /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
```
prprabhu-ms, line 88: Remove all conditional compilation directives from this function. `remoteParticipants` will always be undefined for stable build.

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
```diff
@@ -76,3 +109,17 @@ const overlayPropsWaitingToBeAdmitted = (strings: CallCompositeStrings): LobbyOv
   moreDetails: strings.lobbyScreenWaitingToBeAdmittedMoreDetails,
   overlayIcon: <CallCompositeIcon iconName="LobbyScreenWaitingToBeAdmitted" />
 });
+
+/* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
```
prprabhu-ms, line 113: Remove this conditional compilation directive. This function will be used unconditionally if you accept the suggestion above.

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
prprabhu-ms: Why just the first participant? Should the notice list all the outbound participants being called?
If there is a reason we only name the first participant, please add a comment in the code explaining that.

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
```diff
@@ -76,3 +109,17 @@ const overlayPropsWaitingToBeAdmitted = (strings: CallCompositeStrings): LobbyOv
...
+const overlayPropsOutboundCall = (
+  strings: CallCompositeStrings,
+  participant: RemoteParticipantState
+): LobbyOverlayProps => {
+  const title = isPhoneNumberIdentifier(participant.identifier)
```
prprabhu-ms, line 118: nit: The logic will be more readable if you separate out the two cases entirely:

```
if (isPhoneNumerIdentifier(participant.identifier)) {
  return {
    title: ...
    moreDetails: ...
  }
}

return {
  title: ...
  moreDetails: ...
}
```

anjulgarg (approved):

prprabhu-ms (approved):

change/@internal-react-composites-7452c589-eca7-482e-b144-0572503ede0e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Update lobby screen to reflect to the user who they are calling.",
```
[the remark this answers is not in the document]
dmceachernmsft: the lobby page also handles the 'connecting' state that users land in for when they are joining that call. this page does seem to be pulling double duty for what it is called. we can't change the name of the page right since we expose that state in the API but it seems to be more of a `startingCall` screen or something?

PR #2404, opened 2022-10-13
Suppress camera access error bar error for consumer

mgamis-msft:
# What

Suppress camera access error bar error for consumer.

The errorBarSelector pushes the error message with type 'callCameraAccessDenied' when the DeviceManagerState property deviceAccess.video is false. It is NOT from Calling SDK UFD

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3015756

# How Tested

https://user-images.githubusercontent.com/79475487/195479583-c808c065-a7e4-4be1-b5f6-3a7f9ccd8aa2.mp4

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
+++ b/packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
@@ -64,6 +64,7 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
   const rolePermissions = _usePermissions();
 
   /* @conditional-compile-remove(rooms) */
+  // TODO: move this logic to the error bar selector once role is plumbed from the headless SDK
   if (!rolePermissions.cameraButton) {
     errorBarProps = {
       ...errorBarProps,
--- a/packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
+++ b/packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
@@ -63,6 +63,12 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
   /* @conditional-compile-remove(rooms) */
   const rolePermissions = _usePermissions();
   /* @conditional-compile-remove(rooms) */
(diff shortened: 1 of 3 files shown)


packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -63,6 +63,12 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
   /* @conditional-compile-remove(rooms) */
   const rolePermissions = _usePermissions();
   /* @conditional-compile-remove(rooms) */
+  if (!rolePermissions.cameraButton) {
```
prprabhu-ms, line 66: This splits the business logic of what errors to show between the error bar selector and the configuration page UI component.

The right approach here is to plumb the role from stateful layer into the error bar selector and act on it there.

I understand that for the first public preview, role is provided as a composite argument, so it isn't available to the selector. But this is a very short-term stop-gap.

Please add a TODO comment here to move this logic to the error bar selector once role is plubmed in from headless calling SDK, and create a task for the same under an appropriate user story.

prprabhu-ms (approved):
Nit: I'd remove `[BETA-BLOCKING]` from the PR title. It's not very useful when looking at old PR history.

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
JamesBurnside: (somewhat tangential, we need to know the role in the configuration page (i.e. before a call is joined), which sounds like it won't be possible to come from stateful client? One possibility would be to pass into the adapter ctor though and use in the adapter selectors)

mgamis-msft: Ok I will add a TODO comment when the role is available in the stateful client. 

That's a good thought on the availability of the role in the configuration page. I will have to give it some thought when we remove role from the CallComposite props.

prabhjot-msft (approved):

PR #2410, opened 2022-10-14
Bugfix for Config screen flashing for 2 seconds after leaving a call

anjulgarg:
# What
Bugfix for Config screen flashing for 2 seconds after leaving a call

# Why
This bug happens because we set the callId to undefined in the adapter inside the leaveCall function.
However, call client state takes a while to add that call to endedCalls array and remove it from the active call.
Hence, we need to only set call id to undefined once that operation has completed.

# How Tested
Manually

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
+++ b/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
@@ -128,9 +128,9 @@ class CallContext {
     const oldPage = this.state.page;
     const newPage = getCallCompositePage(call, latestEndedCall);
     if (!IsCallEndedPage(oldPage) && IsCallEndedPage(newPage)) {
-      this.emitter.emit('callEnded', {
-        callId: this.callId
-      });
+      this.emitter.emit('callEnded', { callId: this.callId });
+      // Reset the callId to undefined as the call has ended.
+      this.setCurrentCallId(undefined);
     }
 
     if (this.state.page) {
@@ -381,11 +381,8 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
   public async leaveCall(forEveryone?: boolean): Promise<void> {
     await this.handlers.onHangUp(forEveryone);
     this.unsubscribeCallEvents();
-    this.call = undefined;
     this.handlers = createDefaultCallingHandlers(this.callClient, this.callAgent, this.deviceManager, undefined);
-    this.context.setCurrentCallId(undefined);
-    // Resync state after callId is set
-    this.context.updateClientState(this.callClient.getState());
+    this.call = undefined;
     this.stopCamera();
     this.mute();
   }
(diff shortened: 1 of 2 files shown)


JamesBurnside (approved):
LGTM! Note we should keep this bug active on calling: https://skype.visualstudio.com/SPOOL/_workitems/edit/3006714 @anjulgarg if you found any extra information when debugging this can you add it? From trying this out, at a quick glance, it looks like we simply do not get the callendreason for a long time after clicking hangup so there's a long delay before the user is taken to the end call screen.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -381,11 +381,8 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
   public async leaveCall(forEveryone?: boolean): Promise<void> {
     await this.handlers.onHangUp(forEveryone);
     this.unsubscribeCallEvents();
-    this.call = undefined;
     this.handlers = createDefaultCallingHandlers(this.callClient, this.callAgent, this.deviceManager, undefined);
-    this.context.setCurrentCallId(undefined);
-    // Resync state after callId is set
-    this.context.updateClientState(this.callClient.getState());
+    this.call = undefined;
```
prprabhu-ms, line 385: Should `this.call` also be reset on line 133 then.

This is creating a disconnect between `this.callId` and `this.call`. That might cause other bugs, or is difficult to understand / remember at the least.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -128,9 +128,9 @@ class CallContext {
     const oldPage = this.state.page;
     const newPage = getCallCompositePage(call, latestEndedCall);
     if (!IsCallEndedPage(oldPage) && IsCallEndedPage(newPage)) {
-      this.emitter.emit('callEnded', {
-        callId: this.callId
-      });
+      this.emitter.emit('callEnded', { callId: this.callId });
+      // Reset the callId to undefined as the call has ended.
+      this.setCurrentCallId(undefined);
```
prprabhu-ms, line 133: Do you no longer need to resync state from stateful layer after the `callId` is set (i.e., what happened to the old line 388)?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
anjulgarg: this.call is part of the call adapter, this.callId is part of the callContext

prprabhu-ms:
LGTM to unblock beta.

Please re-request review if you make non-trivial changes based on the discussions we had in a sidebar.

prprabhu-ms (approved):

PR #2411, opened 2022-10-14
Fix group call calling lobby screen strings

dmceachernmsft:
# What

Bug introduced by https://github.com/Azure/communication-ui-library/pull/2403 was causing the ACS call string to render for a second before joining a groupCall. 
# Why

stops flash from happening when joining a group call.

related to: https://skype.visualstudio.com/SPOOL/_workitems/edit/2968937 
# How Tested

Ran locally in all calling modalities


packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
```diff
@@ -71,14 +71,21 @@ export const LobbyPage = (props: LobbyPageProps): JSX.Element => {
...
+   *
+   * We also need to check the participant state since in a group call the remote participants array will populate just before
+   * the user joins.
    */
-  const outboundCallParticipant = remoteParticipants ? remoteParticipants[0] : undefined;
+  const outboundCallParticipant: RemoteParticipantState =
+    remoteParticipants[0] && ['Ringing', 'Connecting'].includes(remoteParticipants[0].state)
+      ? remoteParticipants[0]
+      : undefined;
+  outboundCallParticipant && console.log(outboundCallParticipant.state);
```
JamesBurnside, line 88: remove console log here

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
```diff
@@ -71,14 +71,21 @@ export const LobbyPage = (props: LobbyPageProps): JSX.Element => {
...
    * Only grab the first participant because there will only be one in this situation.
    * when starting a call with multiple people the call goes to the connected state and composite goes directly to
    * videoGallery.
+   *
+   * We also need to check the participant state since in a group call the remote participants array will populate just before
+   * the user joins.
    */
-  const outboundCallParticipant = remoteParticipants ? remoteParticipants[0] : undefined;
+  const outboundCallParticipant: RemoteParticipantState =
+    remoteParticipants[0] && ['Ringing', 'Connecting'].includes(remoteParticipants[0].state)
```
JamesBurnside, line 85: Why do we check only remoteParticipant `0` here?

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
JamesBurnside, line 88: Is this PR missing a commit? It tries grab a outboundCallParticipant then just console logs it currently

packages/react-composites/src/composites/CallComposite/pages/LobbyPage.tsx
dmceachernmsft: you will only be in this position should you be calling 1 person. if you start a call with multiple people you go straight to the connected state and go to the video gallery.

JamesBurnside: Does `Connecting` denote specifically a `Calling...` state?

JamesBurnside: > you will only be in this position should you be calling 1 person. if you start a call with multiple people you go straight to the connected state and go to the video gallery.

Can you update this check to do that? Without it we are relying on (somewhat undocumented) internal calling behavior. Alternatively could this check check that _all_ remote participants are in the Ringing state?

dmceachernmsft: yes, this is because when you are joining a groupCall, just before you connect you will get a participants array from the calling SDK but the state of the participants here will be `Idle`

JamesBurnside: In the `call.state`, there's a property `ringing` - is there a reason to not use that and instead calculate if we are ringing based on the number of remote participants and the state of the remote participant?

JamesBurnside: Chatted with Don offline, when you call a PSTN participant the callstate never enters 'Ringing' state so we're instead watching the call participant state

JamesBurnside (approved):

PR #2415, opened 2022-10-18
Bump `@fluentui/react` dependency to `8.98.3`

JamesBurnside:
# What
Bump `@fluentui/react` dependency to `8.98.3`

# Why
This version has the new react peer dependencies that allow v18 (and all the fluent fixes to support react 18).

**This PR just updates the fluent dependency and does not change our react peer dependency versions yet**

# How Tested
* Ran a quick CallWithChat smoke test locally - verified joining a call with and without camera, sending, editing and deleting messages between two users, sharing screen and putting call on hold. All looked good with no new console errors.
* e2e should be our main point of catching regressions for these updates


packages/react-components/src/components/CameraButton.test.tsx
```diff
@@ -25,9 +25,9 @@ describe('CameraButton strings should be localizable and overridable', () => {
       cameraButton: { offLabel: Math.random().toString(), onLabel: Math.random().toString() }
     });
     const component = mountWithLocalization(<CameraButton showLabel={true} />, testLocale);
-    expect(component.text()).toBe(testLocale.strings.cameraButton.offLabel);
+    expect(component.find('button').text()).toBe(testLocale.strings.cameraButton.offLabel);
```
JamesBurnside, line 28: If you are curious why these unit tests needed updated, Fluent (for A11y reasons) now have the tooltip text rendered in the element but hidden when not hovered:
image

This caused the enzyme `.text()` to pick up the hidden content, so instead we only grab the text from the button element (given this is what we are testing anyway):
image

mgamis-msft (approved):

packages/react-components/src/components/CameraButton.test.tsx
dmceachernmsft: Maybe this is a change to also be called out in the web UI channel so people know going forward this is needed for tests?

dmceachernmsft (approved):

PR #2416, opened 2022-10-18
Add teams call into binding layer

PorterNan:
1. Add teamsHandler to support teams call
2. Add compatibility of provider for TeamsCall and TeamsCallAgent
3. Add 2 new hooks - useTeamsCall, useTeamsCallAgent for non-breaking changes

# What

# Why

# How Tested

# Process & policy checklist


packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1707,6 +1738,16 @@ export type IsSpeakingChangedListener = (event: {
...
 
+// Warning: (ae-internal-missing-underscore) The name "isTeamsCall" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
+export const isTeamsCall: (call: CallCommon) => call is never;
+
+// Warning: (ae-internal-missing-underscore) The name "isTeamsCallAgent" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
+export const isTeamsCallAgent: (callAgent: CallAgentCommon) => callAgent is never;
```
PorterNan, line 1749: Need to remove this from stable

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1291,8 +1312,8 @@ export const createAzureCommunicationChatAdapter: ({ endpoint: endpointUrl, user
 // @public
 export const createAzureCommunicationChatAdapterFromClient: (chatClient: StatefulChatClient, chatThreadClient: ChatThreadClient) => Promise<ChatAdapter>;
 
-// @public
-export const createDefaultCallingHandlers: (callClient: StatefulCallClient, callAgent: CallAgent | undefined, deviceManager: StatefulDeviceManager | undefined, call: Call | undefined) => CallingHandlers;
+// @public (undocumented)
+export const createDefaultCallingHandlers: <AgentType extends CallAgent = CallAgent>(callClient: StatefulCallClient, callAgent: AgentType | undefined, deviceManager: StatefulDeviceManager | undefined, call: CallTypeOf<AgentType> | undefined) => CallHandlersOf<AgentType>;
```
PorterNan, line 1316: There are 2 ways of implement this function -
1. use Generic one but the default CallAgent will be just ACS CallAgent, the benefit of this solution is minimize the api we expose, and also reduce if-else code branch we need, but reading this api needs more time
2. use 2 functions, one for createDefaultCallingHandlers - which returns handlers only for ACS call, then createDefaulTeamstCallingHandlers - which only return handlers for Teams call, this will add up some complexities to number of functions, but api will look better than 1st one

Both of solution will be non-breaking change.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1690,6 +1711,16 @@ export interface IncomingCallState {
...
 
+// Warning: (ae-internal-missing-underscore) The name "isACSCall" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
+export const isACSCall: (call: CallCommon) => call is Call;
+
+// Warning: (ae-internal-missing-underscore) The name "isACSCallAgent" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
+export const isACSCallAgent: (callAgent: CallAgentCommon) => callAgent is CallAgent;
```
PorterNan, line 1722: To be removed in api.md

packages/calling-component-bindings/review/beta/calling-component-bindings.api.md
```diff
@@ -70,16 +72,25 @@ export interface CallClientProviderProps {
     children: React_2.ReactNode;
 }
 
+// Warning: (ae-internal-missing-underscore) The name "CallHandlersOf" should be prefixed with an underscore because the declaration is marked as @internal
```
JamesBurnside, line 75: fix warning

JamesBurnside: (and for CallTypeOf)

packages/calling-component-bindings/review/beta/calling-component-bindings.api.md
```diff
@@ -123,8 +140,11 @@ export type CameraButtonSelector = (state: CallClientState, props: CallingBaseSe
 // @public
 export const cameraButtonSelector: CameraButtonSelector;
 
-// @public
-export const createDefaultCallingHandlers: (callClient: StatefulCallClient, callAgent: CallAgent | undefined, deviceManager: StatefulDeviceManager | undefined, call: Call | undefined) => CallingHandlers;
+// @public (undocumented)
```
JamesBurnside, line 143: all public functions need documented

packages/calling-component-bindings/src/handlers/createACSCallHandlers.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Object containing all the handlers required for calling components.
+ *
+ * Calling related components from this package are able to pick out relevant handlers from this object.
+ * See {@link useHandlers} and {@link usePropsFor}.
+ *
+ * @public
+ */
+export type CallingHandlers = Omit<CallingHandlersCommon, 'onStartCall'> & {
+  onStartCall: (participants: CommunicationIdentifier[], options?: StartCallOptions) => Call | undefined;
```
JamesBurnside, line 31: why doesn't this support all the options it previously supported:
image

JamesBurnside: Is this a breaking change to reduce the supported parameter set?

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -255,6 +255,11 @@ export type CallAdapterUiState = {
     page: CallCompositePage;
 };
 
+// Warning: (ae-internal-missing-underscore) The name "CallAgentCommon" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
+export type CallAgentCommon = CallAgent;
```
prprabhu-ms, line 261: Why is this `@internal` type in the public API?

Same for `CallCommon`.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -475,10 +488,14 @@ export type CallingBaseSelectorProps = {
 };
 
 // @public
-export type CallingHandlers = {
+export type CallingHandlers = Omit<CallingHandlersCommon, 'onStartCall'> & {
```
prprabhu-ms, line 491: We should try very hard to avoid `Omit` types in public API because they are hard to understand for devs.

A couple questions:
* Should `onStartCall` not be included in `CallingHandlersCommon` at all? If the method is different in the two call types, it should not be included in the common interface at all. Treat interface inheritence similar to class inheritence - keep methods additive.
* If not, let's define `CallingHandlers` explicitly, and then add internal type assertions to make sure that the two interfaces agree.
* We slightly prefer `interface` over `type` in declarations.

packages/communication-react/review/stable/communication-react.api.md
prprabhu-ms: I prefer (2) as it is easier for Contoso to understand and aligned with the factory functions you have planned for adapters. The return type can still be generic so that Contoso can write shared code on top.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1707,6 +1738,16 @@ export type IsSpeakingChangedListener = (event: {
     isSpeaking: boolean;
 }) => void;
 
+// Warning: (ae-internal-missing-underscore) The name "isTeamsCall" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
```
prprabhu-ms, line 1743: Remove or mark `@public` if this API is intended to be added here.

prprabhu-ms:
Do you need to regeneration API for beta? I only saw a stable flavor  `communication-react.api.md`

packages/calling-component-bindings/src/hooks/useHandlers.ts
```diff
@@ -17,12 +17,25 @@ import { CallClientContext, useCall, useCallAgent, useDeviceManager } from '../p
...
 
+  // Handle edge case, validate if call and callAgent are the same type (ACS/Teams)
+  if (callAgent && !isACSCallAgent(callAgent)) {
+    if (call && isACSCall(call)) {
+      throw new Error('A TeamsCall must be provided when callAgent is TeamsCallAgent');
+    }
+  }
+
+  if (callAgent && isACSCallAgent(callAgent)) {
+    if (call && !isACSCall(call)) {
```
alkwa-msft, line 35: should we assume if it is not an acs call then its a teams call? I just have a light concern if we ever want to bring other types of calls (highly unlikely) but is it possible to do some other check instead of saying its just not an acs call?

packages/calling-component-bindings/src/handlers/createTeamsCallHandlers.ts
```diff
@@ -0,0 +1,121 @@
...
+ * @beta
+ */
+export const createDefaultTeamsCallingHandlers = memoizeOne(
+  (
+    callClient: StatefulCallClient,
+    callAgent: undefined | /* @conditional-compile-remove(teams-identity-support) */ TeamsCallAgent,
+    deviceManager: StatefulDeviceManager | undefined,
+    call: undefined | /* @conditional-compile-remove(teams-identity-support) */ TeamsCall
+  ): never | TeamsCallingHandlers => {
+    const calingsHandlers = createDefaultCallingHandlersCommon(callClient, deviceManager, call);
```
alkwa-msft, line 50: cal**l**ingHandlers

packages/calling-component-bindings/src/index.ts
```diff
@@ -4,14 +4,23 @@
...
 export type { ErrorBarSelector } from './errorBarSelector';
 /* @conditional-compile-remove(PSTN-calls) */
 export type { HoldButtonSelector } from './callControlSelectors';
 
-export type { CallingHandlers } from './handlers/createHandlers';
+export type { CallHandlersOf, CallTypeOf } from './handlers/createHandlers';
+export type { CallingHandlers } from './handlers/createACSCallHandlers';
+/* @conditional-compile-remove(teams-identity-support) */
+export type { TeamsCallingHandlers } from './handlers/createTeamsCallHandlers';
+export type { CallingHandlersCommon } from './handlers/createHandlersCommon';
```
prprabhu-ms, line 20: Naming nit: To go with `CallingHandlers` and `TeamsCallingHandlers`, I'd name this `CommonCallingHandlers` (as `Teams` is a prefix, so should be `Common`).

prprabhu-ms: Also, should `CallingHandlersCommon` also be a beta-only export?

packages/calling-component-bindings/src/index.ts
```diff
@@ -4,14 +4,23 @@
...
+export { createDefaultTeamsCallingHandlers } from './handlers/createTeamsCallHandlers';
 export type { ParticipantListSelector } from './participantListSelector';
 export type { ParticipantsButtonSelector } from './participantsButtonSelector';
 export type { VideoGallerySelector } from './videoGallerySelector';
 export type { ErrorBarSelector } from './errorBarSelector';
 /* @conditional-compile-remove(PSTN-calls) */
 export type { HoldButtonSelector } from './callControlSelectors';
 
-export type { CallingHandlers } from './handlers/createHandlers';
+export type { CallHandlersOf, CallTypeOf } from './handlers/createHandlers';
```
prprabhu-ms, line 16: These should be beta-only exports.

packages/calling-component-bindings/src/handlers/createACSCallHandlers.ts
```diff
@@ -0,0 +1,73 @@
...
+        options?: AddPhoneNumberOptions
+      ): Promise<void> => {
+        const participant = _toCommunicationIdentifier(userId);
+        if (isPhoneNumberIdentifier(participant)) {
+          call?.addParticipant(participant, options);
+        } else if (isCommunicationUserIdentifier(participant) || isMicrosoftTeamsUserIdentifier(participant)) {
+          call?.addParticipant(participant);
+        }
+      },
+      onRemoveParticipant: async (
```
alkwa-msft, line 65: when we remove a participant will this try to convert the id to a communicationIdentifier? What if they aren't a communications user but a phone or a teams user?

packages/calling-component-bindings/src/hooks/usePropsFor.ts
```diff
@@ -61,7 +61,7 @@ import { ErrorBarSelector, errorBarSelector } from '../errorBarSelector';
 export const usePropsFor = <Component extends (props: any) => JSX.Element>(
```
prprabhu-ms, line 61: Are we saying that we don't support `usePropsFor` for components that need any Teams specific handlers, for now?

What is the plan for when we do have components that need Teams specific handlers (my guess would be to have a separate `usePropsForTeamsComponent` or some such, but want to confirm)?

packages/calling-component-bindings/src/providers/CallProvider.tsx
```diff
@@ -57,5 +60,26 @@ export const CallProvider = (props: CallProviderProps): JSX.Element => <CallProv
  * @public
  */
 export const useCall = (): Call | undefined => {
-  return useContext(CallContext)?.call;
+  const call = useContext(CallContext)?.call;
+  if (call && !isACSCall(call)) {
+    throw new Error('TeamsCall object was provided, try useTeamsCall() instead');
```
prprabhu-ms, line 65: Future proofing: We're not sure here that it's a `TeamsCall`, only that it is not an ACS call. Let's reword this to `'Incorrect call type: must provide a `Call` object.`.

packages/calling-component-bindings/src/providers/CallProvider.tsx
```diff
@@ -57,5 +60,26 @@ export const CallProvider = (props: CallProviderProps): JSX.Element => <CallProv
...
+ *
+ * Useful when implementing a custom component that utilizes the providers
+ * exported from this library.
+ *
+ * @public
+ */
+export const useTeamsCall = (): undefined | /* @conditional-compile-remove(teams-identity-support) */ TeamsCall => {
+  const call = useContext(CallContext)?.call;
+  if (call && isACSCall(call)) {
+    throw new Error('Regular ACS Call object was provided, try useCall() instead');
```
prprabhu-ms, line 82: Same comment as above for future proofing.

packages/calling-component-bindings/src/providers/CallProvider.tsx
```diff
@@ -57,5 +60,26 @@ export const CallProvider = (props: CallProviderProps): JSX.Element => <CallProv
...
+  return call;
+};
+
+/* @conditional-compile-remove(teams-identity-support) */
+/**
+ * Hook to obtain {@link @azure/communication-calling#TeamsCall} from the provider.
+ *
+ * Useful when implementing a custom component that utilizes the providers
+ * exported from this library.
+ *
```
prprabhu-ms, line 76: Let's add to this documentation that you must have previously used the `CallProvider` with a `TeamsCall` to use this hook.

Same for the comment for `useCall`.

packages/calling-component-bindings/src/providers/CallAgentProvider.tsx
```diff
@@ -53,4 +56,28 @@ export const CallAgentProvider = (props: CallAgentProviderProps): JSX.Element =>
  *
  * @public
  */
-export const useCallAgent = (): CallAgent | undefined => useContext(CallAgentContext)?.callAgent;
+export const useCallAgent = (): CallAgent | undefined => {
```
prprabhu-ms, line 59: Same comments about the thrown error and documentation as for `useCall` and `useTeamsCall`.

packages/calling-component-bindings/src/utils/callUtils.ts
```diff
@@ -75,3 +83,31 @@ const memoizedUpdateDisplayName = memoizeFnAll((participantId: string, participa
     return participant;
   }
 });
+
+/**
+ * @private
+ */
+export const isTeamsCallParticipants = (
```
alkwa-msft, line 90: In this function we are determining if all of the participants are all teams users? Does a MicrosoftTeamsUserIdentifier match that condition?

packages/calling-component-bindings/src/providers/CallProvider.tsx
```diff
@@ -57,5 +60,26 @@ export const CallProvider = (props: CallProviderProps): JSX.Element => <CallProv
...
+ * @public
+ */
+export const useTeamsCall = (): undefined | /* @conditional-compile-remove(teams-identity-support) */ TeamsCall => {
+  const call = useContext(CallContext)?.call;
+  if (call && isACSCall(call)) {
```
prprabhu-ms, line 81: Future proofing: Explicity check `isTeamsCall`.

packages/calling-component-bindings/src/providers/CallProvider.tsx
```diff
@@ -57,5 +60,26 @@ export const CallProvider = (props: CallProviderProps): JSX.Element => <CallProv
...
+/* @conditional-compile-remove(teams-identity-support) */
+/**
+ * Hook to obtain {@link @azure/communication-calling#TeamsCall} from the provider.
+ *
+ * Useful when implementing a custom component that utilizes the providers
+ * exported from this library.
+ *
+ * @public
+ */
+export const useTeamsCall = (): undefined | /* @conditional-compile-remove(teams-identity-support) */ TeamsCall => {
```
prprabhu-ms, line 79: This should also be marked `@beta` not `@public`.

Also consider simply conditionally compiling out the entire method (simpler than conditionally compiling the argument and then also the export in `index.ts`. Your call ;)

packages/calling-component-bindings/src/utils/callUtils.ts
```diff
@@ -75,3 +83,31 @@ const memoizedUpdateDisplayName = memoizeFnAll((participantId: string, participa
...
+    }
+  }
+  return true;
+};
+
+/**
+ * @private
+ */
+export const isNonTeamsCallParticipants = (
+  participants: CommunicationIdentifier[]
```
alkwa-msft, line 105: Would be helpful to have a comment and some unit tests

nit. Prefer to avoid having negative names as functions.. just because when someone starts doing !isNonTeamsCallParticipants.. It is really easy to mess up the logic.

packages/communication-react/review/stable/communication-react.api.md
alkwa-msft: I think it is ok to do a little bit of duplication if it helps with readability. For the onStartCall function.

packages/calling-component-bindings/src/hooks/useHandlers.ts
prprabhu-ms: Agreed. I left similar comments elsewhere. A more future proof test would be:

```TypeScript

if (callAgent && isTeamsCallAgent(callAgent)) {
  if (call && !isTeamsCall(call) {
    throw new Error('Incompatible call type: A TeamsCall must be used with a TeamsCallAgent');
  }
}

if (callAgent && isAcsCallAgent(callAgent)) {
  if (call && !isAcsCall(call) {
    throw new Error('Incompatible call type: A Call must be used with a CallAgent');
  }
}
```

packages/calling-component-bindings/src/utils/callUtils.ts
```diff
@@ -75,3 +83,31 @@ const memoizedUpdateDisplayName = memoizeFnAll((participantId: string, participa
...
   }
 });
+
+/**
+ * @private
+ */
+export const isTeamsCallParticipants = (
+  participants: CommunicationIdentifier[]
+): participants is (PhoneNumberIdentifier | MicrosoftTeamsUserIdentifier | UnknownIdentifier)[] => {
+  for (const p of participants) {
```
prprabhu-ms, line 93: The one-liner lover in me is itching to say: `return participants.every(p => !isCommunicationUserIdentifier(p))`

packages/calling-component-bindings/src/utils/callUtils.ts
```diff
@@ -75,3 +83,31 @@ const memoizedUpdateDisplayName = memoizeFnAll((participantId: string, participa
     return participant;
   }
 });
+
+/**
+ * @private
+ */
+export const isTeamsCallParticipants = (
+  participants: CommunicationIdentifier[]
+): participants is (PhoneNumberIdentifier | MicrosoftTeamsUserIdentifier | UnknownIdentifier)[] => {
```
prprabhu-ms, line 92: Nice type restriction.

packages/calling-component-bindings/src/utils/callUtils.ts
```diff
@@ -75,3 +83,31 @@ const memoizedUpdateDisplayName = memoizeFnAll((participantId: string, participa
...
+      return false;
+    }
+  }
+  return true;
+};
+
+/**
+ * @private
+ */
+export const isNonTeamsCallParticipants = (
```
prprabhu-ms, line 104: The naming is a bit confusing, especially when compared with `isTeamsCallParticipants`.

Either rename this to `isAcsCallParticipants` or rename both: `containsNoAcsParticipant()` and `containsNoTeamsParticipant`. I like the latter because it is closer to what the function does and is grammatically correct ("is" should't be used with plural).

packages/calling-component-bindings/src/utils/callUtils.ts
```diff
@@ -75,3 +83,31 @@ const memoizedUpdateDisplayName = memoizeFnAll((participantId: string, participa
...
+  participants: CommunicationIdentifier[]
+): participants is (PhoneNumberIdentifier | CommunicationUserIdentifier | UnknownIdentifier)[] => {
+  for (const p of participants) {
+    if (isCommunicationUserIdentifier(p)) {
```
prprabhu-ms, line 108: Bug? This returns false if you _do_ find an ACS user.

prprabhu-ms (commented):
Partial review.

alkwa-msft (commented):
some small nits but it looks alright. I think Prathmesh and James asked the really big questions. Lets make sure they have approved

packages/calling-component-bindings/src/handlers/createTeamsCallHandlers.ts
```diff
@@ -0,0 +1,121 @@
...
+      },
+      /* @conditional-compile-remove(teams-identity-support) */
+      /* @conditional-compile-remove(PSTN-calls) */
+      onAddParticipant: async (
+        userId: string | CommunicationIdentifier,
+        options?: AddPhoneNumberOptions
+      ): Promise<void> => {
+        const participant = _toCommunicationIdentifier(userId);
+        /* @conditional-compile-remove(teams-identity-support) */
+        const threadId = options?.threadId;
```
prprabhu-ms, line 73: I don't undderstand this. How can `options?: AddPhoneNumberOptions` have a `threadId`?

packages/calling-component-bindings/src/handlers/createTeamsCallHandlers.ts
```diff
@@ -0,0 +1,121 @@
...
+        /* @conditional-compile-remove(teams-identity-support) */
+        if (isPhoneNumberIdentifier(participant)) {
```
prprabhu-ms, line 75: Can you rework this to remove nesting:

```
if (isCommunicationUserIdentifier(participant) {
   throw new Error(...
}
if (isPhoneNumberIdentifier(participant) {
  return call?.addParticipant(...
}

return call?.addParticipant(...)
```

packages/calling-component-bindings/src/handlers/createTeamsCallHandlers.ts
```diff
@@ -0,0 +1,121 @@
...
+          call?.addParticipant(participant, threadId ? { threadId } : undefined);
+        } else {
+          if (isCommunicationUserIdentifier(participant)) {
+            throw new Error('CommunicationIdentifier in Teams call is not supported!');
```
prprabhu-ms, line 79: Why do we have this check in the UI layer? Should the headless SDK throw in this case?

packages/calling-component-bindings/src/handlers/createTeamsCallHandlers.ts
```diff
@@ -0,0 +1,121 @@
...
+          }
+          /* @conditional-compile-remove(teams-identity-support) */
+          call?.addParticipant(participant, threadId ? { threadId } : undefined);
+        }
+      },
+      onRemoveParticipant: async (
+        userId: string | /* @conditional-compile-remove(PSTN-calls) */ CommunicationIdentifier
+      ): Promise<void> => {
+        const participant = _toCommunicationIdentifier(userId);
+        if (isCommunicationUserIdentifier(participant)) {
```
prprabhu-ms, line 89: Same question as above. Why does this check exist in the UI layer as opposed to the headless layer?

packages/calling-component-bindings/src/handlers/createHandlers.ts
```diff
@@ -1,426 +1,58 @@
...
- * Create the default implementation of {@link CallingHandlers}.
- *
- * Useful when implementing a custom component that utilizes the providers
- * exported from this library.
- *
  * @public
  */
 export const createDefaultCallingHandlers = memoizeOne(
-  (
+  <AgentType extends CallAgent | /* @conditional-compile-remove(teams-identity-support) */ TeamsCallAgent = CallAgent>(
```
prprabhu-ms, line 42: Are the types `CallTypeOf` and `CallHandlersOf` _only_ introduced to support a generic `createDefaultCallingHandlers`?

In that case, I'd say we consider an alternative where these (difficult to understand) types are not introduced at the cost of having to know which of two functions `createDefaultCallingHandlers` vs `createDefaultTeamsCallingHandlers` must be called. This function should anyway not be needed frequently by devs.

iiuc, `useHandlers` will continue to work seamlessly for both types of calls and does not need new generic types to be introduced.

packages/calling-component-bindings/src/handlers/createACSCallHandlers.ts
```diff
@@ -0,0 +1,73 @@
...
+        }
+        return callAgent ? callAgent.startCall(participants, options) : undefined;
+      },
+      /* @conditional-compile-remove(PSTN-calls) */
+      onAddParticipant: async (
+        userId: string | CommunicationIdentifier,
+        options?: AddPhoneNumberOptions
+      ): Promise<void> => {
+        const participant = _toCommunicationIdentifier(userId);
+        if (isPhoneNumberIdentifier(participant)) {
```
prprabhu-ms, line 59: Again, I don't understand why we need to stop forwarding `options` to the headless API in some cases, given that the type of `options` is `AddPhoneNumberOptions`.

packages/calling-component-bindings/src/handlers/createACSCallHandlers.ts
prprabhu-ms, line 65: Can this type casting move to the common implementation?

packages/communication-react/review/stable/communication-react.api.md
PorterNan: I see the concerning here - onStartCall is the shared logic between call and teamsCall hanlders, which will be required by adapters/components. In common handlers, not return type is required, but we need to return call object for each 2 specific type otherwise it would be breaking change. I have a better way to refactor this part, lemme take a shot then update here soon

packages/calling-component-bindings/src/index.ts
PorterNan: CallingHandlersCommon would be used and required by some stable feature (imagine components will be asking for CallingHandlersCommon  instead of CallingHandlers)

packages/calling-component-bindings/src/hooks/usePropsFor.ts
PorterNan: Yep, if there is anything teams specific, that would be usePropsForTeamsComponent(or some other names), but looks like calling sdk doesn't have any plan shipping something like this for now

packages/calling-component-bindings/src/hooks/useHandlers.ts
PorterNan: Yep, that's a great point, in previous implementation, I didn't have a perfect way to implement isTeamsCall. But this isn't a problem anymore, lemme apply this comment in next a few iterations in the PR

packages/calling-component-bindings/src/handlers/createACSCallHandlers.ts
PorterNan: which type casting? Is it this one?
```
const participant = _toCommunicationIdentifier(userId);
```

packages/calling-component-bindings/src/handlers/createACSCallHandlers.ts
PorterNan: I believe this is a copy&paste code, lemme see if I can forward this param

packages/communication-react/review/stable/communication-react.api.md
PorterNan: Hmm, maybe I should make it public... this is not smart enough to make cast it back to CallAgent...

packages/calling-component-bindings/src/handlers/createHandlers.ts
PorterNan: You are 100% right! I've refactored the logic to my solution 2 - which doesn't require expose such a generic type

packages/calling-component-bindings/src/handlers/createTeamsCallHandlers.ts
PorterNan: Don't know about this either...looks like it is just the backward compatibility of calling sdk

packages/calling-component-bindings/src/handlers/createTeamsCallHandlers.ts
PorterNan: Because we want to align handler signature to simplify component logic for now - only one or two identifier types are not worthy to create an entirely new set of handlers

packages/calling-component-bindings/src/handlers/createACSCallHandlers.ts
PorterNan: image
Just checked api we are exposing in PSTN calling feature, looks like we don't currently expose options for all the other types of identifier, so I am not changing any logic here for now

packages/calling-component-bindings/src/utils/callUtils.ts
PorterNan: I tried the latter name first, but found the logic is a bit hard to follow when reading it (because there are 2 big 'No's there), so I fallback to name of 'isAcsCallParticipants'

prprabhu-ms:
Looks like this is ready for another round of review tomorrow morning?

packages/acs-ui-common/package.json
```diff
@@ -33,7 +33,8 @@
   },
   "peerDependencies": {
     "react": ">=16.8.0 <18.0.0",
-    "@types/react": ">=16.8.0 <18.0.0"
+    "@types/react": ">=16.8.0 <18.0.0",
+    "@azure/communication-calling": "1.8.0-beta.1 || >=1.4.4"
```
JamesBurnside, line 37: my understanding is acs-ui-common should only be for common utils/types used across both calling and chat. We have a dependency chain within calling where stateful calling packlet is the base dependency that can export these. Otherwise Chat now has an implicit dependency on communication-calling

prprabhu-ms: These should be exported as `@internal` symbols from `calling-component-bindings` instead of `acs-ui-common`.

prprabhu-ms: (Or `calling-stateful-client`, depends on where it is needed.

packages/acs-ui-common/review/stable/acs-ui-common.api.md
```diff
@@ -19,9 +22,15 @@ export type AreParamEqual<A extends (props: any) => JSX.Element | undefined, B e
...
 export type AreTypeEqual<A, B> = A extends B ? (B extends A ? true : false) : false;
 
+// @public
+export type CallAgentCommon = CallAgent;
+
 // @public
 export type CallbackType<KeyT, ArgsT extends unknown[], FnRetT> = (memoizedFn: FunctionWithKey<KeyT, ArgsT, FnRetT>) => FnRetT[];
 
+// @public
+export type CallCommon = Call;
```
prprabhu-ms, line 32: Why is this defined here, *and* In BetaToStableTypes.ts?

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
```diff
@@ -57,16 +53,32 @@ export type CallingHandlersCommon = {
...
+ * Calling related components from this package are able to pick out relevant handlers from this object.
+ * See {@link useHandlers} and {@link usePropsFor}.
+ *
+ * @public
+ */
+// CommonCallingHandlers is an interface compatible with TeamsCallingHandlers and CallingHandlers
+// which is also a common interface required by components and adapters
+// components and adapters only requires onStartCall is a function returning void
+// but in TeamsCallingHandlers and CallingHandlers, it returns different types for non-breaking change reason
+export type CommonCallingHandlers = SharedCallingHandlers & {
```
prprabhu-ms, line 75: This is unfortunate, and I'd guess mighty confusing for Contoso.

does this work?

```

interface SharedCallingHandlers {
  onStartCall(...args): void;
}

interface CallingHandlers {
  onStartCall(...args): Call;
}

interface TeamsCallingHandlers {
  onStartCall(...args): TeamsCall;
}
```

i.e., the common interface has a method that returns `void` and the specific interfaces extend that method with different return types.

packages/calling-component-bindings/src/handlers/createTeamsCallHandlers.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export type TeamsCallingHandlers = SharedCallingHandlers & {
+  onStartCall: (
+    participants: CommunicationIdentifier[],
+    options?: StartCallOptions
+  ) => void | /* @conditional-compile-remove(teams-identity-support) */ TeamsCall;
+};
+
+/**
+ * Create the default implementation of {@link TeamsCallingHandlers} for teams call.
```
JamesBurnside, line 36: > for teams call

These are handlers if you are using a teams identity. You could still have a teams interop call using the acs call objects right?

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
JamesBurnside: > mighty confusing for Contoso

I would add and us developing in this repo. `CommonCallingHandlers` and `SharedCallingHandlers` by name are essentially synonyms

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
```diff
@@ -0,0 +1,406 @@
...
+  return !!prevStream && !!newStream && prevStream.source.id === newStream.source.id;
+};
+
+/**
+ * Create the default implementation of {@link CallingHandlers}.
+ *
+ * Useful when implementing a custom component that utilizes the providers
+ * exported from this library.
+ *
+ * @public
```
JamesBurnside, line 95: should this be public? Wouldn't Contoso either use createCallHandlers or createTeamsCallHandlers?

JamesBurnside (approved):

packages/acs-ui-common/package.json
PorterNan: We are now export everything by 
```
exporting * from 'call-stateful-client'
```

Then we might need to change this pattern a bit to only export types/functions we need, what do you think?

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
PorterNan: @prprabhu-ms That was the first shot I tried, but it does not allow me to override with same parameter but different return value, then I chose this one in the end

@prprabhu-ms @JamesBurnside Do you have a better naming recommendation?

packages/calling-component-bindings/src/handlers/createTeamsCallHandlers.ts
PorterNan: Yep, teams interop is still an ACS call, but joining a teams link as a locator (people still pay ACS calling bill)
Teams Call is the one use AAD token directly (which they pay teams/license)

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
PorterNan: Good catch, they don't need this, we don't actually export this in communication-react, lemme change the mark

packages/acs-ui-common/package.json
PorterNan: Looking at composite package, now following the pattern of index-public.ts to resolve this problem

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
prprabhu-ms: What is the error you got? Typescript seems to be happy with:

```Typescript

interface A {
  foo(x: number): void;
}

interface B extends A {
  foo(x: number): number;
}

interface C extends A {
  foo(x: number): string;
}

function main() {
  const b: B = {
    foo: (x: number) => { return x; }
  }
  iWantA(b);
  const c: C = {
    foo: (x: number) => { return `${x}`; }
  }
  iWantA(c);
}

function iWantA(a: A) {
  a.foo(2);
}
```

PorterNan: Interesting, I tried Interface, and it does override the function, 

image

but it does not work for Type:
image


Then there is one more obvious reason when we want to use Interface over Type - for better function override. Lemme change all the definition to Interface for this case and see if it works well

PorterNan: This is an awesome discussion! Learning something new here

prprabhu-ms (approved):
Sweet! The stable and beta APIs look great now. Thanks for iterating on this.

PR #2417, opened 2022-10-18
Add rooms role into calling stateful client

mgamis-msft:
# What

Add rooms role into calling stateful client

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3021449

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -249,6 +253,16 @@ export class CallContext {
     });
   }
 
+  /* @conditional-compile-remove(rooms) */
+  public setRoleChanged(callId: string, role: ParticipantRole): void {
```
prprabhu-ms, line 257: Should this be named simply `setRole`?

`setRoleChanged` suggests that you're setting a callback of some sort.

prprabhu-ms (approved):

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -320,6 +334,19 @@ export class CallContext {
     });
   }
 
+  /* @conditional-compile-remove(rooms) */
+  public setParticipantRole(callId: string, participantKey: string, role: ParticipantRole): void {
```
PorterNan, line 338: One minor thing to confirm, will participant events always happen before the RoleChange event? Just a bit worry about role event happening to soon so no participant could be found

PorterNan (approved):

packages/calling-stateful-client/src/CallContext.ts
mgamis-msft: That's a really good thought. I will confirm that.

PR #2419, opened 2022-10-19
Added role to selectors

mgamis-msft:
# What

# Why

# How Tested

# Process & policy checklist


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2691,8 +2692,8 @@ export interface RemoteVideoStreamState {
     view?: VideoStreamRendererViewState;
 }
 
-// @beta (undocumented)
-export type Role = 'Presenter' | 'Attendee' | 'Consumer';
+// @beta
+export type Role = 'Unknown' | 'Presenter' | 'Attendee' | 'Consumer' | 'Organizer';
```
JamesBurnside, line 2696: Is unknown from the Calling SDK? I would suggest role being `'Presenter' | 'Attendee' | 'Consumer' | 'Organizer'` only and then it should be passed as undefined it is unknown or not used.
Or is there an SDK role type we can use?

JamesBurnside: From the meeting this morning, as things move to the selector layer and adpater ctor this interface may be removed as well so not blocking on this comment

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -30,14 +30,11 @@ const convertRemoteParticipantsToParticipantListParticipants = (
...
-            toFlatCommunicationIdentifier(participant.identifier),
-            participant.displayName,
-            participant.state,
-            participant.isMuted,
+          return memoizeFn({
+            userId: toFlatCommunicationIdentifier(participant.identifier),
             isScreenSharing,
-            participant.isSpeaking
-          );
+            ...participant
```
JamesBurnside, line 36: this adds properties (like videoStreams) which I don't think we want here?

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -30,14 +30,11 @@ const convertRemoteParticipantsToParticipantListParticipants = (
           const isScreenSharing = Object.values(participant.videoStreams).some(
             (videoStream) => videoStream.mediaStreamType === 'ScreenSharing' && videoStream.isAvailable
           );
-          return memoizeFn(
-            toFlatCommunicationIdentifier(participant.identifier),
-            participant.displayName,
-            participant.state,
-            participant.isMuted,
+          return memoizeFn({
+            userId: toFlatCommunicationIdentifier(participant.identifier),
```
JamesBurnside, line 34: was this broken before?

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -30,14 +30,11 @@ const convertRemoteParticipantsToParticipantListParticipants = (
           const isScreenSharing = Object.values(participant.videoStreams).some(
             (videoStream) => videoStream.mediaStreamType === 'ScreenSharing' && videoStream.isAvailable
           );
-          return memoizeFn(
-            toFlatCommunicationIdentifier(participant.identifier),
-            participant.displayName,
-            participant.state,
-            participant.isMuted,
+          return memoizeFn({
```
JamesBurnside, line 33: careful changing this from a set of parameters to an object. It's passed into a memoize function, but it is creating a new object everytime the memoizeFn is called - will this still be memoized?

packages/communication-react/review/beta/communication-react.api.md
mgamis-msft: Yes, 'Unknown' is from from the Calling SDK also.

packages/calling-component-bindings/src/utils/participantListSelectorUtils.ts
```diff
@@ -9,41 +12,31 @@ import { CallParticipantListParticipant } from '@internal/react-components';
...
-  ): CallParticipantListParticipant => {
-    return convertRemoteParticipantToParticipantListParticipant(
-      userId,
-      displayName,
-      state,
-      isMuted,
-      isScreenSharing,
-      isSpeaking
-    );
+  (data: {
```
anjulgarg, line 15: Similar to what James said above. The first function param is used as the key and this won't be memoized since you are passing an object.

packages/calling-component-bindings/src/participantListSelector.ts
mgamis-msft: OK I reverted back to a set of parameters. I have to create a separate beta version of the function since conditional compilation does not work if the new function parameter is a beta type.

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: Just a note that `react-components` should only list the values they care about, not necessarily reflect the Calling SDK exactly. What is `unknown` to a react-component? When would Contoso, when using a pure component, pass `unknown` instead of undefined?

packages/calling-component-bindings/src/utils/participantListSelectorUtils.ts
```diff
@@ -50,3 +52,56 @@ const convertRemoteParticipantToParticipantListParticipant = (
...
+  return {
+    userId,
+    displayName,
+    state,
+    isMuted,
+    isScreenSharing,
+    isSpeaking,
+    // ACS users can not remove Teams users.
+    // Removing unknown types of users is undefined.
+    isRemovable:
```
JamesBurnside, line 102: avoid duplicate logic, if conditional compile isn't working, instead have convertRemoteParticipantToParticipantListParticipantBeta call convertRemoteParticipantToParticipantListParticipant then apply the beta adaptions on top:
```
convertRemoteParticipantToParticipantListParticipantBeta = (args) => {
  return {
    convertRemoteParticipantToParticipantListParticipant(args),
    // beta things here
  }
}
```

packages/calling-component-bindings/src/participantListSelector.ts
mgamis-msft: Removed the spread. Thanks.

packages/calling-component-bindings/src/utils/participantListSelectorUtils.ts
JamesBurnside: This will also highlight what differs between these two functions without having to compare every line

JamesBurnside: However, we should be updating our conditional compile to be able to work for this util function

JamesBurnside (approved):

packages/calling-component-bindings/src/utils/participantListSelectorUtils.ts
mgamis-msft: Good point. Thanks.

anjulgarg (approved):

PR #2426, opened 2022-10-21
Add TeamsCallAdapter interface

PorterNan:
Add TeamsCallAdapter interface
Add createAzureCommunicationTeamsCallAdapter() function 
Extract common interface between teamsCallAdapter and CallAdapter

# What

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatAdapter.ts
```diff
@@ -369,12 +370,17 @@ export interface CallWithChatAdapterSubscriptions {
...
+export interface CallWithChatAdapterCommon
   extends CallWithChatAdapterManagement,
     AdapterState<CallWithChatAdapterState>,
     Disposable,
     CallWithChatAdapterSubscriptions {}
 
+/**
+ * @public
+ */
+export type CallWithChatAdapter = Omit<CallWithChatAdapterCommon, keyof ACSCallManagement> & ACSCallManagement;
```
JamesBurnside, line 382: note: ARB is quite against the use of Omit and non-flat structures for public types, is there another way this could be defined?

JamesBurnside: (we might need a seperate work item to define a flat structure for the adapter interface as it already extends a few things)

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatAdapter.ts
```diff
@@ -65,7 +66,7 @@ export interface CallWithChatAdapterManagement {
    *
    * @public
    */
-  joinCall(microphoneOn?: boolean): Call | undefined;
+  joinCall(microphoneOn?: boolean): void;
```
JamesBurnside, line 69: this looks like a breaking change

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -77,6 +79,12 @@ import { VideoDeviceInfo } from '@azure/communication-calling';
 import { VideoStreamRenderer } from '@azure/communication-calling';
 import { VideoStreamRendererView } from '@azure/communication-calling';
 
+// @public (undocumented)
+export type ACSCallManagement = {
```
JamesBurnside, line 83: Is there a reason to define seperate types for ACSCallManagement and TeamsCallManagement or can we use a generic?
```
interface CallManagement<CallType> {
    joinCall(microphoneOn?: boolean): <CallType> | undefined;
    startCall(participants: string[], options?: StartCallOptions): <CallType> | undefined;
};
```

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -230,6 +237,10 @@ export type CallAdapterClientState = {
     alternateCallerId?: string;
 };
 
+// @public
+export interface CallAdapterCommon extends AdapterState<CallAdapterState>, Disposable, CallAdapterCallManagement, CallAdapterDeviceManagement, CallAdapterSubscribers {
```
JamesBurnside, line 241: Can we avoid exporting the common interface and just publicly export `CallAdapter` and `TeamsCallAdapter` and use `CallAdapter | TeamsCallAdapter` where this is used? (in the public interface)

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -527,7 +538,12 @@ export type CallErrors = {
...
+
+// Warning: (ae-internal-missing-underscore) The name "CallHandlersOf" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
+export type CallHandlersOf<AgentType extends TeamsCallAgent | CallAgent> = AgentType extends CallAgent ? CallingHandlers : never | /* @conditional-compile-remove(teams-call) */ TeamsCallingHandlers;
```
JamesBurnside, line 546: Public api file shouldn't have internals

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -623,10 +644,19 @@ export interface CallState {
     startTime: Date;
     state: CallState_2;
     transcription: TranscriptionCallFeature;
+    type: 'Teams' | 'ACS';
 }
 
+// Warning: (ae-internal-missing-underscore) The name "CallTypeOf" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
```
JamesBurnside, line 652: Public api file shouldn't have internals

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -239,9 +246,8 @@ export interface BrowserPermissionDeniedStyles extends BaseCustomStyles {
     troubleshootingLink?: ILinkStyles;
 }
 
-// @public
-export interface CallAdapter extends AdapterState<CallAdapterState>, Disposable, CallAdapterCallManagement, CallAdapterDeviceManagement, CallAdapterSubscribers {
-}
+// @public (undocumented)
+export type CallAdapter = Omit<CallAdapterCommon, keyof ACSCallManagement> & ACSCallManagement;
```
PorterNan, line 250: This Omit can be replaced by 'interface' method as we discussed @prprabhu-ms, will be working on it soon

packages/communication-react/review/beta/communication-react.api.md
prprabhu-ms: Can't. `CallComposite` now only needs the `CallAdapterCommon` interface, not `CallAdapter`. This is because the composite only makes use of the common functionality.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -72,6 +72,12 @@ import { VideoDeviceInfo } from '@azure/communication-calling';
 import { VideoStreamRenderer } from '@azure/communication-calling';
 import { VideoStreamRendererView } from '@azure/communication-calling';
 
+// @public (undocumented)
+export type ACSCallManagement = {
```
prprabhu-ms, line 76: Don't use the acronym `ACS` in the API.

prprabhu-ms (commented):
Delete packages/communication-react/temp copy/communication-react.api.json

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -801,41 +881,40 @@ export const createAzureCommunicationCallAdapter = async ({
...
+    /* @conditional-compile-remove(PSTN-calls) */ alternateCallerId
+  });
+  const callAgent = await callClient.createTeamsCallAgent(credential, {
+    undefined
+  });
+  const adapter = createAzureCommunicationTeamsCallAdapterFromClient(callClient, callAgent, locator);
+  return adapter;
+};
+
+type AdapterType<CallType extends 'ACS' | 'Teams'> = CallType extends 'ACS'
```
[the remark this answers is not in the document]
prprabhu-ms: Similar to above, can we define this type unconditionally to reduce the mental load of reading the code?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -925,6 +1028,93 @@ export const useAzureCommunicationCallAdapter = (
...
+  /**
+   * Optional callback called before the adapter is disposed.
+   *
+   * This is useful for clean up tasks, e.g., leaving any ongoing calls.
+   */
+  beforeDispose?: (adapter: CallAdapter) => Promise<void>
+): CallAdapter | undefined => {
+  return useAzureCommunicationCallAdapterGeneric(
+    args,
+    afterCreate as (adapter: CallAdapter) => Promise<CallAdapter>,
```
prprabhu-ms, line 1068: This downcast is not safe. What if Contoso provides a callback that actually returns only a `Promise<CallAdapterCommon>`?

That said, I think we can keep the signature `afterCreate?: (adapter: CallAdapter) => Promise<CallAdapter>` because before this change, Contoso _must have_ used a callback with that signature (this PR only splits the definition of `CallAdapter` into `CallAdapterCommon` and `CallAdapter`) so that this downcast won't be necessary.

Same comment below.

prprabhu-ms (commented):
Overall looks less controversial than the lower layers.

Thanks for working through this.

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -60,7 +62,7 @@ export const END_CALL_PAGES: CallCompositePage[] = [
 ];
 
 /**
- * {@link CallAdapter} state for pure UI purposes.
+ * {@link CallAdapterCommon} state for pure UI purposes.
```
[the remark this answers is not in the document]
PorterNan: CallAdapter is used for ACS already, if we move it to AzureCommunicationCallAdapter, it would be a breaking type change :(
What do you think of CommonCallAdapter? Or AbstractCallAdapter?

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -259,14 +261,16 @@ export interface CallAdapterCallManagement {
...
    */
-  startCall(participants: string[], options?: StartCallOptions): Call | undefined;
+  startCall(participants: string[], options?: StartCallOptions): void;
   /* @conditional-compile-remove(PSTN-calls) */
   /**
    * Start the call.
-   * @param participants - An array of {@link @azure/communication-common#CommunicationIdentifier} to be called
-   * @beta
+   *
+   * @param participants - An array of participant to join
```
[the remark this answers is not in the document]
PorterNan: Nice catch, it should be just a merge conflict

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -238,13 +267,30 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
-    this.handlers = createDefaultCallingHandlers(callClient, callAgent, deviceManager, undefined);
-
     this.onClientStateChange = onStateChange;
 
     this.subscribeDeviceManagerEvents();
 
     this.callClient.onStateChange(onStateChange);
+
+    /* @conditional-compile-remove(teams-identity-support) */
+    if (!_isACSCallAgent(callAgent)) {
```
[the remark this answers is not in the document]
PorterNan: For 1&2, yes, I need to apply some patterns from last 2 PR reviews, just need some time after I finish interface change

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -238,13 +267,30 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
+      this.handlers = createDefaultTeamsCallingHandlers(
+        callClient,
+        callAgent,
+        deviceManager,
+        undefined
+      ) as CallHandlersOf<AgentType>;
```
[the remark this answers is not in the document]
PorterNan: TypeScript is not smart enough to infer the type during the code closure :-(, the type would be unknown if I don't do the cast

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatAdapter.ts
PorterNan: This is for CommonCallAdapter - which is the interface(abstraction) needed by composite, it requires less so it is backward compatible 

For CallAdapter, it is remaining the same, joinCall still return Call | undefined there

prprabhu-ms:
Please IM me when this is ready for another round of reviews. Would like to help get this over the finish line this week.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: That's a point I wanted to discuss - lemme resolve all the other comments from Internal review and this PR, then we can chat about this a bit later
It is just because I found myself had a really hard time to implement a common function in contoso side

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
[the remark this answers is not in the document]
PorterNan: We chose CommonCallAdapter to align with handler layer in the end

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -801,41 +881,40 @@ export const createAzureCommunicationCallAdapter = async ({
...
+};
+
+type AdapterType<CallType extends 'ACS' | 'Teams'> = CallType extends 'ACS'
+  ? CallAdapter
+  : /* @conditional-compile-remove(teams-identity-support) */ TeamsCallAdapter;
+
+/**
+ * @private
+ */
+const useAzureCommunicationCallAdapterGeneric = <CallType extends 'ACS' | 'Teams' = 'ACS'>(
```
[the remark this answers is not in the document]
PorterNan: This is a very smart one! I tried this and it works seamlessly. It also reduces overall code if-else branches, perfect!

PorterNan:
Created a follow-up item to refactor implementation of adapter as we discussed @prprabhu-ms https://skype.visualstudio.com/SPOOL/_workitems/edit/2894791
Will pick it up a bit later once we finish main chunk of CTE

common/config/rush/pnpm-lock.yaml
```diff
@@ -28,12 +28,19 @@ packages:
...
   /@azure/abort-controller/1.1.0:
     dependencies:
-      tslib: 2.4.1
+      tslib: 2.4.0
     dev: false
     engines:
       node: '>=12.0.0'
     resolution:
       integrity: sha512-TrRLIoSQVzfAJX9H1JeFjzAoDGcoK1IYX1UImfceTZpsyYfWr09Ss1aHW1y5TrrR3iq6RZLBwJ3E24uwPhwahw==
+  /@azure/communication-calling/1.8.0-beta.1:
```
JamesBurnside, line 37: revert this (looks like the whole lockfile should be reverted)

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatAdapter.ts
JamesBurnside: Right, but technically anyone using the `CallWithChatAdapterManagement` interface directly will now hit a breaking change.
In actuality, ARB wants us to have flat interfaces anyway, perhaps (best in a separate PR if we do) we leave this interface as is, mark as deprecated and not used, then update the whole interface to be flat (internally we can still use common interfaces to ensure type correctness across different adapters)

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatAdapter.ts
```diff
@@ -386,12 +386,22 @@ export interface CallWithChatAdapterSubscriptions {
...
     CallWithChatAdapterSubscriptions {}
 
+/**
+ * @public
+ */
+export interface CallWithChatAdapter extends CallWithChatAdapterCommon {
+  joinCall(microphoneOn?: boolean): Call | undefined;
+  startCall(participants: string[], options?: StartCallOptions): Call | undefined;
+  /* @conditional-compile-remove(PSTN-calls) */
+  startCall(participants: CommunicationIdentifier[], options?: StartCallOptions): Call | undefined;
```
JamesBurnside, line 402: just something I'm noticing here but not part of this PR, I didn't think you can have two overloaded functions with the same number of arguments. Should this not need to be:
```suggestion
  startCall(participants: CommunicationIdentifier[] | string[], options?: StartCallOptions): Call | undefined;
```

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -954,6 +1055,24 @@ export const createAzureCommunicationCallAdapterFromClient: (
...
+ */
+export const createAzureCommunicationTeamsCallAdapterFromClient = async (
+  callClient: StatefulCallClient,
+  callAgent: TeamsCallAgent,
+  locator: CallAdapterLocator
+): Promise<TeamsCallAdapter> => {
+  const deviceManager = (await callClient.getDeviceManager()) as StatefulDeviceManager;
+  return new AzureCommunicationCallAdapter(callClient, locator, callAgent, deviceManager);
+};
+
```
JamesBurnside, line 1075: this file is now over 1000 lines, can we break up logic in this file into more isolated blocks in their own files

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -350,28 +360,39 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
-  private _joinCall(audioOptions: AudioOptions, videoOptions: VideoOptions): Call {
+  private _joinCall(audioOptions: AudioOptions, videoOptions: VideoOptions): CallTypeOf<AgentType> {
     const isTeamsMeeting = 'meetingLink' in this.locator;
     /* @conditional-compile-remove(rooms) */
     const isRoomsCall = 'roomId' in this.locator;
 
+    /* @conditional-compile-remove(teams-identity-support) */
+    if (_isTeamsCallAgent(this.callAgent)) {
+      if (isTeamsMeeting) {
+        return this.callAgent.join(this.locator as TeamsMeetingLinkLocator, {
```
JamesBurnside, line 371: Avoid this duplicate block which is the same as below. Instead invert the if:
```ts
if (_isTeamsCallAgent(this.callAgent) && !isTeamsMeeting) {
  throw new Error('Locator not supported by TeamsCallAgent');
}
...
if (isTeamsMeeting) {
  ...
```

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -238,13 +248,13 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
...
 
-    this.handlers = createDefaultCallingHandlers(callClient, callAgent, deviceManager, undefined);
-
     this.onClientStateChange = onStateChange;
 
     this.subscribeDeviceManagerEvents();
 
     this.callClient.onStateChange(onStateChange);
+
+    this.handlers = createHandlers(callClient, callAgent, deviceManager, undefined);
```
JamesBurnside, line 257: did this move for a specific reason (i.e. this now _must_ be called after over functions)? If so add a comment that this must be called at the end

change-beta/@azure-communication-react-3ae22222-4a77-4173-80b3-ec236bef5e2f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Add CTE support to adapter layer",
```
prprabhu-ms, line 3: Let's not use `CTE` in the changelog. Can this be more descriptive? `Add TeamsCallAdapter etc. to support composites powered by teams identity` or some such?

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -242,7 +242,13 @@ export interface BrowserPermissionDeniedStyles extends BaseCustomStyles {
 }
 
 // @public
-export interface CallAdapter extends AdapterState<CallAdapterState>, Disposable, CallAdapterCallManagement, CallAdapterDeviceManagement, CallAdapterSubscribers {
+export interface CallAdapter extends CommonCallAdapter {
+    // (undocumented)
```
prprabhu-ms, line 246: All public API methods should be documented.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -689,8 +695,18 @@ export interface CallState {
     type: 'Teams' | 'ACS';
 }
 
+// @public (undocumented)
```
prprabhu-ms, line 698: Document this type and its methods.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2956,6 +2982,16 @@ export interface SystemMessageCommon extends MessageCommon {
     messageType: 'system';
 }
 
+// @beta
+export interface TeamsCallAdapter extends CommonCallAdapter {
+    // (undocumented)
```
prprabhu-ms, line 2987: Document these methods.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
prprabhu-ms: I don't think this moved for any specific reason.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -805,41 +826,38 @@ export const createAzureCommunicationCallAdapter = async ({
...
+  });
+  const adapter = createAzureCommunicationTeamsCallAdapterFromClient(callClient, callAgent, locator);
+  return adapter;
+};
+
+/**
+ * @private
+ */
+const useAzureCommunicationCallAdapterGeneric = <
+  AdapterType extends CallAdapter | /* @conditional-compile-remove(teams-identity-support) */ TeamsCallAdapter
```
prprabhu-ms, line 854: Implement this without conditional compilation. It is internal only, and the type `TeamsCallAdapter` is available here in both build flavors.

packages/react-composites/src/composites/CallComposite/adapter/createHandlers.ts
```diff
@@ -0,0 +1,47 @@
...
+  _isACSCallAgent,
+  _isTeamsCall,
+  _isTeamsCallAgent
+} from '@internal/calling-stateful-client';
+
+/**
+ * @private
+ */
+export type CallHandlersOf<AgentType extends CallAgent | TeamsCallAgent> = AgentType extends CallAgent
+  ? CallingHandlers
```
prprabhu-ms, line 23: Do you need this conditional compilation? Or can you set it up such that for stable build `TeamsCallingHandlers` is `never`?

prprabhu-ms (approved):
The elephant in the room is James' comment about a breaking API change. Otherwise LGTM.

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatAdapter.ts
PorterNan: I mentioned this in another PR for startCall as well, I don't remember the detail of this one, but this was what we end up with in the end, @anjulgarg please correct me if I am wrong

PorterNan: BTW, for "have two overloaded functions with the same number of arguments", we have some examples from "on" functions in calling sdk subscription. (But I still preferred "or" style than override if that is possible)

packages/react-composites/src/composites/CallComposite/adapter/createHandlers.ts
PorterNan: This is the only place to call TeamsCallingHandlers outside of their package, so I guess keeping it here would be nice to avoid adding another type into stable api as a never type (we will have to export TeamsCallingHandlers in communication-react)

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: Previously it was because conditional compilation, but we extract the logic to createHandlers in the end, so I think it's good to move it back to the place

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: Noticed there are 2 teams meeting condition here - one for CTE, one for regular ACS. So _isTeamsCallAgent(this.callAgent) still need to be there as a type guard, but I applied if(!isTeamsMeeting) {throw ""} to avoid an else closure :D

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: Nice catch, @prprabhu-ms and me also discussed this yesterday and we decided to open another work item https://skype.visualstudio.com/SPOOL/_workitems/edit/3067003 for refactoring, because:
1. There is no impact on current api design
2. The refactor would be fairly large, which might involve more merge conflict for this PR (and this PR is already big enough)

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatAdapter.ts
PorterNan: Although I thought there would be very few people directly use CallWithChatAdapterManagement, but I still need to say, you are right, let's not lie to semver number, this is a breaking change.

I tried flattening apis, but it turned out to be a big change for api, and flattening those apis is not the focus of this PR, so I prefer not to do it here in the PR.

The solution to make minimum change here is that I made some changes to current api - I had to add one more interface to isolate "joinCall" "startCall" functions, then once flattening happens, we just need to deprecate all of those together and we are good.

And good news here - we don't have to make api changes to CallWithChat anymore

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -245,16 +245,28 @@ export interface BrowserPermissionDeniedStyles extends BaseCustomStyles {
...
+    startCall(participants: CommunicationIdentifier[], options?: StartCallOptions): Call | undefined;
 }
 
 // @public
 export type CallAdapterCallEndedEvent = {
     callId: string;
 };
 
+// @public @deprecated
+export interface CallAdapterCallManagement extends CallAdapterCallOperations {
```
JamesBurnside, line 261: aha clever way to work around that! I like it!

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -245,16 +245,28 @@ export interface BrowserPermissionDeniedStyles extends BaseCustomStyles {
...
+    joinCall(microphoneOn?: boolean): Call | undefined;
+    startCall(participants: string[], options?: StartCallOptions): Call | undefined;
+    // @beta
+    startCall(participants: CommunicationIdentifier[], options?: StartCallOptions): Call | undefined;
+}
+
 // @public
-export interface CallAdapterCallManagement {
+export interface CallAdapterCallOperations {
```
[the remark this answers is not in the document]
JamesBurnside: Note: we have a ton of these interfaces we currently export, would be great to go through mark them all as deprecated and just export the flat interfaces, but that's a big item of work but this use case highlights how it would reduce tech debt by allowing us to be flexible with interfaces that need only be internal-only

JamesBurnside (approved):

PR #2427, opened 2022-10-21
Update/remove further dependencies that relied on react 17

JamesBurnside:
# What
Update following deps that listed react 17 as peer dep that need updating:
* `@fluentui/react-hooks`
* `@fluentui/react-window-provider`

Also removed unused dependencies that specified react 17 as a peer dep in the lockfile for cleanup. These don't impact the npm package but will make the repo upgrade to react 18 easier.

# Why
Continuation of #2415 to support react 18

To see the full list of items to upgarde see this doc: https://microsoft.sharepoint-df.com/:w:/t/IC3SDK/EYInxewzAmJDvJaoCZhLTgsBvmY_YLpa82v-zTp3WDbhgA?e=DSNdP8

# How Tested
Relying on CI for dep bumps

PR #2428, opened 2022-10-21
Require webpack 5 in our lib. Add minimum webpack and typescript versions to storybook. Update `html-to-parser` version

JamesBurnside:
# What
* Update `html-to-parser` version to the latest version that supports react 18
* Add documentation for minimum typescript and webpack support
* Update storybook manager to use webpack 5 (to avoid https://github.com/fb55/htmlparser2/issues/1237)
* Update tools/check-treeshaking to use babel workaround for: https://github.com/fb55/htmlparser2/issues/1237

# Why
Continuation of #2415 and #2427 to support react 18.

# How Tested
Verified storybook ran locally.


packages/storybook/stories/Troubleshooting.stories.mdx
```diff
@@ -61,3 +61,22 @@ by `npm add`.
...
+|     return getFeed(parseDOM(feed, options));
+| }
+> export * as DomUtils from "domutils";
+| // Old name for DomHandler
+| export { DomHandler as DefaultHandler };
+```
+
+This issue will occur when using `@azure/communication-react` with a webpack version < 5.
+The best course of action is to upgrade your webpack version to 5.
+If this is not possible take a look at these workarounds: https://github.com/fb55/htmlparser2/issues/1237#issuecomment-1182522861.
```
prprabhu-ms, line 82:
```suggestion
If this is not possible, take a look at [these workarounds](https://github.com/fb55/htmlparser2/issues/1237#issuecomment-1182522861).
```

prprabhu-ms (approved):

prprabhu-ms:
How do you feel about holding this PR back until we cut the next stable release branch. It's been a while since the last stable release, and it'd be good to release this in a separate, smaller, stable release, clearly calling out the change.

JamesBurnside:
> How do you feel about holding this PR back until we cut the next stable release branch. It's been a while since the last stable release, and it'd be good to release this in a separate, smaller, stable release, clearly calling out the change.

Sure thing!

prprabhu-ms:
Feel free to merge this when ready. `1.4.0` release branch has been created: https://github.com/Azure/communication-ui-library/tree/release/1.4.0

packages/storybook/package.json
```diff
@@ -53,6 +53,8 @@
     "@storybook/addon-essentials": "~6.5.7",
     "@storybook/addon-links": "~6.5.7",
     "@storybook/addon-storyshots": "~6.5.7",
+    "@storybook/builder-webpack5": "6.5.12",
```
PorterNan, line 56: There are some other packages like @storybook/api are still staying in ~6.5.7, I guess we might want to align them together? (maybe make them all into ~6.5.12?)

PorterNan (approved):

packages/storybook/package.json
JamesBurnside: We probably should but as there's no errors I imagine it will fall into 'isn't broke won't fix it'. In saying that storybook7 is currently in development and on like alpha v42 so easiest to wait for that then do a big upgrade!

PR #2429, opened 2022-10-21
[DTMF dial pad] fix unexpected string behavior

dmceachernmsft:
# What

Fix behavior of dtmf dialpad so end users dont lose track of last tone entered.
https://user-images.githubusercontent.com/94866715/197297184-d441e0f0-92ac-4681-b642-c7f355f6dfcd.mp4
# Why

inputs were being lost after certain characters were entered on the dialpad

https://skype.visualstudio.com/SPOOL/_workitems/edit/3015850
# How Tested

tested locally to verify correct behavior


carocao-msft (approved):

packages/react-composites/src/composites/common/SendDtmfDialpad.tsx
```diff
@@ -40,7 +40,11 @@ export const SendDtmfDialpad = (props: SendDtmfDialpadProps): JSX.Element => {
   const [textFieldValue, setTextFieldValue] = useState<string>();
 
   const onChange = (input: string): void => {
-    setTextFieldValue(input);
+    if (input.length > 25) {
+      setTextFieldValue('.' + '.' + '.' + input.slice(4));
```
anjulgarg, line 44: reading the below code is slightly easier.
```tsx
setTextFieldValue(`...${input.slice(4)}`);
```

packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -340,5 +340,5 @@ export const Dialpad = (props: DialpadProps): JSX.Element => {
 
 const sanitizeInput = (input: string): string => {
   // remove non-valid characters from input: letters,special characters excluding +, *,#
-  return input.replace(/[^\d*#+]/g, '');
+  return input.replace(/[^\d*#+.]/g, '');
```
anjulgarg, line 343: The `.` is a wildcard that matches everything. This regex doesn't seem right to me

anjulgarg: OR are we saying that a dot is also allowed
If yes, why is a dot also allowed?

dmceachernmsft: the dot is now allowed since thats what we are appending to the string to create the `...`

anjulgarg (approved):

PR #2433, opened 2022-10-24
Add troubleshooting guide error bar to call composite config screen 

carocao-msft:
# What
Add troubleshooting guide error bar to call composite config screen 
Screen Shot 2022-10-24 at 8 35 24 AM

# Why
call readiness work

PR #2434, opened 2022-10-24
Add allow camera and microphone access drawer to call composite mobile config screen 

carocao-msft:
# What
Add allow camera and microphone access drawer to call composite mobile config screen 

# Why
users can click on allow access to receive device permission prompt 
before this implementation we don't have anyway to prompt for access on mobile 

# How Tested

https://user-images.githubusercontent.com/96077406/197587394-3d007992-149d-489d-a5bd-b5fef9d9f228.mov

Also tested by setting callreadinessoptin to false in app.tsx and verified the feature disappears
Accessibility tested


packages/react-components/src/components/DomainPermissions.tsx
```diff
@@ -84,13 +84,13 @@ const DomainPermissionsContainer = (props: DomainPermissionsProps): JSX.Element
         </Stack>
       </Stack>
       <Stack styles={textContainerStyles}>
-        <Text styles={primaryTextStyles}>{_formatString(strings.primaryText, { appName: appName })}</Text>
-        <Text styles={secondaryTextStyles}>{strings.secondaryText}</Text>
+        <Text styles={primaryTextStyles}>{_formatString(strings?.primaryText ?? '', { appName: appName })}</Text>
```
dmceachernmsft, line 87: odd situation here but what happens if the strings object is undefined here? does it just display the appName? I think `_formatString` tries to put those two strings together by the `appName` attribute in the primaryText string. does it just return an empty string in this case?

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -108,11 +112,50 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
+  /* @conditional-compile-remove(call-readiness) */
+  const drawerStyle: _DrawerSurfaceStyles = {
+    root: {
+      position: 'absolute',
+      top: 0,
+      left: 0,
+      width: '100%',
+      height: '100%',
+      // apply zindex = 99 so drawer appear over device buttons and other components in the config page
+      zIndex: 99
```
dmceachernmsft, line 131: should we maybe have this just 1 above the highest layer in the composites config page? avoid this magic number?

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -108,11 +112,50 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
         <ErrorBar {...errorBarProps} />
       </Stack>
+
+      {
+        /* @conditional-compile-remove(call-readiness) */
+        mobileView && isDrawerShowing && (
+          <_DrawerSurface onLightDismiss={onLightDismissTriggered} styles={drawerStyle}>
+            <DomainPermissions
+              appName={'app'}
+              onTroubleshootingClick={() => console.log('clicked trouble shooting link')}
```
JamesBurnside, line 147: remove console log here

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
JamesBurnside: Can we use common or existing styles, drawer is shown in other places:
image

carocao-msft: We can't coz the buttons are set as zindex: 11

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
carocao-msft: will link the ontroubleshootingclick with onpermissionTroubleShootingClick in another PR, need this one to be merged first https://github.com/Azure/communication-ui-library/pull/2433, in this pr I passed  onpermissionTroubleShootingClick down from composite to config screen

packages/react-components/src/components/DomainPermissions.tsx
carocao-msft: in the code below string is populated by locale so it wont be undefined, just doing this here to get pass lint errors. We need string prop to be optional

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
dmceachernmsft: could we make another PR to bring that z-index down if it dosen't break anything? would that be considered a breaking change @JamesBurnside since some Contosos might have styled their apps around that?

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
dmceachernmsft: This should be the troubleshootingClick prop from the composite itself I think?

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
JamesBurnside: They should be set to 11 for a reason - every z-index should be something + 1 instead of arbitrary values

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
JamesBurnside: Just a nit, remove this line and then add it appropriately in #2433. Reason being another PR could be blocked or abandoned for any reason so we want to keep the codebase in a good state at all times (not always possible with stacked PRs though but something to strive for)

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
JamesBurnside: We should still aim to use common styles though, if someone needs to add some new css property to all uses of the drawer they'll need to add it in x different places and could easily miss one.
Consider 
```ts
/**
  Drawer styles to be used to house the _DrawerComponent on top of all other content on the screen.
  @param zIndex: this defaults to MAX_Z_INDEX if unset
*/
const drawerContainerStyles = (zIndex: number = MAX_Z_INDEX): IStackStyles => ({
  root: {
   ...etc.
  }
});
```

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -108,11 +112,50 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
+
   return (
     <Stack className={mobileView ? configurationContainerStyleMobile : configurationContainerStyleDesktop}>
       <Stack styles={bannerNotificationStyles}>
         <ErrorBar {...errorBarProps} />
       </Stack>
+
+      {
+        /* @conditional-compile-remove(call-readiness) */
+        mobileView && isDrawerShowing && (
```
JamesBurnside, line 143: if they've already accepted permissions and don't need to accept permissions, will this still always show?

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -108,11 +112,50 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
+          <_DrawerSurface onLightDismiss={onLightDismissTriggered} styles={drawerStyle}>
+            <DomainPermissions
+              appName={'app'}
+              onTroubleshootingClick={() => console.log('clicked trouble shooting link')}
+              onAllowAccessClick={async () => {
+                await adapter.askDevicePermission({ video: true, audio: true });
+                adapter.queryCameras();
+                adapter.queryMicrophones();
+                adapter.querySpeakers();
+                setIsDrawerShowing(false);
```
JamesBurnside, line 153: for a future PR but flagging here for posterity; if they have denied access to the permissions we want to show the appropriate message here

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
carocao-msft: Yes same with enable mic/camera buttons

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
carocao-msft: fixed

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
carocao-msft: fixed

packages/react-components/src/components/DomainPermissions.tsx
carocao-msft: fixed

dmceachernmsft (approved):

packages/react-components/src/components/DomainPermissions.tsx
```diff
@@ -84,14 +88,18 @@ const DomainPermissionsContainer = (props: DomainPermissionsProps): JSX.Element
...
       <Stack styles={textContainerStyles}>
-        <Text styles={primaryTextStyles}>{_formatString(strings.primaryText, { appName: appName })}</Text>
-        <Text styles={secondaryTextStyles}>{strings.secondaryText}</Text>
+        {strings && isValidString(strings?.primaryText) && (
+          <Text styles={primaryTextStyles}>{_formatString(strings.primaryText, { appName: appName })}</Text>
+        )}
+        <Text styles={secondaryTextStyles}>{strings?.secondaryText}</Text>
         {onAllowAccessClick && (
-          <PrimaryButton styles={primaryButtonStyles} text={strings.primaryButtonText} onClick={onAllowAccessClick} />
+          <PrimaryButton styles={primaryButtonStyles} text={strings?.primaryButtonText} onClick={onAllowAccessClick} />
```
JamesBurnside, line 96: You will need the same `strings && isValidString` for each of these, otherwise there will be an empty button with no text here (and apply the same to the secondary text element above)

packages/react-composites/src/composites/CallWithChatComposite/styles/CallWithChatCompositeStyles.ts
```diff
@@ -33,8 +35,12 @@ export const controlBarContainerStyles: IStackStyles = {
...
 };
 
-/** @private */
-export const drawerContainerStyles: IStackStyles = {
+/**
+ * @private
+ * Drawer styles to be used to house the _DrawerComponent on top of other content on the screen.
+ * @param zIndex: this defaults to DEFAULT_Z_INDEX if unset
+ */
+export const drawerContainerStyles = (zIndex: number = DEFAULT_Z_INDEX): IStackStyles => ({
```
JamesBurnside, line 43: Can this just import the `drawerContainerStyles` from the call composite dir?

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -149,8 +166,35 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
           errorBarProps={errorBarProps}
+          /* @conditional-compile-remove(call-readiness) */
+          callReadinessOptedIn={callReadinessOptedIn}
         />
       </Stack>
+
+      {
+        /* @conditional-compile-remove(call-readiness) */
+        mobileView && isDrawerShowing && callReadinessOptedIn && (
+          <_DrawerSurface onLightDismiss={onLightDismissTriggered} styles={drawerContainerStyles(99)}>
```
JamesBurnside, line 177: 99 should be defined somewhere as a constant, even just near the top of this file
```
const DRAWER_HIGH_Z_BAND = 99; // comment on why we choose an arbitrary 99
```

JamesBurnside (approved):

packages/react-components/src/components/DomainPermissions.tsx
```diff
@@ -64,14 +64,18 @@ export interface DomainPermissionsStrings {
    * Primary button text string.
    */
   primaryButtonText: string;
+  /**
+   * aira label describing the content of the container
```
edwardlee-msft, line 68: "Aria label describing..."

JamesBurnside (approved):

PR #2435, opened 2022-10-24
[Lobby Screen] Tests to ensure correct strings are shown for each calling modality

dmceachernmsft:
# What

Tests added for lobby screen in Call Composite
# Why

Tests ensure that the correct strings are shown for each Calling modality.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3021395
# How Tested

Ran tests locally will build snapshots in CI


packages/react-composites/tests/browser/call/hermetic/LobbyScreen.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import { IDS } from '../../common/constants';
+import { dataUiId, isTestProfileStableFlavor, stableScreenshot, waitForSelector } from '../../common/utils';
+
+test.describe('Lobby page tests', async () => {
+  test('Lobby page shows correct strings when joining a group call', async ({ page, serverUrl }) => {
+    const dina = defaultMockRemoteParticipant('Dina');
+    /**
+     * remote participants reflect as this state for a moment before the user connects.
+     *
+     * Dina exists to demonstrate that we don't see the 'Calling' string when joining a group call as a flash.
```
prprabhu-ms, line 21: <3

packages/react-composites/tests/browser/call/hermetic/LobbyScreen.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (initialState.call) {
+      initialState.call.state = 'Connecting';
+    }
+    initialState.page = 'lobby';
+
+    await page.goto(buildUrlWithMockAdapter(serverUrl, initialState));
+
+    await waitForSelector(page, dataUiId(IDS.lobbyScreenTitle));
+
+    expect(await stableScreenshot(page)).toMatchSnapshot('lobbyPage-ACScall.png');
```
prprabhu-ms, line 56: Should this ben renamed to `lobbyPage-outboundCall` (a group call is also an ACS call... and we avoid the use of the acronym ACS).

prprabhu-ms: Also, nitty nit on snapshot names. All the snapshots currently-use-dash-and-word. This name is mixing camelCase with that and messing with me :D

prprabhu-ms (approved):
Thank you!

anjulgarg (approved):

anjulgarg:
Thanks for adding these tests. This was a big peace we weren't testing

packages/react-composites/tests/browser/call/hermetic/LobbyScreen.test.ts
dmceachernmsft: will do! 😊

PR #2436, opened 2022-10-24
Remove DeclarativeCallAgent from stable API

prprabhu-ms:
This is new API needed for a beta-only feature. Let's remove it from stable builds for now.


dmceachernmsft (approved):

PorterNan (approved):

packages/calling-stateful-client/src/StatefulCallClient.ts
```diff
@@ -128,6 +128,7 @@ export interface StatefulCallClient extends CallClient {
     stream: LocalVideoStreamState | RemoteVideoStreamState
   ): void;
 
+  /** @conditional-compile-remove(one-to-n-calling) */
```
anjulgarg, line 131: Why is this added here?

PR #2441, opened 2022-10-24
Regenerate the PNPM shrinkwrap files for beta flavor

prprabhu-ms:
`1.4.0` is a stable release, but the GitHub workflows are failing in an early step where then need to install beta build's dependencies.

Fix it first while I try to figure out how we got into this state.


prprabhu-ms:
#2442 fixes the release branch creation workflow to avoid this manual fixup

PR #2453, opened 2022-10-26
Fix: blank video gallery

JamesBurnside:
# What
Wrap participant subscribers in a try-catch

**Note: this only happened because I was on a stable version of calling but running beta code not stable code. So feel free to reject PR** (however it did cause me much pain)

# Why
I was debugging and suddenly when I joined a call locally I had no one else in the video gallery. I could hear everyone else, chat messages came through but it showed no people in the video gallery or participant pane.

After much debugging I learned the calling sdk onParticipantsUpdated _was_ being called, but for some reason our state never had any remote participants in it.

Turns out the participant subscriber was near-silently throwing an exception (just a hard to spot warning only seen if the azure logger is turned on):
image

The exception is now caught on our side, allowing the participants to be propagated to state, and the error is made loud and proud:
image

# How Tested
Left: same call but without fix, Right: same call but with fix:
image

### DIFF
--- a/packages/calling-stateful-client/src/CallSubscriber.ts
+++ b/packages/calling-stateful-client/src/CallSubscriber.ts
@@ -160,10 +160,18 @@ export class CallSubscriber {
 
   private remoteParticipantsUpdated = (event: { added: RemoteParticipant[]; removed: RemoteParticipant[] }): void => {
     event.added.forEach((participant: RemoteParticipant) => {
-      this.addParticipantListener(participant);
+      try {
+        this.addParticipantListener(participant);
+      } catch (e) {
+        console.error(e);
+      }
     });
     event.removed.forEach((participant: RemoteParticipant) => {
-      this.removeParticipantListener(participant);
+      try {
+        this.removeParticipantListener(participant);
+      } catch (e) {
+        console.error(e);
+      }
     });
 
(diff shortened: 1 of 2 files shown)


packages/calling-stateful-client/src/CallSubscriber.ts
```diff
@@ -160,10 +160,18 @@ export class CallSubscriber {
 
   private remoteParticipantsUpdated = (event: { added: RemoteParticipant[]; removed: RemoteParticipant[] }): void => {
     event.added.forEach((participant: RemoteParticipant) => {
-      this.addParticipantListener(participant);
+      try {
+        this.addParticipantListener(participant);
+      } catch (e) {
+        console.error(e);
```
prprabhu-ms, line 166: We shouldn't have to `console` log the error here.

The real mystery is why the exception that escaped here wasn't bubbling all the way up and getting reported on console already.

prprabhu-ms:
Bumping off my attention set. Please re-assign when you pick this up again.

JamesBurnside:
Closing as I don't feel I have time to investigate this,, the error was only because I beta prop was used in stable, and the error is still technically in the console, just very hard to spot and requires the azure logger to be on

PR #2459, opened 2022-10-27
fixing up the setup section snippets

alkwa-msft:
I found I had to make some changes to the setup section when starting to use some of the samples. I think we should aspire to making the initial setup as close to real as possible.

- Updated the index.tsx replacement file for < React18 to not reference a snippet file
- Added a dependency for @testing-library/react that requires React16 instead of 18

https://skype.visualstudio.com/SPOOL/_workitems/edit/3032434

# Why
Developers are already treading into the unknown. I think we want make sure we make the first try as seamless as possible. We are already asking them to do several steps and having to also fix up some of our snippets is more than I would expect them to do.

# How Tested
Tested locally

# Process & policy checklist


JamesBurnside (approved):

alkwa-msft:
Closing since Prathmesh made a Pr around the same thing

PR #2462, opened 2022-10-27
Clean up storybook quickstart setup page

prprabhu-ms:
* Remove a license header from a code snippet.
* Fix the code snippet to work out of the box in a new CRA app.
* Fix the suggested default dependencies so that a new CRA app works correctly out of the box.
* Update references to beta package of `@azure/communication-calling`.


packages/storybook/stories/snippets/App.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
prprabhu-ms, line 1: It is OK that this file isn't named `*.snippet.tsx` because it gets imported by `ReactIndex.snippet.tsx` and gets tested that way.

alkwa-msft (approved):

carocao-msft (approved):

PR #2463, opened 2022-10-28
Add allow access modal to configuration screen 

carocao-msft:
# What
Add allow access modal to configuration screen 
Screenshot 2022-10-28 at 1 48 29 AM

# Why
This is part of call readiness feature 

# How Tested

https://user-images.githubusercontent.com/96077406/198546424-e8d7b4bd-ba9d-4a72-a429-d83a42aac10c.mov

Also tested by setting callreadinessoptin to false in app.tsx and verified the feature disappears 
Accessibility tested


packages/react-components/src/components/BrowserPermissionDeniedIOS.tsx
```diff
@@ -34,7 +34,7 @@ export interface BrowserPermissionDeniedIOSProps extends BrowserPermissionDenied
   /**
    * Localization strings for BrowserPermissionDeniedIOS component.
    */
-  strings: BrowserPermissionDeniedIOSStrings;
+  strings?: BrowserPermissionDeniedIOSStrings;
```
dmceachernmsft, line 37: If these strings are now optional we should make sure that the component is getting its own localized strings instead of its parent.

dmceachernmsft (commented):
overall looks good! something I wonder though is should that modal be showing for when the microphone permissions are being asked for as well? I was thinking about this watching the video because microphone is the barrier to entry for starting a call.

carocao-msft:
> overall looks good! something I wonder though is should that modal be showing for when the microphone permissions are being asked for as well? I was thinking about this watching the video because microphone is the barrier to entry for starting a call.

It will, the modal will show up if one of the 2 permissions are not set. If it has shown up once, it shouldn't show up again

packages/react-components/src/components/BrowserPermissionDeniedIOS.tsx
carocao-msft: They are using localized strings

carocao-msft: const locale = useLocale().strings.BrowserPermissionDeniedIOS;

dmceachernmsft: Oh I didnt see it in the expanded file, will look again 😊

dmceachernmsft:
> > overall looks good! something I wonder though is should that modal be showing for when the microphone permissions are being asked for as well? I was thinking about this watching the video because microphone is the barrier to entry for starting a call.
> 
> It will, the modal will show up if one of the 2 permissions are not set. If it has shown up once, it shouldn't show up again

I think we need to be a little more controlling here. the permissions for microphone should be a blocker for the modal like it is with the startCall button. that modal disappears the user might think that everything they need is set if we are showing it in the first place.

carocao-msft:
> > > overall looks good! something I wonder though is should that modal be showing for when the microphone permissions are being asked for as well? I was thinking about this watching the video because microphone is the barrier to entry for starting a call.
> > 
> > 
> > It will, the modal will show up if one of the 2 permissions are not set. If it has shown up once, it shouldn't show up again
> 
> I think we need to be a little more controlling here. the permissions for microphone should be a blocker for the modal like it is with the startCall button. that modal disappears the user might think that everything they need is set if we are showing it in the first place.

The modal needs to disappear for user to be able to click on the other enable button. For the scenario when user denied permission, we have a separate screen that will show up and block them from further actions (need permission api for this)

dmceachernmsft:
> > > > overall looks good! something I wonder though is should that modal be showing for when the microphone permissions are being asked for as well? I was thinking about this watching the video because microphone is the barrier to entry for starting a call.
> > > 
> > > 
> > > It will, the modal will show up if one of the 2 permissions are not set. If it has shown up once, it shouldn't show up again
> > 
> > 
> > I think we need to be a little more controlling here. the permissions for microphone should be a blocker for the modal like it is with the startCall button. that modal disappears the user might think that everything they need is set if we are showing it in the first place.
> 
> The modal needs to disappear for user to be able to click on the other enable button. For the scenario when user denied permission, we have a separate screen that will show up and block them from further actions (need permission api for this)

would it make sense then to have if you click on one drop down it queries for both? it seems like a lot of feedback to the end user to have the dialog clear after just one device is enabled.

carocao-msft:
> p down it queries for both? it seems like a lot of feedback to the end user to have the dialog clear after just one device is enabled.

I don't think we should prompt for mic access when user click on allow camera access. For the modal since user already seen it once and read it, and followed what it said to allow device access,  it should not show up again or it would be repetitive?

carocao-msft:
> > > > > overall looks good! something I wonder though is should that modal be showing for when the microphone permissions are being asked for as well? I was thinking about this watching the video because microphone is the barrier to entry for starting a call.
> > > > 
> > > > 
> > > > It will, the modal will show up if one of the 2 permissions are not set. If it has shown up once, it shouldn't show up again
> > > 
> > > 
> > > I think we need to be a little more controlling here. the permissions for microphone should be a blocker for the modal like it is with the startCall button. that modal disappears the user might think that everything they need is set if we are showing it in the first place.
> > 
> > 
> > The modal needs to disappear for user to be able to click on the other enable button. For the scenario when user denied permission, we have a separate screen that will show up and block them from further actions (need permission api for this)
> 
> would it make sense then to have if you click on one drop down it queries for both? it seems like a lot of feedback to the end user to have the dialog clear after just one device is enabled.

The overall flow should be: click on allow camera access, see the modal, click allow, then click on mic access, click allow again then join call. or click on allow mic access, see the modal, click allow, then click on camera access and allow; 
In the case if user clicked denied, a second modal would show up to guide them to go through system preferences to allow access and that is not a dismissable modal

packages/react-components/src/components/BrowserPermissionDenied.tsx
```diff
@@ -94,16 +94,21 @@ const BrowserPermissionDeniedContainer = (props: BrowserPermissionDeniedProps):
...
-          text={strings.primaryButtonText}
-          onClick={onTryAgainClick}
-        />
-        <Link styles={mergeStyleSets(linkTextStyles, styles?.troubleshootingLink)} onClick={onTroubleshootingClick}>
-          {strings.linkText}
-        </Link>
+        <Text styles={primaryTextStyles}>{strings?.primaryText}</Text>
+        <Text styles={secondaryTextStyles}>{strings?.secondaryText}</Text>
+        {onTryAgainClick && (
+          <PrimaryButton
```
JamesBurnside, line 100: Same as https://github.com/Azure/communication-ui-library/pull/2434/files#diff-86f3474befa4ad506ce8c7c7f9d1de9345eb84c5a14726dc9935af0180f0ccdcR91 , each of these needs to check if the string is valid. If not the react element should not be rendered

packages/react-components/src/components/BrowserPermissionDeniedIOS.tsx
```diff
@@ -109,55 +109,55 @@ const BrowserPermissionDeniedIOSContainer = (props: BrowserPermissionDeniedIOSPr
...
             style={{ padding: '1.5rem' }}
           />
         </Stack>
       )}
       <Stack style={{ padding: '1rem', paddingBottom: '1.25rem' }}>
         <Stack styles={textContainerStyles}>
-          <Text styles={primaryTextStyles}>{strings.primaryText}</Text>
-          <Text styles={secondaryTextStyles}>{strings.secondaryText}</Text>
-          {strings.step1Text && (
+          <Text styles={primaryTextStyles}>{strings?.primaryText}</Text>
```
JamesBurnside, line 120: Likewise here, now these strings are all optional they should follow the same pattern https://github.com/Azure/communication-ui-library/pull/2434/files#diff-86f3474befa4ad506ce8c7c7f9d1de9345eb84c5a14726dc9935af0180f0ccdcR91

packages/react-components/src/components/DevicePermissionDropdown.tsx
```diff
@@ -91,6 +95,9 @@ export const _DevicePermissionDropdown = (props: _DevicePermissionDropdownProps)
         if (askDevicePermission) {
           askDevicePermission(constrain ?? { video: true, audio: true });
         }
+        if (onClick) {
```
JamesBurnside, line 98: nit: you can shorten this to `onClick?.()`

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -150,6 +152,14 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
     // only way to dismiss this drawer is clicking on allow access which will leads to device permission prompt
   };
 
+  /* @conditional-compile-remove(call-readiness) */
+  const [isModalShowing, setIsModalShowing] = useState(false);
+
+  /* @conditional-compile-remove(call-readiness) */
+  const onClickEnableDevicePermission = (): void => {
```
JamesBurnside, line 159: memoize

dmceachernmsft (approved):

Issue #2425, opened 2022-10-20
ChatAdapter errors is not thrown to the error listener


DercilioFontes:
**Describe the bug; what happened?**
I implemented the error handler for the `ChatAdapter` on the error listener, but the errors are not coming to it. I can see the error in the console logs.

```typescript
export const errorCodes = [401, 403, 404];

...

const errorHandler = (e: ChatAdapterError) => {
  if (e.innerError.statusCode && errorCodes.includes(e.innerError.statusCode)) {
    chatAdapter?.dispose();
  }
};

...

chatAdapter.on('error', errorHandler);

...

<ChatComposite
    fluentTheme={customTheme}
    onFetchAvatarPersonaData={() => Promise.resolve(avatarPersonaData)}
    options={{ topic: false }}
    adapter={chatAdapter}
/>
```
Image:
![Screen Shot 2022-10-20 at 5 16 58 PM](https://user-images.githubusercontent.com/20172847/197063280-d4093110-dfbb-44d9-bb8a-6a66c3f1e8af.png)

**What are the steps to reproduce the issue?**
1. Start a regular chat
2. Leave one side, ending the thread in the server side
3. Continue typing on the other side

In this case, we implemented to end the chat thread on the server side when one closes the browser, "unload" event. So, no event of removing the participant is done to be listened to on the other side. When the other side continues typing, it gets the error "Not Found." But the `ChatAdapter` doesn't throw it to the error listener.

**What behavior did you expect?**
Errors are to be thrown to the `ChatAdapter` error listener.

**If applicable, provide screenshots:**
Added above

**In what environment did you see the issue?**
    `@azure/communication-chat`: "^1.2.0",
    `@azure/communication-common`: "^2.1.0",
    `@azure/communication-react`: "^1.3.1",

- OS & Device: MacOS 12.6
- Browser: Google Chrome / Mozilla FireFox
- Browser Version: Google Chrome Version 108.0.5355.0 (Official Build) dev (arm64) / Mozilla FireFox 105.0.3 (64-bit)

**Is there any additional information?**
No

dmceachernmsft:
Thanks for raising the issue @DercilioFontes! We have a `afterCreate` function that is an optional property to our `useAzureCommunicationChatAdapter` constructor for the adapter that is meant to be used to subscribe to these events.  this little snippet is a example of how to use this function:

` const errorHandler = (e: ChatAdapterError) => {
[...]

PR #2471, opened 2022-10-28
Chat adapter error update

dmceachernmsft:
# What

Fixes issue where adapter errors aren't being caught
# Why

Issue where `sendTypingIndicator` errors were not being caught by the adapter.

https://github.com/Azure/communication-ui-library/issues/2425
https://skype.visualstudio.com/SPOOL/_workitems/edit/3026772
# How Tested

Added new unit test to validate the error capture.


prprabhu-ms (approved):

packages/react-composites/package.json
```diff
@@ -20,7 +20,7 @@
     "build:e2e:call": "rushx _by-flavor \"webpack-cli build --config ./tests/app/call/webpack.config.js --mode=production --env production\"",
     "build:e2e:chat": "rushx _by-flavor \"webpack-cli build --config ./tests/app/chat/webpack.config.js --mode=production --env production\"",
     "build:e2e:callwithchat": "rushx _by-flavor \"webpack-cli build --config ./tests/app/callwithchat/webpack.config.js --mode=production --env production\"",
-    "test": "rushx _by-flavor \"rushx _test:by-flavor\"",
+    "test": "rush build -t . && rushx _by-flavor \"rushx _test:by-flavor\"",
```
PorterNan, line 23: I got the purpose of adding rush build -t here.
Somehow call rush build sounds like a parent level command

Imagine if you are doing a `rush test`, this `rush build -t` will run for every child repo which might cause some duplication, I guess we just need to give developers expectation that we don't do a full build for the test here

packages/chat-stateful-client/src/StatefulChatThreadClient.ts
```diff
@@ -109,6 +109,14 @@ class ProxyChatThreadClient implements ProxyHandler<ChatThreadClient> {
       case 'listReadReceipts': {
         return createDecoratedListReadReceipts(chatThreadClient, this._context);
       }
+      case 'sendTypingNotification': {
```
PorterNan, line 112: there is a unit test for statefulChatThreadClient:
StatefulCallClient.test.ts in the same folder

And there is a test set named 'errors should be reported correctly from Call when'

Would be nice to add this into the test set

PorterNan (approved):

packages/react-composites/package.json
prprabhu-ms: Hmm.. Actually, I read this wrong earlier. `rushx build` would have been an acceptable trade-off. Running `rush build -t` when developer ran `rushx test` is wrong -- the expectation for a `rushx` command is that it is packlet-local only.

prprabhu-ms: Let's remove this from this PR and figure out if we need better documentation / understand why the test failed without the `rush build -t`.

PR #2481, opened 2022-11-02
CallComposite getting role from Calling SDK

mgamis-msft:
# What

This PR changes the `role` from the CallComposite to `roleHint`. The `roleHint` will determine the role before the call because it will be passed to the PermissionProvider. 
When a Rooms call starts, the CallComposite will obtain the role from the CallingSDK and use it for the PermissionProvder 

# Why

# How Tested

# Process & policy checklist


JamesBurnside:
Pro the move to roleHint 👍
Best in a separate PR but want to check, is the plan to keep it as a Composite property or move to an adapter contructor property?

mgamis-msft:
> Pro the move to roleHint 👍 Best in a separate PR but want to check, is the plan to keep it as a Composite property or move to an adapter contructor property?

Yes, I put it in a separate PR here https://github.com/Azure/communication-ui-library/pull/2482 :)

JamesBurnside (approved):

dmceachernmsft (approved):

PR #2482, opened 2022-11-02
Move roleHint to CallAdapter

mgamis-msft:
# What

This PR removes the prop `roleHint` from CallComposite and adds it to `CallAdapterClientState `.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3030206

# How Tested

Tested on local calling sample

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -878,6 +906,39 @@ export const useAzureCommunicationCallAdapter = (
...
+): Promise<CallAdapter> => {
+  const deviceManager = (await callClient.getDeviceManager()) as StatefulDeviceManager;
+  return new AzureCommunicationCallAdapter(callClient, locator, callAgent, deviceManager);
+};
+
+let createAzureCommunicationCallAdapterFromClientToExport: (
+  callClient: StatefulCallClient,
+  callAgent: CallAgent,
+  locator: CallAdapterLocator,
+  /* @conditional-compile-remove(rooms) */ roleHint?: Role
```
JamesBurnside, line 922: in terms of scaling to more future options, do we want to start making this non-flat? e.g.
```ts
createAzureCommunicationCallAdapterFromClient: (
  callClient,
  callAgent,
  locator,
  options: {
    roleHint: Role
  }
) => Promise<CallAdapter>
```

The moment there's optional arg that's positional in the future this could become:
```ts
createAzureCommunicationCallAdapterFromClient(callClinet, callAgent, locator, undefined /*rolehint*/, undefined /*another option*/, thirdOption)
```

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -360,11 +352,12 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
...
     onFetchParticipantMenuItems,
     options,
-    formFactor = 'desktop',
-    /* @conditional-compile-remove(rooms) */
-    roleHint
+    formFactor = 'desktop'
   } = props;
 
+  /* @conditional-compile-remove(rooms) */
+  const roleHint = adapter.getState().roleHint;
```
JamesBurnside, line 359: Is there a future work item to move roleHint to selectors?

JamesBurnside: Oh this is getting direct from adapter state because it's being used in the permissions prompt not a component

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1505,10 +1506,10 @@ export interface ControlBarProps {
 }
 
 // @public
-export const createAzureCommunicationCallAdapter: ({ userId, displayName, credential, locator, alternateCallerId }: AzureCommunicationCallAdapterArgs) => Promise<CallAdapter>;
+export const createAzureCommunicationCallAdapter: ({ userId, displayName, credential, locator, alternateCallerId, roleHint }: AzureCommunicationCallAdapterArgs) => Promise<CallAdapter>;
 
 // @public
-export const createAzureCommunicationCallAdapterFromClient: (callClient: StatefulCallClient, callAgent: CallAgent, locator: CallAdapterLocator) => Promise<CallAdapter>;
+export const createAzureCommunicationCallAdapterFromClient: (callClient: StatefulCallClient, callAgent: CallAgent, locator: CallAdapterLocator, roleHint?: Role | undefined) => Promise<CallAdapter>;
```
prprabhu-ms, line 1512: The arguments to this factory function are getting out of hands...

Really it should have been:

`createAzureCommunicationCallAdapterFromClient(callClient, callAgent, options)` with the clients explicitly listed and then all options in an option bag.

But we already added the `locator` as a top-level arg here.

I'd recommend still introducing an options bag now and then taking it to ARB to see what they think: 
`createAzureCommunicationCallAdapterFromClient(callClient, callAgent, locator, options)`

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -72,7 +74,11 @@ class CallContext {
   private state: CallAdapterState;
   private callId: string | undefined;
 
-  constructor(clientState: CallClientState, isTeamsCall: boolean) {
+  constructor(
+    clientState: CallClientState,
+    isTeamsCall: boolean,
+    /* @conditional-compile-remove(rooms) */ roleHint?: undefined | /* @conditional-compile-remove(rooms) */ Role
```
prprabhu-ms, line 80: Isn't this enough? `/* @conditional-compile-remove(rooms) */ roleHint?: Role

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -195,15 +202,20 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
     callClient: StatefulCallClient,
     locator: CallAdapterLocator,
     callAgent: CallAgent,
-    deviceManager: StatefulDeviceManager
+    deviceManager: StatefulDeviceManager,
+    /* @conditional-compile-remove(rooms) */ roleHint?: undefined | /* @conditional-compile-remove(rooms) */ Role
```
prprabhu-ms, line 206: Same: `/* @conditional-compile-remove(rooms) roleHint?: Role`.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
prprabhu-ms: Heh... Left the same comment on the API diff.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -878,6 +906,39 @@ export const useAzureCommunicationCallAdapter = (
...
+) => Promise<CallAdapter> = createAzureCommunicationCallAdapterFromClientStable;
+
+/* @conditional-compile-remove(rooms) */
+/**
+ * Create a {@link CallAdapter} using the provided {@link StatefulCallClient} and a provided role
+ * to hint the user of their role specific for Rooms calls.
+ *
+ * @beta
+ */
+createAzureCommunicationCallAdapterFromClientToExport = async (
```
prprabhu-ms, line 932: I don't understand why you had to introduce `createAzureCommunicationCallAdapterFromClientToExport` in this way.

Can't you conditionally compile just the argument to `createAzureCommunicationCallAdapterFromClient` along with the implementation?

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -88,6 +90,12 @@ export type CallAdapterClientState = {
    * Azure communications Phone number to make PSTN calls with.
    */
   alternateCallerId?: string;
+  /* @conditional-compile-remove(rooms) */
+  /**
+   * Optional role of the local user that is assumed before a call is started. This role will determine permissions
```
prprabhu-ms, line 95: Please update this comment to very clearly indicate how `roleHint` is used. In particular, explain why this is named _hint_: That this is used only in the situation when role information is not available from the backend.

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -113,12 +115,7 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     return <Spinner label={'Creating adapter'} ariaLive="assertive" labelPosition="top" />;
   }
 
-  let callInvitationUrl: string | undefined = window.location.href;
-  /* @conditional-compile-remove(rooms) */
-  // If role is defined then the call is a Rooms call so we should not make call invitation link available
-  if (role) {
-    callInvitationUrl = undefined;
-  }
+  const callInvitationUrl: string | undefined = window.location.href;
```
prprabhu-ms, line 118: This seems to be changing business logic of when to show `callInvitationUrl`.

prprabhu-ms (commented):
Very nicely scoped PR. Thanks!

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
mgamis-msft: Yeah, you and James both have a good point :). Changed.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
mgamis-msft: No it was not enough. Functions with beta type arguments still need the beta type reference in stable for some reason. So I had to hack it with an `undefined` union to get cond-comp to work. But it is a moot point, after moving roleHint to an object type argument.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
mgamis-msft: This is a hack I had to do because if I cinditionally compile the just the function argument like so:
image
The stable api changes includes the options argument like so:
image

JamesBurnside: hmm, @prprabhu-ms @PorterNan @anjulgarg is there any change we can make to the babel config to support this?

JamesBurnside (approved):

samples/Calling/src/app/views/CallScreen.tsx
mgamis-msft: My bad. Reverted.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -465,7 +469,6 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
     callInvitationUrl?: string;
     formFactor?: 'desktop' | 'mobile';
```
PorterNan, line 470: Interesting question comes to my mind - should we move  form factor into options once we are allowed to make breaking change? @prprabhu-ms @JamesBurnside

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: Hmm, what it would be like in preprocessed file? Wondering why the hack would work

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -879,23 +907,52 @@ export const useAzureCommunicationCallAdapter = (
...
- * @public
+ * @beta
  */
-export const createAzureCommunicationCallAdapterFromClient = async (
+createAzureCommunicationCallAdapterFromClientToExport = async (
   callClient: StatefulCallClient,
   callAgent: CallAgent,
-  locator: CallAdapterLocator
+  locator: CallAdapterLocator,
+  /* @conditional-compile-remove(rooms) */ options?: { /* @conditional-compile-remove(rooms) */ roleHint?: Role }
```
PorterNan, line 940: We might need a really good jsdoc here to explain how roleHint works, otherwise it might be confusing lots of developers

PorterNan:
Lets see if we could find a better solution of condtional compilation, other than that I am more than happy to approve this

packages/communication-react/review/beta/communication-react.api.md
mgamis-msft: Yeah I agree we can move form factor into the CallComposite options object prop. But once we are allowed to make breaking changes and for another PR

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
mgamis-msft: It looks much better now without the let variable reassignment 👍

prprabhu-ms (approved):
The eventual conditionally compiled code is correct, so let's unblock this PR to make progress.

PR #2485, opened 2022-11-03
Update domain permissions to support individual camera and microphone modals

JamesBurnside:
# What
* Create a base domainpermissions component that each individual permissions component use
* Create three domainpermissions components: CameraAndMicrophone, Camera, Microphone
* Repurpose the Try Again button with an optional Continue Anyway button
  * This will be a little strange on mobile until the prompt behavior is changed in an upcoming PR (but call readiness is opted-out currently so not breaking).
* Update storybook to have appropriate preview args:
   image

# Why
New call readiness designs

# How Tested
Just in storybook, the prompt logic is changing and will be a separate PR

Before:
| DomainPermission Dialog (Before change) |
| -- |
| image |

After:

| CameraAndMicrophone (After changes) | Camera (After changes) | Microphone (After changes) |
| -- | -- | -- |
| image | image | image |

PR #2486, opened 2022-11-03
Add permissions denied SKU to DomainPermissions

JamesBurnside:
# What
* Add Denied type with corresponding strings and icons

Note: We have BrowserPermissions dialogs that are similar, convergence between these will happen in a future PR. This PR just adds the new SKU to the domain permissions modal.

# Why
Newer call readiness designs, continuation of #2485 

# How Tested
Added SKU option to storybook:
image

| CameraAndMicrophone | Camera | Microphone |
| -- | -- | -- |
| image | image | image |


carocao-msft (approved):

packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx
```diff
@@ -19,6 +19,10 @@ export interface CommonDomainPermissionsProps {
    * Name of application calling experience is in.
    */
   appName: string;
+  /**
+   * Type of the Domain Permissions component.
+   */
+  type: 'request' | 'denied';
```
edwardlee-msft, line 25: nit: Does this need to be a required prop? Will we not provide a default value like setting to "request" by default?

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -205,6 +205,24 @@
     "primaryButtonText": "Continue without microphone",
     "ariaLabel": "Allow microphone access"
   },
+  "CameraAndMicrophoneDomainPermissionsDenied": {
+    "primaryText": "Unable to access camera and microphone",
+    "secondaryText": "Click the lock icon in the address bar to grant microphone permissions to this webpage. A page refresh may be required.",
+    "primaryButtonText": "Continue Anyway",
```
edwardlee-msft, line 211: nit: "Continue Anyway" => "Continue anyway"

packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx
JamesBurnside: We could, and open to, but I want to be explicit first until we have a strong argument to provide a default

packages/react-components/src/localization/locales/en-US/strings.json
JamesBurnside: Oh good catch, this should actually be `Continue without x/y` - copy paste mistake!

dmceachernmsft (approved):

packages/react-components/src/localization/locales/en-US/strings.json
JamesBurnside: updated

packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx
JamesBurnside: Not having one might suggest this should be two seperate component API instead:
`CameraDomainPermissionsRequest` and `CameraDomainPermissionsDenied` (that under the hood just use this).
Because this is in beta and we can adjust we'll see how we use this and get an API review then can adjust accordingly

packages/react-components/src/theming/icons.tsx
```diff
@@ -119,6 +119,36 @@ const DomainPermissionCamera20Filled = (): JSX.Element => {
   );
 };
 
+/* @conditional-compile-remove(call-readiness) */
+const DomainPermissionsMicDenied20Filled = (): JSX.Element => {
+  const theme = useTheme();
+  return (
+    <Stack
+      horizontalAlign={'center'}
+      styles={{ root: { borderRadius: '100%', background: theme.palette.themeLighterAlt, padding: '2rem' } }}
```
edwardlee-msft, line 128: nit: this particular style is used 4 times. Maybe a good idea to create a common variable?

edwardlee-msft (approved):

packages/react-components/src/theming/icons.tsx
JamesBurnside: Great callout, made a `icons.styles.ts` file and move common styles to there

PR #2488, opened 2022-11-04
Add support for conditional compilation of browser tests

prprabhu-ms:
After this PR, browser tests can also be conditionally compiled, similar to the packlets.

This PR conditionally compiles a single `describe` block as an example. Follow up PRs will replace all uses of `isTestProfileStableFlavor()` with conditional compilation and deprecate that helper function.

The benefit of this PR is that we will not need to separately remember to enable browser tests when stabilizing a feature.


packages/react-composites/scripts/runBrowserTests.mjs
```diff
@@ -128,13 +150,48 @@ async function runOne(args, composite, hermeticity) {
...
   }
 }
 
 function setup() {
   console.log('Cleaning up output directory...');
-  rmdirSync(BASE_OUTPUT_DIR, { recursive: true });
+  fs.rmSync(BASE_OUTPUT_DIR, { recursive: true, force: true });
+}
+
+async function preprocessTests() {
```
PorterNan, line 162: Nice, so we are running preprocess as a function in this mjs file

PorterNan (approved):

edwardlee-msft (approved):

PR #2490, opened 2022-11-04
Migrate most tests away from runtime check of build flavor

prprabhu-ms:
This PR migrates most browser tests off of runtime checks for build flavor (via `isTestProfileStableFlavor`) to instead use conditional compilation.

The benefit is that the tests will not have to be separately modified when the relevant feature is stabilized.


common/config/env/.env
```diff
@@ -1,2 +1,2 @@
 // THIS IS AN AUTO GENERATED FILE BY `RUSH SWITCH <BETA|STABLE>`. DO NOT MODIFY
-COMMUNICATION_REACT_FLAVOR=beta
```
prprabhu-ms, line 2: Err.... NO.

packages/react-composites/tests/browser/call/hermetic/HoldScreen.test.ts
```diff
@@ -3,13 +3,12 @@
 
 import { expect } from '@playwright/test';
 import { IDS } from '../../common/constants';
-import { dataUiId, isTestProfileStableFlavor, pageClick, stableScreenshot, waitForSelector } from '../../common/utils';
+import { dataUiId, pageClick, stableScreenshot, waitForSelector } from '../../common/utils';
 import { buildUrlWithMockAdapter, defaultMockCallAdapterState, defaultMockRemoteParticipant, test } from './fixture';
 
+/* @conditional-compile-remove(PSTN-calls) @conditional-compile-remove(one-to-n-calling) */
```
dmceachernmsft, line 9: I thought when denoting 2 features in a CC statement it was something like `@conditional-compile-remove(feature 1 feature 2)` when in the same comment?

packages/react-composites/tests/browser/callwithchat/live/CallWithChatComposite.test.ts
```diff
@@ -54,17 +53,17 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
...
-    test.skip(isTestProfileStableFlavor());
     const page = pages[1];
     await pageClick(page, dataUiId('call-with-chat-composite-more-button'));
     const moreButtonShowDialpadButton = await page.$('div[role="menu"] >> text="Show dialpad"');
     await moreButtonShowDialpadButton?.click();
     expect(await stableScreenshot(page)).toMatchSnapshot(`call-with-chat-more-drawer-dtmf-dialpad.png`);
   });
 
+  /* @conditional-compile-remove(PSTN-calls) */
   test('More Drawer menu opens and can choose to be on hold', async ({ pages }) => {
```
dmceachernmsft, line 66: this test is also for 1-n-calling

dmceachernmsft (approved):

common/config/env/.env
JamesBurnside: out of interest, this should be gitignored:
image

anjulgarg (approved):

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -7,16 +7,15 @@ import {
...
   isTestProfileDesktop,
-  isTestProfileStableFlavor,
   pageClick,
   stableScreenshot,
   waitForSelector
 } from '../../common/utils';
 import { buildUrlWithMockAdapter, defaultMockCallAdapterState, defaultMockRemoteParticipant, test } from './fixture';
 
 test.describe('Participant pane tests', async () => {
+  /* @conditional-compile-remove(one-to-n-calling) */
```
edwardlee-msft, line 17: People Pane should actually be conditionally compiled for both PSTN and one-to-n-calling.

edwardlee-msft: I will go back and add the missing conditional compilation for PeoplePane related code. But in the meantime please also add PSTN here :)

edwardlee-msft: https://github.com/Azure/communication-ui-library/pull/2493 for tracking

packages/react-composites/tests/browser/chat/hermetic/FileSharing.test.ts
```diff
@@ -2,12 +2,11 @@
 // Licensed under the MIT license.
 import { expect } from '@playwright/test';
 import { sendMessage, waitForMessageDelivered } from '../../common/chatTestHelpers';
-import { dataUiId, isTestProfileStableFlavor, stableScreenshot, waitForSelector } from '../../common/utils';
+import { dataUiId, stableScreenshot, waitForSelector } from '../../common/utils';
 import { buildUrlForChatAppUsingFakeAdapter, DEFAULT_FAKE_CHAT_ADAPTER_ARGS, test, TEST_PARTICIPANTS } from './fixture';
 
+/* @conditional-compile-remove(file-sharing) */
```
PorterNan, line 8: How does this work? Do we add these test.ts into conditional compilation?

PorterNan: Just checked the config file - we do make unittest works for conditional compilation,  but not for e2e test

PorterNan (approved):

edwardlee-msft (approved):

packages/react-composites/tests/browser/call/hermetic/HoldScreen.test.ts
prprabhu-ms: Nah, we don't support fancy expressions inside the directive. But we do support multiple directives in the same / different comment.

See https://github.com/Azure/communication-ui-library/blob/b7666daf66fdb29053a89ee516a839ece5a7d9b1/packages/acs-ui-common/src/conditional-compilation-sample/index.tsx#L233

PR #2479, opened 2022-11-02
[Call Readiness] Add getEnvironmentInfo check to CallAdapter and Stateful Client

dmceachernmsft:
# What

Add handlers and usage of new getEnvironmentInfo API from calling SDK to Stateful Client and CallAdapter
# Why

Allows the CallComposite to use its new unsupported browser page

https://skype.visualstudio.com/SPOOL/_workitems/edit/2981221
https://skype.visualstudio.com/SPOOL/_workitems/edit/3009502
https://skype.visualstudio.com/SPOOL/_workitems/edit/2980686
# How Tested

https://user-images.githubusercontent.com/94866715/199573013-f034d114-c79a-4b6c-b161-2c506e87328d.mp4

Tested on opera browser in browserStack to verify new page responds to new logic.

PR #2491, opened 2022-11-04
[Call Readiness] add optin feature for unsupported browser.

dmceachernmsft:
# What

adds a "opt in" feature for unsupported browser and `getEnvironmentInfo` handler to `CallAdapter`
# Why

allows for Contoso to opt in or out of the unsupported browser feature.

unblocks this PR: https://github.com/Azure/communication-ui-library/pull/2479
# How Tested

Tested locally on browser stack to verify no flash and as well ran live tests to verify they aren't blocked by feature.


edwardlee-msft (approved):

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -149,6 +149,8 @@ export interface CallAdapterCallManagement {
     createStreamView(remoteUserId?: string, options?: VideoStreamOptions): Promise<void | CreateVideoStreamViewResult>;
     disposeStreamView(remoteUserId?: string, options?: VideoStreamOptions): Promise<void>;
     // @beta
+    getEnvironmentInfo(): Promise<EnvironmentInfo>;
```
JamesBurnside, line 152: this might have to be optional to prevent a breaking change to this interface

JamesBurnside: Although looking at other new ones we added like `sendDtmfTone` these may also need to be optional

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -467,6 +471,13 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
     });
   }
 
+  /* @conditional-compile-remove(unsupported-browser) */
+  public async getEnvironmentInfo(): Promise<EnvironmentInfo> {
+    return await this.asyncTeeErrorToEventEmitter(async () => {
+      return this.callClient.feature(Features.DebugInfo).getEnvironmentInfo();
```
JamesBurnside, line 477: this is not being awaited, either:
```suggestion
      return this.callClient.feature(Features.DebugInfo).getEnvironmentInfo;
```
or
```suggestion
      return await this.callClient.feature(Features.DebugInfo).getEnvironmentInfo();
```

packages/react-composites/review/beta/react-composites.api.md
JamesBurnside, line 152: Is there a reason to only expose `getEnvironmentInfo` and not the `DebugInfo` interface?

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -83,7 +83,8 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     () => ({
       callReadinessOptedIn: callReadinessOptedIn,
       onPermissionsTroubleshootingClick,
-      onNetworkingTroubleShootingClick
+      onNetworkingTroubleShootingClick,
+      unsupportedBrowserOptedIn: false
```
JamesBurnside, line 87: why not opt in on calling sample also?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -900,10 +911,6 @@ export const createAzureCommunicationCallAdapterFromClient = async (
   locator: CallAdapterLocator
 ): Promise<CallAdapter> => {
   const deviceManager = (await callClient.getDeviceManager()) as StatefulDeviceManager;
-  /* @conditional-compile-remove(unsupported-browser) */
-  const feature = callClient.feature(Features.DebugInfo);
-  /* @conditional-compile-remove(unsupported-browser) */
-  await feature.getEnvironmentInfo();
```
JamesBurnside, line 906: Question that could be a bad idea: should we do this anyway to prime the call in the call composite?
Concern is that this is asynchronous:
image
meaning we could flash the configuration page if its not primed

packages/react-composites/review/beta/react-composites.api.md
JamesBurnside: Added making these required again to the major bump list: https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/31318/What-we-would-like-to-break-when-we-do-a-major-bump
But for now I'd recommend going through newly added adapter interfaces and making them optional

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
dmceachernmsft: Problem is if this is primed the adapter will set the page to `unsupportedEnvironement` and when the featrue is opted out you land on a white screen. I don't see value in adding a `clearEnvironmentInfo()` handler to use here, do you think that we should?

JamesBurnside: Hmm, this is a problem. If someone calls getEnvironmentInfo() it will have a sudden side effect that the whole composite will go blank then

samples/Calling/src/app/views/CallScreen.tsx
JamesBurnside: Call readiness if off because the feature isn't complete and its a bad join experience right now, but I think this should be on by default -- or another way to ask the question; when would be the right time to turn it on?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
dmceachernmsft: I will look into making sure the composite is paying attention. I think if Contoso uses this function, they are in a sense opting into the feature. I will make sure the comment for this on the interface acknowledges this.

samples/Calling/src/app/views/CallScreen.tsx
dmceachernmsft: I will update this to reflect that we want it on now.

packages/react-composites/review/beta/react-composites.api.md
dmceachernmsft: from my understanding to use the `DebugInfo` interface you need to use the `feature()` function on the call client which we proxied to get this API. If Contoso wanted the `DebugInfo` interface they could get it off the `StatefulClient` and then create the adapter from there. everything else on the interface didn't seem relevant to the feature so I didn't add it. do you think we should?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
JamesBurnside: > if Contoso uses this function, they are in a sense opting into the feature

If they have explicitly opted out somewhere in our API (in the composite interface), calling a function should never "opt them back in", particularly when the names don't align (nothing in name of `adapter.getEnvironmentInfo` suggest it'll override my opt-out property on a different interface)

packages/react-composites/review/beta/react-composites.api.md
JamesBurnside, line 152: Anything with a `get` should really be idempotent (i.e. have no side effects), because of our stateful design it populates state ... I wonder (at the stateful layer) if we shouldn't be calling this `getEnvironmentInfo` and instead something like populateEnvironmentInfo or fetchEnvironmentInfo to indicate there will be a side effect whereby the state gets populated (and thus further changes could propagate)

JamesBurnside: cc @PorterNan @prprabhu-ms

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -154,6 +155,7 @@ export type AzureCommunicationCallWithChatAdapterArgs = {
     credential: CommunicationTokenCredential;
     locator: CallAndChatLocator | TeamsMeetingLinkLocator;
     alternateCallerId?: string;
+    callingFeatures?: CallAdapterOptionalFeatures;
```
JamesBurnside, line 158: nit: I would keep this consistent and make this features, when it does include chat features the type can be updated to `CallWithChatAdapterOptionalFeatures` and still be a non-breaking change

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -311,15 +311,21 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
...
-        </>
-      );
+  }
+
+  /* @conditional-compile-remove(unsupported-browser) */
+  if (
+    adapter.getState().features?.unsupportedEnvironment &&
+    adapter.getState().environmentInfo?.isSupportedBrowser === false
+  ) {
+    pageElement = (
```
JamesBurnside, line 321: if this is a page in of itself, should we be using adapter state page to drive this?

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatBackedCallAdapter.ts
```diff
@@ -149,6 +151,13 @@ export class CallWithChatBackedCallAdapter implements CallAdapter {
       return this.callWithChatAdapter.addParticipant(participant as CommunicationUserIdentifier);
     }
   }
+  /* @conditional-compile-remove(unsupported-browser) */
+  public async getEnvironmentInfo(): Promise<EnvironmentInfo> {
```
JamesBurnside, line 155: getEnvironmentInfo -> populateEnvironmentInfo

JamesBurnside: and in `throw new Error('getEnvironmentInfo not implemented');`

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
JamesBurnside: if so, we should check that `onCallEnded` fires correctly

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -51,6 +53,7 @@ import {
   DiagnosticChangedEventListner,
   CallAdapterCallEndedEvent
 } from './CallAdapter';
+import { CallAdapterOptionalFeatures } from './CallAdapter';
```
edwardlee-msft, line 56: Should this be CC? 
CallAdapterOptionalFeatures type is a @beta feature.

edwardlee-msft: After speaking with Don privately, he clarified that as long as the Stable API is not impacted it is fine not to CC internally.

packages/react-composites/src/composites/CallComposite/index.ts
```diff
@@ -14,5 +14,41 @@ export type {
   CustomCallControlButtonProps,
   CustomControlButtonProps
 } from './types/CallControlOptions';
-export * from './adapter';
```
edwardlee-msft, line 17: This must have been painful. Was there an automated way to do this? Or did you have to manually export each one individually?

edwardlee-msft (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -310,15 +310,22 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
...
-          {
-            /* @conditional-compile-remove(unsupported-browser) */
-            <UnsupportedBrowserPage onTroubleshootingClick={props.options?.onEnvironmentInfoTroubleshootingClick} />
-          }
-        </>
-      );
+  }
+
+  /* @conditional-compile-remove(unsupported-browser) */
+  if (adapter.getState().features?.unsupportedEnvironment) {
```
JamesBurnside, line 316: this conditional check shouldn't be necessary, the adapter shouldn't provide a page `unsupportedEnvironment` if the feature isn't opted into

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -372,6 +379,11 @@ export const CallComposite = (props: CallCompositeProps): JSX.Element => {
 
   const mobileView = formFactor === 'mobile';
 
+  /* @conditional-compile-remove(unsupported-browser) */
+  if (adapter.getState().features?.unsupportedEnvironment && adapter.populateEnvironmentInfo) {
+    adapter.populateEnvironmentInfo();
```
JamesBurnside, line 384: this is done in the adapter ctor, is this needed here?

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -419,13 +431,17 @@ const holdPageTrampoline = (): string => {
...
-  return 'call';
-};
-
-const getQueryOptions = (options: { /* @conditional-compile-remove(rooms) */ role?: Role }): PermissionConstraints => {
+const getQueryOptions = (options: {
+  /* @conditional-compile-remove(rooms) */ role?: Role;
+  /* @conditional-compile-remove(call-readiness) */ callReadinessOptedIn?: boolean;
+}): PermissionConstraints => {
+  /* @conditional-compile-remove(call-readiness) */
+  if (options.callReadinessOptedIn) {
```
JamesBurnside, line 439: this looks unintentional, was this meant to be added in or artefact after merge from main?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -479,6 +490,16 @@ export class AzureCommunicationCallAdapter implements CallAdapter {
     });
   }
 
+  /* @conditional-compile-remove(unsupported-browser) */
+  public async populateEnvironmentInfo(): Promise<EnvironmentInfo> {
```
JamesBurnside, line 494: actually given we call this in the adapter ctor.. can we make this private?

JamesBurnside: (or perhaps not private, but remove from the CallAdapter interface and just keep on the AzureCommunicationCallAdapter)

JamesBurnside (changes requested):
Seeing if we can reduce the public interface by not adding populateEnvironmentInfo to the CallAdapter

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
dmceachernmsft: this was a artifact from main in a conflict resolution

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
dmceachernmsft: if it isn't on the interface we can't publicly export the handler from the AzureCommunicationsCallWithChatAdapter since it is viewing the adapter as a implemented CallAdapter. should we just remove the function entirely? this would stop Contoso from being able to ask for the adapter to get this data. (even though we ask for it in the constructor)

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -137,10 +140,13 @@ class CallContext {
 
     // As the state is transitioning to a new state, trigger appropriate callback events.
     const oldPage = this.state.page;
+    /* @conditional-compile-remove(unsupported-browser) */
+    const environmentInfo = this.state.features?.unsupportedEnvironment ? this.state.environmentInfo : undefined;
```
JamesBurnside, line 144: this check is unnecessary as the getCallCompositePage does this check (either keep here and remove from getCallCompositePage or remove from here)

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1591,15 +1599,15 @@ export interface ControlBarProps {
 }
 
 // @public
-export const createAzureCommunicationCallAdapter: ({ userId, displayName, credential, locator, alternateCallerId, options }: AzureCommunicationCallAdapterArgs) => Promise<CallAdapter>;
+export const createAzureCommunicationCallAdapter: ({ userId, displayName, credential, locator, alternateCallerId, features, options }: AzureCommunicationCallAdapterArgs) => Promise<CallAdapter>;
 
 // @public
-export const createAzureCommunicationCallAdapterFromClient: (callClient: StatefulCallClient, callAgent: CallAgent, locator: CallAdapterLocator, options?: {
+export const createAzureCommunicationCallAdapterFromClient: (callClient: StatefulCallClient, callAgent: CallAgent, locator: CallAdapterLocator, features?: CallAdapterOptionalFeatures, options?: {
```
JamesBurnside, line 1605: avoid flat optional arguments (only permit one at the end), instead `features` should be inside `options`

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -214,6 +214,8 @@ type CallWithChatScreenProps = {
   devicePermissions?: DevicePermissionRestrictions;
   /* @conditional-compile-remove(call-readiness) */
   callReadinessOptedIn?: boolean;
+  /* @conditional-compile-remove(unsupported-browser) */
+  unsupportedBrowserOptedIn?: boolean;
```
JamesBurnside, line 218: can this be remove now?

samples/CallWithChat/src/app/views/CallScreen.tsx
```diff
@@ -111,7 +112,6 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
       rtl={currentRtl}
       joinInvitationURL={window.location.href}
       formFactor={isMobileSession ? 'mobile' : 'desktop'}
-      /* @conditional-compile-remove(call-readiness) */
```
JamesBurnside, line 114: should this have been removed?

dmceachernmsft: this should be here for the callReadiness opt in work Caroline did I believe

samples/CallWithChat/src/app/views/CallScreen.tsx
```diff
@@ -79,14 +79,19 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     [callIdRef]
   );
 
+  const callAdapterOptions = useMemo(() => {
```
JamesBurnside, line 82: conditional compile this?

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -86,7 +86,10 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
   );
 
   /* @conditional-compile-remove(rooms) */
```
JamesBurnside, line 88: update conditional compile here ton include browser suppotred check

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -141,16 +144,20 @@ type GetCallCompositePageFunction = ((
 export const getCallCompositePage: GetCallCompositePageFunction = (
   call,
   previousCall,
-  environmentInfo?
+  environmentInfo?,
+  features?
 ): CallCompositePage => {
+  console.log(features?.unsupportedEnvironment);
```
JamesBurnside, line 150: remove console log

samples/CallWithChat/src/app/views/CallScreen.tsx
JamesBurnside: (can leave if it doesn't cause a build failure)

dmceachernmsft: added the cc comment. 😊

JamesBurnside (approved):

PR #2497, opened 2022-11-06
Remove a rogue raw reference to env var

prprabhu-ms:
Missed in my original PR #2490


prabhjot-msft (approved):

packages/react-composites/scripts/runBrowserTests.mjs
```diff
@@ -128,13 +150,48 @@ async function runOne(args, composite, hermeticity) {
...
+  fs.writeFileSync(path.join(PREPROCESSED_ROOT, 'tsconfig.json'), '{ "extends": "../tsconfig.preprocess.json" }');
+}
+
+async function preprocessDir(fromDir, toDir) {
+  console.log(`Preprocessing ${fromDir} to ${toDir}`);
+  fs.mkdirsSync(toDir);
+  fs.copySync(fromDir, toDir, { overwrite: true, preserveTimestamps: true });
+  await exec(
+    quote([
+      'npx',
```
PorterNan, line 180: Maybe we can execute command here from package.json file like `npm run preprocess: test` or so (with 2nd parameter)? Good to have a centralized place to store all the commands (I can feel some other reasons we put command here as well, so highly depends on you)

packages/react-composites/tests/browser/call/hermetic/CallReadiness.test.ts
```diff
@@ -6,21 +6,19 @@ import { expect, Page } from '@playwright/test';
...
   isTestProfileMobile,
-  isTestProfileStableFlavor,
   pageClick,
   stableScreenshot,
   waitForPageFontsLoaded,
   waitForSelector
 } from '../../common/utils';
 import type { MockCallAdapterState } from '../../../common';
 
+/* @conditional-compile-remove(call-readiness) */
```
PorterNan, line 16: Nice! How do you figure out all these feature tags? I can imagine it requires lots of code reading (But we don't have to do it any more after your PR!)

PorterNan (approved):

packages/react-composites/scripts/runBrowserTests.mjs
prprabhu-ms: The intention is to keep moving logic from `package.json` `scripts` object to actual node scripts.

We can extract common logic as we use it.

packages/react-composites/tests/browser/call/hermetic/CallReadiness.test.ts
prprabhu-ms: Yeah, just my best guess followed by asking the feature owners.

PR #2498, opened 2022-11-06
Refactor `PeoplePane.test.tsx` and drop dependence on `COMMUNICATION_REACT_FLAVOR` envvar

prprabhu-ms:
The tests were very hard to read due to the difference branches within the test. This PR splits the tests into 3 suites for the three distinct UX's being tested - (1) flyout, (2) side pane (3) full screen with drawer.

It also replaces the use of `COMMUNICATION_REACT_FLAVOR` runtime check with the relevant conditional compilation directive.

There are some tangential changes to the UI snapshots because this PR unifies the fake remote participants used by the tests (the differences were not significant to the tests) to reduce boilerplate.


packages/react-composites/tests/browser/call/hermetic/PeoplePane.test.ts
```diff
@@ -9,185 +9,209 @@ import {
...
-      await pageClick(page, dataUiId(IDS.participantButtonPeopleMenuItem));
-      await page.hover(dataUiId('participant-item') + ' >> nth=0');
-      // There should be at least one participant. Just click on the first.
-      await pageClick(page, dataUiId(IDS.participantItemMenuButton) + ' >> nth=0');
-
-      const injectedMenuItem = await waitForSelector(page, dataUiId('test-app-participant-menu-item'));
-      await injectedMenuItem.waitForElementState('stable', { timeout: perStepLocalTimeout() });
-    }
+    await pageClick(page, dataUiId('call-with-chat-composite-more-button'));
+    const drawerPeopleMenuDiv = await page.$('div[role="menu"] >> text=People');
```
dmceachernmsft, line 195: Would there be any value in making a DataUIId for this button?

prprabhu-ms: I'll refrain from adding that in this PR.

I think there would be value in adding a helper function for this. Raw `$()` expressions are hard to read in a test.

packages/react-composites/tests/browser/call/hermetic/PeoplePane.test.ts
```diff
@@ -9,185 +9,209 @@ import {
...
+const participantListInitialState = (): MockCallAdapterState => {
+  const paul = defaultMockRemoteParticipant('Paul Bridges');
+  addVideoStream(paul, true);
+  paul.isSpeaking = true;
+  const initialState = defaultMockCallAdapterState([
+    paul,
+    defaultMockRemoteParticipant('Eryka Klein'),
+    defaultMockRemoteParticipant('Fiona Harper')
+  ]);
+  addDefaultMockLocalVideoStreamState(initialState);
```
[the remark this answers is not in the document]
prprabhu-ms: We do: 
image

The green boxes are the videos (with a moving bus hidden behind the green mask ;) )

dmceachernmsft (approved):

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -7,16 +7,15 @@ import {
...
   isTestProfileDesktop,
-  isTestProfileStableFlavor,
   pageClick,
   stableScreenshot,
   waitForSelector
 } from '../../common/utils';
 import { buildUrlWithMockAdapter, defaultMockCallAdapterState, defaultMockRemoteParticipant, test } from './fixture';
 
 test.describe('Participant pane tests', async () => {
+  /* @conditional-compile-remove(one-to-n-calling) */
```
edwardlee-msft, line 17: nit: Possible for another PR, but this should include /* @conditional-compile-remove(PSTN-calls) */ as well.

edwardlee-msft (approved):

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
prprabhu-ms: Oh yeah, sorry I should have rebased before tagging you. It's fixed in `main` from an earlier PR.

PR #2502, opened 2022-11-08
[A11y bug] participant joined/left announcement

dmceachernmsft:
# What

Introduces Narrator strings and the Announcer to the MediaGallery.
# Why

Allows for Narrator users to know when participants have left and joined the call through announcements.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3007443
# How Tested

Validated on windows and android that the announcements happen at the correct time.


packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -60,6 +67,32 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
     };
   }, [cameraSwitcherCallback, cameraSwitcherCameras]);
 
+  useMemo(() => {
+    const onPersonJoined = (e) => {
+      setAnnouncerString(locale.participantJoinedNoticeString);
+      setInterval(() => {
```
JamesBurnside, line 73: leave a comment around why this setInterval is needed

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -50,6 +53,10 @@ export interface MediaGalleryProps {
  * @private
  */
 export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
+  const adapter = useAdapter();
+  const locale = useLocale().strings.call;
+  const [announcerString, setAnnouncerString] = useState<string>('');
```
JamesBurnside, line 58: is there a reason to use an empty string here and not `undefined` or `null`?

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -60,6 +67,32 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
+        setAnnouncerString('');
+      }, 5000);
+    };
+    adapter.on('participantsJoined', onPersonJoined);
+    return () => {
+      adapter.off('participantsJoined', onPersonJoined);
+    };
+  }, []);
+
+  useMemo(() => {
```
JamesBurnside, line 83: these add a lot of lines to the render function, consider a creating a helper hook that just returns the string
```
const announcerString = useParticipantJoinedAnnouncement();
```

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -60,6 +67,32 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
+    const onPersonLeft = (e) => {
+      setAnnouncerString(locale.participantLeftNoticeString);
+      setInterval(() => {
+        setAnnouncerString('');
```
JamesBurnside, line 87: what if a participant joins and another one leaves?

dmceachernmsft: The announcer should just read the first string. this is something we should test on the platforms to understand rapid string changes in that component.

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -68,7 +68,9 @@
     "holdScreenLabel": "You're on hold",
     "openDtmfDialpadLabel": "Show dialpad",
     "dtmfDialpadPlaceholderText": "Enter number",
-    "outboundCallingNoticeString": "Calling..."
+    "outboundCallingNoticeString": "Calling...",
+    "participantJoinedNoticeString": "Participant joined",
```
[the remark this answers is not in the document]
JamesBurnside: This is definitely something we should ask about, saying "participant joined" feels like there's so little information its not worth it?
In the bug they mention "Dev has joined the call" which indicates something less generic:
image

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
JamesBurnside: I should rephrase that, what if a user leaves, then 4999ms later a participant joins (i.e. it sets setAnnouncerString('new participant') then immediately calls setAnnouncerString(''))

packages/react-composites/src/composites/localization/locales/en-US/strings.json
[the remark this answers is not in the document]
edwardlee-msft: Let's talk with Aman about this tomorrow to see if we can keep it generic or if it will need to contain specifics.

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -118,3 +126,57 @@ export const useLocalVideoStartTrigger = (isLocalVideoAvailable: boolean, should
...
+     * Announcer component should the string change.
+     */
+    if (timeoutState) {
+      clearTimeout(timeoutState);
+      setTimeoutState(undefined);
+    }
+    setTimeoutState(
+      setTimeout(() => {
+        setAnnouncerString(string);
+        setTimeoutState(undefined);
```
edwardlee-msft, line 159: Pretty nifty!

dmceachernmsft: @prprabhu-ms is responsible for this wizardry! 😊

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -76,18 +81,21 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
-        localVideoViewOptions={localVideoViewOptions}
-        remoteVideoViewOptions={remoteVideoViewOptions}
-        styles={VideoGalleryStyles}
-        layout="floatingLocalVideo"
-        showCameraSwitcherInLocalPreview={props.isMobile}
-        localVideoCameraCycleButtonProps={cameraSwitcherProps}
-        onRenderAvatar={onRenderAvatar}
-      />
+      <>
+        <Announcer announcementString={announcerString} ariaLive={'assertive'} />
```
JamesBurnside, line 85:
```suggestion
        <Announcer announcementString={announcerString} ariaLive={'polite'} />
```
?

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -118,3 +126,58 @@ export const useLocalVideoStartTrigger = (isLocalVideoAvailable: boolean, should
...
+    setTimeoutState(
+      /**
+       * These set timeouts are needed to clear the announcer string in case we have multiple
+       * participants join. Since the narrator will only announce the string in the
+       * Announcer component should the string change.
+       */
+      setTimeout(() => {
+        setAnnouncerString(string);
+        setTimeoutState(undefined);
+      }, 500)
```
JamesBurnside, line 161: no arbitrary constants nit, pull this out as a constant and give a name and a comment for its purpose (you can probably move the comment from the setTimeout as it becomes redundant if the constant is named and commented), e.g.
```tsx
// Delay before telling a screen reader to announce a participant joined/left event.
// The delay is needed ...
const PARTICIPANT_ANNOUNCEMENT_DELAY = 500
```

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -118,3 +126,58 @@ export const useLocalVideoStartTrigger = (isLocalVideoAvailable: boolean, should
...
+ * used by the system narrator.
+ *
+ * @returns string to be used by the narrator and Announcer component
+ */
+const useParticipantChangedAnnouncement = (): string => {
+  const adapter = useAdapter();
+  const locale = useLocale().strings.call;
+  const [announcerString, setAnnouncerString] = useState<string>('');
+  /**
+   * state to track whether there is currently a timer set in the MediaGallery
```
JamesBurnside, line 141: nit: this checks if there is a timer set _in this hook_, not in the media gallery. The media gallery could have other timers unrealted to this hook

JamesBurnside: ideally remove mention of the MediaGallery and move this to its own util file, it'll likely only be used here as we don't want this announcement happening all over the place but the implementation detail is not related to the MediaGallery so not necessary to bloat the file

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -118,3 +126,58 @@ export const useLocalVideoStartTrigger = (isLocalVideoAvailable: boolean, should
...
+  };
+
+  useEffect(() => {
+    const onPersonJoined = (): void => {
+      setParticipantEventString(locale.participantJoinedNoticeString);
+    };
+    adapter.on('participantsJoined', onPersonJoined);
+
+    const onPersonLeft = (): void => {
+      setParticipantEventString(locale.participantLeftNoticeString);
```
JamesBurnside, line 172: we should confirm expected behavior with the a11y team on what happens if someone leaves and joins, from this implementation if someone joins then before 500ms elapses someone leaves - we won't announce that someone had joined

packages/react-composites/src/composites/localization/locales/en-US/strings.json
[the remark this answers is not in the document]
edwardlee-msft: After talking with Aman, the voice over should indicate who has joined the call rather than voicing out generics.
For example:
"Edward has joined the Call"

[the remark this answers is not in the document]
JamesBurnside: Did he happen to mention about if multiple people joined the call?

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
dmceachernmsft: leftovers from some testing oops. changed back to polite

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
dmceachernmsft: so the bug is explicitly because someone joined. I wonder if there is a difference for people leaving? I programmed this in because I was expecting if we announce one we should announce both. I will ping the A11y people about this.

JamesBurnside: If you are pinging him can you ask about if multiple people joined in quick succession, and for extremes what if 50 people joined at once

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+  participants?: RemoteParticipant[]
+): string => {
+  if (participants) {
+    if (participants.length === 1) {
+      return _formatString(localeString, {
+        displayName: participants[0].displayName ? participants[0].displayName : defaultName
+      });
+    } else {
+      let names = '';
+      participants.forEach((p) => {
```
JamesBurnside, line 99: nit: you can nicely condense this:
```ts
const names = participants.map((p) => p.displayName ?? defaultName).join(' ');
```

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+        if (names === '') {
+          names = names + (p.displayName ? p.displayName : defaultName);
```
JamesBurnside, line 101: if multiple people have no display name you'll have a bunch of `defaultName`s being read out

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
JamesBurnside, line 99: What if there were 20 participants joined, do a11y team really want us to read them all out?
Or could we limit to like 3 and do something like, `Amy, Bob, Chris and 1902 others`

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+        names = names + ' ' + (p.displayName ? p.displayName : defaultName);
```
JamesBurnside, line 103: display names could be multiple words so using a `' '` to join isn't the best, consider
```suggestion
        names = names + ', ' + (p.displayName ? p.displayName : defaultName);
```

packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -327,4 +327,16 @@ export interface CallCompositeStrings {
...
+   */
+  participantJoinedNoticeString: string;
+  /**
+   * Notice to be announced be narrator when a participant leaves a call
+   */
+  participantLeftNoticeString: string;
+  /**
+   * string to be used to announce a change in participant if they have no displayName
+   */
+  defaultParticipantChangedString: string;
```
JamesBurnside, line 341: rename suggestion:
```suggestion
  unnamedParticipantAnnouncedNameString: string;
```

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -68,7 +68,10 @@
     "holdScreenLabel": "You're on hold",
     "openDtmfDialpadLabel": "Show dialpad",
     "dtmfDialpadPlaceholderText": "Enter number",
-    "outboundCallingNoticeString": "Calling..."
+    "outboundCallingNoticeString": "Calling...",
+    "participantJoinedNoticeString": "{displayName} joined",
```
JamesBurnside, line 72: this is actually a list of names now?
```suggestion
    "participantJoinedNoticeString": "{displayNames} joined",
```

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+  const locale = useLocale().strings.call;
+  const [announcerString, setAnnouncerString] = useState<string>('');
+  /**
+   * state to track whether there is currently a timer set in the useParticipantChangedAnnouncement hook
+   */
+  const [timeoutState, setTimeoutState] = useState<ReturnType<typeof setTimeout>>();
+
+  useEffect(() => {
+    const setParticipantEventString = (string: string): void => {
+      setAnnouncerString('');
```
JamesBurnside, line 37: this still has the issues where if a participant joins or leaves before the `PARTICIPANT_ANNOUNCEMENT_DELAY` elapses, the previous announcement won't be read out

JamesBurnside, line 37: what happens if names are being read out and a new user joins? Does it finish what it was saying then read out the new user or does it cut off the old announcement (potentially missing names) then read the new announcement?

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -68,7 +68,10 @@
...
+    "participantLeftNoticeString": "{displayName} left",
+    "defaultParticipantChangedString": "participant"
```
JamesBurnside, line 74: consider 'unnamed participant' to align with other places that have default text for participants without names: https://github.com/Azure/communication-ui-library/pull/1978/files#diff-3a3f9e0f0171ef781ad482c272de2534023eb9b9449a5dca3fdbe3f13ee04df6

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
dmceachernmsft: with polite being set it should finish the current name and then queue up another. but if someone joins in the window their announcement might be lost. I am still waiting to hear back what they are wanting us to do when this happens

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
dmceachernmsft: I think limiting it is a good idea. will make that happen.

packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -327,4 +327,20 @@ export interface CallCompositeStrings {
    * Label disaplayed on the lobby screen during a 1:1 outbound call.
    */
   outboundCallingNoticeString: string;
+  /**
+   * Notice to be announced be narrator when a participant joins a call
```
prprabhu-ms, line 331: nit: be -> by (here and below)

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @returns string to be used by the narrator and Announcer component
+ *
+ * @internal
+ */
+export const useParticipantChangedAnnouncement = (): string => {
+  const locale = useLocale().strings.call;
+  const [announcerString, setAnnouncerString] = useState<string>('');
+  const remoteParticipantsObject = useSelector(getRemoteParticipants);
+  const remoteParticipants = Object.values(remoteParticipantsObject ?? {}).filter((p) => {
```
prprabhu-ms, line 23: Recommendation: create a new selector for this (in the same file is OK) so that the filtering only happens when participants actually change.

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * We need to make sure remote participants are in the connected state. If they are not
+     * they might not have their displayName set in the call state just yet.
+     */
+    if (p.state === 'Connected') {
+      return p;
```
prprabhu-ms, line 29: This works because `p` is guaranteed to be truthy here. But let's be explicit and return `true` because the predicate to `Array.prototype.filter()` is expected to return a boolean.

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return;
+  });
+  const [currentParticipants, setCurrentParticipants] = useState<RemoteParticipantState[]>(remoteParticipants);
```
prprabhu-ms, line 33: The naming is a bit confusing here because `currentParticipants` is actually the participants cached from a _previous_ invocation of this function. `previousParticipants` might be a better name.

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  useEffect(() => {
+    const setParticipantEventString = (string: string): void => {
+      setAnnouncerString('');
+      setAnnouncerString(string);
+    };
+    if (remoteParticipants.length < currentParticipants.length) {
```
prprabhu-ms, line 40: This check isn't quite right. What if one participant joins and one leaves between two invocations to this hook?

Granted it's an edge case, but let's do it right. Compute the difference explicitly first and then check.

```TypeScript

const whoLeft = previousParticipants.filter(p => !remoteParticipants.includes(p));
const whoJoined = remoteParticipants.filter(p => !previousParticipants.includes(p));

if (whoLeft.length > 0) {
  // ...
}
if (whoJoined.length > 0) {
  // ...
} 
```

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+      //someone left
+      const whoLeft = currentParticipants.filter((p) => {
+        if (!remoteParticipants.includes(p)) {
```
prprabhu-ms, line 43: I'm a bit wary of leaving behind hidden O(N*N) cost in the codebase, even when we think that N will stay small.

I'd recommend creating sets for `currentParticipants` and `remoteParticipants` first so that you are assured the repeated `.includes` test doesn't blow up.

Funsies:

image
https://www.joelonsoftware.com/2001/12/11/back-to-basics/

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+        createAnnouncmentString(
+          locale.participantLeftNoticeString,
+          locale.unnamedParticipantChangedString,
+          locale.participantsJoinedOverflowString,
+          whoLeft
+        )
+      );
+      setCurrentParticipants(remoteParticipants);
+    } else if (remoteParticipants.length > currentParticipants.length) {
+      // someone joined
```
prprabhu-ms, line 58: The current code doesn't handle people joining and leaving at the same time (we don't have any place where we create a string that handles both cases together).

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Generates the announcement string for when a participant joins or leaves a call.
+ */
+const createAnnouncmentString = (
+  localeString: string,
+  defaultName: string,
+  overflowString: string,
+  participants?: RemoteParticipantState[]
+): string => {
+  if (participants) {
+    console.log(participants);
```
prprabhu-ms, line 90: debugging `console.log`?

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (participants.length <= 3) {
+      const names = participants.map((p) => p.displayName ?? defaultName).join(', ');
+      return _formatString(localeString, { displayNames: names });
+    } else {
```
prprabhu-ms, line 94: nit: Remove the `else` block because you have an early return (good!)

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+      const names = participants
+        .slice(0, 2)
+        .map((p) => p.displayName ?? defaultName)
+        .join(', ');
+      const namesPlusExtra =
+        names + _formatString(overflowString, { numOfParticipants: numberOfExtraParticipants.toString() });
+      return _formatString(localeString, { displayNames: namesPlusExtra });
+    }
+  } else {
+    return _formatString(localeString, { displayNames: defaultName });
```
prprabhu-ms, line 105: I don't understand how this branch will ever be hit. When would you call `createAnnouncementString` without a list of participants?

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+      return _formatString(localeString, { displayNames: names });
+    } else {
+      const numberOfExtraParticipants = participants.length - 3;
+      const names = participants
+        .slice(0, 2)
+        .map((p) => p.displayName ?? defaultName)
+        .join(', ');
+      const namesPlusExtra =
+        names + _formatString(overflowString, { numOfParticipants: numberOfExtraParticipants.toString() });
+      return _formatString(localeString, { displayNames: namesPlusExtra });
```
prprabhu-ms, line 102: This string manipulation is hard to understand for someone translating the locale strings.

I'd recommend some redundancy: 
* One locale string that is used for <=3 participants
* A separate locale string that is used for >3 participants, which has slots for the 3 participants as well as a slot for the number of remaining particpants.

You could take it further and claim that we shouldn't assume how participants are concatenated together (i.e, don't assume that `Donald, Prathmesh` is the only way to concatenate names together, and just have different locale string for each case:

```TypeScript
{
  'oneParticipantJoined': '{} has joined the call',
  'twoParticipantsJoined': '{} and {} have joined the call',
  'threeParticipantsJoined': '{}, {} and {} have joined the call',
  'manyParticipantsJoined': '{}, {}, {} and {} others have joined the call',
}
```

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (participants) {
+    console.log(participants);
+    if (participants.length <= 3) {
+      const names = participants.map((p) => p.displayName ?? defaultName).join(', ');
+      return _formatString(localeString, { displayNames: names });
+    } else {
+      const numberOfExtraParticipants = participants.length - 3;
+      const names = participants
+        .slice(0, 2)
+        .map((p) => p.displayName ?? defaultName)
```
prprabhu-ms, line 98: nit: Might be nicer to first filter out the participants without a `displayName` so that if there are 3 participants with a name, we announce those names, skipping over any participants without a name.

prprabhu-ms:
Is this ready for another round of reviews?

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,141 @@
...
+// Licensed under the MIT license.
+
+import { _formatString } from '@internal/acs-ui-common';
+import { RemoteParticipantState } from '@internal/calling-stateful-client';
+import { useEffect, useMemo, useState } from 'react';
+import { useLocale } from '../../localization';
+import { CallAdapterState } from '../adapter';
+import { useSelector } from '../hooks/useSelector';
+
+const getRemoteParticipantsConnectedSelector = (callState: CallAdapterState): RemoteParticipantState[] => {
```
JamesBurnside, line 11: if its a selector lets keep with all the other selectors in the other packlet

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,141 @@
...
+          .map((p) => p.displayName ?? strings.unnamedParticipantChangedString)
+          .join(', ')
+      : /** if we have no participants with displayNames we just announce one unnamed participant */
+        strings.unnamedParticipantChangedString;
+
+  const numberOfExtraParticipants =
+    names === strings.unnamedParticipantChangedString ? participants.length - 1 : participants.length - 1;
+
+  const namesPlusExtra =
+    names +
```
JamesBurnside, line 135: this will be _really_ hard for contoso (and us) to localize correctly, are we sure that other languages the grammer will work?

JamesBurnside: I would encourage the `participantsOverflowString` to be:
```
"participantsOverflowString": "{displayName1}, {displayName2}, {displayName3}, and {numOfParticipants} other participants have joined"
```

(and then have one for joined and left, and update the name from participantsOverflowString to something like `manyParticipantsJoinedString` per other comment)

JamesBurnside (changes requested):

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
dmceachernmsft: Since this is a custom selector explicitly for this hook should we be storing it with the mediaGallery selectors?

JamesBurnside: I think media gallery selector works, just as long as its in thie folder with the oder selectors
image

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -68,7 +68,16 @@
     "holdScreenLabel": "You're on hold",
     "openDtmfDialpadLabel": "Show dialpad",
     "dtmfDialpadPlaceholderText": "Enter number",
-    "outboundCallingNoticeString": "Calling..."
+    "outboundCallingNoticeString": "Calling...",
+    "participantJoinedNoticeString": "{displayNames} joined",
```
JamesBurnside, line 72:
```suggestion
    "participantJoinedNoticeString": "{displayName} joined",
```
?

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -68,7 +68,16 @@
...
+    "twoParticipantJoinedNoticeString": "{displayName1} and {displayName2} have joined",
+    "threeParticipantJoinedNoticeString": "{displayName1}, {displayName2} and {displayName3} have joined",
+    "participantLeftNoticeString": "{displayNames} left",
+    "twoParticipantLeftNoticeString": "{displayName1} and {displayName2} have left",
+    "threeParticipantLeftNoticeString": "{displayName1}, {displayName2} and {displayName3} have left",
+    "unnamedParticipantChangedString": "unnamed participant",
+    "manyUnnamedParticipantsChanged": "unnamed participant and {numOfParticipants} other participants",
+    "manyParticipantsChanged": "{displayName1}, {displayName2}, {displayName3} and {numOfParticipants} other participants"
```
JamesBurnside, line 80: Open to push back on this, I think it would be clearer to have manyParticipantsLeft and manyParticipantsJoined as its very difficult to know how `manyParticipantsChanged` would be used

packages/react-composites/src/composites/localization/locales/en-US/strings.json
JamesBurnside: (same with other single displayName strings below)

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+      numOfParticipants: (participants.length - 1).toString()
+    });
+    return _formatString(
+      direction === 'joined' ? strings.participantJoinedNoticeString : strings.participantLeftNoticeString,
+      { displayNames: participantsString }
+    );
+  }
+  const oneName = participants[0].displayName ?? strings.unnamedParticipantChangedString;
+  const twoNames = participants.map((p) => p.displayName ?? strings.unnamedParticipantChangedString);
+  const threeNames = participants.map((p) => p.displayName ?? strings.unnamedParticipantChangedString);
```
JamesBurnside, line 109: nit: twoNames and threeNames do exactly the same thing, and one name functionally does the same thing.
Consider 
```
const participantNames = participants.map((p) => p.displayName ?? strings.unnamedParticipantChangedString);

switch (participantNames.length) {
  case 0: return;
  case 1: ...  { displayName: participantNames[0] }
  case 2: ...  { displayName1: participantNames[0], participantNames[1] }
  case 3: ... { displayName1: participantNames[0], participantNames[1], participantNames[2] }
  default: // many participants
    ... { displayName1: participantNames[0], participantNames[1], participantNames[2], numberOfExtraParticipants }
```

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Generates the announcement string for when a participant joins or leaves a call.
+ */
+export const createAnnouncmentString = (
+  direction: 'joined' | 'left',
+  participants: RemoteParticipantState[],
+  strings: ParticipantChangedAnnouncmentStrings
+): string => {
+  /**
+   * Check that we have more than 1 participant, if they all have no displayName return unnamed participants
```
JamesBurnside, line 95: what is the purpose of this? doesn't your switch case below handle everything?

JamesBurnside: It won't catch the chance that 2 unnamed participants and 1 named participant joins the call. It's probably also a rare enough case that reading "unnamed participant, unnamed participant and unnamed participant joined the call" is fine (3 people of that name _did_ join and it also lets the a11y user know that three people joined the call)

packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -327,4 +327,40 @@ export interface CallCompositeStrings {
...
    */
   outboundCallingNoticeString: string;
+  /**
+   * Notice to be announced by narrator when a participant joins a call
+   */
+  participantJoinedNoticeString: string;
+  /**
+   * Notice to be announced by narrator when a participant joins a call
+   */
+  twoParticipantJoinedNoticeString: string;
```
prprabhu-ms, line 337: nit: `twoParticipantsJoinedNoticeString` (plural) etc.

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -76,18 +81,21 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
+          styles={VideoGalleryStyles}
+          layout="floatingLocalVideo"
+          showCameraSwitcherInLocalPreview={props.isMobile}
+          localVideoCameraCycleButtonProps={cameraSwitcherProps}
+          onRenderAvatar={onRenderAvatar}
+        />
+      </>
     );
-  }, [videoGalleryProps, props.isMobile, onRenderAvatar, cameraSwitcherProps]);
+  }, [videoGalleryProps, props.isMobile, onRenderAvatar, cameraSwitcherProps, announcerString]);
```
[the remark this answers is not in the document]
prprabhu-ms: +1

We should use `React.memo` instead of `useMemo` when we're trying to "memoize" components.

This PR didn't add the `useMemo`, but it does add more complexity in an already iffy place here.

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+import { useEffect, useMemo, useState } from 'react';
+import { useLocale } from '../../localization';
+import { CallAdapterState } from '../adapter';
+import { useSelector } from '../hooks/useSelector';
+
+/**
+ * Custom selector for this hook to retrieve all the participants that are currently
+ * connected to the call.
+ */
+const getRemoteParticipantsConnectedSelector = (callState: CallAdapterState): RemoteParticipantState[] => {
```
prprabhu-ms, line 15: Actually make this a selector. See https://github.com/Azure/communication-ui-library/blob/main/docs/architecture/WritingSelectors.md

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+ * sets the announcement string whenever a Participant comes or goes from a call to be
+ * used by the system narrator.
+ *
+ * @returns string to be used by the narrator and Announcer component
+ *
+ * @internal
+ */
+export const useParticipantChangedAnnouncement = (): string => {
+  const locale = useLocale().strings.call;
+  const strings = useMemo(() => {
```
prprabhu-ms, line 49: nitty nit: We usually just use the entire `locale` in the dependency array of hooks (below).
So you don't _really_ need this `useMemo`.

Either way is OK.

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+  const remoteParticipants = useSelector(getRemoteParticipantsConnectedSelector);
+  const [previousParticipants, setPreviousParticipants] = useState<RemoteParticipantState[]>(remoteParticipants);
+
+  useEffect(() => {
+    const setParticipantEventString = (string: string): void => {
+      setAnnouncerString('');
+      setAnnouncerString(string);
+    };
+
+    if (previousParticipants.length > remoteParticipants.length) {
```
prprabhu-ms, line 74: My question about handling the case when folks join and leave at the same time is still open.

We should first compute `whoLeft` and `whoJoined` and only then check for the two cases (`whoLeft.length > 0` and `whoJoined.length > 0`)

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+  direction: 'joined' | 'left',
+  participants: RemoteParticipantState[],
+  strings: ParticipantChangedAnnouncmentStrings
+): string => {
+  /**
+   * Check that we have more than 1 participant, if they all have no displayName return unnamed participants
+   * overflow string
+   */
+  if (participants.filter((p) => p.displayName).length === 0 && participants.length > 0) {
+    const participantsString = _formatString(strings.manyUnnamedParticipantsChanged, {
```
prprabhu-ms: I'd still push back on putting together strings from the locale like we do here.

It's clearer to add multiple strings here as well `manyUnnamedParticipantsJoined` `mainUnnamedParticipantsLeft`

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
JamesBurnside: > Should we never announce unnamed participant

I'd like a11y team to sign off this assumption. There is some value in knowing that someone without a name joined your meeting - should they even be there without a name?

dmceachernmsft:
closing this PR, the work stream is being picked up in this one with some fancy tests: https://github.com/Azure/communication-ui-library/pull/2546

PR #2504, opened 2022-11-08
Split `CHANGELOG.md` files for beta and stable builds

prprabhu-ms:
See also: An alternative approach in #2512

Split the current `CHANGELOG`s for `stable` vs `beta` channel releases. A follow up PR will update the tooling so that future releases correctly update the relevant `CHANGELOG`s.

- Split the change files per build flavor
- Split all existing `CHANGELOG.md` files for the two flavors
- Update `.gitignore` so we don't accidentally commit a build flavor-unaware `CHANGELOG`.
- Add a top-level `CHANGELOG.md` that points to the correct file deep in our tree. The lack of this file has been very confusing for devs, even on our own team.

Note that the `CHANGELOG.json` files are _not_ split as the JSON data is already keyed by the release version.


prprabhu-ms:
closed in favor of #2512

PR #2506, opened 2022-11-09
Remove icon re-registered warning from all 3 composites and storybook landing page

carocao-msft:
# What
Remove icon re-registered warning from all 3 composites and storybook landing page

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3043083

# How Tested
Go to each sample and see if you can see icon reregistered warnings
For calling and callwithchat, the only reregister warning you should be able to see is for icon 'BackSpace', this is due to needing to register backspace icon in HomeScreen.tsx outside of call composite. Remove registerIcons in homescreen then you should not be able to see any  reregister warning=> if contoso is only using Call Composite/ Call with chat composite, should not see warnings 

For chat you can still see reregister warning, which is coming from ChatHeader component outside of chat composite. the warning should disappear when ChatHeader is removed => if contoso is only using Chat Composite, should not see warnings 

For storybook should not see reregister warning on landing page, but still can see reregister warning for composite stories due to needing to register icons for storybook preview and register icons inside composites

# Process & policy checklist


packages/react-components/src/theming/icons.tsx
```diff
@@ -246,8 +243,7 @@ export const DEFAULT_COMPONENT_ICONS = {
   SendBoxSend: <Send20Regular />,
   SendBoxSendHovered: <Send20Filled />,
   VideoTileMicOff: <MicOff16Filled />,
-  /* @conditional-compile-remove(dialpad) */ /* @conditional-compile-remove(PSTN-calls) */
-  BackSpace: <Backspace20Regular />,
```
JamesBurnside, line 250: Given this is still in beta, consider instead updating this icon name to `DialpadBackspace`

JamesBurnside (approved):

packages/react-components/src/theming/icons.tsx
carocao-msft: Updated

anjulgarg (approved):

PR #2517, opened 2022-11-10
`initializeIcons()` in PSTN and 1:N quickstarts

prprabhu-ms:
After this PR, the quickstart applications show icons correctly out of the box.

### DIFF
--- a/change/@internal-storybook-a6d95240-9957-4057-a6ce-6d2a50b18088.json
+++ b/change/@internal-storybook-a6d95240-9957-4057-a6ce-6d2a50b18088.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "initializeIcons() in 1:N and PSTN quickstarts",
+  "packageName": "@internal/storybook",
+  "email": "82062616+prprabhu-ms@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


change/@internal-storybook-a6d95240-9957-4057-a6ce-6d2a50b18088.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
prprabhu-ms, line 2: Unsure if this should be `prerelease` or `none`. It doesn't change anything in the library, but fixes the documentation for a beta feature.

dmceachernmsft: I like pre-release since its about fixing docs for a feature we haven't released yet. Is there value for having a new kind of changelog for doc-changes? like `doc-change` get visibility on when we are improving our storybook?

dmceachernmsft (approved):

change/@internal-storybook-a6d95240-9957-4057-a6ce-6d2a50b18088.json
prprabhu-ms: Will stick to `prerelease` then. A new type is probably not worth it.

PR #2520, opened 2022-11-11
Add type cast to support latest lib.dom version

PorterNan:
# What

There are 3 warnings popping up after doing a `rush update` and it seems because of lib.dom update, which changes some type definition, this PR is to do the right type cast for bypass warnings

Check this build for those 3 warnings: https://github.com/Azure/communication-ui-library/actions/runs/3434222284/jobs/5725340021

image

# Why

# How Tested

# Process & policy checklist


PorterNan:
Merge this to unblock handlers pr

PR #2525, opened 2022-11-14
Call Readiness should guide user through enable device permissions 

carocao-msft:
# What
Auto prompt for access on load
show allow device permission modal
if permission denied show denied modal
if permission granted show nothing
This PR does not include fall back when permission api is not supported
This PR include e2e tests

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3034528

# How Tested
Turn call readiness on in App.tsx
go through sample and test various permission cases (both prompt, both denied, one of them denied, one of them prompt...)
User should always be able to see guidance to enable permissions (either through modals/error bars)
User does not need to enable camera to start call (click on continue without camera to dismiss modal and start call) 

1. prompt for both and deny both

https://user-images.githubusercontent.com/96077406/201764208-a4d37576-f233-49dd-954f-689cdfb3b5f5.mov

2. prompt for both and allow both

https://user-images.githubusercontent.com/96077406/201764248-a748b704-8dd9-4242-b2e1-31bd812dfcaf.mov

3. one prompt and one already denied

https://user-images.githubusercontent.com/96077406/201766615-857dde5a-3b6a-43ee-8c66-468d717c4cab.mov

4. user dismiss prompt window desktop (dismiss is considered as denied)

https://user-images.githubusercontent.com/96077406/201766147-d9ba5ac4-ad12-4dec-8847-b25bf5ca384f.mov

5.  user dismiss prompt window mobile( if dismiss both, still consider as prompt ; if set permission for first prompt and dismiss the second, the second is considered denied ; if dismiss the first and set permission for the second, the first is still considered prompt) 

https://user-images.githubusercontent.com/96077406/201766268-f1b25ec8-6705-490a-99b0-e9db9d9bf708.mov

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -157,18 +163,8 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
-  /* @conditional-compile-remove(call-readiness) */
-  const onLightDismissTriggered = (): void => {
-    // do nothing here
-    // only way to dismiss this drawer is clicking on allow access which will leads to device permission prompt
-  };
-
   /* @conditional-compile-remove(call-readiness) */
-  const [isModalShowing, setIsModalShowing] = useState(false);
-
+  const [isModalShowing, setIsModalShowing] = useState(true);
```
JamesBurnside, line 167: nit: specify what "modal" is here
```suggestion
  const [isAskPermissionsModalShowing, setIsAskPermissionsModalShowing] = useState(true);
```

JamesBurnside, line 167: Is this always showing by default? what if they have accepted permissions previously, or if they have opted out of call readiness?

carocao-msft: we use the video and audio state to determine which modal to show or show nothing, this is just for making modal disappear when dismissed

JamesBurnside: Can you update the name to reflect this? If someone is reading this or debugging they will see this as true and expect a modal to be showing.
Something like hasPermissionsModalBeenDismissed would better reflect this variables purpose

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -303,3 +268,181 @@ const localPreviewTrampoline = (mobileView: boolean, doNotShow?: boolean): JSX.E
...
+  audioState?: PermissionState | undefined
+): boolean | undefined => {
+  /* @conditional-compile-remove(call-readiness) */
+  return audioState ? audioState === 'granted' : microphonePermissionGranted;
+
+  return microphonePermissionGranted;
+};
+
+/* @conditional-compile-remove(call-readiness) */
+const callReadinessModal = (
```
JamesBurnside, line 293: can we move this to its own file? ConfigurationPage is getting long

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -303,3 +268,181 @@ const localPreviewTrampoline = (mobileView: boolean, doNotShow?: boolean): JSX.E
...
+        />
+      );
+    }
+    // if neither state is denied or prompt, we don't need to return anything
+    else {
+      return undefined;
+    }
+  };
+
+  if (mobileView && modal() !== undefined) {
```
JamesBurnside, line 422: I'm not sure what react does here when a react component is used in a conditional, my assumption is it has to perform the same render computation it would do in a render pass but my gut reaction is to avoid this

Something you can do that makes the code slightly more complicated is:
const 


But happy to have push back and leave this as is 
```tsx
const showModal = videoState === 'denied' || videoState === 'prompt || audioState === 'denied' || audioState === 'prompt;
const modal: undefined | () => JSX.Element = !showModal ? undefined : () => {
   ... modal code
}
```

So instead of a function that returns a react node or undefined, the function is undefined or it returns a react element

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -303,3 +268,181 @@ const localPreviewTrampoline = (mobileView: boolean, doNotShow?: boolean): JSX.E
...
+): boolean | undefined => {
+  /* @conditional-compile-remove(call-readiness) */
+  return audioState ? audioState === 'granted' : microphonePermissionGranted;
+
+  return microphonePermissionGranted;
+};
+
+/* @conditional-compile-remove(call-readiness) */
+const callReadinessModal = (
+  mobileView: boolean,
```
JamesBurnside, line 294: when there's so many args passed into a function, consider using named args instead
i.e. fn({arg1: val, arg2: val, arg3: val) instead of fn(arg1, arg2, arg3)

Given this is a JSX Element best to make this a standard react component and update the signature to:
```tsx
const CallReadinessModal = (props: {...}) => JSX.Element
```

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
carocao-msft: made changes and pushed to PR https://github.com/Azure/communication-ui-library/pull/2529

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
carocao-msft: made changes and pushed to PR https://github.com/Azure/communication-ui-library/pull/2529

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
carocao-msft: made changes and pushed to PR https://github.com/Azure/communication-ui-library/pull/2529

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
carocao-msft: made changes and pushed to PR https://github.com/Azure/communication-ui-library/pull/2529

carocao-msft: made changes and pushed to PR https://github.com/Azure/communication-ui-library/pull/2529

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -303,3 +268,181 @@ const localPreviewTrampoline = (mobileView: boolean, doNotShow?: boolean): JSX.E
...
+  cameraPermissionGranted: boolean | undefined,
+  videoState?: PermissionState | undefined
+): boolean | undefined => {
+  /* @conditional-compile-remove(call-readiness) */
+  return videoState ? videoState === 'granted' : cameraPermissionGranted;
+
+  return cameraPermissionGranted;
+};
+
+const micPermissionGrantedTrampoline = (
```
edwardlee-msft, line 282: nit: To be consistent, I would rename this as microphonePermissionGrantedTrampoline (though this will be omitted once this goes into stable)

edwardlee-msft (approved):

PR #2526, opened 2022-11-14
Fix roleHint usage in Calling sample

mgamis-msft:
# What

- Memoized roleHint in sample Calling app
- Exporting AzureCommunicationCallAdapterOptions

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3055004

# How Tested

Local Calling sample testing

# Process & policy checklist


samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -97,7 +97,7 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
       /* @conditional-compile-remove(PSTN-calls) */
       alternateCallerId,
       /* @conditional-compile-remove(rooms) */
-      options: { roleHint }
+      options: roleHint ? { roleHint } : undefined
```
JamesBurnside, line 100: this still needs extracted and memoized

```ts
  const callAdapterOptions: AzureCommunicationCallAdapterOptions = useMemo(() => ({
    roleHint
  }), [roleHint]);
```

JamesBurnside:
update PR title and desc with more details

samples/Calling/src/app/views/CallScreen.tsx
mgamis-msft: The ternary suffices but sure I'll do that for correctness.

alkwa-msft: doesnt this say if roleHint exists then return a new JSON object which contains the roleHint else undefined? so it would always be new?

JamesBurnside:
re-assigning myself to prathmesh for load balancing

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -732,6 +732,21 @@ export type CallAdapterLocator =
   | /* @conditional-compile-remove(rooms) */ RoomCallLocator
   | /* @conditional-compile-remove(teams-adhoc-call) */ /* @conditional-compile-remove(PSTN-calls) */ CallParticipantsLocator;
 
+/* @conditional-compile-remove(rooms) */
```
prprabhu-ms, line 735: nit: Define the type without any conditional compilation so you can use it internally.

It is already conditionally exported from `./index.tsx`, as it should be.

dmceachernmsft (approved):

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -88,6 +90,9 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     [callReadinessOptedIn]
   );
 
+  /* @conditional-compile-remove(rooms) */
+  const callAdapterOptions: AzureCommunicationCallAdapterOptions = useMemo(() => ({ roleHint }), [roleHint]);
```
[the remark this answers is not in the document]
mgamis-msft: I didn't need to. But I felt like I need the options to have a type instead of just being an object.

PR #2527, opened 2022-11-15
Update small packages 

dmceachernmsft:
# What

Updates the lockfile for any new versions of our dependencies
# Why

Allows us to update the UI snapshots per these smaller updates

# How Tested

Update snapshots


JamesBurnside:
These look like they've moved too far apart:
image

JamesBurnside:
Unassigning myself for load balancing

prprabhu-ms (commented):
image

`SendBox` with files now has more space under the text than earlier / figma. Worth looking at why.

prprabhu-ms (commented):
The two pointed out so far are worth looking at. The other two:

* Dialpad numbers/text have extra padding between them
* FileSharing error bar has an extra pixel or 2 height

are OK. They bring the components in alignment with the figma.

JamesBurnside (approved):

packages/react-components/src/components/DevicePermissions/DomainPermissionsScaffolding.tsx
```diff
@@ -116,19 +117,25 @@ export const DomainPermissionsContainer = (props: DomainPermissionsContainerProp
   const theme = useTheme();
   return (
     <Stack style={{ padding: '2rem', maxWidth: '25.375rem' }} aria-label={strings?.ariaLabel}>
-      <Stack horizontal style={{ paddingBottom: '1rem' }} horizontalAlign={'space-between'}>
+      <Stack
+        styles={iconBannerContainerStyles}
+        horizontal
+        horizontalAlign={'center'}
+        verticalFill
+        tokens={{ childrenGap: '2rem' }}
```
prprabhu-ms, line 125: extract a file-level constant for `tokens` so you're not passing in a new object each time.

packages/react-components/src/components/DevicePermissions/DomainPermissionsScaffolding.tsx
```diff
@@ -116,19 +117,25 @@ export const DomainPermissionsContainer = (props: DomainPermissionsContainerProp
   const theme = useTheme();
   return (
     <Stack style={{ padding: '2rem', maxWidth: '25.375rem' }} aria-label={strings?.ariaLabel}>
-      <Stack horizontal style={{ paddingBottom: '1rem' }} horizontalAlign={'space-between'}>
+      <Stack
+        styles={iconBannerContainerStyles}
+        horizontal
+        horizontalAlign={'center'}
```
prprabhu-ms, line 123: This can just be `horizontalAlign="center"`

https://reactjs.org/docs/introducing-jsx.html#specifying-attributes-with-jsx

packages/react-components/src/components/styles/DomainPermissions.styles.ts
```diff
@@ -9,11 +9,23 @@ import { _pxToRem } from '@internal/acs-ui-common';
...
   }
 };
 
+/**
+ * @internal
+ */
+export const iconBannerContainerStyles: IStackStyles = {
+  root: {
+    paddingBottom: '1rem',
+    width: '20rem',
```
JamesBurnside, line 24: is width and margin needed or can we remove? not clear what they're doing here

JamesBurnside: (I would expect width to be `100%` if its needed)

dmceachernmsft: not needed so removed.

prprabhu-ms (approved):

PR #2529, opened 2022-11-15
Fall Back UI Call readiness 

carocao-msft:
# What
Call Readiness will display 2 screens when permission api is not availble
Checking for permissions at start (show up for 2 secs)
Permission denied (when user deny permissions)

# Why
When permission api is not supported ,we fall back to using calling sdk's own device permission
since there are delay on load to get the correct device permission information, we show 'checking for permission' screen for 2 secs
For Calling SDK's own device permission, undefined means permission not set up yet, true means granted and false means denied. However it does not support the case when one of the device is not set up and one of the device is set to denied/granted. 
Due to the above reason, we don't display the 'allow device permission' screen when using  calling sdk's own device permission, to avoid user confusion when brower only prompt for camera access, and modal pop up says "enable mic and camera"

# How Tested
Go to function getDevicePermissionState in utils,
change  .query({ name: 'camera' }) and .query({ name: 'microphone' }) to .query({ name: 'hi' }) and .query({ name: 'hello' })
(change to what ever names that are unsupported) 
Go through calling /callwithchat sample to see if the correct screens show up

https://user-images.githubusercontent.com/96077406/201813383-4d763b77-a09f-45d7-b7a4-a03ed999c707.mov

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -446,3 +479,147 @@ const callReadinessModal = (
...
+    // only way to dismiss this drawer is clicking on allow access which will leads to device permission prompt
+  };
+
+  // When permissions are not set, value is undefined, do nothing here
+  // When permissions are set to denied, value is false, show helper screen
+  const modal = (): JSX.Element | undefined => {
+    if (cameraPermissionGranted === false && microphonePermissionGranted === false) {
+      return (
+        <CameraAndMicrophoneDomainPermissions
+          appName={'app'}
```
JamesBurnside, line 513: is there an ADO item to provide a better value for this?

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -163,8 +163,22 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
   const [isModalShowing, setIsModalShowing] = useState(true);
+
+  // When permission API is not available, we want to show screen saying checking for access (disappears on its own)
+  // then based on permission setting, we show permission denied or nothing
+  /* @conditional-compile-remove(call-readiness) */
+  const [checkPermissionModalShowing, setCheckPermissionModalShowing] = useState(true);
+  /* @conditional-compile-remove(call-readiness) */
+  const [permissionDeniedModalShowing, setPermissionDeniedModalShowing] = useState(false);
+  /* @conditional-compile-remove(call-readiness) */
+  setTimeout(() => {
```
JamesBurnside, line 177: what is this timeout for?

JamesBurnside: Oh I see this auto dismisses the checking. We want to keep the "checking" modal active until the user has either clicked accept or deny the system permission prompt:
image
So the checking modal should be showing and active whilst the systme prompt is displayed

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -202,6 +216,25 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
         )
       }
 
+      {
+        /* @conditional-compile-remove(call-readiness) */ callReadinessOptedIn &&
```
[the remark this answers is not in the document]
carocao-msft: apply to everything under

packages/react-composites/src/composites/CallComposite/components/CallReadinessModal.tsx
```diff
@@ -0,0 +1,357 @@
...
+    isPermissionsModalDismissed,
+    setIsPermissionsModalDismissed,
+    onPermissionsTroubleshootingClick
+  } = props;
+  const onLightDismissTriggered = (): void => {
+    // do nothing here
+    // only way to dismiss this drawer is clicking on allow access which will leads to device permission prompt
+  };
+
+  console.log(videoState, audioState);
```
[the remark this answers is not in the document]
carocao-msft: oops totally missed that! Updated

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -158,13 +145,28 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
   const networkErrors = errorBarProps.activeErrorMessages.filter((message) => message.type === 'callNetworkQualityLow');
+
   /* @conditional-compile-remove(call-readiness) */
-  const [isModalShowing, setIsModalShowing] = useState(true);
+  const [isPermissionsModalDismissed, setIsPermissionsModalDismissed] = useState(true);
+
+  // When permission API is not available, we want to show screen saying checking for access (disappears on its own)
+  // then based on permission setting, we show permission denied or nothing
+  /* @conditional-compile-remove(call-readiness) */
+  const [checkPermissionModalShowing, setCheckPermissionModalShowing] = useState(true);
```
JamesBurnside, line 161: `setCheckPermissionModalShowing` and `permissionDeniedModalShowing` are always inverse of each other (and hence directly correlated). To reduce redundancy can we combine these variables into:

```ts
const [minimumFallbackTimerElapsed, setMinimumFallbackTimerElapsed] = useState(false);
```

then for below it becomes:
```tsx

const forceShowingCheckPermissions = !minimumFallbackTimerElapsed;

<CallReadinessModalFallBack
  // rename this to forceCheckPermissionsModalShowing as this can show even
  // when the prop is false (when the permissions aren't known yet due to system prompt)
  checkPermissionModalShowing={forceShowingCheckPermissions}
/>
```

and then remove the `permissionDeniedModalShowing` param as this will be determined based on the other props passed in

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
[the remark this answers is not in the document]
edwardlee-msft: Great. As long as we are sure that the rest of the conditional statement does not proceed then I am good to approve this pr.

edwardlee-msft (approved):

PR #2531, opened 2022-11-16
managing threads with an explicit user id instead of managing all kno…

alkwa-msft:
…wn threads

# What
Adding an explicit requirement to provide an admin user id in appsettings.json.
Removing the thread mapping we have in our server.
There is a chance an invalid thread could be entered but this is only if users try to do something funny.

# Why
To align with an improved process we have figured out building self host. It is better to hydrate the admin user Id for adding users to threads than maintaining a map of all known threads. The issue was that if the server ever restarted we would lose out on joining those threads. By using a "stable-enough" userId, then we can leverage the threads between server restarts. This also means if someone is building another chat application and they want to debug with the chat hero sample we can just use the same resource and adminUserId.

* There is a potential where if you try to add in a made up threadId then the chat composite or callwithchat composite will load but you will get an error saying you are not part of the thread.

# How Tested
Tested locally and we have used this pattern in other applications

Yes, you will now need to supply a userId for the admin.


samples/Server/src/routes/addUser.ts
```diff
@@ -26,20 +27,29 @@ interface AddUserParam {
...
+      participants: [
+        {
+          id: { communicationUserId: addUserParam.Id },
+          displayName: addUserParam.DisplayName
+        }
+      ]
+    });
+    res.sendStatus(201);
+  } catch (err) {
+    res.sendStatus(404);
```
JamesBurnside, line 51: nit: `500` error?

edwardlee-msft:
This is great! 
One follow up question would be, how do we get a list of chat threads the admin user moderated/ is a part of?

samples/Server/src/routes/addUser.ts
alkwa-msft: The thread was not found so 404? Its not really a server I give up error but rather someone is trying to add themselves to a thread that the adminUserId is not part of. I am open to 500 as well. 500 to me always sounds like completely unexpected on the developer and its a big catch all for the server threw an exception.

alkwa-msft:
> This is great! A plus side to this as well, is we can easily have long running conversations to test with! One follow up question would be, how do we get a list of chat threads the admin user moderated/ is a part of?

I believe there is a listChatThreads function on the chatClient that will allow us to open this up in the future if we want to.

https://learn.microsoft.com/en-us/javascript/api/@azure/communication-chat/chatclient?view=azure-node-latest#@azure-communication-chat-chatclient-listchatthreads

samples/Server/src/routes/addUser.ts
JamesBurnside: Is `thread not found` the only reason this throws an error? If so 404 yeah makes sense but add a comment to mention we return 404 for that reason

samples/Server/src/lib/identityClient.ts
```diff
@@ -8,15 +8,17 @@ import {
...
+import { getResourceConnectionString, getAdminUserId } from './envHelper';
 
 // lazy init to allow mocks in test
 let identityClient: CommunicationIdentityClient | undefined = undefined;
 const getIdentityClient = (): CommunicationIdentityClient =>
   identityClient ?? (identityClient = new CommunicationIdentityClient(getResourceConnectionString()));
 
 // replicate here to allow for mocks in tests
-export const createUser = (): Promise<CommunicationUserIdentifier> => getIdentityClient().createUser();
+export const createUser = (): CommunicationUserIdentifier => {
```
JamesBurnside, line 19: can you rename this to `createAdminUser` ? To avoid confusion with `createUserAndToken` below which creates a non-admin user 
```suggestion
export const createAdminUser = (): CommunicationUserIdentifier => {
```

change-beta/@azure-communication-react-2839d297-3f9b-45d1-8db6-58f7175c5f19.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
JamesBurnside, line 2: This should be `patch` or `none` as it applies to stable releases, I'd recommend `none` as we don't include sample changes in the changelog 
@prprabhu-ms we should add this detail to https://github.com/Azure/communication-ui-library/blob/main/docs/references/release-changelogs.md

samples/Server/appsettings.json
```diff
@@ -8,5 +8,6 @@
   },
   "AllowedHosts": "*",
   "ResourceConnectionString": "REPLACE_WITH_CONNECTION_STRING",
-  "EndpointUrl":"REPLACE_WITH_ENDPOINT_URL"
+  "EndpointUrl":"REPLACE_WITH_ENDPOINT_URL",
+  "AdminUserId": "REPLACE_WITH_ADMIN_USER_ID"
```
JamesBurnside, line 12: update readme instructions on populating this new setting

JamesBurnside: Also add details to the readme on what an admin user id is for in this sample application

samples/Server/src/lib/identityClient.ts
JamesBurnside: Actually `getAdminUser` would be more appropriate as it doesn't create a new admin

JamesBurnside (approved):

samples/Server/src/lib/identityClient.ts
alkwa-msft: I like getAdminUser thanks for the suggestion!

PR #2533, opened 2022-11-16
Show ? icon when unnamed participant

anjulgarg:
# What
Show ? icon when unnamed participant

# Why

# How Tested

# Process & policy checklist


packages/react-composites/tests/browser/call/hermetic/fixture.ts
```diff
@@ -102,7 +102,7 @@ export function defaultMockCallAdapterState(participants?: MockRemoteParticipant
  *
  * Use this to add participants to state created via {@link defaultCallAdapterState}.
  */
-export function defaultMockRemoteParticipant(displayName: string): MockRemoteParticipantState {
+export function defaultMockRemoteParticipant(displayName?: string): MockRemoteParticipantState {
   return {
     identifier: { kind: 'communicationUser', communicationUserId: `8:acs:${displayName}-id` },
```
alkwa-msft, line 107: if we don't supply a displayName will the identifier be 8:acs-id?

Perhaps we need to detach the concept of acs user id from display name?

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -163,11 +163,12 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
   const strings = { ...localeStrings, ...props.strings };
 
   const avatarOptions = {
-    text: displayName,
+    text: displayName ?? strings.displayNamePlaceholder,
```
alkwa-msft, line 166: On the participant list in the ui snapshots we are still seeing UP instead of the person tile image we are seeing in the video gallery. Is this expected?

anjulgarg: Was still working on this PR. It is fixed now.

packages/react-composites/tests/browser/call/hermetic/fixture.ts
anjulgarg: Correct. Doesn't really matter. it's just a mock name,

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -165,4 +165,24 @@ test.describe('Participant pane tests', async () => {
     }, participantStringId);
     expect(await stableScreenshot(page)).toMatchSnapshot('PSTN-participant-pane-callee-name-truncation.png');
   });
+
+  /* @conditional-compile-remove(one-to-n-calling) @conditional-compile-remove(PSTN-calls) */
+  test('Participant shows unknown icon when displayName is missing', async ({ page, serverUrl }, testInfo) => {
```
dmceachernmsft, line 170: Should we also test the case where the displayName is set by the user serendipitously the same as the localized placeholder?

anjulgarg: I think that's a side effect.
We want to test the expected behaviour.
If something changes in the future where our logic is more streamlined, the side effect goes away but our test will start failing.

edwardlee-msft (approved):

dmceachernmsft (approved):

PR #2536, opened 2022-11-17
VideoGallery default layout enzyme tests

mgamis-msft:
# What

Added VideoGallery default layout enzyme tests

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3054370

# How Tested

Run enzyme tests locally

# Process & policy checklist


anjulgarg (approved):

packages/react-components/src/components/VideoGallery.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+describe('VideoGallery default layout', () => {
+  beforeAll(() => {
+    registerIcons({
+      icons: {
+        horizontalgalleryleftbutton: <></>,
+        horizontalgalleryrightbutton: <></>
+      }
+    });
+
+    // Need to mock this because the HorizontalGallery uses this function. But JSDOM does not actually do any
```
prprabhu-ms, line 27: Great comment. Tells me exactly why you're doing the black magic <3

packages/react-components/src/components/VideoGallery.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+const gridVideoTileCount = (root: ReactWrapper<VideoGalleryProps>): number =>
+  root.find(GridLayout).find(StreamMedia).length;
+
+const gridAudioTileCount = (root: ReactWrapper<VideoGalleryProps>): number =>
+  root.find(GridLayout).find(Persona).length;
+
+const createLocalParticipant = (attrs?: Partial<VideoGalleryLocalParticipant>): VideoGalleryLocalParticipant => {
+  return {
+    userId: attrs?.userId ?? `localParticipant-${Math.random()}`,
```
prprabhu-ms, line 151: Let's avoid randomness in tests as much as possible. There's only ever one local participant, right? Can we hard-code the default. The test can provided a non-default value if it wants for a second local participant.

packages/react-components/src/components/VideoGallery.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+const createVideoDivElement = (): HTMLDivElement => {
+  const divElement = document.createElement('div');
+  divElement.innerHTML = '<video>VIDEO</video>';
+  return divElement;
+};
+
+const createRemoteParticipant = (attrs?: Partial<VideoGalleryRemoteParticipant>): VideoGalleryRemoteParticipant => {
+  return {
+    userId: attrs?.userId ?? `remoteParticipant-${Math.random()}`,
```
prprabhu-ms, line 173: Let's avoid randomness in tests as much as possible. I'd just change the type of `attrs` to _require_ `userId` to be set, and then set it from the tests (could be set using index in your Array constructor).

packages/react-components/src/components/VideoGallery.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    expect(root.find(HorizontalGallery).exists()).toBe(false);
+  });
+
+  test('should render max allowed video tiles with streams in the grid ', () => {
+    const localParticipant = createLocalParticipant({
+      videoStream: { isAvailable: true, renderElement: createVideoDivElement() }
+    });
+    // Need to mock hook _useContainerWidth because the returned width is used by HorizontalGallery to decide how many
+    // tiles to show
+    jest.spyOn(responsive, '_useContainerWidth').mockImplementation(() => {
```
prprabhu-ms, line 104: Can we mock this at the top along with `_convertRemToPx`. It'll avoid having to mock for future tests, and keep all the mocking in one place.

It's also a very similar reason for mocking.

prprabhu-ms (approved):
Very well written PR. Thank you!

PR #2546, opened 2022-11-19
[A11y bug] Participant changed announcements mk 2

dmceachernmsft:
# What

Introduces narrator announcements for when participants change in the call
# Why

allows users in a call using narrator to know when participants come and go from a call.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3007443
# How Tested

With the wonderful guidance of @prprabhu-ms  we created a series of tests to ensure the output of the hook is correct when participants come in go in different combinations


change/@internal-react-composites-81715f29-3fde-4133-99e8-70378241e624.json
```diff
@@ -0,0 +1,7 @@
+{
```
JamesBurnside, line 1: cc @prprabhu-ms bug or intentional this only created a stable changefile for this changfile?

JamesBurnside: (would expect to see both a stable and a beta changelog)

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -76,18 +81,21 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
+          styles={VideoGalleryStyles}
+          layout="floatingLocalVideo"
+          showCameraSwitcherInLocalPreview={props.isMobile}
+          localVideoCameraCycleButtonProps={cameraSwitcherProps}
+          onRenderAvatar={onRenderAvatar}
+        />
+      </>
     );
-  }, [videoGalleryProps, props.isMobile, onRenderAvatar, cameraSwitcherProps]);
+  }, [videoGalleryProps, props.isMobile, onRenderAvatar, cameraSwitcherProps, announcerString]);
```
JamesBurnside, line 98: From #2502 - I think we should avoid adding to the depedency array that rerenders video gallery as its our most expensive component

JamesBurnside: Instead below instead of `return VideoGalleryMemoized;` you can 
return 
```
<> 
  <Announcer announcementString={announcerString} ariaLive={'polite'} />
  <VideoGalleryMemoized />
</>
```

Ideally we update the useMemo to a `React.memo` but that is a larger change we should keep out of this PR

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.test.tsx
```diff
@@ -0,0 +1,307 @@
...
+import { CallAdapterProvider } from '../adapter/CallAdapterProvider';
+import { MockCallAdapter } from '../MockCallAdapter';
+import { useParticipantChangedAnnouncement } from './MediaGalleryUtils';
+import Enzyme, { ReactWrapper, mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import { act } from 'react-dom/test-utils';
+import { initializeIcons } from '@fluentui/react';
+import { CommunicationUserKind } from '@azure/communication-common';
+
+const strings = {
```
JamesBurnside, line 18: Are these the same as what is in `COMPOSITE_LOCALE_EN_US`? (Do we need to override the locale with these?)

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+  const [announcerString, setAnnouncerString] = useState<string>('');
+  const currentParticipants = useSelector(getRemoteParticipantsConnectedSelector);
+  const [previousParticipants, setPreviousParticipants] = useState<RemoteParticipantState[]>(currentParticipants);
+
+  const resetAnnoucement = (string: string): void => {
+    // Is this really needed?
+    // React does not promise that we will see a DOM update with '' value.
+    // If we're getting announcer to announce the same name twice in sequence, it's probably keyed off of the
+    // actual DOM node, which will update even if we don't reest to '' first.
+    setAnnouncerString('');
```
JamesBurnside, line 59: resolve comment about this being removed, given two setStates execute consecutively there is an incredibly small chance there would ever be a render pass between them

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Generates the announcement string for when a participant joins or leaves a call.
+ */
+export const createAnnouncmentString = (
+  direction: 'joined' | 'left',
+  participants: RemoteParticipantState[],
+  strings: ParticipantChangedAnnouncmentStrings
+): string => {
+  /**
+   * Filter participants into two arrays to put all the unnamed participants at the back of the
```
JamesBurnside, line 93: nit: this can be done with a sort
```ts
const sortedParticipants = participants.sort((p) => !p.displayName);
```

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+   * names being announced.
+   */
+  const unnamedParticipants = participants.filter((p) => p.displayName === undefined);
+  const namedParicipants = participants.filter((p) => p.displayName);
+  const sortedParticipants = namedParicipants.concat(unnamedParticipants);
+
+  /**
+   * if there are only unnamed participants present in the array announce a special unnamed participants
```
JamesBurnside, line 101: Moving comment from #2502, what if there are two unnamed participants and one named participant?

JamesBurnside: Can we get confirmation from a11y team on how to handle multiple participants without set display names

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+    case 3:
+      return _formatString(
+        direction === 'joined' ? strings.threeParticipantJoinedNoticeString : strings.threeParticipantLeftNoticeString,
+        {
+          displayName1: participantNames[0],
+          displayName2: participantNames[1],
+          displayName3: participantNames[2]
+        }
+      );
+  }
```
JamesBurnside, line 137: nit: add a `case 0` to this to return blank or throw an error as the code that follows this assumes we have at least 3 participants and this function never guaranteed that. Alternatively at the start of this function have a `if (participants.length === 0) return;`

change/@internal-react-composites-81715f29-3fde-4133-99e8-70378241e624.json
prprabhu-ms: Probably needs a merge from `main`.

@dmceachernmsft if you think this branch might be old, could you (1) merge from `main` (2) run `rush changelog` again; (3) delete this file because the last step will create duplicate change-files.

prprabhu-ms:
Replacing myself with @edwardlee-msft  as assignee. I've contributed enough to this PR that it doesn't make sense for me to review it.

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -76,20 +81,27 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
...
-        {...videoGalleryProps}
-        localVideoViewOptions={localVideoViewOptions}
-        remoteVideoViewOptions={remoteVideoViewOptions}
-        styles={VideoGalleryStyles}
-        layout="floatingLocalVideo"
-        showCameraSwitcherInLocalPreview={props.isMobile}
-        localVideoCameraCycleButtonProps={cameraSwitcherProps}
-        onRenderAvatar={onRenderAvatar}
-      />
+      <>
```
JamesBurnside, line 84: remove unecessary `<>` elements

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,163 @@
...
+    () => {
+      const currentIds = currentParticipants.map((p) => toFlatCommunicationIdentifier(p.identifier));
+      const previousIds = previousParticipants.map((p) => toFlatCommunicationIdentifier(p.identifier));
+      const whoJoined = currentParticipants.filter(
+        (p) => !previousIds.includes(toFlatCommunicationIdentifier(p.identifier))
+      );
+      const whoLeft = previousParticipants.filter(
+        (p) => !currentIds.includes(toFlatCommunicationIdentifier(p.identifier))
+      );
+      console.log('run');
```
JamesBurnside, line 68: remove console log

JamesBurnside (approved):
Some small comments but overall looks great! Thanks for doing all these iterations to get this in the best state for our a11y users!

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
dmceachernmsft: A11y team is happy with these strings as long as the user has the opportunity to see the list of participants somewhere else in the app.

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.test.tsx
```diff
@@ -0,0 +1,325 @@
...
```
edwardlee-msft, line 18: Are these strings needed? Aren't these the same as the ones in the en-us string.json?

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -0,0 +1,162 @@
...
+      if (whoJoined.length > 0) {
+        resetAnnoucement(createAnnouncmentString('joined', whoJoined, strings));
+      }
+      if (whoLeft.length > 0) {
+        resetAnnoucement(createAnnouncmentString('left', whoLeft, strings));
+      }
+      // Update cached value at the end.
+      setPreviousParticipants(currentParticipants);
+    },
+    // previousParticipants caches the value of `currenParticipants`. We _don't_ want this
```
edwardlee-msft, line 77: nit: currenParticipants -> currentParticipants

PR #2547, opened 2022-11-19
VideoGallery floating local video layout enzyme tests

mgamis-msft:
# What

Added VideoGallery floating local video layout enzyme tests

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3054533

# How Tested

Ran enzyme tests locally

# Process & policy checklist


packages/react-components/src/components/VideoGallery.test.tsx
```diff
@@ -156,6 +156,129 @@ describe('VideoGallery default layout', () => {
...
+describe('VideoGallery floating local video layout', () => {
+  beforeAll(() => {
+    registerIcons({
+      icons: {
+        horizontalgalleryleftbutton: <></>,
+        horizontalgalleryrightbutton: <></>
+      }
+    });
+
+    // Need to mock this because the HorizontalGallery uses this function. But JSDOM does not actually do any
```
prprabhu-ms, line 168: Now that we're adding these mocks twice, for the same reason, let's add a helper that keeps the reasoning etc. in one place.

packages/react-components/src/components/VideoGallery.test.tsx
```diff
@@ -156,6 +156,129 @@ describe('VideoGallery default layout', () => {
...
+
+    const root = mountVideoGalleryWithLocalParticipant({ localParticipant });
+
+    root.setProps({ layout: 'floatingLocalVideo' });
+    const remoteParticipants = Array.from({ length: 10 }, () =>
+      createRemoteParticipant({
+        videoStream: { isAvailable: false, renderElement: createVideoDivElement() }
+      })
+    );
+    root.setProps({ remoteParticipants });
```
prprabhu-ms, line 191: In all of these, do you not need to wrap the `root.setProps` with `act`?
We have a bunch of `useEffect` blocks in the implementation, so I'd expect to see a warning in the console if you don't use `act` (and you're not quite testing the right thing without all those `useEffect` hooks getting a chance to execute).

https://reactjs.org/docs/test-utils.html#act

prprabhu-ms (approved):

packages/react-components/src/components/VideoGallery.test.tsx
anjulgarg: +1

anjulgarg (approved):

PR #2549, opened 2022-11-22
bugfix: Do not show stale errors when a user rejoins a call

prprabhu-ms:
# What

Before this PR, stale errors were sometimes shown in composites via error bar when a user rejoins a call.

| Before | After |
| --- | --- |
| Before | After |

Repro steps:

- Join a group call using `CallComposite` or `CallWithChatComposite`
- Make sure your camera is in use by another application
- Try turning on video
  - You will see an error bar saying "failed to start video". This is expected.
- Leave the call
- Rejoin call from the end call screen.
  - The "failed to start video" error bar will reappear. *This is a bug*.

This PR adds a prop to the `ErrorBar` component, `ignorePremountErrors`. When set, the component does not show errors that occurred before the component was mounted.

The composites set this prop so that errors from previous calls are ignored.

# How Tested

- New unit-tests
- Make sure bug is fixed with the repro steps (see screen-captures above)
- Make sure error bar is _not_ dismissed when opening people / chat pane in mobile view (because the primary call pane is hidden in that case)
- sanity test hold and lobby screens


dmceachernmsft (approved):

packages/react-components/src/components/ErrorBar.test.tsx
```diff
@@ -119,10 +124,47 @@ describe('ErrorBar dismissal with multiple errors', () => {
   });
 });
 
-const mountErrorBarWithDefaults = (): ReactWrapper => {
+describe.only('ErrorBar handling of errors from previous call or chat', () => {
```
prprabhu-ms: remove only!

packages/react-components/src/components/ErrorBar.tsx
```diff
@@ -243,7 +258,11 @@ export const ErrorBar = (props: ErrorBarProps): JSX.Element => {
     [props.activeErrorMessages, dismissedErrors]
   );
 
-  const toShow = errorsToShow(props.activeErrorMessages, dismissedErrors);
+  const toShow = errorsToShow(
+    props.activeErrorMessages,
+    dismissedErrors,
+    props.ignorePremountErrors ? mountTimestamp.current : undefined
```
PorterNan, line 264: Simply curious, is there a reason we decide to use a time stamp to ignore premount error instead of just clear all error in premounting when ignorePremountErrors === true?

PorterNan (approved):

packages/react-components/src/components/ErrorBar.tsx
prprabhu-ms: We can't clear any errors because they are stored in the stateful client. Uni-directional data flow implies that errors only clear when backend clears them (e.g. UFD backed errors)

PR #2550, opened 2022-11-22
moving check into the while loop condition

alkwa-msft:
# What
There is an issue where if you hit "back" in the browser while you are fetching messages it will cause an issue. We are expecting a DOM element so we can check our scrollTop property but it may not exist. We want to see if the DOM element exists before we try to evaluate and pull in more messages.

# Why
This solves an issue if Contoso decides to hit back button but we are using delays to pull in messages when it makes sense.


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -851,7 +851,11 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
       if (onLoadPreviousChatMessages) {
```
PorterNan, line 851: chatScrollDivRef.current && in above line (line 850) could be removed as well

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -860,14 +864,6 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
     }
   }, [numberOfChatMessagesToReload, onLoadPreviousChatMessages]);
 
-  const handleInfiniteScroll = useCallback((): void => {
```
PorterNan, line 863: Hmm, totally forget why we added another function on top of fetchNewMessageWhenAtTop, looks like just a small check here, good to remove it

PorterNan (approved):
Thanks for catching this async call bug, this was a bug hiding there when all logic seems okay.

This bug is a good lesson that we should be careful about ref.current + async call loop - because the value could be changed somewhere else when looping

edwardlee-msft (approved):

PR #2552, opened 2022-11-23
[Call Readiness] Unsupported browser version UI

dmceachernmsft:
# What

Add new UI for showing in unsupportedBrowser version in components and composites.
# Why

image
image
Allows user to know their browser version is out of date.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3054307
# How Tested

Added new components to storybook and tested composite changes in browser stack.


packages/react-components/src/components/UnsupportedBrowser.tsx
```diff
@@ -43,38 +33,15 @@ export interface UnsupportedBrowserProps {
...
-        >
-          {strings.moreHelpLink}
-        </Link>
-      )}
-    </Stack>
-  );
-};
-
 /**
  * UI to display to the user that the browser they are using is not supported by calling application.
```
prprabhu-ms, line 37: nit: calling is not an "application".

Perhaps say "not supported by the Azure Communication Services calling service"?

packages/react-components/src/components/UnsupportedBrowserVersion.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * Strings for UnsupportedBrowser component
+ *
+ * @beta
+ */
+export interface UnsupportedBrowserVersionStrings {
+  /** String for the primary text */
+  primaryText: string;
+  /** String for the secondary text */
+  secondaryText: string;
+  /** String for the help link */
```
prprabhu-ms, line 20: Is this the URL that the link points to, or the text that is shown for the link?

If it is the URL, perhaps this should be call `moreHelpUrl`.

And in either case, the docstring could be more helpful.

packages/react-components/src/components/UnsupportedBrowserVersion.tsx
```diff
@@ -0,0 +1,47 @@
...
+  moreHelpLink: string;
+}
+
+/**
+ * props for UnsupportedBrowserVersion UI
```
prprabhu-ms, line 25: nit: `Props for {@link UnsupporedBrowserVersion} component.`

packages/react-components/src/components/UnsupportedEnvironment.tsx
```diff
@@ -0,0 +1,64 @@
...
+  mainTextStyles,
+  secondaryTextStyles
+} from './styles/UnsupportedEnvironment.styles';
+import { UnsupportedBrowserStrings } from './UnsupportedBrowser';
+import { UnsupportedBrowserVersionStrings } from './UnsupportedBrowserVersion';
+
+/**
+ * props for UnsupportedBrowser UI
+ *
+ * @beta
```
prprabhu-ms, line 23: This should be `@private`.

A component in the API that accepts `UnsupportedBrowserStrings | UnsupportedBrowserVersionStrings` doesn't make sense to me. This reads more like a `@private` helper component.

packages/react-components/src/components/UnsupportedEnvironment.tsx
```diff
@@ -0,0 +1,64 @@
...
+  /** String overrides for the component */
+  strings: UnsupportedBrowserStrings | UnsupportedBrowserVersionStrings;
+}
+
+/* @conditional-compile-remove(unsupported-browser) */
+const UnsupportedEnvironmentContainer = (props: UnsupportedEnvironmentProps): JSX.Element => {
+  const { onTroubleshootingClick, strings } = props;
+  return (
+    <Stack styles={containerStyles}>
+      <Icon styles={iconStyles} iconName="UnsupportedBrowserWarning" data-ui-id="unsupportedBrowserIcon"></Icon>
```
prprabhu-ms, line 37: nit: `data-ui-id` should be `unsupportedEnvironmentIcon`.

packages/react-components/src/components/UnsupportedEnvironment.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <Text styles={secondaryTextStyles}>{strings.secondaryText}</Text>
+      {onTroubleshootingClick && (
+        <Link
+          styles={linkTextStyles}
+          onClick={() => {
+            onTroubleshootingClick();
+          }}
+          data-ui-id="unsupportedBrowserLink"
+        >
+          {strings.moreHelpLink}
```
prprabhu-ms, line 48: This prop's name `moreHelpLink` suggests that this is link, but the `onTroubleshootingClick` handler name suggests (and in reality Contoso could use it to) that it is really just a button that takes _some_ action.

Consider renaming `moreHelpLink` to `moreHelpText` or even `moreHelpButtonText`.

packages/react-components/src/components/UnsupportedEnvironment.tsx
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * props for UnsupportedBrowser UI
+ *
+ * @beta
+ */
+export interface UnsupportedEnvironmentProps {
+  /** Handler to perform a action when the help link is actioned */
+  onTroubleshootingClick?: () => void;
+  /** String overrides for the component */
+  strings: UnsupportedBrowserStrings | UnsupportedBrowserVersionStrings;
```
prprabhu-ms, line 29: `UnupportedBrowserStrings` and `UnsupportedBrowserVersionStrings` are, in fact, the same shape (have the same fields).

Assuming this is a `@private` component (see comment above), you can define a new type

```TypeScript
interface UnsupportedEnvironmentStrings {
  primaryText: string;
  ...
}
```

and use that. When `UnsupportedBrowser` and `UnsupportedBrowserVersion` forward their own `strings` prop, it will _just work_ because the shape of all those interfaces is the same.

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -249,6 +249,11 @@
     "secondaryText": "Please join this call using a compatible browser.",
     "moreHelpLink": "More help"
   },
+  "UnsupportedBrowserVersion": {
+    "primaryText": "Browser version not supported",
+    "secondaryText": "Please update your browser and then try joining the call again.",
+    "moreHelpLink": "See compatibility requirements"
```
prprabhu-ms, line 255: This default text suggests that the link will point to ACS help docs.

But there is no default implementation for `onTroubleshootingClick()` that does that.

I think we should addd default behavior for that callback so that the component works out of the box, and also so that this default text makes sense.

dmceachernmsft: So the goal here with this link is for contoso to provide their own page with this information since we don't want the component sending end users to the technical documentation for ACS. We don't provide a default behavior for the link and hide it should the callback not be provided.

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -148,7 +148,10 @@ export const getCallCompositePage: GetCallCompositePageFunction = (
   features?
 ): CallCompositePage => {
   /* @conditional-compile-remove(unsupported-browser) */
-  if (features?.unsupportedEnvironment && environmentInfo?.isSupportedBrowser === false) {
+  if (
```
prprabhu-ms, line 151: For cleanliness, I'd extract a trampoline function here:

```TypeScript

function isUnsupportedEnvironmentTrampoline(): boolean {
  return features?.unsupportedBrowser && (environmentInfo?.isSupportedBrowser === false || environmentInfo?.isSupportedBrowserVersion === false);
// No way to detect unsupported environment in stable build, yet.
return false;
}
```

packages/react-composites/src/composites/CallComposite/pages/UnsupportedBrowser.tsx
```diff
@@ -16,7 +14,6 @@ import { useLocale } from '../../localization';
  */
```
prprabhu-ms, line 14: Should this be @private instead of @internal (@internal is for stuff you want to use across packlets, but not export in the public API. @private is for stuff restricted to a single packlet)

packages/react-composites/src/composites/CallComposite/pages/UnsupportedBrowser.tsx
```diff
@@ -34,9 +30,7 @@ export const UnsupportedBrowserPage = (props: UnsupportedBrowserPageProps): JSX.
   /* @conditional-compile-remove(unsupported-browser) */
   const unsupportedBrowserVersionStrings = locale.component.strings.UnsupportedBrowserVersion;
 
-  /* @conditional-compile-remove(unsupported-browser) */
   let pageElement: JSX.Element | undefined;
-  /* @conditional-compile-remove(unsupported-browser) */
   if (!environmentInfo?.isSupportedBrowser) {
     pageElement = (
       <UnsupportedBrowser onTroubleshootingClick={onTroubleshootingClick} strings={unsupportedBrowserStrings} />
```
prprabhu-ms, line 36: Doesn't this complain that `unsupportedBrowserStrings` is undefined?

I expected that you'll need a `getStringsTrampoline()` function that returns empty strings for this to work.

packages/react-composites/src/composites/CallComposite/pages/UnsupportedBrowser.tsx
```diff
@@ -22,22 +25,32 @@ export type UnsupportedBrowserPageProps = {
  */
 export const UnsupportedBrowserPage = (props: UnsupportedBrowserPageProps): JSX.Element => {
   /* @conditional-compile-remove(unsupported-browser) */
-  const { onTroubleshootingClick } = props;
+  const { onTroubleshootingClick, environmentInfo } = props;
```
dmceachernmsft, line 28: All of the CC is needed in this file because the `EnvironmentInfo` type is from the beta calling SDK only so will cause errors in the stable flavor if present. This trickles down to all the other CC elements as well.

dmceachernmsft:
there is extra CC in the composites because of the Calling SDK's type for `EnvironmentInfo` since it is not part of their stable package.

packages/react-components/src/components/UnsupportedBrowserVersion.tsx
```diff
@@ -0,0 +1,47 @@
...
+  moreHelpText: string;
+}
+
+/**
+ * props for {@link UnsupportedBrowserVersion} UI
+ *
+ * @beta
+ */
+export interface UnsupportedBrowserVersionProps {
+  /** Handler to perform a action when the help link is actioned */
```
edwardlee-msft, line 30: nit: a -> an

packages/react-components/src/components/UnsupportedBrowserVersion.tsx
```diff
@@ -0,0 +1,47 @@
...
+  strings: UnsupportedBrowserVersionStrings;
+}
+
+/**
+ * UI to display to the user that the browser they are using is not supported by Azure Communications Calling service.
+ *
+ * @beta
+ */
+export const UnsupportedBrowserVersion = (props: UnsupportedBrowserVersionProps): JSX.Element => {
+  /* @conditional-compile-remove(unsupported-browser) */
```
edwardlee-msft, line 42: Does this need to be conditionally compiled when the export is tagged as @beta?

packages/storybook/stories/INTERNAL/UnsupportedBrowser/UnsupportedBrowser.stories.tsx
```diff
@@ -34,16 +36,20 @@ const getDocs: () => JSX.Element = () => {
     <Stack>
       <SingleLineBetaBanner />
       <Title>Unsupported Browser</Title>
-      <Description>Component to display help information when the app is loaded in a unsuppported browser.</Description>
+      <Description>
+        Component to display help information when the app is loaded in a unsuppported environment. Shows different
```
edwardlee-msft, line 40: nit: a -> an
"... is loaded in an unsupported environment"

packages/storybook/stories/INTERNAL/UnsupportedBrowser/UnsupportedBrowser.stories.tsx
```diff
@@ -34,16 +36,20 @@ const getDocs: () => JSX.Element = () => {
...
+        screens based on what the environment info from the calling SDK is reporting.
```
edwardlee-msft, line 41: nit: change to "based on the environment info the calling SDK is reporting"
possibl change calling SDK to "Azure calling communication service" to keep consistent with your comments in the code.

packages/storybook/stories/INTERNAL/UnsupportedBrowser/UnsupportedBrowser.stories.tsx
```diff
@@ -34,16 +36,20 @@ const getDocs: () => JSX.Element = () => {
...
+      </Description>
       <Heading>Using in a Modal</Heading>
       <Description>This component while used in a modal here can also be used as it's own page.</Description>
```
edwardlee-msft, line 44: nit:
This component can also be used as it's own page or as in a modal as shown here.

packages/react-composites/src/composites/CallComposite/pages/UnsupportedBrowser.tsx
mgamis-msft: +1

packages/react-composites/tests/browser/common/constants.ts
```diff
@@ -29,7 +29,7 @@ export const IDS = {
   holdPage: 'hold-page',
   moreButton: 'call-with-chat-composite-more-button',
   callPage: 'call-page',
-  unsupportedBrowserIcon: 'unsupportedBrowserIcon',
-  unsupportedBrowserLink: 'unsupportedBrowserLink',
+  unsupportedEnvironmentIcon: 'unsupportedEnvironmentIcon',
```
mgamis-msft, line 32: Minor: I can't help wanting to make this consistent with the other id names with 'unsupported-environment-icon' and  'unsupported-environment-link'

packages/react-components/src/components/UnsupportedBrowserVersion.tsx
```diff
@@ -0,0 +1,47 @@
...
+export interface UnsupportedBrowserVersionProps {
+  /** Handler to perform a action when the help link is actioned */
+  onTroubleshootingClick?: () => void;
+  /** String overrides for the component */
+  strings: UnsupportedBrowserVersionStrings;
+}
+
+/**
+ * UI to display to the user that the browser they are using is not supported by Azure Communications Calling service.
+ *
```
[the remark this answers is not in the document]
dmceachernmsft: I missed updating this comment will do that!

edwardlee-msft (approved):

mgamis-msft (approved):

PR #2554, opened 2022-11-23
[Call Readiness] unsupported operating system UI

dmceachernmsft:
# What

Add UI for when the end user joins with a unsupported operating system.
image
image
# Why

Allows contoso to inform the user when they are joining from a unsupported platform.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3054307
# How Tested

Will add snapshot tests in a following PR. BrowserStack does not have a un-orthodox operating system to cover this case.


packages/react-components/src/components/UnsupportedOperatingSystem.tsx
```diff
@@ -0,0 +1,80 @@
...
+  onTroubleshootingClick?: () => void;
+  /** String overrides for the component */
+  strings: UnsupportedOperatingSystemStrings;
+}
+
+/* @conditional-compile-remove(unsupported-browser) */
+const UnsupportedOperatingSystemContainer = (props: UnsupportedOperatingSystemProps): JSX.Element => {
+  const { onTroubleshootingClick, strings } = props;
+  return (
+    <Stack styles={containerStyles}>
```
JamesBurnside, line 50: this is all the same as UnsupportedBrowser, they should all use the same base component. See how all DomainPermissions SKUs use the same underlying DomainPermissions scaffolding component

packages/react-components/src/components/UnsupportedOperatingSystem.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <Icon styles={iconStyles} iconName="UnsupportedBrowserWarning" data-ui-id="unsupportedBrowserIcon"></Icon>
```
JamesBurnside, line 51: Should these also be update
```suggestion
      <Icon styles={iconStyles} iconName="UnsupportedOperatingSystemWarning" data-ui-id="unsupportedOSIcon"></Icon>
```

packages/react-components/src/components/UnsupportedOperatingSystem.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /** String for the primary text */
+  primaryText: string;
+  /** String for the secondary text */
+  secondaryText: string;
+  /** String for the help link */
+  moreHelpLink: string;
+}
+
+/**
+ * props for UnsupportedOperatingSystem UI
```
JamesBurnside, line 35: nit: props -> Props

packages/react-components/src/components/UnsupportedOperatingSystem.tsx
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/* @conditional-compile-remove(unsupported-browser) */
```
JamesBurnside, line 4: these all shouldn't be needed. You can create the pure ui component but just conditional compile the export line

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -148,9 +148,12 @@ export const getCallCompositePage: GetCallCompositePageFunction = (
   features?
 ): CallCompositePage => {
   /* @conditional-compile-remove(unsupported-browser) */
+  console.log(environmentInfo);
```
JamesBurnside, line 151: remove

packages/storybook/stories/INTERNAL/UnsupportedBrowser/snippets/UnsupportedEnvironmentModal.snippet.tsx
```diff
@@ -11,6 +11,7 @@ export const UnsupportedEnvironmentModals: () => JSX.Element = () => {
   const locale = useLocale().strings.UnsupportedBrowser;
   const [modal1Open, setModal1Open] = useState<boolean>(false);
   const [modal2Open, setModal2Open] = useState<boolean>(false);
+  const [modal3Open, setModal3Open] = useState<boolean>(false);
```
JamesBurnside, line 14: modal1/2/3 should have names and not 1/2/3

packages/react-composites/src/composites/CallComposite/pages/UnsupportedBrowser.tsx
```diff
@@ -33,11 +33,20 @@ export const UnsupportedBrowserPage = (props: UnsupportedBrowserPageProps): JSX.
...
   /* @conditional-compile-remove(unsupported-browser) */
-  if (!environmentInfo?.isSupportedBrowser) {
+  if (!environmentInfo?.isSupportedPlatform) {
+    pageElement = (
+      <UnsupportedOperatingSystem
+        onTroubleshootingClick={onTroubleshootingClick}
+        strings={unsupportedOperatingSystemStrings}
+      />
+    );
+  } else if (!environmentInfo?.isSupportedBrowser) {
```
JamesBurnside, line 49: what does L63 show:
`return <Stack styles={containerStyles}>{pageElement}</Stack>;`
?

Can we ever hit that line? If not instead do
```
else {
  throw new Error(...
}
```

packages/react-composites/tests/browser/call/hermetic/UnsupportedBrowserPage.test.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+import type { EnvironmentInfo } from '@azure/communication-calling';
```
anjulgarg, line 4: Is this file not throwing errors in stable?
Due to unused imports

dmceachernmsft: it is not because one of the helper functions in the test is using the type so this should be fine.

packages/react-composites/src/composites/CallComposite/pages/UnsupportedBrowser.tsx
dmceachernmsft: this will return the different unsupportedEvironment UI to the page. I included the error since if the user hits the case that they get to this page without environmentInfo for whatever reason we should throw an error letting them know.

packages/react-components/src/components/UnsupportedOperatingSystem.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * UI to display to the user that the operating system they are using is not supported by Azure Communications Calling service.
+ *
+ * @beta
+ */
+export const UnsupportedOperatingSystem = (props: UnsupportedOperatingSystemProps): JSX.Element => {
+  const { onTroubleshootingClick, strings } = props;
+  return <UnsupportedEnvironment onTroubleshootingClick={onTroubleshootingClick} strings={strings} />;
+  return <></>;
```
JamesBurnside, line 42: remove unreachable line (I assume left over from an old conditional compile line?)

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -254,6 +254,11 @@
     "secondaryText": "Please update your browser and then try joining the call again.",
     "moreHelpText": "See compatibility requirements"
   },
+  "UnsupportedOperatingSystem": {
+    "primaryText": "Operating system not supported",
+    "secondaryText": "Please join this call using a device with a compatible operating system.",
+    "moreHelpText": "See compatibility requirements"
```
JamesBurnside, line 260: highlight this is a button/link:
```suggestion
    "moreHelpLinkText": "See compatibility requirements"
```
Unless we specifically do not do that throughtout the strings file, consistency is better

packages/react-composites/tests/browser/call/hermetic/UnsupportedBrowserPage.test.ts
```diff
@@ -10,16 +11,22 @@ import { buildUrlWithMockAdapter, defaultMockCallAdapterState, test } from './fi
 /* @conditional-compile-remove(unsupported-browser) */
 test.describe('unsupportedBrowser page tests', async () => {
   test('unsupportedBrowser displays correctly without a help link', async ({ page, serverUrl }) => {
-    await page.goto(buildUrlWithMockAdapter(serverUrl, defaultMockUnsupportedBrowserPageState()));
+    const state = defaultMockUnsupportedBrowserPageState();
+    state.environmentInfo = setMockEnvironmentInfo(true, false, false);
```
JamesBurnside, line 15: avoid many unnamed arguments particularly when they are booleans, reading this line what is `true, false, false`?
Instead use named params:
```
setMockEnvironmentInfo({ wellNamedArg1: true, wellNamedArg2: false, wellNamedArg3: false });
```

JamesBurnside (approved):

anjulgarg (approved):

PR #2556, opened 2022-11-24
VideoTile 'onLongTouch' support

anjulgarg:
# What
VideoTile 'onLongTouch' support

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/VideoTile.test.tsx
+++ b/packages/react-components/src/components/VideoTile.test.tsx
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
 import { initializeIcons } from '@fluentui/react';
 import Enzyme, { mount } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
--- a/packages/react-components/src/components/VideoTile.test.tsx
+++ b/packages/react-components/src/components/VideoTile.test.tsx
@@ -1,0 +1,25 @@
+import { initializeIcons } from '@fluentui/react';
+import Enzyme, { mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import React from 'react';
+import { VideoTile } from './VideoTile';
+import { act } from 'react-dom/test-utils';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('VideoTile', () => {
+  beforeAll(() => {
+    initializeIcons();
+  });
+
+  test('onLongTouch should trigger callback', async () => {
+    const mockCallback = jest.fn();
+    const wrapper = mount(<VideoTile onLongTouch={mockCallback} />);
+    await act(async () => {
+      wrapper.simulate('touchstart');
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+      wrapper.simulate('touchend');
+    });
+    expect(mockCallback).toBeCalledTimes(1);
+  });
+});
--- a/packages/react-components/src/components/VideoTile.test.tsx
+++ b/packages/react-components/src/components/VideoTile.test.tsx
@@ -15,6 +15,7 @@ describe('VideoTile', () => {
     initializeIcons();
   });
 
+  /* @conditional-compile-remove(pinned-participants) */
   test('onLongTouch should trigger callback', async () => {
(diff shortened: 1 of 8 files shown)


packages/react-components/src/components/utils/useLongPress.ts
```diff
@@ -29,38 +29,48 @@ export default function useLongPress(
...
-    if (isLongPress) {
-      onLongPress();
-      return;
+    if (!isLongPress) {
+      onClick();
     }
   }
 
   function handleOnKeyDown(): void {
+    if (isMobile) {
```
JamesBurnside, line 46: should `handleOnKeyDown` still work if mobile has a bluetooth keyboard plugged in ?

JamesBurnside: This logic seems to be there to make isMobile only trigger for touch events is that right?
Instead would we pass the start mechanism into `startPressTimer` e.g.
```
startPressTimer('key' | 'mouseClick' | 'touch');
```
then the startPressTimer can handle accordingly?
```
...
setTimeout(() => {
  onLongPress();
  if (pressMechanism === 'touch') {
    onLongTouch();
  } 
}
```
?

packages/react-components/src/components/utils/useLongPress.ts
```diff
@@ -29,38 +29,48 @@ export default function useLongPress(
...
+      return;
+    }
     if (action) {
       setAction(false);
       startPressTimer();
```
JamesBurnside, line 51: (existing bug in this hook)
I think this means we could start two timers, if I mouseDown on something and before mouseUp I activate the keyDown it will trigger a new timer with somewhat undefined behavior on the previous timer

anjulgarg: This scenarios will end up calling onLongPress twice. I think that's the expected behaviour given that the user explicitly provided 2 inputs.

packages/react-components/src/components/utils/useLongPress.ts
anjulgarg: Given that we only have onClick and onLongPress as functionalities in this hook.
I think changing isMobile to touchEventsOnly makes more sense.

packages/react-components/src/components/VideoTile.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { initializeIcons } from '@fluentui/react';
+import Enzyme, { mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import React from 'react';
+import { VideoTile } from './VideoTile';
+import { act } from 'react-dom/test-utils';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('VideoTile', () => {
```
JamesBurnside, line 13: nice test! ❤️

packages/react-components/src/components/utils/useLongPress.ts
```diff
@@ -69,10 +79,6 @@ export default function useLongPress(
   }
 
   function handleOnTouchEnd(): void {
-    if (isMobile) {
-      isLongPress ? onLongPress() : onClick();
-    }
-
     timerRef.current && clearTimeout(timerRef.current);
   }
```
JamesBurnside, line 84: My gut reaction is the result of this hook should be memoized (which annoyingly would involve memoizing each of the functions in this hook)
... do you think it needs memoized?

JamesBurnside (approved):

edwardlee-msft (approved):

packages/react-components/src/components/utils/useLongPress.ts
```diff
@@ -6,21 +6,20 @@ import { useMemo, useRef, useState, useCallback, useEffect } from 'react';
...
-    onClick: () => void;
-    onMouseDown: () => void;
-    onMouseUp: () => void;
-    onTouchStart: () => void;
-    onTouchEnd: () => void;
-    onKeyDown: () => void;
-    onKeyUp: () => void;
-  };
+export default function useLongPress(props: {
+  onClick?: () => void;
```
edwardlee-msft, line 10: nit: move optional props to the bottom.
onLongPress
onClick?
touchEventsOnly?

PR #2557, opened 2022-11-29
Use IContextualMenuProps for VideoTile contextualMenu

anjulgarg:
# What
Use IContextualMenuProps for VideoTile contextualMenu
Screenshot 2022-11-28 at 4 54 20 PM

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/VideoTile.test.tsx
+++ b/packages/react-components/src/components/VideoTile.test.tsx
@@ -1,7 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { initializeIcons } from '@fluentui/react';
+import { initializeIcons, registerIcons } from '@fluentui/react';
+/* @conditional-compile-remove(pinned-participants) */
+import { IconButton } from '@fluentui/react';
 import Enzyme, { mount } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 import React from 'react';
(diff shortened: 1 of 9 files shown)


packages/react-components/src/components/VideoTile.test.tsx
```diff
@@ -28,4 +35,22 @@ describe('VideoTile', () => {
     /* @conditional-compile-remove(pinned-participants) */
     expect(mockCallback).toBeCalledTimes(1);
   });
+
+  test('VideoTile shows more button when contextualMenu is set', async () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const videoTileProps = { displayName: 'John Doe', contextualMenu: { items: [] } } as any;
```
mgamis-msft, line 41: Just curious why we show an icon button if an empty array of items is provided? What does it look like when you click the menu button?

mgamis-msft (approved):

packages/react-components/src/components/VideoTile.test.tsx
anjulgarg: Since we are now allowing Contoso devs to inject the ContextualMenu, we can't enforce a check using an internal object inside that type. Hence, it is unto Contoso to ensure that `contextualMenu` is undefined if they don't want the icon to show.

edwardlee-msft (approved):

PR #2558, opened 2022-11-29
Fix: Local video not showing to local user but still broadcasts to others in the call

JamesBurnside:
# What
Fix stream utils logic:
* `createView` when in "stopping" should flip state back to "rendering"
* `disposeView` when marking the stream as "stopping" should not mark the renderer as undefined nor try dispose it as it is not performing cleanup here, only signifying that `createView` should perform the cleanup when the view has finished being created
* `disposeView` should no-op when the state is "NotRendered" 

## Next Steps
This fix is the smallest delta suitable for a hotfix and fixes the severity of the issue as the local user can always see their feed. For this to be fully complete we need to also:
* Fix VideoGallery trying to render idle and Connecting users (this caused the gallery layout shift that triggered the local video tile to re-render multiple times destroying and recreating the stream)
* Memoize local participant - this is currently re-rendering anytime a remote participant changes
* Apply same fixes to unparented views and remote views to prevent memory leaks
* Show a fallback placeholder in the local video tile that the user's local video is on if the stream fails to be created for whatever reason

# Why
We have a bad (rare) bug where the local video tile would show the suers video as off when it was still broadcasting to everyone else in the call. This fix fixes the race condition logic that caused the stream to not show to the local user.The reason this happened is:
* createView is called, enters async part where view is being created
* destroyView is called, the view is being created so destroyView marks the stream as stopping
* createView is then called again but the state was in stopping so no-oped
* then destroyView was called again incorrectly setting the state to notRendered
* createView is called again (again no-ops)
* the original createView completes but sees the stream should be cleaned and cleans up the local stream - when it should have continued creating the stream because the last call was to createView

More details: https://skype.visualstudio.com/SPOOL/_workitems/edit/3006479

# How Tested
No longer repros locally with the repro scenario:
* Start Teams meeting with camera on
* Join meeting with ACS client A with camera on
* Join just the lobby of the meeting with ACS Client B with no camera on
* Observe: Client A can suddenly no longer see their camera but Teams user can see their camera feed.


packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -217,13 +217,15 @@ async function createViewLocalVideo(
...
   }
 
+  // "Stopping" only happens if the stream was in "rendering" but `disposeView` was called.
+  // Now `createView` has been re-called when can flip the state back to "rendering".
   if (renderInfo.status === 'Stopping') {
     _logEvent(callingStatefulLogger, {
       name: EventNames.LOCAL_STREAM_STOPPING,
       level: 'warning',
-      message: 'LocalVideoStream is in the middle of stopping.'
+      message: 'LocalVideoStream was is in the middle of stopping. Resetting state to "Rendering".'
```
anjulgarg, line 226: was is?

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -217,13 +217,15 @@ async function createViewLocalVideo(
     return;
   }
 
+  // "Stopping" only happens if the stream was in "rendering" but `disposeView` was called.
+  // Now `createView` has been re-called when can flip the state back to "rendering".
```
anjulgarg, line 221: nit:
// Now that `createView` has been re-called, we can flip the state back to "rendering".

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -468,21 +470,46 @@ function disposeViewLocalVideo(context: CallContext, internalContext: InternalCa
...
+  if (renderInfo.status === 'NotRendered') {
+    _logEvent(callingStatefulLogger, {
+      name: EventNames.LOCAL_STREAM_ALREADY_DISPOSED,
+      level: 'info',
+      message: 'LocalVideoStream is already disposed.'
+    });
+    return;
+  }
+
+  // Status is already marked as "stopping" so we can exit early here. This is because stopping only occurs
```
anjulgarg, line 483: This behaviour is so weird. One would assume "Stopping" occurs only when the stream is disposed or stopped essentially. Is this a documented bhevaiour by calling sdk?

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -468,21 +470,46 @@ function disposeViewLocalVideo(context: CallContext, internalContext: InternalCa
...
+  // "stopping" without performing any cleanup. This will tell the `createView` method that it should stop
+  // rendering and clean up the state once the view has finished being created.
+  if (renderInfo.status === 'Rendering') {
+    _logEvent(callingStatefulLogger, {
+      name: EventNames.REMOTE_STREAM_STOPPING,
+      level: 'info',
+      message: 'Remote stream is still rendering. Changing status to stopping.',
+      data: streamLogInfo
+    });
+    internalContext.setLocalRenderInfo(callId, renderInfo.stream, 'Stopping', renderInfo.renderer);
```
anjulgarg, line 506: Looking at the createView method, wouldn't this set the state to "Rendering" again?

JamesBurnside: If it's called again, yep!

For context for this message, their are four renderInfo states: `NotRendered`, `Rendering`, `Stopping`, `Rendered`.
NotRendered is the starting state or end of the cleanup state
Rendering is set when createView is asynchronously creating the render stream
Stopping is set when disposeView is called, but createView is currently asynchronously creating the view so it can't be cleaned up right now.
Rendered is the happy path when the render has finished.

How DisposeView works:
- will try to dispose the view.
- if the view is NotRendered or Stopping, it can no-op and exit early  -- nothing can be done
- if the view is Rendering, it can't dispose the view right now, so it sets the state to Stopping (the createView then checks for this when it finishes creating the view and performs the cleanup instead of disposeView)
- if the view is rendered, this is the happy path and it will do the cleanup

How createView works:
- if the view is Rendered or Rendering it can simply no-op -- nothing can be done
- if the view is in Stopping, it is only in this state because another createView is asynchronously waiting for the view to be completed and a disposeView was subsequently called that set it to stopping. Now that createView has been called again it can change it back to Rendering safely.
- If the view is in NotRendered this is the happy path and it can create the view
	- Now it creates the view but this is asynchronous
	- Once it completes the view if the state is stopping, it should clean up and dispose of the newly created view
	- Otherwise it's happy path time again and can set the state as Rendered 


now for your question
> Looking at the createView method, wouldn't this set the state to "Rendering" again?

Yes! If, and only if, createView is called again before the previous createView's asynchronous view creation has completed

anjulgarg (approved):

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -468,21 +470,46 @@ function disposeViewLocalVideo(context: CallContext, internalContext: InternalCa
...
+    return;
   }
 
+  // Else the state must be in the "Rendered" state, so we can dispose the renderer and clean up the state.
+  internalContext.setLocalRenderInfo(callId, renderInfo.stream, 'NotRendered', undefined);
```
PorterNan, line 511: So previously, we never set an existing renderInfo to NotRendered again anywhere

PorterNan: Ohh, noticed that in the else closure of status==='rendering'

packages/calling-stateful-client/src/StreamUtils.ts
PorterNan, line 506: When a stream is rendering - then when we try to dispose it, no-ops will be performed except a state change - will stream disposing will actually happen in the end somewhere?

JamesBurnside: It is the createView's job to cleanup, when createView finishes the async step it checks if the state was marked Stopping and then does the cleanup (see v. long comment on how it works: https://github.com/Azure/communication-ui-library/pull/2558#discussion_r1035331944)

PorterNan: If someone calls createView then calls disposeView when it is creating, will the view gets disposed in the end without calling createView

JamesBurnside: Yes, the first createView call that is stuck awaiting in the async view creation will perform the cleanup

PorterNan (approved):

PR #2565, opened 2022-12-01
Make VideoStreamRendererView declarative so that updateScalingMode triggers state change

anjulgarg:
# What
Make VideoStreamRendererView declarative so that updateScalingMode triggers state change

# Why

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/StatefulCallClient.ts
```diff
@@ -356,14 +357,21 @@ export const createStatefulCallClientWithDeps = (
...
       callId: string | undefined,
       participantId: CommunicationIdentifier | undefined,
       stream: LocalVideoStreamState | RemoteVideoStreamState,
       options?: CreateViewOptions
     ): Promise<CreateViewResult | undefined> => {
       const participantIdKind = participantId ? getIdentifierKind(participantId) : undefined;
-      return createView(context, internalContext, callId, participantIdKind, stream, options);
+      const result = await createView(context, internalContext, callId, participantIdKind, stream, options);
+      // We only need to declaratify the VideoStreamRendererView object for remote participants.
+      // Because the updateScalingMode only needs to be called on remote participant stream views.
```
mgamis-msft, line 369: mega nit: should be in one sentence

packages/calling-stateful-client/src/StatefulCallClient.ts
```diff
@@ -356,14 +357,21 @@ export const createStatefulCallClientWithDeps = (
...
+      if ('id' in stream && callId && participantId && result) {
+        const participantKey = toFlatCommunicationIdentifier(participantId);
+        result.view = videoStreamRendererViewDeclaratify(result?.view, context, callId, participantKey, stream.id);
```
mgamis-msft, line 372: mega nit: `result?.view` -> `result.view`

packages/calling-stateful-client/src/VideoStreamRendererViewDeclarative.ts
```diff
@@ -0,0 +1,52 @@
...
+        return Reflect.get(target, prop);
+    }
+  }
+}
+
+/**
+ * Creates a declarative Incoming Call by proxying IncomingCall using ProxyIncomingCall.
+ * @param incomingCall - IncomingCall from SDK
+ * @returns proxied IncomingCall
+ */
```
mgamis-msft, line 42: Comment should be changed

mgamis-msft (approved):

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
```diff
@@ -296,9 +296,7 @@ export const createDefaultCommonCallingHandlers = memoizeOne(
         // Hardcoded `scalingMode` since it is highly unlikely that CONTOSO would ever want to use a different scaling mode for screenshare.
         // Using `Crop` would crop the contents of screenshare and `Stretch` would warp it.
         // `Fit` is the only mode that maintains the integrity of the screen being shared.
-        createViewResult = await callClient.createView(call.id, participant.identifier, screenShareStream, {
-          scalingMode: 'Fit'
-        });
+        await callClient.createView(call.id, participant.identifier, screenShareStream, { scalingMode: 'Fit' });
```
PorterNan, line 299: Why do we change this line? I guess the view result is still useful to manage screenShareStream life cycle

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -497,6 +503,26 @@ export class CallContext {
     });
   }
 
+  public setRemoteVideoStreamViewScalingMode(
```
JamesBurnside, line 506: Question:
As Contoso I can do:
```
const view = statefulCallClient.createView()
view.updateScalingMode()
```

when do I call `setRemoteVideoStreamViewScalingMode` over `view.updateScalingMode`?

JamesBurnside: Oh different file and class, never mind!

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
anjulgarg: When a remote participant is screen sharing, their video stream view gets overridden here.
For now, the priority is to use their video stream view.
We need to find a better way to consume both stream and screen share view objects

packages/calling-stateful-client/src/VideoStreamRendererViewDeclarative.ts
```diff
@@ -0,0 +1,52 @@
...
+    this._callId = callId;
+    this._participantId = participantId;
+    this._streamId = _streamId;
+  }
+
+  public get<P extends keyof VideoStreamRendererView>(target: VideoStreamRendererView, prop: P): any {
+    switch (prop) {
+      case 'updateScalingMode': {
+        return async (args: Parameters<VideoStreamRendererView['updateScalingMode']>) => {
+          const result = await target.updateScalingMode(...args);
```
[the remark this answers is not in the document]
anjulgarg: Correct.

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
PorterNan: Yea, I can see some concerns to make it easier here
This is a public facing api, but we change the behavior a bit here, I would still consider this is a behavior breaking change from our side(although I doubt if there is anyone use this, but let's not lie to breaking change definition)

One way I can think of is not changing this, but add an if-condition in parent caller - it's easy to determine if a stream is a remote video stream or just a screen share stream - if it is a screen share stream, don't override it

PorterNan: BTW, @JamesBurnside do you remember why do we add a return value here, I guess there is a final consumer of this return value, but I just can't find it in the code

PorterNan (approved):

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
anjulgarg: Removed this change from this PR
This will be coming along with it's usage in another PR

PR #2573, opened 2022-12-05
Change args type to teams specific

PorterNan:
# What

Change CTE args type according to ACS Internal review comments
Change name of factory function according to ACS Internal review comments

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -918,14 +951,27 @@ const useAzureCommunicationCallAdapterGeneric = <
...
-          credential,
-          displayName,
-          locator,
-          userId,
-          /* @conditional-compile-remove(PSTN-calls) */ alternateCallerId,
-          /* @conditional-compile-remove(rooms) */ options
-        });
+        let newAdapter: AdapterType;
+        if (type === 'AzureCommunication') {
+          if (!('displayName' in args) || !args?.displayName) {
```
JamesBurnside, line 956: this was already checked on L937 - is this needed?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -918,14 +951,27 @@ const useAzureCommunicationCallAdapterGeneric = <
...
+            return;
```
JamesBurnside, line 957: we already checked for displayName on L937, I assume this check is to satiate typescript, if so can this be throw new Error instead of return as this code should be unreachable

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -848,42 +878,42 @@ export const createAzureCommunicationCallAdapter = async ({
...
+  ? CallAdapter
+  : TeamsCallAdapter;
+
 /**
  * @private
  */
 const useAzureCommunicationCallAdapterGeneric = <
-  AdapterType extends CallAdapter | /* @conditional-compile-remove(teams-identity-support) */ TeamsCallAdapter
+  CallType extends 'AzureCommunication' | 'Teams',
+  AdapterType extends AdapterTypeOf<CallType>
```
JamesBurnside, line 909: we should avoid naming a type with a suffix of Type, where we use `AdapterType` can we just use `Adapter` instead?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -848,42 +878,42 @@ export const createAzureCommunicationCallAdapter = async ({
...
+type AdapterTypeOf<CallType extends 'AzureCommunication' | 'Teams'> = CallType extends 'AzureCommunication'
+  ? CallAdapter
+  : TeamsCallAdapter;
+
 /**
  * @private
  */
 const useAzureCommunicationCallAdapterGeneric = <
-  AdapterType extends CallAdapter | /* @conditional-compile-remove(teams-identity-support) */ TeamsCallAdapter
+  CallType extends 'AzureCommunication' | 'Teams',
```
JamesBurnside, line 908: CallType isn't quite right here as we use type of a call to also mean pstn vs group call etc.
Instead can this be something like `AdapterKind` or `IdentityKind`

JamesBurnside (commented):
change looks good, left some nits on naming, waiting for green CI to approve

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -848,42 +878,42 @@ export const createAzureCommunicationCallAdapter = async ({
...
+  AdapterType extends AdapterTypeOf<CallType>
 >(
-  args: Partial<AzureCommunicationCallAdapterArgs>,
-  createAzureCommunicationCallAdapter: (args: AzureCommunicationCallAdapterArgs) => Promise<AdapterType>,
+  args: PartialArgsType<AdapterType>,
   afterCreate?: (adapter: AdapterType) => Promise<AdapterType>,
-  beforeDispose?: (adapter: AdapterType) => Promise<void>
+  beforeDispose?: (adapter: AdapterType) => Promise<void>,
+  type: CallType = 'AzureCommunication' as CallType
```
JamesBurnside, line 914: avoid `type` here as it is a keyword in javascript, suggestion: `identityKind`

JamesBurnside (approved):

edwardlee-msft:
Liking all the new names! Makes it more clear than having to read AzureCommunicationTeams... to know that it's related to Teams

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -937,14 +970,30 @@ const useAzureCommunicationCallAdapterGeneric = <
...
+          }
+          newAdapter = (await createAzureCommunicationCallAdapter({
+            credential,
+            displayName: args.displayName,
+            locator,
+            userId: userId as CommunicationUserIdentifier,
+            /* @conditional-compile-remove(PSTN-calls) */ alternateCallerId: args.alternateCallerId,
+            /* @conditional-compile-remove(rooms) */ options
+          })) as Adapter;
+        } else {
```
edwardlee-msft, line 989: nit: Since this is specifically a teams call adapter, should we wrap this in an "else if"?
Not really necessary in this case, but it would make it more explicit.

edwardlee-msft (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: Nice catch! Else-if makes lots of sense, then I will add a throw error sentence in last else closure, to make it future proof

PR #2575, opened 2022-12-05
[Bugfix] Device dropdown behavior updated for no devices

dmceachernmsft:
# What

Update dropdown behavior on the composite configuration screens when there are Video or Audio devices
# Why

Communicates to the end user that they have Microphones, Camera or Speakers connected to their device.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3024504
# How Tested

Added new snapshot tests to verify the dropdowns behavior should there be no devices available


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -551,6 +551,9 @@ export interface CallCompositeStrings {
     mutedMessage: string;
     networkReconnectMoreDetails: string;
     networkReconnectTitle: string;
+    noCameraLabel: string;
```
JamesBurnside, line 554: nit for consistency with the other strings: 
```suggestion
    noCamerasLabel: string;
```

packages/react-composites/tests/browser/call/hermetic/ConfigurationPage.test.ts
```diff
@@ -32,6 +32,20 @@ test.describe('Call Composite E2E Configuration Screen Tests', () => {
     await waitForCallCompositeToLoad(page);
     expect(await stableScreenshot(page)).toMatchSnapshot('call-configuration-page-with-call-details.png');
   });
+
+  test.only('Configuration screen should show no devices available', async ({ page, serverUrl }) => {
```
JamesBurnside, line 36: remove `.only`

packages/react-composites/src/composites/CallComposite/components/ConfigurationpageCameraDropdown.tsx
```diff
@@ -41,14 +41,14 @@ export const ConfigurationpageCameraDropdown = (props: ConfigurationpageCameraDr
   /* @conditional-compile-remove(call-readiness) */
   const cameraBlockedDropdown = (
     <_DevicePermissionDropdown
+      {...props.dropdownProps}
```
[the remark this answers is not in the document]
dmceachernmsft: this was a change left over from some testing, I reverted it but should we not be doing the spread first than setting the props? or is the idea if something is found from the selector to be able to override it?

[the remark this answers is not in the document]
JamesBurnside: Nope I agree with you the spread should be first so it doesn't accidentally stomp over other variables explicitly defined!
Just wanted to understand if there was a specific field that was causing an issue that had forced the line to be moved (to understand if something bad was hidden) -- but nope you were just making the code better haha.
Feel free to un-revert it or leave as is :)

JamesBurnside:
Separate bug, for speaker selection I don't think we need microphone permission:
image

packages/react-composites/src/composites/CallComposite/components/LocalDeviceSettings.tsx
```diff
@@ -131,14 +134,18 @@ export const LocalDeviceSettings = (props: LocalDeviceSettingsType): JSX.Element
...
+  const hasCameras = props.cameras.length > 0 ? true : false;
+  const hasMicrophones = props.microphones.length > 0 ? true : false;
+  const hasSpeakers = props.speakers.length > 0 ? true : false;
+
   const cameraGrantedDropdown = (
     <Dropdown
       data-ui-id="call-composite-local-camera-settings"
       aria-labelledby={'call-composite-local-camera-settings-label'}
-      placeholder={defaultPlaceHolder}
+      placeholder={!hasCameras ? noCameraLabel : defaultPlaceHolder}
```
mgamis-msft, line 145: nit: `hasCameras ? defaultPlaceholder : noCameraLabel` reads a little better

packages/react-composites/src/composites/CallComposite/components/LocalDeviceSettings.tsx
```diff
@@ -165,9 +172,9 @@ export const LocalDeviceSettings = (props: LocalDeviceSettingsType): JSX.Element
       {roleCanUseMic && (
         <Dropdown
           aria-labelledby={'call-composite-local-sound-settings-label'}
-          placeholder={defaultPlaceHolder}
+          placeholder={!hasMicrophones ? noMicLabel : defaultPlaceHolder}
```
mgamis-msft, line 175: Same nit as above

mgamis-msft (approved):

change/@azure-communication-react-254dcd24-ad35-401b-be39-ae52d3869471.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
JamesBurnside, line 2: this should be minor as we are introducing new strings Contoso may have to override/localize themselves

packages/react-composites/src/composites/CallComposite/components/LocalDeviceSettings.tsx
```diff
@@ -131,14 +134,18 @@ export const LocalDeviceSettings = (props: LocalDeviceSettingsType): JSX.Element
   /* @conditional-compile-remove(call-readiness) */
   const dropdownProps = usePropsFor(_DevicePermissionDropdown);
 
+  const hasCameras = props.cameras.length > 0 ? true : false;
```
JamesBurnside, line 137: nit: the ternaries here redundant, just `props.camera.length > 0` will return true or false

dmceachernmsft:
> For mobile snapshot can we open the device flyout: image
> 
> I'm curious for android where sometimes we combine the speaker and microphone list to ensure that is handled expentantly

This is a new bug filed for Alex P to look at how we might show this, the behavior in this fix disables the buttons based on number of devices in each category. will need some work when Alex knows how we communicate whats happening.

dmceachernmsft:
> Separate bug, for speaker selection I don't think we need microphone permission: image

is this not just showing us that when the microphone permissions are not set the speaker devices don't populate as well?

JamesBurnside:
What's the difference between these two and why is one devices button disabled and one isn't:
image
image

JamesBurnside (approved):

PR #2576, opened 2022-12-06
Cursor should show at the end when clicking in dialpad textfield

carocao-msft:
# What
Cursor should show at the end when clicking in dialpad textfield

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3015832

# How Tested

https://user-images.githubusercontent.com/96077406/205985140-f02ec960-c9a2-4559-ad58-688e3cfd3e73.mov

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/Dialpad/Dialpad.tsx
+++ b/packages/react-components/src/components/Dialpad/Dialpad.tsx
@@ -267,6 +267,7 @@ const DialpadContainer = (props: {
         onChange={(e: any) => {
           setText(e.target.value);
         }}
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         onClick={(e: any) => {
           const input = e.target;
           const end = input.value.length;
--- a/packages/react-components/src/components/Dialpad/Dialpad.tsx
+++ b/packages/react-components/src/components/Dialpad/Dialpad.tsx
@@ -267,6 +267,14 @@ const DialpadContainer = (props: {
         onChange={(e: any) => {
           setText(e.target.value);
         }}
+        onClick={(e: any) => {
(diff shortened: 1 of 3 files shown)


edwardlee-msft:
This is an interesting scenario and not sure if I agree. I can see how this would make sense for a call that is already dialling, but for a call that has yet to be dialled I can see a need to edit the middle of the number.

But regardless, I can see that Alex P provided his input so no issues either.

edwardlee-msft (approved):

carocao-msft:
> This is an interesting scenario and not sure if I agree. I can see how this would make sense for a call that is already dialling, but for a call that has yet to be dialled I can see a need to edit the middle of the number.
> 
> But regardless, I can see that Alex P provided his input so no issues either.

To edit the middle of number you can click and then using the left arrow on keyboard to move cursor to the middle. I was concerned about this as well so I made sure we can still do that

packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -267,6 +267,15 @@ const DialpadContainer = (props: {
...
           setText(e.target.value);
         }}
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        onClick={(e: any) => {
+          const input = e.target;
+          const end = input.value.length;
+
+          // Move focus to end of input field
+          input.setSelectionRange(end, end);
+          input.focus();
```
dmceachernmsft, line 277: This doesn't auto focus the dialpad text box right? Can a keyboard use tab in and out of the box as well?

dmceachernmsft (approved):

PR #2577, opened 2022-12-07
Add teams identity option to sample

PorterNan:
# What

Add teams identity option to sample
Implement an extra component for teams call wrapper

A screenshot for teams identity option:
image

# Why

# How Tested

# Process & policy checklist


samples/Calling/src/app/App.tsx
```diff
@@ -56,6 +56,8 @@ const App = (): JSX.Element => {
   /* @conditional-compile-remove(rooms) */
   const [role, setRole] = useState<Role>();
 
+  const [isTeamsCall, setIsTeamsCall] = useState<boolean>(false);
```
JamesBurnside, line 59: nit: almost worth conditionally compiling this line just so when we update the public samples it obvious whether or not to keep this line

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -34,23 +35,15 @@ export interface CallScreenProps {
...
-    /* @conditional-compile-remove(call-readiness) */ callReadinessOptedIn
-  } = props;
+  const { token, userId, /* @conditional-compile-remove(teams-identity-support) */ isTeamsIdentityCall } = props;
   const callIdRef = useRef<string>();
-  const { currentTheme, currentRtl } = useSwitchableFluentTheme();
-  const isMobileSession = useIsMobile();
+
   const afterCreate = useCallback(
-    async (adapter: CallAdapter): Promise<CallAdapter> => {
+    async <CallAdapterType extends CommonCallAdapter>(adapter: CallAdapterType): Promise<CallAdapterType> => {
```
JamesBurnside, line 46: hmm this is very confusing for a sample app to read

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -71,19 +64,60 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
...
+    <ACSCallScreen afterCreate={afterCreate} credential={credential} {...props} />
+  );
+};
+
+type CommonCallScreenProps = CallScreenProps & {
+  afterCreate?: <CallAdapterType extends CommonCallAdapter>(adapter: CallAdapterType) => Promise<CallAdapterType>;
+  credential: AzureCommunicationTokenCredential;
+};
+
+const TeamsCallScreen = (props: CommonCallScreenProps): JSX.Element => {
```
JamesBurnside, line 86: there's so much Teams-logic vs ACS-logic, for the sample it would be nice if all ACS-call logic was in one file, and all Teams-Identity logic was in one file

JamesBurnside (approved):
Looks good! My only concern is how complicated this makes the sample and how difficult it will be for us to pull apart when we go to update the public samples - but not blocking on that

samples/Calling/src/app/views/CallScreen.tsx
PorterNan: Maybe I can extract these 2 component into a new file

BTW, there is no big chunk of teams logic, the only different here is useAzureCommunicationTeamsCallAdapter vs useAzureCommunicationCallAdapter
But because of we can't call hooks conditionally, these 2 components becomes a must-have middle layer, sigh...

samples/Calling/src/app/views/CallScreen.tsx
PorterNan: Agree, then I will use another option in my mind - extract the common piece into a single util function

samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -50,6 +50,8 @@ export interface HomeScreenProps {
     outboundParticipants?: string[];
     /* @conditional-compile-remove(PSTN-calls) */
     alternateCallerId?: string;
+    /* @conditional-compile-remove(teams-identity-support) */
+    teamsToken?: string;
```
edwardlee-msft, line 54: nit: possibly rename to teamsIdentityToken?

edwardlee-msft (approved):

samples/Calling/src/app/views/CallScreen.tsx
PorterNan: I was moving code around for a while, then I noticed TeamsCallScreen and ACSCallScreen might be too small to stay in a file(especially without the context of callScreen)
But CallScreen file itself is too big to be a single file, then I realized I can extract the composite container part away and keep a nice balance of each file. (We definitely need some separate on the app level tbh)

samples/Calling/src/app/views/HomeScreen.tsx
PorterNan: I checked the word they use in teams scenario, they still call it token or access token just like ACS, so I guess add a prefix "Teams" would align with their naming

For now if you wanna get teams token, there are 2 ways, 

1. go to azure portal then generate one:
image

2. Set up a service following the instruction from https://learn.microsoft.com/en-us/azure/communication-services/quickstarts/manage-teams-identity?pivots=programming-language-javascript, then you would get your teams token when you log into this service using OAuth or some other ways

PR #2583, opened 2022-12-08
[SPIKE] Drawer Menu in Video Tile

anjulgarg:
# What

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/RemoteVideoTile.tsx
+++ b/packages/react-components/src/components/RemoteVideoTile.tsx
@@ -1,8 +1,21 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+<<<<<<< Updated upstream
 import React, { useMemo } from 'react';
 import { CreateVideoStreamViewResult, OnRenderAvatarCallback, ParticipantState, VideoStreamOptions } from '../types';
+=======
+import { IContextualMenuProps, Layer, Stack } from '@fluentui/react';
+import React, { useMemo } from 'react';
+import {
+  CreateVideoStreamViewResult,
+  OnRenderAvatarCallback,
+  ParticipantState,
+  VideoGalleryRemoteParticipant,
+  VideoStreamOptions
+} from '../types';
+import { _DrawerMenu, _DrawerMenuItemProps } from './Drawer';
+>>>>>>> Stashed changes
 import { StreamMedia } from './StreamMedia';
 import {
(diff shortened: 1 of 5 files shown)


packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -90,9 +91,11 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
         showCameraSwitcherInLocalPreview={props.isMobile}
         localVideoCameraCycleButtonProps={cameraSwitcherProps}
         onRenderAvatar={onRenderAvatar}
+        /* @conditional-compile-remove(pinned-participants) */
+        showRemoteVideoTileContextualMenu={!props.isMobile}
```
mgamis-msft, line 95: Shoud we also have a prop for drawerMenuHostId here?

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -38,6 +46,8 @@ export const _RemoteVideoTile = React.memo(
     showLabel?: boolean;
     personaMinSize?: number;
     participantState?: ParticipantState;
+    showRemoteVideoTileContextualMenu?: boolean;
+    hostId?: string;
```
mgamis-msft, line 50: I don't mind this optional prop here since RemoteVideoTile is an internal component. But wouldn't we have to create this prop in VideoGallery so that the RemoteVideoTiles created by the VideoGallery know their host. 

Would providing a `hostId` and seting `showRemoteVideoTileContextualMenu` to true would clicking to the ellipses or a long touch would open both the drawer menu and the contextual menu? If that's true I am still okay with that since RemoteVideoTile is internal

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -163,6 +168,12 @@ export interface VideoGalleryProps {
    * Camera control information for button to switch cameras.
    */
   localVideoCameraCycleButtonProps?: LocalVideoCameraCycleButtonProps;
+  /* @conditional-compile-remove(pinned-participants) */
+  /**
+   * Whether to show the remote video tile contextual menu.
+   * @defaultValue `true`
+   */
+  showRemoteVideoTileContextualMenu?: boolean;
```
anjulgarg, line 176: drawerMenuHostId

PR #2587, opened 2022-12-12
[bugfix] Resolve Mute/Unmute Icon tooltip showing up immediately after the call is joined 

edwardlee-msft:
# What
Requirements:
1. A11y MAS compliance. First tab keypress must fall onto Microphone button.
2. Hide tooltip on initial render
3. Screen Reader must not cause confusion.

Fix:
1. CallComposite: Reorder ControlBar in Call Composite to the top of the DOM tree to have Microphone button in first tab keypress order focus rather than videoTile. Removes unneeded autofocus attribute on microphone button. 

5. CallWithChatComposite: Added a button that is hidden upon component mount in CallWithChat ControlBar to capture the initial tab keypress order focus. 
Note, Call Composite fix does not work for CallWithChat Composite because the focus is set on the CallComposite component inside CallWithChat. This causes all first initial tab keypresses to focus on the VideoTile regardless of DOM tree order. 
In others words for CallWithChat, because contoso interacted with the buttons in the configuration screen, the focus is currently set on CallComposite Component when entering the Call Screen. 
To unbind the focus from CallComposite Component, a button is added in CallWithChat ControlBar that takes initial focus using the autoFocus prop then removed from view once focus is set to the CallWIthChat ControlBar group.

# Why
**https://skype.visualstudio.com/SPOOL/_workitems/edit/3007794**
image

# How Tested
Locally on MacOS Chrome on both Call Composite and CallWithChat Composite

# Process & policy checklist


packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -74,6 +75,12 @@ const inferCallWithChatControlOptions = (
  * @private
  */
 export const CallWithChatControlBar = (props: CallWithChatControlBarProps & ContainerRectProps): JSX.Element => {
+  useEffect(() => {
+    // On mount, button used for initial focus is hidden to prevent screen readers from announcing the initial focus
+    if (document.querySelector('[data-ui-id=call-with-chat-autofocus-hidden-button]') === document.activeElement) {
+      document.activeElement?.toggleAttribute('hidden');
```
dmceachernmsft, line 81: This reference to the `activeElement` should only show as this button when it has focus right? so the chain of events is:
1. user presses tab entering focus to the page
2. focus is on this button so the hook sets the CSS to hidden
3. without the user pressing tab again, because the current item was removed through CSS the focus goes to the microphone button

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -165,6 +172,18 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
...
+                  hidden button to set first tab keypress focus on ControlBar on initial render. 
+                  Unable to set initial focus on Microphone as it causes tooltip to render. On mount, button is
+                  autofocused then immediately hidden
+                */}
+                <ControlBarButton
+                  autoFocus
+                  ariaHidden={true}
+                  data-ui-id={'call-with-chat-autofocus-hidden-button'}
+                  styles={hiddenAutoFocusButtonStyles}
+                  tabIndex={-1}
```
dmceachernmsft, line 185: could we maybe make a constant for this value? give it a name to say its a tab order remover? maybe: `hiddenFromTabOrderTabIndex` or something?

dmceachernmsft:
image
would seem that the hidden buttons still has something of a margin or padding. it is shifting the whole control bar to the right.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
edwardlee-msft: Thanks for the callout but I don't think it is necessary to make it into a constant as it's pretty self explanatory. 
-1 for tabIndex is standard for removing an element from being able to be tabbed into with the keyboard.

JamesBurnside (approved):
Adding context from offline discussion for posterity. The reason the callwithchatcomposite's control bar needs a more nuanced approach than the callcomposite is because the callwithchatcontrolbar exists outside the callcomposite.
When the experience starts, the callcomposite has focus. When the user gets to a screen where the callwithchatcontrolbar is shown, is lives outside the callcomposite component (which has focus) and nothing tells the DOM to move focus to outside of the callcomposite to this new component.

Overall this PR looks great!

dmceachernmsft (approved):

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
JamesBurnside, line 81: why `toggleArtibute` and not set attribute? I assume there's not scenario where we toggle it back to showing?

packages/react-composites/src/composites/CallComposite/styles/CallPage.styles.ts
```diff
@@ -82,3 +82,12 @@ export const bannerNotificationStyles: IStackStyles = {
     pointerEvents: 'auto' // to allow the dismissal or error and warning bars in the notification container
   }
 };
+
+/**
+ * @private
+ */
+export const callArrangementContainerStyles: IStackStyles = {
+  root: {
+    root: { flexDirection: 'column-reverse' } // to allow first initial keyboard focus on ControlBar
```
mgamis-msft, line 91: Why is this double nested in 'root'?

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -153,7 +154,7 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
   return (
     <div ref={containerRef} className={mergeStyles(containerDivStyles)}>
       <Stack verticalFill horizontalAlign="stretch" className={containerClassName} data-ui-id={props.dataUiId}>
-        <Stack horizontal grow>
+        <Stack horizontal grow styles={callArrangementContainerStyles}>
```
mgamis-msft, line 157: I'm surprised the control bar still stacks vertically in this Stack because `horizontal` prop is true. Without the  flexDirection: 'column-reverse' style would the control bar still stack horizontally?
Does moving the CallControl in this Stack avoid initial focus before the first tab?

edwardlee-msft: That's right, without the column-reverse the control bar would stack horizontally.
image

_Does moving the call control bar in this Stack avoid initial focus before the first tab?_
Yes, moving the controlBar below the gallery would prevent it from having first initial focus and first tab focus as well. 
Hence why it was moved above the gallery for Call Composite

mgamis-msft: I see. Thanks.

mgamis-msft (approved):

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
JamesBurnside: to miguel's point here, should `horizontal` prop be removed from this stack then?

PR #2588, opened 2022-12-12
[A11y Bugfix] Add announcements for people button/ people flyout

dmceachernmsft:
# What

introduce participants button copy link announcement
# Why

Provides feedback to a narrator user that they have copied the invite link in the people pane or participants button flyout.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2790797
# How Tested

Validated in stable flavor sample apps.


packages/react-composites/src/composites/common/AddPeopleButton.tsx
```diff
@@ -92,10 +112,14 @@ export const AddPeopleButton = (props: AddPeopleButtonProps): JSX.Element => {
       <Stack tokens={peoplePaneContainerTokens}>
         {inviteLink && (
           <Stack styles={copyLinkButtonStackStyles}>
+            <Announcer announcementString={copyInviteLinkAnnouncerStrings} ariaLive={'polite'} />
```
mgamis-msft, line 115: What if we add a boolean prop that hides the Announcer component by setting the style display: none and then we use a boolean ref instead for that prop . Something like this:
```
const announcerActive = useRef(false);
...
<Announcer announcementString={copyInviteLinkAnnouncerStrings} ariaLive={'polite'} hide={!announcerActive}/>
```
and then the onClick callback will simply be
```
              onClick={() => {
                copy(inviteLink ?? '');
                announcerActive.current = true;
                setTimeout(() => {
                  announcerActive.current = false;
                }, 3000);
              }}
```
Just an idea.

dmceachernmsft: so the Announcer is designed to be invisible on the page through CSS already, but if you make something hidden in CSS through `display: none` the narrator will ignore it.

mgamis-msft: The narrator will still narrate the Announcer string even with CSS `display: none`? Regardless, I don't think my idea is any better. useRef vs useState. Either way a hook is needed to keep state.

mgamis-msft (approved):

carocao-msft (approved):

PR #2590, opened 2022-12-13
Display correct guidance text for device permission error on safari

carocao-msft:
# What
Display correct guidance text for device permission error on safari 

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3086741

# How Tested
Screenshot 2022-12-13 at 10 08 59 AM
Screenshot 2022-12-13 at 10 08 55 AM

# Process & policy checklist


packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx
```diff
@@ -59,7 +59,8 @@ export interface CameraAndMicrophoneDomainPermissionsProps extends CommonDomainP
   /** Strings for use with the {@link CameraAndMicrophoneDomainPermissions} */
   strings?: CameraAndMicrophoneDomainPermissionsStrings;
 }
-
+/* @conditional-compile-remove(call-readiness) */
+const isSafari = navigator.userAgent.indexOf('Safari') > -1 && navigator.userAgent.indexOf('Chrome') <= -1;
```
JamesBurnside, line 63: what is the behavior of chrome on iphone for this issue? 
Chrome and other browsers on ios are just re-skins of safari so I'm curious if they exhibit the same behavior

JamesBurnside: you may also need to filter out android, firefox and chrome alias: https://stackoverflow.com/a/40463096/3047190

carocao-msft: Having a hard time to check chrome on iphone, it's shown as unsupported browser and codespace still doesn't work for me :(
Tested the above code on firefox and chrome and safari on Mac, it's behaving properly. The above code already filters out chrome alias

carocao-msft: That's a neat regex! Updated :) For chrome on iphone still struggling to get codespace working

change-beta/@azure-communication-react-60233904-ae91-42c7-9b22-9394d4780adb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
dmceachernmsft, line 2: Should this not be prerelease? I know its a bug fix but it is for a beta feature. @JamesBurnside maybe you can correct me on this?

packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx
```diff
@@ -70,10 +74,19 @@ export const CameraAndMicrophoneDomainPermissions = (props: CameraAndMicrophoneD
   /* @conditional-compile-remove(call-readiness) */
   const locale = useLocale().strings;
 
+  /* @conditional-compile-remove(call-readiness) */
+  const isSafari = (): boolean => {
```
dmceachernmsft, line 78: you should extract this, and the function on lines 147 and 208 to just be 1 function

carocao-msft: Good call, fixed that

packages/react-composites/src/composites/CallComposite/components/CallReadinessModal.tsx
```diff
@@ -61,6 +68,8 @@ export const CallReadinessModal = (props: {
           return (
             <CameraAndMicrophoneDomainPermissions
               appName={'app'}
+              /* @conditional-compile-remove(unsupported-browser) */
```
dmceachernmsft, line 71: What happens if CallReadiness goes before unsupported browser to GA? will this break the app? these modals need the browser type right? or does it just default to the non safari strings if this isn't release yet?

change-beta/@azure-communication-react-60233904-ae91-42c7-9b22-9394d4780adb.json
edwardlee-msft: I also believe this should be prerelease as well

packages/react-composites/src/composites/CallComposite/components/CallReadinessModal.tsx
carocao-msft: It uses the useragent when unsupported browser is not available

change-beta/@azure-communication-react-60233904-ae91-42c7-9b22-9394d4780adb.json
JamesBurnside: This PR has two parts: one prerelease and one patch
Prerelease for the fix for the domainpermissions modal
Patch for the fix to the general error bar strings
So I'd keep patch here

packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx
```diff
@@ -23,6 +23,10 @@ export interface CommonDomainPermissionsProps {
    * Type of the Domain Permissions component.
    */
   type: 'request' | 'denied' | 'check';
+  /**
+   * Type of the browser domain permission component is rendered in.
+   */
+  browser?: string;
```
JamesBurnside, line 29: browser can't be `string` here - how would contoso know what string value to provide, and what will happen with specific string values?
Consider `browser: 'safari' | 'other' with @defaultValue other`

packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx
```diff
@@ -60,6 +64,13 @@ export interface CameraAndMicrophoneDomainPermissionsProps extends CommonDomainP
   strings?: CameraAndMicrophoneDomainPermissionsStrings;
 }
 
+/* @conditional-compile-remove(call-readiness) */
+const isSafari = (browser: string | undefined): boolean => {
+  /* @conditional-compile-remove(unsupported-browser) */
+  return browser === 'safari';
+  return /^((?!chrome|android|crios|fxios).)*safari/i.test(navigator.userAgent);
```
JamesBurnside, line 71: I wouldn't do this at the pure component level and instead in `CallComposite/components/CallReadinessModal.tsx`

JamesBurnside, line 71: This regex check is in two places, consider exporting (internally) a util function called `_isSafari`:
```
export const _isSafari = (environmentInfo: /** conditional compile environment info type out */ EnvironmentInfo | undefined) {
  return environmentInfo?.environment.browser === 'safari';
  return /^((?!chrome|android|crios|fxios).)*safari/i.test(navigator.userAgent);
}`

File: packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx (hunk 5 above)
Comment (carocao-msft) [reply packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx]: updated

File: packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx (hunk 6 above)
Comment (carocao-msft) [reply packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx]: updated

Comment (carocao-msft) [reply packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx]: Updated

File: packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx [hunk 7]
```diff
@@ -23,6 +23,10 @@ export interface CommonDomainPermissionsProps {
    * Type of the Domain Permissions component.
    */
   type: 'request' | 'denied' | 'check';
+  /**
+   * Whether we are using safari browser or not
+   */
+  isSafari?: boolean;
```
JamesBurnside, line 29: `isSafari` is OK, but if we ever needed to do something specific for firefox having a boolean makes it less extensible

carocao-msft: If we want to move the logic to callreadinessmodal and pass it down to domain permissions, we can only pass along a boolean. Earlier I had the logic inside domain permissions component so that when we need to do something for firefox it also works

JamesBurnside: Why is it we can only pass a boolean? We can do
```
const isSafari = _isSafari(environmentInfo);
<CameraAndMicrophoneDomainPermissions
  browser={isSafari ? 'safari' | 'other'}
/>
```

carocao-msft: In this way we are still limited when we need to do something specific for firefox, since 'other' includes chrome and firefox. Should we have the browser type to be safari | firefox | chrome | edge ?

JamesBurnside: When we need to do firefox we can then add it, but you're correct that `other` isn't so great, perhaps:
`browserHint: 'safari' | 'unset'`
?

JamesBurnside (approved):
We can iterate on the isSafari prop in future so not blocking for this PR, at GA we will likely get feedback from review boards to avoid the boolean - we can always update then - otherwise Pr looks great!

packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx
dmceachernmsft: So I like the idea of spreading it out to define the major browsers explicitly so `other` refers to browsers like Brave, duckduckgo and so on. I'd argue Opera might be in there too but it is prominent enough in the world that Calling could want to support it in the distant future.

dmceachernmsft (approved):

PR #2593, opened 2022-12-14
[CallReadiness] Update composite useAdapted selector to get new StatefulCallClientState

dmceachernmsft:
# What

Updates the `useAdaptedSelector` to get new state variables for environmentInfo and alternateCallerId
# Why

Allows component usePropsFor to get the new state variables for components

https://github.com/Azure/communication-ui-library/pull/2590
# How Tested

made sample app that gets selector data from getEnvironmentInfo base selector and put it in errorbar selector to see if populates on config screen. (carolines PR above does something similar).


carocao-msft (approved):

packages/react-composites/src/composites/CallComposite/hooks/useAdaptedSelector.ts
```diff
@@ -80,7 +82,9 @@ const memoizeState = memoizeOne(
     deviceManager: DeviceManagerState,
     calls: { [key: string]: CallState },
     latestErrors: CallErrors,
-    displayName?: string
+    displayName?: string,
+    alternateCallerId?: string,
+    environmentInfo?: undefined | /* @conditional-compile-remove(unsupported-browser) */ EnvironmentInfo
```
PorterNan, line 87: Is it possible to conditional compile the param instead of the type?

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -119,3 +121,12 @@ export const getDiagnostics = (
  */
 export const getCallState = (state: CallClientState, props: CallingBaseSelectorProps): string =>
   state.calls[props.callId]?.state;
+
+/* @conditional-compile-remove(unsupported-browser) */
+/**
+ * @private
+ */
+export const getEnvironmentInfo = (state: CallClientState): EnvironmentInfo | undefined => {
+  console.log(state);
```
PorterNan, line 130: Remove console.log from here

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -119,3 +121,12 @@ export const getDiagnostics = (
  */
 export const getCallState = (state: CallClientState, props: CallingBaseSelectorProps): string =>
   state.calls[props.callId]?.state;
+
+/* @conditional-compile-remove(unsupported-browser) */
+/**
+ * @private
+ */
+export const getEnvironmentInfo = (state: CallClientState): EnvironmentInfo | undefined => {
```
PorterNan, line 129: Question about this baseSelector - is there any component currently using this selector? It makes sense to put it here if there is, if not, then this selector should stay in composite folder

packages/react-composites/src/composites/CallComposite/hooks/useAdaptedSelector.ts
dmceachernmsft: not without defining a whole type for the function params unfortunately. it wouldn't need it at all if it weren't the EnvironmentInfo type since its not in the stable SDK yet

packages/calling-component-bindings/src/baseSelectors.ts
dmceachernmsft: Caroline is going to be using it in the errorbar component once this is merged in.

PorterNan (approved):

packages/calling-component-bindings/src/baseSelectors.ts
PorterNan: So we got a plan to expose these on component layer, nice!

PR #2594, opened 2022-12-14
Fix no video errors showing on the configuration page

JamesBurnside:
# What
Tee errors to state when starting (or stopping) a local preview fails 
* This is done by tee-ing errors in the `createUnparentedView` method in our stream utils. This is not ideal as we typically do not wrap a function at this level but since there is no `startLocalVideoPreview` method to wrap instead this is the only place that makes sense without adding a new public function.

# Why
Currently on the configuration page, if starting or stopping the video fails there is no error message.
With this change the error bar correctly shows

# How Tested
- Added unit test that ensures that the state has been teed the error when createUnparentedView throws an error.
- Added e2e test for capturing UI (does not test mechanism that ensures on failure the context's error state is set, the unit test does this)
- Ran locally and verified I saw the error bar that matches the snapshot results when camera in use by another application


carocao-msft (approved):

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -597,7 +597,12 @@ export function createView(
     return createViewLocalVideo(context, internalContext, callId, options);
   } else if (!('id' in stream) && !callId) {
     // Render LocalVideoStream that is not part of a Call
-    return createViewUnparentedVideo(context, internalContext, stream, options);
+    // Because it is not part of the call we don't tee errors to state naturally (e.g. via a Call Client function such as startVideo).
+    // We do not have a startLocalPreviewVideo function, so as a workaround we ensure any errors are propagated here.
+    return context.withAsyncErrorTeedToState(
```
dmceachernmsft, line 602: This is cool! So anytime we are passing around the CallContext in the adapter we can also kick the `withAsyncErrorTeedToState`? are there any situations in the chat composite/adapter that we should be thinking about this too?

dmceachernmsft (approved):

packages/calling-stateful-client/src/StreamUtils.ts
JamesBurnside: technically yes, but we should be careful when we call it as we don't want to trigger multiple times for the same error. ideally this is just wrapped around statefulclient handlers but since there isn't one for starting local preview video I added one here

PR #2595, opened 2022-12-15
[bugfix] Change aspect ratio for local video tile to 16:9 to capture full camera view

edwardlee-msft:
# What
LocalVideoTile modal aspect ratio changed to 16:9 to capture full camera view on desktop/landscape.

Teams also has their local video tile 16:9 on desktop with or without a camera being enabled.

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3009419
left and right boundaries were being cropped from local video tile. 
Can cause misinformation for the user on what is being captured and shared to others in the call.
image

# How Tested

# Process & policy checklist


packages/react-components/src/components/VideoGallery/styles/FloatingLocalVideo.styles.ts
```diff
@@ -34,8 +34,9 @@ export const SMALL_FLOATING_MODAL_SIZE_PX = { width: 64, height: 88 };
 
 /**
  * Large floating modal width and height in rem for large screen
+ * Aspect ratio: 16:9
  */
-export const LARGE_FLOATING_MODAL_SIZE_PX = { width: 160, height: 120 };
+export const LARGE_FLOATING_MODAL_SIZE_PX = { width: 215, height: 120 };
```
JamesBurnside, line 39: I like this as a quick fix that covers most use cases, we should definitely keep this change and push this PR in. That being said do we have a plan to also address non-16:9 ratios? (e.g. portrait cameras such as mobiles)

JamesBurnside: (If this just works (everything else plays nicely with us adjusting this width), we could use the incoming stream size we get from beta calling sdk to set the width of the local preview dynamically)

edwardlee-msft: For Mobile, it uses the width and height from line 33
`export const SMALL_FLOATING_MODAL_SIZE_PX = { width: 64, height: 88 };`.

It's not quite the usual 4:3 aspect ratio that cameras use... I tested around with different width and heights for mobile but unfortunately, I am unable to get the full camera view. either the height, width, or both get cropped (zoomed in despite being the correct aspect ratio).

So for now, this fix is primarily for Desktop.

edwardlee-msft: I'll continue to investigate why mobile zooms in despite changing the modal to the correct aspect ratio. 
But, as of right now, the current modal size for mobile local video tiles is relatively accurate. Crops very slightly on left and right.

edwardlee-msft: > (If this just works (everything else plays nicely with us adjusting this width), we could use the incoming stream size we get from beta calling sdk to set the width of the local preview dynamically)

Plays well on desktop! Not so well for mobile from my manual aspect ratio testing.

JamesBurnside:
Existing bug (not caused by this Pr just noticed it in the snapshot updates), the local tile should be in the horizontal gallery and not have a drop shadow:
image

JamesBurnside (approved):

packages/react-components/src/components/VideoGallery/styles/VideoGalleryResponsiveHorizontalGallery.styles.ts
```diff
@@ -4,16 +4,7 @@
...
-
-/**
- * Small floating modal width and height in rem for small screen
- */
-export const SMALL_FLOATING_MODAL_SIZE_PX = { width: 64, height: 88 };
-
-/**
- * Large floating modal width and height in rem for large screen
- */
-export const LARGE_FLOATING_MODAL_SIZE_PX = { width: 160, height: 120 };
```
mgamis-msft, line 16: Fixing my bug on your branch. Lesson is never have the same variable defined twice :)
https://user-images.githubusercontent.com/79475487/207985251-3e3fb24d-0034-4f1c-819d-6fc338a219ea.mp4
Thank goodness for chromatic!

mgamis-msft (approved):

packages/react-components/src/components/VideoGallery/styles/VideoGalleryResponsiveHorizontalGallery.styles.ts
edwardlee-msft: Hahaha, Nice!! :)

PR #2596, opened 2022-12-15
Fix chat pane problem when no parent height is set

PorterNan:
# What

This is the best minimum way to introduce as less change as possible, it based on these facts:
1. Stack or display:flex, the default align-items value is 'normal', which behaves like 'stretch' for each children items,
2.  if we set height to auto, the child item will try to occupy the whole vertical space, which is exactly what we want.

Why 100% not working - because there is no parent item to tell the child div what the 100% is (especially when the parent flex box height is set to auto). Set the root element height will help with this scenario

Before:
image

After:
image

# Why

# How Tested

# Process & policy checklist


edwardlee-msft (approved):
Very nice! Great find Nan

PR #2597, opened 2022-12-15
Device denied screen should show on safari when don't allow is clicked 

carocao-msft:
# What
Device denied screen should show on safari when don't allow is clicked 

# Why
When don't allow is clicked on safari, permissionsAPI returns prompt instead of denied (that is because inside safari settings it's still set to ask when don't allow is clicked.) Calling SDK returns false in this case

So we add an extra layer of checking to make sure we always show the denied screen when calling sdk returns false. 

# How Tested

https://user-images.githubusercontent.com/96077406/207913061-04132871-dabb-4ffc-a9ba-1ce46785b40d.mov

# Process & policy checklist

### DIFF
--- a/change/@azure-communication-react-d74c9c67-8c1e-4d3c-a629-38ddfa99a3fb.json
+++ b/change/@azure-communication-react-d74c9c67-8c1e-4d3c-a629-38ddfa99a3fb.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed bug where on safari browser the device denied screen is not showing when don't allow is clicked ",
+  "packageName": "@azure/communication-react",
+  "email": "carolinecao@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@azure-communication-react-d74c9c67-8c1e-4d3c-a629-38ddfa99a3fb.json
+++ b/change/@azure-communication-react-d74c9c67-8c1e-4d3c-a629-38ddfa99a3fb.json
@@ -1,5 +1,5 @@
 {
-  "type": "patch",
+  "type": "prerelease",
   "comment": "Fixed bug where on safari browser the device denied screen is not showing when don't allow is clicked ",
   "packageName": "@azure/communication-react",
   "email": "carolinecao@microsoft.com",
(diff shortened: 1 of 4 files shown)


packages/react-composites/src/composites/CallComposite/components/CallReadinessModal.tsx
```diff
@@ -35,13 +35,15 @@ export const CallReadinessModal = (props: {
     camera: PermissionState;
     microphone: PermissionState;
   }) => void;
+  cameraPermissionGranted: boolean | undefined;
```
JamesBurnside, line 38: we have two sets of props here that look the same `videoState` and `cameraPermissionGranted` -- what's the difference between these two?

change/@azure-communication-react-d74c9c67-8c1e-4d3c-a629-38ddfa99a3fb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
JamesBurnside, line 2: this should be prerelease at this fix only applies to the call readiness features

packages/react-composites/src/composites/CallComposite/components/CallReadinessModal.tsx
```diff
@@ -249,15 +245,20 @@ export const CallReadinessModalFallBack = (props: {
...
 
-  // When permissions are not set, value is undefined, do nothing here
-  // When permissions are set to denied, value is false, show helper screen
-  const showModal = cameraPermissionGranted === false || microphonePermissionGranted === false;
+  const videoState = permissionsState.camera;
+  const audioState = permissionsState.microphone;
+
+  // When permissions are not set, do nothing here
+  // When permissions are set to denied, show helper screen
+  const showModal = videoState === 'denied' || audioState === 'denied';
```
JamesBurnside, line 253: Can correct me here but I think we will want to show the modal if videoState is `denied` or `prompt` and not just denied

JamesBurnside (approved):

dmceachernmsft (approved):

PR #2599, opened 2022-12-15
Change is Safari prop to browserHint to allow more customization

carocao-msft:
# What
Change is Safari prop to browserHint to allow more customization

# Why
In this way if we want to display some specific text for firefox we don't need to add a new prop

# How Tested
Made sure safari is showing the right text
Chrome and firefox is showing the right text

### DIFF
--- a/packages/communication-react/review/beta/communication-react.api.md
+++ b/packages/communication-react/review/beta/communication-react.api.md
@@ -1419,7 +1419,7 @@ export interface CommonCallingHandlers {
 // @beta
 export interface CommonDomainPermissionsProps {
     appName: string;
-    isSafari?: boolean;
+    browserHint: 'safari' | 'unset';
     onContinueAnywayClick?: () => void;
     onTroubleshootingClick?: () => void;
     styles?: DomainPermissionsStyles;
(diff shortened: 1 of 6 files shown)


JamesBurnside (approved):

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1419,7 +1419,7 @@ export interface CommonCallingHandlers {
 // @beta
 export interface CommonDomainPermissionsProps {
     appName: string;
-    isSafari?: boolean;
+    browserHint: 'safari' | 'unset';
```
dmceachernmsft, line 1422: Can we extract this as a type somewhere we can expand in the future?

packages/react-components/src/components/DevicePermissions/DomainPermissions.tsx
```diff
@@ -24,9 +24,9 @@ export interface CommonDomainPermissionsProps {
    */
   type: 'request' | 'denied' | 'check';
   /**
-   * Whether we are using safari browser or not
+   * Type of the browser used, the domain permission component will show different guidance text based on the browser type
    */
-  isSafari?: boolean;
+  browserHint: 'safari' | 'unset';
```
dmceachernmsft, line 29: Same comment as above

dmceachernmsft (approved):

PR #2601, opened 2022-12-16
[bugfix] Resolve Mute/Unmute Icon tooltip showing up immediately after the call is joined 

edwardlee-msft:
### Round 2 👀
**Note: This is a PR resubmission for a reverted PR https://github.com/Azure/communication-ui-library/commit/513702c4f8e2009be43d02c52a6e6955fd38921a.
This PR Resolved** issue found by Nan where Call Composite was not presenting CallPane in correct sequence. It is now working correctly. Thank you @PorterNan for the quick action and find! 

Call Composite views
| Desktop | Mobile |
|---|---|
|||

CallWithChat Screens below. 
Not impacted by the additional changes but screenshots provided for quicker review:
| Desktop | Mobile |
|---|---|
|||

# What
Requirements:
1. A11y MAS compliance. First tab keypress must fall onto Microphone button.
2. Hide tooltip on initial render
3. Screen Reader must not cause confusion.

Fix:
1. CallComposite: Reorder ControlBar in Call Composite to the top of the DOM tree to have Microphone button in first tab keypress order focus rather than videoTile. Removes unneeded autofocus attribute on microphone button. 

5. CallWithChatComposite: Added a button that is hidden upon component mount in CallWithChat ControlBar to capture the initial tab keypress order focus. 
Note, Call Composite fix does not work for CallWithChat Composite because the focus is set on the CallComposite component inside CallWithChat. This causes all first initial tab keypresses to focus on the VideoTile regardless of DOM tree order. 
In others words for CallWithChat, because contoso interacted with the buttons in the configuration screen, the focus is currently set on CallComposite Component when entering the Call Screen. 
To unbind the focus from CallComposite Component, a button is added in CallWithChat ControlBar that takes initial focus using the autoFocus prop then removed from view once focus is set to the CallWIthChat ControlBar group.

# Why
**https://skype.visualstudio.com/SPOOL/_workitems/edit/3007794**
image

# How Tested
Locally on MacOS Chrome on both Call Composite and CallWithChat Composite Desktop view and mobile view.

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -167,36 +168,38 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
...
+                  containerHeight={containerHeight}
+                  isMobile={props.mobileView}
+                  /* @conditional-compile-remove(one-to-n-calling) */
+                  peopleButtonChecked={activePane === 'people'}
+                  /* @conditional-compile-remove(one-to-n-calling) */
+                  onPeopleButtonClicked={togglePeoplePane}
+                />
+              </Stack.Item>
+            )}
+          <Stack horizontal grow>
```
edwardlee-msft, line 187: Grouped the Gallery with the CallPane to properly display the CallPane on the screen. 
This is the primary fix for the CallPane issue Nan found.

packages/react-composites/src/composites/common/styles/Pane.styles.ts
```diff
@@ -75,6 +75,7 @@ export const hiddenStyles: IStackStyles = {
 export const sidePaneStyles: IStackStyles = {
   root: {
     height: 'auto',
+    width: '100%',
```
PorterNan, line 78: Can we test the scenairo that when the width and height is not set for parent container - just simply remove width = 100% height=100% of .root class in callWithChat sample app and test it

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -74,6 +75,12 @@ const inferCallWithChatControlOptions = (
  * @private
  */
 export const CallWithChatControlBar = (props: CallWithChatControlBarProps & ContainerRectProps): JSX.Element => {
+  useEffect(() => {
```
PorterNan, line 78: consider call the hook named useRef + ref prop in react, it's a more react way, but gives you the same result

(if controlButton doesn't support that, use a basic react element instead,)

packages/react-composites/src/composites/CallWithChatComposite/styles/CallWithChatCompositeStyles.ts
```diff
@@ -43,3 +43,19 @@ export const ChatButtonContainerStyles: IStackStyles = {
     position: 'relative'
   }
 };
+
+/** @private */
+export const hiddenAutoFocusButtonStyles: IStackStyles = {
```
PorterNan, line 48: I like this style to ensure there is nothing unexpected showing for customers

One more parameter to make it even safer opacity: 0;  ;-D

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -156,6 +163,18 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
       <Stack.Item grow>
         <CallAdapterProvider adapter={props.callAdapter}>
           <Stack horizontalAlign="center">
+            {/*
+              hidden button to set first tab keypress focus on ControlBar on initial render.
```
PorterNan, line 167: This is an extremely creative way, nice job! Wondering if we could also use this as a common solution for CallComposite? Wondering if we could have a common solution across the composite which makes each piece more readable (I guess chat might have this problem in the future)

Would be nice to have a common component named HiddenFocusStartPoint (Or anything better) then put a button plus useEffects() there then well comment it, then this would be a nice common solution whenever we need it in composite

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
PorterNan: Also, we don't need a data-ui-id

PorterNan:
I tried to diff those changed screenshot online, but get 0% diff (which means you are doing it perfectly):
image

Consider reverting those visual snapshot changes, and remove "ui change" label

PorterNan (approved):
A smart solution! Learned something new by reviewing this PR today

JamesBurnside (approved):

JamesBurnside:
> I tried to diff those changed screenshot online, but get 0% diff (which means you are doing it perfectly): image
> 
> Consider reverting those visual snapshot changes, and remove "ui change" label

It's weird we're seeing this particular one in a lot of PRs with 0% change

packages/react-composites/src/composites/common/styles/Pane.styles.ts
edwardlee-msft: When root width and height are unset, the height does shrink:

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
edwardlee-msft: I implemented your suggestion and it works AMAZINGLY well! 
Great idea Nan!! Really improves on the scalability and cleanliness of the code as well. 
I've submitted new changes with your suggestion. 

I have tested it extensively to make sure it is working in Call Composite and CallWithChat Composite desktop mobile and with voiceOver on to ensure accessibility is working correctly.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
edwardlee-msft: Removed useEffect as I made the hidden button it's own reusable component as per your great suggestion.
https://github.com/Azure/communication-ui-library/pull/2601/files/eb637f038fb883e9edd0d10cb093a23944775bd3#r1051193481

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -156,6 +157,7 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
       <Stack.Item grow>
         <CallAdapterProvider adapter={props.callAdapter}>
           <Stack horizontalAlign="center">
+            {HiddenFocusStartPoint()}
```
PorterNan, line 160: Could just use this as a component instead of a function like <HiddenFocusStartPoint />, which react might have some optimizations on its own

packages/react-composites/src/composites/common/utils.tsx
```diff
@@ -48,3 +50,47 @@ export const useMinMaxDragPosition = (modalLayerHostId: string, rtl?: boolean):
...
+  }, []);
+  return (
+    <DefaultButton
+      autoFocus
+      ariaHidden={true}
+      styles={isMounted ? invisibleHiddenFocusStartPoint : hiddenFocusStartPointStyles}
+      tabIndex={-1}
+    />
+  );
+};
```
PorterNan, line 72: Since this is just a normal component, might be worthy to put it into a single file (even this is just util component, we even have TypeAssertion.tsx file under common/component, so this would be 100% good)

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
PorterNan: Maybe we could directly apply this to callComposite as well in this PR I guess? Putting an element at the top of DOM tree will work for current case for sure, but we are working on a future proof solution - we can't ensure the element would always stay on the top, even it is reversed order(we might be adding more element around)

But thanks to your common solution, we have one-stop component to solve them all together

packages/react-composites/src/composites/common/utils.tsx
edwardlee-msft: Okay! Let's do that!
Pushed the new changes, HiddenFocusStartPoint is now in it's own file

PorterNan:
I also checked focusZone component from fluent, they have something similar to solve this problem - but that component is a bit too heavy only for this case (and could be a potential behavior change), I am happy to see us introducing a light-weight component to solve this problem (and maybe similar problem in the future)

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
edwardlee-msft: You make a good point that we can't guarantee that the controlBar will always be the first interact-able element in the DOM tree. However, I believe it's good common practice to have tab focus order top down in the DOM Tree.

I'm okay with having the Call Composite use the same solution, but the solution for CallWithChat was only implemented as a workaround as reversing the order in CallWithChat did not resolve the focus issue.

PorterNan: Either way works for me, keep it in best practice is also a good point to me. We just need to keep in mind that, we have an elegant single component implemented by you to solve this problem, and apply it whenever it is not possible to keep the element on the top

(Maybe just add some more comments on in HiddenFocusStartPoint like "This is a util component when we can't ensure the initial element for first tab focus is at the top of dom tree, and it moves the first-tab focus to be on the element right after it")

edwardlee-msft: We can definitely revisit this fix for CallComposite later down the line :)
For now, I will keep it in dom tree order. Thanks a lot Nan!!

PR #2603, opened 2022-12-16
Turn call readiness alwayson (for beta packages only)

JamesBurnside:
# What
Remove callreadinessoptin flag

# Why
Instead it will be alwayson

# How Tested
--


packages/react-composites/src/composites/CallComposite/CallComposite.test.tsx
```diff
@@ -109,11 +109,4 @@ describe('CallComposite device permission test for call readiness opted in/opted
     expect(audioDevicePermissionRequests).toBe(1);
     expect(videoDevicePermissionRequests).toBe(1);
   });
-
-  /* @conditional-compile-remove(call-readiness) */
-  test('Audio and video device permission should still be requested when callreadiness option opted in', async () => {
-    mount(<CallComposite adapter={adapter} options={{ callReadinessOptedIn: true }} />);
```
dmceachernmsft, line 115: Should this test be removed? or should we be changing it to make sure the permissions are requested with CallReadiness on?

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.test.tsx
JamesBurnside: It's identical to the one above it so I removed it

edwardlee-msft (approved):

PR #2604, opened 2022-12-16
[CallReadiness] Fix device permissions request in Calling and CallWIthChat 

dmceachernmsft:
# What

Fixes issue on older iOS browsers that try Calling and CallWithChat.
# Why

Permissions API is not present so request from API trips error unmounting application.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3086057
# How Tested

On Ipad in browserstack tested old version of safari that does not use Permissions API. Got update browser page.


edwardlee-msft (approved):

edwardlee-msft:
Nice quick fix!

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -84,7 +84,11 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
   /* @conditional-compile-remove(call-readiness) */
   const [audioState, setAudioState] = useState<PermissionState | 'unsupported' | undefined>(undefined);
   /* @conditional-compile-remove(call-readiness) */
-  getDevicePermissionState(setVideoState, setAudioState);
+  try {
```
JamesBurnside, line 87: this shouldn't be necessary as getDevicePermissionState should already be catching the error:
image

JamesBurnside (changes requested):

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
dmceachernmsft: I don't think these are working because Navigator.Permissions is undefined
image
this is the error found in browserstack when this function is called.

JamesBurnside: ah makes sense, can you move the try catch still to here so that 'unsupported' is correctly being set?

There's also a couple things that should be fixed about this function:
* Its signature reads as if it should be synchronous as it doesn't return an awaitable promise 
* it's called `getXXXXX` which implies you are trying to get the permissions not populate some side state with this function

Can you update this? Something like:
```
In fact I would expect the try catch here to fail.. I guess the permissions .query must throw before the promise is constructed,
can you update this to be an inline async function instead of using the older promise syntax.
So something like:
```ts
const queryCameraPermissionFromPermissionsAPI = () => {
  try {
     return await navigator.permissions.query({ name: 'camera' as PermissionName });
  catch () {
    return 'unsupported'
  }
}

// another function as above but for mic

export const populateDevicePermissionState = async (
  setVideoState: (state: PermissionState | 'unsupported') => void,
  setAudioState: (state: PermissionState | 'unsupported') => void
): Promise<void> => {
  const [cameraPermission, microphonePermission = await Promise.all([
    queryCameraPermissionFromPermissionsAPI,
    queryMicrophonePermissionFromPermissionsAPI
  ]);

  setAudioState(microphonePermission);
  setVideoState(cameraPermission);
};
```
(syntax could be very wrong here, just typing as an example)

dmceachernmsft: on it!

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -289,35 +289,44 @@ export const isDisabled = (option: boolean | { disabled: boolean } | undefined):
...
 /**
  *
  * This function uses permission API to determine if device permission state is granted, prompt or denied
  * @returns whether device permission state is granted, prompt or denied
- * If permission API is not supported on this browser, do nothing and log out error
+ * If permission API is not supported on this browser, permission state is set to unsupported.
  * @private
  */
-export const getDevicePermissionState = (
+export const getDevicePermissionStateAsync = async (
```
JamesBurnside, line 324: nit: the suffix `Async` isn't needed as the function is now marked async and returns a Promise (but feel free to keep if you like, we have examples in the codebase both with and without the suffix)

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -289,35 +289,44 @@ export const isDisabled = (option: boolean | { disabled: boolean } | undefined):
...
 };
 
+/**
+ * @returns Permissions state for the camera.
+ */
+const queryCameraPermissionFromPermissionsAPI = async (): Promise<PermissionState | 'unsupported'> => {
+  try {
+    return (await navigator.permissions.query({ name: 'camera' as PermissionName })).state;
+  } catch (e) {
+    console.warn('permissions API is not supported by browser', e);
```
JamesBurnside, line 299: don't warn here, we expect the scenario where it fails and return an appropriate value. You can `console.info` this detail if you feel it adds value to know but we don't want to give developers warnings when they use the function in an intended way

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
JamesBurnside, line 324: nit: the prefix `get` is still misleading here, the function doesn't return the values but instead calls the setters passed in, consider `updateDevicePermissionsState`

JamesBurnside (approved):
Left some tiny ignorable nits, this looks great! Thanks for making those updates!

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
dmceachernmsft: I think they should get some information so I will do a console.info()

PR #2610, opened 2022-12-19
Disable pin menu button in VideoGallery when max pinned participants is reached

mgamis-msft:
# What

- Added `maxPinnedRemoteVideoTiles` prop to VideoGallery.
- Disabling pin menu button on remote VideoTile of VideoGallery when max pinned remote video tiles is reached
- Added enzyme tests for pinned participants enzyme tests including those that test prop `maxPinnedRemoteVideoTiles`
- Added 1 hermetic playwright tests to test that CallComposite will limit max pinned remote video tiles to 4
- Added a revision to API View for Pinned Participants here: https://apiview.dev/Assemblies/Review/7276a0d82fce4755b833fc91def19dbc/8a35753f996445c28156396635624f28?diffRevisionId=80605558e3ea4faa8acc2db2edb2e3fa&doc=False&diffOnly=True

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3043119
https://skype.visualstudio.com/SPOOL/_workitems/edit/3050025
https://skype.visualstudio.com/SPOOL/_workitems/edit/3045927
https://skype.visualstudio.com/SPOOL/_workitems/edit/3050071

# How Tested

Local calling sample, storybook, and through enzyme tests and 1 playwright test

# Process & policy checklist


JamesBurnside:
LGTM! Just waiting for green CI before approving. Added update_snapshots label to get snapshots of the new e2e test

packages/react-composites/tests/browser/call/hermetic/VideoGallery.test.ts
```diff
@@ -73,4 +73,26 @@ test.describe('VideoGallery tests', async () => {
...
+    const displayNames = ['Tony Hawk', 'Marie Curie', 'Gal Gadot', 'Margaret Atwood', 'Kobe Bryant', "Conan O'Brien"];
+    const participants = displayNames.map((name) => defaultMockRemoteParticipant(name));
+    const initialState = defaultMockCallAdapterState(participants);
+
+    await page.goto(buildUrlWithMockAdapter(serverUrl, initialState));
+
+    // pin remote video tiles up to the max allowed in the call composite
+    for (let i = 0; i < 4; i++) {
+      // click the more button of an unpinned remote video tile
+      await pageClick(page, dataUiId('video-tile-more-options-button') + ' >> nth=-1');
```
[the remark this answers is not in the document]
JamesBurnside: Looks like this is also causing the update snapshots to fail:
image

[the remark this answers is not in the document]
mgamis-msft: I am trying to click the last instance of the more options button which is presumably in an unpinned remote video tile but for some reason this does not work in CI. 
Running this test locally I am able to get this snapshot.
disabled-pin-menu-button-win32

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -3211,6 +3211,7 @@ export interface VideoGalleryProps {
     localParticipant: VideoGalleryLocalParticipant;
     localVideoCameraCycleButtonProps?: LocalVideoCameraCycleButtonProps;
     localVideoViewOptions?: VideoStreamOptions;
+    maxPinnedRemoteVideoTiles?: number;
```
anjulgarg, line 3214: I am not sure if we ever discussed this prop as part of the public API changes.
Do we really need to introduce this?
I was under the impression that VideoGallery was going to enforce default of 4

JamesBurnside: If its helpful I really like this prop! In there a reason to limit to 4?

JamesBurnside:
I actually worry about disabling the button - it's not clear to the user _why_ it's disabled and how they can go about fixing that, can you check in with design on this?

JamesBurnside:
nit: the pin icon doesn't quite have the correct padding around it, the text is closer to the pin than the pin is to the left hand side:
image

packages/communication-react/review/beta/communication-react.api.md
mgamis-msft: I can hardcode max limit of 4 pinned remote video tiles to disable the pin menu item just to avoid the API change. And we can add this prop in the future from customer signals by reverting the commit hash on this branch.

packages/react-composites/tests/browser/call/hermetic/VideoGallery.test.ts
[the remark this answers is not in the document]
mgamis-msft: Issue fixed. This fails on mobile UI tests because ellipses button is not shown on mobile. Fixed.

mgamis-msft:
> nit: the pin icon doesn't quite have the correct padding around it, the text is closer to the pin than the pin is to the left hand side: image
I think the snapshot you saw might have been outdated because the margins look fine to me.
image
I looked at the margins on storybook and they seem ok as well.
image

JamesBurnside:
Going by the [latest snapshot update](https://github.com/Azure/communication-ui-library/pull/2610/commits/eb94476cf0be744c052acabc982bbc9a27368d88) the horizontal margins are still a little off (by ~30% depending on resolution), and the vertical direction isn't center aligned:
image
^ The icon should be vertically centered with the text and not bottom-aligned?

I see the same issue in storybook:
image


For reference this is what Teams' looks like:
image

Icon alignment is tricky (fluent icons aren't the best). Happy to defer to design and have them sign off or not on if this is an acceptable level of alignment

mgamis-msft:
> Going by the [latest snapshot update](https://github.com/Azure/communication-ui-library/pull/2610/commits/eb94476cf0be744c052acabc982bbc9a27368d88) the horizontal margins are still a little off (by ~30% depending on resolution), and the vertical direction isn't center aligned: image ^ The icon should be vertically centered with the text and not bottom-aligned?
> 
> I see the same issue in storybook: image
> 
> For reference this is what Teams' looks like (notice the pin icon extends ~the same amount above and below the P in Pin): image
> 
> Icon alignment is tricky (fluent icons aren't the best). Happy to defer to design and have them sign off or not on if this is an acceptable level of alignment

I was able to change the icon size so that the icon is vertically aligned with the menu text. But there is now a little more space on the right side due that natural padding of the span element which is what is the menu text is.
image
image
image. Let me know if I should file a bug to fix this. Thanks.

mgamis-msft:
> I actually worry about disabling the button - it's not clear to the user _why_ it's disabled and how they can go about fixing that, can you check in with design on this?

Talked with Alex P and we are now indicating that the limit is reached. Please see updated snapshot.

JamesBurnside:
>  was able to change the icon size so that the icon is vertically aligned with the menu text

The icon in the snapshot is still not vertically centered:
image

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -174,6 +174,7 @@
     "fitRemoteParticipantToFrame": "Fit to frame",
     "fillRemoteParticipantFrame": "Fill frame",
     "pinParticipantForMe": "Pin for me",
+    "pinParticipantForMeWhenLimitReached": "Pin (limit reached)",
```
JamesBurnside, line 177: nit: remove the `when` here, just `pinParticipantForMeLimitReached`

JamesBurnside (approved):

change-beta/@azure-communication-react-12a92b59-bacd-414a-927d-fe69b009f6cf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Added maxPinnedRemoteVideoTiles prop to VideoGallery. Disabling pin menu button on remote VideoTile of VideoGallery when max pinned remote video tiles is reached",
```
prabhjot-msft, line 3: As we are not adding the `maxPinnedRemoteVideoTiles ` prop, the comment seems misleading.

change/@azure-communication-react-3bfec36c-0908-4eb5-b99a-f83fdc6d7f26.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Added maxPinnedRemoteVideoTiles prop to VideoGallery. Disabling pin menu button on remote VideoTile of VideoGallery when max pinned remote video tiles is reached.",
```
prabhjot-msft, line 3: same as above

packages/react-composites/tests/browser/call/hermetic/VideoGallery.test.ts
```diff
@@ -73,4 +81,36 @@ test.describe('VideoGallery tests', async () => {
...
+      // TODO: Test that pin menu item is disabled when maximum remote VideoTiles are pinned in VideoGallery when
+      // drawer menu on long touch has been implemented
+      return;
+    }
+
+    // pin remote video tiles up to the max allowed in the call composite
+    for (let i = 0; i < 4; i++) {
+      // click the last more button on the page which is presumably on an unpinned remote video tile
+      const moreOptionsButton = page.locator(dataUiId('video-tile-more-options-button')).last();
+      await screenshotOnFailure(page, async () => await moreOptionsButton.click({ timeout: perStepLocalTimeout() }));
```
prabhjot-msft, line 106: we can use `waitforselector` helper method to avoid this repetitive code as it takes care of screenshot on failure

packages/react-composites/tests/browser/call/hermetic/VideoGallery.test.ts
```diff
@@ -73,4 +81,36 @@ test.describe('VideoGallery tests', async () => {
...
+      // click pin menu button in contextual menu
+      await pageClick(page, dataUiId('video-tile-pin-participant-button'));
+    }
+    // click the last more button on the page which is presumably on an unpinned remote video tile
+    const moreOptionsButton = page.locator(dataUiId('video-tile-more-options-button')).last();
+    await screenshotOnFailure(page, async () => await moreOptionsButton.click({ timeout: perStepLocalTimeout() }));
```
prabhjot-msft, line 112: same as above

prabhjot-msft (approved):

PR #2611, opened 2022-12-19
Added `remoteVideoTileMenuOptions` in VideoGallery. Added `showRemoteVideoTileMenu` prop to CallCompositeOptions.

mgamis-msft:
# What

- Added `remoteVideoTileMenuOptions` prop to replace prop `showRemoteVideoTileContextualMenu` in VideoGallery to allow developers using VideoGallery to define the remote video tile menu 
- Added `showRemoteVideoTileMenu` prop to CallCompositeOptions to allow developers using the CallComposite to not show remote video tile menu in the VideoGallery
- Reflected these additions in API View revisions for ARB here: https://apiview.dev/Assemblies/Review/7276a0d82fce4755b833fc91def19dbc/8a35753f996445c28156396635624f28?diffRevisionId=80605558e3ea4faa8acc2db2edb2e3fa&doc=False&diffOnly=True

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3091713

# How Tested

Tested with locally on Calling sample

# Process & policy checklist

### DIFF
--- a/packages/communication-react/review/beta/communication-react.api.md
+++ b/packages/communication-react/review/beta/communication-react.api.md
@@ -478,6 +478,7 @@ export type CallCompositeOptions = {
     }) => void;
     onNetworkingTroubleShootingClick?: () => void;
     onEnvironmentInfoTroubleshootingClick?: () => void;
+    remoteVideoTileMenu?: RemoteVideoTileMenuOptions;
 };
 
 // @public
@@ -2823,6 +2824,11 @@ export interface RemoteVideoStreamState {
 }
 
 // @beta
+export interface RemoteVideoTileMenuOptions {
+    isHidden?: boolean;
+}
+
+// @beta
 export type Role = 'Presenter' | 'Attendee' | 'Consumer' | 'Organizer';
 
 // @public
@@ -3201,10 +3207,10 @@ export interface VideoGalleryProps {
     onUnpinParticipant?: (userId: string) => void;
     pinnedParticipants?: string[];
     remoteParticipants?: VideoGalleryRemoteParticipant[];
+    remoteVideoTileMenuOptions?: false | VideoTileContextualMenuProps | VideoTileDrawerMenuProps;
     remoteVideoViewOptions?: VideoStreamOptions;
     showCameraSwitcherInLocalPreview?: boolean;
     showMuteIndicator?: boolean;
-    showRemoteVideoTileContextualMenu?: boolean;
     strings?: Partial<VideoGalleryStrings>;
     styles?: VideoGalleryStyles;
 }
@@ -3272,6 +3278,17 @@ export interface VideoStreamRendererViewState {
 
 // @public
 export const VideoTile: (props: VideoTileProps) => JSX.Element;
+
+// @beta
+export interface VideoTileContextualMenuProps {
+    kind: 'contextual';
+}
+
+// @beta
+export interface VideoTileDrawerMenuProps {
+    hostId?: string;
+    kind: 'drawer';
+}
 
 // @public
 export interface VideoTileProps {
--- a/packages/communication-react/review/beta/communication-react.api.md
+++ b/packages/communication-react/review/beta/communication-react.api.md
@@ -15,6 +15,7 @@ import { CallClientOptions } from '@azure/communication-calling';
 import { CallDirection } from '@azure/communication-calling';
 import { CallEndReason } from '@azure/communication-calling';
 import { CallerInfo } from '@azure/communication-calling';
+import { CallKind } from '@azure/communication-calling';
 import { CallState as CallState_2 } from '@azure/communication-calling';
 import { ChatClient } from '@azure/communication-chat';
 import { ChatClientOptions } from '@azure/communication-chat';
@@ -699,6 +700,7 @@ export interface CallState {
     id: string;
     isMuted: boolean;
     isScreenSharingOn: boolean;
+    kind: CallKind;
     localVideoStreams: LocalVideoStreamState[];
     recording: RecordingCallFeature;
     remoteParticipants: {
@@ -712,7 +714,6 @@ export interface CallState {
     startTime: Date;
     state: CallState_2;
     transcription: TranscriptionCallFeature;
-    type: 'Teams' | 'ACS';
 }
 
 // @public
--- a/packages/react-components/src/components/VideoGallery.tsx
+++ b/packages/react-components/src/components/VideoGallery.tsx
@@ -44,6 +44,13 @@ export const DEFAULT_MAX_REMOTE_VIDEO_STREAMS = 4;
  * Set aside only 6 dominant speakers for remaining audio participants
  */
 export const MAX_AUDIO_DOMINANT_SPEAKERS = 6;
+/**
+ * @private
+ * Default remote video tile menu options
+ */
+export const DEFAULT_REMOTE_VIDEO_TILE_MENU_OPTIONS = {
+  kind: 'contextual'
+};
 
 /**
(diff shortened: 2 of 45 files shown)


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -3294,6 +3295,17 @@ export interface VideoStreamRendererViewState {
...
 export const VideoTile: (props: VideoTileProps) => JSX.Element;
 
+// @beta
+export interface VideoTileContextualMenuProps {
+    type: 'contextual';
+}
+
+// @beta
+export interface VideoTileDrawerMenuProps {
+    hostId?: string;
```
JamesBurnside, line 3305: context menu also has a hostId, would be good to have this for VideoTileContextualMenuProps also

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -3294,6 +3295,17 @@ export interface VideoStreamRendererViewState {
 // @public
 export const VideoTile: (props: VideoTileProps) => JSX.Element;
 
+// @beta
+export interface VideoTileContextualMenuProps {
+    type: 'contextual';
```
JamesBurnside, line 3300: suggestion: `kind` instead of `type` - type is already a keyword in javascript so typically tend to avoid naming variables the same as keywords

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -479,6 +479,7 @@ export type CallCompositeOptions = {
     }) => void;
     onNetworkingTroubleShootingClick?: () => void;
     onEnvironmentInfoTroubleshootingClick?: () => void;
+    showRemoteVideoTileMenu?: boolean;
```
JamesBurnside, line 482: So is the plan to make this `opt-in`?

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -181,8 +182,45 @@ export interface VideoGalleryProps {
...
+ * @beta
+ */
+export interface VideoTileDrawerMenuProps {
+  /**
+   * The menu property type
+   */
+  type: 'drawer';
+  /**
+   * The optional id property provided on an element that the drawer menu should render within when a
+   * remote participant video tile is long touched. If an id is not provided, then a drawer menu will
```
JamesBurnside, line 219: remove `is long touched` here in this comment, that behavior isn't specific to this property and could change
```suggestion
  * remote participant video tile drawer menu is shown. If an id is not provided, then a drawer menu will
```

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -164,6 +164,13 @@ export type CallCompositeOptions = {
    * if this is not supplied, the composite will not show a unsupported browser page.
    */
   onEnvironmentInfoTroubleshootingClick?: () => void;
+  /* @conditional-compile-remove(pinned-participants) */
+  /**
+   * If set to false, remote video tiles in the VideoGallery will not have menu options
+   *
+   * default: true
+   */
+  showRemoteVideoTileMenu?: boolean;
```
JamesBurnside, line 173: from review board. We always try default to false (as then it is the same as not supplying the property), so this should be `hideRemoteVideoTileMenu` with `@default false`

JamesBurnside:
Not this PR, but a bug in the 3 dots, they're below the existing participant name text:
image

And also seem either slightly bigger font or more bolded? (maybe because it's an ellipsis icon and not the font?)

JamesBurnside:
If mobile is long press, does it still show the context menu?
image

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
mgamis-msft: This should apply to both I believe.

mgamis-msft:
> If mobile is long press, does it still show the context menu? image

Yes, you are right. My boolean logic was off. Fixed.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -350,7 +391,9 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
           /* @conditional-compile-remove(PSTN-calls) */
           participantState={participant.state}
           /* @conditional-compile-remove(pinned-participants) */
-          showRemoteVideoTileContextualMenu={props.showRemoteVideoTileContextualMenu}
+          showRemoteVideoTileContextualMenu={
+            remoteVideoTileMenuOptions && remoteVideoTileMenuOptions.kind === 'contextual'
```
mgamis-msft, line 395: I don't the 'remoteVideoTileMenuOptions &&'

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -91,12 +95,20 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
         showCameraSwitcherInLocalPreview={props.isMobile}
         localVideoCameraCycleButtonProps={cameraSwitcherProps}
         onRenderAvatar={onRenderAvatar}
-        /* @conditional-compile-remove(pinned-participants) */
-        showRemoteVideoTileContextualMenu={!props.isMobile}
         // @TODO: Provide props.drawerMenuHostId to VideoGallery when VideoGallery props support it.
+        /* @conditional-compile-remove(pinned-participants) */
+        remoteVideoTileMenuOptions={
+          props.remoteVideoTileMenuOptions?.isHidden ? false : props.isMobile ? false : { kind: 'contextual' }
```
mgamis-msft, line 101: Simplify this double ternary. Also when it is mobile assign { kind: 'drawer', hostId: <someHostId> }.

JamesBurnside (approved):

packages/react-components/src/components/VideoGallery.tsx
prabhjot-msft, line 395: do we need a check here for `remoteVideoTileMenuOptions` as it will be always available as default options

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -1269,6 +1270,11 @@ export type ParticipantsRemovedListener = (event: {
     removedBy: ChatParticipant;
 }) => void;
 
+// @beta
+export interface RemoteVideoTileMenuOptions {
+    isHidden?: boolean;
```
prabhjot-msft, line 1275: is there a reason we have isHidden optional as `remoteVideoTileMenu ` is already optional, so in case it is passed it should have `isHidden` value

dmceachernmsft (approved):

packages/react-components/src/components/VideoGallery.tsx
mgamis-msft: I had the same thought but yes because remoteVideoTileMenuOptions is a union type `false | VideoTileContextualMenuProps | VideoTileDrawerMenuProps` and could be `false`.

packages/communication-react/review/beta/communication-react.api.md
mgamis-msft: Good point but I will leave it off for now and wait for customer signals.

packages/react-composites/review/beta/react-composites.api.md
mgamis-msft: We chose to have a prop `isHidden` in an object to allow for future options to `remoteVideoTileMenu` for the CallComposite. See this discussion in this PR: https://github.com/Azure/communication-ui-library/pull/2611#discussion_r1052444394

PR #2615, opened 2022-12-28
Update calling stateful client and CallAdapter state to be readonly

dmceachernmsft:
# What

Adds readonly properties to the CallAdapter adapter state
# Why

Makes adapter state like pages protected from direct assignment.

https://skype.visualstudio.com/SPOOL/_workitems/edit/2981275
# How Tested

updated tests for readonly state

### DIFF
--- a/packages/calling-stateful-client/src/CallContext.ts
+++ b/packages/calling-stateful-client/src/CallContext.ts
@@ -200,10 +200,11 @@ export class CallContext {
   }
 
   /* @conditional-compile-remove(unsupported-browser) */
-  public setEnvironmentInfo(envInfo: EnvironmentInfo): void {
-    this.modifyState((draft: CallClientState) => {
-      draft.environmentInfo = envInfo;
-    });
+  public setEnvironmentInfo(environmentInfo: EnvironmentInfo): void {
+    this.modifyState((draft: CallClientState) => ({
+      ...draft,
+      environmentInfo
+    }));
   }
 
   public setCallIsScreenSharingOn(callId: string, isScreenSharingOn: boolean): void {
--- a/packages/calling-stateful-client/src/CallContext.ts
+++ b/packages/calling-stateful-client/src/CallContext.ts
@@ -116,18 +116,20 @@ export class CallContext {
   // make sure the state is clean because any left over state (if previous CallAgentDeclarative was disposed) may be
   // invalid.
   public clearCallRelatedState(): void {
-    this.modifyState((draft: CallClientState) => {
-      draft.calls = {};
-      draft.incomingCalls = {};
-      draft.callsEnded = {};
-      draft.incomingCallsEnded = {};
-    });
+    this.modifyState((draft: CallClientState) => ({
+      ...draft,
+      calls: {},
+      incomingCalls: {},
+      callsEnded: {},
+      incomingCallsEnded: {}
+    }));
   }
 
   public setCallAgent(callAgent: CallAgentState): void {
-    this.modifyState((draft: CallClientState) => {
-      draft.callAgent = callAgent;
-    });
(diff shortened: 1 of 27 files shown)


packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -201,12 +201,12 @@ export interface CallAdapterCallOperations {
 
 // @public
 export type CallAdapterClientState = {
-    userId: CommunicationIdentifierKind;
-    displayName?: string;
-    call?: CallState;
+    readonly userId: CommunicationIdentifierKind;
+    readonly displayName?: string;
+    readonly call?: CallState;
     devices: DeviceManagerState;
```
edwardlee-msft, line 207: Particular reason why devices and latestErrors are not readonly as well?

alkwa-msft (approved):

edwardlee-msft:
Nice! Thanks for updating the tests as well with these readonly changes.

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -116,18 +116,20 @@ export class CallContext {
   // make sure the state is clean because any left over state (if previous CallAgentDeclarative was disposed) may be
   // invalid.
   public clearCallRelatedState(): void {
-    this.modifyState((draft: CallClientState) => {
-      draft.calls = {};
-      draft.incomingCalls = {};
-      draft.callsEnded = {};
-      draft.incomingCallsEnded = {};
-    });
+    this.modifyState((draft: CallClientState) => ({
```
edwardlee-msft, line 119: I love and hate that this works haha
My brain says no but logic says yes.  🤪

edwardlee-msft (approved):

dmceachernmsft:
waiting to merge based on feedback on whether this is a breaking change.

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -116,18 +116,20 @@ export class CallContext {
...
+      ...draft,
+      calls: {},
```
PorterNan, line 121: Hmmm, this will break the functionality of immutable and make it not working as intended, draft.[property] = [value] is not just a simple assignment expression, it is a setter function overriden by immer library, not calling it will cause immer.js cannot detect some deep changes in the state tree, which might cause some unexpected error/perf regressions

PorterNan (changes requested):
I 100% agree that we should set a limit that can block developers from changing state directly, and this is a tricky problem because of several reasons:
1. Currently we don't expose a modifyState function for our developers to modify state - so they try some interesting solutions like get state then modify state directly
2. We have immer which needs directly assign value to an object property (it is actually a setter method override, check https://immerjs.github.io/immer/#how-immer-works)
3. Even we do #1, then make CallClientState to readonly will cause problems if they wanna do some modifications using immer

I would consider this as a breaking change, this is already impacting some of our internal implementation, and if developers follow our implementation, they will have to change their code as well

And also, I would say this is more like a feature we need to take step by step to resolve:
1. Expose modifyState function for our developer to modify state, so they don't have to find a tricky solution by themselves
2. After we have 1, consider use statefulClient.getState() -> Readonly<ClientState>, we only restrict return value by getState() function not restrict the whole state type

packages/calling-stateful-client/src/CallContext.ts
dmceachernmsft: I was reading the docs on immer, it seemed like this reassignment would also work. if the state of the draft is updated in any way, when the function terminates it updates the newState to what was changed. or am I misunderstanding something?

dmceachernmsft:
Closing as this is now a KTLO item and needs more investigation to make sure we are doing this properly and not breaking Contoso.

PR #2618, opened 2022-12-29
Update chat stateful client state and adapter to be readonly

dmceachernmsft:
# What

Make state readonly in the Adapter and Chat statefulClient
# Why

Makes the state of the two levels immutable

https://skype.visualstudio.com/SPOOL/_workitems/edit/2981275
related: https://github.com/Azure/communication-ui-library/pull/2615
# How Tested

Ran apps locally to verify client and adapters are still working


dmceachernmsft:
Closing this as its now part of a KTLO item and more work needs to be done to make sure we aren't breaking Contoso.

PR #2619, opened 2022-12-30
[bugfix] Update storybook and samples to disable mobile feature "pull-down to refresh"

edwardlee-msft:
# What
Disable pull-down to refresh on samples and document best practice in storybook on how to handle accidental pull down to refresh triggers on mobile

# Why
On IOS Safari Chat and CallWithChat scrolling on chat page can sometimes cause the whole page view to refresh/scroll.
This can cause accidental page refresh.
https://skype.visualstudio.com/SPOOL/_workitems/edit/3069029

# How Tested
Tested on iPhone SE on CallWIthChat and Chat sample.

# Process & policy checklist

### DIFF
--- a/samples/CallWithChat/src/app/views/CallScreen.tsx
+++ b/samples/CallWithChat/src/app/views/CallScreen.tsx
@@ -36,6 +36,16 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     locator,
     /* @conditional-compile-remove(PSTN-calls) */ alternateCallerId
   } = props;
+
+  // Disables pull down to refresh. Prevents accidental page refresh when scrolling through chat messages
+  // Another alternative: set body style touch-action to 'none'. Achieves same result.
+  useEffect(() => {
+    document.body.style.overflow = 'hidden';
+    return () => {
+      document.body.style.overflow = 'null';
+    };
+  }, []);
+
   const callIdRef = useRef<string>();
   const { currentTheme, currentRtl } = useSwitchableFluentTheme();
   const isMobileSession = useIsMobile();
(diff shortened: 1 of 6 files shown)


samples/CallWithChat/src/app/views/CallScreen.tsx
```diff
@@ -36,6 +36,16 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
     locator,
     /* @conditional-compile-remove(PSTN-calls) */ alternateCallerId
   } = props;
+
+  // Disables pull down to refresh. Prevents accidental page refresh when scrolling through chat messages
+  // Another alternative: set body style touch-action to 'none'. Achieves same result.
+  useEffect(() => {
+    document.body.style.overflow = 'hidden';
```
dmceachernmsft, line 43: does this disable pulldown behavior because there is no overflow on the page?

edwardlee-msft: Yes, essentially on safari ios there is no explicit option to disable pull down to refresh. 
The intermediate solution is to either apply a 
overflow: 'hidden' or a touch-action: 'hidden' to the body element to prevent a pull-down from occurring at all.

edwardlee-msft: I did not apply this fix to the composites and have only applied the fix to the samples as this can 100% affect Contosos use cases if integrated as a modal, iframe, etc..

I added instructions into storybook for best practices for mobile.

dmceachernmsft (approved):

JamesBurnside (approved):

PR #2621, opened 2022-12-30
[Bugfix] Fix video tile strict mode findDOMNode warning

dmceachernmsft:
# What

Remove use of `<Ref>` tag from fluent nothstar
# Why

fixes findDOMNode warning caused by strict mode in the video tile. Under the hood the `<Ref>` tag probably uses `findDOMNode()`. According to react's documentation about the function it should not be used with React.FunctionalComponents. `<Ref>` is a React.FunctionalComponent according the the api file provided by fluent-northstar.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3090254
# How Tested

Ran app locally with new reference structure to make sure the video gallery still works

### DIFF
--- a/packages/react-components/src/components/VideoTile.tsx
+++ b/packages/react-components/src/components/VideoTile.tsx
@@ -246,7 +246,6 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
   const observer = useRef(
     new ResizeObserver((entries): void => {
       const { width, height } = entries[0].contentRect;
-      console.log(entries);
       const personaSize = Math.min(width, height) / 3;
       setPersonaSize(Math.max(Math.min(personaSize, personaMaxSize), personaMinSize));
     })
--- a/packages/react-components/src/components/VideoTile.tsx
+++ b/packages/react-components/src/components/VideoTile.tsx
@@ -4,7 +4,6 @@
 import { Icon, IStyle, mergeStyles, Persona, Stack, Text } from '@fluentui/react';
 /* @conditional-compile-remove(pinned-participants) */
 import { IconButton } from '@fluentui/react';
-import { Ref } from '@fluentui/react-northstar';
 import React, { useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { useIdentifiers } from '../identifiers';
 import { ComponentLocale, useLocale } from '../localization';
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/VideoTile.tsx
```diff
@@ -307,30 +306,30 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+            borderRadius: theme.effects.roundedCorner4,
+            width: '100%',
+            height: '100%'
+          }
+        },
+        styles?.root
+      )}
+      {...longPressHandlersTrampoline}
+    >
+      <div ref={videoTileRef} style={{ width: '100%', height: '100%' }}>
```
PorterNan, line 332: Nice, we should do get rid of dependencies of northstar more and more

Btw, we do we not wrap the Stack element anymore?

samples/Calling/src/app/App.tsx
```diff
@@ -191,18 +191,20 @@ const App = (): JSX.Element => {
...
-          displayName={displayName}
-          callLocator={callLocator}
-          /* @conditional-compile-remove(PSTN-calls) */
-          alternateCallerId={alternateCallerId}
-          /* @conditional-compile-remove(rooms) */
-          roleHint={role}
-          /* @conditional-compile-remove(teams-identity-support) */
-          isTeamsIdentityCall={isTeamsCall}
-        />
+        <React.StrictMode>
```
PorterNan, line 194: I guess this StrictMode element was added for testing purposes? Remove it if there is no specific reason to add this

dmceachernmsft: yup this was added so that should this come up again on the calling side we will know.

packages/react-components/src/components/VideoTile.tsx
dmceachernmsft: so the stack that is wrapping this div is what is setting the height for the tile. I moved the div that is setting the ref inside so that the tile renders as normal. For some reason when the div was outside it would break the video tile and it wouldn't have a height being passed to the reference.

mgamis-msft: nit: put style { width: '100%', height: '100%' } in VideoTile.styles.ts 😅

mgamis-msft (approved):

packages/react-components/src/components/VideoTile.tsx
PorterNan: I guess Ref doesn't create any new element but div does? And that could be the reason, LGTM if this would not create any regressions, but move the style tag into a separated file (./styles/VideoTile.styles)

PorterNan (approved):

PR #2622, opened 2022-12-31
prototype snippet component code

alkwa-msft:
Creating a new wrapper component for our <Source> tag usage when showing snippets.

type Args = {
  key: string;
  code: string;
  startLine: number;
  endLine: number;
  githubUrl: string;
};

we take in these args into our new "snippet" component.

We still pass in the code similar to the Source component. The difference is that we can pass in a start/end line to constrain our code to a specific set of lines in the snippet. Also including a githubUrl where we would link to the snippet in github with the specific lines hightlighted to show where the lines in the snippet fit in the overall working code


JamesBurnside:
I like this! Some notes if we want to make this snippet component. The identation needs fixed:
image

And the text below it needs wrapped in a fluent `<Text>` block so it is the correct font:
image

PR #2624, opened 2023-01-03
[Bugfix] Bound CallReadiness screens to the boundaries of the container

edwardlee-msft:
# What
Use ModalClone and a Layerhost instead of FluentUi provided Modal. 
image

# Why
Modal takes up full width and height of body and not the container it's called within.
https://skype.visualstudio.com/SPOOL/_workitems/edit/3086034
image

# How Tested
Locally on MacOS Safari

# Process & policy checklist


edwardlee-msft:
Note that the changes include an "align-items" prop which causes the width of the modal to be a bit thinner than before:
image

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -82,6 +84,7 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
   const [audioState, setAudioState] = useState<PermissionState | 'unsupported' | undefined>(undefined);
   /* @conditional-compile-remove(call-readiness) */
   getDevicePermissionState(setVideoState, setAudioState);
+  const modalLayerHostId = useId('callReadinessModalLayerhost');
```
JamesBurnside, line 87: We already have a composite-root level modal layer host id, can we use this instead so all modals, context menus etc. going forward all bind to the same root:
https://github.com/Azure/communication-ui-library/blob/d7d036e2d316fdc8081c3858b5a2a0bbf9db7b1b/packages/react-composites/src/composites/CallComposite/CallComposite.tsx#L377

edwardlee-msft: Good idea, I confirmed with FluentUi modal examples that multiple modals use 1 root modal layer host. 
I'll make the change now!

edwardlee-msft: Alright! Done, needed to play around with some transform in the modal but got it to work with the current layerhost. 
Also removed mobileView only for that specific layerhost as I found that it was not needed during my testing with the PIP modal. (but needs to be removed for call readiness modal to render on desktop using that layerhost).

JamesBurnside (approved):

PR #2626, opened 2023-01-04
Add displayName param of adapter

PorterNan:
- Add a required displayName parameter for adapter
- Add an optional displayName of StatefulClient for CTE
- Update the sample code to support it in sample

# What

# Why

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/StatefulCallClient.ts
```diff
@@ -193,7 +193,9 @@ class ProxyCallClient implements ProxyHandler<CallClient> {
             const callAgent = await target.createCallAgent(...args);
             this._callAgent = callAgentDeclaratify(callAgent, this._context, this._internalContext);
```
alkwa-msft, line 194: is it possible for teamsCallAgent to also take in the displayname? we take in a displayname in two different ways based on if its teams or acs?

alkwa-msft: I suppose its ok if we are only going into public preview as we could change this in the future. Its just odd that we have two different ways of setting display name if your using acs or teams

samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -263,23 +263,21 @@ export const HomeScreen = (props: HomeScreenProps): JSX.Element => {
...
             onClick={() => {
-              if (displayName || /* @conditional-compile-remove(teams-identity-support) */ teamsIdentityChosen) {
+              if (displayName) {
                 displayName && saveDisplayNameToLocalStorage(displayName);
                 /* @conditional-compile-remove(one-to-n-calling) */
                 const acsParticipantsToCall = parseParticipants(outboundParticipants);
                 /* @conditional-compile-remove(PSTN-calls) */
                 const dialpadParticipantToCall = parseParticipants(dialPadParticipant);
                 props.startCallHandler({
                   //TODO: This needs to be updated after we change arg types of TeamsCall
```
dmceachernmsft, line 279: is this TODO still something we need to worry about?

dmceachernmsft (approved):

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1664,7 +1665,7 @@ export const createStatefulCallClient: (args: StatefulCallClientArgs, options?:
 export const createStatefulChatClient: (args: StatefulChatClientArgs, options?: StatefulChatClientOptions | undefined) => StatefulChatClient;
 
 // @beta (undocumented)
-export const createTeamsCallAdapter: ({ userId, credential, locator }: TeamsCallAdapterArgs) => Promise<TeamsCallAdapter>;
+export const createTeamsCallAdapter: ({ userId, credential, displayName, locator }: TeamsCallAdapterArgs) => Promise<TeamsCallAdapter>;
```
JamesBurnside, line 1668: from discussion that CTE <--> CTE callees won't see each others' names do we still want to go ahead with this change?

PorterNan:
Close this one because we are seeking for no displayname by default + displayName override

PR #2630, opened 2023-01-05
Fix empty string bug in createAnnouncmentString()

PorterNan:
# What

The 2 sets of array - unnamedParticipants + namedParicipants should exactly match the original participants array.
But there is one case missing - participantName === '' 
So if any empty string participant appears, this bug will cause unexpected exception.

Change the condition a bit to include '' empty string in unnamedParticipant
# Why

# How Tested

# Process & policy checklist


edwardlee-msft:
Create mini fix! 
Note for other reviewers:
Empty strings are "falsy" which means they are considered false in a Boolean context, so you can just use not string

edwardlee-msft (approved):

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/utils/MediaGalleryUtils.ts
```diff
@@ -100,7 +100,7 @@ export const createAnnouncmentString = (
    * Filter participants into two arrays to put all the unnamed participants at the back of the
    * names being announced.
    */
-  const unnamedParticipants = participants.filter((p) => p.displayName === undefined);
+  const unnamedParticipants = participants.filter((p) => !p.displayName);
```
dmceachernmsft, line 103: How does this also cover the case when the `displayName === ''`?

PorterNan: This might feel a bit strange, but in js  ```!'' = true```

PorterNan: @edwardlee-msft 's comment above explains this in a more detailed way 👍

PR #2633, opened 2023-01-05
Modify Avatar to person icon when no displayName

PorterNan:
# What

image

previously it is UM, even it is local participant, which feels a bit weird we give a Unamed Participant an initial avatar (local one), this modify will correct all the avatar for videoTile + participant list for "me"

image

# Why

# How Tested

# Process & policy checklist


JamesBurnside:
Could we add a hermetic test for this? (one of those areas I don't see us testing often)

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -282,7 +282,7 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
 
   const placeholderOptions = {
     userId,
-    text: initialsName || displayName,
+    text: initialsName !== undefined ? initialsName : displayName,
```
JamesBurnside, line 285: tiny nit: this can be written as 
```suggestion
    text: initialsName ?? displayName
```

JamesBurnside (approved):
Looks great! One small suggestion just to add a e2e test for this

packages/react-components/src/components/VideoTile.tsx
PorterNan: Nice, small tip I can repeat here for myself to remember it better 
The difference from || and ?? for string
 1. '' ?? displayName = ''
 2. '' || displayName = displayName

PorterNan:
> Looks great! One small suggestion just to add a e2e test for this

Will open another PR on this once I merge this PR :D

edwardlee-msft (approved):

PorterNan:
Close this because our ci is not happy with single quote in the branch name

PR #2638, opened 2023-01-06
UI Snapshot Test - Verify clicking on VideoTile ellipses opens contextual menu in Desktop and test Pin/Unpin

prabhjot-msft:
# What
UI Snapshot Test - Verify clicking on VideoTile ellipses opens contextual menu in Desktop and test Pin/Unpin

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3021647

# How Tested

# Process & policy checklist


alkwa-msft (approved):

packages/react-composites/tests/browser/call/hermetic/VideoGalleryLayout.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  test
+} from './fixture';
+import { expect } from '@playwright/test';
+import { dataUiId, waitForSelector, stableScreenshot, isTestProfileMobile, pageClick } from '../../common/utils';
+import { IDS } from '../../common/constants';
+
+test.describe('VideoGalleryLayout tests', async () => {
+  /* @conditional-compile-remove(pinned-participants) */
+  test('VideoTile pin/unpin the remote participant', async ({ page, serverUrl }, testInfo) => {
+    test.skip(isTestProfileMobile(testInfo));
```
alkwa-msft, line 18: its not clear from this test that its only meant for desktop besides ignoring this for mobile. 

Should there be a mobile test as well?

prabhjot-msft: yes, there are separate tests for mobile drawer menu on long press. Modified the test title to make it clear that it is only for desktop

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1730,6 +1730,7 @@ export interface _Identifiers {
     typingIndicator: string;
     videoGallery: string;
     videoTile: string;
+    videoTileMoreOptionsButton: string;
```
dmceachernmsft, line 1733: This probably shouldn't be in the API file here. is there some missing CC?

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -192,12 +192,14 @@ const videoTileMoreMenuProps = {
 /* @conditional-compile-remove(pinned-participants) */
 const VideoTileMoreOptionsButton = (props: { contextualMenu?: IContextualMenuProps }): JSX.Element => {
   const { contextualMenu } = props;
+  const ids = useIdentifiers();
```
dmceachernmsft, line 195: is this a new hook that provides us with the data-ui-ids? is this context wrapping the composite internally? I haven't seen this yet? very cool!

mgamis-msft: It is a context wrapping the composite internally and has been in our code for a while :). `_IdentifierProvider` is internal so I think it is ok to assign data-ui-id props directly as a prop to components but we should have a team discussion about it

mgamis-msft (approved):

PR #2640, opened 2023-01-08
Scrollable horizontal gallery when VideoGallery is narrow

mgamis-msft:
# What

Change paginated horizontal gallery with navigation buttons to scrollable horizontal gallery when VideoGallery is narrow.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3094701

# How Tested

Local Calling sample. UI snapshots for Calling mobile should be updated.

# Process & policy checklist


packages/react-components/src/components/VideoGallery/styles/ScrollableHorizontalGallery.style.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export const scrollableHorizontalGalleryStyles: IStackStyles = {
+  root: {
+    width: '100%',
+    minHeight: `${SMALL_HORIZONTAL_GALLERY_TILE_SIZE_REM.height}rem`,
+    overflow: 'scroll',
+    paddingRight: '0.5rem',
+    '-ms-overflow-style': 'none',
+    'scrollbar-width': 'none',
+    '::-webkit-scrollbar': { display: 'none' },
```
JamesBurnside, line 21: concerned about this from an accessibility point of view - are there any a11y consequences to hiding the scrollbar or any a11y principles around scroll behavior? (I'll havea a google around this also and try find out, but would be worth asking the a11y team)

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
```diff
@@ -68,22 +70,34 @@ export const DefaultLayout = (props: DefaultLayoutProps): JSX.Element => {
...
+      return null;
+    }
+    /* @conditional-compile-remove(pinned-participants) */
+    if (isNarrow) {
+      return <ScrollableHorizontalGallery horizontalGalleryElements={horizontalGalleryTiles} />;
+    }
+    return (
+      <VideoGalleryResponsiveHorizontalGallery
+        isNarrow={isNarrow}
+        shouldFloatLocalVideo={true}
```
JamesBurnside, line 84: The local video looks to be occluding the other tiles (sits on top of them) but I cannot drag it off the tile to see behind it in storybook:
image

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
```diff
@@ -68,22 +70,34 @@ export const DefaultLayout = (props: DefaultLayoutProps): JSX.Element => {
...
+        horizontalGalleryElements={horizontalGalleryTiles}
+        styles={styles?.horizontalGallery}
+      />
+    );
+  }, [isNarrow, horizontalGalleryTiles, styles?.horizontalGallery]);
+
   return (
-    <Stack horizontal={false} styles={rootLayoutStyle}>
+    <Stack horizontal={false} styles={rootLayoutStyle} tokens={{ childrenGap: '0.5rem' }}>
```
JamesBurnside, line 92: Anywhere using `tokens={{ childrenGap: '0.5rem' }}`, pull the object into a const defined outside the render body (or useMemo it) to prevent this being a new object on every render
```suggestion
    <Stack horizontal={false} styles={rootLayoutStyle} tokens={tileGap}>
```

packages/react-components/src/components/VideoGallery/styles/ScrollableHorizontalGallery.style.ts
dmceachernmsft: So I got in touch with them and they are saying it required that it is visible. does Alex P have designs for what a scrollbar in the UI lib should look like?

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
JamesBurnside: From the figma this looks correct, but I assume only because it can be dragged away:
image

dmceachernmsft: should we then file this as a bug to finally make the local tile draggable in mobile again? I think it should be doable now that we have that modal

packages/react-components/src/components/VideoGallery/FloatingLocalVideoLayout.tsx
```diff
@@ -118,27 +120,38 @@ export const FloatingLocalVideoLayout = (props: FloatingLocalVideoLayoutProps):
       )
     ) : undefined;
 
+  const horizontalGallery = useMemo(() => {
+    if (horizontalGalleryTiles.length === 0) {
+      return null;
+    }
+    /* @conditional-compile-remove(pinned-participants) */
+    if (isNarrow) {
```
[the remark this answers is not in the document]
mgamis-msft: No, there is no prop for the VideoGallery to know if its on a mobile device. We can add isMobile prop to VideoGallery or some horizontalGallery config prop. But that would require an ARB review.
The isNarrow is determined from the width of the VideoGallery for now.

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
mgamis-msft: @JamesBurnside You can now drag the scrollable horizontal gallery with your mouse. I had to use this nifty package https://www.npmjs.com/package/react-use-draggable-scroll

packages/react-components/src/components/VideoGallery/styles/ScrollableHorizontalGallery.style.ts
mgamis-msft: I will let Alex P know about your concerns and what the A11y team says. Thank you.

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
JamesBurnside: I am still unable to drag it on storybook using mouse - are you able to?

https://user-images.githubusercontent.com/2684369/211616408-99b13386-b558-400e-8b7f-fae5b8f7e299.mp4

Is there a reason we needed to use [npmjs.com/package/react-use-draggable-scroll](https://www.npmjs.com/package/react-use-draggable-scroll) and not use the fluent modal -- the reason to try use the fluent modal is to avoid multiple implementations for the same behavior (each will have their own nuances/bugs so better if we can keep it consistent)

dmceachernmsft (approved):
Very cool functions for dragging the horizontal Gallery!

mgamis-msft:
I am aware that A11y team can raise issues about the lack if scrollbar. Alex P intends for there to be no scrollbar and points to Teams as an example where no scrollbar is present. If the A11y team, raises an issue we will address it by adding an appropriate scrollbar.

JamesBurnside (approved):

PR #2641, opened 2023-01-09
[BugFix] Add isCameraOn to CallAdapter state

dmceachernmsft:
# What

Add new state tracking local participants camera state to the CallAdapter
# Why

Allows Contoso to have better state control of the camera should they want to do something based on it like create their own custom control bar for the CallComposite.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3048158
# How Tested

Ran samples locally checking whether the state of the camera was being reflected appropriately in new variable.


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -182,7 +183,10 @@ class CallContext {
         page: newPage,
         endedCall: latestEndedCall,
         devices: clientState.deviceManager,
-        latestErrors: clientState.latestErrors
+        latestErrors: clientState.latestErrors,
+        isCameraOn:
+          clientState.deviceManager.unparentedViews.find((s) => s.mediaStreamType === 'Video') !== undefined ||
+          this.getState().call?.localVideoStreams.find((s) => s.mediaStreamType === 'Video') !== undefined
```
JamesBurnside, line 189: dont use `getState()` here, just use `call` that we grab above to be consistent
```suggestion
          call?.localVideoStreams.find((s) => s.mediaStreamType === 'Video') !== undefined
```

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -182,7 +183,10 @@ class CallContext {
         page: newPage,
         endedCall: latestEndedCall,
         devices: clientState.deviceManager,
-        latestErrors: clientState.latestErrors
+        latestErrors: clientState.latestErrors,
+        isCameraOn:
+          clientState.deviceManager.unparentedViews.find((s) => s.mediaStreamType === 'Video') !== undefined ||
```
JamesBurnside, line 188: I was going to comment that nit: `!== undefined` is a double negative, consider an is defined instead
```suggestion
          !!clientState.deviceManager.unparentedViews.find((s) => s.mediaStreamType === 'Video') ||
```

But turns out that check is literally a double negative (`!!`) check in itself 😄

JamesBurnside (approved):
Small change to avoid using `getState` as this will be a stale value to the one passed into setState fn, but otherwise looks great! Really like this change as it was a big gap for me during the hackathon

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -182,7 +183,10 @@ class CallContext {
         page: newPage,
         endedCall: latestEndedCall,
         devices: clientState.deviceManager,
-        latestErrors: clientState.latestErrors
+        latestErrors: clientState.latestErrors,
+        isCameraOn: this.state.call
+          ? !!clientState.calls[this.state.call.id]?.localVideoStreams.find((s) => s.mediaStreamType === 'Video')
```
JamesBurnside, line 188: avoid using `this.state` here for the same reason as `getState()`, this function is updating the state so its reasonable to say that the old state is not reliable
(we probably shouldn't be using this.state.page above but that can be a different problem).
We've already calculated the call at the top of the function (L155), can we not use that?

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -202,6 +202,7 @@ export type CallAdapterClientState = {
     alternateCallerId?: string;
     environmentInfo?: EnvironmentInfo;
     roleHint?: Role;
+    isCameraOn?: boolean;
```
PorterNan, line 205: Just thinking of this for the future a bit - Is it possible that camera has more than one state? like "on", "pending" or we wanna expose more property from camera?

Would be nice if we could have a future proof name like: 
cameraStatus: 'on' | 'off'

Because of this name, we can make it like object or add more string as options to make it flexible

PorterNan: Or even more - we might wanna expose microphoneStatus as well - do we wanna merge them together as 
```
devicesStatus: {
microphone:
camera:
}
```

Maybe cameraStatus is good already for us

PorterNan (approved):
Just a minor comment about naming for future proof, this looks good to me

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -208,6 +208,7 @@ export type CallAdapterClientState = {
     endedCall?: CallState;
     isTeamsCall: boolean;
     latestErrors: AdapterErrors;
+    isCameraOn?: 'On' | 'Off';
```
PorterNan, line 211: Would be good to name this to cameraStatus - otherwise ‘On’ | 'Off' would be still only options for this

PR #2642, opened 2023-01-09
[BugFix] Fix `Idle` state participant string in participant list and videoGallery

dmceachernmsft:
# What

Updates the behavior when a participant is in the `idle` state
# Why

The participant list and video gallery suggest to the user that remote users in the `idle` state are in the 'Calling..' state. Participants will no longer say (Calling...) next to them should they go `Connected` -> `Idle`.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3075499
# How Tested

Wrote E2E tests to reflect the idle and connecting state for participants.


packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -301,16 +302,16 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
...
 
 const participantStateStringTrampoline = (
-  props: ParticipantItemProps,
-  strings: ParticipantItemStrings
+  strings: ParticipantItemStrings,
+  participantState?: ParticipantState
 ): string | undefined => {
   /* @conditional-compile-remove(one-to-n-calling) */
   /* @conditional-compile-remove(PSTN-calls) */
-  return props.participantState === 'Idle' || props.participantState === 'Connecting'
```
JamesBurnside, line 309: Why are we only updating this only for idle and not connecting as well?
A Teams user joining a call has the connecting state:
image
(https://skype.visualstudio.com/SPOOL/_workitems/edit/3075499)

It's confusing that participantStateConnecting string is "Calling...":
image

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -228,7 +229,7 @@ export const ParticipantItem = (props: ParticipantItemProps): JSX.Element => {
     setMenuHidden(true);
   };
 
-  const participantStateString = participantStateStringTrampoline(props, strings);
+  const participantStateString = participantStateStringTrampoline(strings, participantState);
```
JamesBurnside, line 232: issue isn't just with the string, when an ACS Teams interop participant joins they go through the following states:
`idle -> connecting -> lobby -> connected`

This means we will try to render a tile during the idle and connecting states, then remove the tile when they hit the lobby, then return the tile once the hit connected

packages/react-components/src/components/ParticipantItem.tsx
dmceachernmsft: Good point! We should actually take a look at the userId type here. 'Connecting' is only important for PSTN users. will take another look at this. fix will address the below comment as well.

dmceachernmsft: This was updated from the changes in the selectors that now filter out all remote participants in the idle state,

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -403,7 +412,7 @@ const participantStateStringTrampoline = (props: VideoTileProps, locale: Compone
   const strings = { ...locale.strings.videoTile, ...props.strings };
   /* @conditional-compile-remove(one-to-n-calling) */
   /* @conditional-compile-remove(PSTN-calls) */
-  return props.participantState === 'Idle' || props.participantState === 'Connecting'
+  return props.participantState === 'Connecting'
```
JamesBurnside, line 415: with the change to the selector, can we remove this?

JamesBurnside: I assume without the change the video gallery will still say "Calling.." for teams interop participants

dmceachernmsft: Done! just waiting for snapshots now

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -27,7 +27,10 @@ const convertRemoteParticipantsToParticipantListParticipants = (
       remoteParticipants
         // temporarily hiding lobby participants in ACS clients till we can admit users through ACS clients
         .filter((participant: RemoteParticipantState) => {
-          return participant.state !== 'InLobby';
+          return (
```
JamesBurnside, line 30: update the comment here to reflect the new behavior
```suggestion
        // Hide lobby participants in ACS clients till we can admit users through ACS clients.
        // Hide Participants in the idle or connecting stages also, unless they are a phoneNumber in which case connecting reflects on outgoing call
        .filter((participant: RemoteParticipantState) => {
          return (
```

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -26,7 +26,10 @@ export const _videoGalleryRemoteParticipantsMemo = (
   return memoizedAllConvertRemoteParticipant((memoizedFn) => {
     return Object.values(remoteParticipants)
       .filter((participant: RemoteParticipantState) => {
-        return participant.state !== 'InLobby';
+        return (
+          (participant.state !== 'InLobby' && participant.state !== 'Idle' && participant.state !== 'Connecting') ||
```
JamesBurnside, line 30: add same comment here

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -306,7 +305,7 @@ const participantStateStringTrampoline = (
 ): string | undefined => {
   /* @conditional-compile-remove(one-to-n-calling) */
   /* @conditional-compile-remove(PSTN-calls) */
-  return props.participantState === 'Idle' || props.participantState === 'Connecting'
+  return props.participantState === 'Connecting'
```
JamesBurnside, line 308: I still feel we should remove this. Connecting shouldn't mean "Calling..." as not all participants who are connecting are being called.
This is a pure UI component so it shouldn't need to adhere to calling's decision of what a connecting participant means, I think instead in the selector we should be mapping a connecting + phoneNumber participant to be in the ringing state
Let me know your thoughts on this

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -342,7 +342,16 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
             {renderElement}
           </Stack>
         ) : (
-          <Stack className={mergeStyles(videoContainerStyles)} style={{ opacity: participantStateString ? 0.4 : 1 }}>
+          <Stack
+            className={mergeStyles(videoContainerStyles)}
+            style={{
```
JamesBurnside, line 347: why do we have both className and style?

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -140,7 +146,7 @@ test.describe('Participant pane tests', async () => {
     const longPaul = defaultMockRemoteParticipant(
       'I have a really really really really long name. Trust me you dont wanna know.'
     );
-    longPaul.state = 'Connecting';
+    longPaul.state = 'Ringing';
```
JamesBurnside, line 149: Test no longer matches test name and the long participant is not truncated:
image

packages/react-composites/tests/browser/call/hermetic/HorizontalGallery.test.ts
```diff
@@ -78,7 +78,7 @@ test.describe('HorizontalGallery tests', async () => {
   test('HorizontalGallery should have 1 PSTN participant in the horizontal gallery', async ({ page, serverUrl }) => {
     const paul = defaultMockRemoteParticipant('Paul Bridges');
     addVideoStream(paul, true);
-    const vasily = defaultMockRemoteParticipant('Vasily Podkolzin');
+    const vasily = defaultMockRemotePSTNParticipant('+15553334444');
```
JamesBurnside, line 81: We should highlight this issue to design - can't see who you are calling:
image

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
```diff
@@ -185,4 +191,114 @@ test.describe('Participant pane tests', async () => {
...
+    await pageClick(page, dataUiId(IDS.moreButton));
+
+    await waitForSelector(page, dataUiId('call-composite-more-menu-people-button'));
+    await pageClick(page, dataUiId('call-composite-more-menu-people-button'));
+
+    expect(await stableScreenshot(page)).toMatchSnapshot('participant-list-connecting-participant-mobile.png');
+  });
+
+  /* @conditional-compile-remove(one-to-n-calling) @conditional-compile-remove(PSTN-calls) */
+  test('Participant should be hidden when idle to connecting desktop', async ({ page, serverUrl }, testInfo) => {
```
JamesBurnside, line 283: can we avoid having a mobile test and a desktop test?
It's hard to tell but looking between them it seems the only difference is the how the people pane is opened. It would be better to have a function `await openPeoplePane(isMobile)` and just have one test

JamesBurnside (changes requested):
Want to check on whether we want a pure component's 'connecting' state to mean 'Calling...' when there is a ringing state for this purpose

packages/react-composites/tests/browser/call/hermetic/HorizontalGallery.test.ts
dmceachernmsft: there was a decision to do it this way since we cannot fit the name and the string on the tile to just put the string since the user can go to the participant list to see who is being called.

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
dmceachernmsft: this is a old image that should be deleted. this test was changed to be for 1:N calling instead of PSTN

packages/react-components/src/components/ParticipantItem.tsx
dmceachernmsft: I like your suggestion to map a connecting PSTN participant to 'ringing' and removing this. before we were operating on the knowledge that PSTN participants never hit the ringing state in the SDK because of network limitations.

packages/react-composites/tests/browser/call/hermetic/ParticipantPane.test.ts
dmceachernmsft: I don't think we can. problem with the `await waitforSelector` and `await pageClick` calls if the selector it is trying to use is not present the function will timeout the test. moving this to another function I don't think will solve this problem.

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -25,18 +25,28 @@ const convertRemoteParticipantsToParticipantListParticipants = (
...
+            (participant.state !== 'InLobby' && participant.state !== 'Idle' && participant.state !== 'Connecting') ||
+            participant.identifier.kind === 'phoneNumber'
+          );
         })
         .map((participant: RemoteParticipantState) => {
           const isScreenSharing = Object.values(participant.videoStreams).some(
             (videoStream) => videoStream.mediaStreamType === 'ScreenSharing' && videoStream.isAvailable
           );
+          const state =
```
JamesBurnside, line 42: add a quick comment what's happening here

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -25,18 +25,28 @@ const convertRemoteParticipantsToParticipantListParticipants = (
...
+            participant.identifier.kind === 'phoneNumber' && participant.state === 'Connecting'
```
JamesBurnside, line 43: do we need to handle Idle state here as well?

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -24,20 +24,33 @@ export const _videoGalleryRemoteParticipantsMemo = (
...
+         * phone users will be in the connecting state until they are connected to the call.
+         */
+        .filter((participant: RemoteParticipantState) => {
+          return (
+            (participant.state !== 'InLobby' && participant.state !== 'Idle' && participant.state !== 'Connecting') ||
+            participant.identifier.kind === 'phoneNumber'
+          );
+        })
+        .map((participant: RemoteParticipantState) => {
+          const state =
```
JamesBurnside, line 40: given this odd logic is in two places, might be worth a helper function for the whole filter & map

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -306,9 +305,7 @@ const participantStateStringTrampoline = (
 ): string | undefined => {
   /* @conditional-compile-remove(one-to-n-calling) */
   /* @conditional-compile-remove(PSTN-calls) */
-  return props.participantState === 'Idle' || props.participantState === 'Connecting'
-    ? strings?.participantStateConnecting
```
JamesBurnside, line 310: Is `participantStateConnecting` used anywhere else or can we remove from the strings.json files?

JamesBurnside (approved):

packages/calling-component-bindings/src/participantListSelector.ts
dmceachernmsft: no like with teams and group calls `idle` is acting like a placeholder state here. so we want to still show it like before even for PSTN participants.

packages/react-components/src/components/ParticipantItem.tsx
dmceachernmsft: I removed it, lets see if something breaks in CI

carocao-msft (approved):

PR #2644, opened 2023-01-10
Add a userProfileProvider to fetch displayName

PorterNan:
# What

Add a userProfileProvider to fetch displayName

Usage:

For composite - pass a fetchProfile function and it will memorize the first return value of specific params and never update (which means can't live update user name)

For component - wrap components with UserProfileProvider, and provide the function to it

For future design - could be onShowPeopleCard or onHover and trigger some actions to render people card for users in the Profile object
# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/index.ts
+++ b/packages/react-components/src/index.ts
@@ -49,3 +49,7 @@ export type {
   VideoStreamOptions,
   ViewScalingMode
 } from './types';
+
+export { _UserProfileProvider } from './profile/ProfileProvider';
+
+export type { _UserProfileProviderProps, UserProfile } from './profile/ProfileProvider';
--- a/packages/acs-ui-common/review/beta/acs-ui-common.api.md
+++ b/packages/acs-ui-common/review/beta/acs-ui-common.api.md
@@ -58,6 +58,9 @@ export const _logEvent: (logger: AzureLogger, event: TelemetryEvent) => void;
 export const _MAX_EVENT_LISTENERS = 50;
 
 // @public
+export const memoizeAllRet: <ArgsT extends unknown[], FnRetT>(fnToMemoize: (...args: ArgsT) => FnRetT) => (...args: ArgsT) => FnRetT;
+
+// @public
 export const memoizeFnAll: <KeyT, ArgsT extends unknown[], FnRetT, CallBackT extends CallbackType<KeyT, ArgsT, FnRetT>>(fnToMemoize: FunctionWithKey<KeyT, ArgsT, FnRetT>, shouldCacheUpdate?: (args1: unknown, args2: unknown) => boolean) => (callback: CallBackT) => FnRetT[];
 
 // @public
--- a/packages/acs-ui-common/src/index.ts
+++ b/packages/acs-ui-common/src/index.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 export { memoizeFnAll } from './memoizeFnAll';
+export { memoizeAllRet } from './memoizeAllRet';
 export {
   fromFlatCommunicationIdentifier,
   toFlatCommunicationIdentifier,
--- a/packages/acs-ui-common/src/memoizeAllRet.ts
+++ b/packages/acs-ui-common/src/memoizeAllRet.ts
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { _safeJSONStringify } from './safeStringify';
+
+/**
+ * The function memoize is to cache all results return by the same function args
+ * It will compare function args to determine whether to return previous value or return a new one
+ * @param  fnToMemoize - the function needs to be memorized
+ * @public
+ */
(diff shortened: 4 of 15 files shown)


packages/react-components/src/index.ts
```diff
@@ -49,3 +49,7 @@ export type {
   VideoStreamOptions,
   ViewScalingMode
 } from './types';
+
+export { _UserProfileProvider } from './profile/ProfileProvider';
```
JamesBurnside, line 53: My feeling is we shouldn't be adding private providers at the UI components layer (and by extension we shouldn't have UI components using the hooks for those providers). UI Components package is meant to be as if it was an external library to the composites, but here the composites are sneakily using a repo-only provider, meaning we further the distance that contoso would have to implement if they were to copy the composites' code.

A provider can exist but I think only at the composite layer. There we can inject the displayName based on the provider value as this is how Contoso would use the UI component.

JamesBurnside: An alternative here could be to have the CallClientProvider take in a displayNameHint, then the usePropsFor can be augmented to use the displayNameHint if we don't have an SDK-driven display name

PorterNan: I think the final target is to expose the provider as a component level api, but we can take it step by step, grab some feedback from composite level then make decision on this

JamesBurnside: I'm confused between providers vs selectors in how customers are meant to use the library.
Does this mean if someone wanted to build their own stateful-client driven CTE experience they would then need:
```
<CallClientProvider>
   <CallProvider>
      <LocalizationProvider>
        <DisplayNameProvider>
           const props = usePropsFor(component);
           <Component {...props} />
       />
    />
  />
/>
```
In this case why isn't CallClientProvider taking in whatever the DisplayNameProvider is doing? And with all these providers why are some props retrieved from the usePropsFor and others are injected by the providers?

PorterNan: Was thinking of this for a while, I almost forget that I got a pretty similar design at the very beginning, I put a displayName hint as a parameter of statefulClient

But later, I noticed getDisplayName() needs to be an async function which will cause some side effects (and you never know when to call it for different participants), so it is not the same pattern we could apply to useSelector (unless we decide to hijack listParticipant() and add data there)

PorterNan: One thing I can think of is the pattern we used for onRenderAvatar in component then onFetchAvatarData in composite

we could add an onRenderDisplayName(userId: string, displayName: string) => JSX.Element to those components who need displayName override with 2 parameter, 1. userId as the user id of that user, 2, displayName for the value coming from sdk (but contoso can choose not to render it and render their own one)

So we do some interesting wrapping and fetching in composite, convert onFetchAvatarData to onRenderDisplayName, then render it inside of component

JamesBurnside: > getDisplayName() needs to be an async function

Can you elaborate on this? As a pretext, we strongly want to avoid any callback being async: https://gist.github.com/prprabhu-ms/9a35b69f0c1a667d8e4de79c544da007

PorterNan: That is a good article, and I fully agree with it, that's why I introduced a memorize-and-forget feature for displayName to make my current case is falling into the preset condition like this:
image

While most part of the article looks great to me, there is a missing part there - sync function cannot be onFetchName, which return a string, otherwise there is no way to notify the update of displayName when re-render is needed (the problem might be covered up because we have so many updates in a call)

If we choose go with sync function, there would be 2 choices - expose onRenderDisplayName directly like what I mentioned for component (also for composite) or add a sync function onFetchName + version props to force update (which might be hard to understand)

Overall, I strongly agree that async function shouldn't appear in component, because it is pure, sync onRender should be there to serve this purposes, for composite, I would still consider async as a low code solution - just need to indicate the function would be only called once and the result would be memorize forever (Or we might want to expose onRender directly for composite)

PorterNan: Lemme summarize what we currently have for potential solution, just for note taking and preparing tmr's session:

1. Sync function - onFetchDisplay + version exposed in composite props - onFetchName to fetch json data, then version to notify the change (usually the finish of outside async call)
2. Sync function - onFetchDisplay + a value to notify change for Provider - we need something to ask Provider to re-render
3. Sync function - onRenderDisplayName, same to component level, but also expose this in composite level, this will allow contoso to control the lifecycle of displayName rendering so they will have an async call in their displayName component (* I prefer this for sync one)
4. Async function - do a deep comparation from the returning object and update only when necessary, in composite, then still use onRenderDisplayName for component
5. Async function - memoize and forget - we only do it for once every id and value never changes,  still use onRenderDisplayName for component (I prefer this one)

Personally I prefer 3 or 5 depending on how much we want to do in composite, 1&2 or 4&5, their api  would be compatible to each other which wouldn't be a breaking change

I've already scheduled a brainstorm session for this, let's walk these through to see what we would like to go ahead with

PorterNan:
Closed, we decided to add the fetch function as parameter in adapter

PR #2648, opened 2023-01-11
Bugfix: Fix camera permission request flash

JamesBurnside:
# What
Ensure that setState is not called until both camera and microphone values are known

# Why
We were setting the permission for one, a render pass would happen, then the permission for the other would be set.
https://skype.visualstudio.com/SPOOL/_workitems/edit/3086747


dmceachernmsft (approved):

edwardlee-msft (approved):

edwardlee-msft:
This is a neat fix!

PR #2650, opened 2023-01-13
Fix rerender of VideoGallery when first participant is pinned

mgamis-msft:
# What
Problem:
The rerender occurs because the VideoGallery is switching between PinnedParticipantLayout and FloatingLocalVideoLayout causing all the remote video tiles to be rerendered.
Solution:
Do pinned participant logic in FloatingLocalVideoLayout and DefaultLayout. Remove PinnedParticipantLayout 

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3104069

# How Tested

Local sample testing
https://user-images.githubusercontent.com/79475487/212332894-85fa0706-54d8-40ef-b863-0fd0b6a80945.mp4

# Process & policy checklist


packages/react-components/src/components/VideoGallery/FloatingLocalVideoLayout.tsx
```diff
@@ -51,23 +51,25 @@ export const FloatingLocalVideoLayout = (props: FloatingLocalVideoLayoutProps):
...
+    parentHeight,
+    /* @conditional-compile-remove(pinned-participants) */ pinnedParticipantUserIds
   } = props;
 
   const theme = useTheme();
 
   const isNarrow = parentWidth ? isNarrowWidth(parentWidth) : false;
 
-  const floatingLocalVideoLayout = useFloatingLocalVideoLayout({
+  const layout = useLayout({
```
JamesBurnside, line 62: nit: `useLayout` is a vague (hard to know what a `layout` entails, I had to go look at the function definition to understand what this hook did), consider something like:
```suggestion
  const { gridParticipants, horizontalGalleryParticipants } = useOrganizedParticipants({
```

(struggling also to come up with a better word, not in love with `useOrganizedParticipants` but I think just having the return value immediately split goes a long way in understanding what this function does)

JamesBurnside (approved):
Love this refactor!

prabhjot-msft (approved):

PR #2651, opened 2023-01-13
Remote video tile keyboard accessible to open menu

mgamis-msft:
# What

Make RemoteVideoTile menu keyboard accessible to open menu

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3091719

# How Tested

Local sample testing
Desktop: https://user-images.githubusercontent.com/79475487/212334291-aa21dea2-2f03-4371-bf71-0faac8b367eb.mp4
Emulated mobile on browser: https://user-images.githubusercontent.com/79475487/212334405-209467be-5593-42d8-b9bd-e368b3e1ea22.mp4

# Process & policy checklist


packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -138,8 +140,26 @@ export const _RemoteVideoTile = React.memo(
...
+        }
+      },
+      [setDrawerMenuItemProps, contextualMenuProps]
+    );
+
     return (
-      <Stack style={remoteVideoTileWrapperStyle}>
+      <Stack
+        /* @conditional-compile-remove(pinned-participants) */
+        tabIndex={remoteVideoTileMenuKind === 'drawer' ? 0 : undefined}
```
dmceachernmsft, line 158: why are we setting the tab order of the video tile to `0` when it will show a drawer? it shouldn't be any different for mobile incase the user uses a bluetooth keyboard.

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -138,8 +140,26 @@ export const _RemoteVideoTile = React.memo(
...
+        /* @conditional-compile-remove(pinned-participants) */
+        onKeyDown={remoteVideoTileMenuKind === 'drawer' ? onKeyDown : undefined}
```
dmceachernmsft, line 160: should this handler only be used when its a drawer over a context menu? the handler you wrote will work for desktop no?

packages/react-components/src/components/RemoteVideoTile.tsx
JamesBurnside: My assumption here is that when menuKind is contextmenu it has the three dots that can be focused and clicked instead?

JamesBurnside (approved):

packages/react-components/src/components/RemoteVideoTile.tsx
mgamis-msft: Yes James is right. When the menuKind is 'contextual' then ellipses button will be present and keyboard accessible so we don't need to focus on the remote video tile. I don't think we want the keyboard user to tab on a remote video tile and then have to tab again to focus on the ellipses button to open the contextual menu:

https://user-images.githubusercontent.com/79475487/212398077-c67cd0ac-dc7b-4428-aef8-3271c8204cf6.mp4

But let me know if this is desired.

packages/react-components/src/components/RemoteVideoTile.tsx
mgamis-msft: Yes, the handler should only be assigned when menuKind = 'drawer'. The RemoteVideoTile will be assigned menuKind = 'drawer' when the VGL is on mobile and menuKind = 'contextual' when the VGL is on desktop. We don't want the handler to be active on Desktop like so:
https://user-images.githubusercontent.com/79475487/212398985-b2a781c6-be38-4525-ba1a-ff46c9faae11.mp4

dmceachernmsft (approved):

PR #2654, opened 2023-01-17
Make common adapter generic according to ARB

PorterNan:
# What

# Why

# How Tested

# Process & policy checklist


JamesBurnside (approved):
Don't recall what we decided on here but both way look good to me so approving 👍

dmceachernmsft (approved):

PorterNan:
Closed, because we no longer need to make this change

PR #2655, opened 2023-01-17
 Change `type` to `kind` for SitePermissions Components

JamesBurnside:
# What
*  Change `type` to `kind` for SitePermissions Components
*  Fix some conditional compiles flags

# Why
Old APIView comment


packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -77,7 +77,7 @@ import { StatefulChatClient } from '@internal/chat-stateful-client';
 import { ChatThreadClient } from '@azure/communication-chat';
 import { useEffect, useRef, useState } from 'react';
 import { _toCommunicationIdentifier } from '@internal/acs-ui-common';
-/* @conditional-compile-remove(unsupported-browser) */
+/* @conditional-compile-remove(rooms) */
```
prabhjot-msft, line 80: is the conditional compile statement for rooms didn't exist already for the import and if no, why do we need it now?

JamesBurnside: after some previous refactoring only rooms uses this import, it should have had both `unsupported-browser` and `rooms` so this change should just remove the `unsupported-browser` but there was a gap where rooms wasn't specified (because we never test our beta compilation compiling only a subset of all features enabled)

prabhjot-msft (approved):

edwardlee-msft (approved):

PR #2658, opened 2023-01-18
Update documentation to specify NodeJS 14.x

emlynmac:
# What
References throughout the documentation reference node version greater than 14.9.
Rush requires version 14.x, which 

# Why
Current default LTS is 16.x, which is not compatible with the project

# How Tested
Documentation only; tested with storybook locally

# Process & policy checklist

No


change-beta/@azure-communication-react-ad25f56f-cdbf-483e-96ec-846c3f466f48.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Update nodeJS wording to require version 14",
```
JamesBurnside, line 3: Instead of requiring node 14, I think we should ensure we are working with LTS always.
Instead can we update the `rush.json` to something like
```json
"nodeSupportedVersionRange": ">=12.13.0 <13.0.0 || >=14.15.0 <18.0.0",
```

@dmceachernmsft do we still have outstanding issues with node 18 or were those fixed?

packages/storybook/stories/CallComposite/Quickstart1ToNDocs.stories.mdx
```diff
@@ -32,7 +32,7 @@ Download the [UI Library quickstart composites](https://github.com/Azure-Samples
 ## Prerequisites
 
 - An Azure account with an active subscription. [Create an account for free](https://azure.microsoft.com/free/?WT.mc_id=A261C142F).
-- [Node.js](https://nodejs.org/) Active LTS and Maintenance LTS versions (Node 14.19.0 and above).
+- [Node.js](https://nodejs.org/) Active LTS and Maintenance LTS versions (Node 14 latest).
```
JamesBurnside, line 35: These docs refer to using the npm package, as opposed to using the repo directly which has rush, so _should_ work with any version 14+ -- if not we should definitely fix that

JamesBurnside (changes requested):

change-beta/@azure-communication-react-ad25f56f-cdbf-483e-96ec-846c3f466f48.json
emlynmac: The rush tool _only_ supports version 14 in the current setup. 
`"nodeSupportedVersionRange": ">=12.13.0 <13.0.0 || >=14.15.0 <15.0.0",`
This limits the rest of the system to 14 also, does it not?

packages/storybook/stories/CallComposite/Quickstart1ToNDocs.stories.mdx
emlynmac: Maybe I was too enthusiastic with find and replace...

The issue I have is that rush only works with 14 currently, which is a little behind the current LTS

change-beta/@azure-communication-react-ad25f56f-cdbf-483e-96ec-846c3f466f48.json
JamesBurnside: Oh yep great point it should filter out odd numbers - but I think it should JustWork if we update that version range. We should also drop anything below 14 as thats deprecated.
Donald was doing some testing with node 18 so keen to know if we can flag that as supported also

emlynmac: Maybe this is something to discuss internally before putting in a PR. I would have thought moving Node versions could have bigger impact - and more testing first.

emlynmac:
Closing pending further discussion / testing

PR #2662, opened 2023-01-19
Remote video tile menu aria labels and announcements

mgamis-msft:
# What

Added Announcer to VideoGallery and passed a callback to toggle string assigned to the announcer to RemoteVideoTile so that when pin/unpin menu items are clicked 'Pinned {participantName}'/'Unpinned {participantName}' will be announced.
Added aria labels to the menu items.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3107424

# How Tested

On Desktop with narrator. For some reason the aria labels are not working on my IPad. If you have any clue why please let me know.

# Process & policy checklist


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -382,6 +396,25 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     [pinnedParticipantsState, setPinnedParticipantsState, onUnpinParticipantHandler]
   );
 
+  /* @conditional-compile-remove(pinned-participants) */
```
dmceachernmsft, line 399: nit: To cut down on CC I don't think that we need to CC this. All of the announcer stuff is in the stable API so we should be able to leave this here. Though it will cause build errors since the toggle function wont be used. this can be fixed by moving the CC a layer into  the internal _RemoteVideoTile. wont block on this though if you don't want to change this.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -423,6 +456,8 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
           isPinned={isPinned}
           /* @conditional-compile-remove(pinned-participants) */
           disablePinMenuItem={pinnedParticipants.length >= MAX_PINNED_REMOTE_VIDEO_TILES}
+          /* @conditional-compile-remove(pinned-participants) */
```
dmceachernmsft, line 459: Referred to in comment above ^

dmceachernmsft (commented):
Should the print label be changing here? does the design support this update?
image

mgamis-msft:
> Should the print label be changing here? image

Yes I have confirmed with the FIgma
image

dmceachernmsft (approved):

edwardlee-msft (approved):

PR #2671, opened 2023-01-23
[BugFix] Update error bar selector for when camera is already in use

dmceachernmsft:
# What

Update error bar messaging in Call and config screens for when camera is in use already
# Why

image

image

https://skype.visualstudio.com/SPOOL/_workitems/edit/3030591
# How Tested

Ran locally with camera app stealing camera.

Also checked in with Calling to validate the different cases when the error for video failed is being surfaced.


packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -126,7 +126,16 @@ export const errorBarSelector: ErrorBarSelector = createSelector(
...
-    appendActiveErrorIfDefined(activeErrorMessages, latestErrors, 'Call.startVideo', 'startVideoGeneric');
+    if (
+      latestErrors['Call.startVideo']?.message === 'Call.startVideo: Video operation failure SourceUnavailableError'
+    ) {
+      appendActiveErrorIfDefined(activeErrorMessages, latestErrors, 'Call.startVideo', 'callCameraAlreadyInUse');
+    } else if (
+      latestErrors['Call.startVideo']?.message === 'Call.startVideo: Video operation failure permissionDeniedError'
+    ) {
+      appendActiveErrorIfDefined(activeErrorMessages, latestErrors, 'Call.startVideo', 'callCameraAccessDenied');
+    }
```
JamesBurnside, line 137: should there be an else `appendActiveErrorIfDefined(activeErrorMessages, latestErrors, 'Call.startVideo', 'startVideoGeneric');` if the messages don't exactly match those errors?

JamesBurnside (approved):

carocao-msft (approved):

PR #2680, opened 2023-01-25
Add onFetchProfile to teams adapter option

PorterNan:
# What

Add onFetchProfile to teams adapter option

- Add a util function to memorize and modify participant ref and generate when something changes
- Add a sync modifier which will modify the state and able to handle side effects
Code sample:
image
Result:
image

# Why

# How Tested

# Process & policy checklist


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2786,6 +2789,11 @@ export type ParticipantsRemovedListener = (event: {
 // @public
 export type ParticipantState = 'Idle' | 'Connecting' | 'Ringing' | 'Connected' | 'Hold' | 'InLobby' | 'EarlyMedia' | 'Disconnected';
 
+// @public
+export type Profile = {
```
dmceachernmsft, line 2793: will this hold more in the future like email and such for a contact card?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -830,6 +843,20 @@ export type AzureCommunicationCallAdapterArgs = {
...
 };
 
+/**
+ * Optional parameters to create {@link AzureCommunicationCallAdapter}
+ *
+ * @beta
+ */
+export type TeamsAdapterOptions = {
+  /**
+   * Use this to hint the role of the user when the role is not available before a Rooms call is started. This value
```
dmceachernmsft, line 853: should this comment be updated?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -1004,6 +1045,7 @@ const useAzureCommunicationCallAdapterGeneric = <
       /* @conditional-compile-remove(PSTN-calls) */
       alternateCallerId,
       /* @conditional-compile-remove(rooms) */
+      /* @conditional-compile-remove(teams-identity-support) */
```
dmceachernmsft, line 1048: should this be dependent on both CC statements here?

packages/communication-react/review/beta/communication-react.api.md
PorterNan: Yep, that is why we named this as Profile, and once major update is allowed, we should add AvatarPersonaData here as well

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: Nice catch, I was adding this for both adapter, but now it is only for teams one

PorterNan: Hmm, interesting question, I looked at the code a bit, this dependency is like "this or that", so either code being available in stable then we should not remove this line anymore - current conditional compilation logic works for this part

Some other dependency is like "This and that", so we need both code to enable the line.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2998,6 +3006,11 @@ export interface SystemMessageCommon extends MessageCommon {
     messageType: 'system';
 }
 
+// @beta
+export type TeamsAdapterOptions = {
+    onFetchProfile?: OnFetchProfileCallback;
```
JamesBurnside, line 3011: should this be a required param? Otherwise we don't display the current users info

PorterNan: This is a good point - I got the same problem and discussed that a bit, according to previous discussions:
1. CTE can work with Teams client (especially 1 to 1, or 1 to n) without this one
2. Displayname could be potentially coming from sdk in the future
So the discussion result is not to make this required, to make it easier to use for one CTE to multiple teams client scenarios and leave the possibility for displayname coming into sdk

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -108,7 +108,19 @@ const TeamsCallScreen = (props: TeamsCallScreenProps): JSX.Element => {
     throw new Error('A MicrosoftTeamsUserIdentifier must be provided for Teams Identity Call.');
   }
 
-  const adapter = useTeamsCallAdapter({ ...adapterArgs, userId, locator }, afterCreate);
+  const adapter = useTeamsCallAdapter(
+    {
+      ...adapterArgs,
+      userId,
+      locator,
+      options: {
```
JamesBurnside, line 116: does this need memoized?

PorterNan: That's a good catch, and this shouldn't be checked into the PR, this is just for test code

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -830,6 +843,20 @@ export type AzureCommunicationCallAdapterArgs = {
...
+ *
+ * @beta
+ */
+export type TeamsAdapterOptions = {
+  /**
+   * Use this to fetch profile information which will override data in {@link CallAdapterState} like display name
+   * The onFetchProfile is fetch-and-forget one time action for each user, once a user profile is updated, the value will be cached
+   * and would not be updated again within the lifecycle of adapter.
+   */
+  onFetchProfile?: OnFetchProfileCallback;
```
JamesBurnside, line 857: (not for this PR, just a thought)
Should we provide any way to know if the information was populated from the calling sdk already?
E.g. I want to come along and only modify profiles were data doesn't already exist.

An alternative to this is in an example usage we pull out of the call remoteparticipants object by default and only do some fancy graph fetch if its undefined and a teams user

samples/Calling/src/app/views/CallScreen.tsx
PorterNan: Reverted

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
PorterNan: That's a good point, we can do (id: string, displayName?: string) => Profile for this case

Let's check this in for beta release first, then I will open a new PR to add this one

dmceachernmsft (approved):

JamesBurnside (approved):
🚀🚀🚀

PR #2681, opened 2023-01-26
[Bugfix][a11y] Fixed leave chat state is not announced by Screen Reader

Leah-Xia-Microsoft:
# What

The status message should be announced by Screen Reader after leaving chat. Screen Reader should announce as "You left the Chat" as seen visually.

Screenshot 2023-01-25 at 17 10 34

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2881374

# How Tested

Tested on Mac with Chrome and Safari

# Process & policy checklist


samples/Chat/src/app/EndScreen.tsx
```diff
@@ -70,6 +71,7 @@ export const EndScreen = (props: EndCallProps): JSX.Element => {
       className={endChatContainerStyle}
     >
       <Stack tokens={upperStackTokens}>
+        <Announcer announcementString={leftCall} ariaLive={'polite'} />
```
dmceachernmsft, line 74: This is good! Another more recommended way to do this would be to verify that the text in the heading is read instead of using an `announcer` just so we don't have a army of them running around in the DOM. you can set the aria-label to be what you gave the `announcer`. sometimes Fluent components (I'm assuming `Text` is from fluent) can have issues if their aria label is not set. their syntax for setting them is a prop like: `ariaLabel={'your string here'}` .

Leah-Xia-Microsoft: > This is good! Another more recommended way to do this would be to verify that the text in the heading is read instead of using an `announcer` just so we don't have a army of them running around in the DOM. you can set the aria-label to be what you gave the `announcer`. sometimes Fluent components (I'm assuming `Text` is from fluent) can have issues if their aria label is not set. their syntax for setting them is a prop like: `ariaLabel={'your string here'}` .

Thanks Donald. Moved it to Text itself.

dmceachernmsft (approved):

emlynmac (approved):

PR #2682, opened 2023-01-26
Console Log for invalid userId for pinned participants

prabhjot-msft:
# What
Console Log for invalid userId for pinned participants

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -300,6 +300,12 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
   /* @conditional-compile-remove(pinned-participants) */
   const [pinnedParticipantsState, setPinnedParticipantsState] = React.useState<string[]>([]);
   /* @conditional-compile-remove(pinned-participants) */
+  props.pinnedParticipants?.forEach((pinParticipant) => {
```
dmceachernmsft, line 303: is there a way that we can tee this to the adapter error state in the composite and show an error bar for it to the user? just so we can let them know something went wrong when they tried to pin someone and it failed?

prabhjot-msft: This error would not occur in case of composites. Contoso can only control pinned participants if they are using video gallery directly. If they use composite they won't be able to pass pinned participants.

dmceachernmsft: oh I see, I read into where this would go. I am still concerned that we are not tracking the error state here correctly. Should this maybe also throw so that Contoso can catch the error themselves?

prabhjot-msft: Screenshot 2023-01-30 at 12 06 36 PM
Changed the console error to throw statement to let Contoso be informed about the issue.

JamesBurnside: This is currently being executed on every render pass of the VideoGallery - this cause the render to exit early if this is detected.
Can we instead throw this error when pinParticipant is called?

dmceachernmsft (approved):

packages/react-components/src/components/VideoGallery.tsx
JamesBurnside: I'm concerned about throwing an error on the direct render pass as this will take down that video gallery's render update - and if Contoso isn't explicitly listening for uncaught errors with a react error boundary (which is non-trivial) they take down their whole app.
Its also quite easy for contoso to provide incorrect participants here, imagine I have 4 pinned participants and one leaves, I should be updating this array but as a developer I could easily make a mistake -- and that mistake will take down the whole video gallery.
I wonder if throwing the error in an async block would work...

JamesBurnside: My preference here would be to console warn in this scenario

JamesBurnside, line 303: This should be in a useEffect block to it only executes when the pinnedParticipants or the remoteParticipants changes

JamesBurnside (changes requested):
concerned about the throw error implications

packages/react-components/src/components/VideoGallery.tsx
prabhjot-msft: done

dmceachernmsft: yeah I didn't think about that. we don't throw in any of our components currently because of this reason right? that in mind I think a warning would suffice. should we update the comment for this prop they can expect that warning if something went wrong just so they know to look for it and they can do something about it?

mgamis-msft:
@prabhjot-msft I agree with James. Let's just do a console.warn in a useEffect when a pinnedParticipant user id is not in the remoteParticipants

packages/react-components/src/components/VideoGallery.tsx
prabhjot-msft: Yeah, I agree. I will change the throw statement to the warning and will update troubleshooting guide as well

JamesBurnside (approved):

PR #2689, opened 2023-01-30
Add `@azure/communication-calling-effects` package

JamesBurnside:
# What
* Add `@azure/communication-calling-effects` package to beta releases only
  * Added as a **peer dep**
* Update `force-build-versions` script to _remove_ beta-only packages

# Why
Needed for background effects feature

# How Tested
Verified when running `force-build-versions` script does indeed remove the calling-effects package:
image

### DIFF
--- a/common/scripts/package-utils.js
+++ b/common/scripts/package-utils.js
@@ -25,7 +25,7 @@ function parsePackage(packagePath) {
 function overrideAllPackages(packagePaths, packageProcessFunc) {
   for (const path of packagePaths) {
     const packageJson = parsePackage(path);
-    newPackageJson = packageProcessFunc(packageJson);
+    const newPackageJson = packageProcessFunc(packageJson);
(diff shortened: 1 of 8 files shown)


common/scripts/package-utils.js
```diff
@@ -66,6 +66,33 @@ const updateAllDepVersions = (versionUpdater, deps/* dependency names to update
...
+const removeDepsFromAllPackages = (deps) => {
+  const packagePaths = findAllPackageJSON(PACKAGES_DIR);
+
+  const removeDependencyByName = (dependencies, depName) => {
+    if (typeof dependencies === 'object') {
+      delete dependencies[depName];
+    }
+  }
+
+  const packageProcessFunc = (packageJson) => {
```
PorterNan, line 84: Nice to see we reuse all our codes using this pattern!

PorterNan (approved):

prabhjot-msft (approved):

PR #2690, opened 2023-01-30
Run `force-build-versions` in CI at all times when switching to stable

JamesBurnside:
# What
* Run `force-build-versions` in CI at all times when switching to stable
* Run `rush update:stable` to ensure the new version definitions are picked up

# Why
`force-build-versions` ensures the crafted package.json only has the required versions.
This is only run on creating a release branch, we should be running it on all CI.

Related, this ensures the remove of #2689 from stable builds will be tested at CI time.

# How Tested
CI runs
Also ran these commands locally

```sh
rush install
node ./common/scripts/force-build-flavor.mjs stable
rush switch-flavor:stable
```

And confirmed `@azure/communication-calling-effects` was removed in the node_modules:
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,8 +66,9 @@ jobs:
         id: switch-flavor
         if: ${{ matrix.flavor == 'stable' }}
         run: |
-          node ./common/scripts/force-build-flavor.mjs stable
           rush switch-flavor:stable
+          node ./common/scripts/force-build-flavor.mjs stable
+          rush update:stable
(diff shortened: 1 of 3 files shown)


PorterNan (approved):
Nice to see we can do this relatively easy - maybe we can add that to .github\workflows\update-snapshots.yml as well?

(Or maybe add a single command like switch-flavor-ci:stable, then use it everywhere in ci?)

dmceachernmsft (approved):

edwardlee-msft (approved):

PR #2691, opened 2023-01-31
[BugFix] update errorbar selector to no longer incorrectly indicate user is on mac

dmceachernmsft:
# What

image

image
Error bar will validate the user is on mac and show correct message. stable flavor just show generic message
# Why

Error bar no longer tells android or windows user they are on macOS and their camera is blocked

https://skype.visualstudio.com/SPOOL/_workitems/edit/3124414
# How Tested

Validated locally issue does not happen anymore on windows and android. tested on mac using browserstack errors still showed mac strings

### DIFF
--- a/packages/calling-component-bindings/src/errorBarSelector.ts
+++ b/packages/calling-component-bindings/src/errorBarSelector.ts
@@ -125,13 +125,13 @@ export const errorBarSelector: ErrorBarSelector = createSelector(
     ) {
       activeErrorMessages.push({ type: 'callMacOsCameraAccessDenied' });
     }
-    /* @conditional-compile-remove(unsupported-browser) */
+
     /**
      * This UFD only works on mac still so we should only see it fire on mac.
      */
     if (
       diagnostics?.media.latest.screenshareRecordingDisabled?.value === true &&
-      environmentInfo?.environment.platform === 'mac'
+      /* @conditional-compile-remove(unsupported-browser) */ environmentInfo?.environment.platform === 'mac'
     ) {
       activeErrorMessages.push({ type: 'callMacOsScreenShareAccessDenied' });
     }
--- a/packages/calling-component-bindings/src/errorBarSelector.ts
+++ b/packages/calling-component-bindings/src/errorBarSelector.ts
@@ -58,7 +58,6 @@ export const errorBarSelector: ErrorBarSelector = createSelector(
       return environmentInfo?.environment.browser === 'safari';
       return /^((?!chrome|android|crios|fxios).)*safari/i.test(navigator.userAgent);
     };
-    console.log(environmentInfo);
 
     // Errors reported via diagnostics are more reliable than from API method failures, so process those first.
     if (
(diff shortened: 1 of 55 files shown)


packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -58,6 +58,7 @@ export const errorBarSelector: ErrorBarSelector = createSelector(
       return environmentInfo?.environment.browser === 'safari';
       return /^((?!chrome|android|crios|fxios).)*safari/i.test(navigator.userAgent);
     };
+    console.log(environmentInfo);
```
carocao-msft, line 61: Remove console log here

packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -114,10 +115,24 @@ export const errorBarSelector: ErrorBarSelector = createSelector(
...
     }
-    if (diagnostics?.media.latest.screenshareRecordingDisabled?.value === true) {
+
+    /**
+     * This UFD only works on mac still so we should only see it fire on mac.
+     */
+    if (
+      diagnostics?.media.latest.screenshareRecordingDisabled?.value === true &&
+      /* @conditional-compile-remove(unsupported-browser) */ environmentInfo?.environment.platform === 'mac'
+    ) {
```
carocao-msft, line 135: Is it ok that this fix will only be stable once the unsupportedBrowser feature moves to stable?

dmceachernmsft: yup, since currently we cannot block screen sharing on windows for now that this will never fire. however that might not always be the case so we should still proceed to add this fix since the string is specific to mac.

carocao-msft (approved):

packages/calling-component-bindings/src/errorBarSelector.ts
JamesBurnside: This fires when you start a screenshare but then cancel on the "choose what to share" dialog on windows -- but it exists in stable releases currently so don't consider this a blocking comment for this PR

JamesBurnside: More info: https://teams.microsoft.com/l/message/19:5e150d250f724d1089c949305efdc8cc@thread.skype/1674868535410?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e9c1fc3-39df-4486-a26a-456d80e80f82&parentMessageId=1674687915063&teamName=Azure%20Communication%20Services&channelName=Web%20UI%20-%20Team&createdTime=1674868535410

packages/calling-component-bindings/src/errorBarSelector.ts
```diff
@@ -114,10 +114,24 @@ export const errorBarSelector: ErrorBarSelector = createSelector(
       }
     }
```
JamesBurnside, line 116: Can't comment on the line, too far up, but L86 I think should also have the macOS check as we are providing a mac speficifc string:
```
    if (diagnostics?.media.latest.microphonePermissionDenied?.value === true) {
      activeErrorMessages.push({ type: 'callMacOsMicrophoneAccessDenied' });
    }
```

JamesBurnside (approved):

PR #2693, opened 2023-02-01
Fix for navigation using keyboard to activate the video tile context men

prabhjot-msft:
# What
Added empty icon render for tab navigation of video tile

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3123445

https://user-images.githubusercontent.com/97124699/215964340-e2aa32d9-0c4a-4185-96aa-c4738898f516.mov

# How Tested

# Process & policy checklist


packages/react-components/src/components/VideoTile.tsx
```diff
@@ -189,18 +187,24 @@ const videoTileMoreMenuProps = {
...
+const VideoTileMoreOptionsButton = (props: {
+  contextualMenu?: IContextualMenuProps;
+  canShowContextMenuButton: boolean;
+}): JSX.Element => {
+  const { contextualMenu, canShowContextMenuButton } = props;
   if (!contextualMenu) {
     return <></>;
   }
+
+  const optionsIcon = canShowContextMenuButton ? 'VideoTileMoreOptions' : 'VideoTileHideOptionsIcon';
```
JamesBurnside, line 199: What is the purpose behind the `VideoTileHideOptionsIcon` ? Are we unable to not render the `<IconButton` when `canShowContextMenuButton` is false?

dmceachernmsft (approved):

packages/react-components/src/components/VideoTile.tsx
prabhjot-msft: It renders the icon button without any icon, so the button is there and it can be navigated using keyboard to activate the context menu. We use similar solution for participant Item where the ellipsis appear on hover.

JamesBurnside (approved):

mgamis-msft (approved):

PR #2697, opened 2023-02-01
Add parameter to hint displayName from SDK

PorterNan:
# What

Add parameter to hint displayName from SDK, so Contoso doesn't have to override it if there is any from sdk
# Why

# How Tested

# Process & policy checklist


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2608,7 +2608,7 @@ export interface NetworkDiagnosticsState {
 }
 
 // @beta
-export type OnFetchProfileCallback = (userId: string) => Promise<Profile | undefined>;
+export type OnFetchProfileCallback = (userId: string, displayName?: string) => Promise<Profile | undefined>;
```
JamesBurnside, line 2611: to make this extendable, should this be profileData instead of displayName?
```suggestion
export type OnFetchProfileCallback = (userId: string, profileData?: Profile) => Promise<Profile | undefined>;
```

JamesBurnside (approved):
This a great solution to make Contoso know what to provide! Love it!

edwardlee-msft (approved):

PR #2704, opened 2023-02-03
[BugFix] Resize local video tile based on local feed size

dmceachernmsft:
# What

Styles for local video tile are based on local stream width
# Why

Allows the local stream to always be fit scaling mode so local user can see the whole of their stream

https://skype.visualstudio.com/SPOOL/_workitems/edit/3042907
# How Tested

Ran locally to see resizing happen


dmceachernmsft:
Closing, improved mobile experience with 9:16 aspect ratio. Maybe something to revisit

PR #2705, opened 2023-02-03
Added stream type to stream logs 

carocao-msft:
# What
Added stream type to stream logs 

# Why
Improve our logs so we can easily see which stream fails

### DIFF
--- a/packages/calling-stateful-client/src/StreamUtils.ts
+++ b/packages/calling-stateful-client/src/StreamUtils.ts
@@ -39,6 +39,7 @@ async function createViewRemoteVideo(
 ): Promise<CreateViewResult | undefined> {
   // Render RemoteVideoStream that is part of a Call
   const streamId = stream.id;
+  const streamType = stream.mediaStreamType;
   let participantKey;
   if (typeof participantId === 'string') {
     participantKey = participantId;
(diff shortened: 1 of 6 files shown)


packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -51,7 +52,8 @@ async function createViewRemoteVideo(
     name: EventNames.CREATING_REMOTE_VIEW,
     level: 'info',
     message: 'Start creating view for remote video.',
-    data: streamLogInfo
+    data: streamLogInfo,
+    streamType: streamType
```
JamesBurnside, line 56: Given this is extra metadata, move this into the streamLogInfo object (i.e. part of the `data` field of the logEvent)

JamesBurnside (changes requested):

JamesBurnside (approved):

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -498,7 +506,11 @@ function disposeViewRemoteVideo(
 }
 
 function disposeViewLocalVideo(context: CallContext, internalContext: InternalCallContext, callId: string): void {
-  const streamLogInfo = { callId };
+  context.setLocalVideoStreamRendererView(callId, undefined);
```
JamesBurnside, line 509: Just a heads up, #2713 moves this line in case this PR is checked in afterwards

edwardlee-msft (approved):

PR #2707, opened 2023-02-03
[Bugfix] Fix sending/delivered/failed message status is hidden for large group

Leah-Xia-Microsoft:
# What

Fix sending/delivered/failed message status is hidden for large group

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2764493

# How Tested

- In a chat thread that has less than 20 participants, all message status (sending/delivered/failed and seen) should be displayed properly
- In a chat thread that has more than 20 participants, sending/delivered/failed message status should be displayed properly. 
- In a large group(more than 20 participants), messages will never be marked as seen.

# Process & policy checklist


packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -203,7 +203,7 @@ export const messageThreadSelector: MessageThreadSelector = createSelector(
     updateMessagesWithAttached(convertedMessages);
     return {
       userId,
-      showMessageStatus: !isLargeGroup,
+      showMessageStatus: true,
```
PorterNan, line 206: What if we set this as undefined? Will that be default on or off?

PorterNan (approved):
This small fix is nice, I highly recommend we adding a hermetic test for this as a following PR, to make sure we always show status and not showing 'seen' status

packages/chat-component-bindings/src/messageThreadSelector.ts
PorterNan: 'Set it as undefined' means not to set it

Leah-Xia-Microsoft: > 'Set it as undefined' means not to set it

Setting it to undefined will cause it to use the default value provided in MessageThread.ts which is false which is contradictory with our current behaviour.

emlynmac (approved):
I'm curious if this should be configurable behaviour.

Leah-Xia-Microsoft:
> I'm curious if this should be configurable behaviour.

Contoso can configure whether to show send status (including read receipt) or not. However, they cannot configure the behaviour based on whether it's a large group.

PR #2710, opened 2023-02-06
[Storybook Update] Add simplified Try ChatComposite story to use fakeAdapter

jimchou-dev:
# What

- Add 'Try Chat Composite' story that uses fakeAdapter instead of needing token from AzureCommunication Token
- Updated bot messages to make more sense for the example context
- Make Show ParticipantsPane working for Basic Example
- Minor typo fix for aria label for Participant Pane
image

# Why

Currently it's a bit complex for Contoso to get started requiring tokens to see the ChatComposite in action, changing basic example to use FakeAdapter to remove this friction
https://skype.visualstudio.com/SPOOL/_workitems/edit/2626713

# How Tested

Tested on Mac Chrome and Safari

# Process & policy checklist


PorterNan (commented):
This is a fairly old topic, I like that we use fake adapter to make playground easier to try out, but this playground is also a good place to allow user to take their first step to see how everything works together, otherwise it would be hard to demonstrate relationship between tokens to an active session

And also lots of people use this as a validation app to connect back to their own app, removing this option would force developers to implement a whole app to try the real ACS services. 

So the idea here is to keep both option - if someone wants to have a quick try, they could use fake adapter, but we will still allow them to connect to our service directly

BTW, if I remember it correct, this work items was targeting on removing the need of connection string, we've resolved it by asking for a token, so we are almost good on this one already - but I think it would be nice we can add fake adapter as an addition to the playground

jimchou-dev:
> This is a fairly old topic, I like that we use fake adapter to make playground easier to try out, but this playground is also a good place to allow user to take their first step to see how everything works together, otherwise it would be hard to demonstrate relationship between tokens to an active session
> 
> And also lots of people use this as a validation app to connect back to their own app, removing this option would force developers to implement a whole app to try the real ACS services.
> 
> So the idea here is to keep both option - if someone wants to have a quick try, they could use fake adapter, but we will still allow them to connect to our service directly
> 
> BTW, if I remember it correct, this work items was targeting on removing the need of connection string, we've resolved it by asking for a token, so we are almost good on this one already - but I think it would be nice we can add fake adapter as an addition to the playground

Sorry wasn't aware of the history for this ticket, updated the PR to be a separate page

JamesBurnside:
Just tried it out and this is awesome!

I wonder, instead of its own story, how do folks feel about this being in the primary doc instead of the static image?
image

We could prepopulate it with some fun messages and avatar images like in the static screenshot

jimchou-dev:
> Just tried it out and this is awesome!
> 
> I wonder, instead of its own story, how do folks feel about this being in the primary doc instead of the static image? image
> 
> We could prepopulate it with some fun messages and avatar images like in the static screenshot

I like that Idea! And maybe the message could include some aka.ms links and message with image screenshot to walk developer through how to do certain things, but I think that can be a new feature task to revamp existing storybook pages (maybe working with PM for that)

edwardlee-msft:
> Just tried it out and this is awesome!
> 
> I wonder, instead of its own story, how do folks feel about this being in the primary doc instead of the static image? image
> 
> We could prepopulate it with some fun messages and avatar images like in the static screenshot

this is a GREAT idea!!

JamesBurnside:
LGTM! Can you get a sign off from PM on this PR as it changes the doc? I'd like their opinion whether this new page differs enough from basic sample to be clear to users

JamesBurnside (approved):

jimchou-dev:
@ddematheu2 can you sign off on this?

edwardlee-msft (approved):

jimchou-dev:
> LGTM! Can you get a sign off from PM on this PR as it changes the doc? I'd like their opinion whether this new page differs enough from basic sample to be clear to users

Confirmed with David, he is good to go with these changes

PR #2711, opened 2023-02-06
Export useFakeChatAdapters et. al. from the react-composites dir

JamesBurnside:
# What
Export useFakeChatAdapters et. al. from the react-composites dir

# Why
So they can be used easily in other packlets (in this case - storybook)

# How Tested
CI should test this.


JamesBurnside:
Closing in favor of #2714

PR #2713, opened 2023-02-07
[BugFix] Update streamUtils to work better with disposing local video

dmceachernmsft:
# What

There was a race condition between CallSubscriber, LocalVideoTile, and Camera button. updated to better handle this. Camera button no longer directly calls dispose view.
# Why
 
Camera button kept losing the race throwing 
image
Video tile result now properly returns saying the view was already disposed of. 

https://skype.visualstudio.com/SPOOL/_workitems/edit/3135605
# How Tested

Ran locally error no longer happens and camera properly turns off. will bug bash with new beta

### DIFF
--- a/packages/calling-stateful-client/src/StreamUtils.ts
+++ b/packages/calling-stateful-client/src/StreamUtils.ts
@@ -507,8 +507,6 @@ function disposeViewLocalVideo(context: CallContext, internalContext: InternalCa
     data: streamLogInfo
   });
 
-  context.setLocalVideoStreamRendererView(callId, undefined);
-
   const renderInfo = internalContext.getLocalRenderInfo(callId);
   if (!renderInfo) {
     _logEvent(callingStatefulLogger, {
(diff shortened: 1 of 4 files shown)


packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -568,6 +563,11 @@ function disposeViewLocalVideo(context: CallContext, internalContext: InternalCa
       data: streamLogInfo
     });
     renderInfo.renderer.dispose();
+
+    // Else the state must be in the "Rendered" state, so we can dispose the renderer and clean up the state.
```
JamesBurnside, line 567: move/update this comment, the `else` is confusing

JamesBurnside (approved):
Let's be sure to give this a hardy bug bash in the beta release bug bash to make sure it doesn't regress anything unexpected

edwardlee-msft (approved):

PR #2714, opened 2023-02-07
Allow use of fake adapters outside of the react-composites packlet

JamesBurnside:
# What
Export useFakeChatAdapters et. al. from the react-composites dir

# Why
So they can be used easily in other packlets (in this case (#2710) - storybook)

# How Tested
[Test CI run](https://github.com/Azure/communication-ui-library/actions/runs/4111009209)


packages/fake-backends/src/FakeChatClient.ts
```diff
@@ -26,8 +26,13 @@ import { getThreadEventTargets } from './ThreadEventEmitter';
 export class FakeChatClient implements IChatClient {
   private realtimeNotificationsEnabled = false;
   private threadClients: FakeChatThreadClient[] = [];
+  private model: Model;
+  private userId: CommunicationIdentifier;
 
-  constructor(private model: Model, private userId: CommunicationIdentifier) {}
```
JamesBurnside, line 30: jest w/ babel didn't like the implicit `private` declaration within the constructor so moved out

edwardlee-msft (approved):

edwardlee-msft:
Wow! This is great!! 
Definitely can see this being useful for other storybook snippets :D

PR #2721, opened 2023-02-08
Fix VideoTile menu button to show on hover when isSpeaking is true

mgamis-msft:
# What

Remove zIndex from VideoTIle and its isSpeaking border. Reordered LayerHost in FloatingLocalVideoLayout to be before the VideoTiles.

# Why

Bug from bug bash where VideoTile does not show menu button on hover when isSpeaking is true.
https://user-images.githubusercontent.com/79475487/217512366-5b7d315b-7546-483d-b745-200dd0f4f327.mp4

# How Tested

Tested VideoTile:
https://user-images.githubusercontent.com/79475487/217512554-13a141fc-d8b3-4107-a3b8-e4b4e340297e.mp4

Tested VideoGallery floating local video is still draggable:
https://user-images.githubusercontent.com/79475487/217512595-51916059-4d05-48ae-96d3-1812f27afc08.mp4

# Process & policy checklist


edwardlee-msft (approved):

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -333,19 +333,17 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
...
+          '&::after': {
             content: `''`,
             position: 'absolute',
-            zIndex: 1,
             border: `0.25rem solid ${theme.palette.themePrimary}`,
             borderRadius: theme.effects.roundedCorner4,
             width: '100%',
-            height: '100%'
+            height: '100%',
+            pointerEvents: 'none'
```
dmceachernmsft, line 346: will this pointer event block any interactions with keyboard if on mobile? or is this just for the border?

mgamis-msft: This won't block mouse or keyboard interactions with the VideoTile. This will just prevent mouse interactions with the [after pseudo-element](https://developer.mozilla.org/en-US/docs/Web/CSS/::after). Needed to do this because it blocks the onHover events on the VideoTile.

dmceachernmsft (approved):

PR #2726, opened 2023-02-08
Switch video seamlessly from different devices

PorterNan:
# What

https://skype.visualstudio.com/SPOOL/_workitems/edit/3136884

Did some investigation, here is the summary:
1. If you join from CTE + teams client, multiple video and screen share stream will be created
2. In our code, we assume there would be only 1 video stream + 1 screen stream
3. When joining in both way, there would be only first video steam picked up by our library, which makes another one not working

Fix:
1. Add a logic to get the "available" stream as our first priority, if not then follow the existing logic

Now video stream can switch seamlessly from 2 different devices - it will pick up the first available stream and render it:

https://user-images.githubusercontent.com/11863655/217670206-bb3d7e53-ade9-4b08-8658-48fd221ecc7e.mp4

# Why

# How Tested

# Process & policy checklist


packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -85,22 +85,21 @@ export const convertRemoteParticipantToVideoGalleryRemoteParticipant = (
...
-      screenShareStream = convertRemoteVideoStreamToVideoGalleryStream(rawVideoStreamsArray[1]);
-    } else {
-      videoStream = convertRemoteVideoStreamToVideoGalleryStream(rawVideoStreamsArray[1]);
-    }
+  const sdkScreenShareStream =
+    Object.values(rawVideoStreamsArray).find((i) => i.mediaStreamType === 'ScreenSharing' && i.isAvailable) ||
+    Object.values(rawVideoStreamsArray).find((i) => i.mediaStreamType === 'ScreenSharing');
+
+  if (!sdkRemoteVideoStream || !sdkScreenShareStream) {
+    throw new Error('Video stream or screen share stream is undefined');
```
JamesBurnside, line 97: Why do they both have to be defined here? I'f i"m not screensharing won't sdkScreenShareStream be undefined?

JamesBurnside (approved):

edwardlee-msft (approved):

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
PorterNan: Every participant joins the call will always have at least 2 streams

Even you don't turn anything on, they will be available=false

Also it will cause much worse problem from previous logic - we direct visit streamArray[1] index of the stream array, since we don't get any serve bug for now, we can confirm there are always at least 2 streams in the array

PorterNan: Hmm, double checked the logic, looks like there is a check for undefined of previous logic, lemme fix this

PR #2728, opened 2023-02-09
Populate localVideoStreamState state with video effects status

JamesBurnside:
# What

When a new local video stream is created we subscribe to changes (effectsStarted, effectsStopped and effectsError), subscriptions happen in two places:
1. Inside a call, we subscribe when call.onUpdateLocalVideoStreams is triggered with `added`
   - we unsubscribe when call.onUpdateLocalVideoStreams is triggered with `removed`
2. When an unparented view is created
   - we unsubscribe when that view is destroyed

### What's not in this PR

- Calling currently do not expose an isActive properties for streams with effects already applied. They are adding this in the near future (marked as TODO in code)
- Errors are not propagated to state (marked as TODO in code)

# Why
Getting video effects status of the local video stream into state

# How Tested
Manually trigger local video streams and verified subscriptions where fired when in config screen and in a cal, and verified localVideoStreamState correctly had the new state properties

image


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -625,20 +653,36 @@ export class CallContext {
...
-  ): void {
+  public setDeviceManagerUnparentedView(args: {
+    localVideoStream: LocalVideoStreamState;
+    view: VideoStreamRendererViewState | undefined;
+    /* @conditional-compile-remove(video-background-effects) */
+    localVideoStreamEffectsAPI: VideoEffectsFeature;
+  }): void {
+    /* @conditional-compile-remove(video-background-effects) */
+    {
+      this._unparentedViewVideoEffectsSubscriber?.unsubscribe();
```
dmceachernmsft, line 664: We are invoking the unsubscribe here to just to make sure there isn't a unparented view already? is there anything stopping us from having more than one unparented view?

dmceachernmsft (approved):

packages/calling-stateful-client/src/CallContext.ts
JamesBurnside: Just a quick failsafe if setDeviceManagerUnparentedView is called twice in a row we don't set up multiple subscriptions

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -625,20 +653,36 @@ export class CallContext {
...
   }
 
-  public setDeviceManagerUnparentedView(
-    localVideoStream: LocalVideoStreamState,
-    view: VideoStreamRendererViewState | undefined
-  ): void {
+  public setDeviceManagerUnparentedView(args: {
+    localVideoStream: LocalVideoStreamState;
+    view: VideoStreamRendererViewState | undefined;
+    /* @conditional-compile-remove(video-background-effects) */
```
PorterNan, line 659: is it because of conditional compilation that we wrap it using an args?

packages/calling-stateful-client/src/CallSubscriber.ts
```diff
@@ -32,6 +34,8 @@ export class CallSubscriber {
   private _participantSubscribers: Map<string, ParticipantSubscriber>;
   private _recordingSubscriber: RecordingSubscriber;
   private _transcriptionSubscriber: TranscriptionSubscriber;
+  /* @conditional-compile-remove(video-background-effects) */
+  private _localVideoStreamVideoEffectsSubscribers: Map<string, LocalVideoStreamVideoEffectsSubscriber>;
```
PorterNan, line 38: Do we have any other events to be subscribed from stream in near future? If yes then it would be nice to name it LocalStreamSubScriber and put it all there

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -625,20 +653,36 @@ export class CallContext {
...
+    localVideoStreamEffectsAPI: VideoEffectsFeature;
+  }): void {
+    /* @conditional-compile-remove(video-background-effects) */
+    {
+      this._unparentedViewVideoEffectsSubscriber?.unsubscribe();
+      this._unparentedViewVideoEffectsSubscriber = new LocalVideoStreamVideoEffectsSubscriber({
```
PorterNan, line 665: I can see why we put a subscriber here - cause we only create localStream in StreamUtil function then it's hard to get the stream and subscribe to it in other functions

But this adds complexity to the overall dependency graph, previously call context is only responsible for getting and setting data and all the other logics are in subscriber/client
But now it's like subscriber->context->subscriber->context, so it could be easily a circular ref when someone doesn't understand this well

I guess either finding a way to create stream outside of context or getting stream from internal context after the creation or even creating subscriber in internal context would be nice

Lets keep CallContext only the final stop for data and logic to flow in ;D

PorterNan (approved):

packages/calling-stateful-client/src/CallSubscriber.ts
JamesBurnside: Oh good point, I'll look at making this more generic so new localVideoStream subscriptions could be easily added

packages/calling-stateful-client/src/CallContext.ts
JamesBurnside: nope, just many positional arguments, some of which may be undefined so I figured it made sense to make them named arguments now. Let me know if you want to revert back to positional args

packages/calling-stateful-client/src/CallContext.ts
JamesBurnside: great point, will look into where else it could go...

packages/calling-stateful-client/src/CallContext.ts
JamesBurnside: oh turns out it was for conditional comp lol, so i guess both reasons haha

packages/calling-stateful-client/src/CallContext.ts
JamesBurnside: Moved the subscription logic into internalContext, not super ideal as then the internalcontext kinda has a handle to the regular context but changes
`subscriber->context->subscriber->context`
to
`subscriber->internalcontext->subscriber->context`
which is a little nicer!

packages/calling-stateful-client/src/CallSubscriber.ts
JamesBurnside: Going to leave this as is. If someone were to create a local video stream subcriber for different events (or a declarativelocalvideostream) they would use this subscriber themselves.

Which from playing around with how this feature should be implemented, could look something like this!~

```ts

import { Features, LocalVideoStream } from '@azure/communication-calling';
import { CallContext } from './CallContext';
import { CallIdRef } from './CallIdRef';
import { convertSdkLocalStreamToDeclarativeLocalStream } from './Converter';
import { LocalVideoStreamVideoEffectsSubscriber } from './LocalVideoStreamVideoEffectsSubscriber';

/**
 * @private
 */
export class ProxyLocalVideoStream implements ProxyHandler<LocalVideoStream> {
  private _context: CallContext;
  private _callIdRef: CallIdRef | 'unparented';
  private _localVideoStreamVideoEffectsSubscriber?: LocalVideoStreamVideoEffectsSubscriber;

  constructor(context: CallContext, callId: CallIdRef | 'unparented') {
    this._context = context;
    this._callIdRef = callId;
  }

  public construct(target: LocalVideoStream): LocalVideoStream {
    const statefulLocalVideoStream = convertSdkLocalStreamToDeclarativeLocalStream(target);
    const effectsApi = target.feature(Features.VideoEffects);
    this._localVideoStreamVideoEffectsSubscriber = new LocalVideoStreamVideoEffectsSubscriber({
      parent: this._callIdRef,
      context: this._context,
      localVideoStream: statefulLocalVideoStream,
      localVideoStreamEffectsAPI: effectsApi
    });
    return new Proxy(target, this);
  }

  public unsubscribe(): void {
    if (this._localVideoStreamVideoEffectsSubscriber) {
      this._localVideoStreamVideoEffectsSubscriber.unsubscribe();
    }
  }

  public get<P extends keyof LocalVideoStream>(target: LocalVideoStream, prop: P): any {
    switch (prop) {
      case 'feature': {
        return async (...args: Parameters<LocalVideoStream['feature']>) => {
          if (args[0] === Features.VideoEffects) {
            const feature = target.feature(Features.VideoEffects);
            return {
              ...feature,
              dispose: (...args: Parameters<typeof feature.dispose>) => {
                this.unsubscribe();
                return feature.dispose(...args);
              }
            };
          } else {
            return target.feature(...args);
          }
        };
      }
      default:
        return Reflect.get(target, prop);
    }
  }
}

/**
 * Creates a declarative LocalVideoStream that is backed by a LocalVideoStream from the SDK.
 * Calling methods on this declarative object triggers state updates in the stateful client.
 */
export const localVideoStreamDeclaratify = (
  view: LocalVideoStream,
  context: CallContext,
  callId: CallIdRef | 'unparented'
): LocalVideoStream => {
  const proxyLocalVideoStream = new ProxyLocalVideoStream(context, callId);
  return new Proxy(view, proxyLocalVideoStream) as LocalVideoStream;
};

export {};
```

packages/calling-stateful-client/src/CallContext.ts
JamesBurnside: With the move of logic to internalcontext this change was removed - resolving!

PR #2731, opened 2023-02-09
Add `getUnparentedLocalVideoStreams()` to the StatefulDeviceManager

JamesBurnside:
# What
Add `getUnparentedLocalVideoStreams()` to the StatefulDeviceManager

# Why
For video backgrounds effect feature.
Handlers need a way to call `localVideoStream.startEffects` but currently we do not expose the video streams created in createView.

This solution isn't ideal but determined to be the best of three available options; this could be done in two other ways:
- expose startEffects in the LocalVideoStreamState -- this goes against the principles of our client state being readonly properties
- return localvideostream when createView is returned -- this means handlers would need to become stateful and store the local video stream returned in create view, or not support usePropsFor and make the application store the local video stream.

TBD: write a doc for API review that summarizes this.

# How Tested
Locally checked that creating a handler on the configuration page could use these newly exposed to start video effects:
image
image


packages/calling-stateful-client/src/DeviceManagerDeclarative.ts
```diff
@@ -16,6 +19,16 @@ export interface StatefulDeviceManager extends DeviceManager {
    * {@link StatefulDeviceManager}. See also {@link DeviceManagerState.selectedCamera}.
    */
   selectCamera: (VideoDeviceInfo) => void;
+
```
edwardlee-msft, line 22: I think this is neat solution. 
I do not quite understand why this is not most ideal solution according to your pr description. 
I'll approve as it seems okay to me, but I may rereview depending on better understanding

edwardlee-msft (approved):

dmceachernmsft (approved):
Very cool solution approving. Just wondering though do we have examples somewhere to see the statefulDeviceManager in use? I've seen the file before but I don't think I have interacted with it yet.

PR #2733, opened 2023-02-09
[Beta][Draft] Add sendBox focus to composites

emlynmac:
# What
Adds the ability to focus the chat send box in our chat composite and call-with-chat composite, updates the samples to use it and puts the new functionality behind a beta flag

# Why
Improvement to the usability and flexibility.

# How Tested
Tested with chat sample app

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -44,8 +47,21 @@ export interface ChatCompositeProps extends BaseCompositeProps<ChatCompositeIcon
    * @defaultValue 'desktop'
    */
   formFactor?: 'desktop' | 'mobile';
+
+  /* @conditional-compile-remove(chat-reference-support) */
+  /**
+   * A callback to use as a reference to the composite.
+   * @beta
+   */
+  compositeRef?: (ref: ChatCompositeRefProps) => void;
```
dmceachernmsft, line 56: Overall this is looking good, a open question I have though is what does this do that we don't already provide with the `autofocus` property in the `ChatCompositeOptions`. We provide this as a way for contoso to tell the composite to steal focus to the send box when it is rendered.

emlynmac: at the current time, nothing. 
I think the original issue was to provide a way to expand upon the boolean auto focus in a way that would allow focus / configuration for different parts within the composite. The ref does that - maybe we need a more concrete feature ask before merging?

PR #2739, opened 2023-02-10
[BugFix BetaBlocking] Fix PSTN start Call

dmceachernmsft:
# What

Update `fromFlatCommunicationIdentifier` function to handle E.164 format phone numbers as Id's
# Why

Allows for startCall handler in Adapter to parse out correct id type.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3141060
# How Tested

Validated calls can be connected again locally. added unit test for util function that parses id


mgamis-msft (approved):

edwardlee-msft:
Thanks for this quick hotfix Don! Master class OCE

edwardlee-msft (approved):

PR #2752, opened 2023-02-15
Add more url parameters for chat sample

palatter:
# What

# Why

# How Tested

# Process & policy checklist


samples/Chat/src/app/utils/getExistingDisplayNameFromURL.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ *
+ * The displayName is extracted from the url
+ * using URLsearchparams.
+ *
+ * @returns The displayName as String
+ *
+ */
+
+export const getExistingDisplayNameFromURL = (): string | null => {
```
dmceachernmsft, line 13: I think each of these queryParams functions can live in the same file? `QueryParamLoginUtils.ts` or something

samples/Chat/src/app/ConfigurationScreen.tsx
```diff
@@ -129,6 +174,16 @@ export default (props: ConfigurationScreenProps): JSX.Element => {
           return;
         }
         setConfigurationScreenState(CONFIGURATIONSCREEN_SHOWING_JOIN_CHAT);
+
+        // Join with URL Parameters 
+        const token = getExistingTokenFromURL();
```
dmceachernmsft, line 179: If we are invoking these here to check if we have all the params we need could we not just pass them into the `joinChatThreadWithExistingUser` callback?

samples/Chat/src/app/ConfigurationScreen.tsx
```diff
@@ -116,6 +124,43 @@ export default (props: ConfigurationScreenProps): JSX.Element => {
     internalSetupAndJoinChatThread();
   }, [name, joinChatHandler, selectedAvatar, setDisplayName, setEndpointUrl, setThreadId, setToken, setUserId]);
 
+  const joinChatThreadWithExistingUser = useCallback(() => {
+    const internalSetupAndJoinChatThread = async (): Promise<void> => {
```
dmceachernmsft, line 128: why do we have a wrapper on this? should this just be the function itself? does this also need to be async?

dmceachernmsft (approved):

samples/Chat/src/app/ConfigurationScreen.tsx
```diff
@@ -129,10 +150,21 @@ export default (props: ConfigurationScreenProps): JSX.Element => {
...
+        // Join with URL Parameters
+        const threadId = getExistingThreadIdFromURL();
+        const token = getExistingTokenFromURL();
+        const userId = getExistingUserIdFromURL();
+        const displayName = getExistingDisplayNameFromURL();
+        const endpointUrl = getExistingEndpointURLFromURL();
+
+        if (token && userId && displayName && threadId && endpointUrl) {
+          joinChatThreadWithExistingUser(token, userId, displayName, threadId, endpointUrl);
+        }
```
JamesBurnside, line 163: Move this line to below `joinChatThreadWithExistingUser` as `joinChatThreadWithExistingUser` is replacing what is set here.

```suggestion
        // Check if we have all the required parameters supplied as query search params.
        const threadId = getExistingThreadIdFromURL();
        const token = getExistingTokenFromURL();
        const userId = getExistingUserIdFromURL();
        const displayName = getExistingDisplayNameFromURL();
        const endpointUrl = getExistingEndpointURLFromURL();

        if (token && userId && displayName && threadId && endpointUrl) {
          joinChatThreadWithExistingUser(token, userId, displayName, threadId, endpointUrl);
          return;
        }
        
        // Else show the join chat screen where a user enters there display name and the other args are collected from the server
        setConfigurationScreenState(CONFIGURATIONSCREEN_SHOWING_JOIN_CHAT);
```

JamesBurnside (approved):

PR #2753, opened 2023-02-16
[StreamUtils updates] Move logging out of stream utils

dmceachernmsft:
# What

Moves the use of the stateful logger out to a separate file from streamUtils.

fixes small bug where we are clearing the renderer state from the internal context for the remote stream before we are disposing of it.
# Why

Simplifies the code in the file to cut down on size

fixes similar case for localtile where renderer is still in state but deleted. causes error for remote participant.

# How Tested

Just logging changes, ran jest tests locally to make sure updates didn't break them and ran samples locally with multiple participants.


packages/calling-stateful-client/src/StreamUtilsLogging.ts
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * helper function to manage logging for stream disposals
+ *
+ * @param eventName Name of event that occured when managing streams
+ * @param streamLogInfo Data about the stream in the event
+ * @returns
+ */
+function _logDisposeStreamEvent(
+  eventName: string,
```
JamesBurnside, line 16: nit: can we use the enum here as a type?
```suggestion
  eventName: EventNames,
```

JamesBurnside (approved):

alkwa-msft (approved):
Much easier to read and we can infer the type of the stream from the function we invoked it from. Great stuff!

PR #2755, opened 2023-02-16
[StreamUtils updates] one CreateView to rule them all

dmceachernmsft:
# What

Wraps `createViewLocalVideo` and `createViewRemoteVideo` into one function
# Why

works towards generalizing streamUtils so it can live in the internalContext.

# How Tested

tested locally, and validated refactor does not break anything in jest tests


packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -30,210 +31,214 @@ export type CreateViewResult = {
...
-    _logCreateRemoteStreamEvent(EventNames.CREATE_REMOTE_STREAM_FAIL, streamLogInfo);
-    internalContext.setRemoteRenderInfo(callId, participantKey, streamId, renderInfo.stream, 'NotRendered', undefined);
-    throw e;
+  if (participantId) {
+    createViewType = 'remote';
+  } else if (callId) {
+    createViewType = 'local';
+  } else {
+    // TODO update for when unparented view.
+    createViewType = 'local';
```
JamesBurnside, line 55: throw error here instead?

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -30,210 +31,214 @@ export type CreateViewResult = {
...
-  }
-
-  if (renderInfo.status === 'Rendering') {
-    // Do not log to console here as this is a very common situation due to UI rerenders while
-    // the video rendering is in progress.
-    _logCreateRemoteStreamEvent(EventNames.REMOTE_STREAM_RENDERING, streamLogInfo);
-    return;
-  }
+  // we can only have 3 types of createView
+  let createViewType: 'local' | 'remote' | 'unparented';
```
JamesBurnside, line 43: this is only set once, consider making constant and set the value at the time of declaration

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -30,210 +31,214 @@ export type CreateViewResult = {
...
-      renderInfo.renderer
-    );
-    return;
-  }
-
-  const renderer = new VideoStreamRenderer(renderInfo.stream);
-
-  internalContext.setRemoteRenderInfo(callId, participantKey, streamId, renderInfo.stream, 'Rendering', undefined);
+  // we will reuse these for local as well but we need to make sure the remote stream is passed in like before.
+  let streamId;
```
JamesBurnside, line 46: same here, these are only set once - if you can make these constant and set at the time of declaration

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -30,210 +31,214 @@ export type CreateViewResult = {
...
-      participantKey,
-      streamId,
-      refreshedRenderInfo.stream,
-      'NotRendered',
-      undefined
-    );
-    context.setRemoteVideoStreamRendererView(callId, participantKey, streamId, undefined);
-    return;
+  // we want to check to see if there is a participantId this will tell us whether its a local stream or a remote one.
+  let participantKey;
```
JamesBurnside, line 66: nit: +here too `const participantKey =`

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -30,210 +31,214 @@ export type CreateViewResult = {
...
-  if (!refreshedRenderInfo) {
-    // RenderInfo was removed. This should not happen unless stream was removed from the call so dispose the renderer
-    // and clean up state.
-    _logCreateRemoteStreamEvent(EventNames.REMOTE_RENDER_INFO_NOT_FOUND, streamLogInfo);
-    renderer.dispose();
-    context.setRemoteVideoStreamRendererView(callId, participantKey, streamId, undefined);
-    return;
+  if (stream) {
+    streamType = stream.mediaStreamType;
+    if (createViewType === 'remote') {
```
JamesBurnside, line 60: local should also have an id

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -30,210 +31,214 @@ export type CreateViewResult = {
...
-      streamId,
-      refreshedRenderInfo.stream,
-      'NotRendered',
-      undefined
-    );
-    context.setRemoteVideoStreamRendererView(callId, participantKey, streamId, undefined);
-    return;
+  // we want to check to see if there is a participantId this will tell us whether its a local stream or a remote one.
+  let participantKey;
+  if (createViewType === 'remote' && participantId) {
```
JamesBurnside, line 67: can we include retrieving local participant id also? (useful for logs)

packages/calling-stateful-client/src/StreamUtils.ts
JamesBurnside: (useful for logs)

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -30,210 +31,214 @@ export type CreateViewResult = {
...
-): Promise<CreateViewResult | undefined> {
-  // Render LocalVideoStream that is part of a Call
-  const renderInfo = internalContext.getLocalRenderInfo(callId);
-
-  const streamType = renderInfo?.stream.mediaStreamType;
+  const streamLogInfo = { callId, participantKey, streamId, streamType };
 
-  const streamLogInfo = { callId, streamType };
+  // make different logging announcement based on whether or not we are starting a local or remote
+  createViewType === 'remote'
```
JamesBurnside, line 78: instead of every log having to check if its local or remote and having different event names, could we do:
```
const streamLogInfo = { ..., createViewType };
_logStreamEvent(EventNames.CreateView, streamLogInfo)
```

packages/calling-stateful-client/src/StreamUtils.ts
JamesBurnside: image

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -30,152 +31,58 @@ export type CreateViewResult = {
...
 
-  const streamLogInfo = { callId, streamType, streamEventType: 'createViewLocal' };
+  // make different logging announcement based on whether or not we are starting a local or remote
+  _logStreamEvent(EventNames.CREATING_VIEW, streamLogInfo);
 
-  _logStreamEvent(EventNames.START_STREAM_RENDERING, streamLogInfo);
+  // if we have a participant Id and a stream get the remote info, else get the local render info from state.
+  const renderInfo =
+    streamEventType === 'createViewRemote' && participantKey
+      ? internalContext.getRemoteRenderInfoForParticipant(callId, participantKey, streamId)
```
JamesBurnside, line 76: not for this pr, this pr already does loads of goodness!
But we should make the `internalContext.set/get` interfaces agnostic of remote vs local to avoid all these ternaries we have throughout.

JamesBurnside (approved):

edwardlee-msft (approved):

PR #2757, opened 2023-02-17
update beta calling dependency

alkwa-msft:
Updating calling dependencies


change/@azure-communication-react-bfa86c01-5c65-4b9b-854b-ec9ae0793ecb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
dmceachernmsft, line 2: would this be a none change? or are package updates considered patches even though they aren't fixing anything?

dmceachernmsft (approved):

PR #2770, opened 2023-02-22
[VerticalGallery] Responsive vertical gallery wrapper

dmceachernmsft:
# What

Draft for responsive wrapper on VerticalGallery component
# Why

# How Tested


packages/react-components/src/components/VideoGallery/styles/VideoGalleryResponsiveVerticalGallery.styles.tsx
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Styles for the VerticalGallery's container set in parent.
+ *
+ * @param shouldFloatLocalVideo whether rendered in floating layout or not
+ * @param isNarrow is being rendered with narrow tile settings.
+ * @returns Style set for VerticalGallery container.
+ */
+export const verticalGalleryContainerStyle = (shouldFloatLocalVideo: boolean, isNarrow: boolean): IStyle => {
+  return {
+    width: `${LARGE_VERTICAL_GALLERY_TILE_SIZE_REM.width + 1}rem`,
```
prabhjot-msft, line 27: why is additional width added here and not when the value is defined

prabhjot-msft (commented):
just a minor comment, otherwise looks good and if there are any visible CSS changes can you add some screenshots

packages/react-components/src/components/VerticalGallery.tsx
```diff
@@ -114,8 +114,14 @@ export const VerticalGallery = (props: VerticalGalleryProps): JSX.Element => {
 
   return (
     <Stack className={mergeStyles(rootStyle, styles?.root)}>
-      <Stack className={mergeStyles(childrenContainerStyle, { '> *': styles?.children })}>
-        {childrenOnCurrentPage}
+      <Stack styles={{ root: childrenContainerStyle }}>
+        {childrenOnCurrentPage.map((child) => {
+          return (
+            <Stack.Item key={childrenOnCurrentPage.indexOf(child)} styles={{ root: styles?.children }}>
```
JamesBurnside, line 120: nit: instead of `indexOf` in `key={childrenOnCurrentPage.indexOf(child)}`, `.map` gives you the index nicely:
```
{childrenOnCurrentPage.map((child, i) => {
   ....
    key={i}
```

packages/react-components/src/components/VideoGallery/styles/VideoGalleryResponsiveVerticalGallery.styles.tsx
dmceachernmsft: the width is added to account for the space needed to hold the individual tiles with a nice gap. I will update the comment 😊

packages/react-components/src/components/ResponsiveVerticalGallery.tsx
```diff
@@ -0,0 +1,112 @@
...
+ */
+const calculateChildrenPerPage = (args: {
+  numberOfChildren: number;
+  containerHeight: number;
+  gapHeightRem: number;
+  controlBarHeight: number;
+}): number => {
+  const { numberOfChildren, containerHeight, gapHeightRem, controlBarHeight } = args;
+
+  const childMinHeightPx = 90;
```
JamesBurnside, line 69: can we use LARGE_VERTICAL_GALLERY_TILE_SIZE_REM instead here? (I assume these values need to be kept consistent)

packages/react-components/src/components/ResponsiveVerticalGallery.tsx
```diff
@@ -0,0 +1,112 @@
...
+        {children}
+      </VerticalGallery>
+    </div>
+  );
+};
+
+/**
+ * Helper function to find the number of children for the VerticalGallery on each page.
+ */
+const calculateChildrenPerPage = (args: {
```
JamesBurnside, line 61: this is the kinda function we should really have a unit test for... (or a unit test for - given these children and this size of vertical gallery I expect `n` children and `m` pages) but I wouldn't block PR on it

JamesBurnside (approved):

packages/react-components/src/components/VerticalGallery.tsx
```diff
@@ -114,8 +114,14 @@ export const VerticalGallery = (props: VerticalGalleryProps): JSX.Element => {
 
   return (
     <Stack className={mergeStyles(rootStyle, styles?.root)}>
-      <Stack className={mergeStyles(childrenContainerStyle, { '> *': styles?.children })}>
-        {childrenOnCurrentPage}
+      <Stack styles={{ root: childrenContainerStyle }}>
```
JamesBurnside, line 117: nit: this object should be memoized

packages/react-components/src/components/VerticalGallery.tsx
```diff
@@ -114,8 +114,14 @@ export const VerticalGallery = (props: VerticalGalleryProps): JSX.Element => {
...
+        {childrenOnCurrentPage.map((child, i) => {
+          return (
+            <Stack.Item key={i} styles={{ root: styles?.children }}>
```
JamesBurnside, line 120: likewise here (memoize styles object being created)

packages/react-components/src/components/VideoGallery/styles/VideoGalleryResponsiveVerticalGallery.styles.tsx
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation.
```
mgamis-msft, line 1: This styles file type should be styles.ts

packages/react-components/src/components/ResponsiveVerticalGallery.tsx
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * number of children = container height - (2* gap height + button height) / childMinHeight
+   *
+   * we want to find the maximum number of children at the smallest size we can fit in the gallery and then resize them
+   * to fill in the space as much as possible
+   */
+
+  const numberOfChildrenInContainer = Math.floor(
+    (containerHeight - (2 * gapHeightPx + controlBarHeightPx)) / childMinHeightPx
+  );
```
mgamis-msft, line 95: This first calculation should be to check if all children will fit into the container height without needing to paginate. So you should not account for the control bar height and the gaps for it. Also you must take into account the gaps between the children like so:
containerHeight = n * childMinHeightPx+ (n - 1) * gapHeightPx -> isolate n and take the floor -> Math.floor((containerWidth + gapWidth) / (childWidth + gapWidth))

packages/react-components/src/components/ResponsiveVerticalGallery.tsx
```diff
@@ -0,0 +1,113 @@
...
+   */
+
+  const maxNumberOfChildrenInContainer = Math.floor((containerHeight + gapHeightPx) / (childMinHeightPx + gapHeightPx));
+  // if all of the children fit in the container just return the number of children
+  if (numberOfChildren <= maxNumberOfChildrenInContainer) {
+    return maxNumberOfChildrenInContainer;
+  }
+
+  /**
+   * For the paginastion we know the container height, the height of the button bar and the 2 times the gap
```
mgamis-msft, line 102: typo: pagination

mgamis-msft (approved):

PR #2774, opened 2023-02-24
Added overflowGalleryLayout prop to VideoGallery

mgamis-msft:
# What

- Added overflowGalleryLayout prop to VideoGallery to control which component to use for overflow participants
- Renamed VideoGalleryResponsiveHorizontalGallery -> OverflowGallery
- Refactor FloatingLocalVideo component to simply have localVideoSize prop instead of using isNarrow boolean prop to decide the size internally

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3151770

# How Tested

Local storybook preview testing. Relying on snapshot tests on CI.

# Process & policy checklist


packages/react-components/src/components/VideoGallery/styles/FloatingLocalVideo.styles.ts
```diff
@@ -38,6 +38,12 @@ export const SMALL_FLOATING_MODAL_SIZE_PX = { width: 58, height: 104 };
  */
 export const LARGE_FLOATING_MODAL_SIZE_PX = { width: 215, height: 120 };
 
+/**
+ * Vertical gallery floating modal width and height in rem
+ * Aspect ratio: 16:9
+ */
+export const VERTICAL_GALLERY_FLOATING_MODAL_SIZE_PX = { width: 144, height: 81 };
```
dmceachernmsft, line 45: should this be 144 x 90? I guess thats not 16: 9 hey?

packages/react-components/src/components/VideoGallery/styles/VideoGalleryResponsiveVerticalGallery.styles.ts
```diff
@@ -13,24 +14,36 @@ import { LARGE_FLOATING_MODAL_SIZE_PX, SMALL_FLOATING_MODAL_SIZE_PX } from './Fl
...
 };
+
+/**
+ * @private
+ */
+export const VERTICAL_GALLERY_TILE_STYLE = {
+  minHeight: `${VERTICAL_GALLERY_TILE_SIZE_REM.minHeight}rem`,
+  minWidth: `${VERTICAL_GALLERY_TILE_SIZE_REM.width}rem`,
+  maxHeight: `${VERTICAL_GALLERY_TILE_SIZE_REM.maxHeight}rem`,
+  maxWidth: `${VERTICAL_GALLERY_TILE_SIZE_REM.width}rem`
```
dmceachernmsft, line 42: I think we can just define this as width and we don't need a min and max here

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -276,7 +290,9 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     /* @conditional-compile-remove(pinned-participants) */
     onUnpinParticipant: onUnpinParticipantHandler,
     /* @conditional-compile-remove(pinned-participants) */
-    remoteVideoTileMenuOptions = DEFAULT_REMOTE_VIDEO_TILE_MENU_OPTIONS
+    remoteVideoTileMenuOptions = DEFAULT_REMOTE_VIDEO_TILE_MENU_OPTIONS,
+    /* @conditional-compile-remove(vertical-gallery) */
+    overflowGalleryLayout = 'VerticalRight'
```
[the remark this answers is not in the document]
mgamis-msft: This should be 'HorizontalBottom'. Good catch!

packages/react-components/src/components/VideoGallery/styles/FloatingLocalVideo.styles.ts
mgamis-msft: Oh I just followed the size in Figma and it is the following currently:
image

packages/react-components/src/components/VideoGallery/styles/VideoGalleryResponsiveVerticalGallery.styles.ts
mgamis-msft: I remember there was an issue with that simplification from the past but let me test if I can make that simplification.

dmceachernmsft (approved):

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -3237,6 +3240,7 @@ export interface VideoGalleryProps {
     onRenderLocalVideoTile?: (localParticipant: VideoGalleryLocalParticipant) => JSX.Element;
     onRenderRemoteVideoTile?: (remoteParticipant: VideoGalleryRemoteParticipant) => JSX.Element;
     onUnpinParticipant?: (userId: string) => void;
+    overflowGalleryLayout?: OverflowGalleryLayout;
```
JamesBurnside, line 3243: nit this is already a VideoGallery so `Gallery` isn't needed in the naming here, to match the existing `layout` props should this instead be `overflowLayout: OverflowGalleryLayout`?

JamesBurnside (approved):

packages/communication-react/review/beta/communication-react.api.md
mgamis-msft: Really good point here. This new prop will pertain to how VideoGallery will layout overflow participants and overflowLayout would be a concise name. But we might want to keep the naming connection of inner overflow components HorizontalGallery, VerticalGallery etc to this new prop so overflowGalleryLayout maintains a stronger connections. We can always update it in beta if we start getting signals we are exporting the components if ARB picks on it.

PR #2780, opened 2023-02-24
[DRAFT] Adapter and Composite API for Video Effects

prabhjot-msft:
# What
Adapter and Composite API for Video Effects

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/communication-react/review/beta/communication-react.api.md
+++ b/packages/communication-react/review/beta/communication-react.api.md
@@ -188,6 +188,16 @@ export type AzureCommunicationChatAdapterArgs = {
 };
 
 // @public
+export interface BackgroundImage {
+    extension?: string;
+    name?: string;
+    url: string;
+}
+
+// @beta
+export type BackgroundUploadHandler = (image: File) => void;
+
+// @public
 export interface BaseCompositeProps<TIcons extends Record<string, JSX.Element>> {
     fluentTheme?: PartialTheme | Theme;
     icons?: TIcons;
@@ -285,11 +295,19 @@ export interface CallAdapterCallOperations {
     // @beta
     resumeCall(): Promise<void>;
     // @beta
+    selectUploadedBackground(backgroundImage: BackgroundImage): void;
+    // @beta
     sendDtmfTone(dtmfTone: DtmfTone_2): Promise<void>;
+    // @beta
+    setCustomBackgroundImages(backgroundImages: BackgroundImage[]): void;
     startCamera(options?: VideoStreamOptions): Promise<void>;
     startScreenShare(): Promise<void>;
+    // @beta
+    startVideoBackgroundEffect(videoBackgroundEffect: VideoBackgroundBlurEffect | VideoBackgroundReplacementEffect): Promise<void>;
     stopCamera(): Promise<void>;
     stopScreenShare(): Promise<void>;
+    // @beta
+    stopVideoBackgroundEffect(): Promise<void>;
     unmute(): Promise<void>;
 }
 
@@ -482,6 +500,7 @@ export type CallCompositeOptions = {
     onNetworkingTroubleShootingClick?: () => void;
     onEnvironmentInfoTroubleshootingClick?: () => void;
     remoteVideoTileMenu?: RemoteVideoTileMenuOptions;
+    videoBackgroundEffects?: VideoBackgroundEffectsOptions;
 };
 
 // @public
@@ -760,11 +779,15 @@ export interface CallWithChatAdapterManagement {
     // @beta
     resumeCall: () => Promise<void>;
     // @beta
+    selectUploadedBackground(backgroundImage: BackgroundImage): void;
+    // @beta
     sendDtmfTone: (dtmfTone: DtmfTone_2) => Promise<void>;
     sendMessage(content: string, options?: SendMessageOptions): Promise<void>;
     sendReadReceipt(chatMessageId: string): Promise<void>;
     sendTypingIndicator(): Promise<void>;
     setCamera(sourceInfo: VideoDeviceInfo, options?: VideoStreamOptions): Promise<void>;
+    // @beta
+    setCustomBackgroundImages(backgroundImages: BackgroundImage[]): void;
     setMicrophone(sourceInfo: AudioDeviceInfo): Promise<void>;
     setSpeaker(sourceInfo: AudioDeviceInfo): Promise<void>;
     startCall(participants: string[], options?: StartCallOptions): Call | undefined;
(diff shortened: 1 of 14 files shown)


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -187,6 +187,16 @@ export type AzureCommunicationChatAdapterArgs = {
     threadId: string;
 };
 
+// @public
+export interface BackgroundImage {
+    extension?: string;
```
JamesBurnside, line 192: do we need `extension`?

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -187,6 +187,16 @@ export type AzureCommunicationChatAdapterArgs = {
...
+    name?: string;
```
JamesBurnside, line 193: I assume this will be used for tooltips? Do we need name for anything else?

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -285,11 +295,19 @@ export interface CallAdapterCallOperations {
     // @beta
     resumeCall(): Promise<void>;
     // @beta
+    selectUploadedBackground(backgroundImage: BackgroundImage): void;
```
JamesBurnside, line 298: nit: would rename to `selectCustomBackground` and avoid the word uploaded

JamesBurnside: In fact for all we should add the prefix `video`
`selectCustomVideoBackground`

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -285,11 +295,19 @@ export interface CallAdapterCallOperations {
...
+    // @beta
     sendDtmfTone(dtmfTone: DtmfTone_2): Promise<void>;
+    // @beta
+    setCustomBackgroundImages(backgroundImages: BackgroundImage[]): void;
     startCamera(options?: VideoStreamOptions): Promise<void>;
     startScreenShare(): Promise<void>;
+    // @beta
+    startVideoBackgroundEffect(videoBackgroundEffect: VideoBackgroundBlurEffect | VideoBackgroundReplacementEffect): Promise<void>;
```
JamesBurnside, line 306: Can we use Calling's types here instead of defining our own?
image

packages/communication-react/review/beta/communication-react.api.md
prabhjot-msft: We don't need extension as of now, but if we are planning to restrict few extensions, we can use extension to filter the backgrounds before adding to the background set. Any thoughts on that?

packages/communication-react/review/beta/communication-react.api.md
prabhjot-msft: yeah, these are only for tooltips to show the name on hover.

packages/communication-react/review/beta/communication-react.api.md
prabhjot-msft: Sure, I will update by adding the prefix `video` to all the methods. But for `selectUploadedBackground` or `selectCustomVideoBackground`, this method is going to add new background to default set, select the background as active effect and then apply it if that is what design implies. Should we have a name that is bit more self explanatory?

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: This could be interpreted as filename (which is likely not what the desired tooltip name would be).
Rename to `tooltipName`/`displayName`/`friendlyName`/`nameHint`?

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: I would say remove this, if we get an image with an unsupported extension we'll simply have to throw an error (well we'll propagate the underlying error that the calling sdk will throw)

JamesBurnside: I like the idea of catching this at compile time -- e.g. we could strongly type this to be:
```suggestion
    extension?: 'png' | 'jpeg' | 'jpg' | etc...;
```

But I don't want Calling SDK to suddenly add support for new types at the service level and we have arbitrarily limited what can be accepted.
I think we have to follow Calling's API on this and simply take any url and try to render it

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: We'll need to wait until we get confirmation from UX design what should happen when a user uploads a custom image (does it auto apply, does it auto select and they need to hit a button called apply, or does it simply get added to the picker selection but not be selected or applied)

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -1001,6 +1029,12 @@ export const createTeamsCallAdapter: ({ userId, credential, locator, options }:
 // @beta
 export const createTeamsCallAdapterFromClient: (callClient: StatefulCallClient, callAgent: TeamsCallAgent, locator: CallAdapterLocator, options?: TeamsAdapterOptions | undefined) => Promise<TeamsCallAdapter>;
 
+// @public
+export interface CustomBackground {
```
JamesBurnside, line 1033: can we flatten this to 
```tsx
VideoBackgroundImage {
  id: string;
  url: string;
  name?: string;
}
```

Or do we need BackgroundImage to be nested in here?

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -1374,6 +1408,27 @@ export function _useFakeChatAdapters(args: _FakeChatAdapterArgs): _FakeChatAdapt
...
+export interface VideoBackgroundBlurEffect {
+    // (undocumented)
+    effectType: 'Blur';
+}
+
+// @beta
+export interface VideoBackgroundEffectsOptions {
+    backgroundImages?: BackgroundImage[];
+    backgroundUploadHandler?: BackgroundUploadHandler;
+    disableVideoEffects?: boolean;
```
JamesBurnside, line 1421: Open note as we can't answer until we get finialized UX designs.
We can't fully disable the video effects at the composite layer, we can only hide the UI buttons.
Depending on the final UX this should likely live on the CallControlsOptions interface
```tsx
<CallComposite options={{ callControls: { videoEffects: false } }} />
```

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -1374,6 +1408,27 @@ export function _useFakeChatAdapters(args: _FakeChatAdapterArgs): _FakeChatAdapt
...
+}
+
+// @public
+export interface VideoBackgroundReplacementEffect {
+    // (undocumented)
+    customBackground: CustomBackground;
```
JamesBurnside, line 1427: we should avoid the word `custom` here as calling never uses custom.
Instead just `background: VideoBackgroundImage`

PR #2783, opened 2023-02-27
Vertical overflow gallery layout unit tests

mgamis-msft:
# What

Added 1 screenshare and 1 pinned participants unit test for VideoGallery with vertical overflow gallery

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3151774

# How Tested

Ran unit tests locally. CI should pass.

# Process & policy checklist


packages/react-components/src/components/VideoGallery.test.tsx
```diff
@@ -198,7 +201,7 @@ describe('VideoGallery floating local video layout tests', () => {
     ).toBe(false);
   });
 
-  test('should render all remote video tiles in the grid', () => {
+  test('should render all remote video tiles in the grid and others to horizontal gallery', () => {
```
dmceachernmsft, line 204: Why was this test name updated? was there always going to be a Horizontal gallery in this test?

packages/react-components/src/components/VideoGallery.test.tsx
```diff
@@ -412,6 +415,92 @@ describe('VideoGallery pinned participants tests', () => {
   );
 });
 
+/* @conditional-compile-remove(vertical-gallery) */
+describe('VideoGallery with vertical overflow gallery tests', () => {
+  beforeAll(() => {
+    mockVideoGalleryInternalHelpers();
+  });
+
+  test('should render remote screenshare and render dominant speaking remote participants in vertical gallery', () => {
```
dmceachernmsft, line 424: Should we also have tests for 1 participant with video on and all others in VerticalGallery?

packages/react-components/src/components/VideoGallery.test.tsx
mgamis-msft: Just to be more specifc to a horizontal gallery since I have now added vertical gallery tests

packages/react-components/src/components/VideoGallery.test.tsx
mgamis-msft: Added.

carocao-msft (approved):

PR #2785, opened 2023-02-27
[Vertical Galleries] Update styles to allow tile vertical resizing

dmceachernmsft:
# What

Allows the remote video tiles to resize on their own up to the max of 144px creating a 1:1 aspect ratio
image
image

# Why

Allows for tiles to resize to show more room for the remote participant.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3151611
# How Tested

Ran locally in storybook, will add hermetic tests to check the client heights of the tiles to make sure they are the right size.

### DIFF
--- a/packages/react-components/src/components/styles/VerticalGallery.styles.ts
+++ b/packages/react-components/src/components/styles/VerticalGallery.styles.ts
@@ -13,9 +13,12 @@ export const VERTICAL_GALLERY_GAP = 0.5;
 /**
  * @private
  */
-export const childrenContainerStyle: IStyle = {
-  width: '100%',
-  gap: `${VERTICAL_GALLERY_GAP}rem`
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/styles/VerticalGallery.styles.ts
```diff
@@ -13,9 +13,12 @@ export const VERTICAL_GALLERY_GAP = 0.5;
 /**
  * @private
  */
-export const childrenContainerStyle: IStyle = {
-  width: '100%',
-  gap: `${VERTICAL_GALLERY_GAP}rem`
+export const childrenContainerStyle = (pageControlBarHeight: number): IStyle => {
```
mgamis-msft, line 16: nit: pageControlBarHeightRem because usually height and width is in px

mgamis-msft (approved):

prabhjot-msft (approved):

PR #2791, opened 2023-03-01
New playwright test project for desktop to set up overflow gallery tests

mgamis-msft:
# What

New playwright test project for desktop to set up overflow gallery tests

# Why

To set up call composite responsive behavior of switching between vertical/horizontal gallery

# How Tested

Relying on CI

# Process & policy checklist


JamesBurnside:
I wouldn't recommend a whole new test project for differing aspect ratio as this will massively increase the number of tests (and not relevant to chat etc.).
Can we instead have multiple tests that change the aspect ratio of the test runner specifically for gallery related tests?

mgamis-msft:
> I wouldn't recommend a whole new test project for differing aspect ratio as this will massively increase the number of tests (and not relevant to chat etc.). Can we instead have multiple tests that change the aspect ratio of the test runner specifically for gallery related tests?

Yes I am currently looking into that but haven't quite figured it out yet. But I thought I would kick off CI to get an idea of the change in our test times.

PR #2792, opened 2023-03-01
[Component and Story] Add background picker component

JamesBurnside:
# What
Add background picker that houses background effect options. Works same as a radio/choice group

# Why
To be used in the video effects feature

# How Tested

Checkout the storybook link in PR to play around with it

Screenshot:
image

### DIFF
--- a/packages/react-components/src/components/VideoEffects/VideoBackgroundEffectsPicker.tsx
+++ b/packages/react-components/src/components/VideoEffects/VideoBackgroundEffectsPicker.tsx
@@ -1,0 +1,121 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { IStyle, Label, mergeStyles, Stack } from '@fluentui/react';
+import React from 'react';
+import { chunk } from '../utils';
+import { _VideoEffectsItem, _VideoEffectsItemProps } from './VideoEffectsItem';
+
+/**
+ * Props for {@link _VideoBackgroundEffectsPicker}
+ * @internal
+ */
+export interface _VideoBackgroundEffectsPickerProps {
+  /**
+   * The options to display in the picker.
+   */
+  options: _VideoBackgroundEffectChoiceOption[];
+
+  /**
+   * The key of the current selected Video Background Effect.
+   * If you provide this, you must maintain selection state by observing onChange events and passing a new value in when changed.
+   */
+  selectedEffectKey?: string;
+
+  /**
+   * Callback to invoke when a Video Background Effect is selected.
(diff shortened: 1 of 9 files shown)


mgamis-msft (approved):

packages/react-components/src/components/VideoEffects/VideoBackgroundEffectsPicker.tsx
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * The options to display in the picker.
+   */
+  options: _VideoBackgroundEffectChoiceOption[];
+
+  /**
+   * The key of the current selected Video Background Effect.
+   * If you provide this, you must maintain selection state by observing onChange events and passing a new value in when changed.
+   */
+  selectedEffectKey?: string;
```
mgamis-msft, line 23: Nice controlled component concept such that if this is not defined the component will manage the state but if defined the user must manage the state through the onChange event.

packages/react-components/src/components/VideoEffects/VideoBackgroundEffectsPicker.tsx
```diff
@@ -0,0 +1,121 @@
...
+ * This functions similar to a radio group of buttons, where the user can select one of the options.
+ *
+ * @internal
+ */
+export const _VideoBackgroundEffectsPicker = (props: _VideoBackgroundEffectsPickerProps): JSX.Element => {
+  const [componentControlledSelectedEffectKey, setComponentControlledSelectedEffectKey] = React.useState<
+    string | undefined
+  >(props.selectedEffectKey);
+
+  const selectedEffect = props.selectedEffectKey ?? componentControlledSelectedEffectKey;
```
mgamis-msft, line 88: If the selectedEffect provided is not present/removed should we console warn like when a pinned participant id is not one of the remote participants as shown [here](https://github.com/Azure/communication-ui-library/blob/5e4689854bd10f701b18a897607e1c5204454358/packages/react-components/src/components/VideoGallery.tsx#L337)?

JamesBurnside: This is an excellent comment! Added. This also pointed me to this: https://github.com/microsoft/fluentui/blob/306fe5e1e38b36e84daf48a23817704416be2bb4/packages/utilities/src/warn/warnControlledUsage.ts#L31. Turns out fluent exports a component for warning against bad controlled component use! It has other warnings too in `useWarning` and only logs in development mode!

prabhjot-msft (approved):

PR #2810, opened 2023-03-10
[VerticalGallery bugfix] Overflow participants showing video streams on different pages

dmceachernmsft:
# What

image
image

# Why

Allows video participants on other pages to have their streams rendered

# How Tested

Ran locally in samples and on mobile device.

### DIFF
--- a/packages/react-components/src/components/VideoGallery/Layout.ts
+++ b/packages/react-components/src/components/VideoGallery/Layout.ts
@@ -30,7 +30,7 @@ export interface LayoutProps {
    * Maximum number of participant remote video streams that is rendered.
    * @defaultValue 4
    */
-  maxRemoteVideoStreams?: number;
+  maxRemoteVideoStreams: number;
   /**
    * Width of parent element
    */
--- a/change-beta/@azure-communication-react-2145cbc1-9060-4898-b40d-321f83de5a68.json
+++ b/change-beta/@azure-communication-react-2145cbc1-9060-4898-b40d-321f83de5a68.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update overflow gallery to render video tiles on multiple pages if video participants present.",
+  "packageName": "@azure/communication-react",
+  "email": "94866715+dmceachernmsft@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@azure-communication-react-2145cbc1-9060-4898-b40d-321f83de5a68.json
+++ b/change/@azure-communication-react-2145cbc1-9060-4898-b40d-321f83de5a68.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update overflow gallery to render video tiles on multiple pages if video participants present.",
+  "packageName": "@azure/communication-react",
+  "email": "94866715+dmceachernmsft@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/react-components/src/components/HorizontalGallery.tsx
+++ b/packages/react-components/src/components/HorizontalGallery.tsx
@@ -4,7 +4,7 @@
 import { DefaultButton, Icon, IStyle, Stack, mergeStyles } from '@fluentui/react';
 import React, { useMemo, useState } from 'react';
 import { useTheme } from '../theming';
-import { BaseCustomStyles } from '../types';
+import { BaseCustomStyles, VideoGalleryRemoteParticipant } from '../types';
 import { rootStyle, childrenContainerStyle, leftRightButtonStyles } from './styles/HorizontalGallery.styles';
 import { useIdentifiers } from '../identifiers';
 import { bucketize } from './utils/overFlowGalleriesUtils';
@@ -31,7 +31,9 @@ export interface HorizontalGalleryStyles extends BaseCustomStyles {
  * {@link HorizontalGallery} Component Props.
  */
 export interface HorizontalGalleryProps {
-  children: React.ReactNode;
+  galleryParticipants: VideoGalleryRemoteParticipant[];
+  onRenderRemoteParticipant: (participant: VideoGalleryRemoteParticipant, isVideoParticipant?: boolean) => JSX.Element;
+  maxRemoteVideoStreams?: number;
   /**
    * Styles for HorizontalGallery
    */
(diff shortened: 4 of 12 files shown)


packages/react-components/src/components/VideoGallery/Layout.ts
```diff
@@ -30,7 +30,7 @@ export interface LayoutProps {
    * Maximum number of participant remote video streams that is rendered.
    * @defaultValue 4
    */
-  maxRemoteVideoStreams?: number;
+  maxRemoteVideoStreams: number;
```
PorterNan, line 33: Why do we make this required? looks like we have default value to 4 when it is undefined?

PorterNan:
It took me a bit while to remember the discussion we had for Layout + Gallery design.

For the bug - Looks like we do the onRender earlier than the pagination, which means it will only render first 4 elements at the first page, for whatever page you are at.

For the fix - we push down the responsibility of onRender to vertical gallery/horizontal gallery, which they now which page they are at and can call onRender accordingly.

It looks to me this solution will work perfectly overall, the only concern here is we are reverting the original purposes of Layout component - the layout component should be lifecycle/render independently, so it can even handle some customized component from our customer (although we are not exposing this for now, but there is possibilities). If we couple layout component with onRender+remotePariticpant, there is not customization available for it anymore

I am thinking of another work around of this, in Layout component we could have, onCurrentDisplayIndexesChanged: (indexes: number[]), so it can pass out the indexes which is current we are showing on the screen, then the rendering responsibility will be still on its parent, for Example 
```
const [displayIndexes, setDisplayIndexes] = useState<number[]>();

const horizontalGalleryTiles = horizontalGalleryParticipants.map((p, index) => {
    return onRenderRemoteParticipant(
      p,
      maxRemoteVideoStreams && maxRemoteVideoStreams >= 0
        ? p.videoStream?.isAvailable && 
            displayIndexes.includes(index) &&  // Check if video is displayed by the layout component 
            activeVideoStreams++ < maxRemoteVideoStreams
        : p.videoStream?.isAvailable
    );
 });

<HorizontalGallery 
  onCurrentDisplayIndexesChanged = {setDisplayIndexs}
/>
```

packages/react-components/src/components/VideoGallery/Layout.ts
dmceachernmsft: since it will be always defined in our internal implementation of the video gallery this will always be true. this is me thinking more about what happens should we export these components externally so contoso can build their own layouts using the galleries. this should be required so that we don't hit errors off the calling SDK.

dmceachernmsft:
> It took me a bit while to remember the discussion we had for Layout + Gallery design.
> 
> For the bug - Looks like we do the onRender earlier than the pagination, which means it will only render first 4 elements at the first page, for whatever page you are at.
> 
> For the fix - we push down the responsibility of onRender to vertical gallery/horizontal gallery, which they now which page they are at and can call onRender accordingly.
> 
> It looks to me this solution will work perfectly overall, the only concern here is we are reverting the original purposes of Layout component - the layout component should be lifecycle/render independently, so it can even handle some customized component from our customer (although we are not exposing this for now, but there is possibilities). If we couple layout component with onRender+remotePariticpant, there is not customization available for it anymore
> 
> I am thinking of another work around of this, in Layout component we could have, onCurrentDisplayIndexesChanged: (indexes: number[]), so it can pass out the indexes which is current we are showing on the screen, then the rendering responsibility will be still on its parent, for example
> 
> ```
> const [displayIndexes, setDisplayIndexes] = useState<number[]>();
> 
> const horizontalGalleryTiles = horizontalGalleryParticipants.map((p, index) => {
>     return onRenderRemoteParticipant(
>       p,
>       maxRemoteVideoStreams && maxRemoteVideoStreams >= 0
>         ? p.videoStream?.isAvailable && 
>             displayIndexes.includes(index) &&  // Check if video is displayed by the layout component 
>             activeVideoStreams++ < maxRemoteVideoStreams
>         : p.videoStream?.isAvailable
>     );
>  });
> 
> <HorizontalGallery 
>   onCurrentDisplayIndexesChanged = {setDisplayIndexs}
> />
> ```

this would work for sure. is there a concern about this getting passed down two layers though? and what about the case where we want to maybe export these components externally as well so contoso can build their own experience with them? passing a react setState here would block us from doing that no?

PorterNan:
> > It took me a bit while to remember the discussion we had for Layout + Gallery design.
> > For the bug - Looks like we do the onRender earlier than the pagination, which means it will only render first 4 elements at the first page, for whatever page you are at.
> > For the fix - we push down the responsibility of onRender to vertical gallery/horizontal gallery, which they now which page they are at and can call onRender accordingly.
> > It looks to me this solution will work perfectly overall, the only concern here is we are reverting the original purposes of Layout component - the layout component should be lifecycle/render independently, so it can even handle some customized component from our customer (although we are not exposing this for now, but there is possibilities). If we couple layout component with onRender+remotePariticpant, there is not customization available for it anymore
> > I am thinking of another work around of this, in Layout component we could have, onCurrentDisplayIndexesChanged: (indexes: number[]), so it can pass out the indexes which is current we are showing on the screen, then the rendering responsibility will be still on its parent, for example
> > ```
> > const [displayIndexes, setDisplayIndexes] = useState<number[]>();
> > 
> > const horizontalGalleryTiles = horizontalGalleryParticipants.map((p, index) => {
> >     return onRenderRemoteParticipant(
> >       p,
> >       maxRemoteVideoStreams && maxRemoteVideoStreams >= 0
> >         ? p.videoStream?.isAvailable && 
> >             displayIndexes.includes(index) &&  // Check if video is displayed by the layout component 
> >             activeVideoStreams++ < maxRemoteVideoStreams
> >         : p.videoStream?.isAvailable
> >     );
> >  });
> > 
> > <HorizontalGallery 
> >   onCurrentDisplayIndexesChanged = {setDisplayIndexs}
> > />
> > ```
> 
> this would work for sure. is there a concern about this getting passed down two layers though? and what about the case where we want to maybe export these components externally as well so contoso can build their own experience with them? passing a react setState here would block us from doing that no?

Accodrding to previous discussions, these layout component should ideally accept any react node, which means developers can even put a normal html element with a YouTube video inside - but binding it with onRender+remotePariticpant would limit this flexibility, so for anyone who doesn't need a deep customization, current PR should work perfectly, but the drawback is it means people have to use it with our our remotePariticpant with no exceptions

packages/react-components/src/components/VideoGallery/Layout.ts
PorterNan: Yep, I got the point that maxRemoteVideoStreams must have a value - but we could always do a if else like

const maxStreams = maxRemoteVideoStreams ?? 4;

Then use value instead

The idea is we should minimize props needed for a component to run, especially for some tech details most people don't really care a lot. Either make it default to 4(more lean to ACS) or infinite (more lean to generic component) should work well

dmceachernmsft:
Better solution with indexing coming!

PR #2812, opened 2023-03-13
Correct MessageThread API definition and underlying logic

emlynmac:
# What
The `MessageThread` component has an `onMessageSend` property, which is documented as passing the message ID.
This is not the case, it is passed through to the underlying component and the parameter is the message body to send

# Why
Sending the message needs the body to send, the API should behave in the same way as all the other onSendMessage implementations and should behave as documented.

# How Tested

# Process & policy checklist


PorterNan (approved):
Thanks for catching this issue which was hiding that for 1+ year!

carocao-msft (approved):

PR #2817, opened 2023-03-15
Add stop caption function to stateful

PorterNan:
# What

Add stop caption function to stateful, and update centralized state
# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/calling-stateful-client/src/CallDeclarativeCommon.ts
+++ b/packages/calling-stateful-client/src/CallDeclarativeCommon.ts
@@ -98,6 +98,11 @@ export abstract class ProxyCallCommon implements ProxyHandler<CallCommon> {
                 this._context.setIsCaptionActive(target.id, true);
                 return ret;
               },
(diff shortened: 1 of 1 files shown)


packages/calling-stateful-client/src/CallDeclarativeCommon.ts
```diff
@@ -98,6 +98,11 @@ export abstract class ProxyCallCommon implements ProxyHandler<CallCommon> {
                 this._context.setIsCaptionActive(target.id, true);
                 return ret;
               },
+              stopCaptions: async (...args: Parameters<TeamsCaptionsCallFeature['startCaptions']>) => {
+                const ret = await captionsFeature.startCaptions(...args);
```
carocao-msft, line 102: Should this line be stopCaptions?

PorterNan: Nice catch! Actually stopCaptions doesn't have any parameter

PR #2818, opened 2023-03-15
[VerticalGallery Bugfix] Video tile rendering controlled by index

dmceachernmsft:
# What

Create indexing array to determine which children are currently in view so the gallery layouts can determine when to render overflow video participants.
# Why

allows more users to see more video participants on each page of the vertical and horizontal gallery

# How Tested

Locally validated the functionality.

### DIFF
--- a/packages/react-components/src/components/HorizontalGallery.tsx
+++ b/packages/react-components/src/components/HorizontalGallery.tsx
@@ -2,7 +2,7 @@
 // Licensed under the MIT license.
 
 import { DefaultButton, Icon, IStyle, Stack, mergeStyles } from '@fluentui/react';
-import React, { useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { useTheme } from '../theming';
 import { BaseCustomStyles } from '../types';
 import { rootStyle, childrenContainerStyle, leftRightButtonStyles } from './styles/HorizontalGallery.styles';
@@ -41,6 +41,10 @@ export interface HorizontalGalleryProps {
    * @defaultValue 5
    */
   childrenPerPage?: number;
+  /**
+   * helper function to choose which tiles to give video to.
+   */
+  setTilesToRender?: (indexes: number[]) => void;
 }
 
 /**
@@ -49,7 +53,7 @@ export interface HorizontalGalleryProps {
  * @returns
  */
 export const HorizontalGallery = (props: HorizontalGalleryProps): JSX.Element => {
-  const { children, childrenPerPage = DEFAULT_CHILDREN_PER_PAGE, styles } = props;
+  const { children, childrenPerPage = DEFAULT_CHILDREN_PER_PAGE, styles, setTilesToRender } = props;
 
   const ids = useIdentifiers();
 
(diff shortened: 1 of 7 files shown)


packages/react-components/src/components/HorizontalGallery.tsx
```diff
@@ -41,6 +41,10 @@ export interface HorizontalGalleryProps {
    * @defaultValue 5
    */
   childrenPerPage?: number;
+  /**
+   * helper function to choose which tiles to give video to.
+   */
+  setTilesToRender?: (indexes: number[]) => void;
```
PorterNan, line 47: This is a callback, so a callback style name would be easier to understand, like onRenderingTile, or onDisplayTile (hmm, can't find a nice name, but it should be start with "on")

packages/react-components/src/components/HorizontalGallery.tsx
```diff
@@ -58,15 +62,20 @@ export const HorizontalGallery = (props: HorizontalGalleryProps): JSX.Element =>
...
   }, [children, childrenPerPage]);
 
-  // If children per page is 0 or less return empty element
-  if (childrenPerPage <= 0) {
-    return <></>;
-  }
-
   const firstIndexOfCurrentPage = page * childrenPerPage;
   const clippedPage = firstIndexOfCurrentPage < numberOfChildren - 1 ? page : lastPage;
   const childrenOnCurrentPage = paginatedChildren[clippedPage];
```
PorterNan, line 81: This can be:
```
const childrenOnCurrentPage = useMemo(() => indexesArrays[clippedPage].map(index => {
  return children[index];
}), [ **dependency array here** ])
```

So even we change some lines of code in the future, those children nodes will be generated strictly followed what is in the array

PorterNan: And don't forget to remove lines from 75-78

packages/react-components/src/components/HorizontalGallery.tsx
```diff
@@ -58,15 +62,20 @@ export const HorizontalGallery = (props: HorizontalGalleryProps): JSX.Element =>
   const numberOfChildren = React.Children.count(children);
   const lastPage = Math.ceil(numberOfChildren / childrenPerPage) - 1;
 
+  const indexesArray: number[][] = useMemo(() => {
```
PorterNan, line 65: Do we need this to be 2d array? I guess we only need the array from the current page one?
So return bucketize([...Array(numberOfChildren).keys()], childrenPerPage)[page]

packages/react-components/src/components/HorizontalGallery.tsx
```diff
@@ -58,15 +62,20 @@ export const HorizontalGallery = (props: HorizontalGalleryProps): JSX.Element =>
...
+    return bucketize([...Array(numberOfChildren).keys()], childrenPerPage);
+  }, [numberOfChildren, childrenPerPage]);
+
+  useEffect(() => {
+    if (setTilesToRender && indexesArray) {
+      setTilesToRender(indexesArray[page]);
```
PorterNan, line 71: I guess any index access of the array should be clippedPage instead of page?

packages/react-components/src/components/VerticalGallery.tsx
```diff
@@ -102,10 +104,21 @@ export const VerticalGallery = (props: VerticalGalleryProps): JSX.Element => {
...
+    return bucketize([...Array(numberOfChildren).keys()], childrenPerPage);
+  }, [numberOfChildren, childrenPerPage]);
+
+  useEffect(() => {
+    if (setTilesToRender && indexesArray) {
+      setTilesToRender(indexesArray[page - 1]);
+    }
+  }, [indexesArray, setTilesToRender, page]);
 
   const paginatedChildren: React.ReactNode[][] = useMemo(() => {
```
PorterNan, line 119: same here, generate paginatedChildren from indexesArray instead

PorterNan: Logic in different Gallery components look similar - extract these logics into a common util function would be nice - but fully depending on you

PorterNan (approved):

mgamis-msft (approved):

PR #2820, opened 2023-03-16
[Do not merge]Captions component in composite 

carocao-msft:
# What
Internal storybook page showing captions component in composite

# Why
This will be a temporary internal story for Alex P to play around with and finalize design 

# How Tested
Screenshot 2023-03-15 at 12 29 42 PM
Storybook 

# Process & policy checklist


edwardlee-msft (approved):

PorterNan (changes requested):
Let's use this PR to generate the storybook link(which is already there), and not to check this into main branch - the composite of caption is a temporary prototype which should live outside of main branch

PR #2821, opened 2023-03-17
[Chat] [Feature] Data Loss Prevention (DLP) BlockedMessage type

jimchou-dev:
# What

image

- Introduced a new message type BlockedMessage for policy violation
- Added `policyViolation` property for ChatMessageWithStatus (that will be introduced by SDK ChatMessage in the future)
- Localization support for default blocked message
- Able to customize text for the BlockedMessage
- Able to customize styling for the BlockedMessage
- Able to use custom renderer to replace BlockedMessage with JSX.Element

image

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3055010

# How Tested

Using ACS chat to join a Teams meeting and send an connection string that will be policy violated

# Process & policy checklist


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -360,14 +373,107 @@ const memoizeAllMessages = memoizeFnAll(
       disableEditing
     };
 
+    /* @conditional-compile-remove(dlp) */
+    // Same logic as switch statement, if statement for conditional compile
+    if (message.messageType === 'blocked') {
```
jimchou-dev, line 378: conditional-compile-remove  doesn't seem to work on `switch` cases, so breaking out to `if` statement. Have same logic as `case 'chat':`

common/config/babel/.babelrc.js
```diff
@@ -48,7 +48,9 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         // feature for blurred background and replace background effects
         'video-background-effects',
         // feature for vertical gallery layouts in VideoGallery
-        'vertical-gallery'
+        'vertical-gallery',
+        // Data loss prevention feature
+        'dlp'
```
JoshuaLai, line 53: nitpick: Looks like the start of the list was kept in alphabetical order...and descends into not. Might be worth at least putting your `dlp` in alphabetical order

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -63,6 +69,32 @@ const extractAttachedFilesMetadata = (metadata: Record<string, string>): FileMet
...
+    status: !isLargeGroup && message.status === 'delivered' && isSeen ? 'seen' : message.status,
+    senderDisplayName: message.senderDisplayName,
+    senderId: messageSenderId,
+    messageId: message.id,
+    clientMessageId: message.clientMessageId,
+    editedOn: message.editedOn,
+    deletedOn: message.deletedOn,
+    mine: messageSenderId === userId,
+    metadata: message.metadata,
+    link: 'https://go.microsoft.com/fwlink/?LinkId=2132837'
```
JoshuaLai, line 94: Question: Does this link take locale into consideration? Is it the same as our privacy violation links where it will automagically change languages?

packages/react-components/src/types/ChatMessage.ts
```diff
@@ -119,6 +123,36 @@ export interface ContentSystemMessage extends SystemMessageCommon {
   content: string;
 }
 
+/* @conditional-compile-remove(dlp) */
+/**
+ * Content blocked message type.
+ *
+ * Content blocked messages will rendered default value, but applications can provide custom strings and icon to renderers.
+ *
+ * @public
```
JoshuaLai, line 132: Should this is marked with `@beta` instead?

common/config/babel/.babelrc.js
jimchou-dev: I think it used to be in alphabetical order, but it ended not being the case so just adding at the end, will change the order when i make another commit

common/config/babel/.babelrc.js
```diff
@@ -24,6 +24,8 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         'demo',
         // dialpad
         'dialpad',
+        // Data loss prevention feature
+        'dlp',
```
PorterNan, line 28: maybe use the full name like "data-loss-prevention"? It would be easier for someone who never knows this to understand the logic when they see conditional compilation

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -225,5 +257,9 @@ const messagesWithContentOrFileSharingMetadata = (message: ChatMessageWithStatus
   if (message.metadata?.['fileSharingMetadata']) {
     return true;
   }
+  /* @conditional-compile-remove(dlp) */
+  if (message.policyViolation) {
```
PorterNan, line 261: maybe rename to isMessageValidToRender

PorterNan: The whole function name*

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -225,5 +257,9 @@ const messagesWithContentOrFileSharingMetadata = (message: ChatMessageWithStatus
   if (message.metadata?.['fileSharingMetadata']) {
```
PorterNan, line 257: Also a fun idea to explore - currently we are not showing deleted message (if you look at deletedOn filter on the top of this function) - is it possible to make DLP message as a special type deleted message(we don't have deleted message type as well), or is there a common path/pattern we can reuse DLP code for deleted message?

PorterNan: Which means we can set messageType = 'deleted' but add a flag to say it is for dlp reason

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -63,6 +69,32 @@ const extractAttachedFilesMetadata = (metadata: Record<string, string>): FileMet
...
+    contentType: sanitizedMessageContentType(message.type),
+    status: !isLargeGroup && message.status === 'delivered' && isSeen ? 'seen' : message.status,
+    senderDisplayName: message.senderDisplayName,
+    senderId: messageSenderId,
+    messageId: message.id,
+    clientMessageId: message.clientMessageId,
+    editedOn: message.editedOn,
+    deletedOn: message.deletedOn,
+    mine: messageSenderId === userId,
+    metadata: message.metadata,
```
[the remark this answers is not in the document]
jimchou-dev: DLP message is just like a regular message, it can also be deleted (so it will completely not be shown and not just the content is removed). But will remove the metadata and contentType

packages/chat-component-bindings/src/messageThreadSelector.ts
jimchou-dev: Interesting idea, but technically DLP message is still a message, and it can be deleted and based on the current logic it will be completely not shown when someone delete's their DLP message. Just like a regular ChatMessage or edited ChatMessage they can be deleted and with current logic if `onDelete` is not undefined then it will be filtered out.

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -733,7 +849,14 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
   } = props;
   const onRenderFileDownloads = onRenderFileDownloadsTrampoline(props);
 
-  const [messages, setMessages] = useState<(ChatMessage | SystemMessage | CustomMessage)[]>([]);
+  const [messages, setMessages] = useState<
+    (
+      | ChatMessage
```
PorterNan, line 854: Add a new type, to merge these all together
type Message = ChatMessage | SystemMessage | CustomMessage | /* @conditional-compile-remove(data-loss-prevention) */ BlockedMessage

(maybe find a better name)

packages/react-components/src/types/ChatMessage.ts
```diff
@@ -119,6 +123,31 @@ export interface ContentSystemMessage extends SystemMessageCommon {
...
+ *
+ * @beta
+ */
+export interface BlockedMessage extends MessageCommon {
+  messageType: 'blocked';
+
+  content?: string | false;
+  linkText?: string;
+  link?: string;
+  editedOn?: Date;
```
PorterNan, line 140: Remove unnecessary properties as mentioned above

packages/react-components/review/beta/react-components.api.md
```diff
@@ -69,6 +69,36 @@ export interface BaseCustomStyles {
...
+export interface BlockedMessage extends MessageCommon {
+    // (undocumented)
+    attached?: MessageAttachedStatus;
+    // (undocumented)
+    clientMessageId?: string;
+    // (undocumented)
+    content?: string | false;
+    // (undocumented)
+    contentType: MessageContentType;
+    // (undocumented)
```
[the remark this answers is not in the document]
jimchou-dev: clientMessageId and editedOn can be removed, but the BlockedMessage may be deleted so will leave the DeletedOn

packages/chat-component-bindings/src/utils/updateMessagesWithAttached.ts
```diff
@@ -11,36 +11,46 @@ export const updateMessagesWithAttached = (chatMessagesWithStatus: Message[]): v
...
-     * Attached === true means it is within a group of messages in the current order
-     * Attached === top/bottom means it is on the top/bottom boundary
-     * Attached === false means it is just a single message
-     * A group of messages: continuous messages that belong to the same sender and not intercepted by other senders.
-     */
-    let attached: boolean | MessageAttachedStatus = false;
-    const previousMessage = index > 0 ? messages[index - 1] : undefined;
-    const nextMessage = index === messages.length - 1 ? undefined : messages[index + 1];
+    if (
+      message.messageType === 'chat' ||
```
emlynmac, line 15: Does this || operator also need to be conditionally compiled out?

packages/react-components/src/components/BlockedMessage.tsx
```diff
@@ -0,0 +1,91 @@
...
+   * @beta
+   */
+  onDisplayDateTimeString?: (messageDate: Date) => string;
+};
+
+/* @conditional-compile-remove(data-loss-prevention) */
+/**
+ * @private
+ */
+export const BlockedMessageContent = (props: BlockedMessageProps): JSX.Element => {
```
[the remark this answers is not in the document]
jimchou-dev: Do you mean something like this that will be used by MessageContentAsRichTextHTML, MessageContentAsText, and BlockedMessageContent to pass in their children?
image

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -360,14 +378,107 @@ const memoizeAllMessages = memoizeFnAll(
       disableEditing
     };
 
+    /* @conditional-compile-remove(data-loss-prevention) */
```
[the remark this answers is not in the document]
jimchou-dev: The logic IS the same, the only reason is because condition-compile-remove does not work for switch case, when merging to stable will share the same switch case with slight logic change

packages/chat-component-bindings/src/utils/updateMessagesWithAttached.ts
jimchou-dev: The in-line @conditional-compile-remove in boolean logic needs to be after the operator || or &&

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -63,6 +69,28 @@ const extractAttachedFilesMetadata = (metadata: Record<string, string>): FileMet
...
+    messageType: 'blocked',
+    createdOn: message.createdOn,
+    content: message.content?.message,
+    status: !isLargeGroup && message.status === 'delivered' && isSeen ? 'seen' : message.status,
+    senderDisplayName: message.senderDisplayName,
+    senderId: messageSenderId,
+    messageId: message.id,
+    deletedOn: message.deletedOn,
+    mine: messageSenderId === userId,
+    link: 'https://go.microsoft.com/fwlink/?LinkId=2132837'
```
PorterNan, line 90: Move this string to packages\chat-component-bindings\src\utils\constants.ts

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -63,6 +69,28 @@ const extractAttachedFilesMetadata = (metadata: Record<string, string>): FileMet
...
+  message: ChatMessageWithStatus,
+  userId: string,
+  isSeen: boolean,
+  isLargeGroup: boolean
+): BlockedMessage => {
+  const messageSenderId = message.sender !== undefined ? toFlatCommunicationIdentifier(message.sender) : userId;
+  return {
+    messageType: 'blocked',
+    createdOn: message.createdOn,
+    content: message.content?.message,
```
PorterNan, line 83: This content is not the same as message content, maybe find a better name for this like:

warningMessage, alternativeText or anything that can help developer understand that is the message to tell you this is blocked but not the message which was deleted

packages/chat-component-bindings/src/messageThreadSelector.ts
PorterNan: Yep, this is the current logic, while the ideal logic is when someone deleted the message the message will be shown as "this message has been deleted". Then it is pretty much like DLP and we might be able to reuse similar code even same type structure - but again, this is just an exploration we could do, not required for this PR

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -204,6 +204,30 @@ export interface BaseCustomStyles {
     root?: IStyle;
 }
 
+// @beta
+export interface BlockedMessage extends MessageCommon {
+    // (undocumented)
+    attached?: MessageAttachedStatus;
+    // (undocumented)
+    content?: string | false;
```
PorterNan, line 212: Do we need false type here? Undefined should be good enough I guess

packages/react-components/src/types/ChatMessage.ts
```diff
@@ -119,6 +123,31 @@ export interface ContentSystemMessage extends SystemMessageCommon {
...
+  content?: string | false;
+  linkText?: string;
+  link?: string;
+  // editedOn?: Date;
+  deletedOn?: Date;
+  senderId?: string;
+  senderDisplayName?: string;
+  status?: MessageStatus;
+  attached?: MessageAttachedStatus;
+  mine?: boolean;
```
PorterNan, line 146: Remove these 2 comment lines

packages/react-components/src/types/ChatMessage.ts
```diff
@@ -119,6 +123,31 @@ export interface ContentSystemMessage extends SystemMessageCommon {
...
+ *
+ * @beta
+ */
+export interface BlockedMessage extends MessageCommon {
+  messageType: 'blocked';
+
+  content?: string | false;
+  linkText?: string;
+  link?: string;
+  // editedOn?: Date;
```
PorterNan, line 140: Same here, remove

packages/react-components/src/components/utils/getParticipantsWhoHaveReadMessage.ts
```diff
@@ -2,6 +2,9 @@
 // Licensed under the MIT license.
 import { ChatMessage, ReadReceiptsBySenderId } from '../..';
 
+/* @conditional-compile-remove(data-loss-prevention) */
+import { BlockedMessage } from '../..';
```
PorterNan, line 6: Use the specific file instead, especially when it is inside of a package, this will reduce some potential circular dependencies and reduce overall import burdens

packages/react-components/src/types/ChatMessage.ts
```diff
@@ -119,6 +123,31 @@ export interface ContentSystemMessage extends SystemMessageCommon {
...
+/**
+ * Content blocked message type.
+ *
+ * Content blocked messages will rendered default value, but applications can provide custom strings and icon to renderers.
+ *
+ * @beta
+ */
+export interface BlockedMessage extends MessageCommon {
+  messageType: 'blocked';
+
```
PorterNan, line 136: Remove empty line here

packages/react-components/src/components/MessageThread.tsx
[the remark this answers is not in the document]
PorterNan: Most likely we will only remove the beta tag then forget to do the cleanup for the code, so it is always to do it earlier than later

 Would be nice to extract this common part into an internal function, and yes, we could still come back later to make it neat, but let's try our best when it is possible

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -361,14 +374,107 @@ const memoizeAllMessages = memoizeFnAll(
...
     switch (message.messageType) {
+      // case 'blocked':
       case 'chat': {
         const myChatMessageStyle =
           message.status === 'failed'
             ? styles?.failedMyChatMessageContainer ?? styles?.myChatMessageContainer ?? FailedMyChatMessageContainer
             : styles?.myChatMessageContainer ?? defaultMyChatMessageContainer;
         const chatMessageStyle = styles?.chatMessageContainer ?? defaultChatMessageContainer(theme);
-        messageProps.messageContainerStyle = message.mine ? myChatMessageStyle : chatMessageStyle;
+        /* @conditional-compile-remove(data-loss-prevention) */
```
PorterNan, line 465: Why do we have a @conditional-compile-remove(data-loss-prevention)  inside of the 'chat' condition?

jimchou-dev: Because the conditional-compile-remove does not work for switch statement case, so extracting out the logic to a `if` before the switch-statement. (
By adding the conditional compile here so the `if` and `case chat` will almost be identical and can more easily be removed when shipping to stable.

packages/communication-react/review/beta/communication-react.api.md
jimchou-dev: By default if nothing is provided it will be 'undefined' and will follow the default behavior with default localization text. For it to not display the warningText/content, they will need to set it to false (eg. if they only want the linkText only but no warningText)

packages/react-components/src/types/ChatMessage.ts
jimchou-dev: was following ChatMessage, but will also remove the line there

packages/react-components/src/components/MessageThread.tsx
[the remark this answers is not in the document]
jimchou-dev: Will refactor out common code

packages/chat-component-bindings/src/messageThreadSelector.ts
jimchou-dev: Rename to warningText

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -235,7 +240,9 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
             // In doing so here we set the target of the flyout to be the message and
             // not the 3-dot menu button to position the flyout correctly.
             setChatMessageActionFlyoutTarget(messageRef);
-            props.onActionButtonClick(message, setMessageReadBy);
+            if (message.messageType === 'chat') {
```
PorterNan, line 243: What would happen if user click 3-dot menu for blocked message?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -361,6 +374,82 @@ const memoizeAllMessages = memoizeFnAll(
...
+            </Flex>
+          )
+        },
+        attached: message.attached,
+        key: _messageKey
+      };
+    };
+
+    /* @conditional-compile-remove(data-loss-prevention) */
+    // Similar logic as switch statement case 'chat', if statement for conditional compile
```
PorterNan, line 442: Add a note here that:
Remove when stabilize the feature

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
jimchou-dev: Good call out, added the check to disable menu if it's local user BlockedMessage.

JoshuaLai (approved):

PR #2823, opened 2023-03-20
[Vertical Gallery] Calculate children utils

dmceachernmsft:
# What

Move children per page functions to utils to enable mocking of them in tests due to DOM constraints.
# Why

Fixes performance issues and testing issues around using a indexes array to determine which tiles in overflow gallery can be rendered.

# How Tested

Ran tests and apps locally to see passing tests and performance results

### DIFF
--- a/packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
+++ b/packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
@@ -12,6 +12,7 @@ import { rootLayoutStyle } from './styles/DefaultLayout.styles';
 import { videoGalleryLayoutGap } from './styles/Layout.styles';
 import { useOrganizedParticipants } from './utils/videoGalleryLayoutUtils';
 import { OverflowGallery } from './OverflowGallery';
+import { DEFAULT_MAX_REMOTE_VIDEO_STREAMS } from '../VideoGallery';
 
 /**
  * Props for {@link DefaultLayout}.
@@ -34,7 +35,7 @@ export const DefaultLayout = (props: DefaultLayoutProps): JSX.Element => {
     screenShareComponent,
     onRenderRemoteParticipant,
     styles,
-    maxRemoteVideoStreams,
+    maxRemoteVideoStreams = DEFAULT_MAX_REMOTE_VIDEO_STREAMS,
     parentWidth,
     /* @conditional-compile-remove(vertical-gallery) */
     parentHeight,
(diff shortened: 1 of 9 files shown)


mgamis-msft (approved):

packages/react-components/src/components/VideoGallery/Layout.ts
```diff
@@ -30,7 +30,7 @@ export interface LayoutProps {
    * Maximum number of participant remote video streams that is rendered.
    * @defaultValue 4
    */
-  maxRemoteVideoStreams?: number;
+  maxRemoteVideoStreams: number;
```
PorterNan, line 33: Still the same question - can we make this optional, if no one provide this - then we use 4 in the component internal logic?

Like 

const maxStreams = maxRemoteVideoStreams ?? 4;

Then use maxStreams in the component

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
```diff
@@ -66,7 +66,10 @@ export const DefaultLayout = (props: DefaultLayoutProps): JSX.Element => {
     );
   });
 
-  const [indexesToRender, setIndexesToRender] = useState<number[]>([0]);
+  /** instantiate indexes available to render with indexes available that would be on first page */
+  const [indexesToRender, setIndexesToRender] = useState<number[]>([
+    ...Array(maxRemoteVideoStreams - activeVideoStreams).keys()
```
PorterNan, line 71: So this is the key to fix the bug, nice!

A small note here: For some interesting components which doesn't strictly follow the order of the array, we might re-render the initial tiles -> dispose them -> create new tiles, we need to take care of this case when those components are here

let's add this description into comment, so someone will handle this part well when developing "not follow order" component

packages/react-components/src/components/VideoGallery/Layout.ts
PorterNan: Yea, for internal component, this is less concern about it. When we are exporting this for external usage we might look back to api design and review it, should be good

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
dmceachernmsft: I dont think this will be affected by order no? this array is strictly telling the render function which which indexes to render, so the tiles that are present on the first page shouldn't matter and then should the page change this array will be updated.

dmceachernmsft: I dont think this will be affected by order no? this array is strictly telling the render function which which indexes to render, so the tiles that are present on the first page shouldn't matter and then should the page change this array will be updated.

packages/react-components/src/components/VideoGallery/Layout.ts
dmceachernmsft: would there be a use case to export the layout? maybe like as a interface?

PR #2843, opened 2023-03-27
Captions component in composite

carocao-msft:
# What
Captions component in call and callwithchat composite

# Why

# How Tested

Uploading demo.mov…

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
+++ b/packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
@@ -188,17 +188,14 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
                 />
               </Stack.Item>
             )}
-          <Stack.Item>
-            <CaptionsBanner />
-          </Stack.Item>
+
           <Stack horizontal grow>
             <Stack.Item grow style={callCompositeContainerFlex()}>
               <Stack.Item styles={callGalleryStyles} grow>
-                {props.onRenderGalleryContent && (
-                  <Stack verticalFill styles={mediaGalleryContainerStyles}>
-                    {props.onRenderGalleryContent()}
-                  </Stack>
-                )}
+                <Stack verticalFill styles={mediaGalleryContainerStyles}>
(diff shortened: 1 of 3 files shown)


packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -154,6 +156,8 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
     };
   }
 
+  const captionsBannerProps = usePropsFor(_CaptionsBanner);
```
PorterNan, line 159: Let's move all hook calls earlier to the top of the component closure

packages/react-composites/src/composites/common/CaptionsBanner.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+/** @private */
+export const CaptionsBanner = (): JSX.Element => {
+  const captionsBannerProps = usePropsFor(_CaptionsBanner);
+
+  return (
+    <>
+      {captionsBannerProps.captions.length > 0 && (
+        <Stack horizontalAlign="center">
+          <Stack.Item style={{ width: '50%' }}>
```
PorterNan, line 17: What does width:50% do here? it feels not that future proof to set width in a Stack.item (which is actually a flexbox), let's use something like flex-grow, flex-shrink or properties designed for flexbox

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -185,6 +190,15 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
                 />
               </Stack.Item>
             )}
+          <Stack.Item>
```
PorterNan, line 193: Use the common component instead

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
carocao-msft: fixed

PorterNan (approved):

PR #2844, opened 2023-03-27
Fixed some captionsBanner naming issues, improved styling and internal storybook

carocao-msft:
# What
Fixed some captionsBanner naming issues, improved styling and internal storybook

# Why
Make captions grid width dynamic
Make internal storybook page look prettier
Fixed build warnings 

# How Tested
Storybook

# Process & policy checklist


packages/react-components/src/components/CaptionsBanner.tsx
```diff
@@ -94,18 +69,22 @@ export const _CaptionsBanner = (props: _CaptionsBannerProps): JSX.Element => {
           ) : (
             <Persona {...personaOptions} />
           );
+          // if display name is too long, replace with ...
+          if (caption.displayName.length > 15) {
+            caption.displayName = `${caption.displayName.substring(0, 16)}...`;
```
edwardlee-msft, line 74: Should this account for the ... in the length as well?
Perhaps we need to substring(0,14) or substring(0,15) to account for the ellipsis

edwardlee-msft (approved):

packages/react-components/src/components/CaptionsBanner.tsx
carocao-msft: Yeah I'm keeping the length to be 15 charaters + ... according to the design

PorterNan, line 74: Can we use css trick to do this? we should not limit how many characters, everything should depend on max length of parent container: https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

Same amount of characters doesn't mean they have same length let's take an extreme example:

10 ‘l’:
llllllllll
10 'P':
PPPPPPPPPP

PorterNan (approved):

PR #2847, opened 2023-03-28
Control bar buttons should stay centered with regard to view port width

carocao-msft:
# What
Modified callwithchat control bar styling so buttons are centered with regard to view port width 

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3183443

# How Tested
callwithchat sample
tested with custom buttons 

https://user-images.githubusercontent.com/96077406/228696496-631274ae-d35a-4ba0-9ab5-ebcf3aad6990.mov

To avoid call control buttons and panel control buttons overlap, at certain width we cannot center call control buttons based on viewport width anymore
So using useRef here to calculate that moment and remove the code which centers the control buttons based on viewport width
Screenshot 2023-03-29 at 8 16 44 PM

Screenshot 2023-03-29 at 8 16 30 PM

This is also why in snapshot updates you see the control bar buttons not centered to the middle, because viewport width is too small

# Process & policy checklist


edwardlee-msft:
In this case shouldn't the buttons be collapsed into a more button?
image

Cause in your second picture I can see that there is a more button.


Perhaps the sample picture used with custom buttons is not an ideal example to portray this fix?

carocao-msft:
> In this case shouldn't the buttons be collapsed into a more button? image
> 
> Cause in your second picture I can see that there is a more button.
> 
> Perhaps the sample picture used with custom buttons is not an ideal example to portray this fix?

The first picture is how everything looks with custom buttons setup, the second one is without.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -155,137 +214,154 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
   );
 
   return (
-    <Stack horizontal className={mergeStyles(callControlsContainerStyles, controlBarContainerStyles)}>
-      <Stack.Item grow>
-        <CallAdapterProvider adapter={props.callAdapter}>
-          <Stack horizontalAlign="center">
-            {/*
+    <div ref={controlBarContainerRef}>
```
dmceachernmsft, line 217: nit: Maybe a different name for this reference? `controlBarSizeRef`?

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -155,137 +214,154 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
...
-                {isEnabled(options.screenShareButton) && (
-                  <ScreenShare
-                    option={options.screenShareButton}
-                    displayType={options.displayType}
-                    styles={screenShareButtonStyles}
-                    /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
-                    disabled={props.disableButtonsForHoldScreen || isDisabled(options.screenShareButton)}
-                  />
-                )}
+                <div ref={controlBarButtonsRef}>
```
dmceachernmsft, line 245: Continuing the nit above: should this be `controlBarContainerRef`?

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatControlBar.tsx
```diff
@@ -155,137 +214,154 @@ export const CallWithChatControlBar = (props: CallWithChatControlBarProps & Cont
...
+                    <EndCall displayType="compact" styles={endCallButtonStyles} />
+                  </ControlBar>
+                </div>
+              </Stack.Item>
+            </Stack>
+          </CallAdapterProvider>
+        </Stack.Item>
+        {!props.mobileView && (
+          <Stack.Item>
+            <div ref={panelsButtonsRef}>
```
dmceachernmsft, line 330: nit: maybe `sidepaneControlsRef`?

dmceachernmsft (approved):

prabhjot-msft (approved):

PR #2849, opened 2023-03-28
Support background effects in config screen

prabhjot-msft:
# What
1. New effects button in config screen
2. Side pane opens on config screen

Screenshot 2023-03-27 at 10 42 13 PM

Screenshot 2023-03-27 at 10 42 28 PM

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3193136
https://skype.visualstudio.com/SPOOL/_workitems/edit/3193138

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/styles/CallConfiguration.styles.ts
```diff
@@ -125,3 +127,25 @@ export const startCallButtonStyleMobile = mergeStyles({
   width: '100%',
   maxWidth: 'unset'
 });
+
+/* @conditional-compile-remove(video-background-effects) */
```
PorterNan, line 131: We don't have to CC it if that is not exported publicly, and it is encouraged to reduce CC comments if possible

PorterNan (approved):
There is a concern that we should keep main branch always working without any blank panel/not working UI part

So I would expect check this in after we have the side effect pane content merged in first

Approved, and look forward to your PR to merge side effect pane content

prabhjot-msft:
> There is a concern that we should keep main branch always working without any blank panel/not working UI part
> 
> So I would expect check this in after we have the side effect pane content merged in first
> 
> Approved, and look forward to your PR to merge side effect pane content

Thanks for bringing that up, I have added the content to the pane, here is the PR https://github.com/Azure/communication-ui-library/pull/2858/

packages/react-composites/src/composites/CallComposite/styles/CallConfiguration.styles.ts
prabhjot-msft: done

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -270,6 +291,22 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
                 {title}
                 {callDescription}
               </Stack.Item>
+              {
+                /* @conditional-compile-remove(video-background-effects) */
+                <DefaultButton
+                  iconProps={{ iconName: 'OptionsVideoBackgroundEffect', styles: { root: { lineHeight: 0 } } }}
```
mgamis-msft, line 297: store the styles in a styles file. This component will re-render needlessly if objects are not static or memoized.

prabhjot-msft: Looks like that styling is not needed, icon is rendering well without that as well, removed it.

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -270,6 +291,22 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
+                /* @conditional-compile-remove(video-background-effects) */
+                <DefaultButton
+                  iconProps={{ iconName: 'OptionsVideoBackgroundEffect' }}
+                  styles={effectsButtonStyles(theme)}
+                  onClick={() => {
+                    if (showVideoEffectsPane) {
+                      setShowVideoEffectsPane(false);
+                    } else {
+                      setShowVideoEffectsPane(true);
+                    }
```
mgamis-msft, line 304: Maybe can simplify this to just:
```
setVideoEffectsPane(!showVideoEffectsPane)
```

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -75,6 +85,15 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
     /* @conditional-compile-remove(call-readiness) */ onNetworkingTroubleShootingClick
   } = props;
 
+  /* @conditional-compile-remove(video-background-effects) */
+  const [showVideoEffectsPane, setVideoEffectsPane] = useState(false);
+  /* @conditional-compile-remove(video-background-effects) */
+  const setShowVideoEffectsPane = useCallback(
```
mgamis-msft, line 91: nit: I don't think you need this callback. You already have the useState setter.

mgamis-msft (approved):

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
prabhjot-msft: done

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
prabhjot-msft: done

PR #2850, opened 2023-03-28
Prevent video gallery overflow when side pane is open in CallComposite

mgamis-msft:
# What

Prevent video gallery overflow when side pane is open in CallComposite by setting min-width to 0 as suggested by the top answer here: https://stackoverflow.com/questions/36230944/prevent-flex-items-from-overflowing-a-container

<------------- Flex container ------------->
<----Video Gallery---><---Side Pane---->

Video Gallery is wrapped in a Stack with grow property set to true which allows overflow. When the width is narrow enough the VideoGallery uses the ScrollableHorizontalGallery causing overflow to the width of the VideoGallery.

# Why

Issue seen in Calling hero sample in this post:
https://teams.microsoft.com/l/message/19:5e150d250f724d1089c949305efdc8cc@thread.skype/1679940265587?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e9c1fc3-39df-4486-a26a-456d80e80f82&parentMessageId=1679940265587&teamName=Azure%20Communication%20Services&channelName=Web%20UI%20-%20Team&createdTime=1679940265587

# How Tested

Reproduced in Calling sample and fixed as shown here:
https://user-images.githubusercontent.com/79475487/228150080-0365d149-02dd-410d-83a4-e2ec5b1266c3.mp4

# Process & policy checklist

### DIFF
--- a/change-beta/@azure-communication-react-241d9d8a-edc9-473e-8bf4-dd5b13a034f1.json
+++ b/change-beta/@azure-communication-react-241d9d8a-edc9-473e-8bf4-dd5b13a034f1.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Prevent overflow of VideoGallery when side pane is open",
+  "packageName": "@azure/communication-react",
+  "email": "miguelgamis@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


alkwa-msft:
Is there any hermetic tests or anything we could build? Even just build later so we can validate this scenario?

dmceachernmsft (approved):

alkwa-msft (approved):

mgamis-msft:
> Is there any hermetic tests or anything we could build? Even just build later so we can validate this scenario?

Yes we can verify that the ResponsiveHorizontalGallery exists, resize window to decrease the VideoGallery width OR open the side pane, then verify that the ScrollableHorizontalGallery exists AND that the width of the composite does not exceed the window width.

PR #2854, opened 2023-03-28
[Chat][Feature] Introduce an options bag when creating adapter

JoshuaLai:
# What
Introduce a new options bag for credentials

# Why
The new ability to download file attachments from ACS requires a token credential to be saved and passed into via the headers. 

# How Tested

# Process & policy checklist


jimchou-dev (approved):
LGTM

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationFileUploadAdapter.ts
```diff
@@ -96,12 +98,12 @@ class FileUploadContext {
  */
 export class AzureCommunicationFileUploadAdapter implements FileUploadAdapter {
   private context: FileUploadContext;
-  private getAuthToken?: () => Promise<string>;
+  private credential?: CommunicationTokenCredential;
```
emlynmac, line 101: Do we need the whole `CommunicationTokenCredential` here, rather than just the `getAuthToken` closure?

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -427,6 +427,14 @@ const convertEventType = (type: string): ChatMessageType => {
   }
 };
 
+/**
+ * Option bag to include when creating AzureCommunicationChatAdapter.
+ * @public
+ */
+export type ChatAdapterOptions = {
+  credential?: CommunicationTokenCredential;
```
emlynmac, line 435: Do we want to lock to the ACS specific type, rather than just passing through the necessary function to get the authToken?

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -872,6 +872,11 @@ export interface CameraButtonStyles extends ControlBarButtonStyles {
 // @public
 export type ChatAdapter = ChatAdapterThreadManagement & AdapterState<ChatAdapterState> & Disposable & ChatAdapterSubscribers;
 
+// @public
+export type ChatAdapterOptions = {
```
emlynmac, line 876: Should this be on the public stable API, or only on the Beta for now?

JoshuaLai: Since the `createAzureCommunicationChatAdapterFromClient` is already public, this new type will also need to be public. 
Creating a new optional parameter after speaking with Nan should be ok.

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -427,6 +439,14 @@ const convertEventType = (type: string): ChatMessageType => {
   }
 };
 
+/**
+ * Option bag to include when creating AzureCommunicationChatAdapter.
```
emlynmac, line 443: "Configuration options for the ChatAdapter" ?

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -828,6 +828,11 @@ export type CallWithChatEvent = 'callError' | 'chatError' | 'callEnded' | 'isMut
 // @public
 export type ChatAdapter = ChatAdapterThreadManagement & AdapterState<ChatAdapterState> & Disposable & ChatAdapterSubscribers & FileUploadAdapter;
 
+// @public
```
PorterNan, line 831: @beta, and we might want to CC this together with iniline-image

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -828,6 +828,11 @@ export type CallWithChatEvent = 'callError' | 'chatError' | 'callEnded' | 'isMut
...
+export type ChatAdapterOptions = {
+    credential?: CommunicationTokenCredential;
```
PorterNan, line 833: So we are providing the credential object instead, is it because it is easier to be refreshed? (I remember there is a optional refresh token embedded in this object)

packages/react-composites/review/stable/react-composites.api.md
```diff
@@ -634,6 +634,11 @@ export type CallWithChatEvent = 'callError' | 'chatError' | 'callEnded' | 'isMut
 // @public
 export type ChatAdapter = ChatAdapterThreadManagement & AdapterState<ChatAdapterState> & Disposable & ChatAdapterSubscribers;
 
+// @public
```
PorterNan, line 637: @beta

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -117,13 +117,25 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
...
     this.chatClient = chatClient;
     this.chatThreadClient = chatThreadClient;
     this.context = new ChatContext(chatClient.getState(), chatThreadClient.threadId);
-    /* @conditional-compile-remove(file-sharing) */
-    this.fileUploadAdapter = new AzureCommunicationFileUploadAdapter(this.context);
+    const credential = options?.credential;
+    if (credential) {
+      /* @conditional-compile-remove(teams-inline-images) */
+      const getAuthToken: () => Promise<string> = async () => {
+        const accessToken = await credential.getToken();
```
PorterNan, line 129: Right! That is the fetch token function I remember

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -117,13 +117,25 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
...
+        return accessToken.token;
+      };
+      /* @conditional-compile-remove(teams-inline-images) */
+      this.fileUploadAdapter = new AzureCommunicationFileUploadAdapter(this.context, getAuthToken);
+    } else {
+      /* @conditional-compile-remove(file-sharing) */
```
PorterNan, line 135: We can do 
let getAuthToken = undefined;
getAuthToken: () => Promise<string> = async () => {}

this.fileUploadAdapter = new AzureCommunicationFileUploadAdapter(this.context, getAuthToken);

So we can save an if-else in CC and less code to clean up

packages/react-composites/src/composites/ChatComposite/index.ts
```diff
@@ -7,7 +7,7 @@ export {
   useAzureCommunicationChatAdapter
 } from './adapter/AzureCommunicationChatAdapter';
 export type { AzureCommunicationChatAdapterArgs } from './adapter/AzureCommunicationChatAdapter';
-
+export type { ChatAdapterOptions } from './adapter/AzureCommunicationChatAdapter';
```
PorterNan, line 10: CC this one

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -588,9 +609,10 @@ export const useAzureCommunicationChatAdapter = (
  */
 export const createAzureCommunicationChatAdapterFromClient = async (
   chatClient: StatefulChatClient,
-  chatThreadClient: ChatThreadClient
+  chatThreadClient: ChatThreadClient,
+  options?: ChatAdapterOptions
```
PorterNan, line 613: CC this one

packages/react-composites/review/beta/react-composites.api.md
JoshuaLai: Yes. There is already a built in way to retrieve a new token if Contoso uses `RefreshTokenCredential`.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1342,7 +1342,9 @@ export const createAzureCommunicationCallWithChatAdapterFromClients: ({ callClie
 export const createAzureCommunicationChatAdapter: ({ endpoint: endpointUrl, userId, displayName, credential, threadId }: AzureCommunicationChatAdapterArgs) => Promise<ChatAdapter>;
 
 // @public
-export const createAzureCommunicationChatAdapterFromClient: (chatClient: StatefulChatClient, chatThreadClient: ChatThreadClient) => Promise<ChatAdapter>;
+export const createAzureCommunicationChatAdapterFromClient: (chatClient: StatefulChatClient, chatThreadClient: ChatThreadClient, options?: {
```
JamesBurnside, line 1345: If the feature is beta we need to avoid this change in the stable API (once it's in a stable release we can't change the name, and this will be reviewed by ARB for any stable release so without the accompanying feature being released we shouldn't expose it).

I assume conditionally compiling the options isn't working? (cc @PorterNan). If not we should be able to create an overload with the new param and only export the overload in beta

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -797,6 +797,11 @@ export type CallWithChatEvent = 'callError' | 'chatError' | 'callEnded' | 'isMut
 // @public
 export type ChatAdapter = ChatAdapterThreadManagement & AdapterState<ChatAdapterState> & Disposable & ChatAdapterSubscribers & FileUploadAdapter;
 
+// @beta
+export type ChatAdapterOptions = {
+    credential?: CommunicationTokenCredential;
```
JamesBurnside, line 802: can we make the name less generic? We already use a credential to create the adapter/statefulchatclient (the ACS token credential), how does this credential differ?

packages/communication-react/review/stable/communication-react.api.md
JoshuaLai: We can create an overload for sure. Some of the feedback I got was to not create an overload cause this is an optional argument anyways and shouldn't be a breaking change. And on top of that this style follows similar to what the Call Adapter is already doing.

PorterNan: Is it working for CC? If not, maybe it is a good sign we should support that for CC, I've seen something similar for 5+ times

PorterNan: Here we go: 
https://github.com/Azure/communication-ui-library/pull/2880

JoshuaLai: Will wait for the above PR to go in before merging this PR.

packages/react-composites/review/beta/react-composites.api.md
JoshuaLai: Perhaps something like this:
```
export type ChatAdapterOptions = {
    teamsInterop?: {
    }
}
```

JoshuaLai: I don't know if I agree with this change, adding extra layer. Now it feels like we are adding an options bag inside an options bag. I worry it might just adding unnecessary complexity

JamesBurnside: Yeah, if we only have the one property lets keep flat 👍
Scope the name to something less generic to distinguish from the chat token credential and allow for other credentials if they're required; something like `teamsInteropFileDownloadCredential` ? (Can this be used for both file attachments and things like inline images?)

JoshuaLai: Right now only inline images (specifically images coming in from teams) needs this credential. Future-wise is a little unknown. I would imagine going forward if we need to fetch anything from the service outside of the scope of ChatSDK (which is very niche) we would need it.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1342,7 +1342,7 @@ export const createAzureCommunicationCallWithChatAdapterFromClients: ({ callClie
 export const createAzureCommunicationChatAdapter: ({ endpoint: endpointUrl, userId, displayName, credential, threadId }: AzureCommunicationChatAdapterArgs) => Promise<ChatAdapter>;
 
 // @public
-export const createAzureCommunicationChatAdapterFromClient: (chatClient: StatefulChatClient, chatThreadClient: ChatThreadClient) => Promise<ChatAdapter>;
+export function createAzureCommunicationChatAdapterFromClient(chatClient: StatefulChatClient, chatThreadClient: ChatThreadClient): Promise<ChatAdapter>;
```
JoshuaLai, line 1345: @JamesBurnside @PorterNan The change to the stable API looks like this

packages/react-composites/review/beta/react-composites.api.md
JamesBurnside: Lets leave this as is then, if it needs changed can be easily done at api review time 👍

packages/communication-react/review/stable/communication-react.api.md
JamesBurnside: LGTM! Thanks for making this change

PR #2856, opened 2023-03-29
Common call control for both callComposite and callWithChatComposite

PorterNan:
# What

1. Extract logic for CallComposite and CallWithChatComposite
2. Extract common options for CallComposite and CallWithChatComposite
3. Add a legacyCallControl options to enable previous callcontrol

TODO:
Add hermetic test for new call control experience

* What could be done even better for next iteration?

Extract chat button into a custom button, and put it back into call with chat composite, so no chat logic in the common composite
# Why

# How Tested

# Process & policy checklist


carocao-msft (approved):

samples/Calling/src/app/views/CallCompositeContainer.tsx
```diff
@@ -19,7 +19,10 @@ export const CallCompositeContainer = (props: CallCompositeContainerProps): JSX.
   const options: CallCompositeOptions = useMemo(
     () => ({
       onPermissionsTroubleshootingClick,
-      onNetworkingTroubleShootingClick
+      onNetworkingTroubleShootingClick,
+      callControls: {
+        legacyControlBarExperience: false
```
prabhjot-msft, line 24: Is this the option to get back to the old control bar?

prabhjot-msft (approved):
Looks good to me, great work Nan :)

samples/Calling/src/app/views/CallCompositeContainer.tsx
PorterNan: Yep, once this is true, the old control bar will be showing up

PR #2858, opened 2023-03-29
Add Blur and remove effects to VideoEffectsPane

prabhjot-msft:
# What
Add Blur and remove effects to VideoEffectsPane
Screenshot 2023-03-29 at 1 16 59 PM

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3204831

# How Tested

# Process & policy checklist


edwardlee-msft (approved):

packages/react-composites/src/composites/common/VideoEffectsPane.tsx
```diff
@@ -12,13 +20,63 @@ import { Panel } from '@fluentui/react';
 export const VideoEffectsPane = (props: {
   showVideoEffectsOptions: boolean;
   setshowVideoEffectsOptions: (showVideoEffectsOptions: boolean) => void;
+  adapter: CallAdapter | CommonCallAdapter;
 }): JSX.Element => {
-  return VideoEffectsPaneTrampoline(props.showVideoEffectsOptions, props.setshowVideoEffectsOptions);
+  const { showVideoEffectsOptions, setshowVideoEffectsOptions } = props;
+  /* @conditional-compile-remove(video-background-effects) */
```
mgamis-msft, line 26: Is this an internal component that will only be used for the video-background-effects feature? If yes, I think we don't need the conditional compilation statements.

packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -314,6 +314,26 @@ export interface CallCompositeStrings {
...
+  /* @conditional-compile-remove(video-background-effects) */
+  /**
+   * Tooltip text for the blur video background effect item
+   */
+  blurBackgroundTooltip?: string;
+  /* @conditional-compile-remove(video-background-effects) */
+  /**
+   * Label for the remove video background effect item
+   */
+  removeBackgroundEffect?: string;
```
mgamis-msft, line 331: is removeBackgroundEffectButtonLabel more descriptive?

prabhjot-msft: done

packages/react-composites/src/composites/common/VideoEffectsPane.tsx
mgamis-msft: But you are already not showing the VideoEffectsPane in stable by having a CC statement in CallWithChatComposite.tsx
image

mgamis-msft: I'll still approve to unblock :) but I still think the CC lines are unnecessary

mgamis-msft (approved):

packages/react-composites/src/composites/common/VideoEffectsPane.tsx
prabhjot-msft: Thanks Miguel, for pointing this out. I will merge this PR to get unblocked and have created a separate item to remove the CC from video effects pane https://skype.visualstudio.com/SPOOL/_workitems/edit/3207057

PR #2860, opened 2023-03-30
Update Captions UI

carocao-msft:
# What
Update Captions UI based on new design
Modify auto scroll logic, do not auto scroll to bottom unless user is already at bottom when new captions come in

# Why
Screenshot 2023-03-30 at 1 34 42 PM

# How Tested
Storybook

# Process & policy checklist


edwardlee-msft (approved):

packages/react-components/review/stable/react-components.api.md
```diff
@@ -122,14 +122,21 @@ export interface CameraButtonStyles extends ControlBarButtonStyles {
     menuStyles?: Partial<CameraButtonContextualMenuStyles>;
 }
 
+// @internal
+export const _Caption: (props: _CaptionProps) => JSX.Element;
+
+// @internal
```
PorterNan, line 128: Why do we put it here? onRenderAvatar should not be for each caption, it should be for the whole composite

packages/react-components/review/beta/react-components.api.md
```diff
@@ -234,14 +234,21 @@ export interface CameraSitePermissionsProps extends CommonSitePermissionsProps {
...
+// @internal
+export interface _CaptionProps extends _CaptionsInfo {
+    onRenderAvatar?: OnRenderAvatarCallback;
+}
+
 // @internal
 export const _CaptionsBanner: (props: _CaptionsBannerProps) => JSX.Element;
 
 // @internal
 export interface _CaptionsBannerProps {
```
PorterNan, line 249: onRenderAvatar?: OnRenderAvatarCallback; should still stay here, as well as CaptionsInfo,
Then you can construct _CaptionProps in the CaptionsBanner for Caption component

PorterNan (approved):

packages/react-components/review/stable/react-components.api.md
PorterNan: Back to this comment, it's okay to put it here, but we should bring it back to _CaptionsBannerProps 
Inheritance seems not necessary between these 2 components

PR #2864, opened 2023-03-31
Fix side pane when composite is narrow

mgamis-msft:
# What

Set composite min width to 30rem to avoid overlap of side pane when it is open. 

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3207599

# How Tested

Local sample testing

# Process & policy checklist


packages/react-composites/src/composites/common/styles/Composite.styles.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+
+/**
+ * @private
```
dmceachernmsft, line 4: nit: Maybe add a small comment here as to what the number is for.

dmceachernmsft (approved):

prabhjot-msft (approved):

PR #2865, opened 2023-03-31
Adapter API changes to add preset backgrounds to adapter

prabhjot-msft:
# What
1. Add preset backgrounds to adapter constructor
2. Add adapter function to set the default background images
3. Add changes to video effects pane for custom backgrounds

Screenshot 2023-03-31 at 11 44 55 AM

# Why
[TASK 3193145](https://skype.visualstudio.com/SPOOL/_workitems/edit/3193145)
[TASK 3193144](https://skype.visualstudio.com/SPOOL/_workitems/edit/3193144)

# How Tested

# Process & policy checklist


carocao-msft (approved):

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatBackedCallAdapter.ts
```diff
@@ -3,6 +3,8 @@
 
 import { CallWithChatAdapter } from './CallWithChatAdapter';
 import { CallAdapter, CallAdapterState } from '../../CallComposite';
+/* @conditional-compile-remove(video-background-effects) */
+import { VideoBackgroundImage } from '../../CallComposite';
```
dmceachernmsft, line 7: Should we not be getting a type like this from the Components package? are statefulClient users unable to use this type?

prabhjot-msft: this is the type of background image is only passed through composites while either initializing `adapter` or by using `updateBackgroundPickerImages`. Otherwise for components more generic form of effects item can be used to add images to picker `_VideoEffectsItem`

dmceachernmsft (approved):

PR #2874, opened 2023-03-31
Add hermetic tests for common call control in call composite

PorterNan:
# What

Add hermetic tests for common call control in call composite

**This will also fix problem for injected button as well** in call composite for new control bar

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/common/Drawer/MoreDrawer.tsx
```diff
@@ -193,6 +193,7 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
   if (drawerSelectionOptions !== false && isEnabled(drawerSelectionOptions?.peopleButton)) {
     drawerMenuItems.push({
       itemKey: 'people',
+      id: 'call-composite-drawer-people-button',
```
dmceachernmsft, line 196: Should this be a data-ui-id? I think this property sets the DOM id we are trying to avoid.

dmceachernmsft (approved):

packages/react-composites/src/composites/common/Drawer/MoreDrawer.tsx
PorterNan: It should be ideally, but seems like contextual menu doesn't allow us to set data-ui-id, so it is a workaround for this issue

carocao-msft (approved):

PR #2875, opened 2023-03-31
Use proxy for caption feature in stateful client

PorterNan:
# What

This is to solve get undefined available language bug of stateful client

(because normal spread operator does not support accessors well)
# Why

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/CallDeclarativeCommon.ts
```diff
@@ -111,3 +90,61 @@ export abstract class ProxyCallCommon implements ProxyHandler<CallCommon> {
...
+    this._call = call;
+  }
+
+  public get<P extends keyof TeamsCaptionsCallFeature>(target: TeamsCaptionsCallFeature, prop: P): any {
+    switch (prop) {
+      case 'startCaptions':
+        return this._context.withAsyncErrorTeedToState(
+          async (...args: Parameters<TeamsCaptionsCallFeature['startCaptions']>) => {
+            const ret = await target.startCaptions(...args);
+            this._context.setIsCaptionActive(this._call.id, true);
```
carocao-msft, line 113: Should we also set selected spoken language here?

PR #2879, opened 2023-04-03
[Chat][Feature] Preliminary implementation of _AtMentionFlyout component

Leah-Xia-Microsoft:
# What

Preliminary implementation of a new internal component _AtMentionFlyout. This PR covers:
- Setting up this new component 
- Using onQueryUpdated to fetch data when query changes
- Basic positioning of the flyout (Full implementation of positioning will be handled in another ticket)

What's not included:
- Mouses event and key events
- Localization
- Dismissing Flyout and wait state for onQueryUpdated
# Why

# How Tested

1. Add the following code to Input
```
        <_AtMentionFlyout
          query={textValue}
          target={inputBoxRef}
        />
```
2. Provide some fakeData for the `suggestions` state in `_AtMentionFlyout`:
```
  const fakeData: AtMentionSuggestion[] = [
    {
      userId: '1',
      suggestionType: 'person',
      displayName: ''
    },
    {
      userId: '2',
      suggestionType: 'person',
      displayName: 'Patricia Adams'
    },
    {
      userId: '3',
      suggestionType: 'person',
      displayName: '1'
    },
    {
      userId: '4',
      suggestionType: 'person',
      displayName: '2'
    }
]
```

# Process & policy checklist


emlynmac:
It would be really nice to have the component in the storybook updates in this initial PR

emlynmac (approved):
First part looking good; more to come

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -132,6 +132,35 @@ export type AreParamEqual<A extends (props: any) => JSX.Element | undefined, B e
 // @public
 export type AreTypeEqual<A, B> = A extends B ? (B extends A ? true : false) : false;
 
+// @beta
+export interface AtMentionDisplayOptions {
+    atMentionSuggestionRenderer?: (suggestion: AtMentionSuggestion) => JSX.Element;
+}
+
+// @internal
```
PorterNan, line 140: If that is exposed in communication-react.api.md, it shouldn't be internal, either make it beta or remove it from here

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -132,6 +132,35 @@ export type AreParamEqual<A extends (props: any) => JSX.Element | undefined, B e
 // @public
 export type AreTypeEqual<A, B> = A extends B ? (B extends A ? true : false) : false;
 
+// @beta
+export interface AtMentionDisplayOptions {
+    atMentionSuggestionRenderer?: (suggestion: AtMentionSuggestion) => JSX.Element;
```
PorterNan, line 137: Change the name to onRenderAtMentionSuggestion to align with other callback names

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -132,6 +132,35 @@ export type AreParamEqual<A extends (props: any) => JSX.Element | undefined, B e
...
+    onDismiss?: () => void;
+    onSuggestionSelected?: (suggestion: AtMentionSuggestion) => void;
+    suggestionItemRenderer?: (suggestion: AtMentionSuggestion, onSuggestionSelected?: (suggestion: AtMentionSuggestion) => void) => JSX.Element;
+    suggestions: AtMentionSuggestion[];
+    target?: React_2.RefObject<Element>;
+    title?: string;
+}
+
+// @beta
+export interface AtMentionLookupOptions {
```
PorterNan, line 151: Any special reason that we separate AtMentionDisplayOptions and AtMentionLookupOptions? it feels that you could merge them into one single option object like (AtMentionOptions) and make them optional. The idea is to reduce the props burden of messagethread

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1732,6 +1732,8 @@ export interface _IdentifierProviderProps {
 
 // @internal
 export interface _Identifiers {
+    atMentionSuggestionItem: string;
```
PorterNan, line 1735: I guess these should not be part of stable change?

packages/react-components/review/beta/react-components.api.md
```diff
@@ -64,6 +64,38 @@ export type AnnouncerProps = {
     ariaLive: 'off' | 'polite' | 'assertive' | undefined;
 };
 
+// @beta
+export interface AtMentionDisplayOptions {
+    atMentionSuggestionRenderer?: (suggestion: AtMentionSuggestion) => JSX.Element;
+}
+
+// @internal
```
PorterNan, line 72: Small tip: it's okay to have @internal here in the react-components.api.md, the one we really care is the final mega package

packages/react-components/review/stable/react-components.api.md
```diff
@@ -775,6 +775,8 @@ export interface _IdentifierProviderProps {
 
 // @internal
 export interface _Identifiers {
+    atMentionSuggestionItem: string;
```
PorterNan, line 778: Again, add CC and remove this from stable change

packages/react-components/src/components/AtMentionFlyout.tsx
```diff
@@ -0,0 +1,181 @@
...
+   */
+  onSuggestionSelected?: (suggestion: AtMentionSuggestion) => void;
+  /**
+   * Callback to invoke when the at mention flyout is dismissed
+   */
+  onDismiss?: () => void;
+  /**
+   * Optional callback to render an item of the atMention suggestions list.
+   */
+  suggestionItemRenderer?: (
```
PorterNan, line 46: onRenderSuggestionItem

packages/react-components/src/identifiers/IdentifierProvider.tsx
```diff
@@ -71,7 +75,9 @@ const defaultIdentifiers: _Identifiers = {
   verticalGalleryVideoTile: 'vertical-gallery-video-tile',
   horizontalGalleryVideoTile: 'horizontal-gallery-video-tile',
   /* @conditional-compile-remove(vertical-gallery) */
-  verticalGalleryPageCounter: 'vertical-gallery-page-counter'
+  verticalGalleryPageCounter: 'vertical-gallery-page-counter',
+  atMentionSuggestionList: 'at-mention-suggestion-list',
```
PorterNan, line 79: CC these 2 strings

packages/communication-react/review/beta/communication-react.api.md
Leah-Xia-Microsoft: @PorterNan Display option is used for displaying mentions while lookup option is used when sending/editing messages. MessageThread happens to need both because both displaying and editing goes through it. I think it would be a bit confusing if we combine them.

packages/communication-react/review/beta/communication-react.api.md
emlynmac: Let's think a bit about naming here.
When we see `on<something>`, they should be used for handling events that happen. This is not the role of every callback, however. In the case where we have an `onCancel` event, the cancellation _has happened_ and was triggered by the user.
In the case of this callback, it is actually providing the rendering function and is not called as an event after the rendering has occurred.

For this reason, we should keep `on<something>` to event handler callbacks, rather than just as a blanket naming of callbacks.

packages/react-components/src/components/AtMentionFlyout.tsx
emlynmac: See previous comment regarding event handlers vs functionality provider callbacks.

packages/communication-react/review/beta/communication-react.api.md
Leah-Xia-Microsoft: @PorterNan These are props for an internal component _AtMentionFlyout so I guess it should be internal instead of beta? By "remove it from here", do you mean manually remove it?  These are auto-generated for me when I ran `rush build -t communication-react`. I saw that other internal props (eg: _DrawerMenuProps) are declared the same way but somehow didn't show up here.

packages/communication-react/review/stable/communication-react.api.md
Leah-Xia-Microsoft: @PorterNan  These are needed in _AtMentionFlyout and since we're not removing _AtMentionFlyout using a feature flag, we cannot remove these. I tried to put a beta comment for these but seems like `rush build` still add them on stable. Any other way I can remove these from stable? Thanks

packages/communication-react/review/beta/communication-react.api.md
PorterNan: For AtMention feature itself, it makes sense to have it separated, but if you look at from a higher layer, it is just all AtMention feature related options, if people feel confused then that means we should give the property a better name

packages/communication-react/review/beta/communication-react.api.md
PorterNan: One of the reason we align with fluent is we reuse lots of their component and most of our contoso customer will use fluent library as well, so it is a lot easier to have same naming convention, we don't want them to fall into different worlds when they are using both. 


Within UI Library internally, for example, there is onRenderContextualItem in fluent, and we wrap it and re-export it as an interface from our side (same name, then we can just spread the object without doing lots of property copy)

To the event problem, there is still ways to differentiate these 2 different cases:
If we wanna say this is happen after the button is rendered, it should be onButtonRendered or onButtonMounted

It's already a convention for a long while in UI library + fluent, so I would say, it would be not ideal to create an exception and have 2 different naming conventions around the UI library

packages/communication-react/review/beta/communication-react.api.md
Leah-Xia-Microsoft: @PorterNan  The two options can potentially grow and combine them will make it hard for Contoso developers to figure out which ones they need for one of these 2 features. For example, for sendBox, we only need AtMentionLookupOptions, it's very easy and clear with  AtMentionLookupOptions  since all the necessary options are defined together and it's more type guarded. While as if we combine them, developers need to figure out which ones are for the lookup features for sendBox among the combined options.

emlynmac: > Any special reason that we separate AtMentionDisplayOptions and AtMentionLookupOptions? it feels that you could merge them into one single option object like (AtMentionOptions) and make them optional. The idea is to reduce the props burden of messagethread

I think what you're really asking for here is to have one `MentionOptions` optional object on `MessageThreadProps` that contains optional `MentionDisplayOptions` and `MentionLookupOptions`.
The two sub types are definitely separate sets of usage, but I see we would want to keep the `MessageThreadProps` as simple as possible.

PorterNan: @Leah-Xia-Microsoft, it's no doubt that sendBox just need one of the option to make it work, but we can define a single prop for sendbox instead of adding too many options to message thread
You can choose to do type MentionOptions = MentionDisplayOptions & MentionLookupOptions to make it work in messageThread

Also @emlynmac's way is a good one, we just need to give those nested options a good property name

packages/react-components/src/components/AtMentionFlyout.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const localeStrings = useLocale().strings.participantItem;
+
+  const [position, setPosition] = useState<Position>({ top: 0, right: 0, bottom: 0, left: 0 });
+  const [hoveredSuggestion, setHoveredSuggestion] = useState<AtMentionSuggestion | undefined>(undefined);
+
+  // Temporary implementation for AtMentionFlyout's position.
+  useEffect(() => {
+    const rect = target?.current?.getBoundingClientRect();
+    const { top = 0, left = 0, right = 0, bottom = 0, height = 0 } = rect ?? {};
+    const flyoutHeight = 212;
```
PorterNan, line 137: Looks like a temporary solution, but still would be nice to move those numbers to constants.ts

change-beta/@azure-communication-react-c9223c17-ac3b-4223-af7b-ceda8bc8052b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
PorterNan, line 2: the change type should be prerelease, since this is only a beta change (including all beta bug fixes and api changes in beta)

PorterNan (approved):

PR #2885, opened 2023-04-03
Replace Coffee Icon with Spinner

prabhjot-msft:
# What
Replace Coffee Icon with Spinner
Screenshot 2023-04-03 at 4 31 35 PM

Screenshot 2023-04-12 at 2 21 53 PM

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3110150

# How Tested

# Process & policy checklist


packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -305,8 +305,6 @@ export type CallCompositeIcons = {
     ErrorBarClear?: JSX.Element;
     HorizontalGalleryLeftButton?: JSX.Element;
     HorizontalGalleryRightButton?: JSX.Element;
-    LobbyScreenConnectingToCall?: JSX.Element;
-    LobbyScreenWaitingToBeAdmitted?: JSX.Element;
```
dmceachernmsft, line 309: Are we okay with these deletes to the API? this would be a breaking change to anyone who overrode that icon. Should we maybe update these icons to be spinners?

packages/react-composites/src/composites/CallComposite/components/LobbyTile.tsx
```diff
@@ -48,15 +48,8 @@ export const LobbyTile = (props: LobbyTileProps): JSX.Element => {
             aria-atomic
           >
             <Stack.Item className={mergeStyles(titleStyle(palette, isVideoReady))}>
-              {props.overlayProps.overlayIcon}
+              <Spinner size={SpinnerSize.large} label={props.overlayProps.title}></Spinner>
             </Stack.Item>
-            <Text
```
dmceachernmsft, line 53: is there an ask to remove this text along with the icon?

dmceachernmsft: Please also make sure this still looks correct when making a PSTN and 1:n call 😊

packages/react-composites/review/beta/react-composites.api.md
alkwa-msft: @prabhjot-msft  can we bring this up in design office hours?

alkwa-msft: Is this something we want to support if they already started customizing this area should we just let them? We are just changing the default behavior?

alkwa-msft (approved):
Lets just make sure we are only changing the default coffee icon behavior and not the actual lobby text

packages/react-composites/review/beta/react-composites.api.md
prabhjot-msft: After design discussion, replaced the coffee icon with spinner

packages/react-composites/src/composites/CallComposite/components/LobbyTile.tsx
prabhjot-msft: We have decided to not use spinner label and use existing text as of now.

PR #2888, opened 2023-04-04
Fix bug where non final captions are displayed as final captions 

carocao-msft:
# What
Fix bug where non final captions are displayed as final captions 

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3211438

# How Tested
Tested with samples

# Process & policy checklist

### DIFF
--- a/packages/calling-stateful-client/src/CallContext.ts
+++ b/packages/calling-stateful-client/src/CallContext.ts
@@ -706,7 +706,6 @@ export class CallContext {
     if (
       latestCaption &&
       latestCaption.resultType !== 'Final' &&
-      latestCaption.timestamp.getTime() === newCaption.timestamp.getTime() &&
       latestCaption.speaker.identifier &&
       newCaption.speaker.identifier &&
       toFlatCommunicationIdentifier(latestCaption.speaker.identifier) ===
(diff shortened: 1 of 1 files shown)


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -706,7 +706,6 @@ export class CallContext {
     if (
       latestCaption &&
       latestCaption.resultType !== 'Final' &&
```
PorterNan, line 708:
```
latestCaption.resultType !== 'Final'
```

Is it this line should be removed? Why does the time comparing need to be removed, I can feel this would cause lots of issues if we remove the time compare

PorterNan (changes requested):
Let's double check the caption behavior after this fix - in stateful client we are merging captions together depending on time because captions don't have their id, if we remove it, the merging function would just be not working

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -701,18 +701,19 @@ export class CallContext {
...
-      latestCaption &&
-      latestCaption.resultType !== 'Final' &&
-      latestCaption.timestamp.getTime() === newCaption.timestamp.getTime() &&
-      latestCaption.speaker.identifier &&
-      newCaption.speaker.identifier &&
-      toFlatCommunicationIdentifier(latestCaption.speaker.identifier) ===
-        toFlatCommunicationIdentifier(newCaption.speaker.identifier)
-    ) {
-      captions.pop();
+    for (let index = captions.length - 1; index >= 0; index--) {
```
PorterNan, line 704: Add a comment to explain this behavior for easier code reading

PorterNan (approved):

PR #2890, opened 2023-04-04
Alkwa/update calling stable dep

alkwa-msft:
# What
Updating stable calling dependency to 1.11.1

# Why
We want to be on the latest calling dependencies for both beta and stable at any time to ensure our customers are getting the best possible calling experience

# How Tested
Ran a build to ensure that it all compiled


packages/calling-stateful-client/package.json
```diff
@@ -38,7 +38,7 @@
     "@azure/communication-calling": "1.12.0-beta.2 || >=1.4.4"
```
dmceachernmsft, line 38: If we want to update the base version, then we need to update this as well.

alkwa-msft: I must have missed this one I will change this one as well

alkwa-msft: should be all updated now. I did a search for "1.12.0-beta.2 || >=1.4.4" and replaced with "1.12.0-beta.2 || >=1.11.1"

emlynmac (approved):

dmceachernmsft (approved):

PR #2891, opened 2023-04-04
Message edit fail error handling and UI

emlynmac:
# What

This PR implements UI changes for a failed message edit and updates the storybook with the changes.

# Why
If a message edit fails, we need to inform the user and allow recovery of the failure.

# How Tested
Storybook.

# Process & policy checklist


packages/react-components/review/beta/react-components.api.md
```diff
@@ -1082,6 +1089,7 @@ export type MessageProps = {
     showDate?: boolean;
     disableEditing?: boolean;
     onUpdateMessage?: UpdateMessageCallback;
+    onCancelMessageEdit?: CancelEditCallback;
```
PorterNan, line 1092: Is cancel a user triggered action or it is triggered by failed errors?

packages/react-components/review/beta/react-components.api.md
```diff
@@ -234,6 +234,11 @@ export interface CameraSitePermissionsProps extends CommonSitePermissionsProps {
 // @beta
 export type CameraSitePermissionsStrings = SitePermissionsStrings;
 
+// @public
+export type CancelEditCallback = (messageId: string, metadata?: Record<string, string>, options?: {
+    attachedFilesMetadata?: FileMetadata[];
```
PorterNan, line 239: Curious about the parameters here - if we wanna minimize the callback param, messageId should be enough, contoso can grab the real message from state themselves.

If we want to provide more info and make it easier to be consumed, I guess pass the whole message object would be good as well

if we want to separate properties one by one into the params. IMO, for priority, message content >>  metadata > fileMetadata

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
```diff
@@ -163,6 +170,11 @@ export const ChatMessageComponentAsEditBox = (props: ChatMessageComponentAsEditB
           id={'submitIconWrapper'}
         />
       </InputBoxComponent>
+      {message.failureReason && (
+        <div className={mergeStyles(chatMessageFailedTagStyle(theme), { padding: '0.5rem' })}>
```
PorterNan, line 174: Let's add padding: '0.5rem' to the existing style directly, or create a style in the style file

packages/react-components/src/types/ChatMessage.ts
```diff
@@ -61,6 +61,7 @@ export interface ChatMessage extends MessageCommon {
   senderId?: string;
   senderDisplayName?: string;
   status?: MessageStatus;
+  failureReason?: string;
```
PorterNan, line 64: Curious about where does this failureReason coming from? From our state.error or somewhere else

PorterNan (commented):
Nice to see we have an isolated error handling system specifically for edit box!

For all the default error handling, any error caught by UI library will naturally flow to state.error then automatically pop up on error bar (but we need to ensure we bubble up the error well)

For this error handling feature, I have 2 minor questions:
1. Is there a special concern that we don't wanna use our basic error bar system (default way of error handling)
2. I guess we wouldn't want to show double error (both in error bar and the message) if point 1 is what we really want, we should do some kinda filtering in error bar

PorterNan (commented):
There are 2 ways of showing the error of update message
1. Change the existing message status to sending -> failed, this might confuse users a bit, cause they don't know if the original message has been sent or just failed to be edited. Just because all of those share the same status icon. But it reuse current UI nicely

2. Add a banner on the message to indicate it's failed to update. There is no confusion, but UI would be a bit heavier

I've tested in teams, it's choosing 1 as the solution, are we saying we will follow teams and choose 1 as well?

If 1 is our answer, curious to see the new logic in selector how we mark a message as "sending" and "failed" then!

packages/react-components/review/beta/react-components.api.md
emlynmac: It is a user triggered action; they need to fire it with the cancel X button on the edit box message.

packages/react-components/review/beta/react-components.api.md
emlynmac: When I was thinking about this, we have to have the message ID. This would allow a user to look up any other message parts if they need them. 
As there's this signature on the edit, it seemed appropriate to use a similar callback signature here. We don't need the message content however as this is a cancel.

packages/react-components/src/types/ChatMessage.ts
emlynmac: It's the prop to provide the context to show in the UI. It will be provided by wherever raises the error, which in the edit case would be when the onUpdateMessage callback throws an error.

emlynmac:
> There are 2 ways of showing the error of update message
> 
> 1. Change the existing message status to sending -> failed, this might confuse users a bit, cause they don't know if the original message has been sent or just failed to be edited. Just because all of those share the same status icon. But it reuse current UI nicely
> 2. Add a banner on the message to indicate it's failed to update. There is no confusion, but UI would be a bit heavier
> 
> I've tested in teams, it's choosing 1 as the solution, are we saying we will follow teams and choose 1 as well?
> 
> If 1 is our answer, curious to see the new logic in selector how we mark a message as "sending" and "failed" then!

The UX has a design, I can share that with you. 
You can also see the working system in the storybook.

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
emlynmac: Line 103 in this file shows the same sort of pattern; adding a padding to the style for layout.
Would it be better to separate the layout and text styling here or use the merge?

Leah-Xia-Microsoft (approved):
LGTM

packages/react-components/review/beta/react-components.api.md
PorterNan: Yea, that makes sense to drop those signature - I believe onEditMessage is getting the metadata when someone upload a new file or remove one or two files from the message, so it makes more sense to get them, because they need to be passed to underlying handler and sdk

When I am talking about sdk - it reminds me that all these onEditMessage/onSendMessage is actually coming from underlying handlers and sdk - will we have a new handler for this onCancelMessage  and connect it to sdk? If not, I would like to see what user case we wanna support when expose this api - it seems like not a required api for this error handling feature

PorterNan (approved):

PR #2893, opened 2023-04-04
Captions on mobile

carocao-msft:
# What
Captions on mobile

# Why

# How Tested
Samples

https://user-images.githubusercontent.com/96077406/229933661-fa38d45b-0311-4357-98b4-ad4e2561e8a4.mov

# Process & policy checklist


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2096,6 +2099,7 @@ export const DEFAULT_COMPOSITE_ICONS: {
     CaptionsIcon: JSX.Element;
     CaptionsOffIcon: JSX.Element;
     CaptionsSettingIcon: JSX.Element;
+    PersonIcon: JSX.Element;
```
PorterNan, line 2102: Let's not make the name so generic, if that is the icon f or the ChangeSpokenLanguage, then we call it ChangeSpokenLanguageIcon or something

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2096,6 +2099,7 @@ export const DEFAULT_COMPOSITE_ICONS: {
     CaptionsIcon: JSX.Element;
     CaptionsOffIcon: JSX.Element;
     CaptionsSettingIcon: JSX.Element;
```
PorterNan, line 2101: BTW, let's name it consistently here - if we call it CaptionsSetting then we call it all CaptionsSetting, if we call it CaptionsSetting*s* then we add all s to the end

I've seen Settings a lot from other UI, maybe we could just name it all as CaptionsSetting*s*

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -341,7 +337,7 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
   /* @conditional-compile-remove(PSTN-calls) */
   const alternateCallerId = callAdapter.getState().alternateCallerId;
   /* @conditional-compile-remove(close-captions) */
-  const isTeamsCall = useSelector(getIsTeamsCall);
+  const isTeamsCall = callAdapter.getState().isTeamsCall;
```
PorterNan, line 340: Will this value always keep unchanged or not? If this could be potentially changed during the call, we should still use selectors

packages/react-composites/src/composites/common/CaptionsBanner.tsx
```diff
@@ -62,13 +62,15 @@ export const CaptionsBanner = (): JSX.Element => {
           captionsBannerProps.isCaptionsOn && (
             <div className={containerClassName}>
               <Stack horizontalAlign="center">
-                <Stack.Item style={{ width: '50%' }}>
+                <Stack.Item style={{ width: props.isMobile ? '90%' : '50%' }}>
```
PorterNan, line 65: Let's extra these 2 constants to the top of the component like 

const mobileViewBannerWidth = '90%';
const desktopVIewBannerWidth = '50%';

packages/react-composites/src/composites/common/Drawer/MoreDrawer.tsx
```diff
@@ -230,6 +261,137 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
     });
   }
 
+  //Captions drawer menu
```
PorterNan, line 264: Move this line under CC comment - so it will be removed in stable

packages/react-composites/src/composites/common/Drawer/MoreDrawer.tsx
```diff
@@ -230,6 +261,137 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
...
+          root: {
+            margin: 0
+          },
+          pill: {
+            backgroundColor: `${theme.palette.themePrimary} !important`
+          },
+          thumb: {
+            backgroundColor: `${theme.palette.white} !important`
+          }
+        };
```
PorterNan, line 330: Let's move all this style into a style file? Maybe MoreDrawer.style.ts or something

packages/react-composites/src/composites/common/Drawer/MoreDrawer.tsx
```diff
@@ -230,6 +261,137 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
...
+            root: {
+              span: {
+                fontWeight: 600,
+                fontSize: _pxToRem(17),
+                lineHeight: _pxToRem(22),
+                color: theme.palette.neutralDark
+              }
+            }
+          }
+        }
```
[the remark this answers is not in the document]
PorterNan: Same here, move to a separated file

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
carocao-msft: The value will keep unchanged since it's indicating if its currently a teams call or ACS call

PorterNan (approved):
Looks great! Thanks for resolving those comments!

PR #2894, opened 2023-04-05
[Chat] Inline Image UI and storybook doc for inline image

jimchou-dev:
# What

* Implemented inline image UI with html-to-react
* Added docs for inline image in MessageThread for storybook

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -66,6 +68,18 @@ type ChatMessageComponentAsMessageBubbleProps = {
...
+   * Optional function to fetch attachments.
+   * @beta
+   */
+  onFetchAttachments?: (attachment: FileMetadata) => Promise<void>;
+  /* @conditional-compile-remove(teams-inline-images) */
+  /**
+   * Optional map of attachment ids to blob urls.
+   * @beta
+   */
+  attachmentsMap?: Record<string, string>;
```
JoshuaLai, line 82: As an update we will need to be a little smarter with this map. Right now any changes will trigger all message components to re-render. 
Suggestion:  Maybe we create a map of maps, keyed by message id. 
```ts
Record<string, [Record<string, string>]>
```

jimchou-dev: I think it'd require a bigger change that may change the api to also take in the message or messageId for onFetchAttachments. Probably should be a separate task for optimization

JoshuaLai: I agree with this. When we start the e2e implementation we can begin optimizing.

JoshuaLai (approved):

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -146,3 +152,44 @@ describe('Message blocked should display default blocked text correctly', () =>
     expect(component.text()).toContain(testLocale.strings.messageThread.blockedWarningText);
   });
 });
+
+/* @conditional-compile-remove(teams-inline-images) */
+describe.only('Message should display inline image correctly', () => {
```
mgamis-msft, line 157: Remove .only here

mgamis-msft (approved):

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -152,3 +160,63 @@ const messageContentAriaText = (props: ChatMessageContentProps): string | undefi
...
+      // logic to check id in map/list
+      const fileMetadata = props.message.attachedFilesMetadata?.find((f) => f.id === node.attribs.id);
+      // if in cache, early return
+      if (props.attachmentsMap && node.attribs.id in props.attachmentsMap) {
+        node.attribs = { ...node.attribs, src: props.attachmentsMap[node.attribs.id] };
+        return processNodeDefinitions.processDefaultNode(node, children, index);
+      }
+      // not yet in cache
+      if (fileMetadata && props.onFetchAttachment && props.attachmentsMap) {
+        props.onFetchAttachment(fileMetadata);
```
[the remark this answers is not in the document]
jimchou-dev: We design this so it doesn't need to await and can render immediately (even though it might be broken image), and when the URl gets updated in the attachmentsMap, will trigger a re-render with the proper image src

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -146,3 +152,44 @@ describe('Message blocked should display default blocked text correctly', () =>
...
+    const onFetchAttachment = async (attachment: FileMetadata): Promise<string> => {
+      return attachment.previewUrl ?? '';
+    };
+    const component = mount(
+      <MessageThread userId="user1" messages={[sampleMessage]} onFetchAttachments={onFetchAttachment} />
+    );
+    await act(async () => {
+      await new Promise((resolve) => setTimeout(resolve, 0));
+      component.update();
+
```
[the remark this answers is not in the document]
jimchou-dev: Sure will add a expect before the await setTimeout for comparing the img src value prior to the update =)

PR #2897, opened 2023-04-06
Fixed bug where getting error when joining call, and bug when users can not get existing spoken language

carocao-msft:
# What
Fixed bug where getting error when joining call, and bug when users can not get existing spoken language
Also fixed this bug : when call is on hold and then resume, will receive duplicate captions 
# Why
Captions were subscribed before call is connected, which resulted in a build error
Update spoken and caption language when receiving new captions (similar behavior to teams), so all users in the call can receive updated spoken/caption language
Fixed some empty checks in the code for spoken language (need to check for spokenLanguage = '') 

# How Tested
Tested with samples 

# Process & policy checklist


packages/calling-stateful-client/src/CallDeclarativeCommon.ts
```diff
@@ -82,7 +82,10 @@ export abstract class ProxyCallCommon implements ProxyHandler<CallCommon> {
                 const ret = await captionsFeature.startCaptions(...args);
 
                 this._context.setIsCaptionActive(target.id, true);
-                this._context.setSelectedSpokenLanguage(target.id, args[0]?.spokenLanguage ?? 'en-us');
+                if (args[0]?.spokenLanguage && args[0]?.spokenLanguage !== '') {
```
PorterNan, line 85: We shouldn't do it here, we should modify the behavior when you click the button, not the behavior of the function itself, functions should stay what they mean to be

+1 to keep `this._context.setSelectedSpokenLanguage(target.id, args[0]?.spokenLanguage);` here

For calling `await captionsFeature.setSpokenLanguage(args[0].spokenLanguage);` it should be happening in where onStartCaption function in the component

packages/react-components/src/permissions/PermissionsProvider.tsx
```diff
@@ -92,7 +92,7 @@ export const _usePermissions = (): _Permissions => useContext(PermissionsContext
  * @beta
  * The role of a call participant.
  */
-export type Role = 'Presenter' | 'Attendee' | 'Consumer' | 'Organizer';
+export type Role = 'Presenter' | 'Attendee' | 'Consumer' | 'Organizer' | 'Co-organizer';
```
PorterNan, line 95: We do we change this in the PR, if this is unrelated, let's open a new one

packages/react-components/src/components/StartCaptionsButton.tsx
```diff
@@ -80,7 +80,7 @@ export const _StartCaptionsButton = (props: _StartCaptionsButtonProps): JSX.Elem
   };
 
   const captionsOptions: _captionsOptions = {
-    spokenLanguage: currentSpokenLanguage ?? 'en-us'
+    spokenLanguage: !currentSpokenLanguage || currentSpokenLanguage === '' ? 'en-us' : currentSpokenLanguage
```
alkwa-msft, line 83: can 'en-us' be a constant somewhere?

packages/calling-stateful-client/src/CallSubscriber.ts
```diff
@@ -168,6 +161,15 @@ export class CallSubscriber {
 
   private stateChanged = (): void => {
     this._context.setCallState(this._callIdRef.callId, this._call.state);
+
+    // subscribe to captions here so that we don't call captions when call is not initialized
+    if (_isTeamsMeetingCall(this._call) && this._call.state === 'Connected') {
```
alkwa-msft, line 166: - is this the only event we can subscribe to after we connect to a call?
- what if we were in two calls and want to have captions in both as we jump between the calls?
- could we potentially expose a new "event" to subscribe to calls and this function only gets fired once the first time we connect to the main call?

packages/react-components/src/components/CaptionsSettingModal.tsx
```diff
@@ -73,8 +73,8 @@ export const _CaptionsSettingModal = (props: _CaptionsSettingModalProps): JSX.El
   const theme = useTheme();
 
   const [selectedItem, setSelectedItem] = useState<IDropdownOption>({
-    key: currentSpokenLanguage !== '' ? currentSpokenLanguage : 'en-us',
-    text: currentSpokenLanguage !== '' ? currentSpokenLanguage : 'en-us'
+    key: currentSpokenLanguage && currentSpokenLanguage !== '' ? currentSpokenLanguage : 'en-us',
```
alkwa-msft, line 76: is currentSpokenLanguage going to be undefined or empty string?

We are checking if its an empty string but we also just said it could be undefined. Can we determine if its only undefined or only an empty string but not potentially both?

alkwa-msft (approved):

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -156,19 +156,13 @@ export const getCaptionsStatus = (state: CallClientState, props: CallingBaseSele
 
 /* @conditional-compile-remove(close-captions) */
 /** @private */
-export const getCurrentCaptionLanguage = (
-  state: CallClientState,
-  props: CallingBaseSelectorProps
-): string | undefined => {
+export const getCurrentCaptionLanguage = (state: CallClientState, props: CallingBaseSelectorProps): string => {
   return state.calls[props.callId]?.captionsFeature.currentCaptionLanguage;
```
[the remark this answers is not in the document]
PorterNan: either do, state.calls[props.callId]?.captionsFeature.currentCaptionLanguage ?? ''
Or leave the return type as string | undefined

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -156,19 +156,13 @@ export const getCaptionsStatus = (state: CallClientState, props: CallingBaseSele
...
 };
 
 /* @conditional-compile-remove(close-captions) */
 /** @private */
-export const getCurrentSpokenLanguage = (
-  state: CallClientState,
-  props: CallingBaseSelectorProps
-): string | undefined => {
+export const getCurrentSpokenLanguage = (state: CallClientState, props: CallingBaseSelectorProps): string => {
   return state.calls[props.callId]?.captionsFeature.currentSpokenLanguage;
```
PorterNan, line 166: Same here, return type not matched

packages/calling-stateful-client/src/CallSubscriber.ts
```diff
@@ -173,6 +165,16 @@ export class CallSubscriber {
 
   private stateChanged = (): void => {
```
PorterNan, line 166: stateChanged() serves for its own purposes, let's not add unrelated code here

create a new private function named initCaptionSubscriber(), then add to stateChanged event:
```
this._call.on('stateChanged', this.initCaptionSubscriber);
```

And don't forget to unsubscribe the event after you creating the subscriber (also you will need to unsubscribe it in unsubscribe() function)

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -688,7 +688,6 @@ export type CallErrors = {
...
-
```
PorterNan, line 691: It is interesting that why it is changed, but just one empty line

packages/react-composites/src/composites/CallComposite/hooks/useHandlers.ts
```diff
@@ -110,6 +110,9 @@ const createCompositeHandlers = memoizeOne(
     /* @conditional-compile-remove(close-captions) */
     onStartCaptions: async (captionsOptions) => {
       await adapter.startCaptions(captionsOptions);
+      if (captionsOptions?.spokenLanguage && captionsOptions.spokenLanguage !== '') {
```
PorterNan, line 113: This logic fix should be more like in the UI(composite) instead of handler,  because once we have the isCaptionGloballyActive in GA, it is also UI's choice to decide whether we show the modal or not.

also we might need to comment why we need a fix there in UI

packages/react-components/src/components/StartCaptionsButton.tsx
```diff
@@ -84,14 +90,17 @@ export const _StartCaptionsButton = (props: _StartCaptionsButtonProps): JSX.Elem
...
   const captionsOptions: _captionsOptions = {
-    spokenLanguage: currentSpokenLanguage ?? 'en-us'
+    spokenLanguage: currentSpokenLanguage === '' ? defaultSpokenLanguage : currentSpokenLanguage
   };
 
   const onToggleStartCaptions = (): void => {
     if (props.checked) {
       onStopCaptions();
     } else {
       onStartCaptions(captionsOptions);
```
PorterNan, line 100: I guess we will need to await the onStartCaptions(captionsOptions);
Then we go ahead to onSetSpokenLanguage(captionsOptions.spokenLanguage);

packages/react-composites/src/composites/common/ControlBar/DesktopMoreButton.tsx
```diff
@@ -56,6 +56,15 @@ export const DesktopMoreButton = (props: DesktopMoreButtonProps): JSX.Element =>
   const startCaptionsButtonProps = useAdaptedSelector(_startCaptionsButtonSelector);
   /* @conditional-compile-remove(close-captions) */
   const startCaptionsButtonHandlers = useHandlers(_StartCaptionsButton);
+  /* @conditional-compile-remove(close-captions) */
+  const startCaptions = useCallback(() => {
+    startCaptionsButtonHandlers.onStartCaptions({
```
PorterNan, line 61: await this one then go ahead to next function

packages/react-composites/src/composites/common/Drawer/MoreDrawer.tsx
```diff
@@ -313,6 +315,9 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
                 startCaptionsButtonHandlers.onStartCaptions({
```
PorterNan, line 315: await onStartCaptions

PorterNan (approved):

packages/react-composites/src/composites/common/Drawer/MoreDrawer.tsx
```diff
@@ -308,11 +310,14 @@ export const MoreDrawer = (props: MoreDrawerProps): JSX.Element => {
           <Toggle
             checked={startCaptionsButtonProps.checked}
             styles={themedToggleButtonStyle(theme, startCaptionsButtonProps.checked)}
-            onChange={() => {
+            onChange={async () => {
```
PorterNan, line 313: Wrap this function with a useCallback and assign it to a const value

packages/react-components/src/components/StartCaptionsButton.tsx
```diff
@@ -84,14 +90,17 @@ export const _StartCaptionsButton = (props: _StartCaptionsButtonProps): JSX.Elem
   };
 
   const captionsOptions: _captionsOptions = {
-    spokenLanguage: currentSpokenLanguage ?? 'en-us'
+    spokenLanguage: currentSpokenLanguage === '' ? defaultSpokenLanguage : currentSpokenLanguage
   };
 
-  const onToggleStartCaptions = (): void => {
+  const onToggleStartCaptions = async (): Promise<void> => {
```
PorterNan, line 96: wrap it with useCallback

PR #2906, opened 2023-04-10
Jinan/add cc for caption

PorterNan:
# What

Add initial CC to caption

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -311,16 +311,10 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
       );
       break;
     /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
-    case holdPageTrampoline():
+    case 'hold':
       pageElement = (
         <>
```
carocao-msft, line 316: Why are we changing hold page logic here?

PorterNan: https://github.com/Azure/communication-ui-library/pull/2905
it's coming from this, need to add new CC feature to make this work, this part is to adjust to that new CC feature

PR #2909, opened 2023-04-11
Fix dismiss dropdown problem when new captions come

PorterNan:
# What

Fix dismiss dropdown problem when new captions come

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/CaptionsSettingModal.tsx
```diff
@@ -80,27 +81,37 @@ export const _CaptionsSettingModal = (props: _CaptionsSettingModalProps): JSX.El
...
 
-  const onClickConfirm = (language: string): void => {
-    if (isCaptionsFeatureActive) {
-      onSetSpokenLanguage(language);
-    } else {
-      onStartCaptions({ spokenLanguage: selectedItem.text });
-    }
-    onDismissTriggered();
-  };
+  const onClickConfirm = useCallback(
```
alkwa-msft, line 90: should we rename this just to onConfirm? Does it matter if its a click or a touch?

alkwa-msft (approved):

PR #2911, opened 2023-04-11
[Bugfix] add participant count to participant list and Composite people pane

dmceachernmsft:
# What

image

# Why

Shows the total participant count in the participant list

https://skype.visualstudio.com/SPOOL/_workitems/edit/3124032
# How Tested

Tested locally. +0 participant string is not there when normal. number is displayed participants subtracted from the total count coming from the SDK.


packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -66,7 +71,11 @@ export const ParticipantListWithHeading = (props: {
   return (
     <Stack className={participantListStack}>
       <Stack.Item styles={subheadingStyleThemed} aria-label={title}>
-        {title}
+        {title}{' '}
```
alkwa-msft, line 74: can we string.format this instead of having a bunch of strings that happen to fit next to eachother?

packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -52,6 +53,10 @@ export const ParticipantListWithHeading = (props: {
 }): JSX.Element => {
   const { onFetchAvatarPersonaData, onFetchParticipantMenuItems, title, participantListProps } = props;
   const theme = useTheme();
+  /* @conditional-compile-remove(total-participant-count) */
+  const adapter = useAdapter();
+  /* @conditional-compile-remove(total-participant-count) */
+  const totalParticipantCount = adapter.getState().call?.totalParticipantCount;
```
alkwa-msft, line 59: maybe there is a way to make this one prop that we get through the selector and we light it up for the participantList selector? totalParticipantCount

packages/react-components/src/localization/locales/en-US/strings.json
```diff
@@ -9,6 +9,9 @@
     "participantStateRinging": "Calling...",
     "participantStateHold": "On hold"
   },
+  "ParticipantList": {
+    "overflowParticipantCount": "+{overflowCount} more"
```
alkwa-msft, line 13: I am curious if we can review the use of "more" im unsure how its going to scale when we think of all of the other localizations

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -116,9 +120,13 @@ export const participantListSelector: ParticipantListSelector = createSelector(
...
       isRemovable: false
     });
+    /* @conditional-compile-remove(total-participant-count) */
+    const totalParticipantCount = partitipantCount;
     return {
       participants: participants,
-      myUserId: userId
+      myUserId: userId,
+      /* @conditional-compile-remove(total-participant-count) */
+      totalParticipantCount: totalParticipantCount
```
alkwa-msft, line 129: awesome!

packages/react-composites/src/composites/common/ParticipantContainer.tsx
dmceachernmsft: yup totally forgot the props are valid here. will change this!

packages/react-composites/tests/app/call/HermeticApp.tsx
```diff
@@ -23,10 +24,37 @@ export function HermeticApp(props: { queryArgs: QueryArgs }): JSX.Element {
 
   useEffect(() => {
     (async (): Promise<void> => {
```
JamesBurnside, line 26: does this need to be async?

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -116,9 +120,13 @@ export const participantListSelector: ParticipantListSelector = createSelector(
       // Local participant can never remove themselves.
       isRemovable: false
     });
+    /* @conditional-compile-remove(total-participant-count) */
+    const totalParticipantCount = partitipantCount;
```
JamesBurnside, line 124: question on this -- does this include participants that are in lobby or earlier states that we typically hide from the participant list?

packages/react-components/review/beta/react-components.api.md
```diff
@@ -1620,9 +1622,17 @@ export type ParticipantListProps = {
...
     styles?: ParticipantListStyles;
     showParticipantOverflowTooltip?: boolean;
+    totalParticipantCount?: number;
+    strings?: ParticipantListStrings;
     participantAriaLabelledBy?: string;
 };
 
+// @beta
+export interface ParticipantListStrings {
+    // (undocumented)
```
JamesBurnside, line 1632: Add documentation

packages/react-composites/tests/app/call/HermeticApp.tsx
```diff
@@ -22,11 +23,38 @@ export function HermeticApp(props: { queryArgs: QueryArgs }): JSX.Element {
...
+          ...queryArgs.mockCallAdapterState,
+          call: {
+            ...call,
+            localVideoStreams: [],
+            remoteParticipants: participants,
+            totalParticipantCount: Object.values(participants).length + 1
+          }
+        };
+        console.log('Creating mock adapter with args', queryArgs.mockCallAdapterState);
+        console.log(callAdapterState.call?.totalParticipantCount);
```
JamesBurnside, line 50: remove console logs

packages/react-composites/tests/app/call/QueryArgs.ts
```diff
@@ -29,6 +29,7 @@ export interface QueryArgs {
   groupId: string;
   token: string;
   displayName: string;
+  makeMeLotsOfPeople: boolean;
```
JamesBurnside, line 32: while I do like the name, its a little unclear on the purpose - are you making me (the test user) into lots of people? or are you making lots of people for the test user .. as other remote participants?

```suggestion
  mockRemoteParticipantCount: number
```

packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -98,3 +103,11 @@ export const ParticipantListWithHeading = (props: {
     </Stack>
   );
 };
+
+const paneTitleTrampoline = (paneTitle: string, totalParticipantCount?: number): string => {
+  /* @conditional-compile-remove(total-participant-count) */
+  const participantCountString = totalParticipantCount ? `(${totalParticipantCount})` : '';
+  /* @conditional-compile-remove(total-participant-count) */
+  return `${paneTitle} ${participantCountString}`;
```
JamesBurnside, line 111: Will this work for RTL languages? (should the (`participantCountString`) be before the title for those languages

JamesBurnside: Also, the override string is:
image

Given its so generic, I don't think we should add to it.

Maybe we could update our strings.json to be `"peoplePaneSubTitle": "In this call ({participantCount})",` instead?

JamesBurnside (changes requested):
Documentation on new apis and perhaps change to how the in this call string is augmented to have the count

packages/react-composites/src/composites/common/ParticipantContainer.tsx
dmceachernmsft: This was done this way for CC reasons since we cannot CC out of the localization files, will look at this again though maybe there is a better way

JamesBurnside (approved):

PR #2914, opened 2023-04-11
Teams adhoc enabled for calling handlers

mgamis-msft:
# What

Enable Teams adhoc calling for calling handler `startCall`

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3132743

# How Tested

Testing on local calling sample:
https://user-images.githubusercontent.com/79475487/231989813-5b13cb69-dc9c-4310-bc4c-b73af4a8fcbf.mp4

# Process & policy checklist


samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -71,6 +71,8 @@ export interface HomeScreenProps {
     teamsToken?: string;
     /* @conditional-compile-remove(teams-identity-support) */
     teamsId?: string;
+    /* @conditional-compile-remove(teams-adhoc-call) */
+    teamsIds?: string[];
```
dmceachernmsft, line 75: can we call this `outboundTeamsUsers` or something so we know it is the identifiers that we are calling?

samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -129,6 +133,8 @@ export const HomeScreen = (props: HomeScreenProps): JSX.Element => {
   const [teamsToken, setTeamsToken] = useState<string>();
   /* @conditional-compile-remove(teams-identity-support) */
   const [teamsId, setTeamsId] = useState<string>();
+  /* @conditional-compile-remove(teams-adhoc-call) */
+  const [teamsIds, setTeamsIds] = useState<string | undefined>();
```
dmceachernmsft, line 137: Can we update this like suggested above?

samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -257,6 +266,18 @@ export const HomeScreen = (props: HomeScreenProps): JSX.Element => {
                 </Stack>
               )
             }
+            {
+              /* @conditional-compile-remove(teams-adhoc-call) */ teamsAdhocChosen && (
+                <Stack>
+                  <TextField
+                    className={outboundTextField}
+                    label={"Teams ID's"}
```
dmceachernmsft, line 274: Can we say "Teams user ID's" and maybe in the placeholder give an example since we need to prefixes to have the handlers parse correctly? or alternatively give a callout like with the alternate callerId showing an example of a good user ID to avoid confusion when others want to try it?

mgamis-msft: Good point. Changed label and added a callout.

dmceachernmsft (approved):

PR #2915, opened 2023-04-11
[Chat] Inline image chatSDK e2e

jimchou-dev:
# What

- Added mapping to expose attachment from new ChatSDK version
- Added default onFetchAttachment props for MessageThread in ChatComposite

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3209160

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -47,6 +47,10 @@ import { useSelector } from './hooks/useSelector';
 import { FileDownloadErrorBar } from './FileDownloadErrorBar';
 /* @conditional-compile-remove(file-sharing) */
 import { _FileDownloadCards } from '@internal/react-components';
+/* @conditional-compile-remove(file-sharing) */
+import { AttachmentDownloadResult } from '@internal/react-components';
+import { ChatAttachment } from '@azure/communication-chat';
+import { string } from 'yargs';
```
JoshuaLai, line 53: Was this import a mistake?

packages/chat-stateful-client/src/EventSubscriber.ts
```diff
@@ -42,7 +42,7 @@ export class EventSubscriber {
     return convertChatMessage({
       id: event.id,
       version: event.version,
-      content: { message: event.message },
+      content: { message: event.message, attachments: event.attachments },
```
JoshuaLai, line 45: Do we need to flag this? Actually can we flag this?

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -394,6 +398,13 @@ export class AzureCommunicationCallWithChatAdapter implements CallWithChatAdapte
   public updateFileUploadMetadata = (id: string, metadata: FileMetadata): void => {
     this.chatAdapter.updateFileUploadMetadata(id, metadata);
   };
+  /* @conditional-compile-remove(teams-inline-images) */
+  async downloadAuthenticatedAttachment(attachmentUrl: string): Promise<AttachmentDownloadResult> {
```
jimchou-dev, line 402: Maybe make this `Promise<AttachmentDownloadResult | undefined>`?

packages/react-composites/src/composites/CallWithChatComposite/adapter/CallWithChatBackedChatAdapter.ts
```diff
@@ -126,6 +128,15 @@ export class CallWithChatBackedChatAdapter implements ChatAdapter {
   public updateFileUploadMetadata = (id: string, metadata: FileMetadata): void => {
     this.callWithChatAdapter.updateFileUploadMetadata(id, metadata);
   };
+
+  /* @conditional-compile-remove(teams-inline-images) */
+  public async downloadAuthenticatedAttachment(attachmentUrl: string): Promise<AttachmentDownloadResult> {
```
jimchou-dev, line 133: same here, maybe make this `Promise<AttachmentDownloadResult | undefined>`?

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -778,7 +778,7 @@ export interface CallWithChatAdapterManagement {
     deleteMessage(messageId: string): Promise<void>;
     disposeStreamView(remoteUserId?: string, options?: VideoStreamOptions): Promise<void>;
     // (undocumented)
-    downloadAuthenticatedAttachment?: (attachmentUrl: string) => Promise<AttachmentDownloadResult>;
+    downloadAuthenticatedAttachment?: (attachmentUrl: string) => Promise<AttachmentDownloadResult | undefined>;
```
PorterNan, line 781: What does it means to return undefined? Like it is failed?

JoshuaLai: At the moment the only time we return `undefined` is if `downloadAuthenticatedAttachment` is not there. Since it is an optional method.

jimchou-dev: For example if an attachment they don't want to process for some reason, because returning `{ blobUrl: '' }` with empty string doesn't really make sense, so would return undefined instead

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -71,6 +73,35 @@ const extractAttachedFilesMetadata = (metadata: Record<string, string>): FileMet
...
+    id: attachment.id,
+    name: attachment.name ?? '',
+    extension: attachment.contentType ?? '',
+    url: attachment.url,
+    previewUrl: attachment.previewUrl
+  }));
+};
+
+/* @conditional-compile-remove(file-sharing) @conditional-compile-remove(teams-inline-images) */
+const extractFilesMetadata = (message: ChatMessageWithStatus): FileMetadata[] => {
```
PorterNan, line 89: I thought e2e is like e2e test :D, it's like get it working end 2 end

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -838,7 +840,9 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
         return;
       }
       const attachmentDownloadResult = await onFetchAttachments(attachment);
-      setInlineAttachments((prev) => ({ ...prev, [attachment.id]: attachmentDownloadResult.blobUrl }));
+      if (attachmentDownloadResult) {
```
PorterNan, line 843: Nice to see an attachment-based update

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -394,6 +398,13 @@ export class AzureCommunicationCallWithChatAdapter implements CallWithChatAdapte
   public updateFileUploadMetadata = (id: string, metadata: FileMetadata): void => {
     this.chatAdapter.updateFileUploadMetadata(id, metadata);
   };
+  /* @conditional-compile-remove(teams-inline-images) */
+  async downloadAuthenticatedAttachment(attachmentUrl: string): Promise<AttachmentDownloadResult | undefined> {
+    if (this.chatAdapter.downloadAuthenticatedAttachment === undefined) {
```
PorterNan, line 403: Why would this api be undefined? I think it is required by chatAdapter type?

PorterNan (approved):

packages/communication-react/review/beta/communication-react.api.md
PorterNan: We current don't make adapter method optional - but would be a good discussion point we can have for the team

Currently the definition for adapter is - that is what exactly needed by composite, and all functions there are required to get composite work nicely

For whoever implement adapter directly - that is not what we support for non-breaking change since it is a very advanced method - but they still have way to get over it, people can just implement an empty function to bypass the type check

JamesBurnside: If you want to keep the method optional, we should still remove the `| undefined`, the case of the callwithchat adapter it should either call chatAdapter.downloadAuthenticatedAttachment or throw an error if thats not implemented

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -227,6 +241,8 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
             onRenderMessage={onRenderMessage}
             /* @conditional-compile-remove(file-sharing) */
             onRenderFileDownloads={onRenderFileDownloads}
+            /* @conditional-compile-remove(teams-inline-images) */
+            onFetchAttachments={onRenderInlineAttachment}
```
JamesBurnside, line 245: Discussed this with Josh, I think this should be separate and we should have `onFetchAttachments` and `onFetchInlineImages` as the types are just two different. But consider this non-blocking comment we can iterate on in future PRs

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -194,6 +196,18 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
     [fileSharing?.downloadHandler]
   );
 
+  /* @conditional-compile-remove(teams-inline-images) */
+  const onRenderInlineAttachment = useCallback(
+    async (attachment: FileMetadata): Promise<AttachmentDownloadResult | undefined> => {
+      if (adapter.downloadAuthenticatedAttachment && attachment.previewUrl) {
+        const blob = await adapter.downloadAuthenticatedAttachment(attachment.previewUrl);
```
JamesBurnside, line 203: is this cached anywhere? Otherwise we are going to call download on every time it tries to render the attachment

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1192,7 +1192,7 @@ export interface ChatAdapterSubscribers {
 export interface ChatAdapterThreadManagement {
     deleteMessage(messageId: string): Promise<void>;
     // (undocumented)
-    downloadAuthenticatedAttachment?: (attachmentUrl: string) => Promise<AttachmentDownloadResult>;
+    downloadAuthenticatedAttachment?: (attachmentUrl: string) => Promise<AttachmentDownloadResult | undefined>;
```
PorterNan, line 1195: I like the e2e implementation a lot, this api part is the only one I have concerns 

The attachmentURL seems not related to any part of chat - if you give me a random url, I can also download it for you, and it's hard for some other contoso developers to understand what is attachmentURL, and where to get it 

What I am thinking is maybe we could name it
downloadAttachment(messageId: string, index?: number) => Promise<AttachmentDownloadResult | undefined>; (they put multiple blobsUrl in the result)

Then the function itself could figure out what to download

I've still approved the PR since you are bug bashing it, let's see if we could make this happen if we haven't done ARB yet, or we could mark this as an item for GA

PorterNan: And for naming - it feels it gets a url then return a blobUrl, so it could be just downloadAttachment no matter it is authenticated or not, just because we are now only using it for authenticatedAttachment doesn't mean it is only designed for non-authenticated one (maybe an option could be involved in the future)

packages/chat-component-bindings/src/messageThreadSelector.ts
JoshuaLai: LOL yes. This PR was gluing the adapter and the composite pieces together

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
JoshuaLai: Yes this is a good call out. Right now in the `MessageThread.tsx` we have a check that a specific `attachment.id` isn't already there, if it is already there we don't other updating the state so it shouldn't kick off a new re-render, so it also shouldn't call download every time either. 
We have plans to make this logic a little smarter too. Coming soon...
```js
  const [inlineAttachments, setInlineAttachments] = useState<Record<string, string>>({});
  const onFetchInlineAttachment = useCallback(
    async (attachment: FileMetadata): Promise<void> => {
      if (!onFetchAttachments || attachment.id in inlineAttachments) {
        return;
      }
      const attachmentDownloadResult = await onFetchAttachments(attachment);
      if (attachmentDownloadResult) {
        setInlineAttachments((prev) => ({ ...prev, [attachment.id]: attachmentDownloadResult.blobUrl }));
      }
    },
    [inlineAttachments, onFetchAttachments]
  );
```

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
JoshuaLai: Looking at some of the comments above I think we will remove the undefined and we will just throw an error instead, like what James recommended and say something like `method not implemented`

PorterNan: If we are throwing error like this, why don't we just mark the download method required in adapter type, it's actually required (because we are throwing errors but we just pretend it is not)

Again, overriding/creating own adapter is not the scenario we support for "non-breaking changes", cause it is not ACS related anymore

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -800,7 +800,7 @@ export interface CallWithChatAdapterManagement {
     deleteMessage(messageId: string): Promise<void>;
     disposeStreamView(remoteUserId?: string, options?: VideoStreamOptions): Promise<void>;
     // (undocumented)
-    downloadAuthenticatedAttachment?: (attachmentUrl: string) => Promise<AttachmentDownloadResult>;
+    downloadAuthenticatedAttachment: (attachmentUrl: string) => Promise<AttachmentDownloadResult>;
```
JoshuaLai, line 803:
```suggestion
    downloadAttachments: (options: { attachmentUrls: string[] }) => Promise<AttachmentDownloadResult>;
```
Discussed with @PorterNan  will update to this shape.

JoshuaLai (approved):

PR #2920, opened 2023-04-12
Captions banner should not show up when on local hold

carocao-msft:
# What
Captions banner should not show up when on local hold

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3217743

# How Tested
Tested with samples 

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/common/CaptionsBanner.tsx
+++ b/packages/react-composites/src/composites/common/CaptionsBanner.tsx
@@ -26,7 +26,7 @@ const mobileViewBannerWidth = '90%';
 const desktopViewBannerWidth = '50%';
 
 /** @private */
-export const CaptionsBanner = (props: { isMobile: boolean; disableBannerForHoldScreen?: boolean }): JSX.Element => {
+export const CaptionsBanner = (props: { isMobile: boolean; hideBanner?: boolean }): JSX.Element => {
   /* @conditional-compile-remove(close-captions) */
   const captionsBannerProps = useAdaptedSelector(_captionsBannerSelector);
   /* @conditional-compile-remove(close-captions) */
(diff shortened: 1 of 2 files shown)


packages/react-composites/src/composites/common/CaptionsBanner.tsx
```diff
@@ -64,7 +64,8 @@ export const CaptionsBanner = (props: { isMobile: boolean }): JSX.Element => {
       }
       {
         /* @conditional-compile-remove(close-captions) */ captionsBannerProps.captions.length > 0 &&
-          captionsBannerProps.isCaptionsOn && (
+          captionsBannerProps.isCaptionsOn &&
+          !props.disableBannerForHoldScreen && (
```
alkwa-msft, line 68: nit.  Double negatives can trip anyone up.

Can we change the prop to be showBanner and then we can pass false into it if localHold is true
then we can more easily say  isCaptionsOn && showBanner && (....

alkwa-msft (approved):

packages/react-composites/src/composites/common/CaptionsBanner.tsx
PorterNan: Agree, we can have a property named "isOnHold" to disable the captions banner, or more straightforward, showBanner

PorterNan (approved):

packages/react-composites/src/composites/common/CaptionsBanner.tsx
carocao-msft: disableBannerForHoldScreen is consistent with the name disableButtonsForHoldScreen which we pass into the control bar for a similar purpose. 
Should we keep this naming pattern or just pass in something like showBanner/ isOnHold?

carocao-msft: Changed name to hideBanner, makes more sense than disablebanner since we are hiding the entire UI.
It's still a double negative but easier to understand compared to disableBanner, also in this case when we conditionally compiled this line out for PSTN, the prop will be of value 'defined' which equals to false. (We don't need to have extra logic to set default value)

PR #2921, opened 2023-04-12
Fix captions ui bug

carocao-msft:
# What
Fix captions ui bug

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3217733
https://skype.visualstudio.com/SPOOL/_workitems/edit/3217735
https://skype.visualstudio.com/SPOOL/_workitems/edit/3217737
https://skype.visualstudio.com/SPOOL/_workitems/edit/3217740

# How Tested
Tested on samples
Screenshot 2023-04-12 at 1 46 45 PM
Screenshot 2023-04-12 at 1 54 43 PM
image

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/styles/Captions.style.ts
+++ b/packages/react-components/src/components/styles/Captions.style.ts
@@ -42,6 +42,21 @@ export const captionContainerClassName = mergeStyles({
  */
 export const captionsBannerClassName = mergeStyles({
   height: _pxToRem(100),
-  overflowY: 'scroll',
+  overflowY: 'auto',
   overflowX: 'hidden'
+});
+
+/**
+ * @private
+ */
+export const captionsContentContainerClassName = mergeStyles({
+  minWidth: 0
+});
+
+/**
+ * @private
+ */
+export const displayNameContainerClassName = mergeStyles({
+  overflow: 'hidden',
+  textOverflow: 'ellipsis'
 });
--- a/packages/react-components/src/components/styles/Captions.style.ts
+++ b/packages/react-components/src/components/styles/Captions.style.ts
@@ -42,5 +42,6 @@ export const captionContainerClassName = mergeStyles({
  */
 export const captionsBannerClassName = mergeStyles({
   height: _pxToRem(100),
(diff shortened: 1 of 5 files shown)


alkwa-msft (approved):

packages/react-components/src/components/styles/Captions.style.ts
```diff
@@ -42,5 +42,6 @@ export const captionContainerClassName = mergeStyles({
  */
 export const captionsBannerClassName = mergeStyles({
   height: _pxToRem(100),
-  overflow: 'scroll'
+  overflowY: 'scroll',
```
alkwa-msft, line 45: We can probably simplify to 

overflow: 'hidden scroll'

image

PorterNan, line 45: This shouldn't be scroll otherwise the scroll bar will always shows even there is only one caption

```suggestion
  overflowY: 'auto',
```

PorterNan (commented):
Tried this a bit:
the caption overflow doesn't behaves right when the displayName becomes too long:
image

I guess we might wanna apply this style to the name:
https://stackoverflow.com/questions/17868503/on-css-if-text-line-is-break-show-dots

it works fine when the name is in a normal length

PorterNan (commented):
image

Caption settings still shows double scroll bar

PorterNan:
> image
> 
> Caption settings still shows double scroll bar

Updates, pushed a fix for this double scroll bar problem

packages/react-components/src/components/styles/Captions.style.ts
carocao-msft: Screenshot 2023-04-13 at 9 56 13 AM
Tried it and looks like IStyle doesn't support 'hidden scroll'

alkwa-msft: cool then lets keep it separate

PorterNan:
NICE!
image

PorterNan (approved):

PR #2923, opened 2023-04-12
Improve checking user agent in e2e tests

JamesBurnside:
# What
Update user agent version to be high enough it should not be outdated.

# Why
Recently our e2e test suddenly failed because the user agent string became too old a version. 

# How Tested
CI should pass


packages/react-composites/playwright.config.common.ts
```diff
@@ -52,6 +52,10 @@ const LOCAL_REPORTERS: ReporterDescription[] = [['list']];
 const SECOND = 1000;
 const MINUTE = 60 * SECOND;
 
+// Using chromium useragent with a very high version to avoid breaking the unsupportedBrowser page
+const DESKTOP_USER_AGENT = 'Windows Chrome/999.0.0.0';
```
PorterNan, line 56: Do we have a single hermetic test for this updated warning page? like set the expired browser version then test if the page shows up

PorterNan (approved):

packages/react-composites/playwright.config.common.ts
JamesBurnside: Yep we do! https://github.com/Azure/communication-ui-library/blob/5cae388f3bd54c64cbc999696f42fc474fa0dc31/packages/react-composites/tests/browser/call/hermetic/UnsupportedBrowserPage.test.ts

alkwa-msft (approved):

PR #2931, opened 2023-04-13
Fix captions dropdown

PorterNan:
# What

Fix captions dropdown dismiss when new captions come

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/CaptionsSettingsModal.tsx
+++ b/packages/react-components/src/components/CaptionsSettingsModal.tsx
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
-import React from 'react';
+import React, { useCallback } from 'react';
 import { useMemo, useState } from 'react';
 import {
   IModalStyles,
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/CaptionsSettingsModal.tsx
```diff
@@ -82,13 +83,14 @@ export const _CaptionsSettingsModal = (props: _CaptionsSettingsModalProps): JSX.
     text: currentSpokenLanguage !== '' ? currentSpokenLanguage : defaultSpokenLanguage
   });
 
-  const dismiss = (): void => {
+  const onDismiss = useCallback((): void => {
```
carocao-msft, line 86: We should keep the name as dismiss

packages/react-components/src/components/CaptionsSettingsModal.tsx
```diff
@@ -82,13 +83,14 @@ export const _CaptionsSettingsModal = (props: _CaptionsSettingsModalProps): JSX.
...
     if (onDismissCaptionsSettings) {
       onDismissCaptionsSettings();
     }
-  };
+  }, [onDismissCaptionsSettings]);
 
-  const confirm = async (languageCode: string): Promise<void> => {
+  const onConfirm = useCallback(async (): Promise<void> => {
```
carocao-msft, line 92: We should keep the name as confirm

carocao-msft (approved):

PR #2935, opened 2023-04-14
Move off `Enzyme`, to `React Testing Library`

JamesBurnside:
# What
Migrate from Enzyme to react testing library

# Why
See: [Enzyme is dead](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl). Enzyme has no, and is unlikely to have, any react 18 support. So we are migrating to a new testing package.

related: https://github.com/Azure/communication-ui-library/discussions/2939, https://github.com/Azure/communication-ui-library/issues/1900

# How Tested
Unit tests in CI should all pass


packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.test.tsx
```diff
@@ -59,19 +65,10 @@ describe('CallWithChatComposite', () => {
...
-    expect(callComposites.length).toBe(1);
-
-    const callComposite = callComposites.first();
-    expect(callComposite.props().fluentTheme).toBe(mockBaseCompositeProps.fluentTheme);
-    expect(callComposite.props().icons).toBe(mockBaseCompositeProps.icons);
-    expect(callComposite.props().locale).toBe(mockBaseCompositeProps.locale);
-    expect(callComposite.props().rtl).toBe(mockBaseCompositeProps.rtl);
-    expect(callComposite.props().onFetchAvatarPersonaData).toBe(mockBaseCompositeProps.onFetchAvatarPersonaData);
-    expect(callComposite.props().onFetchParticipantMenuItems).toBe(mockBaseCompositeProps.onFetchParticipantMenuItems);
+    expect(container.querySelector('[data-ui-id="lobby-page"]')).toBeTruthy();
```
JamesBurnside, line 72: this is the only test I changed _what_ was tested. Before the test simply checked that the props we passed in were the props the composite received -- this doesn't make sense for a test, so I updated this to validate the composite successfully rendered a lobby page as the mock adapter depicts

packages/react-components/src/components/DevicesButton.test.tsx
```diff
@@ -33,17 +30,17 @@ const mockProps: DevicesButtonProps = {
...
-    expect(component.find('button').text()).toBe(testLocale.strings.devicesButton.label);
+    renderWithLocalization(<DevicesButton showLabel={true} {...mockProps} />, testLocale);
+    expect(screen.getByRole('button').textContent).toBe(testLocale.strings.devicesButton.label);
   });
 
   test('Should override button label with `strings` prop', async () => {
     const testLocale = createTestLocale({ devicesButton: { label: Math.random().toString() } });
     const devicesButtonStrings = { label: Math.random().toString() };
-    const component = mountWithLocalization(
+    renderWithLocalization(
```
dmceachernmsft, line 40: so with this change does this mean these tests do actually create a DOM?

packages/react-components/src/components/Dialpad/Dialpad.test.tsx
```diff
@@ -33,39 +42,33 @@ describe('Dialpad tests', () => {
     const testLocale = createTestLocale({
       dialpad: dialpadStrings
     });
-    const component = mountWithLocalization(<Dialpad />, testLocale);
-    expect(component.find('[data-test-id="dialpad-input"]').first().props().placeholder).toBe(
-      testLocale.strings.dialpad.placeholderText
-    );
+    renderWithLocalization(<Dialpad />, testLocale);
+    expect(screen.getByRole('textbox').getAttribute('placeholder')).toBe(testLocale.strings.dialpad.placeholderText);
```
dmceachernmsft, line 46: is `screen` a property that just exists inside the test runner we can query?

dmceachernmsft (approved):
looks great just some general fact finding questions! 😊

packages/react-components/src/components/Dialpad/Dialpad.test.tsx
JamesBurnside: Yeah, its basically a way of saying - hey query what's on the "screen" (screen isn't a real screen, just an in memory interpretation of a screen). Ideally we'd only use screen for queries but it can't query anything arbitrary (like data-ui-id) so for those I've used the `container` returned by the render method

packages/react-components/src/components/VideoTile.test.tsx
```diff
@@ -26,22 +24,33 @@ describe('VideoTile', () => {
...
+    expect(videoTile).toBeTruthy();
+
+    jest.useFakeTimers();
     await act(async () => {
-      wrapper.simulate('touchstart');
-      await new Promise((resolve) => setTimeout(resolve, 1000));
-      wrapper.simulate('touchend');
+      fireEvent.touchStart(videoTile);
+      jest.runAllTimers();
+      fireEvent.touchEnd(videoTile);
```
mgamis-msft, line 35: Thanks for this conversion. I was worried about triggering touch events with React Testing Library. Nicely done.

packages/react-components/src/components/Dialpad/Dialpad.test.tsx
```diff
@@ -1,14 +1,13 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
mgamis-msft, line 3: Just curious why this was necessary

packages/react-components/src/components/utils/testUtils.tsx
```diff
@@ -2,30 +2,51 @@
...
+/** @private */
+export const renderWithLocalization = (
+  node: React.ReactElement,
+  locale: ComponentLocale
+): {
+  rerender: (node: React.ReactElement) => void;
+  container: HTMLElement;
+} => {
+  const { rerender, container } = render(
+    withLiveAnnouncerContext(<LocalizationProvider locale={locale}>{node}</LocalizationProvider>)
```
mgamis-msft, line 25: Why do we have to wrap with LiveAnnouncer?

mgamis-msft (approved):

packages/react-components/src/components/utils/testUtils.tsx
JamesBurnside: Just gets rid of warnings from the `react-aria-live` package. This warning already exists in our current tests, but this removes the warning

packages/react-components/src/components/Dialpad/Dialpad.test.tsx
JamesBurnside: Typescript isn't smart enough to know that:
```
var a: someType | undefined;

expect(a).toBeTruthy();
expect(a.something).toBe(...); // Typescript complains here that a might be undefined, but it actually can't be undefined because of the previous `expect`

// so I instead do:
expect(a).toBeTruthy();
expect(a!.something).toBe(...); // notice the ! after the a, this is the non-null assertion, which by default we disallow as its should be used more sparingly
```

packages/react-components/src/components/utils/testUtils.tsx
JamesBurnside: (we're moving off this package shortly as it does not support react18 so this was a super quick fix to the warning)

PR #2938, opened 2023-04-14
Captions API changes

carocao-msft:
# What
API changes based on ARB

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
+++ b/packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
@@ -266,11 +266,11 @@ export type CaptionsReceivedListener = (event: { captionsInfo: CaptionsInfo }) =
 
 /* @conditional-compile-remove(close-captions) */
 /**
- * Callback for {@link CallAdapterSubscribers} 'captionsPropertyChanged' event.
+ * Callback for {@link CallAdapterSubscribers} 'isCaptionsActiveChanged' event.
  *
  * @beta
  */
-export type CaptionsPropertyChangedEventListener = () => void;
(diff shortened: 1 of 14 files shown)


packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -667,7 +675,7 @@ export interface CallAdapterSubscribers {
   /**
    * Subscribe function for 'captionsPropertyChanged' event.
    */
-  on(event: 'captionsPropertyChanged', listener: PropertyChangedEvent): void;
+  on(event: 'captionsPropertyChanged', listener: CaptionsPropertyChangedEventListener): void;
```
PorterNan, line 678: Maybe just name it like PropertyChangedEventListener?

Or directly do
```suggestion
  on(event: 'captionsPropertyChanged', listener: () => void): void;
```?

Hmmm, () => void is event less characters than PropertyChangedEventListener

PorterNan: Or... we should directly use the same name isCaptionsActiveChanged (is it the right name coming from sdk?)

Then do 

IsCaptionsActiveChangedEventListener = (event: {isActive: boolean} ) => void

 an event listener + an event makes more sense to get new type

PorterNan: just checked what we bubble up from caption feature, the event is isCaptionsActiveChanged

Let's use the same name here and add isActive as event payload 

And don't forget to change AzureCommunicationCallAdapter to update the implementation  la bit

PorterNan (approved):

PR #2948, opened 2023-04-17
Raise Hand POC

rzdor:
# What
This is POC for Raise Hand feature.
I've added implementation for Raise/Lower Hand button to components.
Updated composites and Calling test app to use new element.
Added states and subscribers to hand updates for new published/removed states.

Currently no updates happen on main call screen, but composites already receiving all new published states, so only IU change needed - i will work on it.

image
image

### DIFF
--- a/change-beta/@azure-communication-react-8458eacc-fcc0-4685-923e-6f80fcf76e06.json
+++ b/change-beta/@azure-communication-react-8458eacc-fcc0-4685-923e-6f80fcf76e06.json
@@ -1,5 +1,5 @@
 {
-  "type": "prerelease",
+  "type": "none",
   "comment": "AzureCommunicationCallAdapter emitting transferRequested event and added acceptedTransferCallState propetry to CallAdapterClientState",
   "packageName": "@azure/communication-react",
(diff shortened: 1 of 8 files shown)


JamesBurnside:
This is awesome! ideally we should spit this into two+ PRs, typically we aim for:

1. Stateful client changes
2. React component changes
3. Handlers & Selector changes
4. Composite changes

To get this checked in we'll need to make sure its being conditional compiled for `beta` only builds and not in `stable` builds (see https://github.com/Azure/communication-ui-library/blob/main/docs/references/beta-only-features.md#conditionally-adding-a-feature)

Once the conditional compile flags are added - The stateful client and handlers & selectors look good to check in now!

The react components and composite changes will need some UX input (as you've mentioned in the PR description). We should get design to look at this and create figmas. The composites will need some e2e tests added later on. While that's ongoing it would be great to get the core parts (stateful client and handlers & selectors) checked in and help reduce the PR size.

JamesBurnside:
@rzdor is this being worked on? Please call out what you need to get unblocked here

PR #2950, opened 2023-04-17
Add tool to perform typescript regression checks as part of CI and update documented typescript version

JamesBurnside:
# What
* Add tool (`check-typescript-regression`) that runs our lowest supported version of typescript over our package.
* Update supported typescript version in storybook from `3.7.0` to `3.8.3` as that is the actual lowest version that works (fluent package breaks 3.7)

# Why
We never know if we've broken this, now we do!

# How Tested
Verified that adding types that only newer typescript supports to our public API breaks this build and hence breaks CI.
image


PorterNan (approved):
So the idea of this check is to see if we still support 3.8.3 while other packages is still using 4+

That's a good idea, maybe more compatibility checker should be added if that is possible

.github/workflows/ci.yml
```diff
@@ -106,6 +106,9 @@ jobs:
       # Tree-shaking check
       - name: Treeshaking check
         run: rush build --only @internal/check-treeshaking
+      # Typescript version check
+      - name: Treeshaking check
```
PorterNan, line 110:
```suggestion
      - name: Typescript version compatibility check
```

PorterNan:
How long does it take to do a check here?

.github/workflows/ci.yml
JamesBurnside: updated

JamesBurnside:
> How long does it take to do a check here?

As much as `4seconds` locally

JamesBurnside:
> So the idea of this check is to see if we still support 3.8.3 while other packages is still using 4+
> 
> That's a good idea, maybe more compatibility checker should be added if that is possible

Yep! I was thinking the same, things like webpack, react etc. would also be useful

tools/check-typescript-regression/README.md
```diff
@@ -0,0 +1,7 @@
+# Typescript Regression Checker
+
+This tool tries to run `tsc` over the npm package, to ensure that the typescript definitions are supported in our minimum typescript version.
+
+For example, `typescript` may add new syntax that is not supported in our minimum typescript version and we may expose that in the `.d.ts` type file of our npm package. This tool will catch that.
+
+If this project fails, it should only be because typescript definitions were added that are not compatible with our minimum typescript version and you likely need to use a older typescript syntax.
```
edwardlee-msft, line 7: super little nit:
likely need to use **an** older typescript syntax.

edwardlee-msft (approved):

PR #2952, opened 2023-04-18
[Chat][Bugfix] Fix duplicate onFetchAttachment requests

jimchou-dev:
# What

- Fix duplicate onFetchAttachment requests
- Change onFetchAttachments to be in useEffect
- Update unit test for 2 images

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3230748

# How Tested

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -72,6 +74,16 @@ const MessageContentWithLiveAria = (props: MessageContentWithLiveAriaProps): JSX
...
   const liveAuthor = _formatString(props.strings.liveAuthorIntro, { author: `${props.message.senderDisplayName}` });
+
+  /* @conditional-compile-remove(teams-inline-images) */
+  useEffect(() => {
+    props.message.attachedFilesMetadata?.map((fileMetadata) => {
+      if (props.onFetchAttachment && props.attachmentsMap && props.attachmentsMap[fileMetadata.id] === undefined) {
+        props.onFetchAttachment(fileMetadata);
+      }
+    });
+  });
```
jimchou-dev, line 85: @PorterNan not sure if I need to add a empty dependency array for the useEffect

JoshuaLai (approved):
Tested this fix during bug bash and was working as expected

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -72,6 +74,16 @@ const MessageContentWithLiveAria = (props: MessageContentWithLiveAriaProps): JSX
 
 const MessageContentAsRichTextHTML = (props: ChatMessageContentProps): JSX.Element => {
   const liveAuthor = _formatString(props.strings.liveAuthorIntro, { author: `${props.message.senderDisplayName}` });
+
+  /* @conditional-compile-remove(teams-inline-images) */
+  useEffect(() => {
```
PorterNan, line 79: Add a dependency array to this useEffect, since we don't wanna trigger it everytime when we re-render

https://codedamn.com/news/reactjs/useeffect-dependency#:~:text=component%20gets%20updated.-,useEffect%20Dependency%20array,-The%20useEffect%20manages

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -847,6 +847,7 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
       if (!onFetchAttachments || attachment.id in inlineAttachments) {
         return;
       }
+      setInlineAttachments((prev) => ({ ...prev, [attachment.id]: '' }));
```
PorterNan, line 850: Aha, setting it to empty string is a cool idea, should put a comment here to explain what it is doing

BTW, will empty string be rendered into the img element? Is placeholder still working after you set it to empty string?

PorterNan (approved):
Left one comment and one question this fix, once you resolve them, feel free to check it in

And cool idea to set the value to ""!

packages/react-components/src/components/MessageThread.tsx
jimchou-dev: I believe Josh synced with you offline on this

jpeng-ms (approved):

PR #2954, opened 2023-04-19
Fix captions console error cannot set spoken language because captions has not started 

carocao-msft:
# What
Fix captions console error cannot set spoken language because captions has not started 

# Why
Screenshot 2023-04-18 at 11 48 23 AM

# How Tested
Tested with samples 

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/CaptionsSettingsModal.tsx
+++ b/packages/react-components/src/components/CaptionsSettingsModal.tsx
@@ -95,10 +95,9 @@ export const _CaptionsSettingsModal = (props: _CaptionsSettingsModalProps): JSX.
     if (isCaptionsFeatureActive) {
       onSetSpokenLanguage(languageCode);
     } else {
-      await onStartCaptions({ spokenLanguage: languageCode });
       // set spoken language when start captions with a spoken language specified.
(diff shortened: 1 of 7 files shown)


packages/react-components/src/components/CaptionsSettingsModal.tsx
```diff
@@ -95,10 +95,9 @@ export const _CaptionsSettingsModal = (props: _CaptionsSettingsModalProps): JSX.
     if (isCaptionsFeatureActive) {
       onSetSpokenLanguage(languageCode);
     } else {
-      await onStartCaptions({ spokenLanguage: languageCode });
```
PorterNan, line 98: What is the difference between await and .then()?

Ideally it should be the same
```
foo: ()=> Promise<boolean>;
await foo();
bar();
```

equals 
```
foo: ()=> Promise<boolean>;
foo().then(() => bar(););

```

PorterNan: The only difference here is Promise.resolvev way will ignore the error and not throw it, if I remember it right

carocao-msft: The original implementation with just await throws the above console error. I haven't tried await and then together, I can try that.

PR #2957, opened 2023-04-19
Update steps to fetch strings in release doc

mgamis-msft:
# What

Update release doc to make it simpler it to fetch strings from TD Build and ensure the strings updates are not done on the release branch which never make it back to main.

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3235200

# How Tested

I will work with the next person performing the next release and monitor how effective or ineffective these updates will be.

# Process & policy checklist


.github/workflows/run-td-build.yml
```diff
@@ -6,10 +6,6 @@ on:
   push:
     branches:
       - main
```
dmceachernmsft, line 8: there a reason for this yml edit? should we not be fetching strings from these other branches?

docs/releases/creating-a-release.md
```diff
@@ -13,6 +13,11 @@ _This document applies to beta and stable releases. Alpha releases are created n
 
 _This document applies to normal releases, off of `main`. For hotfixing a prior release on NPM, see [documentation on creating hotfixes](./hotfix-release.md)._
 
+## Before creating a release branch
+
+- Update the strings on `main` these [steps](../references/string-translations.md).
+  - **Note**: String translation can take up to 5 working days to complete. It is a good idea to avoid string changes once before the release process
```
dmceachernmsft, line 19: I thought the turn around was 10?

mgamis-msft: No it's 5 :)

.github/workflows/run-td-build.yml
mgamis-msft: I think all string changes should be done on main and be cherry-picked into the release branch. It's simpler.
String changes on release branches and the corresponding string translations don't ever make it back into main.

TD looks at our English strings on every push and has translations for each of them in their cache. If we have differing English strings in main vs a release branch, the translations in the cache will be on the latest push. I would like to have the translations always be for the English strings on main.

JamesBurnside (approved):

dmceachernmsft (approved):

PR #2959, opened 2023-04-20
Stabilize the scrolling behavior

PorterNan:
# What

Stabilize the scrolling behavior

- Add timestamp+displayName as key to each caption element so react can always find the same element and render into those
- Add overflow-anchor:auto to each element, so any on screen caption will be an anchor for scrolling position when adding/removing elements

Demo:
In the video, we scrolled into an element in the middle of captions list, and you can see it is anchored when we adding new element at the bottom and removing element from the top (yes we have more than 50 elements)
And it is anchored all the way to the top, until the anchor element is popping out of the stack of all captions.

https://user-images.githubusercontent.com/11863655/233261350-24e4235e-c87d-4d33-9cf7-31f8e7291480.mp4

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/styles/Captions.style.ts
```diff
@@ -35,7 +35,7 @@ export const captionClassName = mergeStyles({
 export const captionContainerClassName = mergeStyles({
   marginTop: _pxToRem(6),
   marginBottom: _pxToRem(6),
-  overflowAnchor: 'none'
+  overflowAnchor: 'auto'
```
dmceachernmsft, line 38: Does this then track the message by the key and stops the overflow from scrolling when there is a new caption?

dmceachernmsft (approved):

packages/react-components/src/components/styles/Captions.style.ts
PorterNan: Right! It tracks messages by the dom element itself, and the key is to ensure react keeps all dom elements to be same instances, instead of re-creating them again and again, so that browser can track them well

PorterNan (commented):
Because this PR also changes how we deserialize json, so Date type is converted correctly in error bar - we need to update error bar state to make sure it is always right in this case

carocao-msft (approved):

PR #2965, opened 2023-04-21
[Draft] Transfer call spike

mgamis-msft:
# What

Can test adhoc calling test Teams users and then transfer to another Teams users.
Can test calling a Call Queue as an ACS/CTE user and see that a Teams user set to answer for Call queue will be transferred to ACS/CTE user


JamesBurnside:
`James the Manual Stale PR Bot:` Closing old draft PR that has not seen activity in over 1+month. Please reopen as necessary.

PR #2968, opened 2023-04-21
Fix for tooltip getting hidden under Effects title in video effects pane

prabhjot-msft:
# What
Fix for tooltip getting hidden under Effects title in video effects pane
Before:
bug

After:
Screenshot 2023-04-21 at 1 15 47 PM

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3217334

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/common/VideoEffectsPane.tsx
```diff
@@ -134,6 +134,11 @@ const VideoEffectsPaneTrampoline = (
   const locale = useLocale();
   /* @conditional-compile-remove(video-background-effects) */
   const selectedEffect = useSelector(activeVideoBackgroundEffectSelector);
+  /* @conditional-compile-remove(video-background-effects) */
+  const headerStyles = {
+    zIndex: 0
```
JamesBurnside, line 139: the default zindex for any element is `0`, what does setting this to `0` do?

JamesBurnside: Oh this is the panel's z-index, seems fine to do this though I will be removing use of the panel when the converged panel work is complete

JamesBurnside (approved):

dmceachernmsft (approved):

packages/react-composites/src/composites/common/VideoEffectsPane.tsx
prabhjot-msft: Yes, the panel has Z-index:1 so by setting z-index for header to 0, tooltip can overlay the header.

PR #2969, opened 2023-04-21
[bugfix widget] Remove side pane button container when neither button enabled

dmceachernmsft:
# What

removes container from control bar if both people and chat buttons are hidden in common controls.
# Why

image
image

# How Tested

Tested locally with widget POC app made change here too.


packages/react-composites/src/composites/common/ControlBar/CommonCallControlBar.tsx
```diff
@@ -360,7 +360,7 @@ export const CommonCallControlBar = (props: CommonCallControlBarProps & Containe
             </Stack>
           </CallAdapterProvider>
         </Stack.Item>
-        {!props.mobileView && (
+        {!props.mobileView && (isEnabled(options.peopleButton) || isEnabled(options.chatButton)) && (
```
JamesBurnside, line 363: we support injecting buttons here, what if these buttons are not enabled but a custom button is injected?

dmceachernmsft: good point, will account for that in the logic

packages/react-composites/src/composites/common/ControlBar/CommonCallControlBar.tsx
```diff
@@ -377,41 +377,45 @@ export const CommonCallControlBar = (props: CommonCallControlBarProps & Containe
...
-                    styles={commonButtonStyles}
-                  />
-                )}
-                {isEnabled(options?.chatButton) && chatButton}
-              </Stack>
-            </div>
-          </Stack.Item>
-        )}
+        {!props.mobileView &&
+          (isEnabled(options.peopleButton) ||
```
JamesBurnside, line 381: consider making this a bool above to reduce the amount to parse here
```
const secondaryButtonsExist = isEnabled(options.peopleButton) || isEnabled(options.chatButton) || customButtons['secondary'] !== undefined
```

packages/react-composites/src/composites/common/ControlBar/CommonCallControlBar.tsx
```diff
@@ -377,41 +377,45 @@ export const CommonCallControlBar = (props: CommonCallControlBarProps & Containe
...
+            isEnabled(options.chatButton) ||
+            /* @conditional-compile-remove(control-bar-button-injection) */ customButtons['secondary'] !==
```
JamesBurnside, line 383: remove conditional compile, callwithchat uses this

JamesBurnside (approved):

PR #2976, opened 2023-04-24
Converged Side Pane & Control Bar

JamesBurnside:
# What

- CallComposite SidePaneProvider provides access to sidepane state
- CallCompsoite MainPage is maintaining the state of the sidepane and passing that into the context
- CallComposite Custom button injection code is used by CallWithChatComposite to inject custom chat button
- CallComposite Video Effects uses converged side pane and not fluent panel
- **Internal** CallComposite interfaces added (and used by call with chat composite) to inject side pane content and inject chat button

# Why
Remove much much duplicate usage of components and remove that call with chat composite had its own control bar and its own side pane.

This also fixed several bugs in the callcompsites new control bar:
- rtl 
- being on hold disabling other buttons
- dialpad button not showing
- incorrect padding (separate PR #2978)

# How Tested

Lots across calling and callwithchat, stab and beta, mobile and landscape - but will also organize a bug bash on this as this impacts stable builds.


packages/react-composites/src/composites/common/TabHeader.tsx
```diff
@@ -92,7 +95,7 @@ export const TabHeader = (props: TabHeaderProps): JSX.Element => {
             role={'tab'}
             disabled={props.disablePeopleButton}
           >
-            {strings.peopleButtonLabel}
+            {strings.peopleButtonLabel ?? strings.peopleButtonLabel}
```
JamesBurnside, line 98: rev

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -191,12 +191,39 @@ type MainScreenProps = {
   options?: CallCompositeOptions;
   /* @conditional-compile-remove(rooms) */
   roleHint?: Role;
+  overrideSidePane?: InjectedSidePaneProps;
+  onSidePaneIdChange?: (sidePaneId: string | undefined) => void;
+  mobileChatTabHeader?: MobileChatSidePaneTabHeaderProps;
+};
+
+const isShowing = (overrideSidePane?: InjectedSidePaneProps): boolean => {
```
JamesBurnside, line 199: move to hook

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -371,118 +421,21 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
             grow
             styles={callCompositeContainerStyles(mobileView)}
             // Perf: Instead of removing the video gallery from DOM, we hide it to prevent re-renders.
-            style={callCompositeContainerCSS}
+            // style={callCompositeContainerCSS}
```
JamesBurnside, line 424: remove

packages/react-composites/src/composites/CallComposite/components/SidePane/SidePaneProvider.tsx
```diff
@@ -0,0 +1,136 @@
...
+  children: React.ReactNode;
+};
+
+/**
+ * Provider to provide localized strings for this library's composites.
+ *
+ * @private
+ */
+export const SidePaneProvider = (props: SidePaneProviderProps): JSX.Element => {
+  const [headerRenderer, setHeaderRenderer] = React.useState<(() => JSX.Element) | undefined>();
```
PorterNan, line 66: Is it possible to avoid adding global changing state into provider, it's a huge review push back to do so long time ago - it's another format of global state - our centralized state not single source of truth of global anymore. The conclusion of the discussion was we could use Provider for global static, but we shouldn't put any dynamic changing state there for either debuggability or code structure. It feels this is opening a new pattern that brings us back to that design

if we feel that is a need to add these states into adapter, I will 100% vote for that

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/components/SidePane/SidePane.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const renderingHiddenOverrideContent = overrideSidePane?.hidden && overrideSidePane?.contentRenderer;
+  const renderingOnlyHiddenContent = renderingHiddenOverrideContent && !activeSidePaneId;
+
+  const paneStyles = renderingOnlyHiddenContent
+    ? hiddenStyles
+    : props.mobileView
+    ? availableSpaceStyles
+    : sidePaneStyles;
+
+  let Header =
```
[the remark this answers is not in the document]
JamesBurnside: It can be very much simplified once we remove the two tabs!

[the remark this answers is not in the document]
JamesBurnside: (instead of a renderer, it can just be `heading` and `onClose`

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -139,21 +163,12 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
...
-    (showVideoEffectsOptions: boolean): void => {
-      setVideoEffectsPane(showVideoEffectsOptions);
-    },
-    [setVideoEffectsPane]
-  );
+  const { toggleVideoEffectsPane } = useVideoEffectsPane(props.mobileView);
 
   const [showDrawer, setShowDrawer] = useState(false);
   const onMoreButtonClicked = useCallback(() => {
-    closePane();
```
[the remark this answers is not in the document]
JamesBurnside: Chatted with @dmceachernmsft , this behavior (because of a regression in main branch) does not happen today so I will not fix in this PR, but will do a subsequent PR to fix this.

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -224,124 +213,124 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
...
     }),
-    [callWithChatStrings]
+    [props.callControls]
   );
 
-  /* @conditional-compile-remove(PSTN-calls) */
-  const alternateCallerId = callAdapter.getState().alternateCallerId;
-  /* @conditional-compile-remove(close-captions) */
-  const isTeamsCall = callAdapter.getState().isTeamsCall;
+  const injectedCustomButtonsTrampoline = useMemo(() => {
```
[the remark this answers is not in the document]
JamesBurnside: updated

packages/react-composites/src/composites/CallComposite/components/SidePane/SidePane.tsx
```diff
@@ -0,0 +1,94 @@
...
+    return <EmptyElement />;
+  }
+
+  return (
+    <Stack verticalFill grow styles={paneStyles} data-ui-id="SidePane" tokens={props.mobileView ? {} : sidePaneTokens}>
+      <Header />
+      <Stack.Item verticalFill grow styles={paneBodyContainer}>
+        <Stack horizontal styles={scrollableContainer}>
+          {ContentRender && (
+            <Stack.Item verticalFill styles={scrollableContainerContents}>
```
prabhjot-msft, line 77: Should we maintain a single content variable and it should hold the value of latest content. Is there a purpose of maintaining content and OverrideContent?

packages/react-composites/src/composites/CallComposite/components/SidePane/SidePaneProvider.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const [contentRenderer, setContentRenderer] = React.useState<(() => JSX.Element) | undefined>();
+  const [activeSidePaneId, setActiveSidePaneId] = React.useState<string>();
```
PorterNan, line 68: Discussion from last time - we could still use provider, but just not to add state here
So this provider could take props(either states or normal props) from outside of the composite, then store it here

Or we just remove provider then use a top to down state from composite to each single component

prabhjot-msft (approved):

packages/react-composites/src/composites/CallComposite/components/SidePane/SidePaneProvider.tsx
```diff
@@ -0,0 +1,136 @@
...
+/** @private */
+export type InjectedSidePaneProps =
+  | undefined
+  | {
+      headerRenderer?: () => JSX.Element;
+      contentRenderer?: () => JSX.Element;
+      sidePaneId: string;
+      // Useful to ensure the side pane renders the content of the override even if the side pane is closed.
+      // This avoids remounting the content when the side pane is opened again.
+      hidden?: boolean;
```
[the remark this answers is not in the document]
JamesBurnside: 100% agree, updated to better names (`isActive` and `persistRenderingWhenClosed`):
image

packages/react-composites/src/composites/CallComposite/components/SidePane/SidePaneProvider.tsx
JamesBurnside: Updated for provider to take props for now

PR #2980, opened 2023-04-25
Add video effects warning when camera is turned off

JamesBurnside:
# What
Add video effects warning when camera is turned off

# How Tested
image

e2e test to come in future pr


packages/react-composites/src/composites/common/VideoEffectsPane.tsx
```diff
@@ -135,6 +137,10 @@ const VideoEffectsPaneTrampoline = (
   /* @conditional-compile-remove(video-background-effects) */
   const selectedEffect = useSelector(activeVideoBackgroundEffectSelector);
   /* @conditional-compile-remove(video-background-effects) */
+  const isCameraOn = useSelector(localVideoSelector).isAvailable;
+  /* @conditional-compile-remove(video-background-effects) */
+  const showWarning = !isCameraOn && selectedEffect !== 'none';
```
prabhjot-msft, line 142: Show warning would always be true when user's camera is turned off and effect is selected. After seeing that warning, if user closes and opens pane again but does not select/change another effect, the warning will still be there. Is it the desired behaviour or the warning should only be shown only when the effect is selected/changed.

prabhjot-msft (approved):

packages/react-composites/src/composites/common/VideoEffectsPane.tsx
JamesBurnside: Chatted with Alexp and we're going to persist the warning on close&reopen for now -we'll play around with it in selfhost though and open to changing if we feel the behavior isn't right

PR #2982, opened 2023-04-25
Add support for mention feature

emlynmac:
# What
Adds support for @mention tagging in the chat components

# Why
Customer request to add this functionality.

# How Tested
Storybook; ad-hoc

# Process & policy checklist


packages/storybook/stories/MessageThread/snippets/MessageWithCustomMentionRenderer.snippet.tsx
```diff
@@ -0,0 +1,62 @@
...
+          },
+          lookupOptions: {
+            onQueryUpdated: async (query) => {
+              return Promise.resolve(
+                [
+                  {
+                    id: user1Id,
+                    displayText: 'Kat Larsson'
+                  }
+                ].filter((suggestion) =>
```
JoshuaLai, line 52: Should this snippet contain both users? 
```suggestion
            onQueryUpdated: async (query) => {
              return Promise.resolve(
                [
                  {
                    id: user1Id,
                    displayText: 'Kat Larsson'
                  }, 
                  {
                    id: user2Id,
                    displayText: 'Robert Tolbert'
                  }
                ].filter((suggestion) =>
```

packages/react-components/review/beta/react-components.api.md
```diff
@@ -1251,6 +1220,55 @@ export const _LocalVideoTile: React_2.MemoExoticComponent<(props: {
...
+
+// @internal
+export interface _MentionPopoverProps {
+    activeSuggestionIndex?: number;
+    location?: 'above' | 'below';
+    onDismiss?: () => void;
+    onRenderSuggestionItem?: (suggestion: Mention, onSuggestionSelected: (suggestion: Mention) => void, isActive: boolean) => JSX.Element;
+    onSuggestionSelected: (suggestion: Mention) => void;
+    suggestions: Mention[];
+    target: React_2.RefObject<Element>;
```
JoshuaLai, line 1259: `React_2.RefObject<Element>;` Whats this type?

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
```diff
@@ -66,7 +73,7 @@ export const ChatMessageComponentAsEditBox = (props: ChatMessageComponentAsEditB
   }, []);
 
   const setText = (
-    event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+    event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement> | null,
```
JoshuaLai, line 76: Are we allowed to assign `null` now? Can you explain a little bit more why this case is needed?

packages/storybook/stories/INTERNAL/Chat/InputBox.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+        <Stack
+          className={mergeStyles(
+            borderAndBoxShadowStyle({
+              theme,
+              hasErrorMessage: false,
+              disabled: false
+            })
+          )}
+        >
+          <InputBoxComponent
```
JoshuaLai, line 68: I'm seeing an error here where the property `children` is not being assigned

packages/react-components/review/beta/react-components.api.md
```diff
@@ -1251,6 +1220,55 @@ export const _LocalVideoTile: React_2.MemoExoticComponent<(props: {
...
 
+// @beta
+export interface Mention {
+    displayText: string;
+    icon?: JSX.Element;
+    id: string;
+}
+
+// @beta
+export interface MentionDisplayOptions {
```
JoshuaLai, line 1231: Are we not giving developers the ability to change the highlighted name's colour or font styling at all?

packages/storybook/stories/MessageThread/snippets/MessageWithCustomMentionRenderer.snippet.tsx
emlynmac: The code is an example of what is expected to be passed around. It doesn't contain both users here as we don't want to be able to call out ourselves

packages/react-components/review/beta/react-components.api.md
emlynmac: You might want to look at how refs work with TS and react: https://stackoverflow.com/questions/33796267/how-to-use-refs-in-react-with-typescript
I'm not sure why the tools have chosen to reuse and create React_2, if maybe that is the confusion?

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
emlynmac: Null is needed as sometimes there is no event that can be sent on this API.

packages/storybook/stories/INTERNAL/Chat/InputBox.stories.tsx
emlynmac: Interesting that CI does not see it.

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +167,315 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+    }
+    setSelectionStartValue(updatedStartIndex);
+    setSelectionEndValue(updatedEndIndex);
+  };
+
+  const getInputFieldTextValue = (): string => {
+    /* @conditional-compile-remove(mention) */
+    return inputTextValue;
+    return textValue;
+  };
```
AmyL219, line 459: why here is two `return ` in this function

packages/react-components/review/beta/react-components.api.md
emlynmac: We are; through CSS styling and a custom renderer.

packages/react-components/src/components/InputBoxComponent.tsx
AmyL219: what does` @conditional-compile-remove(mention)` means here?

emlynmac: The `conditional-compile-remove(mention)` removes the next token. In stable this means the second return is the only one there.

PorterNan:
I see lots of UI snapshot update:
image
They are all showing that the focus color border of sendbox is not working anymore, is this expected?

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2708,6 +2690,36 @@ export interface MediaDiagnosticsState {
...
+
+// @beta
+export type MentionOptions = {
+    lookupOptions?: MentionLookupOptions;
+    displayOptions?: MentionDisplayOptions;
+};
+
+// @beta
+export interface MentionPopoverStrings {
+    mentionPopoverHeader: string;
```
PorterNan, line 2720: It sounds pretty similar to https://developer.microsoft.com/en-us/fluentui#/controls/web/callout from fluent? Is it the same thing or different one? If it is the same, let's align the name for that

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2816,7 +2828,7 @@ export type MessageThreadProps = {
     strings?: Partial<MessageThreadStrings>;
     fileDownloadHandler?: FileDownloadHandler;
     onDisplayDateTimeString?: (messageDate: Date) => string;
-    atMentionOptions?: AtMentionOptions;
```
PorterNan, line 2819: Just out of curiosity? Why do we change atMention to mention?

packages/react-components/.eslintignore
```diff
@@ -1,3 +1,4 @@
+@types/html-to-react.d.ts
```
PorterNan, line 1: Why do we need this one? The html-to-react has its own d.ts definition, is it not working?
https://www.npmjs.com/package/html-react-parser?activeTab=code
image

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2233,6 +2212,9 @@ export const DEFAULT_COMPOSITE_ICONS: {
     ChangeSpokenLanguageIcon: JSX.Element;
 };
 
+// @beta
```
PorterNan, line 2215: Do we need to expose this Renderer here? It's hard to find this one because there are no nexus when developers want to use default renderer, can we use the pattern of onRenderMessage?

PorterNan: Also exporting a single function like this defaultRenderer in api.md feels a bit not clean

packages/react-components/package.json
```diff
@@ -42,10 +43,12 @@
     "@internal/northstar-wrapper": "1.5.1-beta.3",
     "copy-to-clipboard": "~3.3.1",
     "html-to-react": "~1.5.0",
+    "lodash": "^4.17.21",
```
PorterNan, line 46: Do we import this in any files? Just can't find the code piece related to this

packages/react-components/package.json
```diff
@@ -42,10 +43,12 @@
...
     "react-aria-live": "^2.0.5",
     "react-linkify": "^1.0.0-alpha",
-    "uuid": "^8.1.0",
-    "react-use-draggable-scroll": "~0.4.7"
+    "react-use-draggable-scroll": "~0.4.7",
+    "textarea-caret-ts": "^4.1.1",
```
PorterNan, line 50: Looks like this package do have its ts definition:
https://www.npmjs.com/package/@types/textarea-caret
Is it possible to reuse this? 
textarea-caret-ts looks like a fork that hasn't been updated since 4 yrs ago

packages/react-components/review/beta/react-components.api.md
```diff
@@ -1251,6 +1220,55 @@ export const _LocalVideoTile: React_2.MemoExoticComponent<(props: {
...
+    onRenderMention?: (mention: Mention) => JSX.Element;
+}
+
+// @beta
+export interface MentionLookupOptions {
+    onQueryUpdated: (query: string) => Promise<Mention[]>;
+    onRenderSuggestionItem?: (suggestion: Mention, onSuggestionSelected: (suggestion: Mention) => void) => JSX.Element;
+    trigger?: string;
```
PorterNan, line 1239: So the trigger is something like '@' or else string defined by customer, does this support multiple strings?

packages/react-components/review/beta/react-components.api.md
PorterNan: Hmmm, the React_2 is all over the code, and I just checked we don't have React(_1), I guess that is a problem of this api extractor

packages/react-components/review/beta/react-components.api.md
```diff
@@ -1251,6 +1220,55 @@ export const _LocalVideoTile: React_2.MemoExoticComponent<(props: {
...
+export type MentionOptions = {
+    lookupOptions?: MentionLookupOptions;
+    displayOptions?: MentionDisplayOptions;
+};
+
+// @internal
+export const _MentionPopover: (props: _MentionPopoverProps) => JSX.Element;
+
+// @internal
+export interface _MentionPopoverProps {
```
PorterNan, line 1252: We set this as an internal props, which means we are not exposing it to customer right now
Any plans on exposing the popover publicly or we will always keep it internal as part of MessageThread

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
PorterNan: Can we make it undefined?
Or even
event?: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
```diff
@@ -66,7 +73,7 @@ export const ChatMessageComponentAsEditBox = (props: ChatMessageComponentAsEditB
...
     newValue?: string | undefined
```
PorterNan, line 77: Remove | undefined, it's duplicated with ?:

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -71,12 +71,12 @@ type ChatMessageComponentAsMessageBubbleProps = {
    * @beta
    */
   onDisplayDateTimeString?: (messageDate: Date) => string;
-  /* @conditional-compile-remove(at-mention) */
+  /* @conditional-compile-remove(mention) */
   /**
-   * Optional props needed to display suggestions in the at mention scenario.
+   * Optional props needed to display suggestions in the mention scenario.
    * @beta
```
PorterNan, line 77: Correct me if I am wrong, is this ChatMessageComponentAsMessageBubbleProps a beta api or not?
If not, we should mark it as @private or @internal for this props
Same as the props above "onDisplayDateTimeString" (although it's unrelated to this PR)

packages/communication-react/review/beta/communication-react.api.md
Leah-Xia-Microsoft: > Just out of curiosity? Why do we change atMention to mention?

That's because we allow customization of the trigger. Contoso can set the trigger to anything so it might not be an @ sign. Since it can be $Mention or #Mention etc, we decided to go with Mention instead of atMention.

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -73,6 +71,7 @@ const MessageContentWithLiveAria = (props: MessageContentWithLiveAriaProps): JSX
 };
 
 const MessageContentAsRichTextHTML = (props: ChatMessageContentProps): JSX.Element => {
+  const htmlToReactParser = Parser();
```
PorterNan, line 74: wrap this with useMemo so we don't have to create new object in every render
Or we could take one step forward - I guess we could just share the same parser and instructions instance between all components?
Is it possible to extract this part into a util function or something?

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -84,12 +83,74 @@ const MessageContentAsRichTextHTML = (props: ChatMessageContentProps): JSX.Eleme
     });
   }, [props]);
 
+  const processNodeDefinitions = ProcessNodeDefinitions();
```
PorterNan, line 86: Same here, extract this part into a util function so no need to create it for each single render

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -84,12 +83,74 @@ const MessageContentAsRichTextHTML = (props: ChatMessageContentProps): JSX.Eleme
...
+        /* @conditional-compile-remove(mention) */
+        if (props.mentionDisplayOptions?.onRenderMention) {
+          // Override the handling of the <msft-mention> tag in the HTML if there's a custom renderer
+          return node.name === 'msft-mention';
+        }
+        return false;
+      },
+      processNode: (node) => {
+        /* @conditional-compile-remove(mention) */
+        if (props.mentionDisplayOptions?.onRenderMention) {
```
PorterNan, line 100: We can pass this part as a paramenter
I would imagine we just need to call a single util function here like 
parseHtmlNode([processors]);

Then we could have 2 processors for now
const mentionProcessors = createMentionProcessor(props.mentionDisplayOptions.onRenderMention);
const inlineImageProcessors = createInlineImageProcessor();

And wrap these 2 create functions with useMemo to avoid duplication and unnecessary re-render

PorterNan: inlineImageProcessors could be a const value outside of the function, so no need for useMemo

packages/react-components/src/components/ChatMessage/MentionSuggestionRenderer.tsx
```diff
@@ -0,0 +1,32 @@
...
+ */
+export const defaultMentionSuggestionRenderer = (suggestion: Mention): JSX.Element => {
+  return (
+    <msft-mention id={suggestion.id} displayText={suggestion.displayText}>
+      {suggestion.displayText}
+    </msft-mention>
+  );
+};
+
+// Add the custom HTML tag to the JSX.IntrinsicElements interface
```
PorterNan, line 20: Hmm, as a sdk, it looks not ideal to pollut global namespace environment, any alternative to avoid this?

PorterNan: This one gives me 2 conflicts ideas in my mind:

1. If someone wants to use the custom html especially with typescript, then we will need this declaration to make it work, although it's not that standard
2. If someone doesn't need this one, they will see this 'msft-mention' when they try to write their own code in react (auto complete will show this one), after they import our package, it might be very confusing during development

PorterNan: It's kinda like - 

1. if we don't pollute global namespace, then developers have to write their own .d.ts file for this custom tag
2. If we pollute global namespace, then it might be confusing for developers who don't know this feature

I am feeling react + typescript is not that ready for custom html tags

PorterNan: Found a very neat way to do custom tag in react:
https://codepen.io/jiangnanhello/pen/vYVWVLv

packages/react-components/src/components/ChatMessage/MentionSuggestionRenderer.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Provides the default implementation for rendering an Mention
+ * @param suggestion - The suggestion to render
+ *
+ * @beta
+ */
+export const defaultMentionSuggestionRenderer = (suggestion: Mention): JSX.Element => {
+  return (
+    <msft-mention id={suggestion.id} displayText={suggestion.displayText}>
```
PorterNan, line 14: Would you provide an example code of how contoso use this custom html tag? Is it by using some kinda document.findElemetByTag or something?

PorterNan: I guess findElementById would be another way, but need a real example to understand this part better

packages/react-components/.eslintignore
emlynmac: I'm actively working with the package owner to get that PR in; I have provided this upstream.

packages/react-components/package.json
emlynmac: textarea-caret-ts is a TS package and has typings already.

packages/react-components/package.json
emlynmac: `InputBoxComponent.tsx` uses it for debounce.

packages/react-components/src/components/ChatMessage/MentionSuggestionRenderer.tsx
emlynmac: The global enables TS devs to use the new tag; we declare it here and is part of this feature and the SDK. As you pointed out, if we don't then any react user will need to create their own.
We've carefully chosen the name of the tag here to avoid clashing.

packages/react-components/review/beta/react-components.api.md
emlynmac: Current plan is internal only.

packages/react-components/review/beta/react-components.api.md
emlynmac: The trigger string is passed in here to the context used. By default it is `@` but it could be something else if you set it

packages/communication-react/review/beta/communication-react.api.md
emlynmac: We have that patten through `onRenderSuggestionItem` so we don't need to expose the default renderer here.

emlynmac: Maybe this needs to be exposed as a simple React component, `Mention`

packages/react-components/src/components/ChatMessage/MentionSuggestionRenderer.tsx
emlynmac: Ok, I have a better solution. Will update....

emlynmac: > Found a very neat way to do custom tag in react: https://codepen.io/jiangnanhello/pen/vYVWVLv

This doesn't compile in the codebase; it needs the extension to the JSX namespace still

packages/react-components/review/beta/react-components.api.md
Leah-Xia-Microsoft: > We set this as an internal props, which means we are not exposing it to customer right now Any plans on exposing the popover publicly or we will always keep it internal as part of MessageThread

The popover will be internal until there's some customer asks.

packages/communication-react/review/beta/communication-react.api.md
emlynmac: We will wrap this up in the same way as `MessageThread`'s `onRenderMessage` function providing the default renderer as a second parameter.

PorterNan: Component sounds like another topic, cause renderer might do some kinda props mapping and more, let's add it to 2nd parameter and keep our discussion on MsftMention component

packages/react-components/package.json
PorterNan: Found it! That file was collapsed by github

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -25,12 +30,16 @@ import {
...
+import { Caret } from 'textarea-caret-ts';
 import { isDarkThemed } from '../theming/themeUtils';
 import { useTheme } from '../theming';
-/* @conditional-compile-remove(at-mention) */
-import { AtMentionLookupOptions } from './AtMentionFlyout';
-
+/* @conditional-compile-remove(mention) */
+import { MentionLookupOptions, _MentionPopover, Mention } from './MentionPopover';
+/* @conditional-compile-remove(mention) */
+import { debounce } from 'lodash';
```
PorterNan, line 40:
```suggestion
import { debounce } from 'lodash/debounce';
```
it's a small trick to enable treeshaking for lodash
You can read more from here if you are interested:
https://medium.com/att-israel/how-to-shrink-tree-shake-lodash-in-bundle-file-for-react-native-90813a995dfa#:~:text=5.%20Having%20said%20that%20%F0%9F%98%89

packages/react-components/package.json
PorterNan: Another concern is download number - [@types/textarea-caret](https://www.npmjs.com/package/@types/textarea-caret) has 10k+ download per week, and [textarea-caret](https://www.npmjs.com/package/textarea-caret) has 150k+ download
And the textarea-caret-ts is only a fork with 900+ download a week.
If it is possible, it's preferred to use more popular package and original package + types definition, cause it might get more validation through more people and easier to get things fixed

PorterNan: But lemme know if there is some APIs specifically in textarea-caret-ts, so doing a package switch will be time consuming

packages/react-components/src/components/ChatMessage/MentionSuggestionRenderer.tsx
PorterNan: image

This should be working well, validated locally, so we do all the type check for this small component props then wrap the custom tag as a react component (and always access it using that component)

The idea is modularized the custom tag, people should not see it unless they import the file

packages/react-components/package.json
emlynmac: While the JS package has more downloads, this is likely due to the fact that the fork is more recent and is TS oriented. JS is still more prevalent than TS so I would expect numbers to be different here.
What we do know is that the package selected is working as expected and is more recently updated. I think for now, we leave this known package in place.

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -174,16 +174,12 @@ const messageContentAriaText = (props: ChatMessageContentProps): string | undefi
...
-};
+const processNodeDefinitions = ProcessNodeDefinitions();
+const htmlToReactParser = Parser();
 
 const processHtmlToReact = (props: ChatMessageContentProps): JSX.Element => {
-  const htmlToReactParser = new Parser();
-
   /* @conditional-compile-remove(teams-inline-images) */
-  const processInlineImage: NodeProcessInstruction = {
+  const processInlineImage: ProcessingInstructionType = {
```
PorterNan, line 182: This processInlineImage instance will be created as a new instance everytime when you call processHtmlToReact, consider extract it out

PorterNan: Would be nice to get these 2 process objects out of the function, it makes this function too long to read, and I am expecting more process instruction will be coming in the future

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -53,8 +62,8 @@ type InputBoxComponentProps = {
   inlineChildren: boolean;
   'data-ui-id'?: string;
   id?: string;
-  textValue: string;
-  onChange: (event: FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string | undefined) => void;
+  textValue: string; // This could be plain text or HTML.
```
PorterNan, line 65: How do we distinguish if this is html or plain text in our code logic?

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -53,8 +62,8 @@ type InputBoxComponentProps = {
...
+  onChange: (event: FormEvent<HTMLInputElement | HTMLTextAreaElement> | null, newValue?: string | undefined) => void;
```
PorterNan, line 66:
```suggestion
  onChange: (event?: FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string | undefined) => void;
```

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -88,8 +97,66 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+      }
+    },
+    [textFieldRef, caretIndex]
+  );
+
+  /* @conditional-compile-remove(mention) */
+  // Parse the text and get the plain text version to display in the input box
+  useEffect(() => {
+    const trigger = mentionLookupOptions?.trigger || defaultMentionTrigger;
+    console.log('textValue <- html', textValue);
```
PorterNan, line 152: There are lots of console.log() here, remove?

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -88,8 +97,66 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+  const [selectionEndValue, setSelectionEndValue] = useState<number | null>(null);
+  /* @conditional-compile-remove(mention) */
+  const [shouldHandleOnMouseDownDuringSelect, setShouldHandleOnMouseDownDuringSelect] = useState<boolean>(true);
+
+  /* @conditional-compile-remove(mention) */
+  // Caret position in the text field
+  const [caretPosition, setCaretPosition] = useState<Caret.Position | undefined>(undefined);
+  /* @conditional-compile-remove(mention) */
+  // Index of where the caret is in the text field
+  const [caretIndex, setCaretIndex] = useState<number | null>(null);
```
PorterNan, line 132: In ts/js we always prefer undefined if it is just initialize with no value (most case we just directly use undefined)

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+  if (!parentTag) {
+    tags.push(tag);
+  } else {
+    if (!parentTag.subTags) {
+      parentTag.subTags = [tag];
+    } else {
+      parentTag.subTags.push(tag);
+    }
+  }
+};
```
PorterNan, line 1490: Let's open a util file or any new file for those functions... this tsx file is just simply too long to read

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+        // the change is a tag or starts before the tag
+        // tag should be removed, no matter if there are subtags
+        result += htmlText.substring(lastProcessedHTMLIndex, tag.openTagIdx) + processedChange;
+        processedChange = '';
+        lastProcessedHTMLIndex = closeTagIdx + closeTagLength;
+        // the change is handled; exit
+        break;
+      } else if (startIndex >= tag.plainTextBeginIndex && oldPlainTextEndIndex < plainTextEndIndex) {
+        // edge case: the change is between tag
+        console.log('updateHTML 1.2 result', result);
```
PorterNan, line 910: remove

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+      //no close tag
+      plainTextEndIndex = tag.plainTextBeginIndex;
+      closeTagIdx = tag.openTagIdx + tag.openTagBody.length;
+      closeTagLength = 0;
+    }
+
+    if (startIndex < plainTextEndIndex) {
+      // change started before the end tag
+      if (startIndex <= tag.plainTextBeginIndex && oldPlainTextEndIndex === plainTextEndIndex) {
+        console.log('updateHTML 1.1 result', result);
```
PorterNan, line 900: remove

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -133,13 +493,100 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+            setInputTextValue(newValue ?? '');
+            /* @conditional-compile-remove(mention) */
+            handleOnChange(e, newValue);
+            /* @conditional-compile-remove(mention) */
+            return;
+            onChange(e, newValue);
+          }}
+          /* @conditional-compile-remove(mention) */
+          onSelect={(e) => {
+            if (caretIndex !== null) {
```
PorterNan, line 508: replace null with undefined if possible

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -133,13 +493,100 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+          // onMouseMove={(e) => {
+          //   console.log('!!!!not equal onMouseMove');
+          // should preventDefault be used?
+          // for handling mouse actions
+          /* @conditional-compile-remove(mention) */
+          // console.log('updateHTML onMouseMove selectionStart', e.currentTarget.selectionStart);
+          // console.log('updateHTML onMouseMove selectionEnd', e.currentTarget.selectionEnd);
+          // console.log('updateHTML onMouseMove selectionDirection', e.currentTarget.selectionDirection);
+          // if (e.currentTarget.selectionStart !== e.currentTarget.selectionEnd) {
+          //   console.log('updateHTML onMouseMove e.currentTarget.selectionStart !== e.currentTarget.selectionEnd');
```
PorterNan, line 556: Remove comments

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +171,317 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+      /* @conditional-compile-remove(mention) */
+      activeSuggestionIndex,
+      /* @conditional-compile-remove(mention) */
+      onSuggestionSelected
+    ]
   );
 
+  /* @conditional-compile-remove(mention) */
+  const debouncedQueryUpdate = useRef(
+    useDebouncedCallback(async (query: string) => {
```
PorterNan, line 283: Why do we useRef then .current here?

PorterNan: Is it just call the hook for once? Ideally react component wouldn't allow you to conditional call a hook just for once

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +171,317 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+
+  /* @conditional-compile-remove(mention) */
+  useEffect(() => {
+    return () => {
+      debouncedQueryUpdate.cancel();
+    };
+  }, [debouncedQueryUpdate]);
+
+  /* @conditional-compile-remove(mention) */
+  const handleOnChange = async (
```
PorterNan, line 302: Any special reason we use async for this function? I don't see any await in the function body

packages/react-components/src/components/InputBoxComponent.tsx
PorterNan: I guess useMeme should be working well if that is the case

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +171,317 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+              targetPositionOffset={caretPosition}
+              onRenderSuggestionItem={mentionLookupOptions?.onRenderSuggestionItem}
+              onSuggestionSelected={onSuggestionSelected}
+              onDismiss={() => {
+                updateMentionSuggestions([]);
+              }}
+            />
+          )
+        }
         <TextField
```
PorterNan, line 485: I can feel all these new logics could be big enough to be one new component (and glad that we got this huge part working well!)

I am thinking of we create a component named TextFieldWithMention
then do this 
const renderTextField = () => {
/* @conditional-compile-remove(mention) */
return <TextFieldWithMention .../>;
return <Texfileld ... />
}

Then you don't have to conditional compile every single lines of new code, most of new logic will be there without CC, you just need to choose whether to import it or not

It's just a small trick for making the feature easier to be stabilized, and easy to adding new logic
And also, we love smaller files ;D

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+            htmlText.substring(lastProcessedHTMLIndex, tag.openTagIdx + tag.openTagBody.length + startChangeDiff) +
+            processedChange;
+
+          processedChange = '';
+          lastProcessedHTMLIndex = tag.openTagIdx + tag.openTagBody.length + endChangeDiff;
+          // the change is handled; exit
+          break;
+        }
+      } else if (startIndex > tag.plainTextBeginIndex && oldPlainTextEndIndex > plainTextEndIndex) {
+        console.log('updateHTML 1.3 result', result);
```
PorterNan, line 973: remove

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+          // no subtags
+          result += htmlText.substring(
+            lastProcessedHTMLIndex,
+            tag.openTagIdx + tag.openTagBody.length + startChangeDiff
+          );
+          lastProcessedHTMLIndex = closeTagIdx;
+          //proceed with the next calculations
+        }
+      } else if (startIndex < tag.plainTextBeginIndex && oldPlainTextEndIndex > plainTextEndIndex) {
+        console.log('updateHTML 1.4 result', result);
```
PorterNan, line 1024: I guess we can just search for all console.log then remove them, I will just ignore the rest of them in the PR

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -88,8 +97,66 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+    },
+    [textFieldRef, caretIndex]
+  );
+
+  /* @conditional-compile-remove(mention) */
+  // Parse the text and get the plain text version to display in the input box
+  useEffect(() => {
+    const trigger = mentionLookupOptions?.trigger || defaultMentionTrigger;
+    console.log('textValue <- html', textValue);
+    const [tags, plainText] = textToTagParser(textValue, trigger);
```
PorterNan, line 153: So tags are those strings following the trigger symbol?

packages/react-components/src/components/InputBoxComponent.tsx
PorterNan: Wrap this with useCallback btw, avoid creating new instance everytime in rendering

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+ */
+const textToTagParser = (text: string, trigger: string): [TagData[], string] => {
+  const tags: TagData[] = []; // Tags passed back to the caller
+  const tagParseStack: TagData[] = []; // Local stack to use while parsing
+
+  let plainTextRepresentation = '';
+
+  let parseIndex = 0;
+  while (parseIndex < text.length) {
+    // console.log('Parsing at index ' + parseIndex + ' of ' + text.length);
```
PorterNan, line 1335: Remove

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+          result += htmlText.substring(
+            lastProcessedHTMLIndex,
+            tag.openTagIdx + tag.openTagBody.length + startChangeDiff
+          );
+          lastProcessedHTMLIndex = closeTagIdx;
+          //proceed with the next calculations
+        }
+      } else if (startIndex < tag.plainTextBeginIndex && oldPlainTextEndIndex > plainTextEndIndex) {
+        console.log('updateHTML 1.4 result', result);
+        // the change starts before  the tag and finishes after it
```
PorterNan, line 1025: Remove extra space

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+
```
PorterNan, line 1026: Remove extra line

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+        // tag should be removed, no matter if there are subtags
+        // no need to save anything between lastProcessedHTMLIndex and closeTagIdx + closeTagLength
+        lastProcessedHTMLIndex = closeTagIdx + closeTagLength;
+        //proceed with the next calculations
```
PorterNan, line 1030: space after //

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+      } else if (startIndex === tag.plainTextBeginIndex && oldPlainTextEndIndex > plainTextEndIndex) {
+        console.log('updateHTML 1.5 result', result);
+        // the change starts in the tag and finishes after it
+        // tag should be removed, no matter if there are subtags
+        result += htmlText.substring(lastProcessedHTMLIndex, tag.openTagIdx);
+        // processedChange shouldn't be updated as it will be added after the tag
+        lastProcessedHTMLIndex = closeTagIdx + closeTagLength;
+        //proceed with the next calculations
```
PorterNan, line 1038: Space after //

packages/react-components/src/components/MentionPopover.tsx
```diff
@@ -0,0 +1,322 @@
...
+      initialsColor: theme.palette.neutralLight,
+      initialsTextColor: theme.palette.neutralSecondary,
+      showOverflowTooltip: false,
+      showUnknownPersonaCoin: !displayName?.trim() || displayName === displayNamePlaceholder
+    };
+
+    return <Persona {...avatarOptions} />;
+  };
+
+  const defaultOnRenderSuggestionItem = (
```
PorterNan, line 247: This could be a function living outside of the react function, so it wouldn't be created everytime

packages/react-components/src/components/MentionPopover.tsx
```diff
@@ -0,0 +1,322 @@
...
+            (changedSelection ?? false) && active
+          )}
+        >
+          {personaRenderer(suggestion.displayText)}
+        </Stack>
+      </div>
+    );
+  };
+
+  const getHeaderTitle = (): string => {
```
PorterNan, line 280: Wrap this with useCallback

packages/react-components/src/components/MentionPopover.tsx
```diff
@@ -0,0 +1,322 @@
...
+    /* @conditional-compile-remove(mention) */
+    return localeStrings.mentionPopover.mentionPopoverHeader;
+    return '';
+  };
+
+  return (
+    <div ref={popoverRef}>
+      <Stack
+        className={mergeStyles(
+          {
```
PorterNan, line 293: Let's move these styles into the style file

PorterNan:
image
I still can see some unnecessary UI snapshot updates for this PR, can we make sure the UI snapshot update is minimum? Ideally it should only impact what we modified

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Based on the presence of tags when parsing

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Tags are the HTML tags found, as the string to the inputbox can be HTML format

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: All of them will be removed in the next PRs for this feature as this is WIP right now

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: A new refactored task will be created to handle this functionality

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: All of them will be removed in the next PRs for this feature as this is WIP right now

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: All of them will be removed in the next PRs for this feature as this is WIP right now

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: All of them will be removed in the next PRs for this feature as this is WIP right now

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: This change will be done in `vhuseinova/mention-update`

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: A new refactoring task will be created to handle this functionality

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: All of them will be removed in the next PRs for this feature as this is WIP right now

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: All of them will be removed in the next PRs for this feature as this is WIP right now

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: All of them will be removed in the next PRs for this feature as this is WIP right now

jimchou-dev (commented):
Still reviewing InputBoxComponent.tsx but will add these comments first

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: It will be fixed in the next PR with removing all console logs

packages/react-components/src/components/MentionPopover.tsx
emlynmac: The split here for the styles is deliberate; we need to enforce the position and location but we have exposed the other styles through the style file.

packages/react-components/src/components/MentionPopover.tsx
```diff
@@ -0,0 +1,322 @@
...
+    [onDismiss]
+  );
+
+  useEffect(() => {
+    if (changedSelection === undefined) {
+      setChangedSelection(false);
+    } else if (changedSelection === false) {
+      setChangedSelection(true);
+    }
+  }, [activeSuggestionIndex, changedSelection]);
```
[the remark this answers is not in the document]
emlynmac: This is to replicate Teams' selection on a mention popup, which changes the active styling based on whether it is the first selection or subsequent.

packages/react-components/src/components/MentionPopover.tsx
```diff
@@ -0,0 +1,322 @@
...
+    window && window.addEventListener('click', dismissPopoverWhenClickingOutside);
+    return () => {
+      window && window.removeEventListener('click', dismissPopoverWhenClickingOutside);
+    };
+  }, [dismissPopoverWhenClickingOutside]);
+
+  // Determine popover position
+  useEffect(() => {
+    const rect = target?.current?.getBoundingClientRect();
+    const maxWidth = 200;
```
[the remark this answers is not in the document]
emlynmac: Maybe, but in a subsequent PR.

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+};
+
+/* @conditional-compile-remove(mention) */
+type TagData = {
+  tagType: string; // The type of tag (e.g. msft-mention)
+  openTagIdx: number; // Start of the tag relative to the parent content
+  openTagBody: string; // Complete open tag body
+  content?: string; // All content between the open and close tags
+  closeTagIdx?: number; // Start of the close tag relative to the parent content
+  subTags?: TagData[]; // Any child tags
```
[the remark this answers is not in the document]
emlynmac: Any nested HTML tag will be this way.

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +648,843 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+};
+
+/* @conditional-compile-remove(mention) */
+type HtmlTagType = 'open' | 'close' | 'self-closing';
+/* @conditional-compile-remove(mention) */
+type HtmlTag = {
+  content: string;
+  startIdx: number;
+  type: HtmlTagType;
+};
```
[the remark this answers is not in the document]
vhuseinova-msft: A new refactoring task will be created to move this functionality

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: Removed

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: Removed

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: Removed

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: Removed

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: Removed

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: Removed

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: Updated

packages/react-components/src/components/InputBoxComponent.tsx
vhuseinova-msft: Updated

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +168,330 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+      updateMentionSuggestions,
+      /* @conditional-compile-remove(mention) */
+      localeStrings
+    ]
+  );
+
+  const onTextFieldKeyDown = useCallback(
     (ev: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>) => {
       // Uses KeyCode 229 and which code 229 to determine if the press of the enter key is from a composition session or not (Safari only)
       if (ev.nativeEvent.isComposing || ev.nativeEvent.keyCode === 229 || ev.nativeEvent.which === 229) {
```
PorterNan, line 224: Could be constant put into constant file

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +635,818 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+  return suggestion.displayText !== '' ? suggestion.displayText : displayNamePlaceholder ?? '';
+};
+
+/* @conditional-compile-remove(mention) */
+type TagData = {
+  tagType: string; // The type of tag (e.g. msft-mention)
+  openTagIdx: number; // Start of the tag relative to the parent content
+  openTagBody: string; // Complete open tag body
+  content?: string; // All content between the open and close tags
+  closeTagIdx?: number; // Start of the close tag relative to the parent content
```
PorterNan, line 1271: closeTagPosition? Index feels like something for array

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -53,8 +62,8 @@ type InputBoxComponentProps = {
   inlineChildren: boolean;
   'data-ui-id'?: string;
   id?: string;
-  textValue: string;
-  onChange: (event: FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string | undefined) => void;
+  textValue: string; // This could be plain text or HTML.
+  onChange: (event?: FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string | undefined) => void;
```
PorterNan, line 66:
```suggestion
  onChange: (event?: FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => void;
```

packages/react-components/src/components/MentionPopover.tsx
emlynmac: Added parameters and an interface to the style.

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Will be part of the upcoming refactor task we chatted about.

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: We can revisit in the upcoming PRs.

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +167,330 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+    return () => {
+      debouncedQueryUpdate.cancel();
+    };
+  }, [debouncedQueryUpdate]);
+
+  /* @conditional-compile-remove(mention) */
+  const handleOnChange = useCallback(
+    async (
+      event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+      updatedValue?: string | undefined
```
PorterNan, line 299:
```suggestion
      updatedValue?: string
```

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +167,330 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+    ): Promise<void> => {
+      let newValue = updatedValue;
```
PorterNan, line 301: let newValue = updateValue ?? '';

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +167,330 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+      }
+      const triggerText = mentionLookupOptions?.trigger ?? defaultMentionTrigger;
+
+      const newTextLength = newValue.length;
+      let selectionEnd = textFieldRef?.current?.selectionEnd || -1;
+      if (selectionEnd < 0) {
+        selectionEnd = 0;
+      } else if (selectionEnd > newTextLength) {
+        selectionEnd = newTextLength - 1;
+      }
```
PorterNan, line 313:
```
selectionEnd = Math.max(0, selectionEnd);
selectionEnd = Math.min(selectionEnd, newTextLength - 1);
```

packages/react-components/src/components/InputBoxComponent.tsx
PorterNan: What if someone intentionally type a text like this:

Hello, today I wanna read book of <Learning C++>

Does the logic strong enough to hold this case

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +167,330 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+      } else {
+        // there are tags in the text value, textValue is html string
+        const { changeStart, oldChangeEnd, newChangeEnd } = findStringsDiffIndexes(
+          inputTextValue,
+          newValue,
+          selectionEnd
+        );
+        // get updated html string
+        const change = newValue.substring(changeStart, newChangeEnd);
+        const [updatedHTML, updatedChangeNewEndIndex] = updateHTML(
```
PorterNan, line 365: So we are parse html to text then parse it back, that's an amazing process

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +167,330 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+            selectionStartValue
+          );
+
+          updatedStartIndex = newSelectionIndex;
+          updatedEndIndex = newSelectionIndex;
+        }
+      }
+    } else if (event.currentTarget.selectionStart !== event.currentTarget.selectionEnd) {
+      // Both e.currentTarget.selectionStart !== selectionStartValue and e.currentTarget.selectionEnd !== selectionEndValue can be true when a user selects a text by double click
+      if (event.currentTarget.selectionStart !== null && event.currentTarget.selectionStart !== selectionStartValue) {
```
PorterNan, line 433: use undefined instead of null
A bit more context:
if you do 
`typeof null`
 in js, the return value will be Object, this will create strange regression when we do the runtime type check

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +167,330 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+          //TODO: here it takes -1 when it shouldn't, update selectionstart and end with mouse move and or touch move
+          updatedStartIndex = findNewSelectionIndexForMention(
+            mentionTag,
+            inputTextValue,
+            event.currentTarget.selectionStart,
+            selectionStartValue ?? -1
+          );
+        }
+      }
+      if (event.currentTarget.selectionEnd !== null && event.currentTarget.selectionEnd !== selectionEndValue) {
```
PorterNan, line 446:
```suggestion
      if (event.currentTarget.selectionEnd !== undefined && event.currentTarget.selectionEnd !== selectionEndValue) {
```

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -104,24 +167,330 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+            mentionTag,
+            inputTextValue,
+            event.currentTarget.selectionEnd,
+            selectionEndValue ?? -1
+          );
+        }
+      }
+    }
+    // e.currentTarget.selectionDirection should be set to handle shift + arrow keys
+    if (event.currentTarget.selectionDirection === null) {
```
PorterNan, line 461:
```suggestion
    if (event.currentTarget.selectionDirection === undefined) {
```

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -133,13 +502,77 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+            // onClick and onMouseDown can't handle clicking on mention event because
+            // onClick has wrong range as it's called after onSelect
+            // onMouseDown doesn't have correct selectionRange yet
+            // so we need to handle onMouseDown to prevent onSelect default behavior
+            /* @conditional-compile-remove(mention) */
+            setShouldHandleOnMouseDownDuringSelect(true);
+          }}
+          onTouchStart={() => {
+            // see onMouseDown for more details
+            /* @conditional-compile-remove(mention) */
```
PorterNan, line 567: Move /* @conditional-compile-remove(mention) */ to the top of onTouchStart

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +634,818 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+
+/* @conditional-compile-remove(mention) */
+/**
+ * Find mention tag if selection is inside of it
+ *
+ * @private
+ */
+const findMentionTagForSelection = (tags: TagData[], selection: number): TagData | undefined => {
+  let mentionTag: TagData | undefined = undefined;
+  for (let i = 0; i < tags.length; i++) {
```
PorterNan, line 646: It's always preferred to use 
tags.forEach if you are not changing the array structure when you are doing the loop

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +634,818 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+ */
+const findNewSelectionIndexForMention = (
+  tag: TagData,
+  textValue: string,
+  selection: number,
+  previousSelection: number
+): number => {
+  if (
+    tag.plainTextBeginIndex === undefined ||
+    tag.tagType !== 'msft-mention' ||
```
PorterNan, line 689: Not fully understand this part - if not 'msft-mention' why do we still return selection?

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +634,818 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+  lastProcessedHTMLIndex: number,
+  processedChange: string,
+  tag: TagData,
+  closeTagIdx: number,
+  closeTagLength: number,
+  plainTextEndIndex: number,
+  startIndex: number,
+  oldPlainTextEndIndex: number,
+  mentionTagLength: number
+): [resultValue: string, updatedChange: string, htmlIndex: number, plainTextSelectionEndIndex: number | null] => {
```
PorterNan, line 738: I am less concerned about private implementation, but we do prefer returning object instead of tuple, it's hard to read from the caller part

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +634,818 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+const updateHTML = (
+  htmlText: string,
+  oldPlainText: string,
+  newPlainText: string,
+  tags: TagData[],
+  startIndex: number,
+  oldPlainTextEndIndex: number,
+  change: string,
+  mentionTrigger: string
+): [string, number | null] => {
```
PorterNan, line 807: Either in this PR or a further work item, use object instead of tuple
the reader just simply can't understand what does the second number mean when not looking deeply into the function

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +634,818 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+          const startOfRemainingText =
+            (lastSubTag.closeTagIdx ?? lastSubTag.openTagIdx) + lastSubTag.tagType.length + 3;
+          const trailingText = currentOpenTag.content.substring(startOfRemainingText);
+          plainTextRepresentation += trailingText;
+        }
+
+        currentOpenTag.plainTextEndIndex = plainTextRepresentation.length;
+        addTag(currentOpenTag, tagParseStack, tags);
+      } else {
+        console.error(
```
PorterNan, line 1362: Throw errors instead of do console.error, otherwise it cannot be caught by contoso app, and cannot be hide by contoso

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +634,818 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+
+/* @conditional-compile-remove(mention) */
+/**
+ * Parse the text and return the tags and the plain text in one go
+ * @param text - The text to parse for HTML tags
+ * @param trigger The trigger to show for the msft-mention tag in plain text
+ *
+ * @returns An array of tags and the plain text representation
+ */
+const textToTagParser = (text: string, trigger: string): [TagData[], string] => {
```
PorterNan, line 1293: Not necessary in this PR, I believe we would need Unit tests for all the html parsing functions just to ensure everything works always right

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -201,3 +634,818 @@ export const InputBoxButton = (props: InputBoxButtonProps): JSX.Element => {
...
+    // Update parsing index; move past the end of the close tag
+    parseIndex = foundHtmlTag.startIdx + foundHtmlTag.content.length;
+  } // While parseIndex < text.length loop
+
+  return [tags, plainTextRepresentation];
+};
+
+/* @conditional-compile-remove(mention) */
+const parseOpenTag = (tag: string, startIdx: number): TagData => {
+  const tagType = tag
```
PorterNan, line 1381: Strings like:

< div /> or 
< hello > should be considered as just plain string instead of tags, do we handle cases like this?

PorterNan: It's like space before after the open tag

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: We will address in the refactor task; I agree.

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: There is a task for this in the backlog; 100% agree.

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Negative testing also needed for sure; part of the test tasks.

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Added to refactor task

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Added to refactor task

packages/react-components/src/components/SendBox.tsx
```diff
@@ -353,7 +353,12 @@ export const SendBox = (props: SendBoxProps): JSX.Element => {
   }, [activeFileUploads, props, localeStrings]);
 
   return (
-    <Stack className={mergeStyles(sendBoxWrapperStyles)}>
+    <Stack
+      className={mergeStyles(
+        sendBoxWrapperStyles,
+        { overflow: 'visible' } // This is needed for the mention popup to be visible
```
PorterNan, line 359: What would be happening if the content of senboxWrapper has real content overflow? Like too many line of text in the box?

PorterNan, line 359: BTW, why don't we add this one directly into sendBoxWrapperStyles

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: We're potentially given HTML that we can only edit as plain text.

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: The typing of the events are using null, so we've just continued that use internally. We can add to the refactor task to use undefined internally, but I'm curious why null is use for the event typing.

packages/react-components/src/components/MentionPopover.tsx
```diff
@@ -0,0 +1,332 @@
...
+    if (title) {
+      return title;
+    }
+    /* @conditional-compile-remove(mention) */
+    return localeStrings.mentionPopover.mentionPopoverHeader;
+    return '';
+  }, [localeStrings, title]);
+
+  return (
+    <div ref={popoverRef}>
```
PorterNan, line 300: BTW, a QQ for this part, why don't we use fluent ui component for this PopOver but implement our own div element instead, it feels like lots of logic we could reuse from fluent ui component

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: This is an unintended behavioural change I think.
The typing on `selectionDirection` is `selectionDirection: "forward" | "backward" | "none" | null;`
In theory it can't be `undefined`, but can be `null`.

PorterNan (approved):
That's a huge PR and a huge feature! Thanks for addressing comments and I am going to spend some time playing it in storybook ;D

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Can you go explain the why here please?

packages/react-components/src/components/MentionPopover.tsx
emlynmac: Fluent 9 has a popover; I think we can move to it when we get to fluent 9

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Ok, so for the purpose that you can't mess with the array directly.

emlynmac: Added to refactor task

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Added to refactor task

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -133,13 +502,77 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+                setSelectionStartValue(e.currentTarget.selectionStart);
+                setSelectionEndValue(e.currentTarget.selectionEnd);
+              }
+            } else {
+              updateSelectionIndexesWithMentionIfNeeded(e);
+            }
+            /* @conditional-compile-remove(mention) */
+            setShouldHandleOnMouseDownDuringSelect(false);
+          }}
+          onMouseDown={() => {
```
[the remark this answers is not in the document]
emlynmac: Added to refactor task

packages/react-components/src/components/MentionPopover.tsx
PorterNan: Let's open an further task them, just in case we forget it someday when we get fluent 9

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: type of selectionEnd is:
```
 /** Gets or sets the end position or offset of a text selection. */
    selectionEnd: number | null;
 ```

**Review by Leah-Xia-Microsoft — APPROVED** [review]

**Review by jimchou-dev — APPROVED** [review]

File: packages/react-components/src/components/InputBoxComponent.tsx (hunk 62 above)
Comment (Leah-Xia-Microsoft) [reply packages/react-components/src/components/InputBoxComponent.tsx:638]: @PorterNan In this case, we are using `break` to break out of the loop early. Foreach doesn't work with break statement. We can use `every` instead, but using a for loop because of break statements is also legitimate.

File: packages/react-components/src/components/InputBoxComponent.tsx [hunk 72]
```diff
@@ -104,24 +167,322 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+
+  const onTextFieldKeyDown = useCallback(
     (ev: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>) => {
       // Uses KeyCode 229 and which code 229 to determine if the press of the enter key is from a composition session or not (Safari only)
       if (ev.nativeEvent.isComposing || ev.nativeEvent.keyCode === 229 || ev.nativeEvent.which === 229) {
         return;
       }
+      if (ev.key === 'ArrowUp') {
+        ev.preventDefault();
+        /* @conditional-compile-remove(mention) */
```
AmyL219, line 228: Where is the key.preventDefault for `ArrowUp` and `ArrowDown` inherit from?

PR #2992, opened 2023-04-27
[CallWithChat] Updating the factory method for creating callwithchatadapter from stateful clients. 

JoshuaLai:
# What
Included new options bags in the factory method when creating from stateful client

# Why
Chat adapter needs the credential in the options bag to perform inline image download. 
So we need a new argument to be able to pass that downstream. 

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/communication-react/review/beta/communication-react.api.md
+++ b/packages/communication-react/review/beta/communication-react.api.md
@@ -182,7 +182,7 @@ export type AzureCommunicationCallWithChatAdapterFromClientArgs = {
     callClient: StatefulCallClient;
     chatClient: StatefulChatClient;
     chatThreadClient: ChatThreadClient;
-    options?: CallWithChatOptions;
+    options?: AzureCommunicationChatAdapterOptions;
 };
 
 // @public
@@ -192,6 +192,11 @@ export type AzureCommunicationChatAdapterArgs = {
     displayName: string;
     credential: CommunicationTokenCredential;
     threadId: string;
+};
+
+// @beta
+export type AzureCommunicationChatAdapterOptions = {
+    credential?: CommunicationTokenCredential;
 };
 
 // @public
@@ -1084,7 +1089,7 @@ export interface CallWithChatControlOptions extends CommonCallControlOptions {
 export type CallWithChatEvent = 'callError' | 'chatError' | 'callEnded' | 'isMutedChanged' | 'callIdChanged' | 'isLocalScreenSharingActiveChanged' | 'displayNameChanged' | 'isSpeakingChanged' | 'callParticipantsJoined' | 'callParticipantsLeft' | 'selectedMicrophoneChanged' | 'selectedSpeakerChanged' | /* @conditional-compile-remove(close-captions) */ 'isCaptionsActiveChanged' | /* @conditional-compile-remove(close-captions) */ 'captionsReceived' | 'messageReceived' | 'messageSent' | 'messageRead' | 'chatParticipantsAdded' | 'chatParticipantsRemoved';
 
 // @beta
-export type CallWithChatOptions = ChatAdapterOptions;
+export type CallWithChatOptions = AzureCommunicationChatAdapterOptions;
 
 // @beta
 export const CameraAndMicrophoneSitePermissions: (props: CameraAndMicrophoneSitePermissionsProps) => JSX.Element;
@@ -1289,11 +1294,6 @@ export type CaptionsReceivedListener = (event: {
 
 // @public
 export type ChatAdapter = ChatAdapterThreadManagement & AdapterState<ChatAdapterState> & Disposable & ChatAdapterSubscribers & FileUploadAdapter;
-
-// @beta
-export type ChatAdapterOptions = {
-    credential?: CommunicationTokenCredential;
-};
 
 // @public
 export type ChatAdapterState = ChatAdapterUiState & ChatCompositeClientState;
@@ -1861,7 +1861,7 @@ export const createAzureCommunicationCallAdapterFromClient: (callClient: Statefu
 export const createAzureCommunicationCallWithChatAdapter: ({ userId, displayName, credential, endpoint, locator, alternateCallerId, callAdapterOptions }: AzureCommunicationCallWithChatAdapterArgs) => Promise<CallWithChatAdapter>;
 
 // @public
-export function createAzureCommunicationCallWithChatAdapterFromClients(callClient: StatefulCallClient, callAgent: CallAgent, callLocator: CallAdapterLocator | TeamsMeetingLinkLocator, chatClient: StatefulChatClient, chatThreadClient: ChatThreadClient, options?: CallWithChatOptions): Promise<CallWithChatAdapter>;
+export function createAzureCommunicationCallWithChatAdapterFromClients(callClient: StatefulCallClient, callAgent: CallAgent, callLocator: CallAdapterLocator | TeamsMeetingLinkLocator, chatClient: StatefulChatClient, chatThreadClient: ChatThreadClient, options?: AzureCommunicationChatAdapterOptions): Promise<CallWithChatAdapter>;
 
 // @public
 export const createAzureCommunicationChatAdapter: ({ endpoint: endpointUrl, userId, displayName, credential, threadId }: AzureCommunicationChatAdapterArgs) => Promise<ChatAdapter>;
--- a/packages/communication-react/review/beta/communication-react.api.md
+++ b/packages/communication-react/review/beta/communication-react.api.md
@@ -182,9 +182,7 @@ export type AzureCommunicationCallWithChatAdapterFromClientArgs = {
     callClient: StatefulCallClient;
     chatClient: StatefulChatClient;
     chatThreadClient: ChatThreadClient;
-    options?: {
-        credential?: CommunicationTokenCredential;
-    };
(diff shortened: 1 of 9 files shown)


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -182,6 +182,9 @@ export type AzureCommunicationCallWithChatAdapterFromClientArgs = {
     callClient: StatefulCallClient;
     chatClient: StatefulChatClient;
     chatThreadClient: ChatThreadClient;
+    options?: {
```
jpeng-ms, line 185: i think we might need to call out this change in storybook as well for `AzureCommunicationCallWithChatAdapterFromClientArgs`

jpeng-ms (approved):

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -1039,7 +1042,9 @@ export const createAzureCommunicationCallWithChatAdapter: ({ userId, displayName
 export const _createAzureCommunicationCallWithChatAdapterFromAdapters: (callAdapter: CallAdapter, chatAdapter: ChatAdapter) => CallWithChatAdapter;
 
 // @public
-export const createAzureCommunicationCallWithChatAdapterFromClients: ({ callClient, callAgent, callLocator, chatClient, chatThreadClient }: AzureCommunicationCallWithChatAdapterFromClientArgs) => Promise<CallWithChatAdapter>;
+export function createAzureCommunicationCallWithChatAdapterFromClients(callClient: StatefulCallClient, callAgent: CallAgent, callLocator: CallAdapterLocator | TeamsMeetingLinkLocator, chatClient: StatefulChatClient, chatThreadClient: ChatThreadClient, options?: {
```
PorterNan, line 1045: Should we add this option to createAzureCommunicationCallWithChatAdapter as well? We should make last param as CallWithChatAdapterOptions instead

Then declare
type CallWithChatAdapterOption = CallAdapterOption && ChatAdapterOption

PorterNan: How does inline image previously work in our callWithChat sample app?

JoshuaLai: We only caught this issue when we just updated selfhost app yesterday. So it technically never worked / tested until today. 
We can for sure create a `type CallWithChatAdapterOption = CallAdapterOption && ChatAdapterOption` instead it might make the implementation a little cleaner. 

We shouldn't need to include it in `createAzureCommunicationCallWithChatAdapter` cause when we user that factory method the chat adapter grabs the credential when it creates the chat client itself and sends it down stream .

packages/communication-react/review/beta/communication-react.api.md
JoshuaLai: Yes thats a good point. I think we created a new section in the storybook to include the factory methods for creating from stateful client, so this can be included in that new section.

packages/react-composites/review/beta/react-composites.api.md
PorterNan: Ohh right! Cause we pass in the token and create credential there

PorterNan (approved):

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -810,6 +811,9 @@ export interface CallWithChatControlOptions extends CommonCallControlOptions {
 // @public
 export type CallWithChatEvent = 'callError' | 'chatError' | 'callEnded' | 'isMutedChanged' | 'callIdChanged' | 'isLocalScreenSharingActiveChanged' | 'displayNameChanged' | 'isSpeakingChanged' | 'callParticipantsJoined' | 'callParticipantsLeft' | 'selectedMicrophoneChanged' | 'selectedSpeakerChanged' | /* @conditional-compile-remove(close-captions) */ 'isCaptionsActiveChanged' | /* @conditional-compile-remove(close-captions) */ 'captionsReceived' | 'messageReceived' | 'messageSent' | 'messageRead' | 'chatParticipantsAdded' | 'chatParticipantsRemoved';
 
+// @beta
+export type CallWithChatOptions = ChatAdapterOptions;
```
JoshuaLai, line 815: @PorterNan So this is what it would look like if we introduced a new options. But since we don't have a `CallAdapterOptions` yet, it isn't a true union yet. So I don't know, when we do create a `CallAdapterOptions` will this be considered a breaking change?

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -853,19 +863,25 @@ export type AzureCommunicationCallWithChatAdapterFromClientArgs = {
...
  * @public
  */
-export const createAzureCommunicationCallWithChatAdapterFromClients = async ({
-  callClient,
-  callAgent,
-  callLocator,
-  chatClient,
-  chatThreadClient
-}: AzureCommunicationCallWithChatAdapterFromClientArgs): Promise<CallWithChatAdapter> => {
+export async function createAzureCommunicationCallWithChatAdapterFromClients(
```
jimchou-dev, line 866: any reason why this change from const to function?

jimchou-dev (approved):

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
JoshuaLai: The underlying functionality is the same, we just need to change it from an arrow function to just this normal function declaration cause it is the only way to include conditional compile flags in the arguments. This is a known limitation, we had to make this change in the ChatAdapter as well

packages/react-composites/review/beta/react-composites.api.md
PorterNan: Just check the code, looks like we have AzureCommunicationCallAdapterOptions, I guess that is a good hint that we should align chatAdapter option name (a name change, we can do it in this PR)

And in the meantime, back to your question - if CallAdapterOption or AzureCommunicationCallAdapterOptions is not needed for now, let's directly use ChatAdapterOptions or AzureCommunicationChatAdapterOptions as the option type, then we can add

export type CallWithChatOptions = AzureCommunicationChatAdapterOptions && AzureCommunicationCallAdapterOptions; later when it is needed, changes like this is not considered as a breaking change

packages/react-composites/review/beta/react-composites.api.md
```diff
@@ -1039,7 +1043,7 @@ export const createAzureCommunicationCallWithChatAdapter: ({ userId, displayName
 export const _createAzureCommunicationCallWithChatAdapterFromAdapters: (callAdapter: CallAdapter, chatAdapter: ChatAdapter) => CallWithChatAdapter;
 
 // @public
-export const createAzureCommunicationCallWithChatAdapterFromClients: ({ callClient, callAgent, callLocator, chatClient, chatThreadClient }: AzureCommunicationCallWithChatAdapterFromClientArgs) => Promise<CallWithChatAdapter>;
+export function createAzureCommunicationCallWithChatAdapterFromClients(callClient: StatefulCallClient, callAgent: CallAgent, callLocator: CallAdapterLocator | TeamsMeetingLinkLocator, chatClient: StatefulChatClient, chatThreadClient: ChatThreadClient, options?: CallWithChatOptions): Promise<CallWithChatAdapter>;
```
PorterNan, line 1046: The option part should be conditional compiled

JoshuaLai: Wait are we sure? Won't that leak out into the stable api?

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -832,6 +833,13 @@ export const useAzureCommunicationCallWithChatAdapter = (
   return adapter;
 };
 
+/* @conditional-compile-remove(teams-inline-images) */
+/**
+ * Configuration options to include when creating CallWithChatAdapter.
+ * @beta
+ */
+export type CallWithChatOptions = ChatAdapterOptions;
```
PorterNan, line 841: Again, if that is just the same type, we can just directly use ChatAdapterOptions, without introducing a new type here for now

PorterNan: Sry, AzureCommunicationChatAdapterOptions, to be more accurate

packages/react-composites/src/composites/CallWithChatComposite/adapter/AzureCommunicationCallWithChatAdapter.ts
```diff
@@ -832,6 +833,13 @@ export const useAzureCommunicationCallWithChatAdapter = (
   return adapter;
 };
 
+/* @conditional-compile-remove(teams-inline-images) */
```
PorterNan, line 836: Don't forget to remove this ;D

JoshuaLai: Oh good catch

PR #2995, opened 2023-04-27
Show video effects errors in the side pane along with the call screen

prabhjot-msft:
# What
Show video effects errors in the side pane along with the call screen
Screenshot 2023-04-27 at 4 04 38 PM

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3242135

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/common/VideoEffectsPane.tsx
```diff
@@ -150,6 +179,14 @@ const VideoEffectsPaneTrampoline = (
       isLightDismiss={true}
       className={mergeStyles(headerStyles)}
     >
+      {showError && latestEffectError && (
```
JamesBurnside, line 182: if we have both a warning and an error - do we want to show both or let the error take precedence?

packages/react-composites/src/composites/common/VideoEffectsPane.tsx
```diff
@@ -130,6 +143,22 @@ const VideoEffectsPaneTrampoline = (
...
+  /* @conditional-compile-remove(video-background-effects) */
+  if (latestEffectError) {
+    if (dismissedError) {
+      showError = latestEffectError?.timestamp > dismissedError.dismissedAt;
+    } else {
+      showError = true;
+    }
+  }
+  /* @conditional-compile-remove(video-background-effects) */
+  useEffect(() => setDismissedError(dismissedError), [dismissedError]);
```
JamesBurnside, line 161: what is the purpose of this useEffect? It looks like any time `dismissedError` changes we call `setDismissedError` (which feels like a no-op as it will set the dimissedError to the value it already is)

packages/react-composites/src/composites/common/VideoEffectsPane.tsx
```diff
@@ -130,6 +143,22 @@ const VideoEffectsPaneTrampoline = (
   selectableVideoEffects?: _VideoEffectsItemProps[],
   onEffectChange?: (effectKey: string) => Promise<void>
 ): JSX.Element => {
+  /* @conditional-compile-remove(video-background-effects) */
+  const [dismissedError, setDismissedError] = useState<DismissedError>();
+  /* @conditional-compile-remove(video-background-effects) */
+  const latestEffectError = useSelector(videoBackgroundErrorsSelector);
+  /* @conditional-compile-remove(video-background-effects) */
+  let showError = false;
```
JamesBurnside, line 151: nitty nit: consider ternary to avoid mutable `let` usage (we opt for const where possible as it lets the reader of the code know this is the only place it gets set and won't have changing values later on)

```ts
const showError = latestEffectError && (latestEffectError.timestamp > dismissedError.dismissedAt);
```

packages/react-composites/src/composites/common/VideoEffectsPane.tsx
prabhjot-msft: In my opinion, we should show both as user has option to dismiss the error and we do want to show warning in case camera is turned off. Lets check with AlexP what does he suggest:
Screenshot 2023-05-01 at 11 51 38 AM

JamesBurnside (approved):

dmceachernmsft (approved):

PR #2988, opened 2023-04-26
Replace fluent n* with js bundle

JamesBurnside:
# What
Replace fluent n* with single js bundle

- Add webpack config to do bundling
  - bundles for both ESM and CJS
- Replace usage of n* with bundle
- Hook up build system to pick up n* wrapper package output

# Why
Short term plan to support react 18 by removing react18 peer dependency (see #1900)

# How Tested

- Bundler should show no significant regression in app size ⚠️ todo: confirm this is reporting correctly/bundler is bundling correctly
- CI snapshots ✅
- All composites in samples apps running dev mode ✅
- All composites and MessageThread in storybook ✅
- All composites in sample apps running production build ✅
- Static html composites ✅
- Locally NPM pack and test package (primarily message thread component in local test app) ✅

PR #2996, opened 2023-04-28
Remove public API usage of ComponentSlotStyle and redefine type internally

JamesBurnside:
# What
Redefine `ComponentSlotStyle` with interoperable type from fluentui

This works because IRawStyle allows any `Record<string, any>` but has defined types for common CSS property (of which ComponentSlotStyle also supports) -- so is just a better version of `Record<string, any>` for styles properties.

# Why
We are removing usage of fluent n-star (see #2988) to support react 18 (see #1900) and we need to not rely on this type.

# How Tested
Checked for interopability:

image
image


emlynmac (approved):
Great step forward to react 9!

packages/react-components/src/types/ComponentSlotStyle.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { IRawStyle } from '@fluentui/react';
+
+/**
```
PorterNan, line 6: Would this be better to put this one in northstar wrapper?

PorterNan (approved):

packages/react-components/src/types/ComponentSlotStyle.ts
JamesBurnside: I did consider that but I'm actually against that as we shouldn't ever have publicly exported a fluent northstar interface and should have defined our own interface/used regular fluents in this package.

JamesBurnside: (It also makes the bundling more difficult if we have a public API coming out of the northstar wrapper the way we rollup the api)

PR #3005, opened 2023-05-01
`react-aria-live` react 18 support

JamesBurnside:
# What
Move react-aria-live to our repo (+small conversion to functional component style)

# Why
Package has a hard peer dep on react 16. We are moving to support react 18 (see: #1900)

# How Tested
* ✅ unit tests added
* ✅ Local testing appears to work as best I can tell:
   | before | after |
   | -- | -- |
   | image | image |
   | image | image |


dmceachernmsft (approved):
This is really cool! this is just for the message thread? is this announcer (since it is much more powerful than the one I made) something we want to start adapting into new components in the furture/update existing announcers?

JamesBurnside:
> This is really cool! this is just for the message thread? is this announcer (since it is much more powerful than the one I made) something we want to start adapting into new components in the furture/update existing announcers?

We can reuse this throughout. I just super quickly copy-converted over the underlying library as it was very small so should be essentially a no-op but we can definitely integrate this into calling as well to avoid multiple announcement. Honestly though we need a stateful-driven announcer instead of the this being at the react-component layer as the react-components don't have context of the whole app to make a good, informed decision about whether to announce or not

emlynmac (approved):

PR #3006, opened 2023-05-01
[Click-to-call] Update video gallery to have better local video tile control

dmceachernmsft:
# What

Introduce API to control the style of local tile rendered in the video gallery
# Why

Allows for control of the video gallery based on different form factors. Also allows the ability to hide the local tile if Contoso is creating a Audio only experience.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3247127
# How Tested

Hidden:
image
Tiny video gallery:
image


packages/react-components/review/beta/react-components.api.md
```diff
@@ -1253,6 +1253,9 @@ export const _LocalVideoTile: React_2.MemoExoticComponent<(props: {
     personaMinSize?: number | undefined;
 }) => JSX.Element>;
 
+// @beta
+export type LocalVideoTileMode = 'mobile' | 'desktop' | 'hidden' | 'responsive';
```
JamesBurnside, line 1257: for hidden, can we just make the localParticipant optional? Or do other video gallery things rely on that?

JamesBurnside, line 1257: `desktop`/`mobile`/`responsive` are vague -- what does it mean to have a tile mobile vs desktop? Could we make the naming more inline with the behavior instead of the platform we expect them to be on?

dmceachernmsft: I like the updates around using the aspect ratio, but for 'matchStream' I think is a little misleading since it isn't changing to match the stream size, its still responding to the size observers making it responsive today. would `'16:9' | '9:16' | 'responsive'` be appropriate?

dmceachernmsft: unfortunately we do provide the local participant to get their screen share. there is a case that Contoso could provide a audio only experience but still allow a screen share stream so we cannot just remove the participant.

JamesBurnside: I guess my question is why is it responsive? Wouldn't it make more sense to always match the incoming stream? (If they are on mobile and portrait vs landscape, their stream will then be indicative of the mobile size)

JamesBurnside: On smaller screens the tile itself could be smaller but that would be tangential to the aspect ratio

JamesBurnside: Oh because we can't always match it if there video is off, hmm

dmceachernmsft: actually match stream makes more sense thinking about it will make these updates.

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -243,6 +251,14 @@ export interface VideoGalleryProps {
    * @defaultValue 'HorizontalBottom'
    */
   overflowGalleryPosition?: OverflowGalleryPosition;
+  /* @conditional-compile-remove(click-to-call) */
+  /**
+   * Determines the aspect ratio of local video tile in the video gallery.
+   * @remarks 'matchStream' will be responsive to the screen orientation and will change between 9:16 (portrait) and
```
JamesBurnside, line 257: For matchstream, what is the aspect ratio when the camera is off?

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -131,6 +131,8 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
         remoteVideoTileMenuOptions={remoteVideoTileMenuOptions}
         /* @conditional-compile-remove(vertical-gallery) */
         overflowGalleryPosition={overflowGalleryPosition}
+        /* @conditional-compile-remove(click-to-call) */
+        localVideoTileSize={props.isMobile ? 'matchStream' : '16:9'}
```
JamesBurnside, line 135: why not matchStream on desktop as well?

packages/react-components/src/components/VideoGallery.tsx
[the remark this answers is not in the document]
dmceachernmsft: Yes from my understanding through the only time the stream aspect ratio will change from a portrait view on mobile to landscape view is when the device orientation changes. the stream updates from 9:16 to 16:9 when you rotate the device. should I update the comment so it is referring to the stream here even though we are truthfully in the implementation using the device aspect ratio? 

there currently is not a good way for us to get the aspect ratio in a performant way from the SDK. media stats API is not a good solution here in my opinion since we have to set up something to ping the stats every so often so the device could rotate and the tile wont update for a second or two depending on the interval set.

dmceachernmsft: it will match the orientation like before. since we aren't getting a stream here it would be using the device information.

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
dmceachernmsft: desktop web cams do not know when the stream has been rotated and the standard aspect ratio for most camera's is 16:9 so we want to enforce this aspect ratio with desktop mode.

packages/react-components/src/components/VideoGallery.tsx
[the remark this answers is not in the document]
JamesBurnside: Yeah, matchStream is a bit of lie then, matchDevice might be more accurate.. but the aspect ratio doesn't match the device, we just estimate it.
I would rename to matchDevice if we can exactly match the device, perhaps `followDeviceOrientation` ?

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
JamesBurnside: Yeah makes sense, if it was matching the stream aspect ratio I would say pick matchStream but as its following orientation then yeah keep 16:9

packages/react-components/src/components/VideoGallery/FloatingLocalVideoLayout.tsx
```diff
@@ -127,12 +135,12 @@ export const FloatingLocalVideoLayout = (props: FloatingLocalVideoLayoutProps):
   const layerHostId = useId('layerhost');
 
   const localVideoSizeRem = useMemo(() => {
-    if (isNarrow) {
+    if (isNarrow && /*@conditional-compile-remove(click-to-call) */ !(localVideoTileSize === '16:9')) {
```
edwardlee-msft, line 138: nit: We use '16:9' more than once, do want to make this a constant instead?

edwardlee-msft (approved):

packages/react-components/src/components/VideoGallery.tsx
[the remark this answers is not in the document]
dmceachernmsft: I like the `followDeviceOrientation` I will make that update I think that makes sense

JamesBurnside (approved):

PR #3009, opened 2023-05-02
[Ready for merge] Add breaking change check CI

PorterNan:
# What

Add breaking change check CI

One intentional breaking demo:
https://github.com/Azure/communication-ui-library/actions/runs/4875247588/jobs/8697210004

# Why

# How Tested

# Process & policy checklist


.github/workflows/breaking-change-check.yml
```diff
@@ -0,0 +1,96 @@
+name: Breaking Change Check
```
JamesBurnside, line 1: Is it necessary to be its own action? (Does it take a long time to perform the check?) Or could we just have this in the CI.yml where we have the treeshaking and typescript checks?

JamesBurnside: Oh because it builds base as well as current then compares - sounds good!

PorterNan: I guess we might need more machines if that is possible... cause now we have 2 teams conributing to the repo, imagine if we have more teams but we still don't have enough machines to run all PRs

packages/communication-react/breaking-change-check/CompareChanges.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+export type CheckBreakingChanges<OLD, NEW extends OLD> = NEW;
+
+export type ExcludeList =
```
PorterNan, line 5: These are the list we intentionally make breaking changes on, we will still check removed property and compatible property names/types
But it will skip the check for new properties in the interface (and we don't consider them as breaking change)

.github/workflows/breaking-change-check.yml
```diff
@@ -0,0 +1,96 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.get-matrix.outputs.matrix }}
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE so job can access it
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      # Ensure node version is great enough
+      - name: Use Node.js v14.x
```
emlynmac, line 31: Should we be using 16 by default now?

.github/workflows/breaking-change-check.yml
```diff
@@ -0,0 +1,96 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: ${{ fromJSON(needs.get_matrix.outputs.matrix) }}
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE so job can access it
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      # Ensure node version is great enough
+      - name: Use Node.js v14.x
```
emlynmac, line 51: 14 is quite old; 18 is the LTS - can we use that instead?

packages/communication-react/breaking-change-check/checkBreakingChanges.ts
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/* eslint-disable @typescript-eslint/ban-types */
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
emlynmac, line 5: Why is this linter warning turned off? unused-vars seems like an easy thing to fix up

PorterNan: We don't use those variable any where, those assignment just for a purely check of compatibility of "current" and "base"

.github/workflows/breaking-change-check.yml
PorterNan: Yep, will align with what we have for CI (version 16) then we might need a fully test for all yml to bump it to 18

packages/communication-react/breaking-change-check/CompareChanges.ts
JamesBurnside: Add something to this effect as a doc comment to this file/export

JamesBurnside (approved):

.github/workflows/breaking-change-check.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: Restore node_modules from cache
+        uses: actions/cache@v3
+        with:
+          path: common/temp/pnpm-store
+          key: ${{ runner.os }}-${{ hashFiles('common/config/rush/pnpm-lock.yaml') }}
+      # Install dependencies
+      - name: Install rush
+        run: npm install -g @microsoft/rush@5.47.0
+      - name: Install dependencies
+        run: rush install
```
jimchou-dev, line 65: Maybe add retry attempts
`rush install --max-install-attempts 3`

PR #3013, opened 2023-05-02
[bugfix] Resolve overlap between tooltip and list of devices on mobile configuration page

edwardlee-msft:
# What
Disable tooltips on buttons ONLY on mobile configuration page

# Why
Button tooltips in mobile configuration page overlaps with the list of devices.

# How Tested
iOS safari using Calling sample

# Process & policy checklist


dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/components/LocalPreview.tsx
```diff
@@ -114,13 +114,28 @@ export const LocalPreview = (props: LocalPreviewProps): JSX.Element => {
...
+              props.mobileView ? { tooltipOnContent: '', tooltipOffContent: '', tooltipDisabledContent: '' } : {}
+            }
             styles={localPreviewButtonStyle}
           />
           <CameraButton
             data-ui-id="call-composite-local-device-settings-camera-button"
             {...cameraButtonProps}
             showLabel={true}
             disabled={!cameraPermissionGranted || !hasCameras}
+            // disable tooltip as it obscures list of devices on mobile
```
JamesBurnside, line 128: from an a11y point of view, is this OK to do if users are using a bluetooth keyboard on mobile?

edwardlee-msft: I can neither confirm nor deny if this would be an issue with A11y. 
However, my understanding is, tooltips are good idea if the buttons do not have labels. In this case, on the configuration page for mobile, the buttons do have labels.

edwardlee-msft: Also a small blurp about tooltips.
"Although tooltips are mainly limited to desktop computers and laptops, they do have a sister element that is common on touchscreen devices: popup tips. Both tooltips and popup tips have the same goal: to provide helpful, additional content."

JamesBurnside: Ah excellent point! 10/10 response, resolving comment and heartily approving

JamesBurnside (approved):

PR #3016, opened 2023-05-03
Alkwa/update node references

alkwa-msft:
# What
There are a wide range of old versions of Node. We are updating all of our references to 16.19.0

# Why
We want to snap to something that is more modern as this has raised issue with some developers. 

# How Tested
Built locally.


JamesBurnside (approved):

common/config/rush/pnpm-lock.yaml
```diff
@@ -135,7 +135,7 @@ packages:
     dependencies:
       '@azure/core-http': 2.3.2
       '@azure/core-tracing': 1.0.0-preview.13
-      '@azure/logger': 1.0.4
```
JoshuaLai, line 138: Is downgrading the logger intentional?

JoshuaLai (approved):

PR #3019, opened 2023-05-03
Update localization strings for video effects

JamesBurnside:
# What
Update localization strings for video effects

# Why
Needed done

# How Tested
Ran locally and checked en-us was picked up correctly:

image

image


prabhjot-msft (approved):

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/Localization.test.ts-snapshots/localized-call-configuration-page-Desktop-Chrome-linux.png
emlynmac, line 1: In the new screenshot, the French _Effets_ has been replaced with English _Effects_.
Is this intended or something not working?

JamesBurnside: Yeah, because I've changed the name of the effects property it hasn't yet been localized, once we have the localized strings this will be french again.
But this reminded me I should've removed the old string from the other loc files! Will update

emlynmac: So is the process here to wait for the localization as it affects the screenshot or to update the screenshot once we have the localization?

JamesBurnside: Really we should update the e2e tests to use a mocked locale to prevent things like this and remove the dependency on touchdown, but so far we've just been checking in the letting it update in time (which is bad because it doesn't get reviewed at PR time)

edwardlee-msft (approved):

PR #3020, opened 2023-05-03
Filter the errors based on whether the side pane is currently showing the video effects side pane

prabhjot-msft:
# What
Filter the errors based on whether the side pane is currently showing the video effects side pane

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3242143

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -206,7 +218,25 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
   const forceShowingCheckPermissions = !minimumFallbackTimerElapsed;
 
   /* @conditional-compile-remove(video-background-effects) */
-  const { toggleVideoEffectsPane } = useVideoEffectsPane(props.updateSidePaneRenderer, mobileView);
+  const [dismissedVideoEffectsError, setDismissedVideoEffectsError] = useState<DismissedError>();
```
JamesBurnside, line 221: to avoid this copy pasted between callarrangement and configuration page, could this logic be moved to the `useVideoEffectsPane`?

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -206,7 +218,25 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
+  /* @conditional-compile-remove(video-background-effects) */
+  const onDismissVideoEffectError = useCallback((error: AdapterError) => {
+    setDismissedVideoEffectsError(dismissVideoEffectsError(error));
+  }, []);
+  /* @conditional-compile-remove(video-background-effects) */
+  const latestVideoEffectError = useSelector(videoBackgroundErrorsSelector);
+  /* @conditional-compile-remove(video-background-effects) */
+  const showVideoEffectError =
```
JamesBurnside, line 229: nit: instead of passing `showError : true | false` to the videoeffectspane and having the video effects pane have to use a selector to retreive the error, the errors are already retrieved here so the logic would be streamlined if this passed the error to videoPane instead of a showError boolean

JamesBurnside (approved):
minor suggestions but looks great!

carocao-msft (approved):

PR #3021, opened 2023-05-04
Fixed start captions console error and add loading screen 

carocao-msft:
# What
Screenshot 2023-05-03 at 6 54 36 PM

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3235039

# How Tested
Tested on samples 

# Process & policy checklist


edwardlee-msft (approved):

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -101,6 +101,10 @@ export interface CaptionsCallFeatureState {
    * whether captions is on/off
    */
   isCaptionsFeatureActive: boolean;
+  /**
+   * whether start captions button is clicked or now
+   */
+  startCaptionsClicked: boolean;
```
PorterNan, line 107: It could be a better name, it is not just representing the UI clicked, but also when someone call the function

Maybe like:
startCaptionsInProgress would work

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -749,6 +749,16 @@ export class CallContext {
...
+
+  /* @conditional-compile-remove(close-captions) */
+  setStartCaptionsInProgress(callId: string, startCaptionsInProgress: boolean): void {
+    this.modifyState((draft: CallClientState) => {
+      const call = draft.calls[this._callIdHistory.latestCallId(callId)];
+      if (call) {
+        call.captionsFeature.startCaptionsInProgress = startCaptionsInProgress;
+      }
+    });
+  }
```
PorterNan, line 761: Add extra line here

packages/calling-stateful-client/src/CallDeclarativeCommon.ts
```diff
@@ -122,6 +123,7 @@ class ProxyTeamsCaptionsFeature implements ProxyHandler<TeamsCaptionsCallFeature
           async (...args: Parameters<TeamsCaptionsCallFeature['stopCaptions']>) => {
             const ret = await target.stopCaptions(...args);
             this._context.setIsCaptionActive(this._call.id, false);
+            this._context.setStartCaptionsInProgress(this._call.id, false);
```
PorterNan, line 126: Is it possible that caption will be always in progress and never return back to normal? How do we handle scenarios like this?

packages/react-composites/src/composites/common/CaptionsBannerMoreButton.tsx
```diff
@@ -51,15 +51,20 @@ export const CaptionsBannerMoreButton = (props: CaptionsBannerMoreButtonProps):
   );
   /* @conditional-compile-remove(close-captions) */
   const moreButtonContextualMenuItems: IContextualMenuItem[] = [];
+  /* @conditional-compile-remove(close-captions) */
```
PorterNan, line 54: Same here, please refer to the "shouldSetSpokenLanguage" comment above

packages/calling-stateful-client/src/CallDeclarativeCommon.ts
carocao-msft: That means start captions failed and it will stay on the in progress screen, will see error in console too similar to what happens when start call fail

packages/react-components/src/components/CaptionsSettingsModal.tsx
```diff
@@ -90,15 +90,20 @@ export const _CaptionsSettingsModal = (props: _CaptionsSettingsModalProps): JSX.
     }
   }, [onDismissCaptionsSettings]);
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
carocao-msft: My idea is since spoken language doesn't change, it's ok if setspokenlanguage is called multiple times. Calling setspokenlanguage(english) twice will not change the spoken language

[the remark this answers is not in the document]
PorterNan: That is true for 1 customer scenario, but imagine if you have 2 more customers, it could be an easy raise condition to trigger this regression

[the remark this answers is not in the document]
carocao-msft: Since we always require the modal to set language when starting captions, we don't have a scenario where selectedItem.key is different across participants?

packages/react-components/src/components/StartCaptionsButton.tsx
```diff
@@ -93,16 +93,25 @@ export const _StartCaptionsButton = (props: _StartCaptionsButtonProps): JSX.Elem
...
   }, [currentSpokenLanguage]);
 
+  const [hasSetSpokenLanguage, setHasSetSpokenLanguage] = useState(false);
+
   const onToggleStartCaptions = useCallback(async (): Promise<void> => {
     if (props.checked) {
       onStopCaptions();
     } else {
       await onStartCaptions(options);
+    }
```
PorterNan, line 103: This code is kinda duplicated for 2 components - I guess we just need it living in one place then we should be good? A place that wouldn't be disposed during the composite life cycle

packages/react-composites/src/composites/common/CaptionsBannerMoreButton.tsx
```diff
@@ -50,16 +50,30 @@ export const CaptionsBannerMoreButton = (props: CaptionsBannerMoreButtonProps):
     [localeStrings]
   );
   /* @conditional-compile-remove(close-captions) */
+  const [hasSetSpokenLanguage, setHasSetSpokenLanguage] = useState(false);
+  /* @conditional-compile-remove(close-captions) */
   const moreButtonContextualMenuItems: IContextualMenuItem[] = [];
+  /* @conditional-compile-remove(close-captions) */
+  useEffect(() => {
```
PorterNan, line 57: Yea... the third place we have the same logic

packages/react-components/src/components/StartCaptionsButton.tsx
PorterNan: Cause it is not depending on the click of the button anymore - it only depends on setHasSetSpokenLanguage and isCaptionsFeatureActive

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/ClosedCaptions.test.ts-snapshots/caption-toggle-button-Desktop-Chrome-linux.png
PorterNan, line 1: There is some visual change here, I guess that is by deisgned?

PorterNan (approved):

PorterNan (approved):

PR #3024, opened 2023-05-04
Fix bug where error bar overlay with pane

carocao-msft:
# What
Fix bug where error bar overlay with pane

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3241539

# How Tested
Screenshot 2023-05-04 at 1 35 22 PM

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -149,7 +149,8 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
       display: isMobileWithActivePane ? 'none' : 'flex',
       minWidth: props.mobileView ? 'unset' : `${compositeMinWidthRem}rem`,
       width: '100%',
-      height: '100%'
+      height: '100%',
+      position: 'relative'
```
JamesBurnside, line 153: nice fix! Can we add an e2e test for this? It would be very easy for someone to refactor this and accidental regress

JamesBurnside (approved):

dmceachernmsft (approved):

PR #3028, opened 2023-05-05
Update configuration side pane styles and close on start call

JamesBurnside:
# What

- Put side pane in a panel so it overlaps content per figma
- Close the side pane when the start call is clicked

# How Tested

image


packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -331,13 +345,23 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
                 className={mobileWithPreview ? startCallButtonStyleMobile : undefined}
-                onClick={startCallHandler}
+                onClick={startCall}
                 disabled={disableStartCallButton}
               />
             </Stack>
           </Stack>
         </Stack>
-        <SidePane mobileView={props.mobileView} updateSidePaneRenderer={props.updateSidePaneRenderer} />
+        <Panel
```
prabhjot-msft, line 354: Are we switching back to fluent panel for Config screen? In case we would like to switch content for side pane in future, we will have to introduce new panel for that.

dmceachernmsft, line 354: This panel also seems to overlap the whole window, do we have a way to make sure it only overlaps on the composite screen and not all of Contoso's application? I believe I brought this up as a possible bug @prabhjot-msft?

JamesBurnside: There's a layer host property set to scope the panel to the composite - lmk if this doesn't solve what you're referring to

JamesBurnside: Yeah the figma's show it above the content so this was simplest just to host the side pane in a panel.
Sorry not following the second part of the comment - what do you mean by "we will have to introduce new panel for that"?

prabhjot-msft: If we plan on adding another panel content for example for sound effects, would we need to create a new panel for sound effects or we would be able to override content in the same panel

JamesBurnside: Nope same panel, this just adjusts how it is hosted in the configuration pane, the content is tangential to that

JamesBurnside: (on re-looking, the isopen param would need to be updated though but I'll leave as is for now)

prabhjot-msft: Sounds good

mgamis-msft:
Looks great!

mgamis-msft (approved):

PR #3030, opened 2023-05-05
[Chat] Adding new unit test testing factory method from state

JoshuaLai:
# What
Adding new test to check for construction from stateful client
# Why
Testing all entry points from the adapter 

# How Tested
Unit test
In the package: `npm run test AzureCommunicationChatAdapter.test.ts `

# Process & policy checklist


jimchou-dev (approved):

packages/react-composites/src/composites/CallWithChatComposite/adapter/TestUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+import { GroupChatCallLocator, MeetingLocator, RoomLocator, PushNotificationData } from '@azure/communication-calling';
+import {
+  CallState,
+  CallClientState,
+  StatefulCallClient,
+  createStatefulCallClient,
+  CallErrors,
+  CreateViewResult
+} from '@internal/calling-stateful-client';
+import EventEmitter from 'events';
```
PorterNan, line 41: Add js doc @private to it or remove the comment

packages/react-composites/src/composites/ChatComposite/adapter/StubChatClient.ts
```diff
@@ -59,6 +64,78 @@ export class StubChatClient implements PublicInterface<ChatClient> {
...
+  const mockEventHandlersRef = { value: {} };
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const mockChatClient: ChatClient = {} as any;
+
+  mockChatClient.createChatThread = async (request) => {
+    return {
+      chatThread: {
+        id: 'chatThreadId',
+        topic: request.topic,
+        createdOn: new Date(0),
```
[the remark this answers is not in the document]
PorterNan: Date.now() might be creating a flaky result for every render, a static number would be nice (0 feels a bit too early tbh), initialize a real date with UTC string would be cool, but not 100% necessary

packages/react-composites/src/composites/CallWithChatComposite/adapter/TestUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+/**
+ *
+ */
+export class MockCallClient {
```
PorterNan, line 45: We don't have a mock client implementation before? If so, this is a nice one! Have you tried this one:
https://jestjs.io/docs/es6-class-mocks#replacing-the-mock-using-mockimplementation-or-mockimplementationonce

so this will allow you only mock those functions really need to be implemented for your code

JoshuaLai: It looks like he have written similar implementations, not exactly the same, but they are in the statefulclient package.

packages/react-composites/src/composites/ChatComposite/adapter/StubChatClient.ts
```diff
@@ -59,6 +64,78 @@ export class StubChatClient implements PublicInterface<ChatClient> {
     return;
   }
 }
+/**
+ * @returns
+ */
+export const createStatefulChatClientMock = (threadClient: PublicInterface<ChatThreadClient>): StatefulChatClient => {
```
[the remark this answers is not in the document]
JoshuaLai: Moved these to the mock folder. But looking at this a bit closer, we seem to have some mocks in the composite folder and some  random ones in the mock folder. So perhaps we need a refactor task to just clean up where things belong

Leah-Xia-Microsoft (approved):

PR #3036, opened 2023-05-10
[Click to Call] New API for hiding local tile in composite

dmceachernmsft:
# What

Allows contoso to set CallCompositeOptions to choose the position of the local video tile in the composite
# Why

Further expands customizability of the Call and CallWithChat composites

https://skype.visualstudio.com/SPOOL/_workitems/edit/3247128
# How Tested

Ran locally, tests to come if API is approved


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2679,6 +2680,11 @@ export interface LocalVideoStreamVideoEffectsState {
     isActive: boolean;
 }
 
+// @beta
+export interface LocalVideoTileOptions {
```
JamesBurnside, line 2684: Should LocalVideoTileSize be moved into here as that is also a LocalVideoTile option

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2679,6 +2680,11 @@ export interface LocalVideoStreamVideoEffectsState {
...
+    position?: 'grid' | 'floating' | 'hidden';
+}
+
 // @beta
 export type LocalVideoTileSize = '9:16' | '16:9' | 'hidden' | 'matchStream';
```
JamesBurnside, line 2689: `hidden` is now in two places, can we remove hidden from here now?

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -102,6 +102,19 @@ export interface RemoteVideoTileMenuOptions {
...
+
+/* @conditional-compile-remove(click-to-call) */
+/**
+ * Options for the local video tile in the Call composite.
+ *
+ * @beta
+ */
+export interface LocalVideoTileOptions {
+  /**
+   * Position of the local video tile. If unset will render the local tile in the floating local position.
```
JamesBurnside, line 114: When overflow gallery is showing `floating` isn't quite right as its docked to the bottom right

JamesBurnside, line 114: Can we somehow make clear this only affects the call screen and not the config page nor the PiPiP

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -102,6 +102,19 @@ export interface RemoteVideoTileMenuOptions {
...
+   */
+  position?: 'grid' | 'floating' | 'hidden';
```
JamesBurnside, line 116: what does grid refer to with respect to grid vs overflow gallery?

JamesBurnside: And does floating mean it will always be floating and never docked anymore?

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: Oo this is just for composites

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: never mind the new property is just for composites

mgamis-msft (approved):

mgamis-msft (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
dmceachernmsft: So this might be a naming issue with the videoGallery layouts but it will follow the `default layout` when `grid` is selected. so anything that that local tile would do in the default layout will be true here.

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
dmceachernmsft: yup I will update the comment! thanks for pointing this out.

packages/communication-react/review/beta/communication-react.api.md
dmceachernmsft: we still need to pass it into the local tile options in the component form the composite. I am not sure how we could pull it out.

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
edwardlee-msft, line 116: Can we have comments explicit on what grid and floating mean? 
Position wise, where is it, where does grid place it. 
If grid is default layout, should we just name it `default`?

dmceachernmsft: I think mentioning it in the comments where it will land is correct, but for the composite default is 'floating' so we can't use that terminology here.

edwardlee-msft (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -102,6 +102,26 @@ export interface RemoteVideoTileMenuOptions {
...
+ *
+ * @beta
+ */
+export interface LocalVideoTileOptions {
+  /**
+   * Position of the local video tile. If unset will render the local tile in the floating local position.
+   *
+   * @defaultValue 'floating'
+   * @remarks 'grid' - local video tile will be rendered in the grid view of the videoGallery.
+   * 'floating' - local video tile will be rendered in the floating position and will observe overflow gallery.
```
JamesBurnside, line 118: nit remove fullstop as sentence continues on next line?
```suggestion
   * 'floating' - local video tile will be rendered in the floating position and will observe overflow gallery
```

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -123,7 +132,7 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
         localVideoViewOptions={localVideoViewOptions}
         remoteVideoViewOptions={remoteVideoViewOptions}
         styles={VideoGalleryStyles}
-        layout="floatingLocalVideo"
+        layout={layoutBasedOnTilePosition}
```
JamesBurnside, line 135: actually, here the local video tile is rewiring the entire layout including all remote tiles

JamesBurnside (changes requested):

JamesBurnside (approved):

PR #3038, opened 2023-05-10
Fix lint command issue of "No files matching the pattern "/*/**/*.tsx" were found" on Mac OS

Leah-Xia-Microsoft:
# What

On Mac OS 13.3.1, rush lint fails to run due to an error of "No files matching the pattern "/*/**/*.tsx" were found".
This PR fixes this issue.

# Why

# How Tested

CI pipeline of the upcoming PRs should pass

# Process & policy checklist

### DIFF
--- a/change/@azure-communication-react-0318bc34-55d1-4d98-a106-ee5dbbf0deca.json
+++ b/change/@azure-communication-react-0318bc34-55d1-4d98-a106-ee5dbbf0deca.json
@@ -1,7 +1,0 @@
-{
-  "type": "patch",
-  "comment": "Fix CI issue of \"No files matching the pattern \"*/**/*.tsx\" were found\"",
-  "packageName": "@azure/communication-react",
-  "email": "107075081+Leah-Xia-Microsoft@users.noreply.github.com",
-  "dependentChangeType": "patch"
-}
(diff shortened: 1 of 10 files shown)


PorterNan (changes requested):
We only lint ts, tsx files, js is not our target code, and we don't usually lint those files under root folder, this change only hide the issue because it makes the command to find some js file, but it doesn't fully solve the real problem

change/@azure-communication-react-0318bc34-55d1-4d98-a106-ee5dbbf0deca.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix CI issue of \"No files matching the pattern \"*/**/*.tsx\" were found\"",
+  "packageName": "@azure/communication-react",
+  "email": "107075081+Leah-Xia-Microsoft@users.noreply.github.com",
+  "dependentChangeType": "patch"
```
PorterNan, line 6: Anything does not impact our production code would be considered as "none"

Only those PRs patch the ui library package would be type "patch"

PorterNan (approved):
Synced with Leah offline and she had a full test of this command (thanks!)

The command should be running well in CI, but not compatible with Mac OS, adding js is to align with other repos, and it does catch the lint error for those files in src folder

vhuseinova-msft (approved):

palatter (approved):

PR #3042, opened 2023-05-11
[FHL] Add incoming calls to the Call adapter

dmceachernmsft:
# What

image
Add a incoming call adapter to the internals of the call adapter
# Why

This allows us to use the Call and CallWithChat adapters to accept incoming calls.

Contoso would have to subscribe to the incoming Call event and get the incoming calls out of the adapter state and create their own UI (or copy pasta our example in storybook - seems pretty popular!) 
image
Contoso could do this to auto switch into any new calls.

# How Tested

Ran locally <- found bug where caller gets stuck in connecting state will look to see if thats an us problem.
image

UPDATE: fixed the above issue. If you process the same call twice the adapter gets confused and stuck in the last state the call holds. Fix was to only process a new call in the updated calls event if the direction of the call was `incoming` so that the adapter will switch to the new call.


edwardlee-msft:
Ohhhh fun!!!

PR #3048, opened 2023-05-15
 Blur video background effect tests

prabhjot-msft:
# What
 Blur video background effect tests

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3230841
https://skype.visualstudio.com/SPOOL/_workitems/edit/3230845

# How Tested

# Process & policy checklist


packages/react-composites/tests/app/lib/MockCallAdapter.ts
```diff
@@ -152,7 +152,24 @@ export class MockCallAdapter implements CallAdapter {
   }
 
   blurVideoBackground(): Promise<void> {
-    throw new Error('blurVideoBackground not implemented.');
+    this.modifyState((draft: CallAdapterState) => {
+      if (!draft.call) {
```
JamesBurnside, line 156: if there's no call, this should set the unparented video stream of the devicemanager

packages/react-composites/tests/app/lib/MockCallAdapter.ts
```diff
@@ -152,7 +152,24 @@ export class MockCallAdapter implements CallAdapter {
...
+        throw new Error('Call not found');
+      } else if (draft.call.localVideoStreams.length > 0) {
+        draft.call.localVideoStreams = [
```
JamesBurnside, line 159: here `blurBackground` is setting the video stream to something new. Should it not only be updating fields related to video effects?

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -273,6 +279,7 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
         split={props.split ?? props.enableDeviceSelectionMenu}
         aria-roledescription={props.enableDeviceSelectionMenu ? strings.cameraButtonSplitRoleDescription : undefined}
         splitButtonAriaLabel={props.enableDeviceSelectionMenu ? splitButtonAriaString : undefined}
+        splitButtonMenuProps={splitButtonMenuProps}
```
JamesBurnside, line 282: if someone was previously passing in `splitButtonMenuProps` this will overwright that, could this instead be:
```suggestion
        splitButtonMenuProps={props.splitButtonMenuProps ?? splitButtonMenuProps}
```

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -248,6 +250,10 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
     }
   };
 
+  const splitButtonMenuProps: IButtonProps = {
```
JamesBurnside, line 253: avoid defining consts on the render method, pull this outside the render body (otherwise use useMemo)

mgamis-msft: +1

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/VideoEffects.test.ts-snapshots/video-effects-side-pane-blur-selected-camera-on-Desktop-Chrome-linux.png
JamesBurnside, line 1: noted the bug here on the alignment 👍 great snapshots to catch this!

packages/react-composites/tests/app/lib/MockCallAdapter.ts
mgamis-msft: The blurVideoBackground adapter method will call the calling sdk method which can't be done for the MockCallAdapter. I guess it will be hard to mock without directly setting the video stream in the call adapter state.

packages/react-components/src/components/CameraButton.tsx
prabhjot-msft: done

packages/react-composites/tests/app/lib/MockCallAdapter.ts
prabhjot-msft: done

packages/react-composites/tests/app/lib/MockCallAdapter.ts
prabhjot-msft: Rather than replacing the video stream, I am replacing the view now.

packages/react-components/src/components/CameraButton.tsx
prabhjot-msft: done

packages/react-components/src/components/CameraButton.tsx
```diff
@@ -248,6 +250,14 @@ export const CameraButton = (props: CameraButtonProps): JSX.Element => {
     }
   };
 
+  const splitButtonMenuProps: IButtonProps = useMemo(
+    () => ({
+      ...props.splitButtonMenuProps,
+      className: 'camera-split-button'
```
JamesBurnside, line 256: swap these two lines around, otherwise we overwright the className passed in (and we are only using the className for ui test)

JamesBurnside (approved):

mgamis-msft (approved):

PR #3050, opened 2023-05-15
Merge React 18 feature branch

JamesBurnside:
# What

Merge of base PRs:

- #2988
- #3005
- #3045

# Why

Merge short term solution for react 18, see https://github.com/Azure/communication-ui-library/discussions/2939 & https://github.com/Azure/communication-ui-library/issues/1900.

# How Tested

Selfhosting and bug bashes.


packages/storybook/stories/storybook.test.ts
```diff
@@ -45,7 +45,7 @@ afterAll(() => {
   console.error = consoleError;
 });
 
-describe('storybook snapshot tests', () => {
+describe.skip('storybook snapshot tests', () => {
```
JamesBurnside, line 48: these are not working on react 18, they fail with a non-descript "I give up" error ☹️

packages/acs-ui-common/package.json
```diff
@@ -34,16 +34,16 @@
...
+    "@types/react": ">=16.8.0 <19.0.0"
   },
   "devDependencies": {
     "@babel/cli": "~7.16.0",
     "@babel/core": "~7.16.0",
     "@microsoft/api-documenter": "~7.12.11",
     "@microsoft/api-extractor": "~7.18.0",
     "@types/jest": "~26.0.22",
-    "@types/react": "^16.9.49",
+    "@types/react": "18.2.0",
```
PorterNan, line 46: Do we need to run react 16 compatibility test by any chance?
BTW, why not use ^ to get latest update for react 18

packages/communication-react/package.json
```diff
@@ -65,8 +63,9 @@
     "copy-preprocess": "cpx \"../*/preprocessed/**\" ./preprocess-dist && renamer --find preprocessed --replace src ./preprocess-dist/**/*",
     "build": "rushx check-deps && rushx _by-flavor \"rushx _build:by-flavor\"",
     "build:cjs": "rollup -c --silent",
-    "build:esm": "rushx _if-preprocess && rushx copy-original && rushx preprocess && rushx copy-preprocess  && ttsc -project tsconfig.preprocess.json || (if-env COMMUNICATION_REACT_FLAVOR=beta && ttsc)",
+    "build:esm": "rushx _if-preprocess && rushx copy-original && rushx preprocess && rushx copy-preprocess  && ttsc -project tsconfig.preprocess.json || (if-env COMMUNICATION_REACT_FLAVOR=beta && ttsc) && rushx copy-northstar",
     "build:watch": "",
+    "copy-northstar": "cpx \"../northstar-wrapper/dist/dist-esm/**/*\" dist/dist-esm/northstar-wrapper/src --clean",
```
PorterNan, line 68: Just out of curiosity, why do we need to do this copy and paste?

packages/northstar-wrapper/src/index.ts
```diff
@@ -1,8 +1,19 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import * as northStar from '@fluentui/react-northstar';
+import {
```
PorterNan, line 4: Any concerns that we don't do import * ? is it for treeshaking

packages/northstar-wrapper/package.json
```diff
@@ -3,36 +3,43 @@
   "version": "1.5.1-beta.4",
   "private": "true",
   "description": "Package to wrap the Fluent NorthStar UI library, for short-term React 18 compatibility",
+  "module": "dist/dist-esm/index.js",
```
PorterNan, line 6: I guess we will generate both esm and cjs code in our npm package? How much bigger our npm packages will be in the end

packages/northstar-wrapper/package.json
```diff
@@ -3,36 +3,43 @@
...
+  "main": "dist/dist-cjs/index.js",
+  "types": "dist/northstar-wrapper.d.ts",
   "scripts": {
-    "build": "tsc",
+    "build": "rushx build:esm && rushx build:cjs",
+    "build:esm": "tsc && webpack -c ./webpack.esm.config.js && rushx api-extractor",
```
PorterNan, line 11: If that is esm, will webpack bundle them into one file or several files like all the other packages?

packages/acs-ui-common/package.json
JamesBurnside: This should be pinned to the same as react so this should be whatever react is. Really they should probably all be `~` or `^` but we don't have good repo-wide guidenence on what developers should choose. Should we set a standard now and update guidence in the wiki?

packages/communication-react/package.json
JamesBurnside: We need the webpack output from the build. Currently our meta package rebuilds everything itself and hence tsc doesn't use the webpack output. This was the simplest way I found to have the webpack bits in the final npm package (but open if you find a way to instead have tsc grab the webpack output instead of retranspiling the src code)

packages/northstar-wrapper/src/index.ts
JamesBurnside: The question is why we would do `*` at all. We don't want to allow exports from `n*` so we want to keep this to a minimal set such that we don't allow developers to accidentally introduce new reliances on northstar

packages/northstar-wrapper/package.json
JamesBurnside: Still one file, but a file that has treeshaking support (treeshaking can still strip out unused exports in the same file)

packages/acs-ui-common/package.json
JamesBurnside: And yeah we will need that, I have something cooking up on the side but haven't had time for it, will be similar to the typescript check
image

packages/storybook/stories/storybook.test.ts
emlynmac: Maybe storybook 7 works?

emlynmac (commented):
Just some questions, but looks good!

packages/react-composites/tests/browser/call/hermetic/VerticalGalleryResizing.test.ts
```diff
@@ -87,7 +87,7 @@ test.describe('Vertical gallery resizing tests', async () => {
     await waitForSelector(page, dataUiId(IDS.verticalGalleryVideoTile));
 
     // check initial tile number to be correct.
-    expect(await page.locator(dataUiId(IDS.verticalGalleryVideoTile)).count()).toBe(2);
+    await waitForSelectorCount(page, dataUiId(IDS.verticalGalleryVideoTile), 2);
```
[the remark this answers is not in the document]
JamesBurnside: nope, it waits for the correct count and fails if the condition isn't met. Change here allows two things:
- the count is now waited for (so can be in subsequent render passes)
- the action will now screenshot on a failure

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/ParticipantPane.test.ts-snapshots/PSTN-call-screen-with-dialpad-Mobile-Android-Landscape-linux.png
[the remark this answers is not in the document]
JamesBurnside: As far as I can tell its a react 18 optimization. Before with react 16 it took an extra render pass before that focus shows up and our tests took the screenshot before that was happening. With react 18 it handles bulk state changes better and passes focus more efficiently so they tests happen to take their screenshot after the component has taken focus. From running locally the focus is expected

packages/storybook/stories/storybook.test.ts
JamesBurnside: Hopefully, although we actually were considering removing these tests as no one has reported finding a regression with them and they aren't fun to have to update or review - but that's a side conversation

packages/acs-ui-common/package.json
PorterNan: > 

Yea, would be nice to always get latest bit of the current compatible version, cause it might be something unexpected happening there (very rare though) then we can catch it earlier than our customer (cause customers prefer those "^" for versions)

JamesBurnside: I will keep this change for a separate PR, made a bug for it for tracking [Bug 3279379](https://skype.visualstudio.com/SPOOL/_workitems/edit/3279379): [Web] Repo should have guidence on using a pinned dep version vs `^` or `~`

packages/react-components/src/components/MessageStatusIndicator.tsx
```diff
@@ -123,7 +123,7 @@ export const MessageStatusIndicator = (props: MessageStatusIndicatorProps): JSX.
           calloutProps={{ ...calloutProps }}
           styles={hostStyles}
         >
-          {strings.sendingAriaLabel && <LiveMessage message={strings.sendingAriaLabel} aria-live="polite" />}
+          {strings.sendingAriaLabel && <LiveMessage message={strings.sendingAriaLabel} ariaLive="polite" />}
```
dmceachernmsft, line 126: Small thing but did we do any announcer testing here? we have had bugs around this component in the past not announcing states

JamesBurnside: Just the testing done here: https://github.com/Azure/communication-ui-library/pull/3005. Confirmed the DOM shows the same behavior as before

packages/react-components/src/components/Announcer/AnnouncerContext.ts
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react';
+
+/** @private */
+export type AnnouncerContextType = {
+  announceAssertive: (message: string, id: string) => void;
+  announcePolite: (message: string, id: string) => void;
+};
+
+/** @private */
+const AnnouncerContext = React.createContext<AnnouncerContextType>({
```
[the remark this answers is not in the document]
JamesBurnside: I will leave this to a later discussion. This doesn't add a new provider, just moves it from an external package into our codebase

packages/northstar-wrapper/src/index.ts
JamesBurnside: Nothing needs done it will automatically work - but actually wih it didn't. Someone _**should not**_ be adding to this list. We want to get further away from n* not be more reliant on it,

packages/communication-react/tsconfig.preprocess.json
```diff
@@ -10,7 +10,8 @@
       "@internal/calling-stateful-client": ["calling-stateful-client/src"],
       "@internal/chat-stateful-client": ["chat-stateful-client/src"],
       "@internal/react-components": ["react-components/src"],
-      "@internal/react-composites": ["react-composites/src"]
+      "@internal/react-composites": ["react-composites/src"],
+      "@internal/northstar-wrapper": ["northstar-wrapper/src"]
```
PorterNan, line 14: This is for preprocess version, do we preprocess northstar as well?

packages/communication-react/package.json
PorterNan: Yea.. within the internal packages, it will try to use src folder instead (and that's why we use copy-preprocess), btw, I guess doing copy-northstar in the end means we want it to be packaged and consumed by customer, but our own build process doesn't need it?

packages/react-composites/src/composites/CallComposite/MockCallAdapter.ts
```diff
@@ -98,7 +98,7 @@ export class MockCallAdapter implements CallAdapter {
     throw Error('createStreamView not implemented');
   }
   disposeStreamView(): Promise<void> {
-    throw Error('disposeStreamView not implemented');
+    return Promise.resolve();
```
dmceachernmsft, line 101: why do we need an implementation here now? is there one needed for createView as well?

packages/northstar-wrapper/package.json
PorterNan: Yea... I agree with production build should always use esm, but still there would be some app living in cjs world for some legacy reasons, so we might still want to support that well

For npm package, that is just a question out of curiosity, I did receive a feedback from webchat team asking about the  npm package size, but it shouldn't matter too much in the end

packages/communication-react/package.json
JamesBurnside: Yeah that's correct, it does mean we have gaps where parts of our build system is using the webpack output and other parts just directly grabs the n* code from the northstar-wrapper/node_modules :/

packages/react-composites/tests/app/lib/HiddenChatComposites.tsx
```diff
@@ -12,23 +12,67 @@ import { ChatAdapter, ChatComposite } from '../../../src';
  * The added chat composites are hidden, but tests can interact with them programmatically.
  */
 export const HiddenChatComposites = (props: { adapters: ChatAdapter[] }): JSX.Element => {
+  const [hiddenCompositesProps, setHiddenCompositesProps] = useState<HiddenChatCompositesProps[]>([]);
```
dmceachernmsft, line 15: maybe dumb question but what is the reason for the hidden composites again? just to send messages in the tests? (wont block this PR on this but would love to know why we have these)

packages/communication-react/tsconfig.preprocess.json
JamesBurnside: There's no conditional compiles for n*, and given we should not be introducing new usage of this library there shouldn't be in future so stable vs beta didn't need supported in this package

packages/react-composites/tests/browser/call/hermetic/HorizontalGalleryResizing.test.ts
```diff
@@ -87,13 +87,13 @@ test.describe('Height gallery resizing tests', async () => {
     await waitForSelector(page, dataUiId(IDS.horizontalGalleryVideoTile));
 
     // check initial tile number to be correct.
-    expect(await page.locator(dataUiId(IDS.horizontalGalleryVideoTile)).count()).toBe(2);
+    await waitForSelectorCount(page, dataUiId(IDS.horizontalGalleryVideoTile), 2);
```
dmceachernmsft, line 90: love this refactoring

packages/communication-react/tsconfig.preprocess.json
JamesBurnside: (so I guess this line isn't needed? I just copy-pasted anywhere `@internal` packages were referenced)

packages/northstar-wrapper/src/index.ts
PorterNan: Are we saying we developers could import NorthStar component from our package? I guess we don't expose them as a public api right?

packages/react-components/src/components/Announcer/Announcer.tsx
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft Corporation.
```
PorterNan, line 1: The reason why we have these files is because it's not exportable from northstar, I guess?

packages/northstar-wrapper/package.json
JamesBurnside: True, and we do support cjs, we just don't optimize any bundles. However anyone using this in cjs will actually receive a benefit because they don't have to rollup the entire n* library, the small pieces needed are now directly bundled by us so smaller package overall (unless of course they use both us and n* in their app and are bundling using cjs)

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.test.tsx
```diff
@@ -320,15 +320,10 @@ const setup = (
...
-  act(() => {
-    root = create(<TestComponent {...initialArgs} />);
-  });
+  const { rerender } = render(<TestComponent {...initialArgs} />);
 
   if (updatedArgs) {
-    act(() => {
-      root.update(<TestComponent {...updatedArgs} />);
-    });
+    rerender(<TestComponent {...updatedArgs} />);
```
PorterNan, line 326: This is the new format from new react test package?

packages/northstar-wrapper/src/index.ts
JamesBurnside: nope, just internal developers to this repo, not npm package users

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.test.tsx
PorterNan: Sry the render one

PorterNan: Ohh, you mentioned react 18 change its way to render components

packages/react-components/src/components/Announcer/Announcer.tsx
JamesBurnside: See here: https://github.com/Azure/communication-ui-library/pull/3005. The component we used has react 16 as peer dependency but it is a tiny package so we are just taking the code from it directly

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.test.tsx
JamesBurnside: yeah

packages/react-composites/tests/app/chat/index.tsx
```diff
@@ -16,4 +16,9 @@ const params = Object.fromEntries(urlSearchParams.entries());
 initializeFileTypeIcons();
 initializeIconsForUITests();
 
-ReactDOM.render(params.fakeChatAdapterArgs ? <FakeAdapterApp /> : <LiveTestApp />, document.getElementById('root'));
+const domNode = document.getElementById('root');
```
PorterNan, line 19: Yay, with this code piece, I guess we are good to use newest CRA to create projects!

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.test.tsx
JamesBurnside: Yeah newer syntax with the newer testing library. the render blocks don't need to be in act blocks

packages/react-composites/tests/app/chat/index.tsx
JamesBurnside: We are!

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/ParticipantPane.test.ts-snapshots/PSTN-call-screen-with-dialpad-Mobile-Android-Landscape-linux.png
dmceachernmsft, line 1: Did we always have this CSS when the box has focus? I thought the underline just turned blue

packages/react-composites/src/composites/CallComposite/MockCallAdapter.ts
JamesBurnside: One of the tests called it as the tile logic now completes its unmount before the test finishes

packages/react-composites/tests/app/lib/HiddenChatComposites.tsx
JamesBurnside: Yeah just to send messages and read receipts. They don't really have to be hidden tbh, and I wonder if we could replace them with just adapter uses and fire off read events and messages and such through the adapter interface

packages/react-composites/tests/browser/call/hermetic/VerticalGalleryResizing.test.ts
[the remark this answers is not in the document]
PorterNan: So previous one will be waiting for dataUiId(IDS.verticalGalleryVideoTile), if that is showing up, it will start counting the number

But the current behavior would be wait until it is 2 count, and failed it that is not (and take screenshot I guess?)

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/ParticipantPane.test.ts-snapshots/PSTN-call-screen-with-dialpad-Mobile-Android-Landscape-linux.png
JamesBurnside: I believe you already asked this one 😉 https://github.com/Azure/communication-ui-library/pull/3045#issuecomment-1546025173

packages/react-composites/tests/browser/call/hermetic/VerticalGalleryResizing.test.ts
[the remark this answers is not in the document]
PorterNan: I am good with this one, this might miss some "delay render"(like first second it renders one element, but later render 2 elements which meet the condition) issues, but also allow use to get more stable UI snapshot

It could have either make our pipeline faster or slower, would like to see the final number when we have this in main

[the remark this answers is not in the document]
JamesBurnside: I believe the previous one actually did no waiting, it just immediately checked what was rendered. We now instead wait for the condition to become true because the resize observer kicks in after mounting so the tiles we show changes across a couple render passes

packages/react-composites/tests/browser/callwithchat/hermetic/Chat.test.ts
```diff
@@ -42,11 +42,11 @@ test.describe('CallWithChat Composite CallWithChat Page Tests', () => {
     await waitForMessageSeen(page);
 
     // Start typing a message
-    await startTypeMessageFromHiddenChatComposite(page, chatRemoteParticipant, 'I agree!');
+    await typeMessageFromHiddenChatComposite(page, chatRemoteParticipant, 'I agree!');
     // Wait for indicator to show up
     await waitForAndHideTypingIndicator(page, APP_UNDER_TEST_ROOT_SELECTOR);
     // Stop typing and send a message
-    await stopTypingAndSendMessageFromHiddenChatComposite(page, chatRemoteParticipant);
+    await typeAndSendMessageFromHiddenChatComposite(page, chatRemoteParticipant, 'I agree!');
```
PorterNan, line 49: Looks like we are typing 'I agree' for twice and I guess this is changing the test behaivor?

packages/react-composites/tests/browser/common/hermeticChatTestHelpers.ts
```diff
@@ -15,7 +15,10 @@ import { dataUiId, perStepLocalTimeout, screenshotOnFailure } from './utils';
  */
 export async function temporarilyShowHiddenChatComposite(page: Page, participant: ChatParticipant): Promise<void> {
   await withHiddenChatCompositeInForeground(page, participant, async () => {
-    return;
+    // wait for messages to have loaded
+    await waitForSelector(page, dataUiId('chat-composite-message'));
+    // sleep for 100ms to give time for the messagethread hook to send the read receipt
```
PorterNan, line 20: Maybe we need a hidden element in html to notify that "I have successfully sent the read receipt at what time point"

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/ParticipantPane.test.ts-snapshots/PSTN-call-screen-with-dialpad-Mobile-Android-Landscape-linux.png
PorterNan, line 1: Noticed there is a focus change here, is it by designed?

packages/react-composites/tests/browser/callwithchat/hermetic/Chat.test.ts
JamesBurnside: The logic for how the hidden composite is hidden has changed. We now no longer render it entirely whereas before we just marked visibility as hidden. This is because react is much better now and hooks like triggering read receipts get triggered when the composite is hidden now. Because we unmount anything typed and not sent it lost, so I've updated the functions to by type* and typeAndSend*

packages/react-composites/tests/browser/common/hermeticChatTestHelpers.ts
JamesBurnside: Yeah :/ Would be nice to use the adapter and not composites for these actually and use adapter interface to send the read receipt

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/ParticipantPane.test.ts-snapshots/PSTN-call-screen-with-dialpad-Mobile-Android-Landscape-linux.png
JamesBurnside: Yep see here: https://github.com/Azure/communication-ui-library/pull/3045#issuecomment-1546025173

packages/northstar-wrapper/package.json
PorterNan: Lets open a work item to validate that later including find a solution for treeshaking localization part

packages/react-composites/tests/browser/callwithchat/hermetic/Chat.test.ts
PorterNan: Could be a work item not to use hidden composite but adapter/clients

packages/react-composites/src/composites/CallComposite/MockCallAdapter.ts
dmceachernmsft: ahhh okay that makes sense

packages/react-composites/tests/app/lib/HiddenChatComposites.tsx
dmceachernmsft: yeah could be a interesting refactoring of the tests. I don't think its anything worth blocking this over though since are already using this.

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/ParticipantPane.test.ts-snapshots/PSTN-call-screen-with-dialpad-Mobile-Android-Landscape-linux.png
dmceachernmsft: I did and totally missed it! that makes sense to me. I am also just remembering the CSS as it turns out!

PorterNan (approved):

dmceachernmsft (approved):

samples/Chat/tsconfig.json
```diff
@@ -9,6 +9,7 @@
       "@internal/acs-ui-common": ["../../../packages/acs-ui-common/src"],
       "@internal/calling-stateful-client": ["../../../packages/calling-stateful-client/src"],
       "@internal/chat-stateful-client": ["../../../packages/chat-stateful-client/src"],
+      "@internal/northstar-wrapper": ["../../../packages/northstar-wrapper/src"],
```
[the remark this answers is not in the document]
JamesBurnside: Good question! We avoid imports in the code to make sure we only use publicly available exports so that when we migrate the code to the samples nothing breaks. But internally it needs to reference the packlets for quicker building (doesn't have to first build the downstream packlets) and for hot reloading

emlynmac (approved):

PR #3055, opened 2023-05-16
[Chat][Mention] Resolve accessibility issues with mention popover

palatter:
# What

- When using narrator/voice over, the focus should be put on the first item in the popover when it appears.
- Each item in the list should have a role of menu item

# Why

These accessibility issues needed to be addressed.

# How Tested

- Locally run storybook and test with 'SendBox - AtMention' example
- Use narrator/voice-over with Chrome on Mac OS by pressing CMD + F5
- Open popover

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/MentionPopover.tsx
+++ b/packages/react-components/src/components/MentionPopover.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 import React, { useEffect, useRef, useState, useCallback } from 'react';
-import { Persona, PersonaSize, Stack, mergeStyles, useTheme } from '@fluentui/react';
+import { Persona, PersonaSize, Stack, mergeStyles, useTheme, FocusZone } from '@fluentui/react';
 import {
   mentionPopoverContainerStyle,
   headerStyleThemed,
@@ -251,6 +251,7 @@ export const _MentionPopover = (props: _MentionPopoverProps): JSX.Element => {
     (suggestion: Mention, onSuggestionSelected: (suggestion: Mention) => void, active: boolean): JSX.Element => {
       return (
         <div
+          role={'menuitem'}
           data-is-focusable={true}
           /* @conditional-compile-remove(mention) */
           data-ui-id={ids.mentionSuggestionItem}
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/MentionPopover.tsx
```diff
@@ -314,18 +315,20 @@ export const _MentionPopover = (props: _MentionPopoverProps): JSX.Element => {
...
-          className={suggestionListStyle}
-        >
-          {suggestions.map((suggestion, index) => {
-            const active = index === activeSuggestionIndex;
-            return onRenderSuggestionItem
-              ? onRenderSuggestionItem(suggestion, onSuggestionSelected, active)
-              : defaultOnRenderSuggestionItem(suggestion, onSuggestionSelected, active);
-          })}
-        </Stack>
+        <FocusZone shouldFocusOnMount={true}>
```
Leah-Xia-Microsoft, line 318: When MentionPopover is presented, the focused element should still be the inputbox.
The first element of the popover should be "active", not focused.

palatter: So without the focus zone, it won't read out the first item in the list, similar to the participant pane. You're right though, it does affect the focus of the cursor. 

I'll look into the active param you mentioned.

change-beta/@azure-communication-react-26897390-5615-4ae3-8818-289d5617a094.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
dmceachernmsft, line 2: This change type should be `pre-release` since we have not released this yet(?)

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -162,14 +162,18 @@ const extractContent = (s: string): string => {
 };
 
 const messageContentAriaText = (props: ChatMessageContentProps): string | undefined => {
-  return props.message.content
+  // Strip all html tags from the content for aria.
+  const regexForStrippingHtml = /<[^>]*>|<script/gm;
+  const htmlStrippedContent = props.message.content?.replaceAll(regexForStrippingHtml, '');
```
dmceachernmsft, line 167: I'm guessing we are stripping out any possible HTML here since when images or responses were sent it would read out the tags?

dmceachernmsft (approved):
Looks great!

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
emlynmac: Mentions are encoded in a <msft-mention> tag, which we don't want read out. More generally, we don't want tags read too

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -677,6 +678,17 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
     [setTargetSelection, targetSelection, setShouldHandleOnMouseDownDuringSelect, interactionStartPoint, tagsValue]
   );
 
+  /* @conditional-compile-remove(mention) */
+  const announcerText = useMemo(() => {
+    if (activeSuggestionIndex === undefined) {
+      return undefined;
+    }
+    const currentMention = mentionSuggestions[activeSuggestionIndex ?? 0];
```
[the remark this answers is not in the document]
emlynmac: It's actually the mention, not the label. The label string is what's returned from the `useMemo` and is stored in the announcerText.

AmyL219:
Are there any other Accessibility PRs for atMention?
I've noticed a few things that might concern the Accessibility team:

- After selecting the `Unnamed participant`, VoiceOver only reads participant.
- When the AtMention Popover is displayed, PopOver reads 5 options to indicate the presence of the popover.
- We haven't handled the `tab` button yet.

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -274,19 +275,19 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
         return;
       }
       if (ev.key === 'ArrowUp') {
```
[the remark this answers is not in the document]
AmyL219: No, it didn't work for me. I have demo this this morning

PR #3056, opened 2023-05-16
Filter messageReceived notification in adapter by threadId

AmyL219:
# What

Filter messageReceived notification in adapter by threadId

# Why

The composite is specifically designed for single-threaded use. To avoid confusion when a user participates in multiple threads, we filter messageReceived notifications by threadId.

# How Tested

Use same token in joining two different threads. Observed if the adapter only emit the event for the message specific to this chat thread

# Process & policy checklist


emlynmac (approved):

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -369,6 +369,12 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
   }
 
   private messageReceivedListener(event: ChatMessageReceivedEvent): void {
+    const isCurrentChatAdapterThread = event.threadId === this.chatThreadClient.threadId;
+
+    if (!isCurrentChatAdapterThread) {
```
JoshuaLai, line 374: nipick: do we even need the `const` in this case? I guess it makes it slightly clearer and easier to read. 
```suggestion
    if (event.threadId !== this.chatThreadClient.threadId) {
```

JoshuaLai (approved):

packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
AmyL219: I think using a descriptive variable name here could improve readability.

AmyL219:
Based on our previous discussion, we're treating this behavior update in the adapter as a bug fix. As this might require Contoso to modify their implementation code, let me know if there are any additional steps to consider during the release. cc @JoshuaLai

JoshuaLai:
@AmyL219 lets just make a not of it. When we create the release notes for the release branch we will provide a more detailed description if need be.

PR #3060, opened 2023-05-16
Fix issue where after changing background, the previous background persisted for a split second

JamesBurnside:
# What

Apply effects to local video stream before starting the camera in the call.

# Why

Fix issue where after changing background, the previous background persisted for a split second

# How Tested

Locally, all working except turning off the background when the camera is off. This happen before this change also and appears to be a bug in the SDK, this will need to be a separate bug.


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -542,9 +545,30 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
...
-        await this.handlers.onToggleCamera(options);
+        // First kick off the effect on the local device before starting the camera in the call.
+        // This prevents the effect not being applied for a brief moment when the camera is started.
         /* @conditional-compile-remove(video-background-effects) */
-        await startSelectedVideoEffect(this);
+        {
+          const selectedEffect = this.getState().selectedVideoBackgroundEffect;
+          const videoDeviceInfo = this.getState().devices.selectedCamera || this.getState().devices.cameras[0];
+          if (selectedEffect && videoDeviceInfo) {
+            const stream = new SDKLocalVideoStream(videoDeviceInfo);
```
prabhjot-msft, line 555: Do we need to create a new effect here? We have underlying handlers, shouldn't we call that. Another question about video device info, here the first camera is selected. Should the effect be applied to all video streams.

JamesBurnside: Good questions!

> We have underlying handlers, shouldn't we call that

The handlers we have only act on active streams - as they should. Here we are applying an effect to a local video stream before turning it on. The only reason we can do this is we know we are turning that camera on right away.

> Could we update the handlers to always work on inactive cameras?

Yes, but then we would need extra logic to start and stop effects when the camera is switched. Here we don't need to worry becuase we are immediately turning on the camera. We could also move this logic into the onToggleCamera handler but that muddies its functionality and removes the reflective properties of the function mapping to the according react component.

> the first camera is selected. Should the effect be applied to all video streams

We shouldn't because we are applying the effect to the camera (i.e. the stream) we are immediately about to turn on (I'll update this bit code to use a common function that onStartCamera uses so they always use the same logic of grabbing the selected camera).

prabhjot-msft (approved):
Looks good, just a minor comment if we could move the code for starting effect in a util method, rather than keeping in the start camera adapter function.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
edwardlee-msft, line 555: Nice this a neat fix.
Non blocking question: 
Creating a new SDKLocalVideoStream doesn't "turn on" the camera does it, not until we onToggleCamera?
Confirming cause I don't know if there would be any concerns with getting video stream prior to actually turning it on. Would be less of a concern if we had a pre local preview view that contoso can view before actually turning on the camera.

edwardlee-msft (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
JamesBurnside: It doesn't, but only because of the calling sdk's implementation (it waits until createView is called). Really we should assume it _does_ turn on and use the camera because we are acquiring a stream. In this case its OK to do so because we are calling this on an onStartCamera function so camera access is expected

PR #3063, opened 2023-05-17
Add the beta release validation in CI

PorterNan:
# What

Add the beta release validation in CI

# Why

# How Tested

# Process & policy checklist


common/config/rush/command-line.json
```diff
@@ -74,6 +83,17 @@
...
+    {
+      "commandKind": "bulk",
+      "name": "build",
+      "summary": "Build all projects that haven't been built, or have changed since they were last built",
+      "description": "This command is similar to \"rush rebuild\", except that \"rush build\" performs...",
+      "safeForSimultaneousRushProcesses": false,
+      "enableParallelism": true,
+      "ignoreDependencyOrder": false,
+      "ignoreMissingScript": false,
+      "allowWarningsInSuccessfulBuild": true
```
PorterNan, line 95: We ignore warning as failure check for api updates, because we have a following step in GitHub action which checks file changes for those updates

For beta-release build, generate a new api file is always expected

common/config/babel/.babelrc.js
```diff
@@ -58,8 +58,19 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
         // feature for blurred background and replace background effects
         'video-background-effects',
         // feature for vertical gallery layouts in VideoGallery
-        'vertical-gallery'
+        'vertical-gallery',
+        // a demo feature flag for those beta feature not ready for beta release
+        'in-progress-beta-feature-demo'
```
emlynmac, line 63: Do we need to communicate how and when to use this flag?

alkwa-msft: I think its worthwhile to add some documentation on how to use this just so there isn't any miscommunication about adding the entry to the wrong list or how to debug something wierd going on. Getting a first version of thoughts down might be good to help us understand if there are any gaps in documentation. @PorterNan

PorterNan: Definitely, we are doing this as an experiment for next release, and will have a more official version after that, once we polish this well and make this ready for general usage, we will have have a meeting talk about this and collect more feedbacks

common/config/babel/.babelrc.js
```diff
@@ -58,8 +58,19 @@ process.env['COMMUNICATION_REACT_FLAVOR'] === 'stable' &&
...
+      ],      
+      // A list of in progress beta feature.
+      // These features are still beta feature but "in progress"
+      // causing a build failure, but they are ignored by the preprocessing step.
+      inProgressFeatures: [
```
PorterNan, line 68: Take a note here from offline chat with Donald - this could be named as 'alpha' or something to have a more straightforward meaning

Will bring this up with our whole team when we talk about the proposal

PR #3072, opened 2023-05-18
Update stateful to handle new ActiveEffect calling sdk update

JamesBurnside:
# What

- Update activeEffect when the callsubscriber receives the localVideoStreamUpdated event
- Update the api to just use activeEffect same as calling SDK

# Why

Updates to match calling sdk api

# How Tested

Ran locally and could change effects


JamesBurnside:
making changes

JamesBurnside:
Going to update this to handle array of streams

PR #3074, opened 2023-05-18
Update stateful to handle array of active background effects

JamesBurnside:
# What

- Update the api to be an array of activeEffects same as calling SDK
  - Remove isActive in favor of just the name to match calling sdk API.
- Update activeEffects when the callsubscriber receives the localVideoStreamUpdated event

# Why

Updates to match calling sdk api

# How Tested

Ran locally and could change effects.


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -275,19 +275,13 @@ export class CallContext {
   }
 
   /* @conditional-compile-remove(video-background-effects) */
-  public setCallLocalVideoStreamVideoEffects(
-    callId: string,
-    videoEffects: Partial<LocalVideoStreamVideoEffectsState>
-  ): void {
+  public setCallLocalVideoStreamVideoEffects(callId: string, videoEffects: LocalVideoStreamVideoEffectsState): void {
```
prabhjot-msft, line 278: should the videoEffects be of type `videoeffectname[]` not sure if we still need `LocalVideoStreamVideoEffectsState`

JamesBurnside: Great point, I also thought upon this. My feeling for keeping the nested `videoEffects: LocalVideoStreamVideoEffectsState` in the stateful client was that the videoEffects returned by the calling sdk is an object. Currently it just has activeEffects array, but in future there is a potential to have more properties

prabhjot-msft (approved):

mgamis-msft (approved):

PR #3076, opened 2023-05-18
Fix positioning of video effects picker tiles when there is a row with a differing count

JamesBurnside:
# What

Ensure layout is correct when the tiles don't fill the width (see updated snapshots)

# How Tested

See updated snapshots


JamesBurnside:
(draft until snapshots are updated)

prabhjot-msft (approved):

edwardlee-msft (approved):

PR #3082, opened 2023-05-19
[bugfix] SpokenLanguage list drawer menu MobileView 75% height 

edwardlee-msft:
# What

SpokenLanguage List Drawer Menu mobileview set to 75% height.
Screenshot 2023-05-19 at 15 44 33
Screenshot 2023-05-19 at 15 44 16

Edit:
Applied the fix to drawerSurface component instead of only the spokenLanguageDrawer as per comments suggested.
Tested, and current ui is not impacted. However, now, any contextual menu that spans the whole view vertically will not be restricted to 75%.
image
image

# Why
SpokenLanguage Drawer Menu encompassed the full screen on mobile, causing issues with dismissing the menu.
https://skype.visualstudio.com/SPOOL/_workitems/edit/3233595
image

# How Tested
MacOS Chrome mobile view, CallWithChat Composite.


JamesBurnside:
Great fix! Will review any snapshots updates before approving

JamesBurnside:
One thing I've seen - because vertical space is so limited in landscape, instead of setting height to 75%, we could alternatively (just for landscape) set width to be 75% so the user can tap to either side of the drawer to dismiss, but regardless this is much better and a great fix!

packages/react-composites/src/composites/common/Drawer/SpokenLanguageDrawer.styles.ts
```diff
@@ -10,12 +10,23 @@ import { _DrawerMenuStyles } from '@internal/react-components';
  */
 export const spokenLanguageDrawerStyles = (theme: Theme): _DrawerMenuStyles => ({
```
JamesBurnside, line 11: should we always apply this fix to all drawers? (I assume all can take up the whole screen given enough options supplied)

packages/react-composites/src/composites/common/Drawer/SpokenLanguageDrawer.styles.ts
```diff
@@ -10,12 +10,23 @@ import { _DrawerMenuStyles } from '@internal/react-components';
...
     overflow: 'auto'
   },
   drawerSurfaceStyles: {
+    root: {
+      // Targets FocusTrapZone in drawer specific to spokenLanguageDrawer so it does not impact
+      // components using FocusTrapZone.
+      // Setting percentage to Height to transform a container does not work unless the
+      // direct parent container also has a Height set other than 'auto'.
+      '> div:nth-child(2)': {
+        height: '75%',
```
dmceachernmsft, line 23: I wonder if we are to apply this to all drawers that this should be max height and let the other drawers shrink if need be?

packages/react-composites/src/composites/common/Drawer/SpokenLanguageDrawer.styles.ts
edwardlee-msft: Applied the fix to all drawers :).

packages/react-composites/src/composites/common/Drawer/SpokenLanguageDrawer.styles.ts
edwardlee-msft: That's right, after I surfaced it to the DrawerSurface Component I switched it to maxHeight.

edwardlee-msft:
Will need to make some changes. I will add a prop so that it doesn't impact other drawerSurfaces being used by call readiness and dialpad.

carocao-msft (approved):

JamesBurnside (approved):

PR #3087, opened 2023-05-23
[Chat] Refactoring mention feature

Leah-Xia-Microsoft:
# What

- Move Tag related helper function out of InputBoxComponent into MentionTagHelper file
- Create a new TextFieldWithMention component
- Move mention related logic out of InputBoxComponent into TextFieldWithMention
- Separate some function into smaller chunks
- Address issues from previous PR:

https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186382630
https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186386618
https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186442301
https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186447075
https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186444434

Adjust use of `null` to `undefined` where at all possible.
https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186429934

User of forEach() instead of for i...
https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186439960

Review logic
https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186441062

Use of conditional compilation:
https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186431425
https://github.com/Azure/communication-ui-library/pull/2982#discussion_r1186431647

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
```diff
@@ -0,0 +1,957 @@
...
+  }
+  return { result, updatedChange: processedChange, htmlIndex: lastProcessedHTMLIndex, plainTextSelectionEndIndex };
+};
+
+/**
+ * Closing tag information
+ *
+ * @private
+ */
+export type ClosingTagInfoResult = {
```
emlynmac, line 246: Why name this way? 
`ClosingTagInfo` is better maybe?

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
```diff
@@ -0,0 +1,957 @@
...
+};
+
+/**
+ * Get closing tag information if exists otherwise return information as for self closing tag
+ *
+ * @private
+ * @param tag - Tag data.
+ * @returns Closing tag information for the provided tag.
+ */
+export const getTagClosingTagInfo = (tag: TagData): ClosingTagInfoResult => {
```
emlynmac, line 259: `getClosingTag` ?

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
```diff
@@ -0,0 +1,957 @@
...
+  }
+  return { plainTextEndIndex, closeTagIndex, closeTagLength };
+};
+
+/**
+ * Props for update HTML function
+ *
+ * @private
+ */
+export type UpdateHTMLProps = {
```
emlynmac, line 283: Given that this is only used explicitly in one place, do we need to define it separately, rather than just inline with the method parameters?

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
```diff
@@ -0,0 +1,957 @@
...
+  // the end of current selection, should be a valid position in the input field
+  currentSelectionEnd: number;
+};
+
+/**
+ * Result of finding strings diff indexes function
+ *
+ * @private
+ */
+export type DiffIndexesResult = {
```
emlynmac, line 610: `IndexChanges`?

emlynmac (approved):
Some minor naming questions, but nothing that can't wait.

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
PorterNan, line 1: We use utils more than helpers in the repo, so would be nice to align with all the other names

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -0,0 +1,629 @@
...
+              }
+            }
+          }
+        }
+      }
+    },
+    [debouncedQueryUpdate, mentionLookupOptions]
+  );
+
+  const getUpdatedNewValueForOnChange = useCallback(
```
PorterNan, line 427: Is all the value consumed by this function all coming from parameter? if yes, would like to see a separated functions for this

PorterNan: I mean a const function not living in the component function body

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -0,0 +1,629 @@
...
+const DEFAULT_MENTION_TRIGGER = '@';
+
+/**
+ * Props for the TextFieldWithMention component.
+ *
+ * @private
+ */
+export interface TextFieldWithMentionProps {
+  textFieldProps: ITextFieldProps;
+  dataUiId: string | undefined;
```
PorterNan, line 33:
```suggestion
  dataUiId?: string;
```

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -1,12 +1,7 @@
 // Copyright (c) Microsoft Corporation.
```
PorterNan, line 1: Finally this file is not 1500+ lines anymore! Thanks for making this one smaller

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -0,0 +1,629 @@
...
+        if (mentionSuggestions.length > 0) {
+          const newActiveIndex =
+            activeSuggestionIndex === undefined
+              ? 0
+              : Math.min(activeSuggestionIndex + 1, mentionSuggestions.length - 1);
+          setActiveSuggestionIndex(newActiveIndex);
+        }
+      }
+      if (ev.key === 'Enter' && (ev.shiftKey === false || !supportNewline)) {
+        ev.preventDefault();
```
PorterNan, line 197: Should we have the preventDefault() function in the condition below, or no?

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -0,0 +1,629 @@
...
+      return;
+    }
+    const suggestions = (await mentionLookupOptions?.onQueryUpdated(query)) ?? [];
+    if (suggestions.length === 0) {
+      setActiveSuggestionIndex(undefined);
+    } else if (activeSuggestionIndex === undefined) {
+      setActiveSuggestionIndex(0);
+    }
+    updateMentionSuggestions(suggestions);
+  }, 500);
```
PorterNan, line 225: Make this a constant like DEFAULT_MENTION_TRIGGER, put it at the very beginning of the file

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -0,0 +1,629 @@
...
+          tag: mentionTag,
+          textValue: inputTextValue,
+          currentSelectionIndex: selectionIndex,
+          previousSelectionIndex: selectionValue ?? inputTextValue.length
+        });
+        return newSelectionIndex;
+      }
+      return undefined;
+    },
+    []
```
PorterNan, line 259: Looks like no dependencies are needed for this function, consider move it outside of the component function instead of using useCallback

To make give them a rank:

const function outside of component body (if possible) > useCallback > functions without useCallback

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -0,0 +1,629 @@
...
+  // Update selections index in mention to navigate by words
+  const updateSelectionIndexesWithMentionIfNeeded = useCallback(
+    (
+      event: FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+      inputTextValue: string,
+      selectionStartValue: number | undefined,
+      selectionEndValue: number | undefined,
+      tagsValue: TagData[]
+    ): void => {
+      const currentSelectionStart =
```
PorterNan, line 271: This main part of the function could be separated out as a single const function outside of the component body 

The purpose of this function is to 
1. calculate updatedStartIndex and updatedEndIndex 
2. Set those 2 values into the state

Only 2nd part is this component related while the main part for 1st purpose could be extract as a single function (util function maybe?)

The main purpose here is to make component body as minimized as possible

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -0,0 +1,629 @@
...
+              event.currentTarget.selectionDirection
+            );
+          }
+          setSelectionStartValue(mentionTag.plainTextBeginIndex);
+          setSelectionEndValue(mentionTag.plainTextEndIndex ?? mentionTag.plainTextBeginIndex);
+        } else {
+          setSelectionStartValue(currentSelectionStart);
+          setSelectionEndValue(currentSelectionEnd);
+        }
+      } else {
```
PorterNan, line 369: Feeling it would be more readable if we could do rename + extract some code lines into functions like
if(...) {
   handleSelectionByMouseDown();
} else {
  handleSelectionByKeyboard();
}

Comments would work very well for sure here, but it would be always better if the code structure can make it self-commented

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
PorterNan: Yep, no dependencies array for useCallback would be a good indicator that this function might be able to be moved outside of component function body

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
```diff
@@ -0,0 +1,957 @@
+// Copyright (c) Microsoft Corporation.
```
PorterNan, line 1: These code are more like logics for handling html + tags, I wouldn't be too worried about overall structure here, but would be nice to get unit tests for some important functions so no one will mess them up in the future

Not for this PR though, could be another PR for some new work item

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
Leah-Xia-Microsoft: @emlynmac I asked Viktoriia the same question in the previous PR. It is because ClosingTagInfoResult is the result type for `getTagClosingTagInfo`.
However, since this is also only used explicitly in one place, I'll define it inline with the method parameters.

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
Leah-Xia-Microsoft: > 

I think it make sense to keep the info part since what we get from this function is the information of a closing tag (plainTextEndIndex: number; closeTagIndex: number; closeTagLength: number) and not the closing tag itself?
Or maybe we can rename it to getClosingTagData to align with TagData type?

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
```diff
@@ -0,0 +1,957 @@
...
+  const displayNamePlaceholder = localeStrings.participantItem.displayNamePlaceholder;
+  return suggestion.displayText !== '' ? suggestion.displayText : displayNamePlaceholder ?? '';
+};
+
+/**
+ * Represents a tag in the text.
+ *
+ * @private
+ */
+export type TagData = {
```
vhuseinova-msft, line 741: Should this be moved to the top of the file for better visibility?

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
vhuseinova-msft: it may be better to use`ChangeIndexes` as it's used to store info about indexes connected to a change

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
```diff
@@ -0,0 +1,957 @@
...
+  }
+  return { updatedHTML: result, updatedSelectionIndex: changeNewEndIndex };
+};
+
+/**
+ * Props for finding strings diff indexes
+ *
+ * @private
+ */
+export type DiffIndexesProps = {
```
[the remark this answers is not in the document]
vhuseinova-msft: As we disscussed previously, separate props types are created to setup parameters correctly (this change was requested in the previous PR)

cc @emlynmac, @Leah-Xia-Microsoft

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
vhuseinova-msft: As we disscussed previously, separate props types are created to setup parameters correctly (this change was requested in the previous PR)

cc @emlynmac, @Leah-Xia-Microsoft

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
```diff
@@ -0,0 +1,957 @@
...
+
+/**
+ * Adds a tag to the tag stack and updates the parent tag's subtags.
+ * @param tag - The tag to add.
+ * @param tagStack - The tag stack.
+ * @param tags - The array of tags.
+ *
+ * @private
+ */
+export const addTag = (tag: TagData, parseStack: TagData[], tags: TagData[]): void => {
```
vhuseinova-msft, line 926: Should separate props type be created in oder not to mix up `parseStack` and `tags`?

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
Leah-Xia-Microsoft: > function

Moved it out to the mentionTagHelpers.ts

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -0,0 +1,629 @@
...
+      setCaretIndex(undefined);
+      // shouldHandleOnMouseDownDuringSelect should be set to false after the last mouse down event.
+      // it shouldn't be updated in onMouseUp
+      // as onMouseUp can be triggered before or after onSelect event
+      // because its order depends on mouse events not selection.
+      setShouldHandleOnMouseDownDuringSelect(false);
+      if (isEnterKeyEventFromCompositionSession(ev)) {
+        return;
+      }
+      if (ev.key === 'ArrowUp') {
```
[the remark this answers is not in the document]
Leah-Xia-Microsoft: > So this arrowUp and down is handling the logic when mentionSuggestions list is being rendered - and the way we determine whether the mentionSuggestions list is being rendered is using mentionSuggestions.length > 0 condition
> 
> please correct me if I am wrong

That is correct.

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
Leah-Xia-Microsoft: > These code are more like logics for handling html + tags, I wouldn't be too worried about overall structure here, but would be nice to get unit tests for some important functions so no one will mess them up in the future
> 
> Not for this PR though, could be another PR for some new work item

We do have separated tickets for adding unit tests for all these functions.

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
Leah-Xia-Microsoft: > Should we have the preventDefault() function in the condition below, or no?

Will wrap it in the condition below

packages/react-components/src/components/TextFieldWithMention/mentionTagHelpers.ts
Leah-Xia-Microsoft: > Should separate props type be created in order not to mix up `parseStack` and `tags`?

I feel for this one we can leave it inline. It's not used anywhere else and it's only 3 parameters.

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -0,0 +1,512 @@
...
+  getValidatedIndexInRange,
+  htmlStringForMentionSuggestion,
+  textToTagParser,
+  updateHTML,
+  updateSelectionIndicesWithMentionIfNeeded
+} from './mentionTagUtils';
+import { isEnterKeyEventFromCompositionSession } from '../utils/keyboardNavigation';
+
+const DEFAULT_MENTION_TRIGGER = '@';
+const DEBOUNCE_TIME_INTERVAL = 500;
```
emlynmac, line 25:
```suggestion
const DEBOUNCE_TIME_INTERVAL_MS = 500;
```

emlynmac (commented):
Let's get this updated with the main changes and then get it done.

PR #3089, opened 2023-05-23
OVC implementation in VideoGallery

vriosrada-msft:
# What

# Why

# How Tested

# Process & policy checklist


packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -78,6 +78,12 @@ export const getIsScreenSharingOn = (state: CallClientState, props: CallingBaseS
 export const getIsMuted = (state: CallClientState, props: CallingBaseSelectorProps): boolean | undefined =>
   state.calls[props.callId]?.isMuted;
 
+/**
+ * @private
+ */
+export const getOptimalVideoCount = (state: CallClientState, props: CallingBaseSelectorProps): number | undefined =>
+  state.calls[props.callId]?.optimalVideoCountFeature.optimalVideoCount;
```
JamesBurnside, line 85: Is this a beta feature currently or in stable builds of calling sdk? If its beta we'll need to add conditional compile flags [(adding beta only features)](https://github.com/Azure/communication-ui-library/blob/main/docs/references/beta-only-features.md#adding-features-to-be-released-in-the-beta-channel-only)

packages/calling-stateful-client/src/Converter.ts
```diff
@@ -125,6 +125,9 @@ export function convertSdkCallToDeclarativeCall(call: CallCommon): CallState {
       currentCaptionLanguage: '',
       currentSpokenLanguage: '',
       isCaptionsFeatureActive: false
+    },
+    optimalVideoCountFeature: {
+      optimalVideoCount: 1
```
JamesBurnside, line 130: when a call first comes in, is there already a known optimalVideoCount? e.g. could this be:
```suggestion
      optimalVideoCount: call.feature(Features.OptimalVideoCount).optimalVideoCount
```
?

JamesBurnside:
This is looking great! One major thing is the conditional compile flags that will need set as I believe this is only in beta versions of the calling package.
You will also need to run `rush changelog` to generate a change file (this is used to auto generate an initial changelog for release), as well as build the repo to update the `api.md` files (run `rush build -t @azure/communication-react`, you should only need to do this for beta builds as this feature is only beta)
More details on PR requirements and instructions here: https://github.com/Azure/communication-ui-library/blob/main/docs/contributing-guide/6.%20submitting-a-pr.md#prerequisites-to-publishing-a-pull-request

packages/calling-component-bindings/src/baseSelectors.ts
vriosrada-msft: Addressed this comment by adding conditional compile flags

packages/calling-stateful-client/src/Converter.ts
vriosrada-msft: Coming from the SDK if the call is initializing the default value at the beginning will be 1 as well

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -43,7 +43,7 @@ import { VerticalGalleryStyles } from './VerticalGallery';
  * @private
  * Currently the Calling JS SDK supports up to 4 remote video streams
```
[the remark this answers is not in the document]
vriosrada-msft: updated to 9

packages/calling-stateful-client/src/Converter.ts
JamesBurnside: Is there a reason to not do `optimalVideoCount: call.feature(Features.OptimalVideoCount).optimalVideoCount`? Just in case this implementation detail of it starting as `1` changes in future?

packages/react-components/src/components/VideoGallery/Layout.ts
```diff
@@ -28,7 +28,7 @@ export interface LayoutProps {
   screenShareComponent?: JSX.Element;
   /**
    * Maximum number of participant remote video streams that is rendered.
-   * @defaultValue 4
+   * @defaultValue 9
```
JamesBurnside, line 31: @PorterNan would this be considered a breaking change? Should we leave the default as `4` in the VideoGallery, but at the composite level choose 9?

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -216,7 +216,7 @@ export interface VideoGalleryProps {
   strings?: Partial<VideoGalleryStrings>;
   /**
    * Maximum number of participant remote video streams that is rendered.
-   * @defaultValue 4
+   * @defaultValue 9
```
JamesBurnside, line 219: @PorterNan would this be considered a breaking change? Should we leave the default as `4` in the VideoGallery, but at the composite level choose 9?

packages/calling-stateful-client/src/Converter.ts
vriosrada-msft: just analyzed this comment and addressed as suggested

packages/calling-component-bindings/src/providers/CallProvider.tsx
```diff
@@ -66,7 +66,7 @@ export const useCall = (): Call | undefined => {
   if (call && !_isACSCall(call)) {
     throw new Error('Incorrect call type: Must provide a Regular Call object.');
   }
-  return call;
+  return call as Call;
```
alkwa-msft, line 69: @PorterNan  do we need to do this "cast"?

packages/react-components/src/components/VideoGallery.tsx
alkwa-msft: I think we can keep the default value at 4. The actual OVC value will be larger and that is fine.. up to 9.. eventually up to 25, and 49 but if we have 4 and it snaps to this if OVC fails then we should be ok?

common/config/babel/.babelrc.js
```diff
@@ -77,8 +79,8 @@ process.env['COMMUNICATION_REACT_FLAVOR'] !== 'beta' &&
         'click-to-call',
         // Mention feature
         'mention',
-        // Feature for call transfer
-        'call-transfer'
```
JamesBurnside, line 81: perhaps this got overwritten in a merge? keep this as well as the new change

packages/calling-component-bindings/src/providers/CallProvider.tsx
dmceachernmsft: we shouldn't need to.

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
```diff
@@ -115,6 +115,7 @@ export interface CallState {
     isScreenSharingOn: boolean;
     kind: CallKind;
     localVideoStreams: LocalVideoStreamState[];
+    optimalVideoCountFeature: OptimalVideoCountFeatureState;
```
JamesBurnside, line 118: Are we expecting more related state in the OVC feature? If not just make `optimalVideoCount: number` top level

packages/react-components/src/components/VideoGallery/Layout.ts
alkwa-msft, line 31: Can we maintain the same defaultValue for # of streams in VideoGallery and the Composite? Can we bring up this question in our Feature meeting if we don't know which value to snap to and why?

packages/calling-component-bindings/src/providers/CallProvider.tsx
vriosrada-msft: Not needed, just added to avoid warnings in VSCode I can remove if necessary

packages/calling-stateful-client/src/OptimalVideoCountSubscriber.ts
```diff
@@ -0,0 +1,54 @@
...
+/* @conditional-compile-remove(optimal-video-count) */
+import { CallIdRef } from './CallIdRef';
+
+/* @conditional-compile-remove(optimal-video-count) */
+/**
+ * Subscribes to a Optimal Video Count Feature events and updates the call context appropriately.
+ * @private
+ */
+export class OptimalVideoCountSubscriber {
+  private _parent: CallIdRef | 'unparented';
```
JamesBurnside, line 18: I don't think this can be unparented. Unparented refers to this existing outside of a call (we needed this for local video streams as they exist on configuration pages). Can this be:
```suggestion
  private _callIdRef: CallIdRef;
```

common/config/babel/.babelrc.js
vriosrada-msft: Fixed

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
```diff
@@ -80,7 +80,7 @@ export const DefaultLayout = (props: DefaultLayoutProps): JSX.Element => {
    * this case when those components are here
    */
   const [indexesToRender, setIndexesToRender] = useState<number[]>([
-    ...Array(maxRemoteVideoStreams - activeVideoStreams).keys()
+    ...Array(Math.max(maxRemoteVideoStreams - activeVideoStreams, 1)).keys()
```
JamesBurnside, line 83: Not super clear why there is always atleast 1 index here, can you add a comment

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
vriosrada-msft: There are plans to have the number of the videos rendered there as well. @alkwa-msft what do you think about this suggestion? Both makes sense, keeping it will help to use the new values easily in the future. But, having `optimalVideoCount` is simpler and cleaner

packages/react-composites/src/composites/CallWithChatComposite/adapter/TestUtils.ts
```diff
@@ -238,6 +238,10 @@ function createMockCall(mockCallId: string): CallState {
     /* @conditional-compile-remove(call-transfer) */
     transfer: {
       acceptedTransfers: {}
+    },
+    /* @conditional-compile-remove(optimal-video-count) */
+    optimalVideoCountFeature: {
+      optimalVideoCount: 1
```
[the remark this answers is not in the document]
vriosrada-msft: Changed to 4

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
vriosrada-msft: Added the  comment and the number changed to 4. In case the current number of videos rendered is greater than the max, we will render just 4

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/LoadingVideo.test.ts-snapshots/video-gallery-with-loading-spinners-Mobile-Android-Portrait-linux.png
JamesBurnside, line 1: Can these be removed from the changelist now there are (seemingly) no changes?
locally you can do something like this to pull the changes from main back into your branch:
```
git checkout main -- *.png
```

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
```diff
@@ -53,7 +51,7 @@ const _useOrganizedParticipants = (props: OrganizedParticipantsArgs): OrganizedP
           participants: videoParticipants,
           dominantSpeakers,
           lastVisibleParticipants: visibleGridParticipants.current,
-          maxDominantSpeakers: maxRemoteVideoStreams
+          maxDominantSpeakers: maxRemoteVideoStreams || 4
```
JamesBurnside, line 54: can we remove the `|| 4` and just ensure that this is a required prop to this hook and the value is always passed in? There's a lot of defaults floating around that aren't tied to constants, if `4` ever needs updated it difficult to find all the spots where to update.

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
```diff
@@ -78,9 +78,10 @@ export const DefaultLayout = (props: DefaultLayoutProps): JSX.Element => {
    * For some components which do not strictly follow the order of the array, we might
    * re-render the initial tiles -> dispose them -> create new tiles, we need to take care of
    * this case when those components are here
+   * Note: If for some reason the current number of videos rendered is more than the maximum allowed, 4 videos will be rendered
```
JamesBurnside, line 81: Did this happen in testing? I'm concerned we can get to this point and have `maxRemoteVideoStreams - activeVideoStreams` greater than `maxRemoteVideoStreams` to require the Math.max.
Also, why is the limit 4 here? Isn't `maxRemoteVideoStreams` a flexible number?

packages/calling-stateful-client/src/Converter.ts
```diff
@@ -141,6 +141,10 @@ export function convertSdkCallToDeclarativeCall(call: CallCommon): CallState {
     /* @conditional-compile-remove(call-transfer) */
     transfer: {
       acceptedTransfers: {}
+    },
+    /* @conditional-compile-remove(optimal-video-count) */
+    optimalVideoCountFeature: {
+      optimalVideoCount: call.feature(Features.OptimalVideoCount).optimalVideoCount || 4
```
JamesBurnside, line 147: Can we remove the `|| 4` here - if the SDK result can be undefined then we should have this as undefined and others layers can place defulat values when it doesn't exist. The state here should be a reflection of what is in the calling SDK

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
dmceachernmsft: if we know that there are more properties relating to OVC coming then we will want to keep it like this. If we were to top level this and then want to make a interface to wrap the state together for them this would be a breaking change to the API. I happy leaving this as is knowing that.

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -389,6 +391,16 @@ export class CallContext {
     });
   }
 
+  /* @conditional-compile-remove(optimal-video-count) */
+  public setOptimalVideoCount(callId: string, optimalVideoCount: number): void {
+    this.modifyState((draft: CallClientState) => {
+      const call = draft.calls[this._callIdHistory.latestCallId(callId)];
+      if (call) {
+        call.optimalVideoCountFeature.optimalVideoCount = optimalVideoCount;
```
dmceachernmsft, line 399: I think we should rename this state to be something like `optimalVideoCountState` just so that we don't get confused about invoking the Feature API from the SDK here. We just had this discussion in reviews with the `transfer` work so we will have precedent here.

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
dmceachernmsft: I believe there was a decision made that we were to use the old constant. so if we are providing a default should OVC fail the video gallery should fall back to the original value of `DEFAULT_MAX_REMOTE_VIDEOSTREAMS`

packages/react-composites/src/composites/CallComposite/MockCallAdapter.ts
```diff
@@ -216,6 +216,10 @@ const defaultCallAdapterState: CallAdapterState = {
     /* @conditional-compile-remove(call-transfer) */
     transfer: {
       acceptedTransfers: {}
+    },
+    /* @conditional-compile-remove(optimal-video-count) */
+    optimalVideoCountFeature: {
+      optimalVideoCount: 4
```
dmceachernmsft, line 222: Lets look at using a constant for some of these values

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
dmceachernmsft, line 81: Nit: should this be something along the lines of: 'If for some reason the current number of videos rendered is more than the maximum allow only the max number will show, if that max is less than 4 we will render 4 videos'?

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
```diff
@@ -38,7 +36,7 @@ const _useOrganizedParticipants = (props: OrganizedParticipantsArgs): OrganizedP
   const {
     remoteParticipants = [],
     dominantSpeakers = [],
-    maxRemoteVideoStreams = DEFAULT_MAX_REMOTE_VIDEOSTREAMS,
+    maxRemoteVideoStreams,
```
dmceachernmsft, line 39: Do we still need to provide a default here?

dmceachernmsft (approved):

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
mgamis-msft: I don't think we should be using a minimum of 4 here. These video tile indices to render for the overflow gallery. If we are already rendering the maximum in the grid view we should not be rendering a minimum of 4.

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
```diff
@@ -78,9 +78,10 @@ export const DefaultLayout = (props: DefaultLayoutProps): JSX.Element => {
...
    */
   const [indexesToRender, setIndexesToRender] = useState<number[]>([
-    ...Array(maxRemoteVideoStreams - activeVideoStreams).keys()
+    ...Array(Math.max(maxRemoteVideoStreams - activeVideoStreams, 4)).keys()
```
mgamis-msft, line 84: This is only the initial value of the indices of overflow gallery to render. The array of indices will change to the indices of the first page of the overflow gallery when the overflow gallery is rendered. I think this is not needed.

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
mgamis-msft: I think we still need a default here is still needed here for stable. I don't think it's bad to keep the default in beta alsol.

mgamis-msft (changes requested):
Looks great overall. Just a couple of things in the DefaultLayout.tsx file that need changing for me. We don't need to keep a minimum of 4 in the overflow gallery.

packages/react-components/src/components/VideoGallery/DefaultLayout.tsx
vriosrada-msft: Tested and works well without initialization, I will push the code soon to address this

mgamis-msft (approved):

PR #3090, opened 2023-05-23
[Chat] Fix long Seen display name as ellipsis

jimchou-dev:
# What

- Fix long Seen display name as ellipsis (truncate at 320px confirmed with UX)
image
- Set Chat sample to not show Participant Pane by default (1 liner fix)

# Why

Long Seen display name big https://skype.visualstudio.com/SPOOL/_workitems/edit/3240569
Participant Pane showing by default in Chat Sample https://skype.visualstudio.com/SPOOL/_workitems/edit/3240567

# How Tested

# Process & policy checklist


palatter (approved):
Looks good!

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
```diff
@@ -147,7 +147,16 @@ export const ChatMessageActionFlyout = (props: ChatMessageActionFlyoutProps): JS
         subMenuProps: {
           id: 'chat-composite-message-contextual-menu-read-name-list',
           items: messageReadByList ?? [],
-          calloutProps: preventUnwantedDismissProps
+          calloutProps: preventUnwantedDismissProps,
+          styles: concatStyleSets({
```
palatter, line 151: I notice some places in the code they define the styles separately in their own variable?

JoshuaLai (approved):
Tested looks good to me

packages/react-components/src/components/ChatMessage/ChatMessageActionsFlyout.tsx
jimchou-dev: In this same file they are also using concatStyle so for consistency I did the same here. And I think it make sense for certain styles that are used multiple places to be a variable.

PR #3095, opened 2023-05-23
Add github action to deploy to feature-staging test apps

JamesBurnside:
# What

GH action deploys whatever branch it is on to three new test apps

# Why

We want to stop using the beta validation apps for this.

# How Tested

tbd


JamesBurnside:
Close in favor of #3168

PR #3097, opened 2023-05-24
[Accessibility][MessageThread]Embed the EditBox inside a Chat.Message to enable the Accessibility

AmyL219:
# What

Integrate the EditBox into a Chat.Message component. This modification grants the EditBox the ability to inherit the [ChatMessageBehavior](https://github.com/microsoft/fluentui/blob/master/packages/fluentui/accessibility/src/behaviors/Chat/chatMessageBehavior.ts) from Chat.Message, ensuring it can gain focus when a user presses the 'Tab' key.

# Why

The buttons within the EditBox lose their tab functionality when a user is editing a message. In other words, when a user edits a message and presses the 'Tab' key, the focus shifts outside the EditBox.

The problem arises due to the EditBox being directly embedded within a FluentUI Chat List. From my observation, the Chat list only focuses on the most recent message and disregards other focusable items that aren't considered as Chat.Message.

[Bug 3239498](https://skype.visualstudio.com/SPOOL/_workitems/edit/3239498): [P2][Web][Chat Composite]: Inappropriate keyboard focus order is observed on multiline message edit when user is editing the message.
# How Tested

# Process & policy checklist


JoshuaLai (approved):
Tested the fix. Looks good to me

palatter:
Also tested, looks good.

palatter (approved):

PR #3098, opened 2023-05-24
Add cancelEdit callback to the adapter layer

emlynmac:
# What
Some time ago, a proper edit cancel callback was added to the MessageThread component. It is not necessary to implement this within our chat composite from a logic point of view, it was more for implementors _not_ using an ACS backend to be able to get proper state updates.

This PR passes the handler through to the composite

# Why
Developer experience completeness

# How Tested
Chat sample

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
+++ b/packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
@@ -193,7 +193,7 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
   }
 
   async fetchInitialData(): Promise<void> {
-    // If get properties fails we dont want to try to get the participants after.
+    // If get properties fails we don't want to try to get the participants after.
     await this.asyncTeeErrorToEventEmitter(async () => {
       await this.chatThreadClient.getProperties();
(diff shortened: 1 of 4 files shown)


packages/react-composites/src/composites/ChatComposite/adapter/AzureCommunicationChatAdapter.ts
```diff
@@ -284,6 +284,10 @@ export class AzureCommunicationChatAdapter implements ChatAdapter {
     });
   }
 
+  async cancelMessageEdit(/*messageId: string*/): Promise<void> {
```
PorterNan, line 287: Is this function called by any composite code or impact our own state? If not, I would say it is currently no need to expose it here

The adapter is a layer designed 100% fit for composite, it satisfies and only satisfies what is needed by composite - if someone wants to do more, they can involke stateful client with usePropsFor or directly use stateful client

cancelMessageEdit is currently a pure UI thing, no impacts on either adapter states or client states, so it is not something we would need to adapter layer, at least for now

emlynmac: Sure, we can skip it then; the reason for this PR was a code review item from the original PR. 

It might be nices to be able to see how to implement the functionality if needed.

PR #3101, opened 2023-05-25
[Chat] Add mouse / touch selection to mention text

emlynmac:
# What
Selection of a mention will now snap to words on mouse up.

# Why
Closer to teams behaviour

# How Tested
Storybook

# Process & policy checklist


packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -449,36 +452,51 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
       selectionEndValue: number | null
     ): void => {
       /* @conditional-compile-remove(mention) */
```
JoshuaLai, line 454: nit pick: Do we need a conditional compile flag here? Since it looks like the whole `const handleOnSelect` is already being flagged.

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -611,6 +629,51 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
   };
 
+  /* @conditional-compile-remove(mention) */
+  // Adjust the selection range based on a mouse / touch interaction
+  const handleOnMove = useCallback(
+    (event) => {
+      // Should we do anything?
+      if (interactionStartPoint !== undefined) {
+        // And did selection change?
+        if (
```
JoshuaLai, line 639: A little messy having the two nested `if` cause technically it looks like it can just be one giant `if`
```
if (interactionStartPoint &&
        // And did selection change?
          event.currentTarget.selectionStart !== null &&
          (event.currentTarget.selectionStart !== targetSelection?.start ||
            event.currentTarget.selectionEnd !== targetSelection?.end)
        )
```

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -611,6 +629,51 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+          event.currentTarget.selectionStart !== null &&
+          (event.currentTarget.selectionStart !== targetSelection?.start ||
+            event.currentTarget.selectionEnd !== targetSelection?.end)
+        ) {
+          let targetStart = event.currentTarget.selectionStart;
+          let targetEnd = event.currentTarget.selectionEnd;
```
JoshuaLai, line 645: nitpick: I think we can move these declarations outside the `if` might make the if a little more managable / easier to read

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -127,9 +127,16 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
...
+  // for onMouseDown event is not updated yet and the selection range for mouse click/taps will be
+  // updated in onSelect event if needed.
   const [shouldHandleOnMouseDownDuringSelect, setShouldHandleOnMouseDownDuringSelect] = useState<boolean>(true);
-
+  /* @conditional-compile-remove(mention) */
+  // Point of start of touch/mouse selection
+  const [interactionStartPoint, setInteractionStartPoint] = useState<{ x: number; y: number } | undefined>();
+  /* @conditional-compile-remove(mention) */
+  // Target selection from mouse movement
+  const [targetSelection, setTargetSelection] = useState<{ start: number; end: number | null } | undefined>();
```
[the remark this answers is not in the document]
emlynmac: When the event returns with a null there. The typings allow null, so this is just passthrough

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: No, will pull

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Indeed, updated.

packages/react-components/src/components/InputBoxComponent.tsx
emlynmac: Usually, you want to keep your variables as scoped as you can.
I don't really see why you'd want to move these outside the `if` statement?

emlynmac: Except to simplify the if statement. Moving

JoshuaLai (approved):

jimchou-dev (approved):

PR #3103, opened 2023-05-25
[bugfix] Set MaxHeight to drawerSurface and drawerMenu

edwardlee-msft:
# What

SpokenLanguage List Drawer Menu mobileview set to 75% height.
Screenshot 2023-05-19 at 15 44 33
Screenshot 2023-05-19 at 15 44 16

Edit:
Applied the fix to drawerSurface component instead of only the spokenLanguageDrawer as per comments suggested.
Tested, and current ui is not impacted. However, now, any contextual menu that spans the whole view vertically will not be restricted to 75%.
image
image

# Why

SpokenLanguage Drawer Menu encompassed the full screen on mobile, causing issues with dismissing the menu.
https://skype.visualstudio.com/SPOOL/_workitems/edit/3233595
image

# How Tested

# Process & policy checklist


edwardlee-msft:
tracked here https://github.com/Azure/communication-ui-library/pull/3082/files

PR #3108, opened 2023-05-26
Add the leaving page

PorterNan:
# What

This is a fix/feature for https://skype.visualstudio.com/SPOOL/_workitems/edit/3030570/
Add the leaving page
leaving

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -327,12 +331,22 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
         />
       );
       break;
+    case 'leaving':
+      pageElement = (
+        <NoticePage
```
JamesBurnside, line 336: Any concerns we don't show mic and camera status on this page so users may think they're no longer transmitting these? (I can't recall if they proved media could get through or not in this state)

packages/react-composites/src/composites/CallComposite/styles/CallComposite.styles.ts
```diff
@@ -45,3 +45,11 @@ export const drawerContainerStyles = (zIndex: number = DEFAULT_Z_INDEX): IStackS
     zIndex: zIndex
   }
 });
+
+/**
+ * @private
+ * Page container style for leaving page
+ */
+export const leavingPageStyle: IStyle = {
+  background: '#FAF9F8'
```
dmceachernmsft, line 54: Should this colour come from the theme?

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
PorterNan: Because we can no longer come back to the calling status for any chance - so stop showing the bar would be the best choice according to what Alex P designed

Disconnecting might still transmit video and audio for some milliseconds, but it has almost 100% chance to stop them correctly cause it is just a local operation, the only thing could be stuck is network related operation - saying that your stop all transmission but your tile is still on others' video gallery

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -50,6 +50,7 @@ export type CallCompositePage =
   | /* @conditional-compile-remove(PSTN-calls) */ 'hold'
   | 'joinCallFailedDueToNoNetwork'
   | 'leftCall'
+  | 'leaving'
```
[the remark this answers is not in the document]
PorterNan: The string from the UI design shows in the page is leaving, so maybe we just keep leaving as the page - cause we have leftCall as the next page

packages/react-composites/src/composites/CallComposite/styles/CallComposite.styles.ts
PorterNan: Good catch, will update this part soon

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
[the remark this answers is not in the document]
JamesBurnside: I like leaving as it mimic leftCall

JamesBurnside:
(needs e2e snapshot, can be separate PR)

dmceachernmsft (approved):

PR #3113, opened 2023-05-29
[Chat] Add unit tests for Mention in MessageThread component

jimchou-dev:
# What

- Unit tests for Mention in MessageThread component
- Unit test for editing message with Mention and selection in MessageThread Component
- Added devDependency for @testing-library/dom and @testing-library/user-event

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3278889

# How Tested

# Process & policy checklist


packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +225,290 @@ describe('Message should display inline image correctly', () => {
...
+    const user2Id = 'user2';
+    const user2Name = 'Robert Tolbert';
+
+    const messages: ChatMessage[] = [
+      {
+        messageType: 'chat',
+        senderId: user1Id,
+        senderDisplayName: 'Kat Larsson',
+        messageId: Math.random().toString(),
+        content: `Hey <msft-mention id="${user2Id}" displayText="${user2Name}"></msft-mention>, can you help me with my internet connection?`,
```
Leah-Xia-Microsoft, line 255: I think we need to have the display name inserted between the open and closing tag:
`Hey <msft-mention id="${user2Id}" displayText="${user2Name}">{user2Name}</msft-mention>, can you help me with my internet connection?`

jimchou-dev: Technically we don't need it between as part of the message content, as it will parse the `id` and `dispalyText` from the `msft-mention` attribute, then in the `defaultOnMentionRender` it will inject the proper tag with the `displayText` also in the content to be:
```
<msft-mention id="${user2Id}" displayText="${user2Name}">{user2Name}</msft-mention>
```
I purposely omit it so I can assert/expect the displayText is injected properly as part of `msft-mention` textContent

emlynmac: This is actually a little bit of a strange behaviour. The text between the tag should be rendered and in the SendBox, the enclosed text is used as the displayed text, ignoring the display text parameter.

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +225,290 @@ describe('Message should display inline image correctly', () => {
...
+
+    const { container, rerender } = render(
+      <MessageThread
+        userId={user2Id}
+        messages={messages}
+        onUpdateMessage={onUpdateMessageCallback}
+        mentionOptions={{
+          displayOptions: {
+            onRenderMention: (mention, defaultOnMentionRender) => {
+              return <span key={Math.random().toString()}>{defaultOnMentionRender(mention)}</span>;
```
emlynmac, line 428: Math.random should be used for randomness. For test we should be fully controlling the keys / IDs to expected values.

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +225,290 @@ describe('Message should display inline image correctly', () => {
...
+
+    rerender(
+      <MessageThread
+        userId={user2Id}
+        messages={messages}
+        onUpdateMessage={onUpdateMessageCallback}
+        mentionOptions={{
+          displayOptions: {
+            onRenderMention: (mention, defaultOnMentionRender) => {
+              return <span key={Math.random().toString()}>{defaultOnMentionRender(mention)}</span>;
```
emlynmac, line 496: Same here; avoid math.random unless you need some randomness

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +225,290 @@ describe('Message should display inline image correctly', () => {
...
+    expect(everyoneMentionContextMenuItem.classList.contains('ms-Persona-primaryText')).toBe(true);
+
+    // Check that user1Name is an option
+    const user1MentionContextMenuItem = await screen.findByText(user1Name);
+    expect(user1MentionContextMenuItem.classList.contains('ms-Persona-primaryText')).toBe(true);
+
+    // Select mention from popover for user1Name, verify plain text not contain mention html tag
+    fireEvent.click(user1MentionContextMenuItem);
+    expect(editBox.innerHTML).toContain(user1Name);
+    expect(editBox.innerHTML).not.toContain('msft-mention');
```
emlynmac, line 474: msft-mention is defined in a constant; we should use that.

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT license.
 
 import React from 'react';
+/* @conditional-compile-remove(mention) */
+import { v1 as generateGUID } from 'uuid';
```
emlynmac, line 6: Just a general thing with respect to UUIDs. v1 is fine for test / simple scenarios but if this were to be used where uniqueness is truly needed, v4 or v5 should be used.

packages/react-components/src/components/MessageThread.test.tsx
emlynmac: So as we're dropping the displayText parameter and using the text in the tags, we should have our tests reflect that.
It's also important to remember that browsers will render any custom tag in this way - using the content between the tags rather than the parameter - so we should be consistent

packages/react-components/src/components/MessageThread.test.tsx
jimchou-dev: Will keep it as v1 uuid since it's being used in several other places in the repo already and should be sufficient for testing purpose like you said, but thanks for sharing more details regarding v4/v5 UUID being unique

packages/react-components/src/components/MessageThread.test.tsx
emlynmac: so now that we walked back that change, we _do_ need the tag content to contain the displayText - and we should ensure that this is the case when rendered

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,292 @@ describe('Message should display inline image correctly', () => {
...
+    const user2Name = 'Robert Tolbert';
+    const user3Name = 'Sam Fisher';
+
+    const messages: ChatMessage[] = [
+      {
+        messageType: 'chat',
+        senderId: user1Id,
+        senderDisplayName: 'Kat Larsson',
+        messageId: generateGUID(),
+        content: `Hey <msft-mention id="${user2Id}" displayText="${user2Name}"></msft-mention>, can you help me with my internet connection?`,
```
emlynmac, line 311: Same here, add the tag body

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,292 @@ describe('Message should display inline image correctly', () => {
...
+          }
+        }}
+      />
+    );
+
+    expect(container.querySelector(`#${user2Id}`)?.nodeName.toLowerCase()).toEqual(MSFT_MENTION);
+    expect(container.querySelector(`#${user2Id}`)?.textContent).toEqual(user2Name);
+
+    // edit message
+    const message1ContentAfterEdit = `Hey <msft-mention id="${user2Id}" displayText="${user2Name}"></msft-mention> and <msft-mention id="${user3Id}" displayText="${user3Name}"></msft-mention>, can you help me with my internet connection?`;
```
emlynmac, line 349: And all other cases :)

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,292 @@ describe('Message should display inline image correctly', () => {
...
+    fireEvent.click(menuButton);
+
+    // Click on Edit ContextMenuItem
+    const editButton = await screen.findByText('Edit');
+    fireEvent.click(editButton);
+
+    // Type ' @' in edit box to show mentions popover menu
+    const editBox = await screen.getByPlaceholderText('Edit your message');
+    await waitFor(async () => {
+      userEvent.keyboard(' @');
```
vhuseinova-msft, line 464: Should 'await' be added here to wait while the whole text is added to the input box?

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,292 @@ describe('Message should display inline image correctly', () => {
...
+    };
+
+    const { container, rerender } = render(
+      <MessageThread
+        userId={user2Id}
+        messages={messages}
+        onUpdateMessage={onUpdateMessageCallback}
+        mentionOptions={{
+          displayOptions: {
+            onRenderMention: (mention, defaultOnMentionRender) => {
```
emlynmac, line 431: Just curious; why use the custom renderer option here?

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,292 @@ describe('Message should display inline image correctly', () => {
...
+    if (!messageBubble) {
+      fail('it should not reach here');
+    }
+    expect(messageBubble.innerHTML).not.toContain(user1Name);
+    expect(messageBubble.innerHTML).not.toContain(MSFT_MENTION);
+
+    // Click on ... button to trigger context menu
+    const menuButton = container.querySelector('[data-ui-id="chat-composite-message-action-icon"]');
+    if (!menuButton) {
+      fail('it should not reach here');
```
emlynmac, line 453: Maybe a more descriptive fail message so we can get to the specific case quickly?

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,292 @@ describe('Message should display inline image correctly', () => {
...
+    // Select mention from popover for user1Name, verify plain text not contain mention html tag
+    fireEvent.click(user1MentionContextMenuItem);
+    expect(editBox.innerHTML).toContain(user1Name);
+    expect(editBox.innerHTML).not.toContain(MSFT_MENTION);
+
+    // Submit edited message
+    const submitButton = await screen.findByLabelText('Submit');
+    fireEvent.click(submitButton);
+
+    // Veerify message has new edited content includes mention html tag
```
emlynmac, line 484:
```suggestion
    // Verify message has new edited content includes mention HTML tag

```

packages/react-components/src/components/MessageThread.test.tsx
jimchou-dev: Removed, it was me not understanding how to use the options code

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,258 @@ describe('Message should display inline image correctly', () => {
...
+    const expectedOnRenderMentionCount = 2;
+    let onRenderMentionCount = 0;
+
+    rerender(
+      <MessageThread
+        userId={user2Id}
+        messages={messages}
+        mentionOptions={{
+          displayOptions: {
+            onRenderMention: (mention, defaultOnMentionRender) => {
```
emlynmac, line 337: Do we need a custom renderer?

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,258 @@ describe('Message should display inline image correctly', () => {
...
+              onRenderMentionCount++;
+              return <span key={generateGUID()}>{defaultOnMentionRender(mention)}</span>;
+            }
+          }
+        }}
+      />
+    );
+
+    expect(onRenderMentionCount).toEqual(expectedOnRenderMentionCount);
```
emlynmac, line 346: And stronger here to compare the IDs captured to the expected values

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,258 @@ describe('Message should display inline image correctly', () => {
...
+    fireEvent.click(menuButton);
+
+    // Click on Edit ContextMenuItem
+    const editButton = await screen.findByText('Edit');
+    fireEvent.click(editButton);
+
+    // Type ' @' in edit box to show mentions popover menu
+    const editBox = await screen.getByPlaceholderText('Edit your message');
+    await waitFor(async () => {
+      await userEvent.keyboard(' @');
```
emlynmac, line 435: Can we have a test that includes overriding the trigger and verifying that?

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -225,3 +227,258 @@ describe('Message should display inline image correctly', () => {
...
+    });
+
+    // Check that Everyone is an option
+    const everyoneMentionContextMenuItem = await screen.findByText('Everyone');
```
emlynmac, line 439: Should this look up the locale-specific text?

packages/react-components/src/components/MessageThread.test.tsx
jimchou-dev: Not needed but want to test that re-rendering gets called twice

packages/react-components/src/components/MessageThread.test.tsx
emlynmac, line 435: Can we have a test that overrides the trigger and verifies the functionality?

packages/react-components/src/components/MessageThread.test.tsx
jimchou-dev: I think comparing the id doesn't really add much value, since the message.content will also have the id. So if we test for Id we can't tell since it's same before and after the onRenderMention render

packages/react-components/src/components/MessageThread.test.tsx
jimchou-dev: Can you be more specific? override what trigger and what functionality are we verifying?

change-beta/@azure-communication-react-8734ce38-a27b-4774-994a-ceb59f0e8a5c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
vhuseinova-msft, line 2: The type should be none

emlynmac (approved):

PR #3131, opened 2023-05-30
Prerelease beta/1.5.1 beta.5

prabhjot-msft:
# What
Prerelease beta/1.5.1 beta.5

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/common/VideoEffectsPane.tsx
```diff
@@ -146,7 +146,7 @@ const VideoEffectsPaneTrampoline = (
   /* @conditional-compile-remove(video-background-effects) */
   return (
     <Stack horizontalAlign="center">
-      {videoEffectError && (
+      {videoEffectError && isCameraOn && (
```
alkwa-msft, line 149: is this correct? I thought bug fixes should be coming int from main into the release branch and so we shouldnt be seeing any new code when we merge back into main?

JamesBurnside (approved):

JoshuaLai (approved):

packages/react-composites/src/composites/common/VideoEffectsPane.tsx
PorterNan: Is it in main branch already or not? Which fix is this coming from? Wondering why this is introduced here

PorterNan (approved):

PR #3132, opened 2023-05-30
[Chat] Team to ACS file sharing

jimchou-dev:
# What

- Rename conditional-compile-remove for `teams-inline-images` to `teams-inline-images-and-file-sharing` as they are tightly coupled feature (most file changes are just renaming)
- Teams file attachment `teamsImage` will be converted to teamsInlineImage and inject `` at the end of message content
- Teams file attachment `file` will be mapped to `fileSharing` and display `_FileDownloadCards`
- Fix fileSharing to check attachmentType not only `first` in list but `some` file to be `fileSharing`
- Fix registering file extension Icon dynamically
- FileDownloadCards differentiating ACS-to-ACS vs Teams-to-ACS file sharing by FileMetadata.previewUrl will be `undefined` for ACS-to-ACS file sharing
  - ACS-to-ACS `FileMetadata.previewUrl `will be `undefined`
  - Teams-to-ACS does not support direct download will have `ChatAttachment.previewUrl` will be mapped to the Url, and will hide Download Icon, when click on FileCard will open url to SharePoint

image

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3286174
https://skype.visualstudio.com/SPOOL/_workitems/edit/3286175

# How Tested

# Process & policy checklist


packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -95,7 +95,26 @@ const mapAttachmentType = (attachmentType: AttachmentType): FileMetadataAttachme
...
 };
 
-/* @conditional-compile-remove(file-sharing) @conditional-compile-remove(teams-inline-images) */
+const processTeamsImageContent = (message: ChatMessageWithStatus): string | undefined => {
+  if (sanitizedMessageContentType(message.type).includes('html') && message.content?.attachments) {
+    let teamsImageHtmlContent = '';
+    const attachments: ChatAttachment[] = message.content?.attachments;
+    attachments.forEach((attachment) => {
+      if (attachment.attachmentType === 'teamsImage') {
+        teamsImageHtmlContent += `\r\n<p><img alt="image" src="" itemscope="${attachment.contentType}" id="${attachment.id}"></p>`;
```
jpeng-ms, line 105: Do you want to have some CSS somewhere to cap the width of these images?

jimchou-dev: I tested with a 4000x2000 dimension image and it seems to be fine (I'm guessing that it uses previewUrl so AMS resizes it)?
image

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -168,8 +168,16 @@ export const _FileDownloadCards = (props: _FileDownloadCards): JSX.Element => {
   );
 
   const fileDownloadHandler = useCallback(
-    async (userId, file) => {
+    async (userId, file: FileMetadata) => {
       if (!props.downloadHandler) {
+        /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+        if (file.previewUrl) {
```
jpeng-ms, line 174: don't think this is a good idea though, might be more reliable to check based on file type ("teamsFile" vs. "fileSharing")?

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -95,7 +95,27 @@ const mapAttachmentType = (attachmentType: AttachmentType): FileMetadataAttachme
...
+      }
+    });
+    if (teamsImageHtmlContent === '') {
+      return message.content?.message;
+    }
+    const content = (message.content?.message ?? '') + teamsImageHtmlContent;
+    const trialingEmptyLine = '\r\n<p>&nbsp;</p>';
+    return content.endsWith(trialingEmptyLine) ? content.slice(0, content.lastIndexOf(trialingEmptyLine)) : content;
```
jpeng-ms, line 113: what is this for?

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -202,23 +212,30 @@ export const _FileDownloadCards = (props: _FileDownloadCards): JSX.Element => {
...
+            .map((file) => (
+              <_FileCard
+                fileName={file.name}
+                key={file.name}
+                fileExtension={file.extension}
+                actionIcon={
+                  showSpinner ? (
+                    <Spinner size={SpinnerSize.medium} aria-live={'polite'} role={'status'} />
+                  ) : true &&
+                    /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */ !file.previewUrl ? (
```
jpeng-ms, line 230: same thing here, i think using the actual enum type is more reliable. i don't think the assumption `if preview url exist, then apply teams file logic` is solid...

packages/chat-component-bindings/src/messageThreadSelector.ts
jimchou-dev: When Teams send inline image and attachment, it will have duplicated 'newLine' `<p>` tags this is to trim trailing new lines

packages/react-components/src/components/FileDownloadCards.tsx
jimchou-dev: In UI Component it has lost the information 'teamsFile' as it will convert 'file' to 'fileSharing' in the binding layer. The FileMetadataAttachmentType is only 'fileSharing', 'teamsInlineImage', and 'unknown'

packages/react-components/src/components/FileDownloadCards.tsx
jimchou-dev: In UI Component it has lost the information 'teamsFile' as it will convert 'file' to 'fileSharing' in the binding layer

jpeng-ms: right that's what i meant, i thought we were going to introduce "teamsFile" in addition to "fileSharing"?

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -95,7 +95,27 @@ const mapAttachmentType = (attachmentType: AttachmentType): FileMetadataAttachme
...
 };
 
-/* @conditional-compile-remove(file-sharing) @conditional-compile-remove(teams-inline-images) */
+const processTeamsImageContent = (message: ChatMessageWithStatus): string | undefined => {
+  /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+  if (sanitizedMessageContentType(message.type).includes('html') && message.content?.attachments) {
+    let teamsImageHtmlContent = '';
+    const attachments: ChatAttachment[] = message.content?.attachments;
+    attachments.forEach((attachment) => {
+      if (attachment.attachmentType === 'teamsImage') {
```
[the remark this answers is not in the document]
jimchou-dev: Updated

packages/react-components/src/components/FileDownloadCards.tsx
jpeng-ms: i don't think the idea of being agnostic makes sense here. At the end of the day, there are two types of files and two set of logics for these types. How does it make sense to differentiate them solely based on the content of preview url property? 

Moreover, there are teams file specific logic on the UI layer as well, I do think a new type is justified, which is something i thought we have agreed yesterday and you even mentioned ARB review for this specifically?

packages/chat-component-bindings/src/messageThreadSelector.ts
jpeng-ms: ah i see, if this is the case, do we want to wrap this into some if check so it only applies when we have both inline images and attachments

packages/react-components/src/components/FileDownloadCards.tsx
jpeng-ms: resolving this to continue the same conversion in the other comment i left

packages/react-components/src/components/FileDownloadCards.tsx
jpeng-ms: Also if we use the previewUrl to differentiate different file types, it might restrict how we can extend the attachment feature. For example, Dominik and I had discussed about how to incorporate the concept of preview Type to attachment files. So that we can have a preview card rendered in addition to attachment files such as this: 
image. 

If we were support this in the future, then Teams file attachment logic would be applied to ACS file attachments since both of them would have `previewUrl`

jimchou-dev: For now in the binding layer if attachment type is FileSharing will map previewUrl (sharePoint url) to url, and will still have previewUrl that will be used to hide 'Download' Icon.
Future refactoring task to extract into different FileMetadata arch type will happen in ADO 3242058

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -73,19 +73,19 @@ const extractAttachedFilesMetadata = (metadata: Record<string, string>): FileMet
...
-const extractInlineImageFilesMetadata = (attachments: ChatAttachment[]): FileMetadata[] => {
+/* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+const extractTeamsAttachmentsMetadata = (attachments: ChatAttachment[]): FileMetadata[] => {
   return attachments.map((attachment) => ({
     attachmentType: mapAttachmentType(attachment.attachmentType),
     id: attachment.id,
     name: attachment.name ?? '',
     extension: attachment.contentType ?? '',
-    url: attachment.url,
+    url: attachment.attachmentType === 'file' && attachment.previewUrl ? attachment.previewUrl : attachment.url,
```
JoshuaLai, line 83: nitpick: lets have this logic written into a new helper method like `mapAttachmentUrl` or something along those lines. Just so in the future if / when this grows it has a dedicated place for the logic.

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -95,7 +95,28 @@ const mapAttachmentType = (attachmentType: AttachmentType): FileMetadataAttachme
   return 'unknown';
 };
 
-/* @conditional-compile-remove(file-sharing) @conditional-compile-remove(teams-inline-images) */
+const processTeamsImageContent = (message: ChatMessageWithStatus): string | undefined => {
+  /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
```
JoshuaLai, line 99: I think you can move this conditional compile out above the `const`. I'm actually a little surprised the stable build didn't throw an error saying this was unused. Maybe cause this is ` | undefined`?

JoshuaLai: Oh I see in the stable it wold always just return `message.content?.message` which is what it used to do anyways...ok so this is fine

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -193,7 +193,9 @@ export const _FileDownloadCards = (props: _FileDownloadCards): JSX.Element => {
   if (
     !fileMetadata ||
     fileMetadata.length === 0 ||
-    /* @conditional-compile-remove(teams-inline-images) */ fileMetadata[0].attachmentType !== 'fileSharing'
+    /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */ !fileMetadata.some(
```
JoshuaLai, line 196: If we are making this change it is safe to remove the above comment 
```
// Its safe to assume that if the first item in the fileMetadata is not a fileSharing type we don't want to display the FileDownloadCard.
  // Since you can't have both fileSharing and teamsInlineImage in the same message.
```

JoshuaLai: Or at least update the comment so we understand why this check is in there.

packages/chat-component-bindings/src/messageThreadSelector.ts
jimchou-dev: We can't move this outside the const because `processTeamsImageContent` is being called in stable
image
But will rename it to something more generic so conditional-compile-remove will make sense

JoshuaLai (approved):

AmyL219 (approved):

jpeng-ms (approved):

PR #3137, opened 2023-06-01
Dismiss video effects error when camera is off

prabhjot-msft:
# What
Dismiss video effects error when camera is off

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3288350

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -230,7 +235,7 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
   }
 
   /* @conditional-compile-remove(video-background-effects) */
-  if (useIsParticularSidePaneOpen('videoeffects') && props.errorBarProps) {
+  if ((useIsParticularSidePaneOpen('videoeffects') || !isCameraOn) && props.errorBarProps) {
```
JamesBurnside, line 238: We haven't used this anywhere in the codebase yet, but instead of hiding the error when the camera is off, could we clear the error in the the handler for toggleCamera using `newClearCallErrorsModifier`?
image

Honestly no idea if that works just curious if its possible.

JamesBurnside: The way its currently implemented, does the error re-show when the camera is turned back on?

edwardlee-msft: is this supposed to be !isCameraOn?
The title indicates that it should dismiss when camera is off, but this is settings an activemessage when camera is off.

I see that in videoEffectsPane.tsx line 149 new change it does what I expected.
Correct me if I'm wrong here.

edwardlee-msft, line 238: is this supposed to be !isCameraOn?
The title indicates that it should dismiss when camera is off, but this is setting an activemessage when camera is off.

I see that in videoEffectsPane.tsx line 149 new change it does what I expected.
Correct me if I'm wrong here.

prabhjot-msft: - For 1st suggestion, using `newClearCallErrorsModifier`, this modifier is not actually exposed in the calling stateful layer. If we are planning to expose this modifier, it would be called in the stateful layer mostly in the call context by introducing new set errors method. We can trigger this method whenever we want to delete certain errors. I can pursue this as a separate task if we feel strongly about this approach.

- For 2nd question, as part of the second change for auto-dismiss(https://github.com/Azure/communication-ui-library/pull/3160). Whenever camera is turned on, it will reset a new active effect and if the error is generated again it would be shown and otherwise no error will be shown.

prabhjot-msft: Yeah, so in this scenario, if `videoeffectsSidepane` is `open` and `camera` is `off` we are filtering out the video effects error from latest errors.
For videoeffectsPane, if the `videoeffectsError` exists and `camera is on` we are showing the error.

JamesBurnside (approved):

edwardlee-msft (approved):

PR #3139, opened 2023-06-02
[Chat]Improve Accessibility for Files Sharing feature

AmyL219:
# What

Improve Accessibility for Files Sharing feature

# Why

# How Tested

Test in Storybook Message Thread is the easiest approach

# Process & policy checklist


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -244,6 +244,9 @@ export interface MessageThreadStrings {
   /* @conditional-compile-remove(data-loss-prevention) */
   /** String for policy violation message removal details link */
   blockedWarningLinkText: string;
+  /* @conditional-compile-remove(file-sharing) */
```
JoshuaLai, line 247: @jimchou-dev can correct me if I am wrong but I think we also want a `/* @conditional-compile-remove(teams-inline-images-and-file-sharing) */` here as well. The teams file sharing use the same rails so it should also be applicable

JoshuaLai (approved):
The change looks fine there. There is just one tiny thing I want Jim to to verify

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -244,5 +252,5 @@ const DownloadIconTrampoline = (): JSX.Element => {
 const useLocaleStringsTrampoline = (): _FileDownloadCardsStrings => {
   /* @conditional-compile-remove(file-sharing) */
```
jimchou-dev, line 253: add @conditional-compile-remove(teams-inline-images-and-file-sharing) here as well

packages/react-components/src/components/MessageThread.tsx
jimchou-dev: Yes correct and this applies to other places as well to add both
`/* @conditional-compile-remove(file-sharing) @conditional-compile-remove(teams-inline-images-and-file-sharing) */`

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -195,16 +195,14 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
         fileMetadata={message['attachedFilesMetadata'] || []}
         downloadHandler={fileDownloadHandler}
         /* @conditional-compile-remove(file-sharing) */
-        strings={{ downloadFile: props.strings.downloadFile ?? locale.strings.messageThread.downloadFile }}
+        strings={{ downloadFile: props.strings.downloadFile, fileCardGroupMessage: props.strings.fileCardGroupMessage }}
```
alkwa-msft, line 198: nit. Can we use a syntax similar to this const { strings } = props;
so then we can just say fileCardGroupmessage: strings.fileCardGroupMessage?

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -161,12 +163,18 @@ export const _FileDownloadCards = (props: _FileDownloadCards): JSX.Element => {
...
-      // Return download button without aria label
-      return props.strings?.downloadFile ?? '';
     },
     [props.strings?.downloadFile, localeStrings.downloadFile]
   );
 
+  const fileCardGroupDescription = useMemo(
+    () => () => {
+      const fileGroupLocaleString = props.strings?.fileCardGroupMessage ?? localeStrings.fileCardGroupMessage;
+      return _formatString(fileGroupLocaleString, { fileCount: `${fileMetadata.length}` });
```
jimchou-dev, line 173: If there are 1 attachedimage and 2 attached file it will announce there is 3 attachments.
Maybe do a filter `filemetadata.filter(file => file.attachmentType === 'fileSharing').length` to get only attachments

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -195,16 +195,14 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
         fileMetadata={message['attachedFilesMetadata'] || []}
         downloadHandler={fileDownloadHandler}
         /* @conditional-compile-remove(file-sharing) */
```
jimchou-dev, line 197: add conditional compile for @conditional-compile-remove(teams-inline-images-and-file-sharing) here as well

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
AmyL219: Good catch!

packages/react-components/src/components/FileDownloadCards.tsx
AmyL219: Good catch!

packages/react-components/src/components/MessageThread.tsx
AmyL219: updated

jimchou-dev (approved):
LGTM

PR #3141, opened 2023-06-02
Call composite transfer page

mgamis-msft:
# What
- Added TransferPage to Call composite and showing it when CallAdapterState.acceptedTransferCall is defined
- Added new strings
- Updated existing AvatarPersona component to update when userId changes to undefined

# Why
To show a page when a call is being transferred as per [Figma](https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?type=design&node-id=9186-105927&t=4uQjGHFeztaDTUvy-4)
image

https://skype.visualstudio.com/SPOOL/_workitems/edit/3248904
https://skype.visualstudio.com/SPOOL/_workitems/edit/3253807
https://skype.visualstudio.com/SPOOL/_workitems/edit/3253809

# How Tested
Local calling sample:
https://github.com/Azure/communication-ui-library/assets/79475487/38b8cea3-b558-4521-9997-24a343f02de0

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/pages/TransferPage.tsx
```diff
@@ -0,0 +1,225 @@
...
+  // Reduce the controls shown when mobile view is enabled.
+  const callControlOptions = props.mobileView
+    ? reduceCallControlsForMobile(props.options?.callControls)
+    : props.options?.callControls;
+
+  /* @conditional-compile-remove(call-transfer) */
+  const pageState: TransferPageState = useMemo(() => {
+    if (transferCall !== undefined) {
+      if (['Ringing', 'Connected'].includes(transferCall.state)) {
+        return 'connecting';
```
dmceachernmsft, line 57: is this when we are being connected to the 'transfer' that is going to transfer us?

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/pages/TransferPage.tsx
mgamis-msft: The accepted transfer call that is already in CallClientState is in the state 'Ringing' or 'Connected' state. That is when the transfer target actually sees our call so that is when we are showing that we are in the 'connecting' page state as shown in this Figma.
image
Maybe the page state 'connecting' is easily confused with the call state. Will try think of a better name.

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -77,7 +77,11 @@ export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
 
   useEffect(() => {
     (async () => {
-      if (dataProvider && userId) {
+      if (!userId) {
+        setData(undefined);
```
mgamis-msft, line 81: We should set persona data to undefined so that custom avatar is undone when userId changes to undefined

packages/react-composites/src/composites/CallComposite/pages/TransferPage.tsx
mgamis-msft: Good question. I changed the pageState variable to pageSubject. It is too easy to confuse with the call state. Page subject is basically who should be shown in the Transfer page, the transferor or the transfer target.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -677,6 +677,10 @@ export interface CallCompositeStrings {
     startCaptionsButtonTooltipOnContent?: string;
     threeParticipantJoinedNoticeString: string;
     threeParticipantLeftNoticeString: string;
+    transferPageTransferorText: string;
```
PorterNan, line 680: Make these strings optional - which is most common feedback coming from ARB

packages/react-composites/src/composites/CallComposite/pages/TransferPage.tsx
```diff
@@ -0,0 +1,231 @@
...
+import { useSelector } from '../hooks/useSelector';
+import { getRemoteParticipants } from '../selectors/baseSelectors';
+import { toFlatCommunicationIdentifier } from '@internal/acs-ui-common';
+import { AvatarPersona, AvatarPersonaDataCallback } from '../../common/AvatarPersona';
+/* @conditional-compile-remove(call-transfer) */
+import { getTransferCall } from '../selectors/baseSelectors';
+import { reduceCallControlsForMobile } from '../utils';
+import { LobbyPageProps } from './LobbyPage';
+
+/* @conditional-compile-remove(call-transfer) */
```
PorterNan, line 28: Do we need to CC this line, I guess no?

PorterNan: NVM, if you don't CC this, it will be an unused line

packages/react-composites/src/composites/CallComposite/pages/TransferPage.tsx
```diff
@@ -0,0 +1,231 @@
...
+  const transferCall = useSelector(getTransferCall);
+
+  // Reduce the controls shown when mobile view is enabled.
+  const callControlOptions = props.mobileView
+    ? reduceCallControlsForMobile(props.options?.callControls)
+    : props.options?.callControls;
+
+  /* @conditional-compile-remove(call-transfer) */
+  // page subject is which should be participant shown in the transfer page depending on the transfer call state
+  const pageSubject: TransferPageSubject = useMemo(() => {
```
PorterNan, line 55: So at the very beginning, the page shows transferor then once we have the transferTarget, the page will show the transferTarget instead?

packages/react-composites/src/composites/CallComposite/pages/TransferPage.tsx
```diff
@@ -0,0 +1,231 @@
...
+        <Stack horizontal horizontalAlign="center" verticalAlign="center" tokens={{ childrenGap: '0.5rem' }}>
+          <Spinner size={SpinnerSize.large} styles={{ circle: { borderWidth: '0.125rem' } }} />
+          <Text className={mergeStyles({ textAlign: 'center', fontSize: '1rem' })}>{statusString}</Text>
+        </Stack>
+      </Stack>
+    </div>
+  );
+};
+
+const videoContainerStyles: IStyle = {
```
PorterNan, line 209: Consider open a separated css file for these Istyles

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -139,7 +139,8 @@ type GetCallCompositePageFunction = ((
     unsupportedBrowserInfo?: {
       environmentInfo?: EnvironmentInfo;
       unsupportedBrowserVersionOptedIn?: boolean;
-    }
+    },
+    /* @conditional-compile-remove(call-transfer) */ transferCall?: CallState
```
PorterNan, line 143: Is this line under the /* @conditional-compile-remove(unsupported-browser) */ flag?
I guess it would be not independently removable then?

packages/react-composites/src/composites/common/AvatarPersona.tsx
PorterNan: Under which case userId would be undefined? Is it when transferring the call but not connected yet? If so, why would the component still use the same old instance? Shouldn't that be a new instance of component + participant with undefined userId?

PorterNan (approved):

PorterNan (commented):
A question maybe not related to this PR:
Have we added this feature to in-progress feature in .babelrc? Not sure if we want this one in next beta release or not

packages/react-composites/src/composites/common/AvatarPersona.tsx
mgamis-msft: Yes you are right. The transfer target participant should not have a userId that is undefined. Reverting these changes.

packages/react-composites/src/composites/CallComposite/pages/TransferPage.tsx
mgamis-msft: Yes, that's right

PR #3142, opened 2023-06-02
[Chat] Add tests for MentionPopover rendering

vhuseinova-msft:
# What

Added unit tests for showing and hiding mention popover

# Why

Increate tests coverage for the mention feature

# How Tested

# Process & policy checklist


packages/react-components/src/components/InputBoxComponent.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+      <InputBoxComponent
+        inlineChildren={true}
+        textValue={changedValue}
+        placeholderText="Enter a message"
+        onChange={(_, newValue) => {
+          changedValue = newValue ?? '';
+        }}
+        maxLength={1000}
+      >
+        <div></div>
```
emlynmac, line 25: I'm curious as to why the div is needed here?

vhuseinova-msft: `InputBoxComponent` has a required children props and can't be used as a self closing tag

packages/react-components/src/components/InputBoxComponent.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+    {
+      id: '1',
+      displayText: 'Test User1'
+    },
+    {
+      id: 'everyone',
+      displayText: 'Everyone'
+    }
+  ];
+  const trigger = '@';
```
jimchou-dev, line 52: Maybe add a test for trigger with other character that is not '@' or with trigger that is more than 1 character eg '@@'

vhuseinova-msft: Added!

jimchou-dev (approved):
LGTM

PR #3148, opened 2023-06-05
Custom data model to allow changing display name of call queues

mgamis-msft:
# What
Custom data model to change display name of call queues

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3248534

# How Tested
Calling sample test
https://github.com/Azure/communication-ui-library/assets/79475487/3eb1c9a8-1bec-4a5b-9741-5001998aa104

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/OnFetchProfileCallback.ts
```diff
@@ -48,10 +44,7 @@ export const createProfileStateModifier = (
         if (profile?.displayName && originalParticipants[key].displayName !== profile?.displayName) {
           cachedDisplayName[key] = profile?.displayName;
         }
-        shouldNotifyUpdates = true;
       }
-      // notify update only when there is a change, which most likely will trigger modifier and setState again
-      shouldNotifyUpdates && notifyUpdate();
```
mgamis-msft, line 54: @PorterNan do you know why we notify an update here? This caused an infinite cycle when I set onFetchProfile on the CallAdapterOptions

PorterNan: Cause displayName could be just undefined at the very beginning then becomes available later
The workflow is like
async getDisplayName -> displayName still undefined -> render component -> finished async call and get displayName and render it, we need to notify update to adapter to tell it the avatar state has been changed

PorterNan: We need to ensure displayName is always the same from the Onfetch functions (and the key is not undefined I guess?), let's have an offline session together to debug this one

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -1050,6 +1046,12 @@ export type CommonCallAdapterOptions = {
    * Default set of background images for background image picker.
    */
   videoBackgroundImages?: VideoBackgroundImage[];
+  /**
+   * Use this to fetch profile information which will override data in {@link CallAdapterState} like display name
+   * The onFetchProfile is fetch-and-forget one time action for each user, once a user profile is updated, the value will be cached
+   * and would not be updated again within the lifecycle of adapter.
+   */
+  onFetchProfile?: OnFetchProfileCallback;
```
dmceachernmsft, line 1054: Is this something that can be overridden for just normal users as well?

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -224,3 +226,12 @@ const videoBackgroundImages = [
     tooltipText: 'Custom Background'
   }
 ];
+
+/* @conditional-compile-remove(teams-adhoc-call) */
+const onFetchProfile = async (userId: string, defaultProfile?: Profile): Promise<Profile | undefined> => {
+  // TODO: Explore using Graph API to differentiate Call queues from other Voice app bots
+  if (userId.startsWith('28:orgid:')) {
```
dmceachernmsft, line 233: is this always true? I believe all bots use this prefix

packages/react-composites/src/composites/CallComposite/adapter/OnFetchProfileCallback.ts
mgamis-msft: Thanks for the clarification. I reverted the changes here and the infinite loop I was seeing is no longer happening. My mistake.

mgamis-msft: I see . Thanks for the clarification. Turns out, I only needed to set the `shouldNotifyUpdates` flag to true when the display name is different from the cached display name.

samples/Calling/src/app/views/CallScreen.tsx
edwardlee-msft: Don is right, Bots use this prefix.

edwardlee-msft: actually, bots use 28: prefix. The orgid is something different. 28:orgid could be specific to call queue as they would technically be the organizer of the meeting. Can we confirm if other bots can be organizers?

dmceachernmsft: Oh this is a good point ^ I think if call queues are special and get the orgid moniker like users this might actually be good, is there someone on VoiceApps you can reach out to about this @mgamis-msft?

mgamis-msft: The 28: prefix is actually for resource accounts in Microsoft 365. Each resource account can be assigned to a call queue or an auto attendant. Even Graph API can't know which Voice App a resource account is assigned to. 
image.
I will remove this code for our Calling sample but the point of this PR is now Contoso should be able to change the display name of their call queue if they know the resource account user id which they can do using Graph API.

edwardlee-msft (approved):

dmceachernmsft (approved):

PR #3155, opened 2023-06-06
[Chat] Add tests for SendBox

vhuseinova-msft:
# What

Added tests for sending a mention as part of the SendBox component value

# Why

Increase tests coverage for the mention feature

# How Tested

# Process & policy checklist


packages/react-components/src/components/SendBox.test.tsx
```diff
@@ -6,6 +6,12 @@ import { SendBox } from './SendBox';
 import { renderWithLocalization, createTestLocale } from './utils/testUtils';
 import { screen } from '@testing-library/react';
 import { registerIcons } from '@fluentui/react';
+// /* @conditional-compile-remove(mention) */
```
Leah-Xia-Microsoft, line 9: Would this still work if the conditional compile comment is commented out?

vhuseinova-msft: Goot catch! It looks like it works but I'll fix it

Leah-Xia-Microsoft (approved):

PR #3156, opened 2023-06-06
[Chat] introducing new type for future scalability

JoshuaLai:
# What
Per refactor conversation:
Introducing a 2 new types: `CustomFileMetadata` and `TeamsInteropFileMetadata` 
The old `FileMetadata` is now `export type FileMetadata = CustomFileMetadata | TeamsInteropFileMetadata;`
This gives us more flexibility to introduce new types in the future. 

# Why
Instead of introducing new optional properties in the future we can instead just introduce a new type to perform type safe checks against. 

# How Tested

# Process & policy checklist


packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -220,7 +220,7 @@ describe.only('Message should display image and attachment correctly', () => {
       onFetchAttachmentCount++;
       return [
         {
-          blobUrl: attachment.previewUrl ?? ''
+          blobUrl: (attachment as TeamsInteropFileMetadata).previewUrl ?? ''
```
JamesBurnside, line 223: nit: avoid `as` as much as possible as it circumvents the type guarding. Instead if you use your helper function `isTeamsInteropFileMetadata` it will strongly the type the attachment:

```
if (isTeamsInteropFileMetadata(attachement) {
  // you can now use attachment.previewUrl without typecasting
}
```

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -57,6 +77,23 @@ export interface FileMetadata {
...
+  | /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */ TeamsInteropFileMetadata;
+
+/* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+/**
+ * @beta
+ * @param fileMetadata
+ * @returns TeamsInteropFileMetadata
+ */
+export const isTeamsInteropFileMetadata = (fileMetadata: FileMetadata): fileMetadata is TeamsInteropFileMetadata => {
+  return (fileMetadata as TeamsInteropFileMetadata).id !== undefined;
```
JamesBurnside, line 95: can we use the FileMetadataAttachmentType here instead of `id`, `id` seems like something other types could have in future

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2024,6 +2024,13 @@ export interface CustomControlButtonProps {
     text?: string;
 }
 
+// @beta (undocumented)
+export interface CustomFileMetadata {
```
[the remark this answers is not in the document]
JoshuaLai: We've had a lot of discussions about our names and wanting them to the acs agnostic. If thats the case we may steer away from the term `TeamsInteropFileMetadata` all together. That being said, I think we can make a case for this specific situation cause after all it is teams specific behaviour we are designing for.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2654,6 +2651,9 @@ export type IsSpeakingChangedListener = (event: {
     isSpeaking: boolean;
 }) => void;
 
+// @beta (undocumented)
+export const isTeamsInteropFileMetadata: (fileMetadata: FileMetadata) => fileMetadata is TeamsInteropFileMetadata;
```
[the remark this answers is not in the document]
JoshuaLai: No, I think if there is a need we can always expose it later. But I don't see Contoso needing it at the moment

packages/react-components/src/components/MessageThread.test.tsx
JoshuaLai: Updated to se the helper where ever possible

packages/react-components/src/components/FileDownloadCards.tsx
JoshuaLai: Checking against the type now

packages/communication-react/review/beta/communication-react.api.md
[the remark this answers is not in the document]
JoshuaLai: Removed the export

JamesBurnside:
A little overloaded right now, reassigning to @PorterNan

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -3503,6 +3501,30 @@ export interface TeamsCallingHandlers extends CommonCallingHandlers {
...
+    attachmentType: 'fileSharing';
+    extension: string;
+    // (undocumented)
+    id: string;
+    name: string;
+    url: string;
+}
+
+// @beta (undocumented)
+export interface TeamsInteropImageFileMetadata {
```
JoshuaLai, line 3516:
```suggestion
export interface ImageFileMetadata {
```

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -3503,6 +3501,30 @@ export interface TeamsCallingHandlers extends CommonCallingHandlers {
     onStartCall: (participants: CommunicationIdentifier[], options?: StartCallOptions) => undefined | /* @conditional-compile-remove(teams-identity-support) */ TeamsCall;
 }
 
+// @beta (undocumented)
+export interface TeamsInteropFileMetadata {
+    // (undocumented)
+    attachmentType: 'fileSharing';
+    extension: string;
+    // (undocumented)
+    id: string;
```
JoshuaLai, line 3510: attachmentType: 'fileSharingWithOptions';
options?: { "string": "string" }

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2089,6 +2089,14 @@ export interface CustomControlButtonProps {
     text?: string;
 }
 
+// @beta (undocumented)
+export interface CustomFileMetadata {
```
JoshuaLai, line 2093:
```suggestion
export interface FileSharingMetadata {
```

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -3503,6 +3501,30 @@ export interface TeamsCallingHandlers extends CommonCallingHandlers {
     onStartCall: (participants: CommunicationIdentifier[], options?: StartCallOptions) => undefined | /* @conditional-compile-remove(teams-identity-support) */ TeamsCall;
 }
 
+// @beta (undocumented)
+export interface TeamsInteropFileMetadata {
```
JoshuaLai, line 3505:
```suggestion
export interface FileSharingMetadataWithOptions {
```

packages/react-composites/tests/app/chat/FakeAdapterApp.tsx
```diff
@@ -125,8 +125,7 @@ const handleFileUploads = (adapter: ChatAdapter, fileUploads: _MockFileUpload[])
         name: file.name,
         extension: file.extension,
         url: file.url,
-        attachmentType: 'fileSharing',
-        id: ''
+        attachmentType: 'fileSharing'
```
jimchou-dev, line 128: should this be `attachmentType: file.attachmentType`?

JoshuaLai: Yep updated

jimchou-dev (approved):

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2688,6 +2699,19 @@ export interface _Identifiers {
     videoTile: string;
 }
 
+// @beta
+export interface ImageFileMetadata {
```
PorterNan, line 2703: I feel this one should extend FileSharingMetadata right?

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2515,21 +2515,32 @@ export interface FileDownloadError {
...
+// @beta
+export interface FileSharingMetadata {
+    attachmentType: 'fileSharing';
+    extension: string;
+    name: string;
+    url: string;
+}
+
+// @beta
+export interface FileSharingMetadataWithOptions {
```
PorterNan, line 2532: FileSharingMetadataWithOptions is an extended interface of FileSharingMetadata, but only with one more id and options

Is it possible to make id optional? then directly add "options" into FileSharingMetadata so we don't need FileSharingMetadataWithOptions?

BTW, the property name options sounds a bit not accurate, it's more like metadata or payload

packages/communication-react/review/beta/communication-react.api.md
JoshuaLai: No unfortunately because we assign the `attachmentType` in `FileSharingMetadata` to be `fileSharing` this won't work.

jimchou-dev: We cannot do that because `FileSharingMetadata` have `attachmentType` and when extending we cannot assign it a different value.
Originally we implemented them to extend from `BaseFileMetadata`, but it felt weird with the stable is using FileMetadata already so reverted the change.
```
interface BaseFileMetadata {
  name: string;
  url: string;
  extension: string;
}
```
We plan to add the base type and extend from it when the feature is closer to GA so stable will not feel odd.

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -2532,21 +2541,19 @@ export interface FileDownloadError {
...
+// @beta
+export interface FileSharingMetadata extends BaseFileMetadata {
     // (undocumented)
-    id: string;
-    name: string;
+    attachmentType: 'fileSharing';
     // (undocumented)
-    previewUrl?: string;
-    url: string;
+    payload?: Record<string, string> | undefined;
```
PorterNan, line 2554: You don't need to add | undefined if the property is defined by `:?`

packages/react-components/review/beta/react-components.api.md
```diff
@@ -1202,6 +1209,14 @@ export interface _Identifiers {
     videoTile: string;
 }
 
+// @beta
+export interface ImageFileMetadata extends BaseFileMetadata {
+    // (undocumented)
+    attachmentType: 'inlineImage' | 'attachedImage';
```
PorterNan, line 1215: Is it possible to use only one attachmentType for one type
The attachmentType property is more like a `kind` property for identify the object type, so a one to one mapping would be nice
or we use another parameter name as `kind` and this one is dedicated for attachedImage or inlineImage

packages/react-components/review/stable/react-components.api.md
```diff
@@ -716,10 +723,10 @@ export interface FileDownloadError {
...
 
 // @beta
-export interface FileMetadata {
-    extension: string;
-    name: string;
-    url: string;
+export type FileMetadata = FileSharingMetadata;
+
+// @beta
+export interface FileSharingMetadata extends BaseFileMetadata {
```
PorterNan, line 729: Looks like this is a beta type leak? Or please mark these classes as //@public if there are for stable

PorterNan (approved):

packages/react-components/review/beta/react-components.api.md
JoshuaLai: Confirmed that we can remove the `attachedImage` type since it is not being used

PR #3157, opened 2023-06-07
[Bugfix] useAzureCommunicationCallAdapter multiple call client fix

dmceachernmsft:
# What

Protects the hook from creating a second call client before the creation of the first adapter. now prints out information to new storybook section.
image

# Why

image

https://skype.visualstudio.com/SPOOL/_workitems/edit/3241530
# How Tested

image


packages/storybook/stories/Troubleshooting.stories.mdx
```diff
@@ -80,3 +80,19 @@ You may need an appropriate loader to handle this file type, currently no loader
 This issue will occur when using `@azure/communication-react` with a webpack version < 5.
 The best course of action is to upgrade your webpack version to 5.
 If this is not possible, take a look at [these workarounds](https://github.com/fb55/htmlparser2/issues/1237#issuecomment-1182522861).
+
+## Runtime errors
+
+### Failed to create call agent, already exists for this ACS id
+
+```
+Uncaught (in promise) CallError: CallClient.createCallAgent: Failed to create call agent, call agent for this ACS Id already exists
```
JamesBurnside, line 89: this shouldn't happen anymore right? It should be the new console warn message?

dmceachernmsft: They might see this error from other functions that are invoked as well like the direct create functions. I think that there is value in providing some information for whats going on here.

packages/storybook/stories/Troubleshooting.stories.mdx
```diff
@@ -80,3 +80,19 @@ You may need an appropriate loader to handle this file type, currently no loader
...
+    at toCallError (CallContext.ts:739:16)
+    at CallContext.eval (CallContext.ts:704:35)
+    at Generator.throw (<anonymous>)
+    at rejected (CallContext.ts:21:65)
+```
+
+We can see this error at times when some calls to construct our `AzureCommunicationCallAdapter` or `StatefulCallClient`. This can be caused by passing the arguments to these functions more than once or too often.
```
edwardlee-msft:
```suggestion
This error can occur when constructing `AzureCommunicationCallAdapter` or `StatefulCallClient`, often caused by passing arguments more than once or too frequently to these functions.
```

JamesBurnside (approved):

PR #3158, opened 2023-06-07
[Chat] Add unit tests for Mention Popover position

Leah-Xia-Microsoft:
# What

Added unit tests for Mention Popover position

# Why

Increase tests coverage for the mention feature

# How Tested

# Process & policy checklist


packages/react-components/src/components/MentionPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+      bottom: 0,
+      left: 0
+    };
+
+    renderMentionPopoverComponent(targetRect, targetPositionOffset, 'above');
+
+    const mentionPopover = await screen.findByTestId(mentionSuggestionListContainerId);
+    const elements = document.getElementsByClassName(mentionPopover.className);
+    const style = window.getComputedStyle(elements[0]);
+    expect(style.position).toBe('absolute');
```
vhuseinova-msft, line 81: I'm wondering if it worth to have snapshop tests for this functionality

Leah-Xia-Microsoft: > I'm wondering if it worth to have snapshop tests for this functionality

Chatted with Calling team, we need to do a lot of setup to be able to add snapshots for component only features. We'll focus on unit test for Public Preview.

packages/react-components/src/components/MentionPopover.tsx
```diff
@@ -299,6 +300,7 @@ export const _MentionPopover = (props: _MentionPopoverProps): JSX.Element => {
   return (
     <div ref={popoverRef}>
       <Stack
+        data-testid={testIds.mentionSuggestionListContainer}
```
emlynmac, line 303: We should not have test code populating shipped components

Leah-Xia-Microsoft: > We should not have test code populating shipped components

We need the testid as  query when there's no display value of the component. https://testing-library.com/docs/queries/about

Adding test id to the shipped components should be normal. We have `data-ui-id` everywhere which is used for end to end testing. This data-testid is for unit testing.

vhuseinova-msft (approved):

emlynmac (approved):

PR #3167, opened 2023-06-08
[Chat] Mention logic unit tests

emlynmac:
# What
Add unit tests for the mention logic handlers

# Why
Verify robustness and confidence in future changes

# How Tested
Command line `rushx test`

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.test.ts
+++ b/packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.test.ts
@@ -128,6 +128,15 @@ describe('Mention logic should be robust and accurate', () => {
     expect(parsed?.plainText).toEqual(comparisonString);
   });
 
+  test('Using /> in the text does not break parsing', () => {
+    const comparisonString = 'x/>y and z';
+    let parsed: { tags: TagData[]; plainText: string } | undefined;
+    expect(() => {
+      parsed = textToTagParser(comparisonString, '@');
+    }).not.toThrow();
+    expect(parsed?.plainText).toEqual(comparisonString);
+  });
+
   test('Basic HTML generation works', () => {
     const htmlString = htmlStringForMentionSuggestion({ id: '1', displayText: 'Everyone' }, localeStrings);
     expect(htmlString).toEqual('<msft-mention id="1" displayText="Everyone">Everyone</msft-mention>');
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.test.ts
```diff
@@ -0,0 +1,387 @@
...
+  textToTagParser,
+  updateHTML
+} from './mentionTagUtils';
+
+import { COMPONENT_LOCALE_EN_US } from '../../localization/locales';
+
+describe('Mention logic should be robust and accurate', () => {
+  const localeStrings = COMPONENT_LOCALE_EN_US.strings;
+
+  const basicMention = 'Hello <msft-mention id="1" displayText="Everyone">everyone</msft-mention>!';
```
Leah-Xia-Microsoft, line 23: Can the displayName be as an attr be different than the text wrapped up between the tags?

emlynmac: Yes, it can be edited to be that way

Leah-Xia-Microsoft (approved):

Leah-Xia-Microsoft (approved):

packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.test.ts
```diff
@@ -0,0 +1,387 @@
...
+    const comparisonString = 'x<y and z';
+    let parsed: { tags: TagData[]; plainText: string } | undefined;
+    expect(() => {
+      parsed = textToTagParser(comparisonString, '@');
+    }).not.toThrow();
+    expect(parsed?.plainText).toEqual(comparisonString);
+  });
+
+  test('Using > in the text does not break parsing', () => {
+    const comparisonString = 'x>y and z';
```
vhuseinova-msft, line 123: Should we also add a test for `/>`?

vhuseinova-msft (approved):

vhuseinova-msft (approved):

packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.test.ts
emlynmac: We should probably add that when we get the handling of the input of < and > working

PR #3169, opened 2023-06-08
Remove flakey snapshots to regenerate them

edwardlee-msft:
# What

Remove flakey snapshots to regenerate new snapshots for these two pages.
(Might be safer to regenerate new snapshots for the whole set. Will confirm after running update_snapshots)

# Why

Flakey Visual regression tests. Failing with only 1 pixel difference.

# How Tested

# Process & policy checklist


JamesBurnside (approved):
I see no changes to any of these snapshots .. approved! 😄

jimchou-dev (approved):

PR #3170, opened 2023-06-08
[Click to Call] stoybook updates

dmceachernmsft:
# What

Sets up storybook controls and docs for the Composite and VideoGallery changes for c2c
image
image

# Why

add doc coverage for the new configurations

https://skype.visualstudio.com/SPOOL/_workitems/edit/3285735
# How Tested

Storybook local run


packages/react-components/src/localization/locales/de-DE/strings.json
```diff
@@ -329,4 +329,4 @@
     "leftNavButtonAriaLabel": "Vorherige Seite",
     "rightNavButtonAriaLabel": "Nächste Seite"
   }
-}
\ No newline at end of file
+}
```
prabhjot-msft, line 332: are we adding any strings in this change, there are updates in string files.

prabhjot-msft (approved):

edwardlee-msft (approved):

packages/react-components/src/localization/locales/de-DE/strings.json
dmceachernmsft: we are not its one of those weird things github does sometimes for no reason. will occasionally for no reason remove the new lines at the end of all the localization files.

PR #3174, opened 2023-06-09
[Chat] Handle angle brackets in mention messages

Leah-Xia-Microsoft:
# What

Handle the use cases that < or > or /> is used in messages that contains mentions

# How Tested

In send box, type "<xyz/>" and then select a mention from the mention list.

# Process & policy checklist

### DIFF
--- a/change-beta/@azure-communication-react-20ceb770-a1a1-47e2-902d-b8cd5d4210a9.json
+++ b/change-beta/@azure-communication-react-20ceb770-a1a1-47e2-902d-b8cd5d4210a9.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Add unit tests for mention logic",
+  "packageName": "@azure/communication-react",
+  "email": "3941071+emlynmac@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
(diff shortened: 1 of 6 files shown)


change-beta/@azure-communication-react-20ceb770-a1a1-47e2-902d-b8cd5d4210a9.json
```diff
@@ -0,0 +1,7 @@
+{
```
vhuseinova-msft, line 1: Should this PR have another destination branch?

packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.ts
```diff
@@ -905,24 +906,45 @@ const parseOpenTag = (tag: string, startIdx: number): TagData => {
...
-  const tagStartIndex = text.indexOf('<', startIndex);
-  if (tagStartIndex === -1) {
-    // No more tags
-    return undefined;
+const findNextHtmlTag = (text: string, startIndex: number, tagType = MSFT_MENTION_TAG): HtmlTag | undefined => {
+  const newText = text.substring(startIndex);
+  let type: HtmlTagType = 'open';
+  let regexResult;
+
+  const openTagRegex = new RegExp(`<${tagType}(.*?)>`, 'i');
```
emlynmac, line 914: How much of a difference in performance is there between the regex and the previous implementation?

PR #3176, opened 2023-06-09
Alkwa/fix identifier pstn identifier common code

alkwa-msft:
# What
We found an issue where long phone numbers such as 13 digit phone numbers were unable to connect with the UI Library PSTN capabilities.

# Why
There was a common library change which did not require us to remove the + character as part of forming the rawId. We have updated our code to not rip off the + symbol.

# How Tested
Tested locally. Verified the isPhoneNumberIdentifier still continues to work with the + symbol. Validated my phone number worked.


packages/acs-ui-common/src/identifier.ts
```diff
@@ -24,8 +24,7 @@ export const toFlatCommunicationIdentifier = (identifier: CommunicationIdentifie
  */
 export const fromFlatCommunicationIdentifier = (id: string): CommunicationIdentifier => {
   // if the id passed is a phone number we need to build the rawId to pass in
-  const rawId = id.indexOf('+') === 0 ? '4:' + id.slice(1) : id;
-  return createIdentifierFromRawId(rawId);
+  return createIdentifierFromRawId('4:' + id);
```
dmceachernmsft, line 27: do we know that this is a valid raw ID? I thought for it to be the raw one it needed to be `4:xxxxxxxxxx`?

dmceachernmsft (approved):

mgamis-msft (approved):

packages/acs-ui-common/src/identifier.ts
alkwa-msft: I think I had to fix up some of the logic but the big change now is that they are not going to change the "id" so now we need to include the + as well.

PR #3177, opened 2023-06-09
[Chat][Accessibility] Accessibility improvement for EditBox

AmyL219:
# What

1. Resolve issue with ArrowUp and ArrowDown keys not functioning inside the edit box
2. Resolve issue with Tab key not working on small screens

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
+++ b/packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
@@ -134,6 +134,11 @@ export const ChatMessageComponentAsEditBox = (props: ChatMessageComponentAsEditB
           placeholderText={strings.editBoxPlaceholderText}
           textValue={textValue}
           onChange={setText}
+          onKeyDown={(ev) => {
+            if (ev.key === 'ArrowUp' || ev.key === 'ArrowDown') {
+              ev.stopPropagation();
+            }
+          }}
           onEnterKeyDown={() => {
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
```diff
@@ -134,6 +134,11 @@ export const ChatMessageComponentAsEditBox = (props: ChatMessageComponentAsEditB
           placeholderText={strings.editBoxPlaceholderText}
           textValue={textValue}
           onChange={setText}
+          onKeyDown={(ev) => {
```
emlynmac, line 137: Does this still enable any mention suggestion to be selected via up / down arrows?

AmyL219: It shouldn't affect any existed function inside the Edit box, but just the FluentUI Chat Message behavior. 

`stopPropagation is a method that prevents the event from being propagated or "bubbling up" through the DOM.`

emlynmac: And it was good when you tested it?

AmyL219: Tested, and looks good

vhuseinova-msft (approved):

PR #3179, opened 2023-06-09
Update side pane heading styles and correct video effects side pane width

JamesBurnside:
# What

- Increase font size and padding of the desktop side pane header
- Update close button padding to match
- Decrease the side pane width for the video effects

# Why

Match new figmas

# How Tested
Locally - see updated snapshots to see visual changes


dmceachernmsft (commented):
image
looks like there is some alignment issues here

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -347,6 +349,8 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
             </Stack.Item>
             <SidePane
               mobileView={props.mobileView}
+              /* @conditional-compile-remove(video-background-effects) */
+              maxWidth={isVideoPaneOpen ? '19.5rem' : undefined}
```
prabhjot-msft, line 353: Does the maxwidth number `19.5rem` is provided by design or we are maitaining max height from our side, just curious :)

prabhjot-msft (commented):
I have noticed a slight difference in margin from designs. In our current design we have lesser margin and also AlexP pointed in one of the bug bashes.
Screenshot 2023-06-14 at 1 48 01 PM
bug-margin
Bug: https://skype.visualstudio.com/SPOOL/_workitems/edit/3280992

As of current implementation, looks like header is shifted along with the content. Maybe we can check with AlexP
Screenshot 2023-06-14 at 1 49 15 PM

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
JamesBurnside: Yeah, should've been 17.5 though my bad! Updated

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/VideoEffects.test.ts-snapshots/custom-video-effects-config-screen-pane-none-selected-Desktop-Chrome-linux.png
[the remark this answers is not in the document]
JamesBurnside: Yeah figma has a tiny offset:
image

dmceachernmsft (approved):

JamesBurnside:
> I have noticed a slight difference in margin from designs. In our current design we have lesser margin and also AlexP pointed in one of the bug bashes.
> As of current implementation, looks like header is shifted along with the content. Maybe we can check with AlexP 

Updated to have the tighter margins

prabhjot-msft (approved):

PR #3182, opened 2023-06-12
Alkwa/add caret dependencies

alkwa-msft:
# What
Updating all @azure dependencies to use ^ so we will automatically take in minor/patch updates beyond the version specified in our package.json files

# Why
We had inconsistent setup of ~ and ^ in our package.json and with the Azure guidelines it should be safe to automatically take both minor and patch updates with the no breaking change semver agreement.

# How Tested
Built locally with both flavors


packages/calling-stateful-client/package.json
```diff
@@ -30,17 +30,17 @@
...
     "immer": "9.0.6"
   },
   "peerDependencies": {
-    "@azure/communication-calling": "1.13.2-beta.1 || >=1.13.1"
+    "@azure/communication-calling": "1.14.1-beta.1 || >=1.13.1"
   },
   "devDependencies": {
-    "@azure/communication-calling": "1.13.2-beta.1 || >=1.13.1",
+    "@azure/communication-calling": "1.14.1-beta.1 || >=1.13.1",
     "@azure/core-auth": "1.3.2",
```
jimchou-dev, line 44: Is this one missed or on purpose

packages/communication-react/package.json
```diff
@@ -89,7 +89,7 @@
     "_by-flavor": "rushx _current-flavor && env-cmd -f ../../common/config/env/.env --use-shell"
   },
   "devDependencies": {
-    "@azure/communication-calling": "1.13.2-beta.1 || >=1.13.1",
+    "@azure/communication-calling": "1.14.1-beta.1 || >=1.13.1",
     "@azure/communication-calling-effects": "1.0.0-beta.2",
     "@azure/communication-chat": "1.3.2-beta.2 || >=1.2.0",
     "@azure/core-auth": "1.3.2",
```
jimchou-dev, line 95: Is this one missed or on purpose

mgamis-msft (approved):

packages/communication-react/package.json
alkwa-msft: which "one" are you referring to?  If its about adding a caret, because its a beta version and doesn't follow semver you cant use a caret on it.

alkwa-msft: updated to include @azure/core-auth ^1.4.0

https://www.npmjs.com/package/@azure/core-auth?activeTab=versions

jimchou-dev (approved):

PR #3183, opened 2023-06-12
Alkwa/log device enumeration

alkwa-msft:
# What
We have had customers asking about why it takes so long to have the "start call" button to light up on the configuration screen. In local testing we haven't seen anything close to 30 seconds being raised by customers. (I am seeing 1.8s). My hypothesis is that device enumeration is client-side dependent (devices + cpu).

# Why
Before we can try to apply any larger fix we need to understand where the perf hit is coming from. In the same timeframe if I am seeing 1.8s and someone else is seeing 30s then we can determine that the difference is the time it takes for that end user to enumerate through their own devices.

# How Tested
Tested locally with the correct azure logger setting so we could see the logs of how long it takes to enumerate through devices and ask for permission


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -332,6 +334,7 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
     this.locator = locator;
     this.deviceManager = deviceManager;
     const isTeamsMeeting = 'meetingLink' in this.locator;
+    this._logger = createClientLogger('communication-react:composite');
```
alkwa-msft, line 337: are we open to adding AzureLogger to the composite packlet and using it in our composites?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -483,26 +486,41 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
   }
 
   public async queryCameras(): Promise<VideoDeviceInfo[]> {
+    const startTime = new Date().getTime();
     return await this.asyncTeeErrorToEventEmitter(async () => {
-      return this.deviceManager.getCameras();
+      const cameras = await this.deviceManager.getCameras();
```
alkwa-msft, line 491: the enumeration itself takes time so we have to break apart the return from the computation of the API call so we can measure before we return

edwardlee-msft (approved):

PR #3190, opened 2023-06-12
Alkwa/log device enumeration2

alkwa-msft:
# What
We have had customers asking about why it takes so long to have the "start call" button to light up on the configuration screen. In local testing we haven't seen anything close to 30 seconds being raised by customers. (I am seeing 1.8s). My hypothesis is that device enumeration is client-side dependent (devices + cpu).

# Why
Before we can try to apply any larger fix we need to understand where the perf hit is coming from. In the same timeframe if I am seeing 1.8s and someone else is seeing 30s then we can determine that the difference is the time it takes for that end user to enumerate through their own devices.

# How Tested
Tested locally with the correct azure logger setting so we could see the logs of how long it takes to enumerate through devices and ask for permission


edwardlee-msft (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -302,6 +303,7 @@ export type AdapterStateModifier = (state: CallAdapterState) => CallAdapterState
 export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTeamsCallAgent = CallAgent>
   implements CommonCallAdapter
 {
+  private _logger: AzureLogger;
```
PorterNan, line 306: This should be working fine, but "communication-react:composite" sounds like a logger running all across the whole composite package, so it would be nice to have a separated file then we are able to import it in any file we want in this repo.
Here is a good example for chat stateful client
/workspaces/communication-ui-library/packages/chat-stateful-client/src/Logger.ts

PorterNan (approved):

PR #3196, opened 2023-06-13
add raise hand component with conditional build

rzdor:
# What

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
+++ b/packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
@@ -51,6 +51,14 @@ export interface CommonCallingHandlers {
   onStopScreenShare: () => Promise<void>;
   onToggleScreenShare: () => Promise<void>;
   onHangUp: (forEveryone?: boolean) => Promise<void>;
+  /* @conditional-compile-remove(raise-hands) */
+  onRaiseHand: () => Promise<void>;
+  /* @conditional-compile-remove(raise-hands) */
+  onLowerHand: () => Promise<void>;
+  /* @conditional-compile-remove(raise-hands) */
+  onLowerHands: (userId: string[]) => Promise<void>;
+  /* @conditional-compile-remove(raise-hands) */
+  onToggleRaiseHand: () => Promise<void>;
   /* @conditional-compile-remove(PSTN-calls) */
   onToggleHold: () => Promise<void>;
   /* @conditional-compile-remove(PSTN-calls) */
@@ -262,6 +270,36 @@ export const createDefaultCommonCallingHandlers = memoizeOne(
     const onToggleScreenShare = async (): Promise<void> =>
       call?.isScreenSharingOn ? await onStopScreenShare() : await onStartScreenShare();
 
+    /* @conditional-compile-remove(raise-hands) */
+    const onRaiseHand = async (): Promise<void> => await call?.feature(Features.RaiseHand)?.raiseHand();
+
+    /* @conditional-compile-remove(raise-hands) */
+    const onLowerHand = async (): Promise<void> => await call?.feature(Features.RaiseHand)?.lowerHand();
+
+    /* @conditional-compile-remove(raise-hands) */
+    const onLowerHands = async (userIds: string[]): Promise<void> => {
+      if (userIds.length > 0) {
+        await call?.feature(Features.RaiseHand)?.lowerHands([]);
+      }
+    };
+
+    /* @conditional-compile-remove(raise-hands) */
+    const onToggleRaiseHand = async (): Promise<void> => {
+      const raiseHandFeature = call?.feature(Features.RaiseHand);
+      const localUserId = callClient.getState().userId;
+      const isLocalRaisedHand = raiseHandFeature
+        ?.getRaisedHands()
+        .find(
+          (publishedState) =>
+            toFlatCommunicationIdentifier(publishedState.identifier) === toFlatCommunicationIdentifier(localUserId)
+        );
+      if (isLocalRaisedHand) {
+        await raiseHandFeature?.lowerHand();
+      } else {
+        await raiseHandFeature?.raiseHand();
+      }
+    };
+
     const onHangUp = async (forEveryone?: boolean): Promise<void> =>
       await call?.hangUp({ forEveryone: forEveryone === true ? true : false });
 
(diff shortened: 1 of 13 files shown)


packages/calling-component-bindings/src/hooks/usePropsFor.ts
```diff
@@ -99,6 +103,8 @@ export type GetSelector<Component extends (props: any) => JSX.Element | undefine
   ? CameraButtonSelector
   : AreEqual<Component, typeof ScreenShareButton> extends true
   ? ScreenShareButtonSelector
+  : AreEqual<Component, typeof RaiseHandButton> extends true
```
PorterNan, line 106: Is this button private or will be exposed as a public component?

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
```diff
@@ -51,6 +51,14 @@ export interface CommonCallingHandlers {
   onStopScreenShare: () => Promise<void>;
   onToggleScreenShare: () => Promise<void>;
   onHangUp: (forEveryone?: boolean) => Promise<void>;
+  /* @conditional-compile-remove(raise-hands) */
+  onRaiseHand: () => Promise<void>;
+  /* @conditional-compile-remove(raise-hands) */
+  onLowerHand: () => Promise<void>;
+  /* @conditional-compile-remove(raise-hands) */
+  onLowerHands: (userId: string[]) => Promise<void>;
+  /* @conditional-compile-remove(raise-hands) */
```
PorterNan, line 60: Is that is onToggle style then we just have that without onRaise/onLower

or we just use two function styles ( onRaise/onLower) without onToggle

(just choose one style of those)

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
```diff
@@ -262,6 +270,36 @@ export const createDefaultCommonCallingHandlers = memoizeOne(
     const onToggleScreenShare = async (): Promise<void> =>
       call?.isScreenSharingOn ? await onStopScreenShare() : await onStartScreenShare();
 
+    /* @conditional-compile-remove(raise-hands) */
+    const onRaiseHand = async (): Promise<void> => await call?.feature(Features.RaiseHand)?.raiseHand();
```
PorterNan, line 274: Some indent issues here, would be nice to run rush lint: fix to fix all of these

PorterNan: NVM, looks like the sentence wrap problem in my local editor

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
```diff
@@ -462,6 +500,14 @@ export const createDefaultCommonCallingHandlers = memoizeOne(
       onStartLocalVideo,
       onDisposeRemoteStreamView,
       onDisposeLocalStreamView,
+      /* @conditional-compile-remove(raise-hands) */
+      onRaiseHand,
+      /* @conditional-compile-remove(raise-hands) */
+      onLowerHand,
+      /* @conditional-compile-remove(raise-hands) */
+      onLowerHands,
+      /* @conditional-compile-remove(raise-hands) */
```
PorterNan, line 509: Again, either onToggle or (on/off) style is enough, duplicated api is not needed

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -57,6 +61,8 @@ const memoizedAllConvertRemoteParticipant = memoizeFnAll(
     userId: string,
     isMuted: boolean,
     isSpeaking: boolean,
+    /* @conditional-compile-remove(raise-hands) */
+    raisedHand: RaisedHand | undefined,
```
PorterNan, line 65:
```suggestion
    raisedHand?: RaisedHand,
```

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -77,6 +85,8 @@ export const convertRemoteParticipantToVideoGalleryRemoteParticipant = (
   userId: string,
   isMuted: boolean,
   isSpeaking: boolean,
+  /* @conditional-compile-remove(raise-hands) */
+  raisedHand: RaisedHand | undefined,
```
PorterNan, line 89:
```suggestion
  raisedHand?: RaisedHand,
```

We prefer ?: style instead of undefined

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -61,6 +61,12 @@ export interface ParticipantItemStrings {
   menuTitle: string;
   /** Label for the remove button in participant menu  */
   removeButtonLabel: string;
+  /* @conditional-compile-remove(raise-hands) */
+  /** Label for the raised hand icon in participant state stack  */
+  raisedHandIconLabel: string;
```
PorterNan, line 66: Make this 2 strings optional

packages/react-components/src/components/ParticipantsButton.tsx
```diff
@@ -121,6 +121,11 @@ export interface ParticipantsButtonProps extends ControlBarButtonProps {
    * Callback to render the context menu for each participant
    */
   onRemoveParticipant?: (userId: string) => void;
+  /* @conditional-compile-remove(raise-hands) */
+  /**
+   * Callback to render the context menu for each participant
+   */
+  onLowerHands?: (userId: string[]) => void;
```
PorterNan, line 128: I feel like this is just for lowering one participant hand, can we name it as onLowerHand? And make a handler to match this funciton name

PorterNan:
https://azure.github.io/communication-ui-library/iframe.html?viewMode=docs&id=statefulclient-reacthooks-usepropsfor--page#usepropsfor-calling-app-example

Can we test all of these apis for selector and usePropsFor working well based on this code snippet? We don't currently have too much code to test these apis, so would be nice if you could test them locally

PorterNan:
> https://azure.github.io/communication-ui-library/iframe.html?viewMode=docs&id=statefulclient-reacthooks-usepropsfor--page#usepropsfor-calling-app-example
> 
> Can we test all of these apis for selector and usePropsFor working well based on this code snippet? We don't currently have too much code to test these apis, so would be nice if you could test them locally

https://github.com/Azure/communication-ui-library/tree/main/samples/ComponentExamples
And this is the sample you might wanna take a look if you wanna test this locally

PR #3197, opened 2023-06-13
[Chat] Fix for a Storybook issue when messages can't be sent in Message thread

vhuseinova-msft:
# What

- Fix for endless load of past messages issue in storybook
- Fix for flickering issue for avatars

# Why

Fix issues for MessageThread page in storybook

# How Tested

Storybook

# Process & policy checklist


packages/storybook/stories/MessageThread/MessageThread.stories.tsx
```diff
@@ -298,7 +299,10 @@ const MessageThreadStory = (args): JSX.Element => {
 
   const onLoadPreviousMessages = async (): Promise<boolean> => {
```
emlynmac, line 300: Can this be called multiple times before we have all the messages retrieved?
I thought I recall that the messages can come in pages and that the promise should return a falsey value to decide if more are needed?

AmyL219:
Uhh, the storybook from this [PR](https://60c7ae6891f0e90039d7cd54-gypngdevyk.chromatic.com/) still has the flashing avatar issue

https://github.com/Azure/communication-ui-library/assets/75647512/ea05a25e-2a85-4725-8911-a0703249cab0

packages/storybook/stories/MessageThread/MessageThread.stories.tsx
emlynmac: Ah yes, this is the storybook. Nm!

emlynmac (approved):

emlynmac:
> Uhh, the storybook from this [PR](https://60c7ae6891f0e90039d7cd54-gypngdevyk.chromatic.com/) still has the flashing avatar issue
> 
>  Screen.Recording.2023-06-13.at.1.51.09.PM.mov

Yup, the flickery icons seems to be there for the first two messages.

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -986,7 +986,13 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
         }
       }
     }
-  }, [numberOfChatMessagesToReload, onLoadPreviousChatMessages]);
+  }, [
+    numberOfChatMessagesToReload,
+    onLoadPreviousChatMessages,
+    isLoadingChatMessagesRef,
```
JamesBurnside, line 992: Just a note that adding these Refs to the use* dependency array shouldn't impact any bug, the ref from `useRef` cannot change (so its fine to include it but it should be the same behavior as not including it)

vhuseinova-msft: You're right, reverted these changes and pushed the update for the PR

emlynmac (approved):

AmyL219 (approved):

PR #3203, opened 2023-06-14
Alkwa/fix captions error

alkwa-msft:
# What
We are seeing a captions error in the console when we are trying to end a call. This is odd as captions hasn't even been started and I am not in an interop call. I thought about using a better check from a condition in the stateful client but I ended up just mirroring the same condition we use when the captions are "subscribed"

# Why
We are trying to unregister from captions when we haven't even registered for captions in an interop call

# How Tested
Tested locally in an acs group call. Validated that with the unsubscribe condition we are not seeing the error anymore.
Validated with teams meeting interop. No error
Validate with teams meeting as a Teams user. No Error.

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
+++ b/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
@@ -891,6 +891,17 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
     }
   }
 
+  /* @conditional-compile-remove(close-captions) */
+  private unsubscribeFromCaptionEvents(): void {
+    if (this.call && this.call.state === 'Connected' && _isTeamsMeetingCall(this.call)) {
+      this._call?.feature(Features.TeamsCaptions).off('captionsReceived', this.captionsReceived.bind(this));
+      this._call
+        ?.feature(Features.TeamsCaptions)
+        .off('isCaptionsActiveChanged', this.isCaptionsActiveChanged.bind(this));
+      this.call?.off('stateChanged', this.subscribeToCaptionEvents.bind(this));
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


carocao-msft (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -891,6 +891,17 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
...
   }
 
+  /* @conditional-compile-remove(close-captions) */
+  private unsubscribeFromCaptionEvents(): void {
+    if (this.call && this.call.state === 'Connected' && _isTeamsMeetingCall(this.call)) {
+      this._call?.feature(Features.TeamsCaptions).off('captionsReceived', this.captionsReceived.bind(this));
+      this._call
+        ?.feature(Features.TeamsCaptions)
+        .off('isCaptionsActiveChanged', this.isCaptionsActiveChanged.bind(this));
+      this.call?.off('stateChanged', this.subscribeToCaptionEvents.bind(this));
```
PorterNan, line 901: This line already be executed at line 890, but I guess no harm to do this again

PorterNan (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
alkwa-msft: I will write a small task perhaps we can get the fix when we do the captions GA work.

alkwa-msft: https://skype.visualstudio.com/SPOOL/_workitems/edit/3306211

PR #3205, opened 2023-06-14
[Chat] Handle angle bracket with mention message

Leah-Xia-Microsoft:
# What

Workaround for angle brackets in mention messages.

# Why

When user type in "<x @Mentioned User" into the input box, it should be displayed properly

# Process & policy checklist


packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.ts
```diff
@@ -877,6 +884,12 @@ const findNextHtmlTag = (text: string, startIndex: number): HtmlTag | undefined
     // No close tag
     return undefined;
   }
+
+  const textBeforeTagStart = text.substring(0, tagStartIndex + 1);
+  const skippedAngleBrackets = textBeforeTagStart?.match(/&lt;/g);
```
emlynmac, line 889: We might need to keep a list of user-updated indices of angle brackets to handle this properly.

PR #3221, opened 2023-06-16
Alkwa/move util func to common

alkwa-msft:
# What
Moving a common util function from the composites directory to the acs-ui-common packlet.

# Why
There was an optimization suggested in the discussion below to reduce the bundle package size for those only building for a  chat experience
https://github.com/Azure/communication-ui-library/discussions/3208


dmceachernmsft (approved):

packages/acs-ui-common/review/beta/acs-ui-common.api.md
```diff
@@ -51,6 +51,9 @@ export interface _IObjectMap<T> {
     [key: string]: T;
 }
 
+// @public
```
JamesBurnside, line 54: this should be `@internal`

alkwa-msft: will make this change

JamesBurnside (approved):

PorterNan (approved):
https://github.com/Azure/communication-ui-library/pull/3225
There is a PR opened for this to get a better production-like treeshaked bundle - so we can see the real number change from the comment once it is in

JamesBurnside:
Wow! Yeah lets check this in!
image

PR #3222, opened 2023-06-16
[Bug fixes] Local video tile screensharing bug fixes

dmceachernmsft:
# What

image

add local tile to over flow gallery in default layout
image
image

render local tile when screen sharing alone in a call
# Why

Allows for the local tile to be visible when:
- gallery is in default layout
- user is screensharing in a call alone when preparing for a meeting / presentation.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3277628
https://skype.visualstudio.com/SPOOL/_workitems/edit/3135146
# How Tested

Storybook validation


packages/react-components/src/components/VideoGallery/OverflowGallery.tsx
```diff
@@ -96,7 +96,7 @@ export const OverflowGallery = (props: {
         onFetchTilesToRender={onFetchTilesToRender}
         onChildrenPerPageChange={onChildrenPerPageChange}
       >
-        {overflowGalleryElements}
+        {overflowGalleryElements ? overflowGalleryElements : [<></>]}
```
mgamis-msft, line 99: Are you doing this so that an empty VerticalGallery will be rendered and push down the local video tile?

packages/react-components/src/components/VideoGallery/OverflowGallery.tsx
```diff
@@ -108,7 +108,7 @@ export const OverflowGallery = (props: {
     onChildrenPerPageChange?.(3);
     return (
       <ScrollableHorizontalGallery
-        horizontalGalleryElements={overflowGalleryElements}
+        horizontalGalleryElements={overflowGalleryElements ? overflowGalleryElements : [<></>]}
```
mgamis-msft, line 111: Same as above?

packages/react-components/src/components/HorizontalGallery.tsx
```diff
@@ -76,9 +76,12 @@ export const HorizontalGallery = (props: HorizontalGalleryProps): JSX.Element =>
   const clippedPage = firstIndexOfCurrentPage < numberOfChildren - 1 ? page : lastPage;
 
   const childrenOnCurrentPage = useMemo(() => {
-    return indexesArray[clippedPage].map((index) => {
-      return React.Children.toArray(children)[index];
-    });
+    if (indexesArray[0] !== undefined) {
```
mgamis-msft, line 79: I'm curious what situation would the first index be undefined

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
```diff
@@ -96,8 +98,19 @@ const _useOrganizedParticipants = (props: OrganizedParticipantsArgs): OrganizedP
...
+  )[] => {
+    if (isScreenShareActive && localParticipant) {
+      const localParticipantPlusOverflow = [localParticipant].concat(
+        visibleGridParticipants.current.concat(visibleOverflowGalleryParticipants.current)
+      );
+      // If screen sharing is active, assign video and audio participants as overflow gallery participants
+      /* @conditional-compile-remove(PSTN-calls) */ /* @conditional-compile-remove(one-to-n-calling) */
+      return localParticipantPlusOverflow.concat(callingParticipants);
+      return localParticipantPlusOverflow;
+    } else if (isScreenShareActive) {
```
mgamis-msft, line 113: I don't want to make the ordering and organizing of remote participants more complicated. Is it possible to prepend the local video tile to the return overflow gallery participants of this hook?

packages/react-components/src/components/HorizontalGallery.tsx
dmceachernmsft: When there is no overflow gallery tiles this will be undefined and will throw an error so we only want to be able to call the map if there is a tile here and the overflow is present

packages/react-components/src/components/VideoGallery/OverflowGallery.tsx
dmceachernmsft: yup! just more protections against undefined children in the gallery. this was prone to blow up if I didn't add these

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
dmceachernmsft: I found it easier to do the mapping in here than in the result of this hook since it is also returning the grid organization too

packages/react-components/src/components/VerticalGallery.tsx
```diff
@@ -118,9 +118,12 @@ export const VerticalGallery = (props: VerticalGalleryProps): JSX.Element => {
   const firstIndexOfCurrentPage = (page - 1) * childrenPerPage;
   const clippedPage = firstIndexOfCurrentPage < numberOfChildren - 1 ? page : lastPage;
   const childrenOnCurrentPage = useMemo(() => {
-    return indexesArray[clippedPage - 1].map((index) => {
-      return React.Children.toArray(children)[index];
-    });
+    if (indexesArray[0] !== undefined) {
```
prabhjot-msft, line 121: Should we check length of index array rather than first element being undefined

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -400,7 +400,13 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     const initialsName = !localParticipant.displayName ? '' : localParticipant.displayName;
 
     return (
-      <Stack key="local-video-tile-key" tabIndex={0} aria-label={strings.localVideoMovementLabel} role={'dialog'}>
+      <Stack
+        style={{ width: '100%', height: '100%' }}
```
prabhjot-msft, line 404: Should we define these styles in a constant or css file?

packages/react-components/src/components/VerticalGallery.tsx
dmceachernmsft: Because this is a doubly nested array it will always have a truthy reference so we need to check to see if the first index is defined to do that undefined check.

packages/react-components/src/components/VideoGallery.tsx
dmceachernmsft: yup I can find a place for these!

prabhjot-msft (approved):

mgamis-msft (approved):

PR #3223, opened 2023-06-16
[Click to Call] ResizeObserver fix

dmceachernmsft:
# What

Makes video tile not update on the resizeObserver reference updates
# Why

mitigates the resize error loop while still using a useLayoutEffect

# How Tested

tested in click to call JS quick start


mgamis-msft (approved):

JamesBurnside (approved):
No harm in checking this in (cleaner code regardless) so approving! But I think this can't be the cause of the bug 🤔🤔

PR #3225, opened 2023-06-16
Get treeshaked build in production mode

PorterNan:
# What

Get treeshaked build in production mode

# Why

The current webpack mapping supports hot reload really well for daily development, but it blocks webpack treeshaking. So we conditionally choose the mapping when it is in development mode for getting more production-like bundles

# How Tested

# Process & policy checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -455,6 +455,8 @@ jobs:
       - name: Run dependency update for stable
         if: ${{ matrix.flavor == 'stable' }}
         run: rush update:stable
+      - name: Build dependencies
+        run: rush build -t @azure/communication-react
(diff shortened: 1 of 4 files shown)


PorterNan (commented):
This result is interesting... For call and callWithChat... looks like we don't get a lot better after treeshaking enabled, but chat surely does

JamesBurnside (approved):

JamesBurnside:
> This result is interesting... For call and callWithChat... looks like we don't get a lot better after treeshaking enabled, but chat surely does

I guess that makes sense. CallWithChat uses everything, whereas now we have chat without calling and calling without chat

alkwa-msft (approved):

PR #3226, opened 2023-06-17
[Chat] Handle mention messages with angle brackets and ampersand

Leah-Xia-Microsoft:
# What

When user enter < > or &, we should ignore them in the textToTagParser but still display them as is.

# Why

User input x<y & y>z might break the parse if we don't ignore <, > or &

# How Tested

# Process & policy checklist


change-beta/@azure-communication-react-17247566-1bb6-4c63-bf6d-4bc8941ace2f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
vhuseinova-msft, line 2: Could you update it to be 'none'?

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -159,8 +159,7 @@ export const TextFieldWithMention = (props: TextFieldWithMentionProps): JSX.Elem
       const mention = htmlStringForMentionSuggestion(suggestion, localeStrings);
 
       // update plain text with the mention html text
-      const newPlainText =
-        inputTextValue.substring(0, currentTriggerStartIndex) + mention + inputTextValue.substring(selectionEnd);
+      const newPlainText = inputTextValue.substring(0, currentTriggerStartIndex);
```
vhuseinova-msft, line 162: It looks like `inputTextValue.substring(selectionEnd);` will be lost in case if mention is added in the middle of the text

packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.ts
```diff
@@ -342,13 +347,18 @@ export const updateHTML = (props: UpdateHTMLProps): { updatedHTML: string; updat
     if (startIndex <= tag.plainTextBeginIndex) {
       // change start is before the open tag
       // Math.max(lastProcessedPlainTextTagEndIndex, startIndex) is used as startIndex may not be in [[previous tag].plainTextEndIndex - tag.plainTextBeginIndex] range
-      const startChangeDiff = tag.plainTextBeginIndex - Math.max(lastProcessedPlainTextTagEndIndex, startIndex);
+      const startChangeDiff = getChangeDiff(
```
vhuseinova-msft, line 350: Is it possible to use a separate parameter (like `mentionTagLength`), calculate it once for all subtags and reuse instead of calculating it for every tag?

packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.ts
```diff
@@ -376,29 +386,31 @@ export const updateHTML = (props: UpdateHTMLProps): { updatedHTML: string; updat
             if (startIndex !== tag.plainTextBeginIndex && startIndex !== closingTagInfo.plainTextEndIndex) {
               // mention tag should be deleted when user tries to edit it in the middle
               result += htmlText.substring(lastProcessedHTMLIndex, tag.openTagIndex) + processedChange;
-              changeNewEndIndex = tag.plainTextBeginIndex + processedChange.length;
+              changeNewEndIndex = tag.plainTextBeginIndex + revertEscapedHtmlCharacters(processedChange).length;
```
vhuseinova-msft, line 389: The property from[ this comment](https://github.com/Azure/communication-ui-library/pull/3226/files#r1234236489) can probably be used for `revertEscapedHtmlCharacters` call too

packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.ts
```diff
@@ -923,3 +936,31 @@ const addTag = (tag: TagData, parseStack: TagData[], tags: TagData[]): void => {
...
+const getLengthDiffForEscapedHtmlChars = (text: string, startIndex: number, endIndex: number): number => {
+  const subText = text.substring(startIndex, endIndex);
+  const escapedAngleBracketCount = subText.match(/(<)|(>)/g)?.length || 0;
+  let escapedCharLength = escapedAngleBracketCount * ('&lt;'.length - '<'.length);
+  const escapedAmpersandCount = subText.match(/&/g)?.length || 0;
+  escapedCharLength += escapedAmpersandCount * ('&amp;'.length - '&'.length);
+  return escapedCharLength;
+};
+
+const revertEscapedHtmlCharacters = (text: string): string => {
```
emlynmac, line 964: `unEscapeHtmlCharacters` ?

packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.ts
Leah-Xia-Microsoft: Talked to Viktoriia async. We'll keep this as is since the calculation is depends on different startIndex and endIndex. Also, for each tag, we're calculation different portion of the string.

packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.ts
```diff
@@ -305,11 +304,15 @@ type UpdateHTMLProps = {
...
-    props;
+  const { htmlText, oldPlainText, tags, startIndex, oldPlainTextEndIndex, change, mentionTrigger } = props;
   if (tags.length === 0 || (startIndex === 0 && oldPlainTextEndIndex === oldPlainText.length - 1)) {
     // no tags added yet or the whole text is changed
-    return { updatedHTML: newPlainText, updatedSelectionIndex: undefined };
+    const changeWithSkippedChars = escapeHTMLChars(change);
+    const updatedHTML =
+      escapeHTMLChars(oldPlainText.substring(0, startIndex)) +
+      changeWithSkippedChars +
+      oldPlainText.substring(oldPlainTextEndIndex);
```
vhuseinova-msft, line 314: We should use escaped version for `substring(oldPlainTextEndIndex)` too or use substrings from `htmlText` instead
Otherwise, in case of `hi < [change] <3` where [change] is `change` value, the second `<` won't be escaped

packages/react-components/src/components/TextFieldWithMention/mentionTagUtils.ts
```diff
@@ -376,29 +385,30 @@ export const updateHTML = (props: UpdateHTMLProps): { updatedHTML: string; updat
...
             processedChange = '';
           } else {
+            // When change is empty, it means that this change is a deletion
             const updateMentionTagResult = handleMentionTagUpdate({
               htmlText,
               oldPlainText,
               lastProcessedHTMLIndex,
               processedChange,
-              change,
+              change: change,
```
vhuseinova-msft, line 411: minor: parameter name is not required here as parameter and value names are the same

vhuseinova-msft (approved):

emlynmac (approved):
LGTM

PR #3228, opened 2023-06-19
Fix bug where user ends up in 2 calls when failed transfer request is accepted

mgamis-msft:
# What
Switch the call when the accepted transfer call is connected and hang up the old call.

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3305535

# How Tested
Tested in local sample:
Note: video is 2mins long because I wait for the transferor's transfer request to timeout before accepting it
https://github.com/Azure/communication-ui-library/assets/79475487/5ed5756d-48b1-4092-8181-6e9d25c0c899

Updated video testing 4 scenarios
1. Accepting a transfer request that has not timed out
2. Rejecting a transfer request that has timed out
3. Accepting a transfer request that has timed out but transfer target declines call
4. Accepting a transfer request that has timed out and transfer target accepts call

https://github.com/Azure/communication-ui-library/assets/79475487/589c30c3-4c7f-457e-a77a-d8ed2cec3f46

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -370,6 +370,31 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
       }
 
       this.context.updateClientState(clientState);
+
+      /* @conditional-compile-remove(call-transfer) */
+      const acceptedTransferCallState = this.context.getState().acceptedTransferCallState;
+
+      /* @conditional-compile-remove(call-transfer) */
+      // If there has been an accepted transfer call that is now in the connected state we should ensure we leave the current call.
```
JamesBurnside, line 378: What does the SDK do to the current call? Is it placed on hold or do they leave the user also connected? (this feels like something that should be handled at the SDK layer)

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -370,6 +370,31 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
...
+        acceptedTransferCallState &&
+        acceptedTransferCallState.state === 'Connected' &&
+        acceptedTransferCallState.id !== this.call?.id
+      ) {
+        const cAgent = callAgent as CallAgent;
+        const transferCall = cAgent.calls.find((call) => (call as Call).id === acceptedTransferCallState.id);
+        if (transferCall) {
+          const currentCall = this.call;
+          this.processNewCall(transferCall);
+          // Wait 2 seconds and then ensure we hang up the old call if it is still connected
```
JamesBurnside, line 390: Why do we need to wait 2 seconds and not hang up immediately?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -370,6 +370,31 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
       }
 
       this.context.updateClientState(clientState);
+
+      /* @conditional-compile-remove(call-transfer) */
+      const acceptedTransferCallState = this.context.getState().acceptedTransferCallState;
+
+      /* @conditional-compile-remove(call-transfer) */
+      // If there has been an accepted transfer call that is now in the connected state we should ensure we leave the current call.
+      // There is a possibility we accepted a transfer which the transferor thinks failed.
```
JamesBurnside, line 379: This seems like it shouldn't be possible, does the SDK not throw an error/return a failure if the `call.accept()` is called after the timeout?

mgamis-msft: No it does not. I will ask during office hours.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
mgamis-msft: They leave the user connected. So there is actually 2 calls with a connected state.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
mgamis-msft: If I hang up right away, the transferring Teams user thinks the transfer fails. See short video. 
https://github.com/Azure/communication-ui-library/assets/79475487/0d6bc717-2d95-4aad-85c3-2d51a10f4ef4

I am essentially allowing the transferring Teams user to end the call and recognize that the transfer is complete. I will update the code comment.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -370,6 +370,32 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
...
+        acceptedTransferCallState &&
+        acceptedTransferCallState.state === 'Connected' &&
+        acceptedTransferCallState.id !== this.call?.id
+      ) {
+        const cAgent = callAgent as CallAgent;
+        const transferCall = cAgent.calls.find((call) => (call as Call).id === acceptedTransferCallState.id);
+        if (transferCall) {
+          const oldCall = this.call;
+          this.processNewCall(transferCall);
+          // Wait 2 seconds to allow transferor to recognize the transfer is complete and end the call. But if the old
```
dmceachernmsft, line 390: is there not another way we can wait for the state of the old call to change? if there is an issue with the transfer does the state not go to `none` or something?

mgamis-msft: Unfortunately no the old call state is 'Connected' during the transfer and still 'Connected' when it fails.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
JamesBurnside: Can you confirm with calling if this is expected? I struggle to see the scenario where you want to be active in two separate calls at once. If it is expected (to be in `Connected` with two calls at once), can calling SDK provide a snippet on how developers are expected to handle this? e.g. Should the developer manually go `on hold` in one call?

mgamis-msft: So Teams user perform a safe transfer which is "wait for transfer completion to end if success and continue to original call if failure". So on the latter when the transfer request times out, the original call is still connected. The ACS user accepts the transfer and essentially makes an outgoing call to the transfer target.
There is no mention of holding as transferee from the documentation
image
From https://learn.microsoft.com/en-us/azure/communication-services/how-tos/calling-sdk/transfer-calls.
I think even if I hold the original call, I would still have a similar problem because we have 1 call in hold state and the other in connected state.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -370,6 +370,34 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
...
+          const oldCall = this.call;
+          this.processNewCall(transferCall);
+          // Do not hang up right away because we want the Transfer API to remove the call on a successful transfer. If
+          // we hang up right away the Transfer API will recognize the transfer as a failure even though we are now in
+          // the transfer call
+          setTimeout(() => {
+            if (oldCall?.state === 'Connected') {
+              oldCall.hangUp();
+            }
+          }, 2000);
```
JamesBurnside, line 398: nit: can you add a comment why two seconds even just somethig like `Arbitrary wait time before hanging up. Time derived from manual testing.` just to highlight that 2 seconds was a number we came up with

JamesBurnside (approved):

dmceachernmsft (approved):

mgamis-msft:
Verified this fix works in a bug bash

PR #3229, opened 2023-06-19
Enable Node 18 Tooling support

emlynmac:
# What
This change enables compilation and usage under Node 18 and rolls the CI system onto node 18 too.

# Why
Node 18 is the current LTS version and node 16 will be EoL in Sept 2023

# How Tested
Build / storybook

# Process & policy checklist


packages/storybook/scripts/build.mjs
```diff
@@ -10,7 +10,9 @@ async function main() {
     console.log('Skipping storybook build for stable build');
     return;
   }
-  await exec(quote(['npx', 'build-storybook', '--quiet', '--loglevel', 'warn', ...process.argv.slice(2)]));
+  await exec(quote(['npx', 'build-storybook', '--quiet', '--loglevel', 'warn', ...process.argv.slice(2)]), {
+    NODE_OPTIONS: '--openssl-legacy-provider'
```
JamesBurnside, line 14: can you add comments why we have the NODE_OPTIONS here (i.e. to support node 18 and should be removed if we update to storybook 7)

JamesBurnside (approved):
I think we should go with this, given that our production site is just bundled JS I assume this node options only impacts local development?

packages/storybook/scripts/build.mjs
```diff
@@ -10,7 +10,9 @@ async function main() {
     console.log('Skipping storybook build for stable build');
     return;
   }
-  await exec(quote(['npx', 'build-storybook', '--quiet', '--loglevel', 'warn', ...process.argv.slice(2)]));
+  await exec(quote(['npx', 'build-storybook', '--quiet', '--loglevel', 'warn', ...process.argv.slice(2)]), {
+    NODE_OPTIONS: '--openssl-legacy-provider' // Storybook needs this for use with node 18; resolved in storybook v7
```
JamesBurnside, line 14: Will this break if someone builds storybook locally with node 16?

emlynmac: Yes, I think so.

emlynmac: Yes, this will break for node <18.

JamesBurnside: Hmm, instead of passing the option, can we have this script set an environment variable?
https://stackoverflow.com/a/76178571/3047190

JamesBurnside: Or have the script check the node version and only set this if node 18+

emlynmac: Ha, I just implemented this!

JamesBurnside (approved):

jimchou-dev (approved):
LGTM

rush.json
```diff
@@ -6,7 +6,7 @@
     // Setting this to false for the react northstar-wrapper package
     "strictPeerDependencies": false
   },
-  "nodeSupportedVersionRange": ">=12.13.0 <13.0.0 || >=14.15.0 <15.0.0 || >=16.19.0 <17.0.0",
+  "nodeSupportedVersionRange": ">=12.13.0 <13.0.0 || >=14.15.0 <15.0.0 || >=16.19.0 <19.0.0",
```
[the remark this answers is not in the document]
emlynmac: Technically anything less that 16 is EoL. 
This is only for the rush tool.

palatter (approved):

vhuseinova-msft (approved):

PR #3233, opened 2023-06-20
[Chat] Fix for a cursor position during message editing

vhuseinova-msft:
# What

- Fixed an issue with a cursor position when multiline parameter is set to true

# Why

The cursor position should be set to the end of the string instead of the beggining of it when edit is started

# How Tested

Storybook

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/InputBoxComponent.tsx
+++ b/packages/react-components/src/components/InputBoxComponent.tsx
@@ -178,6 +178,12 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
         value={textValue}
         onChange={onChange}
         onKeyDown={onTextFieldKeyDown}
+        onFocus={(e) => {
+          // Fix for setting the cursor to the correct position when multiline is true
+          // This approach should be reviewed during migration to FluentUI v9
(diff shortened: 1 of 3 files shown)


emlynmac (approved):

packages/react-components/src/components/InputBoxComponent.tsx
```diff
@@ -178,6 +178,12 @@ export const InputBoxComponent = (props: InputBoxComponentProps): JSX.Element =>
         value={textValue}
         onChange={onChange}
         onKeyDown={onTextFieldKeyDown}
+        onFocus={(e) => {
+          // Fix for setting the cursor to the correct position when multiline is true
+          // This approach should be reviewed during migration to FluentUI v9
+          e.currentTarget.value = '';
```
palatter, line 184: Why do you have to set this to an empty string first?

vhuseinova-msft: Because otherwise the value change will be ignored as `e.currentTarget.value` is already equal to `textValue` and cursor position (`selectionStart`&`selectionEnd`) won't be reset

palatter (approved):

PR #3238, opened 2023-06-20
[Chat] Fixed Storybook Sendbox showing alert before clear the InputField

Leah-Xia-Microsoft:
# What

Fixed Storybook Sendbox showing alert before clear the InputField
# Why

Storybook should reflect the right behaviour of the real implementation

# How Tested

1. Go to Storybook SendBox page
2. Send a message either from the Docs -> Mention or Preview tab
3. Observe that the input field is cleared out before the alert is shown

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/utils/delay.ts
+++ b/packages/react-components/src/components/utils/delay.ts
@@ -9,8 +9,3 @@ export const delay = (delay: number): Promise<void> => {
     setTimeout(resolve, delay);
   });
 };
-
-/**
- * @private
- */
-export const delayForStorybookSendBoxSendButton = 300;
--- a/packages/react-components/src/components/utils/delay.ts
+++ b/packages/react-components/src/components/utils/delay.ts
@@ -9,3 +9,8 @@ export const delay = (delay: number): Promise<void> => {
     setTimeout(resolve, delay);
   });
 };
+
+/**
+ * @private
+ */
+export const delayForStorybookSendBoxSendButton = 300;
--- a/packages/storybook/stories/SendBox/SendBox.stories.tsx
+++ b/packages/storybook/stories/SendBox/SendBox.stories.tsx
@@ -98,7 +98,7 @@ const getDocs: () => JSX.Element = () => {
 };
 
 const SendBoxStory = (args): JSX.Element => {
-  const timeoutRef = React.useRef<any>();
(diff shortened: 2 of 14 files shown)


packages/react-components/src/components/utils/delay.ts
```diff
@@ -9,3 +9,8 @@ export const delay = (delay: number): Promise<void> => {
     setTimeout(resolve, delay);
   });
 };
+
+/**
+ * @private
+ */
+export const delayForStorybookSendBoxSendButton = 300;
```
emlynmac, line 16: This can just live in the storybook file.

packages/storybook/stories/SendBox/SendBox.stories.tsx
```diff
@@ -98,11 +98,17 @@ const getDocs: () => JSX.Element = () => {
 };
 
 const SendBoxStory = (args): JSX.Element => {
+  const timeoutRef = React.useRef<any>();
```
emlynmac, line 101: any can probably be replaced with something better

packages/storybook/stories/SendBox/snippets/Mentions.snippet.tsx
```diff
@@ -25,24 +25,32 @@ const suggestions: Mention[] = [
...
-            });
-            return Promise.resolve(filtered);
-          }
-        }}
-      />
-    </div>
-  </FluentThemeProvider>
-);
+export const MentionsExample: () => JSX.Element = () => {
+  const timeoutRef = React.useRef<any>();
```
emlynmac, line 29: Type can be improved here?

packages/storybook/stories/SendBox/SendBox.stories.tsx
```diff
@@ -98,11 +98,17 @@ const getDocs: () => JSX.Element = () => {
...
+  const delayForSendButton = 300;
   return (
     <div style={{ width: '31.25rem' }}>
       <SendBoxComponent
         disabled={args.disabled}
-        onSendMessage={async (message) => alert(`sent message: ${message} `)}
+        onSendMessage={async (message) => {
+          timeoutRef.current = setTimeout(() => {
```
vhuseinova-msft, line 108: Probably not very valid situation, but we may need to cancel timer in case if another one is created

vhuseinova-msft (approved):

PR #3239, opened 2023-06-20
[bugfix][a11y][update] Associate in this call with username for React 16

edwardlee-msft:
# What

Replace useId with a custom communication react id generator to support React 16.

Screenshot 2023-06-20 at 11 11 01
Screenshot 2023-06-20 at 11 11 13
Screenshot 2023-06-20 at 11 12 01

# Why

useId is a new a hook introduced in React 18. Our 1.x.x library needs to support React 16+.
https://skype.visualstudio.com/SPOOL/_workitems/edit/3118296

# How Tested
MacOs CallWithChat sample voiceover

# Process & policy checklist


packages/react-composites/src/composites/common/utils.tsx
```diff
@@ -53,3 +53,12 @@ export const useMinMaxDragPosition = (modalLayerHostId: string, rtl?: boolean):
  * @private
  */
 export const defaultSpokenLanguage = 'en-us';
+
+/**
+ * @private
+ *  * Generate a unique id
+ * TODO: Replace with useId() once React 18 becomes a required dependency.
+ */
+export const generateUniqueId = (): string => {
```
dmceachernmsft, line 62: are we placing this hook in both files with the hope to replace them some day? should we not put this in the `acs-ui-common` package so we can use these in other places too if need be?

dmceachernmsft (approved):

packages/react-composites/src/composites/common/utils.tsx
edwardlee-msft: Good catch, yeah that's right. I moved it to acs-ui-common now! :D

packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -51,7 +52,7 @@ export const ParticipantListWithHeading = (props: {
   onFetchParticipantMenuItems?: ParticipantMenuItemsCallback;
 }): JSX.Element => {
   const { onFetchAvatarPersonaData, onFetchParticipantMenuItems, title, participantListProps } = props;
-  const subheadingUniqueId = useId();
```
JamesBurnside, line 54: Can you use useId from fluent instead?

edwardlee-msft: Good suggestion! Changed to use just that

JamesBurnside (approved):

PR #3240, opened 2023-06-20
[Chat] Fix selection is incorrect after deletion

Leah-Xia-Microsoft:
# What

When deleting chars in the middle of a text, the cursor is not at the right index.
This PR is to fix that behavior

# How Tested

1. Paste in some text such as "Hello world" or have some text that contains mention
2. Deleting all the "o"s in the string

# Process & policy checklist

### DIFF
--- a/change-beta/@azure-communication-react-db7d6273-de66-40ff-abe3-9e4fca3c5312.json
+++ b/change-beta/@azure-communication-react-db7d6273-de66-40ff-abe3-9e4fca3c5312.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed selection is incorrect after deletion",
+  "packageName": "@azure/communication-react",
+  "email": "107075081+Leah-Xia-Microsoft@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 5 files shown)


change-beta/@azure-communication-react-db7d6273-de66-40ff-abe3-9e4fca3c5312.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
vhuseinova-msft, line 2: I think it should still be none

vhuseinova-msft (approved):

palatter (approved):

emlynmac (approved):

change-beta/@azure-communication-react-db7d6273-de66-40ff-abe3-9e4fca3c5312.json
Leah-Xia-Microsoft: > I think it should still be none

I think bug fixes should be patch?
None is for things that does not affect our production code at all.

     Prerelease - bump prerelease version
     Patch      - bug fixes; no API changes.
     Minor      - small feature; backwards compatible API changes.
     None       - this change does not affect the published package in any way.

vhuseinova-msft: Mention feature is still work in progress and should be use 'none' type. As soon as we release it, we should use 'patch' type

PR #3243, opened 2023-06-21
[Chat] Hide display mention suggestion items with empty display text

Leah-Xia-Microsoft:
# What

When the display text is empty, we should't show the suggestion item in the list with the default suggestion renderer

# How Tested

# Process & policy checklist


vhuseinova-msft (approved):

jpeng-ms (commented):
so i was testing the storybook, when flyout showed up, and i tap enter key without selecting anyone, `@Unnamed Participant` would show up: 

image


https://github.com/Azure/communication-ui-library/assets/109105353/40a7b3f7-8ce0-45a0-b349-a0ec2d3fc9e7

jpeng-ms (approved):

PR #3247, opened 2023-06-21
Use more specific messages for endcall page when adhoc calling fails

mgamis-msft:
# What
Use more specific messages for endcall page when adhoc calling fails. The Calling subcode descriptions from this table were used to decide on the string. https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/32783/Code-Subcode

# Why
Give more information to user when adhoc call fails

# How Tested
Local calling sample test
image

# Process & policy checklist


dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -548,3 +553,44 @@ const getQueryOptions = (options: { /* @conditional-compile-remove(rooms) */ rol
...
+const getEndedCallStrings = (
+  locale: CompositeLocale,
+  endedCall?: CallState
+): { title: string; moreDetails?: string; disableStartCallButton: boolean } => {
+  let title = locale.strings.call.leftCallTitle;
+  let moreDetails = locale.strings.call.leftCallMoreDetails;
+  let disableStartCallButton = false;
+  /* @conditional-compile-remove(teams-adhoc-call) */
+  switch (endedCall?.callEndReason?.subCode) {
+    case 10037:
```
prabhjot-msft, line 566: what do these codes signify and should we store these somewhere rather than hardcoding here?

prabhjot-msft (approved):
small comment, otherwise looks good!

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
mgamis-msft: Sorry I didn't see this message before merging. But the codes are detailed here https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/32783/Code-Subcode.

PR #3252, opened 2023-06-22
[Alternative stopgap fix] Limit accepting a transfer request to 5 seconds to mitigate timeout failure

mgamis-msft:
# What
Limit accepting a transfer request to 5 seconds to mitigate timeout
Alternative stopgap fix to #3228 

Downsides: 
- we will have reject telemetry data that comes from user accepting after 5 seconds
- not sophisticated at all

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3305535

# How Tested
https://github.com/Azure/communication-ui-library/assets/79475487/2f0bf3d5-4adb-4f78-b397-0d4e77a65f44

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
+++ b/packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
@@ -995,19 +995,26 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
 
   /* @conditional-compile-remove(call-transfer) */
   private transferRequested(args: TransferRequestedEventArgs): void {
+    const acceptanceDeadline = new Date(new Date().getTime() + 5000).getTime();
     const newArgs = {
       ...args,
       accept: (options: AcceptTransferOptions) => {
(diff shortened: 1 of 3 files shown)


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -995,19 +995,26 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
 
   /* @conditional-compile-remove(call-transfer) */
   private transferRequested(args: TransferRequestedEventArgs): void {
+    const acceptanceDeadline = new Date(new Date().getTime() + 5000).getTime();
```
dmceachernmsft, line 998: do we think that 5sec is enough time here? on another note, should this be in the code snippet for contoso as a best practice?

mgamis-msft: No, the code snippet to use a popup would be removed and I would mention that request needs to be accepted or rejected right away

PR #3259, opened 2023-06-23
Remove webpack4 dependency from storybook

emlynmac:
# What
Update webpack4 dependency to webpack5

# Why
Preparing for storybook v7

# How Tested
CI

# Process & policy checklist

### DIFF
--- a/common/config/rush/common-versions.json
+++ b/common/config/rush/common-versions.json
@@ -61,10 +61,6 @@
 
       // Typescript check project's dependency -- this is our minimum supported typescript version of our npm package.
       "3.8.3"
-    ],
-    "webpack": [
-      // All projects should use this version by default
-      "5.38.1"
     ]
   }
 }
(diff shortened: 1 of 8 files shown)


packages/storybook/package.json
```diff
@@ -111,7 +111,6 @@
     "ts-node": "^9.1.1",
     "typescript": "4.3.5",
     "uuid": "^8.1.0",
-    "webpack": "5.76.0",
-    "webpack4": "npm:webpack@^4.44.2"
```
JamesBurnside, line 115: Can we remove webpack from the `common-versions.json` now? I think all uses of differing versions of webpack are now resolved..

emlynmac: Will update and test.

common/config/rush/common-versions.json
```diff
@@ -64,11 +64,7 @@
     ],
     "webpack": [
```
JamesBurnside, line 65: If there's no differing versions, we should be able to remove this entirely (this field is just to allow different versions of the same dependency across the repo) - but understand if you want to keep that out of this PR in case it causes issues

dmceachernmsft (approved):

PR #3260, opened 2023-06-23
[Chat] New file metadata types

jimchou-dev:
# What

- Introduce BaseFileMetadata that is used for different type fileMetadata types

# Why

# How Tested

# Process & policy checklist


JoshuaLai (approved):
Went through the PR and looks good. 
The tests comparing against the existence of `id` is not ideal. We will discuss what a better solution in the next iteration.

PR #3275, opened 2023-06-26
Raise hand feature react composite

rzdor:
What

This is POC for Raise Hand feature.
I've added implementation for Raise/Lower Hand button to components.
Updated composites and Calling test app to use new element.
Added states and subscribers to hand updates for new published/removed states.

image
image

Why
Implementation for SDK raise hand feature

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -396,6 +396,29 @@ export interface CallAdapterCallOperations {
...
+  lowerHand(): Promise<void>;
+  /* @conditional-compile-remove(raise-hand) */
+  /**
+   * Lower hand for participants from the call.
+   *
+   * @param userId - List of Ids of the participant
+   *
+   * @public
+   */
+  lowerHands(userIds: string[]): Promise<void>;
```
alkwa-msft, line 421: I am curious would there ever be an option where you can lower 3 peoples but not the other 4? I question if it needs these parameters and if it should just be an all or nothing?

PR #3286, opened 2023-06-28
Hoist error tracking to composite level

JamesBurnside:
# What

Hoist error dismissed tracking to composite level

# Why

We had two issues:

1. Errors would reappear on different pages. If I dismissed an error on the lobby page, it would reappear on the call page. This was because the error bar would be unmounted and remounted thus losing the fact it had previously tracked an error had been dismissed.
2. Errors now appear in the side pane instead of the call screen, but the same error shows on the call screen if the side pane is closed without dismissing the error. Because these are two separate components there was no shared knowledge that an error was dismissed.

# How Tested
Locally:

https://github.com/Azure/communication-ui-library/assets/2684369/1b82d165-cda0-42e5-aa4c-d34138e4cd32


packages/react-composites/src/composites/CallComposite/types/ErrorTracking.ts
```diff
@@ -0,0 +1,13 @@
...
+import { ErrorType } from '@internal/react-components';
+
+/** @private */
+export interface ErrorTrackingInfo {
+  mostRecentlyActive: Date;
+  lastDismissedAt?: Date;
+}
+
+/** @private */
+export type TrackedErrors = Partial<Record<ErrorType, ErrorTrackingInfo>>;
```
dmceachernmsft, line 13: Not requesting a change here but we should think about the typing here. We had a ARB comment recently around types defined as `{ [key: string]: Object }` that is just essentially just a `Record<string, Object>`. do we  know which typing we want to align with to not confuse the ARB? (We currently observe the former)

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/types/ErrorTracking.ts
JamesBurnside: Great point, no idea on this. Do you have a link to the apiview comment? They should update the azure policies if they have a strong preference but I'm also curious what the difference is!
(This is private so at least doesn't matter for this PR but is a good callout to highlight for everyone across the repo)

edwardlee-msft (approved):

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -256,6 +265,19 @@ const MainScreen = (props: MainScreenProps): JSX.Element => {
     onSidePaneIdChange?.(sidePaneRenderer?.id);
   }, [sidePaneRenderer?.id, onSidePaneIdChange]);
 
+  // Track the last dismissed errors of any error kind to prevent errors from re-appearing on subsequent page navigation
+  // This works by tracking the most recent timestamp of any active error type.
+  // And then tracking when that error type was last dismissed.
+  const activeErrors = usePropsFor(ErrorBar).activeErrorMessages;
+  const [trackedErrors, setTrackedErrors] = useState<TrackedErrors>({} as TrackedErrors);
+  useEffect(() => {
+    setTrackedErrors((prev) => updateTrackedErrorsWithActiveErrors(prev, activeErrors));
```
prabhjot-msft, line 274: What do the tracked errors signify and are these internal to the call composite. Shouldn't we only have active and dismissed errors maintained globally?

packages/react-composites/src/composites/CallComposite/components/SidePane/useVideoEffectsPane.tsx
```diff
@@ -52,32 +51,27 @@ export const useVideoEffectsPane = (
...
-  }, [dismissedVideoEffectsError, latestVideoEffectError, activeVideoEffect]);
+  /* @conditional-compile-remove(video-background-effects) */
+  const latestVideoEffectError = latestErrors.find((error) => error.type === 'unableToStartVideoEffect');
 
   const onRenderContent = useCallback((): JSX.Element => {
     return (
       <VideoEffectsPaneContent
-        onDismissError={onDismissVideoEffectError}
-        activeVideoEffectError={activeVideoEffectError}
-        setActiveVideoEffect={setActiveVideoEffect}
```
prabhjot-msft, line 77: We have removed set active video effects, don't we need that information for the pane anymore?

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
JamesBurnside: TrackedErrors is the a mapping of any errors that have been triggered to information about the timestamp when they were last active, and the timestamp when it was last dismissed. `trackedErrors` maps the error type to an object containing:
```
  mostRecentlyActive: Date;
  lastDismissedAt?: Date;
```

So, for example, when a new error comes in, that error will be added to set:
```
trackedErrors = {
  newError1: {
    lastActive: Date.now()
  }
}
```

Then if another error comes in:
```
trackedErrors = {
  newError1: {
    lastActive: <oldTimestampA>
  },
  newError2: {
    lastActive: Date.now()
  }
}
```

Then if error1 is dismissed but is still active (e.g. no headphones are still plugged in but the error was dismissed):
```
trackedErrors = {
  newError1: {
    lastActive: <oldTimestampA>,
    dismissedAt: Date.now()
  },
  newError2: {
    lastActive: <oldTimeStampB>
  }
}
```

If error1 is resolved and no longer exists in stateful, its removed from the set of errors being tracked:
```
trackedErrors = {
  newError2: {
    lastActive: <oldTimeStampB>
  }
}
```

If it reappears again, it gets added as normal, and will reappear to the user to be dismissed again:
```
trackedErrors = {
  newError1: {
    lastActive: Date.now()
  },
  newError2: {
    lastActive: <oldTimeStampB>
  }
}
```

JamesBurnside: The tracking is local to the CallComposite, if someone plugged the same adapter into a new CallComposite the errors would reappear to be dimissed

PR #3287, opened 2023-06-28
[Doc] Add multi-thread code snippet

AmyL219:
# What
Based on customer signals, we've included a basic code snippet demo to show how to use our components to start multiple chat threads.

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/docs/code-examples/MultiThreadExample.tsx
+++ b/docs/code-examples/MultiThreadExample.tsx
@@ -1,0 +1,119 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import React, { useEffect, useState } from "react";
+import {
+  ChatClientProvider,
+  ChatThreadClientProvider,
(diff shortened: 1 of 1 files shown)


emlynmac:
Let's get this code into the samples directory, so that it can be validated by CI and worked into the examples formally.

PorterNan (commented):
Do we have a code-example folder under docs/ before, it feels it is not a place for tsx file to live, this is an official place for documentation and explaining how things are working, if we need something here, it should be a .md file with some sections talking about how to use multi thread with composite

Either we could put it in storybook page, or under our samples folder, if we feel this code piece is not polished enough, then we can put it in discussion show and tell, polished this a bit more then put into our samples or storybook folder

AmyL219:
Closed the PR as the sample code has been post inside [`discussion show and tell`](the https://github.com/Azure/communication-ui-library/discussions/3300)

PR #3291, opened 2023-06-28
Resolve issue where the number zero appeared in edit box.

palatter:
# What

Add greater than zero check for fileUploadRender

# Why

Resolve issue where the number zero appeared in edit box.

image

# How Tested

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageComponentAsEditBox.tsx
```diff
@@ -98,7 +98,8 @@ export const ChatMessageComponentAsEditBox = (props: ChatMessageComponentAsEditB
 
   const onRenderFileUploads = useCallback(() => {
     return (
-      attachedFilesMetadata?.length && (
+      !!attachedFilesMetadata?.length &&
```
emlynmac, line 101: should this be just a check for not undefined, as this would then allow the removal of the optional `?` in the length call?

palatter: Pushed a change for this.

Leah-Xia-Microsoft (approved):
Tested on the stroybook link of this PR

PR #3304, opened 2023-06-29
[Skip Screens] Fix Call Adapter JoinCall API to also allow start with video

dmceachernmsft:
# What
Fix Call Adapter JoinCall API to also allow start with video

# Why
Enable contoso to start call with config skipped and video on by calling adapter.joinCall( true, true) 

# How Tested
Tested with samples 

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -1524,6 +1526,9 @@ export const createAzureCommunicationCallAdapterFromClient: (
   options?
 ): Promise<CallAdapter> => {
   const deviceManager = (await callClient.getDeviceManager()) as StatefulDeviceManager;
+  deviceManager.getCameras();
```
JamesBurnside, line 1529: I can't remember, does this trigger the camera permissions browser popup?

JamesBurnside: (if so we can't call this in the adapter constructor)

carocao-msft: I tried it on incognito mode and didn't see any pop up aside from call readiness modals

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -793,7 +793,7 @@ export interface CallAdapterCallManagement extends CallAdapterCallOperations {
    *
    * @public
    */
-  joinCall(microphoneOn?: boolean): Call | undefined;
+  joinCall(microphoneOn?: boolean, cameraOn?: boolean): Call | undefined;
```
JamesBurnside, line 796: A list of optional arguments is bad, but we can't change the `microphoneOn` one now..

I approve this change. It might be worth asking sdk reviewers on the side if its better to have what we have, or should we introduce another function instead like:
```ts
joinCallWithOptions(options: {microphoneOn?: boolean, cameraOn?: boolean});
```

and have:

```ts
joinCall(); // new function
joinCall(microphoneOn?: boolean) // mark as deprecated
joinCallWithOptions(options) //new function
```

But don't block PR on this

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -1524,6 +1526,9 @@ export const createAzureCommunicationCallAdapterFromClient: (
...
+  deviceManager.getMicrophones();
```
JamesBurnside, line 1530: what's the reason we need to getmics and getspeakers for this change?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
JamesBurnside: For skipping config screen and joining immediately should it not skip the call readiness modals?

JamesBurnside: What's the reason we need to call this in the constructor also? Could we call it on joinCall if cameraOn is true?

carocao-msft: Tried incognito mode without skipping config screen and the only pop up I see is call readiness. When config screen is skipped and I go into the call without permissions set, I see this 
Screenshot 2023-06-30 at 2 21 11 PM
When permissions are set before going into the call, I see no pop up
I think this  should be expected? 
The reason we call getCameras and getMics here is because we found that before, we are getting device info at the middle of call connecting,  in this way we can get the devices at the correct time

packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
dmceachernmsft: I do like the idea of overloads here as well, the options bag version would be a good why to bring us closer to the SDK without all the complexity of needing the streams to do it.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
dmceachernmsft: without these fetches if we call `joinCall()` to skip the screen there will be no camera or mics in the adapter. we should be fetching the devices as soon as the `DeviceManager` is made

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
dmceachernmsft: this should yes, something to test @carocao-msft 👀

JamesBurnside: That sounds right re the permissions thanks for checking!
We don't await the getDevices calls so there's no guarantee we will have device information and it could still popup during the connecting phase.
Perhaps best to await them all here (use await Promise.all to avoid sequential waiting)?

JamesBurnside: (sadly we can't call it on joinCall becuase it's not async fn :/)

carocao-msft: Updated :)

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -554,15 +554,17 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
...
       throw new Error('You are already in the call!');
     }
 
     return this.teeErrorToEventEmitter(() => {
       const audioOptions: AudioOptions = { muted: !(microphoneOn ?? this.getState().isLocalPreviewMicrophoneEnabled) };
       // TODO: find a way to expose stream to here
-      const videoOptions = { localVideoStreams: this.localStream ? [this.localStream] : undefined };
+      const selectedCamera = getSelectedCameraFromAdapterState(this.getState());
+      const localStream = selectedCamera ? new SDKLocalVideoStream(selectedCamera) : undefined;
+      const videoOptions = localStream && cameraOn ? { localVideoStreams: [localStream] } : {};
```
JamesBurnside, line 567: This is a breaking change.

If a customers code turns on the camera then they have:
```
joinCall(true);
```
They want to join the call with cameraOn. However, with this change we will now interpret this as:
```
joinCall(true, false);
```
Which is not their intention.

Perhaps the parameter of this function should be `camera?: true | false | 'auto'` where auto is the default?
Maybe another argument for creating a new function to handle this called something like joinCallWithOptions

carocao-msft: If customer call joinCall(true), they will only join call with mic on.
And calling joinCall() means joining with mic off and camera off

JamesBurnside: Today though, if a customer calls `joinCall(true)` they join the call with micOn, and with camera on if this.localStream is true

dmceachernmsft: sorry I don't follow why this is breaking @JamesBurnside are you trying to say that if you have code to turn on the camera in contoso's application that setting the second value as false will break that on skip screens?

PR #3319, opened 2023-07-05
[Skip Screens] New Call Adapter JoinCall API to also allow start with video

carocao-msft:
# What

JoinCallWithOptions to allow start call with video and mic 

# Why

This is an extension from this PR https://github.com/Azure/communication-ui-library/pull/3304
Directly modifying joinCall can be a breaking change to customers 
So we will need to implement a new joinCall api to allow call start with both mic and video

#Question
Do we need to mark this as beta and then stablize it? 
It's mainly a copy of the existing joinCall API aside from the new prop and videoStream changes (the videoStream changes are marked as a todo item in the original joinCall API, however we cannot change it without breaking some customer's code) 

# How Tested
Tested with samples 

# Process & policy checklist


change/@azure-communication-react-16281876-dbd6-44fc-bcb2-29f5eb9f7109.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
dmceachernmsft, line 2: If we can go to GA with this then this should be marked as minor

packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -459,7 +460,6 @@ export const getBackgroundEffectFromSelectedEffect = (
     ? new BackgroundReplacementEffect({ backgroundImageUrl: selectedEffect.backgroundImageUrl })
     : undefined;
 
-/* @conditional-compile-remove(video-background-effects) */
```
dmceachernmsft, line 462: Why did this get removed?

carocao-msft: The function used for this pr is conditionally compiled under video-background-effects, so need to remove it to make it work in stable

dmceachernmsft (approved):

edwardlee-msft (approved):

PR #3321, opened 2023-07-05
[Storybook] Improve Storybook Sidebar Navigation and Content Arrangement

AmyL219:
# What

- To improve navigation, I've implemented a workaround that enables auto-scrolling within the [ChatComposite section](https://60c7ae6891f0e90039d7cd54-rgihqllbwl.chromatic.com/?path=/docs/composites-chat-basicexample--basic-example), [CallComposite](https://60c7ae6891f0e90039d7cd54-rgihqllbwl.chromatic.com/?path=/docs/composites-call-customdatamodelexample--custom-data-model-example) and [CallwithChatComposite](https://60c7ae6891f0e90039d7cd54-rgihqllbwl.chromatic.com/?path=/docs/composites-call-with-chat-jointeamsmeeting--join-teams-meeting) when users either click on the sidebar or access a subsection via URL.

- I've separated the single-page documentation for IncomingCallAlertsDocs into distinct sections for [Modal](https://60c7ae6891f0e90039d7cd54-rgihqllbwl.chromatic.com/?path=/docs/examples-incomingcallalerts-incomingcallmodal--incoming-call-modal) and [Toast](https://60c7ae6891f0e90039d7cd54-rgihqllbwl.chromatic.com/?path=/docs/examples-incomingcallalerts-incomingcalltoast--incoming-call-toast). The [original version](https://azure.github.io/communication-ui-library/?path=/docs/examples-incomingcallalerts-incomingcalltoast--incoming-call-toast) is here.
- The Teams beta features([File sharing](https://main--60c7ae6891f0e90039d7cd54.chromatic.com/?path=/docs/examples-teamsinterop-filesharing--file-sharing) and [Inline Image](https://main--60c7ae6891f0e90039d7cd54.chromatic.com/?path=/docs/examples-teamsinterop-inlineimage--inline-image)) has been moved to the '[Use Case](https://60c7ae6891f0e90039d7cd54-rgihqllbwl.chromatic.com/?path=/docs/usecases--page#teams-interop-use-case)s' section and is highlighted with a styled banner.
- Inside the [Use Cases](https://60c7ae6891f0e90039d7cd54-rgihqllbwl.chromatic.com/?path=/docs/usecases--page#calling-use-cases) chart link Teams Interop section

# Why
To enhance user experience and improve ease of navigation within the Storybook

# How Tested

Storybook tests

# Process & policy checklist


JoshuaLai:
Not specifically an issue per say: But it does seem odd to me that the order of the menu items on the right aren't in the same order as how they are laid on out on the page. This feels odd I would tap on `theme example` and I would jump to a part of the page that is above `join existing chat thread`. I would expect the layout of the page to be the same order as the side pane menu.

JoshuaLai:
Right now on all the pages, it seems a little inconsistent (maybe this is something we need to bring up with the MSDocs team) which items we have a link for on the right side bar, and which ones we don't. For example the screenshot below. And this is inconsistent through the whole storybook. Maybe not a problem to be solved for this PR but I would like this better thought out. 
image

AmyL219:
> Right now on all the pages, it seems a little inconsistent (maybe this is something we need to bring up with the MSDocs team) which items we have a link for on the right side bar, and which ones we don't. For example the screenshot below. And this is inconsistent through the whole storybook. Maybe not a problem to be solved for this PR but I would like this better thought out. 

> Not specifically an issue per say: But it does seem odd to me that the order of the menu items on the right aren't in the same order as how they are laid on out on the page. This feels odd I would tap on `theme example` and I would jump to a part of the page that is above `join existing chat thread`. I would expect the layout of the page to be the same order as the side pane menu.


I understand your concerns and we can certainly note them for enhancement during our MDX migration-[wiki page](https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/43675/Migration-to-Storybook-7). The reason why some items have a link and others don't is due to the 'Preview' tab. Sections like 'Custom Date Format' which we want to provide a 'Preview' tab for, have a sidebar. These are the Storybook 'stories' that allow users to interact with and try out the components.

dmceachernmsft (approved):

JoshuaLai (approved):

PR #3322, opened 2023-07-05
[Chat] Mention editing fixes

vhuseinova-msft:
# What

Fix issues:
1.  with a cursor position when a mention is selected by mouse.

- type something
- navigate to the middle of the text
- select a mention by mouse and check that cursor position is after the mention

2. when another mention is added before the previous one

- add a mention
- navigate before that mention
- add another mention
- check that mention is added

3. when a new text is typed before and after mention

- type something before mention, check that html text is correct
- type something after the mention, check that html text is correct

4. when last name in mention is selected and a key is tapped

- add a mention
- select last name only
- type something
- check that the text is correct

# Why

# How Tested

Storybook

# Process & policy checklist


Leah-Xia-Microsoft (approved):
Tested on Storybook

PR #3324, opened 2023-07-06
[Chat] Fixed Video Not Scale Properly

jpeng-ms:
# What
Addressed the issue where video messages not being scaled properly in message container.

Note that currently the video content would appear to be blank and this is expected and it needs to be addressed in the backend.

# Why

|  Before  | After  |
|---|---|
|  image | image  |

# How Tested
run call with chat sample, join a Teams call, send a video message from Teams side, verify if blank video tag is scaled properly in container.

# Process & policy checklist

- [X] I have updated the project documentation to reflect my changes if necessary.
- [X] I have read the [CONTRIBUTING](https://github.com/Azure/communication-ui-library/blob/main/CONTRIBUTING.md) documentation.


jimchou-dev (approved):
LGTM

vhuseinova-msft (approved):

PR #3325, opened 2023-07-07
Update communication common

dmceachernmsft:
# What

Updates Communication Common package based on Calling's need for beta version.
# Why

# How Tested

# Process & policy checklist


JamesBurnside (changes requested):
The `force-build-flavors.mjs` will need updated in this PR
(bug on our repo: we should have something in the CI that fails if we have a dependency that differs between beta and stable and is not part of the list in the force-build-flavors file)

PR #3332, opened 2023-07-12
Fix unread chat messages badge on the CallWithChatComposite being cleared when the call goes on hold

JamesBurnside:
# What

* Move unread messages tracking in callwithchatcomposite to the composite level instead of component level to ensure state is not lost when the button is unmounted.
* Move the tracking logic into a hook for better separation

# Why

* Unread message count was being lost when the call went on hold

# How Tested

Verified locally

image

image

image


packages/react-composites/src/composites/CallWithChatComposite/ChatButton/useUnreadMessagesTracker.ts
```diff
@@ -0,0 +1,40 @@
...
+  }, [chatAdapter, setUnreadChatMessagesCount, isChatPaneVisible, unreadChatMessagesCount]);
+
+  return unreadChatMessagesCount;
+};
+
+/**
+ * Helper function to determine if the message in the event is a valid one from a user.
+ * Display name is used since system messages will not have one.
+ */
+const validNewChatMessage = (message): boolean =>
```
edwardlee-msft, line 39: nit: Would this break with CTE users that have unnamed participant? (Due to cte bug and no graph integration with calling sdk yet).
cc @PorterNan

edwardlee-msft (approved):

packages/react-composites/src/composites/CallWithChatComposite/ChatButton/useUnreadMessagesTracker.ts
JamesBurnside: Great spot! I won't change in this PR as this function was just moved but I think you're right!

carocao-msft (approved):

packages/react-composites/src/composites/CallWithChatComposite/ChatButton/useUnreadMessagesTracker.ts
JamesBurnside: Filed a bug https://skype.visualstudio.com/SPOOL/_workitems/edit/3342861

PR #3338, opened 2023-07-12
Effects pane header misaligned

prabhjot-msft:
# What
Effects pane header misaligned

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3331285

# How Tested
Before:
MicrosoftTeams-image (27)

After:
Screenshot 2023-07-12 at 3 09 39 PM

# Process & policy checklist


packages/react-composites/src/composites/common/styles/ParticipantContainer.styles.ts
```diff
@@ -40,7 +40,7 @@ export const sidePaneHeaderStyles: IStackItemStyles = {
   root: {
     lineHeight: '1.5rem',
     fontSize: '1.125rem',
-    padding: '0.75rem 0.75rem 0.75rem 0.25rem',
+    padding: '0.75rem 0.75rem 0.75rem 0.50rem',
```
JamesBurnside, line 43: Just to confirm this change is by design, the current figma's have it slightly out hence the current behavior:
image

mgamis-msft: Seeing the same thing in Figma. Did Alex P file this bug?

prabhjot-msft: That is reported by AlexP in the bug bash, let me double check

prabhjot-msft: I talked to Alex, the Figma is updated with the alignment fix

mgamis-msft:
I thinnk we change the style of the VideoEffectPane component's root Stack like so:
image
so that we don't affect the alignment in the other side panes like participant pane as seen in the snapshot updates. This is what it looks like from my testing
image

mgamis-msft (approved):

carocao-msft (approved):

PR #3340, opened 2023-07-13
Added capabilites feature to stateful client

mgamis-msft:
# What
Added capabilities feature to stateful client

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3336592

# How Tested
Console.log capabilities properties in Rooms call for Presenter:
image
and Consumer:
image

# Process & policy checklist


packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -133,6 +135,19 @@ export interface TranscriptionCallFeatureState {
...
+/**
+ * State only version of {@link @azure/communication-calling#CapabilitiesFeature}
+ *
+ * @beta
+ */
+export interface CapabilitiesFeatureState {
+  /**
+   * Proxy of {@link @azure/communication-calling#CapabilitiesFeature.capabilities}.
+   */
+  capabilities: ParticipantCapabilities;
```
dmceachernmsft, line 148: Do we want to call this `localParticipantCapabilities`?

dmceachernmsft (approved):
Small naming question otherwise looks great!

common/config/babel/.babelrc.js
```diff
@@ -66,15 +66,19 @@ process.env['COMMUNICATION_REACT_FLAVOR'] !== 'beta' &&
...
       // A list of in progress beta feature.
       // These features are still beta feature but "in progress"
       // causing a build failure, but they are ignored by the preprocessing step.
       inProgressFeatures: [
         // Demo feature. Used in live-documentation of conditional compilation.
         // Do not use in production code.
-        'in-progress-beta-feature-demo'
+        'in-progress-beta-feature-demo',
+        // Feature for capabilities
+        'capabilities'
```
alkwa-msft, line 81: do we want to commit this change as  well?

alkwa-msft (approved):

packages/calling-stateful-client/src/CallClientState.ts
dmceachernmsft: I know, something just seems a little funky that the type has participant in the name but not the definition itself. was just thinking readability sake, not blocking 😊

PR #3345, opened 2023-07-14
[bugfix][a11y] Label side pane tab header currently active tab as aria-label

edwardlee-msft:
# What
Aria-label side pane tab header to the currently active tab that is open. Then revert once another tab is active.
| Active tab | Not active |
| :--------: | :--------: |
|  |  |
|  |  |

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/2790227
Unaware which sidepane pane is currently "active" through voice narration.

# How Tested
Locally on mobile view.
Also on Android device.

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/common/TabHeader.tsx
+++ b/packages/react-composites/src/composites/common/TabHeader.tsx
@@ -79,11 +79,7 @@ export const PeopleAndChatHeader = (props: PeopleAndChatHeaderProps): JSX.Elemen
             onClick={onChatButtonClicked}
             styles={mobilePaneButtonStylesThemed}
             checked={activeTab === 'chat'}
-            ariaLabel={
-              activeTab === 'chat'
-                ? strings.sidePaneActiveTabAriaLabel.replace('{header}', strings.chatButtonLabel)
-                : ''
-            }
+            aria-selected={activeTab === 'chat'}
             role={'tab'}
             disabled={props.disableChatButton}
           >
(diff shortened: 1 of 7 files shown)


packages/react-composites/src/composites/common/TabHeader.tsx
```diff
@@ -79,6 +79,11 @@ export const PeopleAndChatHeader = (props: PeopleAndChatHeaderProps): JSX.Elemen
             onClick={onChatButtonClicked}
             styles={mobilePaneButtonStylesThemed}
             checked={activeTab === 'chat'}
+            ariaLabel={
+              activeTab === 'chat'
+                ? strings.sidePaneActiveTabAriaLabel.replace('{header}', strings.chatButtonLabel)
```
JamesBurnside, line 84: Out of interest, would setting [aria-selected](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected) property do the same thing without updating the string?

packages/react-composites/src/composites/localization/locales/en-US/strings.json
```diff
@@ -23,6 +23,7 @@
     "copyInviteLinkActionedAriaLabel": "Invite link copied",
     "defaultPlaceHolder": "Select an option",
     "dismissSidePaneButtonLabel": "Close",
+    "sidePaneActiveTabAriaLabel": "{header} currently selected",
```
mgamis-msft, line 26: 'People currently selected' and 'Chat currently selected' might be a bit confusing and might be mistranslated. Can we change this "'{header}' tab is currently selected'" unless Alex P specifically wanted this string.

jimchou-dev (approved):

packages/react-composites/src/composites/common/TabHeader.tsx
edwardlee-msft: That's a great suggestion!
Let me try this and report back

packages/react-composites/src/composites/localization/locales/en-US/strings.json
edwardlee-msft: Changed to use aria-selected instead.

packages/react-composites/src/composites/common/TabHeader.tsx
edwardlee-msft: It worked! Thanks James, this is a better cleaner fix.

JamesBurnside (approved):

PR #3348, opened 2023-07-17
[Chat] Fix mention popover position incorrect when insert mention in the middle of long text

Leah-Xia-Microsoft:
# What

The top prop of the relativePosition we get back from the Caret library is the distance between the top of the scroll view to the cursor position.
We use to adjust this value by subtract the difference between the scrollView height - the contentHeight.
The problem of this approach is that it's not considering the current scroll position.
Instead, when the view is scrollable, we should use relativePosition.top subtracted by scrollTop to get the relative position to the content view

From MDN:
The Element.scrollTop property gets or sets the number of pixels that an element's content is scrolled vertically.
An element's scrollTop value is a measurement of the distance from the element's top to its topmost visible content. When an element's content does not generate a vertical scrollbar, then its scrollTop value is 0.

# Why

Fix mention popover position incorrect when insert mention in the middle of long text

# How Tested

Steps:

1. add a very long text
2. scroll to the middle of the text
3. add a mention
Observe the position of the Mention Popover

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
+++ b/packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
@@ -562,9 +562,8 @@ export const TextFieldWithMention = (props: TextFieldWithMentionProps): JSX.Elem
         // Update the caret position, used for positioning the suggestions popover
         const textField = event.currentTarget;
         const relativePosition = Caret.getRelativePosition(textField);
-        const adjustOffset = Math.max(0, textField.scrollHeight - textField.clientHeight);
-        if (relativePosition.top > adjustOffset) {
-          relativePosition.top -= adjustOffset;
+        if (textField.scrollHeight > textField.clientHeight) {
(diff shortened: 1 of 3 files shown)


emlynmac (approved):

packages/react-components/src/components/TextFieldWithMention/TextFieldWithMention.tsx
```diff
@@ -562,9 +562,8 @@ export const TextFieldWithMention = (props: TextFieldWithMentionProps): JSX.Elem
         // Update the caret position, used for positioning the suggestions popover
         const textField = event.currentTarget;
         const relativePosition = Caret.getRelativePosition(textField);
-        const adjustOffset = Math.max(0, textField.scrollHeight - textField.clientHeight);
```
vhuseinova-msft, line 565: |after scrolling and typing the position looks okay|when open the page for the first time, paste text and start to add a first mention, the mention popover is a little bit higher|
|-|-|
| image|image |

PR #3351, opened 2023-07-18
[A11y bugfix] Fix video effects pane tab ordering on call screen

dmceachernmsft:
# What

Pulls focus to the video effects pane on the call screen when the side pane is opened
# Why

Moves focus from the camera button to the sidepane so the user can more efficiently select their video effect.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3327121
# How Tested

Locally in test apps

### DIFF
--- a/packages/react-components/src/components/VideoEffects/VideoBackgroundEffectsPicker.tsx
+++ b/packages/react-components/src/components/VideoEffects/VideoBackgroundEffectsPicker.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { IStyle, Label, mergeStyles, Stack } from '@fluentui/react';
+import { FocusZone, IStyle, Label, mergeStyles, Stack } from '@fluentui/react';
 import { useWarnings } from '@fluentui/react-hooks';
 import React from 'react';
 import { chunk } from '../utils';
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/VideoEffects/VideoBackgroundEffectsPicker.tsx
```diff
@@ -136,7 +136,9 @@ export const _VideoBackgroundEffectsPicker = (props: _VideoBackgroundEffectsPick
           data-ui-id="video-effects-picker-row"
         >
           {options.map((option) => (
-            <_VideoEffectsItem {...option} key={option.itemKey} itemKey={option.itemKey} />
+            <FocusZone key={option.itemKey} shouldFocusOnMount={!!(option.itemKey === 'none')}>
```
mgamis-msft, line 139: mega nit: can 
```
shouldFocusOnMount={!!(option.itemKey === 'none')}
```
be simplified to
```
shouldFocusOnMount={option.itemKey === 'none'}
```

mgamis-msft (approved):

packages/react-components/src/components/VideoEffects/VideoBackgroundEffectsPicker.tsx
prabhjot-msft: Not sure why do we need this condition here `shouldFocusOnMount={!!(option.itemKey === 'none')`

dmceachernmsft: I tried this the other way and it did not work. the !! makes the check truthy like an if statement. Typescript was throwing a compiler error with out it 
image

prabhjot-msft: yeah, there can be a case if the active effect is updated after creating the adapter, `none` would be the first item but not the active effect

prabhjot-msft (approved):

PR #3359, opened 2023-07-21
Update captions banner width and height to increase readability

carocao-msft:
# What
Update captions banner width and height to increase readability

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3245322
https://skype.visualstudio.com/SPOOL/_workitems/edit/3245318

# How Tested
Tested with samples 

# Process & policy checklist


change-beta/@azure-communication-react-df2fc150-ffbd-4e1a-bb90-ad1ea5304bf2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
dmceachernmsft, line 2: If this is for Captions GA is this pre-release?

packages/react-components/src/components/styles/Captions.style.ts
```diff
@@ -65,9 +65,13 @@ export const captionsBannerClassName = (formFactor: 'default' | 'compact'): ISta
 /**
  * @private
  */
-export const loadingBannerClassName = mergeStyles({
-  height: _pxToRem(100)
-});
+export const loadingBannerClassName = (formFactor: 'default' | 'compact'): IStackStyles => {
```
dmceachernmsft, line 68: Can we change the naming for these styles? I don't agree that they are a classname, we should be calling this `loadingBannerStyles` there are a lot of other classes like this.

edwardlee-msft: I agree with this as well.

edwardlee-msft (approved):

change-beta/@azure-communication-react-df2fc150-ffbd-4e1a-bb90-ad1ea5304bf2.json
carocao-msft: It's feature related changes, I think it should be prerelease?

PR #3360, opened 2023-07-21
[RoomsGA] Remove roleHint and permissions provider rooms

dmceachernmsft:
# What

Removes the RoleHint from the adapter API and the permissions provider that was providing the role to rooms
- is a rooms call is now stored in the adapter
- component bindings is now tracking for the buttons the role of the participant in a rooms call
- disables the button if they do not have the right permissions <- this can also be checked against in the composite allowing to hide the buttons as well
- this rips out the permission provider out of the UI library and replaces the fetching of roles from the SDK. (sorry about the size of changes)
# Why

simplifies the API for Rooms making it easier for people to start a rooms call without providing a role hint on the UI side.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3348000
# How Tested

Tested locally. just a draft for now will update tests and such later.


packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -94,6 +98,8 @@ export type ParticipantListSelector = (
   participants: CallParticipantListParticipant[];
   myUserId: string;
   totalParticipantCount?: number;
+  /* @conditional-compile-remove(rooms) */
+  myRole?: Role;
```
mgamis-msft, line 102: Do we need this to show the local user's role in the participant list?

packages/calling-component-bindings/src/callControlSelectors.ts
```diff
@@ -109,12 +115,23 @@ export type ScreenShareButtonSelector = (
...
+    /* @conditional-compile-remove(rooms) */ role
+  ) => {
     return {
       checked: isScreenSharingOn,
       /* @conditional-compile-remove(PSTN-calls) */
-      disabled: callState === 'InLobby' ? true : callState === 'Connecting' ?? false
+      disabled:
+        (callState === 'InLobby' ? true : callState === 'Connecting' ?? false) ||
+        /* @conditional-compile-remove(rooms) */ role === 'Consumer' ||
+        /* @conditional-compile-remove(rooms) */ role === 'Attendee'
```
mgamis-msft, line 134: I will override the changes in this file in [PR](https://github.com/Azure/communication-ui-library/pull/3374/files#diff-5fc9f58576485df70f1f052afe532582e08f93d9ff12bc0f35df0111c0475abf) but this is a good intermediate.

packages/calling-component-bindings/src/participantListSelector.ts
dmceachernmsft: was to simplify the internal logic of the participant list. we can also map against the users Id.

packages/calling-component-bindings/src/callControlSelectors.ts
dmceachernmsft: I think we can still do the mapping through roles as a condition as well. since rooms might not respect the capabilities the same way as interop

packages/react-composites/src/composites/CallComposite/components/buttons/Devices.tsx
```diff
@@ -65,19 +69,16 @@ export const Devices = (props: {
...
   );
 };
 
 /* @conditional-compile-remove(rooms) */
-const getLabelFromPermissions = (
-  permissions: _Permissions,
-  locale: CompositeLocale
-): Partial<DevicesButtonStrings> | undefined => {
-  if (!permissions.cameraButton && !permissions.microphoneButton) {
+const getLabelFromPermissions = (role: Role, locale: CompositeLocale): Partial<DevicesButtonStrings> | undefined => {
```
mgamis-msft, line 80: Rename to getLabelFromRole?

mgamis-msft (approved):

edwardlee-msft (approved):

mgamis-msft:
Snapshots look good!

PR #3371, opened 2023-07-25
Updated Beachball config with new prompts

JoshuaLai:
# What
Introduce two new prompts to `rush changelog` 
New area:
image
New workstream:
image

# Why
Introduce new custom properties generated into the `.json` for us to index on to drive clarity on where an entry should belong. 

# How Tested
run `rush changelog`
validate the json generated has the new fields as expected 

# Process & policy checklist


common/config/beachball/changelog-custom-renders.ts
```diff
@@ -1,9 +1,13 @@
-import { PackageChangelogRenderInfo, ChangelogEntry } from 'beachball';
+import { ChangelogEntry, PackageChangelogRenderInfo } from 'beachball';
 import { getPrNumber, repoDetails } from './github-functions';
 
 const repoUrl = `https://github.com/${repoDetails.owner}/${repoDetails.repo}`;
+let features: ChangelogEntry[] = [];
```
JamesBurnside, line 5: Does these need to be mutable global variables? Can they be moved inside the `renderPackageChangelog` fn

JamesBurnside (approved):

common/config/beachball/changelog-custom-renders.ts
JoshuaLai: Ah this is a good catch. Will update it now.

palatter (approved):
LG

PR #3372, opened 2023-07-25
Storybook, Composites and Sample App references to `react-components-v2`

JamesBurnside:
# What

Add missing references to v2 packlet

# Why

For local development

# How Tested

..


packages/calling-component-bindings/package.json
```diff
@@ -33,6 +33,7 @@
     "@internal/acs-ui-common": "2.0.0-beta.0",
     "@internal/calling-stateful-client": "2.0.0-beta.0",
     "@internal/react-components": "2.0.0-beta.0",
+    "@internal/react-components-v2": "2.0.0-beta.0",
```
vhuseinova-msft, line 36: Should it be "2.0.0-beta.1"?

vhuseinova-msft (approved):

packages/calling-component-bindings/package.json
JamesBurnside: Good question! The reason it's `.0` is the at release time we bump the version, so at release time these will be bumped to `.1` and the first release will be 2.0.0-beta.1

emlynmac (approved):

PR #3376, opened 2023-07-26
v9 example template component

JamesBurnside:
# What

Add an example component and update the necessary deps in the package for development

# Why

For use as a template and drive folder structure discussions

# How Tested

Unit tests added


emlynmac (approved):
LGTM; will add this to the docs too.

vhuseinova-msft (approved):

PR #3380, opened 2023-07-27
[Chat] Add ImageGallery as a new UI component

Leah-Xia-Microsoft:
# What

Created a new UI component called ImageGallery

# Why

Light theme:
Screenshot 2023-07-27 at 11 01 29

Dark theme:
Screenshot 2023-07-27 at 11 04 20

# How Tested

# Process & policy checklist


packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Props for {@link ImageGallery}.
+ *
+ * @beta
+ */
+export interface ImageGalleryImageProps {
+  imageUrl: string;
+  downloadFileName: string;
+  altText?: string;
```
jpeng-ms, line 74: what would alt text be normally? the image title?

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const downloadButtonTitleString = 'Download';
+  const closeString = 'Close';
+  return (
+    <Layer hostId={modalLayerHostId}>
+      <Modal
+        titleAriaId={title}
+        isOpen={images.length > 0}
+        onDismiss={onDismiss}
+        isBlocking={false}
```
jpeng-ms, line 115: sorry what's isBlocking and why it's set to false?

Leah-Xia-Microsoft: > sorry what's isBlocking and why it's set to false?

It is for determining whether the dialog can be light dismissed by clicking outside the dialog (on the overlay).
I just removed it since it's defaulted to false

packages/react-components/src/components/ImageGallery.tsx
Leah-Xia-Microsoft: > what would alt text be normally? the image title?

Yes I think it's quite common to use the image title as altText. However, in our case, the image title will be the sender's display name and we don't want to use that as the altText

jpeng-ms: i see, so we won't be setting any alt text then?

Leah-Xia-Microsoft: > i see, so we won't be setting any alt text then?

Good point. Teams web app set it to "image" for inline images. I'll follow that behaviour.

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/**
+ * Props for {@link ImageGallery}.
+ *
+ * @beta
+ */
+export interface ImageGalleryImageProps {
+  imageUrl: string;
+  downloadFileName: string;
```
jimchou-dev, line 77: Can this just be called fileName? (same with other references for `downloadFileName`)

Leah-Xia-Microsoft: > Can this just be called fileName? (same with other references for `downloadFileName`)

Would it be less clear to Contosos what fileName is used for?

JoshuaLai: This is the file name developers will define for what the saved file should be right? Maybe we can name it as much `saveasFilename` or something along those lines

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -0,0 +1,171 @@
...
+  titleIcon?: JSX.Element;
+}
+
+/**
+ * Props for {@link ImageGallery}.
+ *
+ * @beta
+ */
+export interface ImageGalleryProps {
+  images: Array<ImageGalleryImageProps>;
```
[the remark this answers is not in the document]
JoshuaLai: Also at least if when developers do pass in an array, maybe they don't want to start on the first index, this will give them that flexibility

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -0,0 +1,171 @@
...
+export const ImageGallery = (props: ImageGalleryProps): JSX.Element => {
+  const { images, modalLayerHostId, onImageDownloadButtonClicked, onDismiss, styles } = props;
+  const theme = useTheme();
+  const isDarkTheme = isDarkThemed(theme);
+
+  const downloadButtonTitleString = 'Download';
+  const closeString = 'Close';
+  const defaultAltText = 'image';
+
+  if (images.length <= 0) {
```
JoshuaLai, line 115: Does it make sense to throw an error in this case?

Leah-Xia-Microsoft: > Does it make sense to throw an error in this case?

It made sense to return an empty div when there's no images in the array. However, now that we switched to images.length <= startIndex, seems like we should find a way to throw an error to Contoso developer but not breaking the UI for end user?

packages/react-components/src/components/styles/ImageGallery.style.ts
```diff
@@ -0,0 +1,185 @@
...
+ */
+export const downloadIcon: IIconProps = {
+  iconName: 'Download'
+};
+
+/**
+ * @private
+ */
+export const overlayStyles = {
+  root: { background: 'rgba(0, 0, 0, 0.85)' }
```
dmceachernmsft, line 22: nit: I would do this with the `opacity` CSS property, if you need the colour lets try to do that with the theme like you do below.

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -0,0 +1,178 @@
...
+  image?: IStyle;
+}
+
+/**
+ * Props for {@link ImageGallery}.
+ *
+ * @beta
+ */
+export interface ImageGalleryImageProps {
+  imageUrl: string;
```
dmceachernmsft, line 78: nit can we add TSDocs for these props?

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/**
+ * Props for {@link ImageGallery}.
+ *
+ * @beta
+ */
+export interface ImageGalleryProps {
+  images: Array<ImageGalleryImageProps>;
+  onDismiss: () => void;
```
dmceachernmsft, line 92: Nit: like above please add TSDocs for all these props

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -0,0 +1,178 @@
...
+   * @Example
+   * ```
+   * <ImageGallery styles={{ image: { background: 'blue' } }} />
+   * ```
+   */
+  styles?: ImageGalleryStylesProps;
+}
+
+/**
+ * @beta
```
dmceachernmsft, line 110: Please add a TSDoc comment here for what the imageGallery component is

packages/react-components/src/components/styles/ImageGallery.style.ts
```diff
@@ -0,0 +1,185 @@
...
+};
+
+/**
+ * @private
+ */
+export const focusTrapZoneStyle = mergeStyles({
```
dmceachernmsft, line 28: Instead of using `mergeStyles` everywhere can we strongly type these styles so they aren't misused elsewhere? Fluent provides style types for their components like: `IFocusTrapZoneStyles` or `IStackStyles`

jimchou-dev (approved):

PR #3387, opened 2023-07-31
[v2] Update API-extractor

emlynmac:
# What
Updates API extractor dependency to latest

# Why
Ongoing upkeep

# How Tested
Build / CI

# Process & policy checklist


packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -233,7 +233,7 @@ export abstract class ProxyCallAgentCommon {
        * 2. Get the incoming call from the new state and it's ID
        * 3. Use `callAgent.incomingCalls` and filter an incoming call ID to get a declarative incoming call object
        */
-      case 'incomingCalls': {
+      case 'incomingCalls' as any: {
```
JamesBurnside, line 236: what was the reason this as any was needed?

JamesBurnside (approved):

packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
emlynmac: To fix a TypeScript type error.

JamesBurnside: Lol, more asking why an update to api-extractor caused a new typescript error. And generally we want to avoid as any but relize those common interfaces are sometimes don't work too well and need it

emlynmac: I don't think this update caused the error, but there was a complaint from the linter in this PR.

vhuseinova-msft (approved):

jimchou-dev (approved):

PR #3392, opened 2023-08-01
Add localization to v2 packlet

JamesBurnside:
# What

Add localization infra:

- locale provider
- userLocale hook
- example component updated to have usage

# Why

Localization support

# How Tested
n/a


change/@azure-communication-react-74a88979-2128-4bde-8f4f-5c366d44e9f4.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
emlynmac, line 2: Minor or none?

vhuseinova-msft (approved):

packages/react-components-v2/src/ExampleComponent/index.ts
```diff
@@ -2,4 +2,5 @@
 // Licensed under the MIT license.
 
 export * from './ExampleComponent';
+export * from './ExampleComponent.strings';
 export * from './ExampleComponent.types';
```
emlynmac, line 6: Should we split types out into a separate index.d.ts?

change/@azure-communication-react-74a88979-2128-4bde-8f4f-5c366d44e9f4.json
JamesBurnside: Yeah should be none

packages/react-components-v2/src/localization/LocalizationProvider/index.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export * from './LocalizationProvider';
+export * from './LocalizationProvider.types';
```
emlynmac, line 5: Do we want to break out types into .d.ts file?
ApiExtractor?

emlynmac (approved):
If the api-extractor is not throwing errors then we're good to go

PR #3401, opened 2023-08-02
Intermediate fix for capabilities bug from calling beta

mgamis-msft:
# What
Intermediate fix for bug from calling beta where capabilities in interop is mostly 'NotInitialized' as shown below for ACS users:
image
This is fixed in local tgz builds being provided to me by Calling SDK engineer but not in their latest beta.

To fix this, we are falling back to using the role to infer if buttons should be enabled or disabled

# Why
The buttons in the control bar are disabled for interop scenario
image

# How Tested
Tested locally:
https://github.com/Azure/communication-ui-library/assets/79475487/68962b6c-7dd8-4377-938b-5ea59fb419ca

# Process & policy checklist


packages/calling-component-bindings/src/callControlSelectors.ts
```diff
@@ -41,11 +41,25 @@ export type MicrophoneButtonSelector = (
...
+    getDeviceManager,
+    /* @conditional-compile-remove(capabilities) */ getCapabilites,
+    /* @conditional-compile-remove(capabilities) */ getRole
+  ],
+  (
+    callExists,
+    isMuted,
+    deviceManager,
+    /* @conditional-compile-remove(capabilities) */ capabilities,
+    /* @conditional-compile-remove(capabilities) */ role
```
dmceachernmsft, line 56: should this be CC'd with Rooms?

dmceachernmsft (approved):

packages/calling-component-bindings/src/callControlSelectors.ts
mgamis-msft: It should be for capabilities since role is provided in Teams interop scenario. Also when I stabilize the capabilities feature, I want this code to be stabilized too (given that it is still needed when calling sdk fixes this interop bug)

prabhjot-msft (approved):

PR #3406, opened 2023-08-03
[Chat] Open image gallery when tap on an inline image

Leah-Xia-Microsoft:
# What

- Added the functionality to open ImageGallery when an inline image is clicked
- Dismiss ImageGallery
- Added e2e tests

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -208,6 +220,59 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
     [adapter]
   );
 
+  /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+  const onInlineImageClicked = useCallback(
+    async (attachment: FileMetadata, imageName?: string, senderId?: string): Promise<void> => {
```
JoshuaLai, line 225: I don't think we need to pass the `senderId` into this method. We only use it internally for the `AvatarPersona` is it possible to just reuse the `OnRenderAvatarCallback`?

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -159,6 +164,10 @@ export const ImageGallery = (props: ImageGalleryProps): JSX.Element => {
             src={image.imageUrl}
             className={mergeStyles(imageStyle, styles?.image)}
             alt={image.altText || defaultAltText}
+            onError={() => {
+              console.log('Unable to load Image');
```
JoshuaLai, line 168: Might not be a bad idea to have an `onError` in the props so we can send an error back up. Just a thought

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
Leah-Xia-Microsoft: > I don't think we need to pass the `senderId` into this method. We only use it internally for the `AvatarPersona` is it possible to just reuse the `OnRenderAvatarCallback`?

The OnRenderAvatarCallback also takes a `userId` prop which is the sender's id. We are at the Chat composite level and we don't know anything about the message that's been clicked (including the sender) unless it's passed in with the onInlineImageClicked callback

packages/react-components/src/components/ImageGallery.tsx
Leah-Xia-Microsoft: > Might not be a bad idea to have an `onError` in the props so we can send an error back up. Just a thought

I like this idea. Will update the API

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -208,6 +220,59 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
+          {
+            ...galleryImage,
+            imageUrl: fullSizeAttachments[attachment.id]
+          }
+        ]);
+        return;
+      }
+
+      if (attachment.attachmentType === 'inlineImage' && attachment.url) {
+        const blob = await adapter.downloadAttachments({ attachmentUrls: [attachment.url] });
```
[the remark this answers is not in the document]
Leah-Xia-Microsoft: > does it mean we download the image every time when user opens the gallery? do we want to cache it?

We do cache it with a map called `fullSizeAttachments`

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
```diff
@@ -91,6 +91,12 @@ type ChatMessageComponentProps = {
    */
   onFetchAttachments?: (attachment: FileMetadata) => Promise<void>;
   /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+  /**
+   * Optional callback called when an inline image is clicked.
+   * @beta
+   */
+  onInlineImageClicked?: (attachment: FileMetadata, imageName?: string, senderId?: string) => Promise<void>;
```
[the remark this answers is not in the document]
Leah-Xia-Microsoft: > does contoso need to pass in alt text in this for accessibility?

If Contoso is only using MessageThread as a component, they'll create their own ImageGallery (if they decide to use it) and pass altText there)
If Contoso is using our Chat/CallWithChat composite, I don't think we need to provide that much flexibility to set the alt text. It will be defaulted to `image`.

change-beta/@azure-communication-react-ccb6fbb5-9b88-462c-9be8-a99ad73b17b4.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "none",
+  "area": "feature",
+  "workstream": "ImageGallery",
+  "comment": "Adds functionality to open image gallery when tap on an inline image",
```
JoshuaLai, line 5:
```suggestion
  "comment": "Added functionality to open image gallery when tapping on an inline image",
```

change/@azure-communication-react-ccb6fbb5-9b88-462c-9be8-a99ad73b17b4.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "none",
+  "area": "feature",
+  "workstream": "ImageGallery",
+  "comment": "Adds functionality to open image gallery when tap on an inline image",
```
JoshuaLai, line 5:
```suggestion
  "comment": "Added functionality to open image gallery when tapping on an inline image",
```

packages/react-components/src/components/styles/ImageGallery.style.ts
```diff
@@ -18,10 +18,10 @@ export const downloadIcon: IIconProps = {
 /**
  * @private
  */
-export const overlayStyles = (theme: Theme): IOverlayStyles => {
+export const overlayStyles = (theme: Theme, isDarkThemed: boolean): IOverlayStyles => {
   return {
     root: {
-      background: theme.palette.white,
+      background: isDarkThemed ? theme.palette.white : theme.palette.black,
```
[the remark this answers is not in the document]
Leah-Xia-Microsoft: > does it make sense to have `isDarkThemed` be a property in `Theme`

Theme is a fluentUI type so we should not add isDarkThemed as a property

change-beta/@azure-communication-react-ccb6fbb5-9b88-462c-9be8-a99ad73b17b4.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "minor",
+  "area": "feature",
+  "workstream": "ImageGallery",
+  "comment": "Add functionality to open image gallery when user taps on an inline image",
```
jpeng-ms, line 5: sorry just wanted to double check, we are only showing 1 image per gallery right for this release? Are you going to support multiple images in the gallery later on (similar to teams behaviour)?

Leah-Xia-Microsoft: > sorry just wanted to double check, we are only showing 1 image per gallery right for this release? Are we going to support multiple images in the gallery later on (similar to teams behaviour)?

That's correct. We're supporting one image per gallery for Public Preview. We'll have multiple images support if there's customer asks.

packages/react-components/src/components/styles/ImageGallery.style.ts
```diff
@@ -120,12 +124,21 @@ export const imageContainer: IStyle = {
...
   height: '100%',
   width: '100%'
 };
 
+/**
+ * @private
+ */
+export const brokenImageStyle = (theme: Theme, isDarkThemed: boolean): IStyle => {
+  return {
+    color: isDarkThemed ? theme.palette.black : theme.palette.white
```
JoshuaLai, line 138: Let leave the same comment here about the theme.

JoshuaLai (approved):

JoshuaLai:
Also lets remember to note that we still want the tests for themes in an upcoming PR.

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -159,6 +168,11 @@ export const ImageGallery = (props: ImageGalleryProps): JSX.Element => {
             src={image.imageUrl}
             className={mergeStyles(imageStyle, styles?.image)}
             alt={image.altText || defaultAltText}
+            onError={(event) => {
+              console.log('Unable to load Image');
```
PorterNan, line 172: Remove log from PR

jpeng-ms (approved):
Looks good!

PR #3411, opened 2023-08-03
[Custom Layouts] Movable overflow gallery

dmceachernmsft:
# What

Introduces new controls to move the overflow gallery to the top of the video gallery 
image

# Why
Allows the user to be more in control of how the video gallery is presented
We want to enable the ability for the end user to be able to move the overflow gallery around.

https://skype.visualstudio.com/SPOOL/_workitems/edit/3362677
https://skype.visualstudio.com/SPOOL/_workitems/edit/3362659
https://skype.visualstudio.com/SPOOL/_workitems/edit/3362656
# How Tested

tested locally


edwardlee-msft:
Looking great!  @dmceachernmsft 
There is only one issue that I see, there is probably some missing styling that needs to be applied to have the button aligned and filled:
image

packages/react-composites/src/composites/common/ControlBar/CommonCallControlBar.tsx
```diff
@@ -336,6 +338,7 @@ export const CommonCallControlBar = (props: CommonCallControlBarProps & Containe
                             isCaptionsSupported={props.isCaptionsSupported}
                             /* @conditional-compile-remove(close-captions) */
                             onCaptionsSettingsClick={openCaptionsSettingsModal}
+                            onSetUserSetOverflowGalleryPosition={props.onSetUserSetOverflowGalleryPosition}
```
mgamis-msft, line 341: I guess we are not allowing top gallery position for mobile view?

mgamis-msft (approved):

mgamis-msft:
Looks great!

packages/react-composites/src/composites/common/ControlBar/CommonCallControlBar.tsx
dmceachernmsft: yes is only supported for desktop, the only views that will be for mobile are "gallery" (default) "large count gallery" and "dynamic" floating local video

packages/react-composites/src/composites/CallComposite/pages/CallPage.tsx
```diff
@@ -119,6 +128,8 @@ export const CallPage = (props: CallPageProps): JSX.Element => {
       dataUiId={'call-page'}
       latestErrors={props.latestErrors}
       onDismissError={props.onDismissError}
+      /* @conditional-compile-remove(gallery-layouts) */
+      onSetUserSetOverflowGalleryPosition={setUserSetOverflowGalleryPosition}
```
JamesBurnside, line 132: `onSetUserSet`? 
Not convinced on userSet but if its staying, update this to `onUserSetOverflowGalleryPositionChange` to avoid the double set

dmceachernmsft: (naming was never my strong suit!)

PR #3413, opened 2023-08-04
[Chat] Remove northstar dependency

emlynmac:
# What
Remove the northstar fluent wrapper and use fluent 8/9 components in place.

# Why
Allow full react 18 support.
Supersedes #3200 

# How Tested
Storybook

# Process & policy checklist


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -1275,8 +1209,6 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
       messages,
       showMessageDate,
       showMessageStatus,
-      onRenderAvatar,
-      isNarrow,
```
emlynmac, line 1279: The isNarrow adjustments of the avatar overlap have been lost; need to get this back into the rendering method of the chatMessage from f9

PR #3415, opened 2023-08-08
[Chat]Add unit tests for Image Gallery component

Leah-Xia-Microsoft:
# What

Add unit tests for Image Gallery component

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/ImageGallery.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+describe('ImageGallery default layout tests', () => {
+  beforeAll(() => {
+    registerIcons({
+      icons: {
+        cancel: <></>,
+        download: <></>
+      }
+    });
+  });
+  const titleIconTestId = 'titleIconTestId';
```
jimchou-dev, line 18: Does this id make sense to be used for both `imageInfo` and `imageInfo2` would it be better for them to be different so you can compare which one you are comparing against?

jimchou-dev (approved):

JoshuaLai (approved):

PR #3417, opened 2023-08-09
Update stable Call and CallWithChat composites to use the same control bar

mgamis-msft:
# What
Removed conditional compilation of common control bar in Call and CallWithChat composites so that both composites use the new control bar for stable

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3370372

# How Tested
Local Calling sample:
image
Local CallWithChat sample:
image

# Process & policy checklist


JamesBurnside:
This doesn't look great, can you ensure there's a bug that blocks stable release until this is fixed:
image

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -459,6 +466,7 @@ export type CallControlOptions = (CommonCallControlOptions & {
     participantsButton?: boolean | {
         disabled: boolean;
     };
+    legacyControlBarExperience?: boolean;
```
JamesBurnside, line 469: Do we have UI tests for this?

JamesBurnside (approved):

packages/communication-react/review/stable/communication-react.api.md
mgamis-msft: No but will add that.

mgamis-msft: Edit: My mistake. All previous tests in CallControls.test.ts use the legacyControlBarExperience so we are still taking snapshots of the old control bar. The new snapshots  with the new control bar in this PR were generated because the tests with legacyControlBarExperience equalling false were added to stable (ie. the conditional compilation statement was removed in the test.)

mgamis-msft:
> This doesn't look great, can you ensure there's a bug that blocks stable release until this is fixed (is vertical gallery stable?): image

Sure. I created this bug to verify if the control position is correct for mobile landscape for the September stable release https://skype.visualstudio.com/SPOOL/_workitems/edit/3377341. FYI: this is what the control bar looks like for CallWithChat currently as verified by this UI snapshot: https://github.com/Azure/communication-ui-library/blob/main/packages/react-composites/tests/browser/snapshots/stable/tests/browser/callwithchat/hermetic/CallControls.test.ts-snapshots/user-set-control-bar-button-options-Mobile-Android-Landscape-linux.png

edwardlee-msft (approved):

PR #3419, opened 2023-08-09
[BugFix] local tile bye bye

dmceachernmsft:
# What

Fixes issue where turning off your screenshare stream disposes of local render info removing the stream from the stateful client but not disposing the stream in the call. now we only delete the local render info if the local steam being updated is video
# Why

Big privacy issue!!

https://skype.visualstudio.com/SPOOL/_workitems/edit/3360686
# How Tested

Locally see video
https://github.com/Azure/communication-ui-library/assets/94866715/9d5dad98-dc45-4465-9ff1-c25f0addf9e0


mgamis-msft (approved):

change-beta/@azure-communication-react-a10e1ac6-f68d-4b21-b2d5-83afb5c7fa04.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "patch",
+  "area": "fix",
+  "workstream": "Fixes issue where when local participant stops screen share their local stream would disappear but still be showing to others in the call",
+  "comment": "fix found?",
```
mgamis-msft, line 5: nit: I think what you entered for workstream should be entered for comment so that we capture this bug fix comment in the changelog

dmceachernmsft: oops fixing

packages/calling-stateful-client/src/CallSubscriber.ts
```diff
@@ -316,6 +316,9 @@ export class CallSubscriber {
         undefined,
         convertSdkLocalStreamToDeclarativeLocalStream(event.removed[0])
       );
+      if (event.removed[0].mediaStreamType === 'ScreenSharing') {
```
JamesBurnside, line 319: Can you add a comment about this line and that we need to support local screen share streams

JamesBurnside (approved):

packages/calling-stateful-client/src/CallSubscriber.ts
dmceachernmsft: will do!

PR #3425, opened 2023-08-10
Add "shake to send logs" to sample apps

JamesBurnside:
# What

- Add shake detection in sample apps using shake.js
- On iOS prompt for permissions to detect shakes
- On shake the prompt asks to upload logs
- Afterwards asks if they wish to email the log link

# Why

Quickly get logs from mobile device when a bad issue repros

# How Tested
| | | |
| -- | -- | -- | 
| image |image | image |


samples/Calling/src/app/utils/ShakeToSendLogs.tsx
```diff
@@ -0,0 +1,238 @@
...
+              Your logs were successfully uploaded. You can access them{' '}
+              <Link href={blobUrl} target="_blank">
+                here
+              </Link>
+              .
+            </Text>
+            <DialogFooter>
+              <PrimaryButton
+                text="Send as email"
+                onClick={() => window.open(`mailto:?body=${encodeURIComponent(blobUrl ?? '')}`)}
```
dmceachernmsft, line 203: is this email something that we provide at send time? or something we setup into the server?

samples/Calling/src/app/utils/ShakeToSendLogs.tsx
```diff
@@ -0,0 +1,238 @@
...
+const originalConsoleLog = console.log;
+const originalAzureLoggerLog = AzureLogger.log;
+const consoleLogs: string[] = ['---- LOGS START ----'];
+
+/**
+ * Track console logs for pushing to a debug location.
+ * This is particularly useful on mobile devices where the console is not easily accessible.
+ */
+const startRecordingLogs = (): void => {
+  console.log = (...args: unknown[]) => {
```
dmceachernmsft, line 19: do these assignments mean that any console logs inside the application and lib that we put will get trapped here?

dmceachernmsft (approved):

samples/Calling/src/app/utils/ShakeToSendLogs.tsx
JamesBurnside: its a standard protocol that just opens your default mail client with the body attached! So doesn't send the email itself just opens a client for you to send it

samples/Calling/src/app/utils/ShakeToSendLogs.tsx
JamesBurnside: All console logs that use console.log across the whole `window` will get caught by this function. This function does call the original console.log so they will be outputted to the console, but we just also will store them in the logs array

emlynmac (approved):

PR #3427, opened 2023-08-10
Remove Fluent NorthStar

emlynmac:
# What
Remove deprecated fluent N* library and replace with fluent 9 controls.

# Why
NorthStar is not fully compatible with React 18

# How Tested
CI, storybook

# Process & policy checklist


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -91,7 +92,8 @@ import { SyntheticEvent } from 'react';
 import { TeamsCall } from '@azure/communication-calling';
 import { TeamsCallAgent } from '@azure/communication-calling';
 import { TeamsMeetingLinkLocator } from '@azure/communication-calling';
-import { Theme } from '@fluentui/react';
+import { Theme as Theme_2 } from '@fluentui/react';
```
vhuseinova-msft, line 95: I'm wondering if it's possible to improve the types in this file?

packages/react-components/package.json
```diff
@@ -35,12 +35,15 @@
   },
   "dependencies": {
     "@fluentui/react": "~8.110.4",
+    "@fluentui/react-components": "~9.27.4",
     "@fluentui/react-file-type-icons": "8.8.13",
     "@fluentui/react-hooks": "~8.6.12",
     "@fluentui/react-icons": "~2.0.194",
+    "@fluentui/react-migration-v0-v9": "~9.0.17",
```
vhuseinova-msft, line 42: Do we still need these 2 dependencies?

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -293,39 +308,119 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     handleOnInlineImageClicked
   ]);
 
+  const attached = message.attached === 'top' ? 'top' : message.attached === false ? 'top' : 'center';
```
vhuseinova-msft, line 311: can be updated to
```suggestion
  const attached = message.attached === 'top' || message.attached === false ? 'top' : 'center';
```

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -293,39 +308,119 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
...
+                messageContainerClasses?.body
+              )
+            }}
+            data-ui-id="chat-composite-message"
+            onTouchStart={() => setWasInteractionByTouch(true)}
+            onPointerDown={() => setWasInteractionByTouch(false)}
+            onKeyDown={() => setWasInteractionByTouch(false)}
+            onBlur={() => setWasInteractionByTouch(false)}
+            onClick={() => {
+              if (!wasInteractionByTouch) {
```
vhuseinova-msft, line 402: `OnClick` can be a separate func to remove duplications

packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -118,6 +163,46 @@ export const FailedMyChatMessageContainer: ComponentSlotStyle = {
...
+  body: {
+    ...defaultChatItemMessageContainer,
+    backgroundColor: 'rgb(199, 224, 244)'
+  }
+});
+
+/**
+ * @private
+ */
+export const useChatMyMessageClasses: (messageState?: MessageStatus) => Record<'body', string> = (
```
vhuseinova-msft, line 179: As `useChatMyMessageClasses` and `useChatMessageClasses` do the same, should we reuse code instead?
same for `useChatFailedMyMessageClasses` and `useChatFailedMessageClasses`, `chatNormalMyMessageClasses` and `chatNormalMessageClasses`

vhuseinova-msft:
Not sure if it's expected

| Was | Now |
|------|------|
| image | image |

vhuseinova-msft:
It looks like delete button doesn't work in context menu for a message in the MessageThread

vhuseinova-msft:
Message with a Mention is broken
image

packages/communication-react/review/beta/communication-react.api.md
emlynmac: It's auto-generated, so maybe in the future. There are lots of people on the api-extractor github asking questions about this.

packages/react-components/src/components/styles/MessageThread.styles.ts
emlynmac: I moved the original logic that was in each of these out to the theme; these do indeed need to be refactored as a result.

PR #3432, opened 2023-08-11
[BugFix, TotalParticipantCount] Fix participant count when alone, fix sample to allow joining existing call

dmceachernmsft:
# What

Fixes issue where if you're alone in a call the count has the unformatted string
Fixes issue where you could not join an existing call in the Calling sample
# Why

image

https://skype.visualstudio.com/SPOOL/_workitems/edit/3373830
# How Tested

Tested locally, added e2e tests for scenario


packages/react-composites/src/composites/common/ParticipantContainer.tsx
```diff
@@ -56,7 +56,9 @@ export const ParticipantListWithHeading = (props: {
   const subheadingUniqueId = useId();
   const theme = useTheme();
   /* @conditional-compile-remove(total-participant-count) */
-  const totalParticipantCount = participantListProps.totalParticipantCount;
+  const totalParticipantCount = participantListProps.totalParticipantCount
+    ? participantListProps.totalParticipantCount
+    : 1;
```
JamesBurnside, line 61: Why is `totalParticipantCount` undefined if you're the only one in the call? Shouldn't it correctly be 1?

carocao-msft (approved):

packages/react-composites/src/composites/common/ParticipantContainer.tsx
JamesBurnside: Another Q - In `paneTitleTrampoline` below, what happens on stable? Presumably the string is still `In this call {numberOfPeople}` so if we conditionally compile the `formatString` call won't we see in stable builds `In this call {numberOfPeople}`?

dmceachernmsft: it comes back from the sdk as undefined when you're alone for some reason, not totally sure why.

dmceachernmsft: great question, I believe when I originally did it it was not there as expected but I will take look

dmceachernmsft: oh right in stable we are still using the old control bar so it is not used at all 
image

JamesBurnside: Oh this will be impacted by this active PR then! #3408

JamesBurnside: Actually callwithchat stable already should use the people pane too

JamesBurnside: > it comes back from the sdk as undefined when you're alone for some reason, not totally sure why.

Sounds like a bug :L

dmceachernmsft: image
yeah seems like it

dmceachernmsft: > Actually callwithchat stable already should use the people pane too

good point will see what happens there

dmceachernmsft: yup was broken, I think I have a better fix coming

prabhjot-msft (approved):

JamesBurnside (approved):

PR #3434, opened 2023-08-11
Upgrade calling beta to 1.16.2-beta.1 and stable to 1.15.2

mgamis-msft:
# What
There are many changes related to captions and capabilities in the beta and stable releases of Calling SDK

# Why
Latest and greatest from Calling SDK

# How Tested
Local calling sample

# Process & policy checklist


packages/react-components/src/components/VideoGallery/FloatingLocalVideoLayout.tsx
```diff
@@ -147,9 +147,13 @@ export const FloatingLocalVideoLayout = (props: FloatingLocalVideoLayoutProps):
         ? SHORT_VERTICAL_GALLERY_FLOATING_MODAL_SIZE_REM
         : VERTICAL_GALLERY_FLOATING_MODAL_SIZE_REM;
     }
-    /*@conditional-compile-remove(click-to-call) */
-    if ((overflowGalleryTiles.length > 0 || screenShareComponent) && overflowGalleryPosition === 'HorizontalBottom') {
-      return localVideoTileSize === '16:9' || !isNarrow ? LARGE_FLOATING_MODAL_SIZE_REM : SMALL_FLOATING_MODAL_SIZE_REM;
+    /*@conditional-compile-remove(click-to-call) */ {
+      /* @conditional-compile-remove(vertical-gallery) */
```
dmceachernmsft, line 151: Why did we add this CC statement here?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -997,20 +1003,23 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
 
   /* @conditional-compile-remove(close-captions) */
   private subscribeToCaptionEvents(): void {
-    if (this.call && this.call.state === 'Connected' && _isTeamsMeetingCall(this.call)) {
-      this.call?.feature(Features.TeamsCaptions).on('captionsReceived', this.captionsReceived.bind(this));
-      this.call?.feature(Features.TeamsCaptions).on('isCaptionsActiveChanged', this.isCaptionsActiveChanged.bind(this));
-      this.call?.off('stateChanged', this.subscribeToCaptionEvents.bind(this));
+    if (this.call && this.call.state === 'Connected' && this.context.getState().isTeamsCall) {
+      const captionsFeature = this.call?.feature(Features.Captions) as unknown as TeamsCaptions;
```
dmceachernmsft, line 1007: Can we not do the subscription to the feature through the statefulClient and the CallSubscriber? @carocao-msft is this the intended way to do this?

packages/react-composites/src/composites/CallComposite/components/MediaGallery.tsx
```diff
@@ -157,9 +157,12 @@ export const MediaGallery = (props: MediaGalleryProps): JSX.Element => {
         remoteVideoTileMenuOptions={remoteVideoTileMenuOptions}
         /* @conditional-compile-remove(vertical-gallery) */
         overflowGalleryPosition={overflowGalleryPosition}
-        /* @conditional-compile-remove(rooms) */
+        /* @conditional-compile-remove(click-to-call) */
```
dmceachernmsft, line 160: I think all 3 features should be part of this API removal and whichever goes first releases the prop on the video gallery. this cc wont work if c2c doesn't go before the other two

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
carocao-msft: This is intended but this code change is for captions GA, it shouldn't be leaked here.

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
```diff
@@ -96,7 +98,7 @@ export interface CommonCallingHandlers {
 /**
  * options bag to start captions
  *
- * @beta
+ * @public
```
carocao-msft, line 101: There is a lot of caption related changes here, they are for captions GA and the corresponding calling version is going to be release on 8/21. The changes should not be here

mgamis-msft: Good catch. Will change this back to beta

PR #3436, opened 2023-08-11
Fix for calling sample remote participants missing from call

prabhjot-msft:
# What
Fix for calling sample remote participants missing from call

# Why

# How Tested

# Process & policy checklist


samples/Calling/src/app/App.tsx
```diff
@@ -117,7 +117,9 @@ const App = (): JSX.Element => {
             /* @conditional-compile-remove(PSTN-calls) */
             setAlternateCallerId(callDetails.alternateCallerId);
             let callLocator: CallAdapterLocator | undefined =
-              callDetails.callLocator || getTeamsLinkFromUrl() || getGroupIdFromUrl();
+              getTeamsLinkFromUrl() ||
+              getGroupIdFromUrl() ||
+              /* @conditional-compile-remove(rooms) */ callDetails.callLocator;
```
dmceachernmsft, line 122: Looks like this has already been fixed?

PR #3440, opened 2023-08-11
Fix nightly alpha build version deps

JamesBurnside:
# What

Run enforce build flavor to ensure the correct `||` dep is chosen

# Why

Missed in #3418


prabhjot-msft (approved):

JamesBurnside:
(added this week OCEs to PR)

.github/workflows/nightly-cd.yml
```diff
@@ -101,6 +101,10 @@ jobs:
       - name: Ensure all package versions are consistent
         run: rush ensure-consistent-versions
 
+      # Alpha version run latest
+      - name: Force build flavor to `beta`
```
emlynmac, line 105: Is this to mitigate somebody checking in the .env file set differently?

JamesBurnside: It's poorly named, it sets all the deps in the package.json. So where we have `@azure/communication-calling: 1.x.x-beta.1 || 1.y.y` it rewrites and outputs only `@azure/communication-calling: 1.x.x-beta.1`

emlynmac (approved):

PR #3453, opened 2023-08-15
Add check to make sure `DeviceOrientationEvent` exists before referencing it

JamesBurnside:
## What

Add check to make sure `DeviceOrientationEvent` exists before referencing it

## Why

Threw an error in safari


vhuseinova-msft:
tested on MacOS Safari and Edge, everything works!

vhuseinova-msft (approved):

PR #3461, opened 2023-08-16
[Chat] Update onInlineImageClicked API

Leah-Xia-Microsoft:
# What

As discussed in the internal review, we'll change the onInlineImageClicked into
`onInlineImageClicked?: (attachmentId: string, messageId: string) => Promise<void>;`

Change the default initialsTextColor for titleIcon to white
# Why

# How Tested

# Process & policy checklist


JoshuaLai:
It looks like the avatars in the top changed style slightly in the recordings. Probably because we are now calling RenderAvatar callback.

Leah-Xia-Microsoft:
> It looks like the avatars in the top changed style slightly in the recordings. Probably because we are now calling RenderAvatar callback.

Yes. I've updated the description of the PR as well to include the avatar color changes.
The RenderAvatar default initialsTextColor to white, which we will follow here as well.
So the snapshots update is expected.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -221,11 +224,35 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
-      attachment: FileMetadata,
-      onRenderTitleIcon?: (personaProps?: IPersonaProps) => JSX.Element
-    ): Promise<void> => {
-      const titleIcon = onRenderTitleIcon && onRenderTitleIcon();
+    async (attachmentId: string, messageId: string): Promise<void> => {
+      const messages = messageThreadProps.messages?.filter((message) => {
+        return message.messageId === messageId;
+      });
+      if (!messages || messages.length <= 0) {
+        console.log(`Message not found with messageId ${messageId}`);
```
jimchou-dev, line 232: nitpick: we probably don't want to have these console.logs (similar for other console.logs in this PR)

packages/storybook/stories/MessageThread/MessageThread.stories.tsx
```diff
@@ -350,10 +350,32 @@ const MessageThreadStory = (args): JSX.Element => {
...
+
+    if (!attachments || attachments.length <= 0) {
+      return Promise.reject(`Attachment not found with id ${attachmentId}`);
+    }
+
+    const attachment = attachments[0];
+    attachment.name = chatMessage.senderDisplayName || '';
     const title = 'Message Thread Image';
+    const titleIcon = (
+      <Persona text={chatMessage.senderDisplayName} size={PersonaSize.size32} hidePersonaDetails={true} />
```
jimchou-dev, line 374: nitpick: break into multiple lines for readability

jimchou-dev (approved):

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
Leah-Xia-Microsoft: > nitpick: we probably don't want to have these console.logs (similar for other console.logs in this PR)

We currently don't have a logging system in Chat to only show error messages to Contoso developers.
I've seen MessageThread using console.log to display errors.
I'll bring this in the retro to introduce a proper logging system but I think it is out of the scope of this feature.

packages/storybook/stories/MessageThread/MessageThread.stories.tsx
Leah-Xia-Microsoft: > nitpick: break into multiple lines for readability

Prettifier automatically set it back to one line. It might be at the breaking point according to our linter rule.

JoshuaLai (approved):
I think it may also be good in the future to include a snapshot / test of the gallery without the render avatar being passed in.

PR #3465, opened 2023-08-16
[Chat] Add new storybook page for image gallery

JoshuaLai:
# What
Introduce new page for Image Gallery
image

# Why
New documentation 

# Process & policy checklist


packages/storybook/stories/ImageGallery/snippets/ImageGallery.snippet.tsx
```diff
@@ -0,0 +1,36 @@
...
+      imageUrl: source
+    };
+    setGalleryImages([galleryImage]);
+  };
+
+  return (
+    <FluentThemeProvider>
+      <div style={{ width: '31.25rem' }}>
+        <img alt="image" src="images/inlineImageExample1.png" onClick={imgClickedHandler} />
+      </div>
```
Leah-Xia-Microsoft, line 24: I don't think we need to display a placeholder image to be clicked on.
I think we can directly display the ImageGallery component without needing to click on anything.

packages/storybook/stories/ImageGallery/ImageGallery.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return <div style={{ width: '31.25rem' }}></div>;
+};
+
+export const ImageGallery = ImageGalleryStory.bind({});
+
+export default {
+  id: `${COMPONENT_FOLDER_PREFIX}-imagegallery`,
+  title: `${COMPONENT_FOLDER_PREFIX}/Image Gallery`,
+  component: ImageGalleryComponent,
+  argTypes: {},
```
Leah-Xia-Microsoft, line 47: We probably need to add some controls here for images and startIndex?

packages/storybook/stories/ImageGallery/ImageGallery.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <Canvas mdxSource={ImageGalleryExampleText}>
+        <ImageGalleryExample />
+      </Canvas>
+    </>
+  );
+};
+
+const ImageGalleryStory = (): JSX.Element => {
+  return <div style={{ width: '31.25rem' }}></div>;
+};
```
Leah-Xia-Microsoft, line 39: The Preview tab displays a empty screen. Should we also have <ImageGallery /> inside the ImageGalleryStory?

packages/storybook/stories/ImageGallery/snippets/ImageGallery.snippet.tsx
Leah-Xia-Microsoft: > I'm not sure how this will work. The image gallery is a modal so it will take over the entire screen. We will need an onClick event to launch the gallery in some way or else it would be just shown all the time.

That's a good point. Do you think having a button of "Open Image Gallery" would be more clear in this case than having an clickable image?
Contoso might be confused that the placeholder image is the gallery itself since the convention of storybook page is to display the component directly.

JoshuaLai: Discussed offline

packages/storybook/stories/ImageGallery/ImageGallery.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const galleryImage: ImageGalleryImageProps = {
+    title: 'Image',
+    saveAsName: 'images/inlineImageExample1.png',
+    imageUrl: 'images/inlineImageExample1.png'
+  };
+  const galleryImages = [galleryImage];
+  return (
+    <div style={{ width: '31.25rem' }}>
+      <ImageGalleryComponent
+        isOpen={args.disabled}
```
Leah-Xia-Microsoft, line 49: Should this be enabled instead of disabled? And why don't we just use isOpen instead of giving a new name to it?

packages/storybook/stories/ImageGallery/ImageGallery.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <Description>Component will render a fullscreen modal using a set image source.</Description>
+      <Canvas mdxSource={ImageGalleryExampleText}>
+        <ImageGalleryExample />
+      </Canvas>
+    </>
+  );
+};
+
+const ImageGalleryStory = (args): JSX.Element => {
+  // const [galleryImages, setGalleryImages] = useState<Array<ImageGalleryImageProps> | undefined>(undefined);
```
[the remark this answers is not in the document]
JoshuaLai: Oh good catch

packages/storybook/stories/ImageGallery/ImageGallery.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+  component: ImageGalleryComponent,
+  argTypes: {
+    disabled: controlsToAdd.displayImageGallery,
+    // Hiding auto-generated controls
+    isOpen: hiddenControl,
+    images: hiddenControl,
+    onDismiss: hiddenControl,
+    onImageDownloadButtonClicked: hiddenControl,
+    onError: hiddenControl,
+    startIndex: hiddenControl
```
Leah-Xia-Microsoft, line 76: Should we also give controls to images and startIndex?
And can we use isOpen control rather than adding a disabled control?

packages/storybook/stories/ImageGallery/ImageGallery.stories.tsx
JoshuaLai: `args.disabled` is the argument passed in from storybook. So we are not using a new name

packages/storybook/stories/ImageGallery/snippets/ImageGallery.snippet.tsx
```diff
@@ -0,0 +1,37 @@
...
+import React, { useState } from 'react';
+
+export const ImageGalleryExample: () => JSX.Element = () => {
+  const [galleryImages, setGalleryImages] = useState<Array<ImageGalleryImageProps>>([]);
+  const imgClickedHandler = (event: React.MouseEvent<HTMLImageElement>): void => {
+    event.stopPropagation();
+
+    const img = event.currentTarget;
+    const source = img.src;
+    const title = 'Image';
```
Leah-Xia-Microsoft, line 11: Would be nice to add titleIcon to storybook as well?

packages/storybook/stories/ImageGallery/ImageGallery.stories.tsx
Leah-Xia-Microsoft: > `args.disabled` is the argument passed in from storybook. So we are not using a new name

I think the prop name in args are defined by us. It might be better to have isOpen={args.isOpen} or isOpen={args.displayImageGallery}
And correspondingly, in the argTypes section, we can have
  `argTypes: {
        isOpen: controlsToAdd.isOpen,
        ...
    }`
or
  `argTypes: {
        displayImageGallery: controlsToAdd.displayImageGallery,
        ...
    }`

packages/storybook/stories/ImageGallery/ImageGallery.stories.tsx
JoshuaLai: I'm not sure if we can give them the ability to pass in an image. We only load in images from our own images directory so users wouldn't be able load their own.

palatter:
When I navigate back from preview, it shows this overtop of the page?

image

packages/storybook/stories/controlsUtils.ts
```diff
@@ -240,6 +240,8 @@ export const controlsToAdd = {
   },
   disabled: { control: 'boolean', defaultValue: false, name: 'Disable component' },
   displayName: { control: 'text', defaultValue: 'John Smith', name: 'Display Name' },
+  displayTitleImageGallery: { control: 'text', defaultValue: 'Image', name: 'Image Gallery Title' },
+  displayImageGallery: { control: 'boolean', defaultValue: false, name: 'Display Image Gallery' },
```
palatter, line 244: Should the default value be 'false' here?

Leah-Xia-Microsoft (approved):
LGTM

palatter:
With the preview gone, looks good now!

palatter (approved):

PR #3468, opened 2023-08-16
Scaling mode of participants update when screen sharing is turned off by remote user

prabhjot-msft:
# What
Before: When user selects a rendering option and there is a screen share, the rendering switches back to default.

https://github.com/Azure/communication-ui-library/assets/97124699/9a2d26de-6ce3-4c6e-a116-17edd0da234b

After: This fixes the issue, there is a slight delay while rendering option is switched.

https://github.com/Azure/communication-ui-library/assets/97124699/7d5f5266-7aad-4b61-bf70-d2e3d4c893da

# Why

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -266,6 +266,9 @@ export interface RemoteVideoStreamState {
    * API. This can be undefined if the stream has not yet been rendered and defined after createView creates the view.
    */
   view?: VideoStreamRendererViewState;
+  /* @conditional-compile-remove(pinned-participants) */
+  /** sclaing mode selected by user */
```
JamesBurnside, line 270:
```suggestion
    /** scaling mode selected by user */
```

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -614,6 +614,9 @@ export class CallContext {
           if (stream && stream.view) {
             stream.view.scalingMode = scalingMode;
           }
+          if (stream) {
+            stream.selectedScalingMode = scalingMode;
```
JamesBurnside, line 618: This works really well, but we are now storing UI logic in the stateful client which I think we have to avoid (cc @PorterNan).
We might have to move tracking inside the video gallery, or (though this would leave react-component users in the dust) moving this state storage to adapter layer

JamesBurnside:
Removing self as I will be OOO for a bit after monday

packages/calling-stateful-client/src/CallContext.ts
PorterNan: Yup, looks it's impacting the behavior coming from sdk so moving this to adapter/composite would be cool, thanks for pointing this out!!

dmceachernmsft:
Removing self as I am OOO after tuesday, excited for this work though!

packages/calling-stateful-client/src/CallContext.ts
prabhjot-msft: I am not sure if we can move the tracking to the composite layer as we are providing this feature at both levels: Composites & Components(Video Gallery & Video Tile). So, if we store the changes in adapter layer and  contoso is building experience using components. The value of scaling mode would be incorrect for them

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -54,6 +55,7 @@ export const _RemoteVideoTile = React.memo(
     drawerMenuHostId?: string;
     onPinParticipant?: (userId: string) => void;
     onUnpinParticipant?: (userId: string) => void;
+    updateScalingModeForRemoteUser?: (userId: string, scalingMode: ViewScalingMode) => void;
```
PorterNan, line 58: Is this a public api change? I guess not

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -374,6 +375,24 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
   /* @conditional-compile-remove(pinned-participants) */
   const [pinnedParticipantsState, setPinnedParticipantsState] = React.useState<string[]>([]);
   /* @conditional-compile-remove(pinned-participants) */
+  const [selectedScalingModeState, setselectedScalingModeState] = React.useState<{ [key: string]: VideoStreamOptions }>(
```
PorterNan, line 378:
```suggestion
  const [selectedScalingModeState, setselectedScalingModeState] = React.useState<Record<string, VIdeoStreamOptions>}>(
```

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -374,6 +375,24 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
+    (remoteUserId: string, scalingMode: ViewScalingMode) => {
+      setselectedScalingModeState((current) => ({
+        ...current,
+        [remoteUserId]: {
+          scalingMode,
+          isMirrored: remoteVideoViewOptions?.isMirrored
+        }
+      }));
+    },
+    [remoteVideoViewOptions]
```
PorterNan, line 393:
```suggestion
    [remoteVideoViewOptions?.isMirrored]
```

So we don't depend on an object but a boolean value, so no updated will be needed most of time

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -569,19 +595,21 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
...
+      remoteVideoViewOptions,
+      pinnedParticipants,
       onCreateRemoteStreamView,
       onDisposeRemoteVideoStreamView,
-      remoteVideoViewOptions,
-      localParticipant,
       onRenderAvatar,
       showMuteIndicator,
       strings,
-      /* @conditional-compile-remove(pinned-participants) */ drawerMenuHostId,
```
PorterNan, line 579: Looks like we removed lots of CC comment, Is it on purpose?

packages/react-components/src/components/RemoteVideoTile.tsx
prabhjot-msft: No, just added it to remote video tile which is internal component

packages/react-components/src/components/VideoGallery/useVideoTileContextualMenuProps.ts
```diff
@@ -37,6 +38,7 @@ export const useVideoTileContextualMenuProps = (props: {
     isPinned,
     onPinParticipant,
     onUnpinParticipant,
+    updateScalingModeForRemoteUser,
```
PorterNan, line 41: onUpdateScalingMode

let's change all the callback name like this

PorterNan (approved):
Since there are not public api changes, I am good with it, but please lemme know if there is any api change then I need to come back and discuss

Also, left some minor comments to be resolved

mgamis-msft (approved):

PR #3469, opened 2023-08-16
[Cherry pick]Updating localized resource files 

vhuseinova-msft:
Cherry picked from commit 0dec467fa8ae588dbc358b9671af6ff058d3b1be (#3467)

# What

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/localization/locales/zh-CN/strings.json
+++ b/packages/react-composites/src/composites/localization/locales/zh-CN/strings.json
@@ -165,8 +165,8 @@
     "unableToResolveTenantMoreDetails": "请检查正确输入参与者 ID",
     "participantIdIsMalformedTitle": "参与者 ID 的格式不正确",
     "participantIdIsMalformedMoreDetails": "请检查参与者 ID 的格式正确",
-    "moreButtonGalleryControlLabel": "Gallery options",
-    "moreButtonGalleryPositionToggleLabel": "Move gallery to top"
+    "moreButtonGalleryControlLabel": "库选项",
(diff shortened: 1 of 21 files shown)


packages/react-composites/src/composites/localization/locales/zh-CN/strings.json
```diff
@@ -165,8 +165,8 @@
     "unableToResolveTenantMoreDetails": "请检查正确输入参与者 ID",
     "participantIdIsMalformedTitle": "参与者 ID 的格式不正确",
     "participantIdIsMalformedMoreDetails": "请检查参与者 ID 的格式正确",
-    "moreButtonGalleryControlLabel": "Gallery options",
-    "moreButtonGalleryPositionToggleLabel": "Move gallery to top"
+    "moreButtonGalleryControlLabel": "库选项",
```
PorterNan, line 168: Not related to this PR, this Simplified Chinese translation looks a bit strange while traditional Chinese looks more correct

@mgamis-msft do you know where I can send a request to correct this?

PorterNan (approved):

packages/react-composites/src/composites/localization/locales/zh-CN/strings.json
mgamis-msft: Sorry just saw this. But there are two ways documented here https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/25949/Localization?anchor=filing-an-issue-with-translated-strings

PR #3475, opened 2023-08-18
Capabilities change notification bar

mgamis-msft:
# What
Show a notification bar when capabilities change

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3360665

# How Tested
Tested on local CallWIthChat sample in a Teams interop call in video below. There is also another Teams user getting notifications in the same call for comparison.
https://github.com/Azure/communication-ui-library/assets/79475487/e1dd8242-dbca-4f39-bc2b-46fdb1ee0c79

# Process & policy checklist


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -614,6 +615,12 @@ export interface CallCompositeStrings {
     cameraOffBackgroundEffectWarningText?: string;
     cameraPermissionDenied: string;
     cameraTurnedOff: string;
+    capabilityShareScreenGrantedDueToRoleChangeToPresenter?: string;
+    capabilityShareScreenLostDueToRoleChangeToAttendee?: string;
+    capabilityTurnVideoOnGrantedDueToMeetingOption?: string;
+    capabilityTurnVideoOnLostDueToMeetingOption?: string;
```
JamesBurnside, line 621: nit: perhaps `meeting` -> `call` for these (and maybe option -> capability?)
```suggestion
    capabilityTurnVideoOnLostDueToCallCapability?: string;
```

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
```diff
@@ -381,6 +388,14 @@ export const CallArrangement = (props: CallArrangementProps): JSX.Element => {
                         />
                       </Stack>
                     )}
+                    {
+                      /* @conditional-compile-remove(capabilities) */
+                      props.capabilitiesNotificationBarProps && (
+                        <Stack styles={bannerNotificationStyles}>
+                          <CapabilitiesChangeNotificationBar {...props.capabilitiesNotificationBarProps} />
```
JamesBurnside, line 395: Concerned here that the notification tracking is stored within the CapabilitiesChangeNotificationBar. This might mean as we transition from lobby -> call if there was a notification and user dismissed it it might be retriggered? (or lost, depending on the tracking logic).

Could you test what happens if the capabilities change in the lobby and see if the CapabilitiesChangeNotificationBar still shows in the connected state, and as a different test stays dimissed in the CapabilitiesChangeNotificationBar if dismissed?

packages/react-composites/src/composites/CallComposite/components/CapabilitiesNotficationBar.tsx
```diff
@@ -0,0 +1,275 @@
...
+};
+
+/* @conditional-compile-remove(capabilities) */
+const activeNotificationsToShow = (
+  activeNotifications: CapabalityChangeNotification[],
+  dismissedNotifications: DismissedNotification[],
+  mountTimestamp?: Date
+): CapabalityChangeNotification[] => {
+  const dismissed: Map<ParticipantCapabilityName, DismissedNotification> = new Map();
+  for (const error of dismissedNotifications) {
```
JamesBurnside, line 206: why `error` here?

packages/react-composites/src/composites/CallComposite/components/CapabilitiesNotficationBar.tsx
```diff
@@ -0,0 +1,275 @@
...
+  });
+};
+
+/* @conditional-compile-remove(capabilities) */
+const dismissNotification = (
+  dismissedNotification: DismissedNotification[],
+  toDismiss: CapabalityChangeNotification
+): DismissedNotification[] => {
+  const now = new Date(Date.now());
+  for (const error of dismissedNotification) {
```
JamesBurnside, line 236: error -> notification

packages/react-composites/src/composites/CallComposite/components/CapabilitiesNotficationBar.tsx
```diff
@@ -0,0 +1,275 @@
...
+/**
+ * Notification bar for capabilities changes
+ * @beta
+ */
+export const CapabilitiesChangeNotificationBar = (props: CapabilitiesChangeNotificationBarProps): JSX.Element => {
+  const locale = useLocale();
+
+  // Timestamp for when this component is first mounted.
+  // Never updated through the lifecycle of this component.
+  const mountTimestamp = useRef(new Date(Date.now()));
```
JamesBurnside, line 78: Will the calling sdk always populate the capabilities before this mounts? Is there a chance we get into a (for example) lobby page _then_ the SDK populates the capabilities (i.e. goes from `undefined -> [capabilities]`) in which case would we be incorrectly showing a capability changed notification?

packages/communication-react/review/beta/communication-react.api.md
alkwa-msft: Alot of these capabilities are aligned to match the capability being set in Teams. I believe in Teams they be using the "meeting" terminology which can be confusing. Aligning to similar wording could be helpful for those having to navigate from the Teams side. @mgamis-msft  can you share with us how we are coming up with these labels and from where were we inspired to use these labels?

packages/react-composites/src/composites/CallComposite/components/CapabilitiesNotficationBar.tsx
```diff
@@ -0,0 +1,275 @@
...
+export interface CapabilitiesChangeNotificationBarProps extends IMessageBarProps {
+  capabilitiesChangeInfo?: CapabilitiesChangeInfo;
+  participantRole?: ParticipantRole;
+}
+
+/* @conditional-compile-remove(capabilities) */
+/**
+ * @beta
+ */
+export interface CapabalityChangeNotification {
```
alkwa-msft, line 33: There is a lot of logic here for a UI component.

Can we separate the data layer from the UI a bit more?
Have something in a lower level that collect the various capability issues and then maintains some data structure that the UI can read through a selector? When we dismiss from the UI we push the signal down to remove that notification from this data structure so if you don't dismiss the notification between screens it will still be showing the most important thing?

packages/communication-react/review/beta/communication-react.api.md
mgamis-msft: I chose MeetingOption because the reasons capabilities can change is defined in the Calling SDK API below:
```
/**
 * Capabilities Changed Reason
 * @public
 */
export declare type CapabilitiesChangedReason = 'RoleChanged' | 'UserPolicyChanged' | 'MeetingOptionOrOrganizerPolicyChanged';
```
reference: https://www.npmjs.com/package/@azure/communication-calling/v/1.16.3-beta.1?activeTab=code

I thought adding the reason to the string key would be more specific since we probably we another string if turnVideoOn is lost through another reason. I also though 'capabilityTurnVideoOnLostDueToMeetingOptionOrOrganizerPolicy' is too long unless you think it is okay.

packages/react-composites/src/composites/CallComposite/components/CapabilitiesNotficationBar.tsx
```diff
@@ -0,0 +1,276 @@
...
+            {message}
+          </MessageBar>
+        );
+      })}
+    </Stack>
+  );
+};
+
+/* @conditional-compile-remove(capabilities) */
+const getNotificationMessage = (
```
dmceachernmsft, line 150: Should this be `getChangedCapabilityNotification`?

packages/react-composites/src/composites/CallComposite/components/CapabilitiesNotficationBar.tsx
```diff
@@ -0,0 +1,276 @@
...
+    dismissed.activeSince === undefined && active.get(dismissed.capabilityName) === undefined;
+
+  if (dismissedNotifications.some((dismissed) => shouldDeleteDismissal(dismissed))) {
+    return dismissedNotifications.filter((dismissed) => !shouldDeleteDismissal(dismissed));
+  }
+  return dismissedNotifications;
+};
+
+/* @conditional-compile-remove(capabilities) */
+const activeNotificationsToShow = (
```
dmceachernmsft, line 200: Same as above, should we add the CapabilitiesChanged prefix to all these?

packages/communication-react/review/beta/communication-react.api.md
mgamis-msft: As of right now, a capability change with the reason `UserPolicyChanged` is not something that is possible because user policies can only be applied to Teams users. That could change in the future but if we were to receive this reason then we would not show a notification.

packages/react-composites/src/composites/CallComposite/components/CapabilitiesNotficationBar.tsx
mgamis-msft: Moved the logic out of the UI component and moved the management of the notification collection and dismissals in CallArrangement.

packages/react-composites/src/composites/CallComposite/components/CapabilitiesNotficationBar.tsx
mgamis-msft: Removed the mount timestamp.

packages/react-composites/src/composites/CallComposite/components/CallArrangement.tsx
mgamis-msft: I hoisted a a similar tracking logic to what you did for errorbars but in CallArrangement. Is it better if I also put this logic in callcomposite along with the errorbar tracking? I chose CallArrangement because that is where CapabilitiesNotificationBar is used and where I pass the processed notifications to the CapabilitiesNotificationBar UI component. Also, I would not propagate props all the way from the CallComposite.

edwardlee-msft (approved):

prabhjot-msft (approved):

PR #3485, opened 2023-08-21
Capabilities change handler

mgamis-msft:
# What
- Added capabilitiesChanged to CallAdapter.
- Turning off camera when capability for camera is lost

# Why
- Allow Contoso to handle the capabilitiesChanged event.
- Copying behaviour of Teams to turn off camera when capability for camera is lost

# How Tested
Local Interop testing.
https://github.com/Azure/communication-ui-library/assets/79475487/7b1b93b6-1399-4ede-a8ac-c4516dc08c53

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -1158,6 +1169,14 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
     this.emitter.emit('transferRequested', newArgs);
   }
 
+  /* @conditional-compile-remove(capabilities) */
+  private capabilitiesChanged(data: CapabilitiesChangeInfo): void {
+    if (!data.newValue.turnVideoOn?.isPresent) {
```
alkwa-msft, line 1174: is newValue ever undefined?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -1054,6 +1063,8 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
     this.call?.on('stateChanged', this.subscribeToCaptionEvents.bind(this));
     /* @conditional-compile-remove(call-transfer) */
     this.call?.feature(Features.Transfer).on('transferRequested', this.transferRequested.bind(this));
+    /* @conditional-compile-remove(capabilities) */
+    this.call?.feature(Features.Capabilities).on('capabilitiesChanged', this.capabilitiesChanged.bind(this));
```
alkwa-msft, line 1067: I am a little cautious about chains with ? after what happened on Friday when we tried to get raised hand. Can we confirm that if we can resolve the call then we can always resolve the capabilities feature and if so we can guarantee that it will never be undefined?

mgamis-msft: What happened on Friday was a renaming of properties from Calling SDK. For example, property `ParticipantCapabilities.muteUnmuteMic` was renamed `ParticipantCapabilities.unmuteMic` in one of their beta releases. Are we supposed to prepare for any possible renaming Callling SDK will do? How can I guarantee that a property in the API is then for one version will be renamed in the next version of Calling SDK?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
mgamis-msft: newValue is a Record and can't be undefined but can be empty. But I should be checking that turnVideoOn present in the Record explicitly with an isPresent value of `false`. I should not coercing a Boolean. Fixed.

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -1129,6 +1137,14 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
     this.emitter.emit('transferRequested', newArgs);
   }
 
+  /* @conditional-compile-remove(capabilities) */
+  private capabilitiesChanged(data: CapabilitiesChangeInfo): void {
+    if (data.newValue.turnVideoOn?.isPresent === false) {
+      this.stopCamera();
```
dmceachernmsft, line 1143: Should we also be looking at the microphone state here to? Mute the participant if the capability is removed?

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
mgamis-msft: The calling SDK mutes the mic when the unmute capability changed to not present but yeah we also make sure on our side.

prabhjot-msft (approved):
looks good, hoping the existing comments are resolved before merge!

alkwa-msft (approved):

PR #3486, opened 2023-08-21
Improve shake to send logs

JamesBurnside:
# What

- fix iOS asking for permissions everytime page is opened
  - Checks if permission was already granted
- log all important logging events (added console.* events and window.onerror and window.unhandledrejection)
- do not log on desktop
  - This is because the local console logs in chrome was showing all logs coming from the shaketosendlogs line and not the original line (had to expand the callstack to see the original line which isn't good devx experience)

# Why

Improvements

# How Tested

Locally on desktop (windows, chrome), android chrome and ios safari. Permission prompt correctly only showed on safari and only when the permission state was cleared after hard resetting the browser app.


samples/Chat/src/app/utils/useIsMobile.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
```
JamesBurnside, line 1: previously didn't exist in the chat sample, this is a direct copy that call sample and callwithchat samples have

alkwa-msft (approved):

JamesBurnside:
This PR is gtg if anyone wants to complete it while I am OOO

dmceachernmsft (approved):

palatter (commented):
A lot of the logging code is duplicated between callWithChat, call, and chat.

Would it be possible to split this out into a common area of the code?

palatter (approved):
A lot of the logging code is duplicated between callWithChat, call, and chat.

Would it be possible to split this out into a common area of the code?

Apart from that, everything looks and tested on iPhone 14 Pro iOS 17 Beta

PR #3491, opened 2023-08-21
Alkwa/add carets to stable dependencies

alkwa-msft:
# What
Before when setting dependencies for calling or chat we used >= which allowed for major version changes as well. Instead we should be using ^ so we don't take on all possible forward changes.

# Why
With major version bumps none of the existing features will work since they are expecting a potentially different API Surface.

### DIFF
--- a/samples/ComponentExamples/package.json
+++ b/samples/ComponentExamples/package.json
@@ -22,7 +22,7 @@
     "_typecheck": "rushx _by-flavor \"rushx _if-preprocess && echo skip || (if-env COMMUNICATION_REACT_FLAVOR=beta && tsc)\""
   },
   "dependencies": {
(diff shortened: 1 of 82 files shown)


JoshuaLai (approved):

palatter (approved):

samples/ComponentExamples/package.json
```diff
@@ -68,4 +68,4 @@
     "webpack-cli": "~4.10.0",
     "webpack-dev-server": "4.8.1"
   }
-}
\ No newline at end of file
+}
```
PorterNan, line 71: And some missing line switch at the end of json file (I guess there is a eslint difference)

PorterNan (approved):

PR #3496, opened 2023-08-22
[Bug Fix] People button should announce selected state

carocao-msft:
# What
People button should announce selected state

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/2790321

# How Tested
Tested with samples 

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/Strings.tsx
```diff
@@ -213,6 +213,11 @@ export interface CallCompositeStrings {
    */
   peopleButtonLabel: string;
   /* @conditional-compile-remove(one-to-n-calling) @conditional-compile-remove(PSTN-calls) */
+  /**
+   * control bar People button label when checked
+   */
+  selectedPeopleButtonLabel: string;
```
dmceachernmsft, line 219: Why is this CC'd under these two?

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/Strings.tsx
carocao-msft: because peopleButtonLabel is cc under those two

edwardlee-msft (approved):

PR #3502, opened 2023-08-23
[Chat] Allowing iOS browsers to open images in new tab

JoshuaLai:
# What
The image gallery has a bug that has a download and view prompt. When users tap on `view` they are removed from the call with no way back. 
Implemented a workaround to allow users to open the image as a pop up and they can than save it from there. 
# Why
Gives users the ability to save the image on iOS devices of their own accord. 

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -288,15 +288,28 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
   );
 
   /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
-  const onImageDownloadButtonClicked = useCallback((imageUrl: string, saveAsName: string): void => {
-    // Create a new anchor element
-    const a = document.createElement('a');
-    // Set the href and download attributes for the anchor element
-    a.href = imageUrl;
-    a.download = saveAsName || 'download';
```
AmyL219, line 296: Would only add `a target='_blank'` attribute to the anchor (<a>) tag work here?
```
const a = document.createElement('a');
    
    a.href = imageUrl;

    a.target = '_blank';  // This instructs the browser to open the link in a new tab/window.

    a.download = saveAsName || 'download';
    
    document.body.appendChild(a);
    
    a.click();
    
    a.remove();
```

JoshuaLai: No on iOS browsers this does not work. The prompt is controlled by the native iOS, and because it sees that file type is a `png` it will always give the `View` option. So we try to open it as a pop up instead using the `FileReader` This will have another prompt instead, asking users if they want to `Allow` or `Block`. If they hit `Allow` they are navigated to a new page for just the image where users will have the opportunity to save it themselves.

AmyL219: Oh...I see. That's tricky. 
 I briefly checked Outlook's mobile web to see their approach. It seems they utilize a `Route` to create a new tab with a download URL, which is used for both previewing and downloading the content. But we don't have `Route` set up inside the component.

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -288,15 +288,28 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
-    // Programmatically click the anchor element to trigger the download
-    a.click();
-    a.remove();
+  const onImageDownloadButtonClicked = useCallback(async (imageUrl: string, saveAsName: string): Promise<void> => {
+    if (isIOS()) {
+      const reader = new FileReader();
+      reader.onload = () => {
+        window.open(imageUrl, '_blank');
+      };
+      const blob = await fetch(imageUrl).then((r) => r.blob());
```
Leah-Xia-Microsoft, line 297: For composite, the imageUrl we pass in is already the blob url. Would fetch from a blob url work here?

JoshuaLai: Yes, it does work, but this is a good point the reader really does nothing when I can just open the image url for the blob.

Leah-Xia-Microsoft (approved):

AmyL219 (approved):
Looks good to me for public preview. In future, we should think about manually opening downloads in a new tab to prevent edge cases may not limited to iOS

PR #3504, opened 2023-08-24
[Chat]Enable keyboard navigation for inlineImage

Leah-Xia-Microsoft:
# What

In message bubble, users cannot navigating within the message to different inline images using keyboard.
This PR is to enable this behaviour.

Here's a demo of what's changed.

https://github.com/Azure/communication-ui-library/assets/107075081/5ada03f4-9f95-41e1-ab1d-8792481d4c81

# Why

# How Tested

In a messageThread, using keyboard to navigate to a message bubble that contains multiple inline images
Press Enter key
Then press tab key to navigating between the inline images
Press Enter key again will open the selected inline image in a gallery view
Press Esc key will return the focus to the message bubble

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -220,7 +220,20 @@ const processInlineImage = (props: ChatMessageContentProps): ProcessingInstructi
     };
     /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
     return (
-      <span onClick={handleOnClick} data-ui-id={node.attribs.id}>
+      <span
+        data-ui-id={node.attribs.id}
+        onClick={handleOnClick}
+        tabIndex={0}
+        role="button"
```
JoshuaLai, line 227: Is this correct? Should we be treating this as a button? Won't it make accessibility incorrect?

Leah-Xia-Microsoft: > Is this correct? Should we be treating this as a button? Won't it make accessibility incorrect?

This is the wrapper element of the img tag and it is acting as a button.
However, you brought up a good point that I should add aria-label to the img tag so the voice over can announce it more accurately.

palatter (approved):
Looks good but @edwardlee-msft should take a look as well

jpeng-ms (approved):
looks good!

edwardlee-msft (approved):

PR #3506, opened 2023-08-25
[Chat]Add ChatTheme for ImageGallery specific colors

Leah-Xia-Microsoft:
# What

ImageGallery requires some specific colors.
We add them into the new ChatTheme so Contosos can control them through themes.

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -23,10 +23,9 @@ import {
   titleBarContainerStyle,
   titleStyle
 } from './styles/ImageGallery.style';
-import { useTheme } from '../theming/FluentThemeProvider';
-import { isDarkThemed } from '../theming/themeUtils';
 /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
 import { useLocale } from '../localization';
+import { darkTheme } from '../theming';
```
emlynmac, line 28: We probably don't want to hardcode the darktheme here.
Instead, use the `useTheme()` hook, so that if the theme is overridden then it picks up the values from the current active theme.

packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1069,6 +1069,12 @@ export type ChatParticipantListSelector = (state: ChatClientState, props: ChatBa
 // @public
 export type ChatReturnProps<Component extends (props: any) => JSX.Element> = GetChatSelector<Component> extends (state: ChatClientState, props: any) => any ? ReturnType<GetChatSelector<Component>> & Common<ChatHandlers, Parameters<Component>[0]> : never;
 
+// @public
+export interface ChatTheme {
+    // (undocumented)
```
emlynmac, line 1074: Can we make sure this is not `(undocumented)`?

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -104,17 +103,16 @@ export interface ImageGalleryStrings {
  */
 export const ImageGallery = (props: ImageGalleryProps): JSX.Element => {
   const { isOpen, images, onImageDownloadButtonClicked, onDismiss, onError, startIndex = 0 } = props;
-  const theme = useTheme();
-  const isDarkTheme = isDarkThemed(theme);
+  const theme = useTheme() as unknown as ChatTheme;
```
emlynmac, line 106: This cast should not be needed; does the useTheme() hook type need updating?

Leah-Xia-Microsoft: Chatted offline. We'll keep the current way of casting until Fluent 9 migration

JoshuaLai:
I noticed you were still using the inline-image conditional compile. I think we should use the image gallery one instead.

packages/react-components/src/theming/themes.ts
```diff
@@ -19,12 +19,37 @@ export interface CallingTheme {
   };
 }
 
+/**
+ * Custom Fluent theme palette used by chat related components in this library.
+ *
+ * @public
+ */
+export interface ChatTheme {
```
emlynmac, line 27: Should this interface be @beta and then the conditional compile changes applied to the theme object instead?

JoshuaLai (approved):

PR #3507, opened 2023-08-25
[Chat] Turning on the conditional conditional flag

JoshuaLai:
# What
Turning on the in flight feature so to to interfere with the next beta release. 

# Process & policy checklist


packages/react-components/src/components/ImageGallery.test.tsx
```diff
@@ -1,11 +1,15 @@
...
+/* @conditional-compile-remove(image-gallery) */
 import React from 'react';
+/* @conditional-compile-remove(image-gallery) */
 import { ImageGallery, ImageGalleryImageProps } from './ImageGallery';
+/* @conditional-compile-remove(image-gallery) */
 import { render, screen } from '@testing-library/react';
+/* @conditional-compile-remove(image-gallery) */
 import { registerIcons } from '@fluentui/react';
 
+/* @conditional-compile-remove(image-gallery) */
```
Leah-Xia-Microsoft, line 12: We don't need these 5 CCs because ImageGallery.tsx is not under CC so except the last two tests, all the other tests will pass

JoshuaLai: Good call out. Removed.

edwardlee-msft (approved):

Leah-Xia-Microsoft (approved):

PR #3512, opened 2023-08-25
Ruslanz/Disable raise hand button in lobby

rzdor:
Disable raise hand button in lobby.
Small changes in paddings


carocao-msft (approved):

prabhjot-msft (approved):
Looks good, since CSS is changed, there is expected snapshot update. Do we have snapshot tests for raised hands?

PR #3514, opened 2023-08-25
Send Box and Edit Box styles update

vhuseinova-msft:
# What

- Updated styles for edit message after migration to FluentUI v9

| After migration to v9 | After styles fixes |
| ------------- | ------------- |
| image | image |
- SendBox and EditBox button style updates

| Was | Now |
| ------------- | ------------- |
| image | image |
| image | image) |

# Why

Ui requirement changed and updates for FluentUI v9

# How Tested

Storybook

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
```diff
@@ -48,11 +48,6 @@ type ChatMessageComponentProps = {
    * Whether the status indicator for each message is displayed or not.
    */
   showMessageStatus?: boolean;
-  /**
-   * Inline the accept and reject edit buttons when editing a message.
-   * Setting to false will mean they are on a new line inside the editable chat message.
-   */
-  inlineAcceptRejectEditButtons: boolean;
```
emlynmac, line 55: What's the story with deleting this here?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -1067,7 +1065,6 @@ export const MessageThread = (props: MessageThreadProps): JSX.Element => {
             message={messageProps.message}
             userId={props.userId}
             remoteParticipantsCount={participantCount ? participantCount - 1 : 0}
-            inlineAcceptRejectEditButtons={!isNarrow}
```
emlynmac, line 1070: The `isNarrow` rendering is something that needs to be put back on from the initial PR which didn't implement this

vhuseinova-msft: no, in this case we don't need it. It will be needed for the ChatMessage, that's why I haven't removed it fully.

packages/react-components/src/components/styles/ChatMessageComponent.styles.ts
```diff
@@ -92,13 +96,36 @@ export const chatMessageMenuStyle = mergeStyles({
...
  * @private
  */
-export const chatMessageEditContainerStyle = {
-  margin: 0,
-  padding: 0,
-  maxWidth: 'unset',
-  minWidth: 'unset',
-  backgroundColor: 'transparent'
-};
+export const useChatMessageEditContainerStyles = makeStyles({
```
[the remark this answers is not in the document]
vhuseinova-msft: It looks like @private is used correctly here: https://jsdoc.app/tags-private.html. Name is updated

palatter:
Noticed this one padding difference

image

palatter (approved):
Made one comment about minor padding difference on the system message. Otherwise looks good!

vhuseinova-msft:
> Noticed this one padding difference
> 
> image

Updated padding for now. The main issue is that this is v8 component and FluentProvider (v9) sets a default line-height style

palatter (approved):
Padding looks good now!

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -51,6 +49,7 @@ import { FileMetadata } from './FileDownloadCards';
 /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
 import { AttachmentDownloadResult } from './FileDownloadCards';
 import { useTheme } from '../theming';
+import { createV9Theme, useFluentV9Wrapper } from './../theming/v9ThemeShim';
```
emlynmac, line 52: When will this be available upstream?

packages/react-components/src/theming/themeDuplicates.ts
```diff
@@ -0,0 +1,1167 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { BrandVariants } from '@fluentui/react-components';
+
```
emlynmac, line 5: Let's have a comment here to explain (link to) where this has come from and why we need to copy from fluent UI

vhuseinova-msft: Updated

packages/react-components/src/components/MessageThread.tsx
vhuseinova-msft: I will update this to a separate object (as you advised) in the next PR

packages/react-components/src/theming/themeDuplicates.ts
```diff
@@ -0,0 +1,187 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// These color records are required for createV9Theme
+// For more info, check https://react.fluentui.dev/iframe.html?viewMode=docs&id=concepts-migration-from-v8-components-theme-migration--page#compatible-themes
+
+/**
+ * @private
+ */
+export const grey: Record<Greys, string> = {
```
emlynmac, line 10: Does `Greys` need to be imported from somewhere?

PR #3515, opened 2023-08-25
[Chat] Fixing zoom on image in image gallery

JoshuaLai:
# What
On mobile device when end users pinch to zoom on the image, users are unable to dismiss the gallery. 
Enabling double tap to use the default behaviour will go back to initial scale. 

# Why
Introducing the default behaviour on mobile devices so they can dismiss the image gallery. 

# How Tested
User A join teams interop call with chat with User B on mobile platform. 
User A send an image in chat thread
User B receives the image, and pinch and zooms
User B double taps to zoom back out. 

Known issue, it seems after double tap you don't go completely back to scale 100%. You still need to zoom out slightly.
# Process & policy checklist


Leah-Xia-Microsoft (approved):
Tested. Works well. Thanks Josh.

jpeng-ms (approved):

PR #3535, opened 2023-08-31
Fix up scrolling and mention styling

emlynmac:
# What
Ensure scroll to bottom still works and mentions still look the same

# Why

# How Tested
Storybook

# Process & policy checklist


packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -56,11 +56,10 @@ export const useChatStyles = makeStyles({
     paddingBottom: '0.5rem',
     // paddingTop: '0.8rem',
```
vhuseinova-msft, line 57: Do we still need `paddingTop`?

emlynmac: Pro'lly not now. Removed!

vhuseinova-msft (approved):

PR #3536, opened 2023-08-31
[Chat] Add documentation for how to cover the safe area on iOS devices

Leah-Xia-Microsoft:
# What

To have the image gallery covers the safeArea on iOS devices, the only way is through modifying the global meta tag.
We don't want to add this to our composite since the control should be given to the Contoso's app.
Adding documentation to demo how Contosos can make Image Gallery covers the safe area on iOS devices.

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/storybook/stories/INTERNAL/ImageGallery/ImageGallery.stories.tsx
+++ b/packages/storybook/stories/INTERNAL/ImageGallery/ImageGallery.stories.tsx
@@ -33,7 +33,7 @@ const getDocs: () => JSX.Element = () => {
       </Canvas>
       <Heading>Cover iOS safe area</Heading>
       <Description>
-        To have the image gallery covers the safeArea on iOS devices, we can add `viewport-fit=cover` to the viewport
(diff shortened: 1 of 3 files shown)


packages/storybook/stories/INTERNAL/ImageGallery/ImageGallery.stories.tsx
```diff
@@ -30,6 +31,12 @@ const getDocs: () => JSX.Element = () => {
       <Canvas mdxSource={ImageGalleryExampleText}>
         <ImageGalleryExample />
       </Canvas>
+      <Heading>Cover iOS safe area</Heading>
+      <Description>
+        To have the image gallery covers the safeArea on iOS devices, we can add `viewport-fit=cover` to the viewport
```
palatter, line 36:
```suggestion
        To have the image gallery cover the safeArea on iOS devices, we can add `viewport-fit=cover` to the viewport
```

palatter (approved):
Looks good. One small suggestion on spelling.

JoshuaLai (approved):

PR #3537, opened 2023-08-31
OVC and 3x3 from beta to GA

vriosrada-msft:
# What

# Why

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/Converter.ts
```diff
@@ -27,7 +27,6 @@ import { CaptionsInfo } from './CallClientState';
 import { _isACSCall } from './TypeGuards';
 import { CallCommon, IncomingCallCommon } from './BetaToStableTypes';
 
-/* @conditional-compile-remove(video-background-effects) */
```
vriosrada-msft, line 30: Removing this just because now we will be using `Features`

PR #3545, opened 2023-09-01
[Chat] Fix image gallery is slow with large images

Leah-Xia-Microsoft:
# What

We use to not opening the image gallery until the image is loaded.
This makes opening the image gallery looks a bit slow for large images
This PR is to change the behavior to open the gallery with no image first, then wait for the load.

Before: https://microsoft-my.sharepoint.com/:v:/p/leahxia/EaYP44McEHhFrMCUalmNOh8BKjWd7chJMSGicY5vPJxY8g?e=s0cN3R
After: https://microsoft-my.sharepoint.com/:v:/p/leahxia/EfOFHPj516hCutGe9DwPSqMBPfdREojpuIxVNjnnuxCOog?e=cMRZVr

This PR also fixes the issue that the titleIcon and the title is too far apart from each other

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -267,6 +270,8 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
             imageUrl: fullSizeAttachments[attachment.id]
           }
         ]);
+        setIsImageGalleryOpen(true);
```
JoshuaLai, line 273: Why do we need to set true again here?

vhuseinova-msft (approved):

jpeng-ms (approved):

PR #3559, opened 2023-09-06
Update all versioning to caret

PorterNan:
# What

Update all versioning to caret

# Why

# How Tested

# Process & policy checklist


packages/acs-ui-common/package.json
```diff
@@ -38,31 +38,31 @@
...
   },
   "devDependencies": {
-    "@babel/cli": "~7.16.0",
-    "@babel/core": "~7.16.0",
-    "@microsoft/api-documenter": "~7.12.11",
-    "@microsoft/api-extractor": "~7.18.0",
-    "@types/jest": "~29.5.2",
+    "@babel/cli": "^7.16.0",
+    "@babel/core": "^7.16.0",
+    "@microsoft/api-documenter": "^7.12.11",
```
emlynmac, line 43: We have seen in the past that minor versions of these two components can have different behaviour. 
Perhaps leave these dev tools as ~ rather than ^?

PorterNan: Agree! I can't find a very strong reason that some app will use this package on their side

PorterNan: All updated!

JamesBurnside (approved):

PR #3569, opened 2023-09-07
[Chat] Remove unnecessary FocusTrapZone for ImageGallery

Leah-Xia-Microsoft:
# What

On MacOS Safari, an overlay layer is added to each image and with the FocusTrapZone, the width of this layer is set to the width of the FocusTrapZone which is sometimes larger than then image.
Since we don't need to trap focus for the image area, we're removing this componnet.

# Why

# How Tested

# Process & policy checklist


packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -157,35 +156,23 @@ export const ImageGallery = (props: ImageGalleryProps): JSX.Element => {
   const renderBodyWithLightDismiss = (): JSX.Element => {
     return (
       <Stack className={mergeStyles(bodyContainer)} onClick={() => props.onDismiss()}>
-        <FocusTrapZone
-          onKeyDown={(e) => {
-            if (e.key === 'Escape' || e.key === 'Esc') {
```
vhuseinova-msft, line 162: Does this disable dismiss when Escape is pressed?

packages/react-components/src/components/ImageGallery.tsx
```diff
@@ -157,35 +156,23 @@ export const ImageGallery = (props: ImageGalleryProps): JSX.Element => {
   const renderBodyWithLightDismiss = (): JSX.Element => {
     return (
       <Stack className={mergeStyles(bodyContainer)} onClick={() => props.onDismiss()}>
-        <FocusTrapZone
```
vhuseinova-msft, line 160: It looks like I don't the issue with additional padding anymore. But the background is opaque now. Is it expected?
image

Leah-Xia-Microsoft: > It looks like I don't the issue with additional padding anymore. But the background is opaque now. Is it expected? image

Yes, the background of the ImageGallery modal should be opaque:
image

packages/react-components/src/components/ImageGallery.tsx
Leah-Xia-Microsoft: > Does this disable dismiss when Escape is pressed?

Removing this would not disable Escape key to dismiss. Those line was needed only to overwrite FocusTrapZone's onKeyDown behaviour.

vhuseinova-msft (approved):

JoshuaLai (approved):

PR #3578, opened 2023-09-11
Remove Fluent NorthStar

emlynmac:
# What
Remove NorthStar fluent dependency and replace with fluent 9 controls.
Also update the styling on system messages to a smaller font and reduced vertical padding.
Added localized timestamp by default to chat messages

# Why
Fluent n* is deprecated; not compatible with react 18

# How Tested
Storybook, CI, manual testing.

# Process & policy checklist

Potentially has some impact to custom styling implementations requiring the components previously tied to n* fluent


packages/react-composites/playwright.config.hermetic.ts
```diff
@@ -11,10 +11,7 @@ const config: PlaywrightTestConfig = {
   workers: process.env.CI ? 3 : undefined,
 
   // All these tests are hermetic. It is safe to run them in parallel.
-  fullyParallel: true,
-
-  // TODO(prprabhu): Remove retries once super flakey tests are fixed or removed.
```
JamesBurnside, line 16: Yay if we can do this! Might be worth moving to its own PR and running the stress test action over it

packages/react-components/src/theming/v9ThemeShim.ts
```diff
@@ -0,0 +1,230 @@
...
+  const baseTheme = baseThemeV9 ?? webLightTheme;
+  return {
+    ...baseTheme,
+    ...mapAliasColors(themeV8.palette, themeV8.isInverted),
+    ...mapShadowTokens(themeV8.effects),
+    ...mapBorderRadiusTokens(themeV8.effects),
+    colorBrandBackground2: themeV8.palette.themeLight, // updated from palette.themeLighterAlt
+    colorBrandBackground2Hover: themeV8.palette.themeLight, // updated from palette.themeLighterAlt
+    colorBrandBackground2Pressed: themeV8.palette.themeLight, // updated from palette.themeLighterAlt
+    colorStatusWarningBackground3: '#D83B01',
```
JamesBurnside, line 223: Is this color always static even in dark vs light vs high contrast?

packages/react-components/src/components/utils/Datetime.ts
```diff
@@ -7,50 +7,26 @@ import { MessageThreadStrings } from '../MessageThread';
...
-    hours = hours - 12;
-  }
-  if (hours === 0) {
-    hours = 12;
-  }
-  if (minutes.length < 2) {
-    minutes = '0' + minutes;
-  }
-  return hours.toString() + ':' + minutes + ' ' + (isAm ? 'a.m.' : 'p.m.');
+  return messageDate.toLocaleTimeString([], { hour: 'numeric', minute: '2-digit' });
```
JamesBurnside, line 10: 😍

packages/react-components/src/components/styles/v8StyleShim.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { IStyle } from '@fluentui/react';
+import { ComponentSlotStyle } from '../../types';
+
+/**
+ * Function to create a React.CSSProperties object from a v8 style object.
+ * This function is still not ideal
+ * as v8Style can use pseudo-class selectors that style objects can't process correctly.
```
JamesBurnside, line 10: Just making a note to try remember to bug bash this and what the impact would be

packages/react-components/src/components/styles/EditBox.styles.ts
```diff
@@ -40,6 +33,6 @@ export const inputBoxIcon = mergeStyles({
  */
 export const editBoxStyleSet = {
   root: {
-    width: '100%'
+    minWidth: '30vw'
```
JamesBurnside, line 36: dont use vw - the composite may or may not be 100% of the viewport width/height so this likely doesn't behave as expected when the composite is < 100% of the width of the webpage

packages/react-components/src/components/ChatMessage/ChatMessageActionMenu.tsx
```diff
@@ -25,41 +24,31 @@ export const chatMessageActionMenuProps = (menuProps: {
...
-          </Ref>
-        ),
-
-        key: 'menuButton',
-        indicator: false
-      }
-    ]
+    children: (
+      <div
+        tabIndex={0} //make div focusable, as Icon below is migrated to v9, this can be deleted
```
JamesBurnside, line 38: per comment, can this be deleted?

JamesBurnside (approved):
Just one main thing to remove use of `vw` - but looks great!!

packages/react-composites/playwright.config.hermetic.ts
emlynmac: Removed here because of the triplicate screen checks wasting time on the build.

packages/react-components/src/components/styles/EditBox.styles.ts
emlynmac: What would you suggest here for layout of the edit box @JamesBurnside ?

packages/react-components/src/components/ChatMessage/ChatMessageActionMenu.tsx
emlynmac: We're still using the v8 icons (https://developer.microsoft.com/en-us/fluentui#/controls/web/icon).
v9 Icons (https://react.fluentui.dev/?path=/docs/concepts-developer-icons-icons--page) are from the react-icons package and we've not migrated.

packages/react-composites/tests/browser/snapshots/stable/tests/browser/callwithchat/hermetic/OverflowGallery.test.ts-snapshots/overflow-gallery-with-chat-pane-open-rtl-Desktop-Chrome-linux.png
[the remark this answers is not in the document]
palatter: I'll look into this issue

packages/react-components/src/theming/v9ThemeShim.ts
vhuseinova-msft: Yes, it is the one that was used for mentions

packages/react-components/src/components/SystemMessage.tsx
```diff
@@ -41,7 +41,7 @@ export const SystemMessage = (props: SystemMessageProps): JSX.Element => {
   return (
     <Stack horizontal className={mergeStyles(props?.containerStyle as IStyle)} tabIndex={0}>
       {Icon}
-      <Text style={{ wordBreak: 'break-word' }} role="status" title={content}>
+      <Text style={{ wordBreak: 'break-word' }} role="status" title={content} variant="small">
```
dmceachernmsft, line 44: nit: Can we re-type this as `variant={'small'}`?

packages/react-components/src/theming/v9ThemeShim.ts
```diff
@@ -0,0 +1,230 @@
...
+    colorNeutralStrokeDisabled: palette.neutralQuaternaryAlt,
+    colorNeutralStrokeInvertedDisabled: whiteAlpha[40],
+    colorTransparentStroke: 'transparent',
+    colorTransparentStrokeInteractive: 'transparent',
+    colorTransparentStrokeDisabled: 'transparent',
+    colorNeutralStrokeAlpha: inverted ? whiteAlpha[10] : blackAlpha[5],
+    colorNeutralStrokeAlpha2: whiteAlpha[20],
+    colorStrokeFocus1: palette.white,
+    colorStrokeFocus2: palette.black,
+    colorNeutralShadowAmbient: 'rgba(0,0,0,0.12)',
```
dmceachernmsft, line 173: are these hard coded since there are no matching properties from a v8 theme?

packages/react-components/src/theming/v9ThemeShim.ts
dmceachernmsft: is there a concern if contoso wants to override this they can't because of the hard coding?

dmceachernmsft (approved):

packages/communication-react/package.json
```diff
@@ -28,11 +28,14 @@
     "@azure/core-paging": "^1.5.0",
     "@azure/core-rest-pipeline": "^1.11.0",
     "@azure/logger": "^1.0.4",
+    "@fluentui/react-components": "^9.30.0",
     "@fluentui/react": "^8.110.4",
     "@fluentui/react-file-type-icons": "8.8.13",
     "@fluentui/react-hooks": "^8.6.12",
     "@fluentui/react-icons": "^2.0.194",
+    "@fluentui/react-migration-v8-v9": "^9.4.0",
```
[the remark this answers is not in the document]
emlynmac: I think this can be removed as it's unused...

packages/react-components/src/theming/v9ThemeShim.ts
vhuseinova-msft: not now as we haven't made it public before. So, when we decide to change this, we can reuse v8 color here.

PR #3591, opened 2023-09-18
[Gallery Layouts] Large Gallery

dmceachernmsft:
# What

image
Introduces new gallery mode to render up to 49 tiles at once
# Why

Shows more tiles in larger meetings

https://skype.visualstudio.com/SPOOL/_workitems/edit/3404481
# How Tested

Locally, E2E tests to come


packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+    dominantSpeakers,
+    maxRemoteVideoStreams: maxStreamsTrampoline(),
+    isScreenShareActive: !!screenShareComponent,
+    maxOverflowGalleryDominantSpeakers: screenShareComponent
+      ? childrenPerPage.current - ((pinnedParticipantUserIds.length + 1) % childrenPerPage.current)
+      : childrenPerPage.current,
+    /* @conditional-compile-remove(pinned-participants) */ pinnedParticipantUserIds,
+    /* @conditional-compile-remove(gallery-layouts) */ layout: 'largeGallery'
+  });
+  console.log('gridParticipants', gridParticipants);
```
JamesBurnside, line 76: remove console log

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+        shouldFloatLocalVideo={false}
+        overflowGalleryElements={overflowGalleryTiles}
+        horizontalGalleryStyles={styles?.horizontalGallery}
+        /* @conditional-compile-remove(vertical-gallery) */
+        verticalGalleryStyles={styles?.verticalGallery}
+        /* @conditional-compile-remove(vertical-gallery) */
+        overflowGalleryPosition={overflowGalleryPosition}
+        onFetchTilesToRender={setIndexesToRender}
+        onChildrenPerPageChange={(n: number) => {
+          childrenPerPage.current = n;
```
JamesBurnside, line 128: should this callback trigger a re-render? (currently it uses useRef so won't)

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+    return parentWidth && parentHeight
+      ? calculateMaxTilesInLargeGrid(parentWidth, parentHeight)
+      : maxRemoteVideoStreams;
+    return maxRemoteVideoStreams;
+  };
+
+  // This is for tracking the number of children in the first page of overflow gallery.
+  // This number will be used for the maxOverflowGalleryDominantSpeakers when organizing the remote participants.
+  // We need to add the local participant to the pinned participant count so we are placing the speakers correctly.
+  const childrenPerPage = useRef(4);
```
JamesBurnside, line 63: where did `4` come frome? use defined constants with names for arbitrary numbers. 
Why 4 and not 5 like horiztonal gallery?
image

packages/react-components/src/components/VideoGallery.tsx
```diff
@@ -684,6 +686,10 @@ export const VideoGallery = (props: VideoGalleryProps): JSX.Element => {
     if (layout === 'speaker') {
       return <SpeakerVideoLayout {...layoutProps} />;
     }
+    /* @conditional-compile-remove(gallery-layouts) */
+    if (layout === 'largeGallery') {
```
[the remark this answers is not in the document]
JamesBurnside: And for dynamic layout, when there's a huge grid of >49 participants, should it not also overflow tiles?

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+        isShort={isShort}
+        shouldFloatLocalVideo={false}
+        overflowGalleryElements={overflowGalleryTiles}
+        horizontalGalleryStyles={styles?.horizontalGallery}
+        /* @conditional-compile-remove(vertical-gallery) */
+        verticalGalleryStyles={styles?.verticalGallery}
+        /* @conditional-compile-remove(vertical-gallery) */
+        overflowGalleryPosition={overflowGalleryPosition}
+        onFetchTilesToRender={setIndexesToRender}
+        onChildrenPerPageChange={(n: number) => {
```
JamesBurnside, line 127: nit: pass in memoized fn

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+    </Stack>
+  );
+};
+
+const overflowGalleryTrampoline = (
+  gallery: JSX.Element | null,
+  galleryPosition?: 'HorizontalBottom' | 'VerticalRight' | 'HorizontalTop'
+): JSX.Element | null => {
+  /* @conditional-compile-remove(gallery-layouts) */
+  return galleryPosition !== 'HorizontalTop' ? gallery : <></>;
```
JamesBurnside, line 176: we don't show a overflow gallery if its horizontal top?

JamesBurnside: Oh its to do with placement of the gallery, avoid putting one check for HorizontalTop in a callback function, but the other just inline:
image

JamesBurnside: Out of interest, instead of having to do this check and having the overflow sorta defined in two places, can you set `reversed` property on the stack so children are in the opposite order (so `reversed={galleryPosition === 'HorizontalTop'}`) -- not sure how it would impact the a11y keyboard ordering though

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
dmceachernmsft: This is from the vertical-gallery work, this should not trigger a rerender from what I remember

packages/react-components/src/components/VideoGallery.tsx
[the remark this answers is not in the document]
dmceachernmsft: I have addressed this difference in the utils function 😊 this was a gap I realised as well

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
dmceachernmsft: there is a bug with the fluent stacks where that wont work sadly, as for the callback its just for CC reasons since vertical gallery is also not GA'd, that is part of this effort

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Props for {@link LargeGalleryLayout}.
+ *
+ * @private
+ */
+export type LargeGalleryProps = LayoutProps;
+
+const DEFAULT_CHILDREN_PER_PAGE = 5;
+/* @conditional-compile-remove(gallery-layouts) */
+const REM_TO_PIXEL = 16;
```
mgamis-msft, line 27: Could you use _convertRemToPx instead? Note: you will need to mock this function for any unit tests like [this](https://github.com/Azure/communication-ui-library/blob/70626a43d2c7d7a7a044cf366b16ce7a07b155f5/packages/react-components/src/components/ResponsiveVerticalGallery.test.tsx#L126)

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
JamesBurnside: can't tell what exactly this does.. so I trust you :)

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
JamesBurnside: if others have it and we see no issue, might be OK to leave it, will leave this up to you

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
```diff
@@ -33,6 +33,11 @@ export interface OrganizedParticipantsResult {
 }
 
 const DEFAULT_MAX_OVERFLOW_GALLERY_DOMINANT_SPEAKERS = 6;
+const DEFAULT_MAX_VIDEO_SREAMS = 4;
+/* @conditional-compile-remove(large-gallery) */
+const MAX_VIDEO_STREAMS_FOR_LARGE_GALLERY = 49;
```
JamesBurnside, line 38: should this be 48? (if you (local) is one of them?)

JamesBurnside (approved):

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
dmceachernmsft: good catch!

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
```diff
@@ -42,14 +47,30 @@ const _useOrganizedParticipants = (props: OrganizedParticipantsArgs): OrganizedP
...
+    /* @conditional-compile-remove(gallery-layouts) */
+    if (maxRemoteVideoStreams > MAX_GRID_PARTICIPANTS_NOT_LARGE_GALLERY) {
+      return MAX_GRID_PARTICIPANTS_NOT_LARGE_GALLERY;
+    } else {
+      return maxRemoteVideoStreams;
+    }
+    return maxRemoteVideoStreams;
+  };
+
+  const maxRemoteVideoStreamsToUse = calculateMaxRemoteVideoStreams();
```
mgamis-msft, line 72: Don't we calculate this in LargeGalleryLayout?

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
```diff
@@ -0,0 +1,191 @@
...
+  galleryPosition?: 'HorizontalBottom' | 'VerticalRight' | 'HorizontalTop'
+): JSX.Element | null => {
+  /* @conditional-compile-remove(gallery-layouts) */
+  return galleryPosition !== 'HorizontalTop' ? gallery : <></>;
+  return gallery;
+};
+
+/* @conditional-compile-remove(gallery-layouts) */
+const calculateMaxTilesInLargeGrid = (parentWidth: number, parentHeight: number): number => {
+  const xAxisTiles = Math.floor(parentWidth / (VERTICAL_GALLERY_TILE_SIZE_REM.width * REM_TO_PIXEL));
```
mgamis-msft, line 186: Why are we using vertical gallery tile size here?

dmceachernmsft: these tiles are smaller and used as a good benchmark for what a 'smallest' tile could be, think I should make a new constant?

packages/react-components/src/components/VideoGallery/utils/videoGalleryLayoutUtils.ts
dmceachernmsft: this is to cap the number of tiles for when OVC eventually supports more than 9

packages/react-components/src/components/VideoGallery/LargeGalleryLayout.tsx
mgamis-msft: I see. That's alright. Maybe we can refactor the name of VERTICAL_GALLERY_TILE_SIZE_REM to something like SMALLEST_TILE_SIZE_REM. But for a refactor PR after once everything is in.

mgamis-msft (approved):

PR #3595, opened 2023-09-19
[BugFix] Default message timestamp to use locales

emlynmac:
# What
Use the `toLocale[Time/Date]String()` methods on `Date` to obtain a properly localized default message timestamp string

# Why
Current implementation is English 12hr format only and for anything else, users must override the renderer.

# How Tested
Storybook, CI

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/utils/Datetime.ts
+++ b/packages/react-components/src/components/utils/Datetime.ts
@@ -7,50 +7,26 @@ import { MessageThreadStrings } from '../MessageThread';
  * @private
  */
 export const formatTimeForChatMessage = (messageDate: Date): string => {
-  let hours = messageDate.getHours();
-  let minutes = messageDate.getMinutes().toString();
-  const isAm = hours < 12;
-  if (hours > 12) {
-    hours = hours - 12;
-  }
-  if (hours === 0) {
-    hours = 12;
-  }
-  if (minutes.length < 2) {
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/utils/Datetime.ts
```diff
@@ -7,50 +7,26 @@ import { MessageThreadStrings } from '../MessageThread';
...
- * Given a message date object in ISO8601 and a current date object, generates a user friendly timestamp text like the
- * following:
- *
- * 1:30 p.m.
- * Yesterday 1:30 p.m.
- * Monday 1:30 p.m.
- * 2021-01-10 1:30 p.m.
+ * Given a message date object in ISO8601 and a current date object, generates a user friendly timestamp text
+ * using the system locale.
+ * <time in locale format>.
```
vhuseinova-msft, line 23: can be added for all examples: `e.g. 1:30pm`

vhuseinova-msft (approved):

packages/react-components/src/components/utils/Datetime.ts
emlynmac: Now we have a variable format, it's not really possible to define what will happen. 24hr vs 12hr, different date separators for different locales etc.
I've added templated examples but really can't go into every possibility :)

PR #3604, opened 2023-09-21
Star Survey inside call composite 

carocao-msft:
# What
Star Survey inside call composite 

# Why
End of call survey 

# How Tested
Tested with samples 

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/components/StarSurvey.tsx
+++ b/packages/react-composites/src/composites/CallComposite/components/StarSurvey.tsx
@@ -1,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import React from 'react';
+ /* @conditional-compile-remove(end-of-call-survey) */
+import {useState} from 'react';
+ /* @conditional-compile-remove(end-of-call-survey) */
+import {
+  _StarSurvey,
+  _StarSurveyStrings
+} from '@internal/react-components';
+ /* @conditional-compile-remove(end-of-call-survey) */
+import { _captionSettingsSelector } from '@internal/calling-component-bindings';
+ /* @conditional-compile-remove(end-of-call-survey) */
+import { useLocale } from '../../localization';
+ /* @conditional-compile-remove(end-of-call-survey) */
+import { useHandlers } from '../hooks/useHandlers';
+
+/** @private */
+export const StarSurvey = (): JSX.Element => {
+/* @conditional-compile-remove(end-of-call-survey) */
+  const starSurveyHandler = useHandlers(StarSurvey)
+  /* @conditional-compile-remove(end-of-call-survey) */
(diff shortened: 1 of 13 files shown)


packages/react-composites/src/composites/CallComposite/pages/NoticePage.tsx
```diff
@@ -42,6 +45,8 @@ export function NoticePage(props: NoticePageProps): JSX.Element {
       data-ui-id={props.dataUiId}
       aria-atomic
     >
+      {/* @conditional-compile-remove(end-of-call-survey) */
+     <StarSurvey/>}
```
dmceachernmsft, line 49: This looks like this will always render no matter which kind of notice page this is. Should we be making sure that this is something that is only on the end call screen? this will show if you were kicked as well for example, is this something we want?

carocao-msft: I posted the question in the feature weekly, it feels like it makes more sense to show across the notice page? Calling wants to get as many data as possible :) but I am also not certain at a end user stand point this will be annoying

packages/react-composites/src/composites/CallComposite/components/StarSurvey.tsx
edwardlee-msft, line 1: Does everything in here need to be conditionally compiled?
StarSurvey is internal so it's probably not needed.

Or is there something up the tree of commands which causes errors if this is not conditionally compiled?

packages/react-composites/src/composites/CallComposite/pages/NoticePage.tsx
dmceachernmsft: the question to me for including it when the user is removed or disconnects is whether contoso would want feedback in those situations. Thinking about feedback from being removed from the call they might want to leave negative feedback that is "less than constructive." if we also look at it from the Teams perspective they only request feedback when the call ends naturally, if the end call reason was a disconnection the SDK team can report that.

carocao-msft: https://github.com/Azure/communication-ui-library/pull/3608/files Addressed this issue in this pr, survey is only shown on end call page now

packages/react-composites/src/composites/CallComposite/components/StarSurvey.tsx
carocao-msft: Those import statements causes errors if not conditionally compiled so we have to cc all of them. We do this for other internal composite components too

dmceachernmsft (approved):

edwardlee-msft (approved):

carocao-msft:
Holding off merging this to prevent massive PR forming  on branch [carocao/survey](https://github.com/Azure/communication-ui-library/tree/carocao/survey)

PR #3607, opened 2023-09-22
API tag survey component

carocao-msft:
# What
API tag survey component.
# Why

End of call survey

# How Tested
Tested with storybook
Screenshot 2023-12-04 at 1 10 17 AM

# Process & policy checklist


packages/react-components/review/beta/react-components.api.md
```diff
@@ -2176,6 +2225,13 @@ export interface StreamMediaProps {
     videoStreamElement: HTMLElement | null;
 }
 
+// @internal (undocumented)
```
prabhjot-msft, line 2228: Shouldn't this be documented?

packages/react-components/review/beta/react-components.api.md
```diff
@@ -1327,6 +1373,9 @@ export interface ImageGalleryStrings {
     downloadButtonLabel: string;
 }
 
+// @internal (undocumented)
```
prabhjot-msft, line 1376: This as well, I think we should document these

packages/storybook/stories/INTERNAL/Survey/TagsSurvey.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const callIssuesToTag = {
+    overallRating: {
+      CallCannotJoin: 'I could not join call',
+      CallCannotInvite: 'I could not invite others into the call',
+      HadToRejoin: 'I had to rejoin the call',
+      CallEndedUnexpectedly: 'Call ended for me unexpectedly',
+      OtherIssues: 'I was having other issues with the call'
+    },
+    audioRating: {
+      NoLocalAudio: 'The other side could not hear any sound',
```
dmceachernmsft, line 65: are these going to be localized? or is this something that only exists for Contoso to add their own strings to?

packages/react-components/review/beta/react-components.api.md
```diff
@@ -2185,6 +2184,72 @@ export interface StreamMediaProps {
     videoStreamElement: HTMLElement | null;
 }
 
+// @public
+export interface SurveyCategories {
```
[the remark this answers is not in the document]
carocao-msft: It's mapping from calling sdk values to localized strings. But both key value and string value are used in the components, so I have a separate type for them similar to how we handled CaptionsAvailableLanguageStrings.ts

packages/react-components/review/beta/react-components.api.md
```diff
@@ -2196,6 +2261,24 @@ export interface SystemMessageCommon extends MessageCommon {
     messageType: 'system';
 }
 
+// @internal
+export const _TagsSurvey: (props: _TagsSurveyProps) => JSX.Element;
+
+// @internal
+export interface _TagsSurveyProps {
+    callIssuesToTag: SurveyIssues;
+    categoriesToHeader: SurveyCategories;
```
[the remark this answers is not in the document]
carocao-msft: We use both the keys and values here similar to the mapping between language code and language names in captions. I think we should set it up the same way as how we set up captions? 
Screenshot 2023-12-04 at 11 47 10 AM

packages/react-components/src/components/Survey/TagsSurvey/TagsSurvey.tsx
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @internal
+ */
+export interface _TagsSurveyStrings {
+  /**
+   * Survey question
+   */
+  tagsSurveyQuestion?: string;
+  /**
+   * Confirm Button Label
```
dmceachernmsft, line 28: nit: I don't think this comment is correct

dmceachernmsft (approved):

PR #3608, opened 2023-09-22
Survey inside composite

carocao-msft:
# What
End of call survey inside composite

# Why
end of call survey

# How Tested
Tested with samples
Screenshot 2023-12-04 at 1 53 12 AM

# Process & policy checklist


packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -599,6 +599,7 @@ export type CallCompositeOptions = {
     galleryOptions?: {
         layout?: VideoGalleryLayout;
     };
+    survey?: boolean;
```
JamesBurnside, line 602: Is there a method for a developer to get the survey results? Will it come through as adapter state or do we need a callback here like
```suggestion
    survey?: boolean | {
      onResultsSubmit: (results: SurveyResults) => Promise<void>
    };
```

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -747,6 +749,10 @@ export interface CallCompositeStrings {
     startCaptionsButtonOnLabel?: string;
     startCaptionsButtonTooltipOffContent?: string;
     startCaptionsButtonTooltipOnContent?: string;
+    TagsSurveyCancelButtonAriaLabel: string;
```
JamesBurnside, line 752: these are PascalCase where the rest of the props are camelCase. lowercase the first letter in each of these

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -616,11 +617,11 @@ export interface CallCompositeProps extends BaseCompositeProps<CallCompositeIcon
 export interface CallCompositeStrings {
     blurBackgroundEffectButtonLabel?: string;
     blurBackgroundTooltip?: string;
+    callIssuesToTags: CallIssuesToTags;
```
JamesBurnside, line 620: `callIssuesToTags` is confusing to know what this is for. perhaps
```suggestion
    surveyIssues: SurveyIssues;
```

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -733,6 +734,7 @@ export interface CallCompositeStrings {
     soundLabel: string;
     spokenLanguageStrings?: SpokenLanguageStrings;
     starRatingAriaLabel: string;
+    starRatingCancelButtonAriaLabel: string;
```
JamesBurnside, line 737: does survey need two different cancel button aria labels? could this just be surveyCancelButtonAriaLabel and remove TagsSurveyCancelButtonAriaLabel

JamesBurnside: same with cancel and submit, instead just have surveyCancelButtonLabel and surveySubmitButtonLabel

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: ~~Should it have the ability to choose tag survey vs star survey?~~

```suggestion
    survey?: boolean | {
      onResultsSubmit: (results: SurveyResults) => Promise<void>;
      surveyKind: 'tags' | 'star'
    };
```

Ah I see it is done based on the star rating they provide

packages/react-components/src/components/Survey/StarSurvey/StarSurvey.tsx
```diff
@@ -84,6 +84,8 @@ export interface _StarSurveyProps {
   unselectedIcon?: string;
   /** Function to send StarSurvey results*/
   onSubmitSurvey?: (survey: _CallSurvey) => Promise<_CallSurveyResponse | undefined>;
+  /** function called on confirm button click*/
+  onConfirmStarSurvey?:(ratings:number) => void
```
JamesBurnside, line 88: how does this differ to onSubmitSurvey?

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -234,6 +234,12 @@ export type CallCompositeOptions = {
      */
     layout?: VideoGalleryLayout;
   };
+  /* @conditional-compile-remove(end-of-call-survey) */
+  /**
+   * Show call survey at end of call
```
JamesBurnside, line 239: nit
```suggestion
   * Show call survey at the end of a call.
```

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
```diff
@@ -234,6 +234,12 @@ export type CallCompositeOptions = {
...
+   * @defaultValue true
```
JamesBurnside, line 240: we should name it to default to false, so instead of survey, have hideSurvey. (see other comment about adding more survey options also, maybe `hideSurvey`/`hidden`/`doNotShow` is inside a surveyOptions bag)

samples/Calling/src/app/views/CallScreen.tsx
```diff
@@ -66,6 +66,7 @@ export const CallScreen = (props: CallScreenProps): JSX.Element => {
   const afterCallAdapterCreate = useCallback(
     async (adapter: CallAdapter): Promise<CallAdapter> => {
       subscribeAdapterEvents(adapter);
+      adapter.joinCall({microphoneOn: true})
```
JamesBurnside, line 69: was this test code to remove?

JamesBurnside:
I feel `tag` isn't descriptive of what this survey entails.. maybe `IssueSurvey`?

packages/communication-react/review/beta/communication-react.api.md
prabhjot-msft, line 602: For consistency around naming, should it be something like `surveyOptions`

carocao-msft: It's enabling and disabling a feature similar to errorBar, so I followed the naming and named it as survey

samples/Calling/src/app/views/CallScreen.tsx
carocao-msft: oops that's a mistake gonna remove that

packages/react-composites/src/composites/CallComposite/CallComposite.tsx
carocao-msft: I was following what was done with another feature errorBar, where we have it default true and when passing into false, error bars will not show. Survey is similar to errorBar since they are both optional features, so I followed the naming convention

packages/react-components/src/components/Survey/StarSurvey/StarSurvey.tsx
carocao-msft: Onsubmitsurvey is what's called to submit the survey result, onConfirm is an optional callback users can pass in to do something when the confirm button is clicked

packages/communication-react/review/beta/communication-react.api.md
carocao-msft: Those strings are for 2 different surveys, users might want to have different strings for confirm button and cancel button on those surveys?

carocao-msft:
> I feel `tag` isn't descriptive of what this survey entails.. maybe `IssueSurvey`? (don't love that either though..)

The survey type is called API tag survey so I named it to tag, not sure what would be a better name...

packages/communication-react/review/beta/communication-react.api.md
JamesBurnside: Ah yes sorry you are correct! I wrote this before I realized the star survey flows into this survey and it isn't one or the other

packages/react-composites/src/composites/CallComposite/pages/NoticePage.tsx
```diff
@@ -24,6 +30,42 @@ export interface NoticePageProps {
...
   disableStartCallButton?: boolean;
   pageStyle?: IStyle;
+  /* @conditional-compile-remove(end-of-call-survey) */
+  /**
+   * Options for end of call survey
+   */
+  surveyOptions?: {
+    /**
+     * Hide call survey at the end of a call.
+     * @defaultValue true
```
JamesBurnside, line 40: is the default value `true`? (that we hide by default?)
If so this should be `showSurvey` and default value false

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/pages/NoticePage.tsx
dmceachernmsft: I like the idea that we should be showing by default as we have been with other features. To make this work do they need to provide information that would make this a case for hiding by default? I wonder if we can conditionally render the survey if they provide the callback needed to submit the survey and remove this API all together?

packages/react-composites/src/composites/common/SurveyPaneContent.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+        {showTagsSurvey && <TagsSurvey issues={issues} onConfirm={onIssuesSelected} />}
+      </Stack>
+      {showSubmitFeedbackButton && (
+        <Stack style={{ borderTop: 'solid', borderTopColor: 'lightgrey' }} horizontalAlign="end">
+          <PrimaryButton
+            style={{ marginTop: '2rem' }}
+            onClick={() => {
+              const surveyResults: CallSurvey = { overallRating: { score: ratings } };
+              if (issuesSelected?.overallRating) {
```
dmceachernmsft, line 89: nit: is this better represented as a switch on the `issueSelected`?

dmceachernmsft (approved):

PR #3609, opened 2023-09-22
[FluentUI] RTL Fix

palatter:
# What

If a fluentUI theme is applied outside of the composite, it will be overidden by the fluentUI theme in the baseComposite. Props like 'rtl' will be lost then if not also given to chat composite. Added logic to prevent the provider from being applied multiple times.

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/theming/FluentThemeProvider.tsx
+++ b/packages/react-components/src/theming/FluentThemeProvider.tsx
@@ -48,6 +48,7 @@ const defaultTheme: Theme = {
 
 /** Theme context for library's react components */
 const ThemeContext = createContext<Theme>(defaultTheme);
+const ProviderContext = createContext<boolean>(false);
 
 /**
  * Provider to apply a Fluent theme across this library's react components.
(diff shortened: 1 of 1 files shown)


packages/react-components/src/theming/FluentThemeProvider.tsx
```diff
@@ -78,3 +92,10 @@ export const FluentThemeProvider = (props: FluentThemeProviderProps): JSX.Elemen
  * @public
  */
 export const useTheme = (): Theme => useContext(ThemeContext);
+
+/**
+ * This is used to prevent the provider from being applied multiple times.
+ *
+ * @private
+ */
+const useProvider = (): boolean => useContext(ProviderContext);
```
emlynmac, line 101: maybe `alreadyHasThemeProvider`?

JamesBurnside: prefix with `use` (react hooks nomenclature) `useAlreadyHasThemeProvider`

packages/react-components/src/theming/FluentThemeProvider.tsx
```diff
@@ -60,15 +61,28 @@ const ThemeContext = createContext<Theme>(defaultTheme);
 export const FluentThemeProvider = (props: FluentThemeProviderProps): JSX.Element => {
   const { fluentTheme, rtl, children } = props;
 
+  /**
+   * Pass only the children if we previously wrapped.
+   * This is to prevent the provider from being applied
+   * multiple times and wiping out configuration like RTL
+   * FluentThemeProvider
+   */
+  const alreadyWrapped = useProvider();
```
JamesBurnside, line 70: this is called outside of the `<ProviderContext` - does this ever return true? My understanding the useContext has to be called from a component that is a child of `<ProviderContext />`

JamesBurnside: oh i see how this works, never mind!

packages/react-components/src/theming/FluentThemeProvider.tsx
```diff
@@ -48,6 +48,7 @@ const defaultTheme: Theme = {
 
 /** Theme context for library's react components */
 const ThemeContext = createContext<Theme>(defaultTheme);
+const ProviderContext = createContext<boolean>(false);
```
JamesBurnside, line 51: suggestion: update the name from just `Provider` as that is super generic.
Perhaps `SingleUsageCheckProvider`

JamesBurnside (approved):

emlynmac (approved):

PR #3611, opened 2023-09-22
[Gallery layouts] Mobile gallery e2e testing

dmceachernmsft:
# What

Adds e2e test for mobile gallery layout selection
# Why

adds testing to cover mobile options

https://skype.visualstudio.com/SPOOL/_workitems/edit/3417199
# How Tested

Ran tests locally to see the generated snapshots


edwardlee-msft:
The horizontal gallery buttons look like they are missing in the snapshots

edwardlee-msft (approved):

packages/react-composites/tests/browser/call/hermetic/VideoGallery.test.ts
```diff
@@ -216,4 +216,24 @@ test.describe('VideoGallery tests', async () => {
...
+        mockRemoteParticipantCount: '20'
+      })
+    );
+
+    await waitForSelector(page, dataUiId(IDS.moreButton));
+    expect(await stableScreenshot(page)).toMatchSnapshot('dynamic-layout-mobile.png');
+    await pageClick(page, dataUiId(IDS.moreButton));
+    await page.locator('span:has-text("Gallery options")').click();
+    await page.locator('span:has-text("Gallery layout")').click();
+    expect(await stableScreenshot(page)).toMatchSnapshot('default-layout-mobile.png');
```
mgamis-msft, line 237: Can we call the snapshot 'gallery-layout-mobile' instead?

mgamis-msft:
What's logical difference between the gallery layout and the dynamic layout?
gallery layout
image
dynamic layout
image
Dynamic layout only allows maximum of 4 grid tiles?

packages/react-composites/tests/browser/call/hermetic/VideoGallery.test.ts
dmceachernmsft: I want to keep the snapshots referencing the API type since its for our internal testing

dmceachernmsft:
> What's logical difference between the gallery layout and the dynamic layout? gallery layout image dynamic layout image Dynamic layout only allows maximum of 4 grid tiles?

since there are no video participants in the dynamic layout they will have the same number of tiles, I can edit the test so that there are video participants as well to highlight this. Since OVC is not set in the hermetic tests it falls back to the gallery default of 4 for the max grid participants.

mgamis-msft (approved):

PR #3614, opened 2023-09-22
RTL paddings fix

vhuseinova-msft:
# What

Fixed paddings for rtl

# Why

# How Tested

Storybook

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -309,6 +309,12 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
     message.attached === 'top' || message.attached === false
       ? chatMessageStyles.bodyWithAvatar
       : chatMessageStyles.bodyWithoutAvatar,
+    // handle rtl margins for message without avatar
+    theme.rtl && !(message.attached === 'top' || message.attached === false)
+      ? shouldOverlapAvatarAndMessage
+        ? chatMessageStyles.avatarOverlapRTL
```
palatter, line 315: Can we keep the naming consistent between the one above and this one? I'm good with either name.

So maybe something like:

chatMessageStyles.bodyWithAvatarRTL
chatMessageStyles.bodyWithoutAvatarRTL

palatter (approved):
One minor thought about naming. Otherwise looks good!

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
vhuseinova-msft: Updated

emlynmac (approved):

PR #3619, opened 2023-09-24
default render of participant list item will yield none for presence

alkwa-msft:
# What
Why are showing presence but it is confusing since it looks like M365 presence but it is not. We should be defaulting to none and letting customers choose to put in their own notion of presence.


alkwa-msft:
I was thinking about just wiping away all of presence but it is part of our API and removing it would be a break. We will still keep in all of our storybook documentation. All we are doing with this change is ensuring the default presence is "none" unless further specified by Contoso. 

It brings an interesting question if we should be using these UI components for presence at all. This UI for presence clearly denotes Microsoft "presence" and does it seem odd it could be used for another companies presence or what they ever want to use it at all? @JamesBurnside any thoughts?

JamesBurnside (approved):

JamesBurnside:
> I was thinking about just wiping away all of presence but it is part of our API and removing it would be a break. We will still keep in all of our storybook documentation. All we are doing with this change is ensuring the default presence is "none" unless further specified by Contoso.
> 
> It brings an interesting question if we should be using these UI components for presence at all. This UI for presence clearly denotes Microsoft "presence" and does it seem odd it could be used for another companies presence or what they ever want to use it at all? @JamesBurnside any thoughts?

I like the change as you have it. We want better graph support and having presence is a step in that direction. Presence doesn't have a lot of value in an active call (as you are rarely not "busy" or "do not disturb" in a call) so for callcomposite and callwithchatcomposite its not worth it that much. However for chat composite and the raw UI element I can see more value. But as you said we need to keep it to avoid the API breakage anyways.

dmceachernmsft (approved):

PR #3622, opened 2023-09-25
Shard playwright updates for chat and use Ubuntu 20.04

emlynmac:
# What
Add sharding to the playwright tests for the composites and create a separate chat update-snapshots job.

# Why
Flaky updates.

# How Tested
CI

# Process & policy checklist


.github/workflows/update-chat-snapshots.yml
```diff
@@ -139,12 +392,12 @@ jobs:
         # should not silently overwrite updates to the target branch.
         run: |
           git add packages/react-composites/*.png
-          git commit -m 'Update packages/react-composites ChatComposite browser test snapshots'
-          git pull origin ${{ needs.get_target_branch.outputs.target }} --no-rebase --no-edit
+          git commit -m 'Update packages/react-composites ChatComposite browser test snapshots 4/4'
+          git pull origin ${{ needs.get_target_branch.outputs.target }} --rebase --no-edit
```
vhuseinova-msft, line 396: Rebase is usually not a very good approach for public branches. Do we really need it here?

emlynmac: I was running into issues where one task had committed files out of order.

.github/workflows/update-chat-snapshots.yml
```diff
@@ -52,10 +52,268 @@ jobs:
...
+      - name: Build packages/react-composites browser tests
+        working-directory: ./packages/react-composites
+        run: rushx build:e2e:chat
+      - name: Install packlets used by browser tests
+        run: |
+          cd packages/react-composites
+          rush build -t .
+      - name: Update chat snapshot
+        id: update-chat-snapshots
+        run: rushx test:e2e:chat:update:3
```
vhuseinova-msft, line 274: Can we use matrix here?

PR #3625, opened 2023-09-26
[Chat] Updating storybook anchors to use ref

JoshuaLai:
# What
Investigation on anchors for message thread page. 
Pros:
- ~This implementation works consistently on Safari browsers~
- ~Chromium based browsers it doesn't seem as consistent, but it is an improvement.~

Cons: 
- ~Chromium browsers does still run into issues some times~
- A lot of boilerplate code will need to clean up 

# Why
Anchor in links should be working better with this implementation


packages/storybook/stories/MessageThread/MessageThread.stories.tsx
```diff
@@ -97,7 +97,72 @@ const mentionTag = `
...
+  const refProps = useRef(null);
+
+  const scrollToRef = (ref): void => {
+    setTimeout(() => {
+      ref.current.scrollIntoView({ behavior: 'auto' });
+    }, 1000);
+  };
+
+  useEffect(() => {
+    const url = document.URL;
```
AmyL219, line 125: The scroll issue in Edge and Chrome might stem from the URL not being retrieved correctly on this page. 
For instance, upon entering the URL `http://localhost:6006/?path=/docs/ui-components-messagethread--message-thread#display-file-attachments-with-messages`, the `document.URL` yields "http://localhost:6006/iframe.html?viewMode=docs&id=ui-components-messagethread--message-thread" instead.

By updating from `const url = document.URL;` to `const url = window.top.location.href;`, the functionality is restored in Edge and Chrome. However, this might introduce url fetching failure case in my memory. It would be worthy to confirm with Web expert how to achieve the URL path

packages/storybook/stories/MessageThread/MessageThread.stories.tsx
```diff
@@ -97,7 +97,70 @@ const mentionTag = `
...
+
+  useEffect(() => {
+    const url = window.top ? window.top.location.href : window.location.href;
+    if (url.includes('default-messagethread') && refDefaultMessageThread.current) {
+      scrollToRef(refDefaultMessageThread);
+    } else if (url.includes('messagethread-with-message-date') && refWithMessageDate.current) {
+      scrollToRef(refWithMessageDate);
+    } else if (url.includes('system-message') && refSystemMessage.current) {
+      scrollToRef(refSystemMessage);
+    } else if (url.includes('blocked-message') && refBlockedMessage.current) {
```
AmyL219, line 130: In the URL, both `refWithCustomizedBlockedMessageContainer` and `refBlockedMessage` have the substring `blocked-message`. If you attempt to scroll to `refWithCustomizedBlockedMessageContainer` using the URL `https://60c7ae6891f0e90039d7cd54-jojdfeztms.chromatic.com/?path=/docs/ui-components-messagethread--message-thread#messages-with-customized-blocked-message-container`, it will fail. This is because `refBlockedMessage` matches `blocked-message` first and thus, it scrolls to `refBlockedMessage` before anything else.

To address this, we need to prioritize the more specific path, `messages-with-customized-blocked-message-container`, over the general `blocked-message` path. The revised order should be:

```javascript
if (
  url.includes('messages-with-customized-blocked-message-container') &&
  refWithCustomizedBlockedMessageContainer.current
) {
  scrollToRef(refWithCustomizedBlockedMessageContainer);
} else if (url.includes('blocked-message') && refBlockedMessage.current) {
  scrollToRef(refBlockedMessage);
}
A similar approach may be needed for custom-message.

JoshuaLai: Oh good catch. Yes I based it off of order of item. There might be a case as well for simplicity to group them both under the same header. But I will put in our solution for now

jimchou-dev (approved):

PR #3626, opened 2023-09-26
Fix up padding when action button is not visible

emlynmac:
# What

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/ChatMessage/ChatMessageActionMenu.tsx
+++ b/packages/react-components/src/components/ChatMessage/ChatMessageActionMenu.tsx
@@ -28,27 +28,23 @@ export const chatMessageActionMenuProps = (menuProps: {
   onActionButtonClick: () => void;
   theme: Theme;
 }): ChatMessageActionMenuProps | undefined => {
-  const { enabled, forceShow } = menuProps;
+  const { ariaLabel, enabled, forceShow, theme } = menuProps;
   const showActionMenu = enabled || forceShow;
 
   const actionMenuProps: ChatMessageActionMenuProps = {
     children: (
       <div
-        tabIndex={0} //make div focusable, as Icon below is migrated to v9, this can be deleted
+        tabIndex={showActionMenu ? 0 : undefined} //make div focusable, as Icon below is migrated to v9, this can be deleted
         key="menuButton"
         data-ui-id="chat-composite-message-action-icon"
         ref={menuProps.menuButtonRef}
         onClick={showActionMenu ? () => menuProps.onActionButtonClick() : undefined}
         style={{ margin: showActionMenu ? '1px' : 0, minHeight: showActionMenu ? undefined : '30px' }}
         role="button"
-        aria-label={menuProps.ariaLabel}
+        aria-label={showActionMenu ? ariaLabel : undefined}
       >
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/ChatMessage/ChatMessageActionMenu.tsx
```diff
@@ -39,16 +38,18 @@ export const chatMessageActionMenuProps = (menuProps: {
...
-        <Icon
-          iconName="ChatMessageOptions"
-          aria-label={menuProps.ariaLabel}
-          styles={iconWrapperStyle(menuProps.theme, menuProps.forceShow)}
-        />
+        {showActionMenu ? (
+          <Icon
+            iconName="ChatMessageOptions"
+            aria-label={menuProps.ariaLabel}
+            styles={iconWrapperStyle(menuProps.theme, menuProps.forceShow)}
```
vhuseinova-msft, line 50: we probably can use only `forceShow` (without `menuProp`)

packages/react-components/src/components/ChatMessage/ChatMessageActionMenu.tsx
```diff
@@ -39,16 +38,18 @@ export const chatMessageActionMenuProps = (menuProps: {
         key="menuButton"
         data-ui-id="chat-composite-message-action-icon"
         ref={menuProps.menuButtonRef}
-        onClick={() => menuProps.onActionButtonClick()}
-        style={{ margin: '1px' }}
+        onClick={showActionMenu ? () => menuProps.onActionButtonClick() : undefined}
+        style={{ margin: showActionMenu ? '1px' : 0, minHeight: showActionMenu ? undefined : '30px' }}
         role="button"
         aria-label={menuProps.ariaLabel}
```
vhuseinova-msft, line 44: It looks like it will be visible for accessibility even when it's hidden because of:
- `tabIndex={0}`
- `aria-label={menuProps.ariaLabel}`

emlynmac: Nice catch

packages/react-components/src/components/ChatMessage/ChatMessageActionMenu.tsx
```diff
@@ -28,27 +28,24 @@ export const chatMessageActionMenuProps = (menuProps: {
...
       <div
-        tabIndex={0} //make div focusable, as Icon below is migrated to v9, this can be deleted
+        tabIndex={showActionMenu ? 0 : undefined} //make div focusable, as Icon below is migrated to v9, this can be deleted
         key="menuButton"
         data-ui-id="chat-composite-message-action-icon"
         ref={menuProps.menuButtonRef}
-        onClick={() => menuProps.onActionButtonClick()}
-        style={{ margin: '1px' }}
+        onClick={showActionMenu ? () => menuProps.onActionButtonClick() : undefined}
+        style={{ margin: showActionMenu ? '1px' : 0, minHeight: showActionMenu ? undefined : '30px' }}
```
palatter, line 42: Should we have this hardcoded px value defined in a styles file?

palatter (approved):
Looks good. One small question about where to put style value

PR #3633, opened 2023-09-28
[Chat] Long Display name truncation

JoshuaLai:
# What
Previously long display names would not be truncated and would cover the message thread. 
Implemented new logic to truncate after 50 characters after speaking with design. 
image

# Why
Don't have long display names cover the message thread


packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -147,6 +147,13 @@ const getUsersElement = (
 ): JSX.Element => {
   const userElements: JSX.Element[] = [];
   typingUsers.forEach((user, index) => {
+    if (user.displayName) {
+      let truncatedDisplayName = user.displayName;
+      if (truncatedDisplayName.length > 50) {
+        truncatedDisplayName = truncatedDisplayName.substring(0, 50) + '...';
+      }
+      user.displayName = truncatedDisplayName;
```
jimchou-dev, line 155: does this logic make sense that each user can have up to 50 characters? so if 2 people have long name total typing indicator name would be 100 char long? Just something to think about but this is a start

jimchou-dev (approved):

packages/react-components/src/components/TypingIndicator.tsx
```diff
@@ -147,6 +147,13 @@ const getUsersElement = (
...
+    }
```
Leah-Xia-Microsoft, line 156: Question: if we assign the truncatedDisplayName back to user.displayName, would that change the message object? If so, would it affect other areas such as sender's name in the message bubble?

Leah-Xia-Microsoft, line 156: Maybe we can simply the code as below:
`  typingUsers.forEach((user, index) => {
    let truncatedDisplayName = user.displayName;

    if (truncatedDisplayName && truncatedDisplayName.length > 50) {
      truncatedDisplayName = truncatedDisplayName.substring(0, 50) + '...';
    }
    userElements.push(
      onRenderUser ? (
        onRenderUser(user)
      ) : (
        <Text className={mergeStyles(userDisplayNameStyles)} key={`user-${index}`}>
          {truncatedDisplayName}
        </Text>
      )
    );
    userElements.push(<Text key={`comma-${index}`}>{`${delimiter}`}</Text>);
  });`

packages/react-components/src/components/TypingIndicator.tsx
JoshuaLai: Yep I discussed this with Alex as well. Originally the plan was to have it based on the width %. But it would be difficult and require a refactor of our typing indicator. So instead we will do a char limit on the display name to 50.

packages/react-components/src/components/TypingIndicator.tsx
JoshuaLai: Good suggestion!

Leah-Xia-Microsoft (approved):

PR #3636, opened 2023-09-29
[FluentUI] Correct padding between messages

palatter:
# What

**ISSUE**

image

**CORRECT**

image

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/styles/MessageThread.styles.ts
+++ b/packages/react-components/src/components/styles/MessageThread.styles.ts
@@ -143,7 +143,7 @@ export const useChatMyMessageStyles = makeStyles({
   },
   body: {
     paddingBottom: '10px',
-    marginTop: '0.125rem',
+    marginTop: '1rem',
     maxWidth: '100%',
     minWidth: `${CHAT_MESSAGE_CONTAINER_MIN_WIDTH_REM}rem`,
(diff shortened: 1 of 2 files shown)


emlynmac (approved):

packages/react-components/src/components/styles/MessageThread.styles.ts
```diff
@@ -156,15 +157,19 @@ export const useChatMyMessageStyles = makeStyles({
       visibility: 'visible'
     },
     '& msft-mention': {
-      color: '#D83B01',
+      color: tokens.colorStatusWarningBackground3,
       fontWeight: 600
     }
   },
+  attached: {
```
vhuseinova-msft, line 164: It can be renamed to bodyAttached or something like this to show that it's style for the body

vhuseinova-msft:
It looks like the padding is not the same between Message and MyMessage components ( we should probably decreased the padding for message component as previously it was set to match MyMessage ones)
image

vhuseinova-msft:
Menu button has incorrect position for first MyMessage component

image

palatter:
> Menu button has incorrect position for first MyMessage component
> 
> image

Good catch!

vhuseinova-msft (approved):

PR #3640, opened 2023-09-30
Black border fix for messages

vhuseinova-msft:
# What

Remove the black border for focus-visible selector

# Why

# How Tested

- Open Storybook
- Select Chat message root components
- Set selector to `focus-visible`

# Process & policy checklist


palatter (approved):
Tested safari on Iphone. Looks good.

So is the reason we have to override this style with a 0 border because that is the default appearance for the fluentUI control now?

vhuseinova-msft:
> Tested Safari on iPhone. Also tested Chrome Macbook storybook. Looks good.
> 
> So is the reason we have to override this style with a 0 border because that is the default appearance for the fluentUI control now?

I think it might be FluentUI components changes themselves or the components updates caused an user agent to recognize the need to show the outline. From the :focus-visible documentation:
`The :focus-visible pseudo-class applies while an element matches the :focus pseudo-class and the UA ([User Agent](https://developer.mozilla.org/en-US/docs/Glossary/User_agent)) determines via heuristics that the focus should be made evident on the element.`

PR #3653, opened 2023-10-05
[Bug fix]Decrease captions banner width when screen is too small to prevent more button overlapping with scroll bar 

carocao-msft:
# What
Decrease captions banner width when screen is too small to prevent more button overlapping with scroll bar 

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3413763

# How Tested

https://github.com/Azure/communication-ui-library/assets/96077406/1ac53032-744e-4d7a-b22c-2e66ae4f1cd6

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/common/CaptionsBanner.tsx
+++ b/packages/react-composites/src/composites/common/CaptionsBanner.tsx
@@ -3,7 +3,7 @@
 
 import React from 'react';
 /* @conditional-compile-remove(close-captions) */
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
 /* @conditional-compile-remove(close-captions) */
 import { _CaptionsBanner, _CaptionsBannerStrings } from '@internal/react-components';
 /* @conditional-compile-remove(close-captions) */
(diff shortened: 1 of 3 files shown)


packages/react-composites/src/composites/common/CaptionsBanner.tsx
```diff
@@ -63,6 +61,23 @@ export const CaptionsBanner = (props: { isMobile: boolean }): JSX.Element => {
   const captionsBannerStrings: _CaptionsBannerStrings = {
     captionsBannerSpinnerText: strings.captionsBannerSpinnerText
   };
+  /* @conditional-compile-remove(close-captions) */
+  const { innerWidth: width } = window;
+  /* @conditional-compile-remove(close-captions) */
+  const [windowWidth, setWindowWidth] = useState(width);
```
edwardlee-msft, line 67: Is it possible to use window.innerWidth in this useState?
I'm guessing it's not cause you have line 65.

edwardlee-msft (approved):

prabhjot-msft (approved):

PR #3656, opened 2023-10-05
[Gallery layouts] more gallery cc fixes for beta release

dmceachernmsft:
# What

Same as last PR more CC to fix snapshot updates for beta release
# Why

gets stuck on layout not supported

# How Tested

Tested locally

### DIFF
--- a/packages/react-composites/tests/browser/call/hermetic/VideoGallery.test.ts
+++ b/packages/react-composites/tests/browser/call/hermetic/VideoGallery.test.ts
@@ -168,8 +168,11 @@ test.describe('VideoGallery tests', async () => {
     expect(await stableScreenshot(page)).toMatchSnapshot('gallery-controls-open.png');
     await page.locator('button:has-text("Speaker")').click();
     expect(await stableScreenshot(page)).toMatchSnapshot('speaker-layout.png');
+    /* @conditional-compile-remove(gallery-layout-composite) */
     await pageClick(page, dataUiId(IDS.moreButton));
+    /* @conditional-compile-remove(gallery-layout-composite) */
     await page.locator('button:has-text("View")').click();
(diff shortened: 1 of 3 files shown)


palatter (approved):

JoshuaLai (approved):

JamesBurnside (approved):

packages/react-composites/tests/browser/call/hermetic/VideoGallery.test.ts
```diff
@@ -255,6 +258,7 @@ test.describe('VideoGallery tests', async () => {
 
     await waitForSelector(page, dataUiId(IDS.moreButton));
     await pageClick(page, dataUiId(IDS.moreButton));
+    await page.locator('span:has-text("View")');
```
JamesBurnside, line 261: Is this needed?

JamesBurnside: Or should the below line be `await (await page.locator('...')).click()`

dmceachernmsft: just to help with flakeyness seems to get too eager if it only has the one. I think we need to take another look at the mobile snapshots and interacting with the bottom sheet drawer since we don't use the data-ui-id's there.

JamesBurnside: Hmm, yeah dont need to fix in this PR, but if thats the case this will just hide the flakiness more and not garaunteed to fix it

palatter: Going to disable auto merge in case you need to change this @dmceachernmsft

PR #3671, opened 2023-10-11
Cherry Pick 1.9.0-Beta.1 Pre Release into 1.9.0 Release

palatter:
Cherry picking the pre release branch from 1.9.0-beta.1 as it was not merged into main before the 1.9.0 stable branch was cut.


dmceachernmsft:
Since this happened does that mean that the versioning in the package.json's is incorrect in the new release?

palatter:
> > Since this happened does that mean that the versioning in the package.json's is incorrect in the new release?
> 
> My assumption is this is because we created a stable release before the merge of the beta prerelease branch back into main?

Yes. That's correct. We are missing the step in the instructions for merging the pre-release back to the release. I think this PR was created automatically at one point with the workflow but due to security restrictions, it must be done manually now.

PR #3672, opened 2023-10-11
[Chat] Menu button accessibility fix

vhuseinova-msft:
# What

- fix an issue when a menu button wasn't accessible sometimes with Narrator 
- revert some changes for disabling chat messages groups for Narrator to match the previous behaviour

# Why

# How Tested

Storybook

# Process & policy checklist


palatter (approved):
Looks good. Tested with Chrome on Mac.

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
```diff
@@ -404,16 +400,18 @@ const MessageBubble = (props: ChatMessageComponentAsMessageBubbleProps): JSX.Ele
...
-              className: chatItemMessageContainerClassName,
-              style: { ...createStyleFromV8Style(messageContainerStyle) },
+            root={{
+              className: chatMessageStyles.root,
               // make body not focusable to remove repetitions from narrators.
               // inner components are already focusable
               tabIndex: -1,
               role: 'none'
             }}
+            author={<Text className={chatMessageAuthorStyle}>{message.senderDisplayName}</Text>}
```
emlynmac, line 410: Nit: it would be nice to have these alphabetized, or at least the same line as the previous version so comparing is easier.

emlynmac (approved):

packages/react-components/src/components/ChatMessage/ChatMessageComponentAsMessageBubble.tsx
vhuseinova-msft: As discussed, will leave as it is for now and we can review this later (possible with adding a rule to linter)

PR #3673, opened 2023-10-11
Removed workaround for mouseDown event in jest tests

vhuseinova-msft:
# What

- Removed workaround for mouseDown event in jest tests
- Removed duplicated strict mode from Calling sample app

# Why

# How Tested

Ran tests

# Process & policy checklist

### DIFF
--- a/samples/Calling/src/app/App.tsx
+++ b/samples/Calling/src/app/App.tsx
@@ -204,18 +204,16 @@ const App = (): JSX.Element => {
         return <Spinner label={'Getting user credentials from server'} ariaLive="assertive" labelPosition="top" />;
       }
       return (
-        <React.StrictMode>
-          <CallScreen
-            token={token}
-            userId={userId}
-            displayName={displayName}
-            callLocator={callLocator}
-            /* @conditional-compile-remove(PSTN-calls) */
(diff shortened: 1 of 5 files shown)


samples/Calling/src/app/App.tsx
```diff
@@ -204,18 +204,16 @@ const App = (): JSX.Element => {
         return <Spinner label={'Getting user credentials from server'} ariaLive="assertive" labelPosition="top" />;
       }
       return (
-        <React.StrictMode>
```
JamesBurnside, line 207: keep strict mode?

JamesBurnside (approved):

emlynmac (approved):
We should leave strict mode enabled but good to remove the workaround

samples/Calling/src/app/App.tsx
vhuseinova-msft: @JamesBurnside, @emlynmac it is enabled in `index.tsx` for all Sample apps. The one that is deleted in this PR, is additional call in `App.tsx` file which is not needed anymore.

JamesBurnside: Ah! 💖 the extra cleanup!

PR #3674, opened 2023-10-11
[Dependency updates] calling bindings fix

dmceachernmsft:
# What

Updates Call Adapter to use the new types in start call so that it can make calls to Auto attendants, Call Queues and Teams users
# Why

Users were seeing issues when trying to make calls to Call Queues and AA's since there was no cloud value in the identifier when passed to start call as an `UnknownIdentifier`

# How Tested

Tested locally


packages/calling-component-bindings/review/beta/calling-component-bindings.api.md
```diff
@@ -190,7 +191,7 @@ export interface CommonCallingHandlers {
     // (undocumented)
     onSetSpokenLanguage: (language: string) => Promise<void>;
     // (undocumented)
-    onStartCall: (participants: (CommunicationUserIdentifier | PhoneNumberIdentifier | UnknownIdentifier)[], options?: StartCallOptions) => void;
+    onStartCall: (participants: (CommunicationUserIdentifier | PhoneNumberIdentifier | UnknownIdentifier | MicrosoftTeamsAppIdentifier)[], options?: StartCallOptions) => void;
```
alkwa-msft, line 194: why cant we just make this CommunicationIdentifier and save ourselves from having to consider every type? Is it because we just dont want to support calling a teams user directly?

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -882,6 +883,11 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
         return backendId as PhoneNumberIdentifier;
       } else if (isCommunicationUserIdentifier(backendId)) {
         return backendId as CommunicationUserIdentifier;
+      } else if (isMicrosoftTeamsAppIdentifier(backendId)) {
+        return backendId as MicrosoftTeamsAppIdentifier;
+      } else if (isMicrosoftTeamsUserIdentifier(backendId)) {
+        console.log(backendId);
```
alkwa-msft, line 889: can we remove this console.log?

alkwa-msft (approved):

edwardlee-msft (approved):

PR #3676, opened 2023-10-12
[Bug Fix] Fix issue where Call with Chat adapter would create multiple clients

dmceachernmsft:
# What

Create reference in `useAzureCommunicationCallWithChatAdapter` to make sure that we do not make a second adapter if we are in the middle of creating an adapter already
# Why

We do this already with the CallAdapter fixes old issue in this adapter as well

Also removes the open in other tab logic from sample apps for now as this is not something that should be done with strict mode. Calling SDK supports this functionality and will be used in future.

# How Tested

Tested locally 
image


samples/Calling/src/app/App.tsx
```diff
@@ -84,17 +82,17 @@ const App = (): JSX.Element => {
...
   useEffect(() => {
     if (isMobileSession && isLandscapeSession) {
       console.log('ACS Calling sample: Mobile landscape view is experimental behavior');
     }
   }, [isMobileSession, isLandscapeSession]);
 
-  if (isMobileSession && isAppAlreadyRunningInAnotherTab) {
-    return <PageOpenInAnotherTab />;
-  }
+  // if (isMobileSession && isAppAlreadyRunningInAnotherTab) {
```
JamesBurnside, line 93: remove

samples/Calling/src/app/App.tsx
```diff
@@ -84,17 +82,17 @@ const App = (): JSX.Element => {
 
   const isMobileSession = useIsMobile();
   const isLandscapeSession = isLandscape();
-  const isAppAlreadyRunningInAnotherTab = useSecondaryInstanceCheck();
+  // const isAppAlreadyRunningInAnotherTab = useSecondaryInstanceCheck();
```
prabhjot-msft, line 85: commented code

prabhjot-msft (approved):

PR #3678, opened 2023-10-12
[Bugfix] Fix issue where bottom sheet was not selecting first item

dmceachernmsft:
# What

Updates the keys of each of the drawer items to be unique to the drawer item and adds a focus zone to each
image

# Why

Makes React stop focusing on an item so the focus zone can pull focus to the first item in the list

https://skype.visualstudio.com/SPOOL/_workitems/edit/3426604
# How Tested

locally


packages/react-components/src/components/Drawer/DrawerMenuItem.tsx
```diff
@@ -49,6 +50,10 @@ export interface _DrawerMenuItemProps {
    * A unique id set for the standard HTML id attibute
    */
   id?: string;
+  /**
+   * Property to set the focus since this is the first item in the menu
+   */
+  shouldFocusFirst?: boolean;
```
JamesBurnside, line 56: nit: `shouldFocusOnMount` instead? As this is a single menuItem, it doesn't really have the concept of "first"

JamesBurnside (approved):

edwardlee-msft (approved):

PR #3679, opened 2023-10-12
[Chat] Updating the API to cache based on MessageId

JoshuaLai:
# What
Updating to caching of the inline images to be by message id instead of attachment id. 
This API allows us to fetch all images at the same time

# Why
Improve performance of the composite when we fetch images. 
This reduces the amount of re-renders we perform on the MessageThread component. 

# How Tested
Check the network tab on browser inspector. 
Make sure the images on only being requested once. 

# Process & policy checklist

Breaking changes on the download attachments API


change-beta/@azure-communication-react-8ae2b34f-b599-4075-95df-5298be05892f.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "prerelease",
+  "area": "improvement",
+  "workstream": "InlineImages",
+  "comment": "Updating to map to attachments to message id",
```
Leah-Xia-Microsoft, line 5: We might want to have the changelog starting with Update instead of updating.

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -78,16 +80,18 @@ const MessageContentWithLiveAria = (props: MessageContentWithLiveAriaProps): JSX
...
-      ) {
-        props.onFetchAttachment(fileMetadata);
-      }
+    const attachments = props.message.attachedFilesMetadata?.filter((fileMetadata) => {
+      return fileMetadata.attachmentType === 'inlineImage';
     });
+
+    if (props.attachmentsMap && attachments) {
+      attachments.forEach((fileMetadata) => {
+        if (props.onFetchAttachment && props.attachmentsMap && props.attachmentsMap[fileMetadata.id] === undefined) {
```
Leah-Xia-Microsoft, line 89: We might be able to remove the check for props.attachmentsMap since it's been checked above.

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -286,11 +287,12 @@ describe('Message should display image and attachment correctly', () => {
         }
       ]
     };
-    const onFetchAttachment = async (attachment: FileMetadata): Promise<AttachmentDownloadResult[]> => {
+    const onFetchAttachment = async (attachment: FileMetadata[]): Promise<AttachmentDownloadResult[]> => {
```
jimchou-dev, line 290: array should use plural `attachments`

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -33,7 +35,7 @@ type ChatMessageContentProps = {
   /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
   attachmentsMap?: Record<string, string>;
   /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
-  onFetchAttachment?: (attachment: FileMetadata) => Promise<void>;
+  onFetchAttachment?: (attachment: FileMetadata[], message: Message) => Promise<void>;
```
Leah-Xia-Microsoft, line 38: Should we rename attachment into attachments since it takes an array now?

packages/storybook/stories/MessageThread/snippets/WithInlineImageMessage.snippet.tsx
```diff
@@ -14,16 +14,17 @@ import React, { useState } from 'react';
 export const MessageThreadWithInlineImageExample: () => JSX.Element = () => {
   const [galleryImages, setGalleryImages] = useState<Array<ImageGalleryImageProps>>([]);
 
-  const onFetchAttachment = async (attachment: FileMetadata): Promise<AttachmentDownloadResult[]> => {
+  const onFetchAttachment = async (attachment: FileMetadata[]): Promise<AttachmentDownloadResult[]> => {
```
jimchou-dev, line 17: array should use plural `attachments`

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -216,11 +216,12 @@ describe('Message should display image and attachment correctly', () => {
         }
       ]
     };
-    const onFetchAttachment = async (attachment: FileMetadata): Promise<AttachmentDownloadResult[]> => {
+    const onFetchAttachment = async (attachments: FileMetadata[]): Promise<AttachmentDownloadResult[]> => {
```
Leah-Xia-Microsoft, line 219: Missing message prop?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -909,21 +909,25 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
   const [readCountForHoveredIndicator, setReadCountForHoveredIndicator] = useState<number | undefined>(undefined);
 
   /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
-  const [inlineAttachments, setInlineAttachments] = useState<Record<string, string>>({});
+  const [inlineAttachments, setInlineAttachments] = useState<Record<string, Record<string, string>>>({});
   /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
   const onFetchInlineAttachment = useCallback(
-    async (attachment: FileMetadata): Promise<void> => {
-      if (!onFetchAttachments || attachment.attachmentType !== 'inlineImage' || attachment.id in inlineAttachments) {
+    async (attachment: FileMetadata[], message: Message): Promise<void> => {
```
Leah-Xia-Microsoft, line 915: Same for the attachment -> attachments

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -78,16 +80,18 @@ const MessageContentWithLiveAria = (props: MessageContentWithLiveAriaProps): JSX
...
-      if (
-        props.onFetchAttachment &&
-        props.attachmentsMap &&
-        fileMetadata.attachmentType === 'inlineImage' &&
-        props.attachmentsMap[fileMetadata.id] === undefined
-      ) {
-        props.onFetchAttachment(fileMetadata);
-      }
+    const attachments = props.message.attachedFilesMetadata?.filter((fileMetadata) => {
+      return fileMetadata.attachmentType === 'inlineImage';
```
jimchou-dev, line 84: keep in mind, future GA will just have attachmantType as `file` and `image`, determining if an image is inlineImage will be based on attachmentType == image AND content contains  that that has matching id in chat.content.attachments

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -909,21 +909,25 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
...
 
-      setInlineAttachments((prev) => ({ ...prev, [attachment.id]: '' }));
+      const messageId = message.messageId;
       const attachmentDownloadResult = await onFetchAttachments(attachment);
-      if (attachmentDownloadResult[0]) {
-        setInlineAttachments((prev) => ({
-          ...prev,
-          [attachment.id]: attachmentDownloadResult[0].blobUrl
-        }));
+      const list = inlineAttachments[messageId];
```
jimchou-dev, line 922: use a better naming than just `list` sounds too generic

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -909,21 +909,25 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
...
   const onFetchInlineAttachment = useCallback(
-    async (attachment: FileMetadata): Promise<void> => {
-      if (!onFetchAttachments || attachment.attachmentType !== 'inlineImage' || attachment.id in inlineAttachments) {
+    async (attachment: FileMetadata[], message: Message): Promise<void> => {
+      if (!onFetchAttachments) {
         return;
       }
 
-      setInlineAttachments((prev) => ({ ...prev, [attachment.id]: '' }));
+      const messageId = message.messageId;
```
Leah-Xia-Microsoft, line 920: If all we need here is messageId, should we only pass in messageId instead of the whole message?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -1202,7 +1210,7 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
             /* @conditional-compile-remove(image-gallery) */
             onInlineImageClicked={onInlineImageClicked}
             /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
-            attachmentsMap={inlineAttachments}
+            attachmentsMap={inlineAttachments[messageProps.message.messageId]}
```
jimchou-dev, line 1213: can we do something like to avoid setting the props in 1184
```
attachmentsMap={inlineAttachments[messageProps.message.messageId] ?? {} }
```
since most message will not have image

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -909,21 +909,25 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
...
       const attachmentDownloadResult = await onFetchAttachments(attachment);
-      if (attachmentDownloadResult[0]) {
-        setInlineAttachments((prev) => ({
-          ...prev,
-          [attachment.id]: attachmentDownloadResult[0].blobUrl
-        }));
+      const list = inlineAttachments[messageId];
+      for (const result of attachmentDownloadResult) {
+        const attachmentId = result.attachmentId;
+        const blobUrl = result.blobUrl;
```
Leah-Xia-Microsoft, line 925:
```suggestion
        const { attachmentId, blobUrl } = result;
```

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -909,21 +909,25 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
...
+      const messageId = message.messageId;
       const attachmentDownloadResult = await onFetchAttachments(attachment);
-      if (attachmentDownloadResult[0]) {
-        setInlineAttachments((prev) => ({
-          ...prev,
-          [attachment.id]: attachmentDownloadResult[0].blobUrl
-        }));
+      const list = inlineAttachments[messageId];
+      for (const result of attachmentDownloadResult) {
+        const attachmentId = result.attachmentId;
```
Leah-Xia-Microsoft, line 924:
```suggestion
       const { attachmentId, blobUrl } = result;
```

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -216,12 +216,20 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
-      }
-      return [{ blobUrl: '' }];
+    async (attachment: FileMetadata[]): Promise<AttachmentDownloadResult[]> => {
+      const targets = attachment.filter((obj) => {
+        return obj.attachmentType === 'inlineImage' && obj.previewUrl;
+      });
+
+      const entry: Record<string, string> = {};
+      targets.forEach((target) => {
+        if (target.attachmentType === 'inlineImage' && target.previewUrl) {
```
Leah-Xia-Microsoft, line 226: We've checked this in the filter function above

packages/react-composites/src/composites/ChatComposite/ChatScreen.tsx
```diff
@@ -216,12 +216,20 @@ export const ChatScreen = (props: ChatScreenProps): JSX.Element => {
...
-    async (attachment: FileMetadata): Promise<AttachmentDownloadResult[]> => {
-      if (attachment.attachmentType === 'inlineImage' && attachment.previewUrl) {
-        const blob = await adapter.downloadAttachments({ attachmentUrls: [attachment.previewUrl] });
-        return blob;
-      }
-      return [{ blobUrl: '' }];
+    async (attachment: FileMetadata[]): Promise<AttachmentDownloadResult[]> => {
+      const targets = attachment.filter((obj) => {
+        return obj.attachmentType === 'inlineImage' && obj.previewUrl;
+      });
```
Leah-Xia-Microsoft, line 222: Seems like this filter function is unnecessary since in the forEach loop below, we can just loop through the attachment instead of the targets.

Leah-Xia-Microsoft:
Since the downloadAttachments function from adapter has changes, should we also update the onInlineImageClicked function which uses adapter.downloadAttachments as well?

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -149,6 +149,8 @@ export type AreTypeEqual<A, B> = A extends B ? (B extends A ? true : false) : fa
 
 // @beta (undocumented)
 export interface AttachmentDownloadResult {
+    // (undocumented)
```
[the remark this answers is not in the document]
JoshuaLai: This is the APIView but good point I should add a document to the interface

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
JoshuaLai: Yep we will make this refactor after we GA the new chat sdk with this change

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
JoshuaLai: This throw a linter error (and honestly I am not sure why) about the attachmentsMap being undefined. So I will need to leave this condition in there

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -78,16 +80,18 @@ const MessageContentWithLiveAria = (props: MessageContentWithLiveAriaProps): JSX
...
     });
+
+    if (props.attachmentsMap && attachments) {
+      attachments.forEach((fileMetadata) => {
+        if (props.onFetchAttachment && props.attachmentsMap && props.attachmentsMap[fileMetadata.id] === undefined) {
+          props.onFetchAttachment([fileMetadata], props.message);
```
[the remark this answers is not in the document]
JoshuaLai: It doesn't matter, its actually a bit more efficient this way because as long as a message is even edited the attachment id changes for all items. So if we detect any change at all we will need to refresh. I suppose if we wanted to be really efficient all we need to do was to check if the first item has change and if so make the full request. But IMO this is just slightly safer.

packages/react-components/src/components/MessageThread.test.tsx
JoshuaLai: No this one does not require the message prop

packages/react-components/src/components/MessageThread.tsx
JoshuaLai: Good idea!

Leah-Xia-Microsoft (approved):
LGTM

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -909,21 +909,22 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
...
   /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
   const onFetchInlineAttachment = useCallback(
-    async (attachment: FileMetadata): Promise<void> => {
-      if (!onFetchAttachments || attachment.attachmentType !== 'inlineImage' || attachment.id in inlineAttachments) {
+    async (attachments: FileMetadata[], messageId: string): Promise<void> => {
+      if (!onFetchAttachments) {
         return;
       }
+      const attachmentDownloadResult = await onFetchAttachments(attachments);
+      const listOfAttachments = inlineAttachments[messageId];
```
jimchou-dev, line 920: Maybe do this instead if message id is not yet in the map
```
const listOfAttachments = inlineAttachments[messageId] ?? {}
```
So the `if` check in line 1183 can be removed

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -1177,6 +1178,10 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
   // To rerender the defaultChatMessageRenderer if app running across days(every new day chat time stamp need to be regenerated)
   const defaultChatMessageRenderer = useCallback(
     (messageProps: MessageProps) => {
+      /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+      if (inlineAttachments[messageProps.message.messageId] === undefined) {
+        setInlineAttachments((prev) => ({ ...prev, [messageProps.message.messageId]: {} }));
+      }
```
jimchou-dev, line 1184: In conjunction in the comment above for line 920, this `if` check here can be removed? since if it's not in the dictionary/object, it will take `inlineAttachments[messageProps.message.messageId] ?? {}` to be empty automatically?

jimchou-dev (approved):

PR #3694, opened 2023-10-19
[Dependency Updates] Update eslint, prettier and roll back rush to .108.0

emlynmac:
# What
Dependency updates

# Why
Keeping updated with latest

# How Tested
CI, local build.

# Process & policy checklist


common/config/rush/command-line.json
```diff
@@ -6,8 +6,8 @@
       "name": "prettier",
       "summary": "Used by the pre-commit Git hook. This command invokes Prettier to reformat staged changes.",
       "autoinstallerName": "rush-prettier",
-      "shellCommand": "pretty-quick --staged",
-      "safeForSimultaneousRushProcesses": true
+      "shellCommand": "npx lint-staged",
```
emlynmac, line 9: pretty-quick works only with prettier v2. For prettier v3, we use lint-staged.

packages/calling-stateful-client/src/TestUtils.ts
```diff
@@ -347,7 +347,7 @@ export const createMockLocalVideoStream = (): LocalVideoStream =>
     switchSource: Promise.resolve,
     /* @conditional-compile-remove(video-background-effects) */
     feature: () => createMockVideoEffectsAPI()
-  } as unknown as LocalVideoStream);
+  }) as unknown as LocalVideoStream;
```
vhuseinova-msft, line 350: Just wanted to confirm that the meaning of the expression stays the same

packages/communication-react/breaking-change-check/checkBreakingChanges.ts
```diff
@@ -78,7 +78,7 @@ type BreakingChangeCallClientStateTest = CheckBreakingChanges<
 >;
 
 // Stateful client test start
-
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-constraint
 type StatefulClientTestFunction<Client extends any> = (client: NestedPartial<OmitState<Partial<Client>>>) => Client;
```
vhuseinova-msft, line 82: Can we just remove `extends any` here?

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -1913,7 +1913,7 @@ export const COMPONENT_LOCALE_ZH_CN: ComponentLocale;
 export const COMPONENT_LOCALE_ZH_TW: ComponentLocale;
 
 // @public
-export type ComponentIcons = Record<keyof typeof DEFAULT_COMPONENT_ICONS, JSX.Element>;
+export type ComponentIcons = Record<keyof typeof DEFAULT_COMPONENT_ICONS, React_2.JSX.Element>;
```
vhuseinova-msft, line 1916: I'm wondering if `React_2.` can be deleted here too

emlynmac: This is a generated file.

packages/calling-stateful-client/src/TestUtils.ts
emlynmac: It's a change in prettier / eslint syntax

packages/communication-react/review/beta/communication-react.api.md
vhuseinova-msft: Yep,  just noticed that React_2 was deleted in most of the places and wanted to check if we can do anything with this  one too

vhuseinova-msft (approved):

packages/communication-react/breaking-change-check/checkBreakingChanges.ts
emlynmac: this would be a question for @PorterNan. IIRC, the casting was part of the determining if there is a break.

edwardlee-msft (approved):

packages/react-components/src/theming/icons.tsx
```diff
@@ -131,7 +132,7 @@ const SitePermissionMic20Filled = (): JSX.Element => {
 };
 
 /* @conditional-compile-remove(call-readiness) */
-const SitePermissionCamera20Filled = (): JSX.Element => {
+const SitePermissionCamera20Filled = (): React.JSX.Element => {
```
edwardlee-msft, line 135: Noticed that all these have React prepended now.
Can we import in order not to have to call React.JSX.Element all the time rather just JSX.Element?

emlynmac: This is a deprecation. We should be adopting this throughout.

PR #3699, opened 2023-10-20
Add ability to add a logo to the configuration page

JamesBurnside:
# What
Add ability to add a logo to the configuration page

# Why
Custom branding feature

# How Tested
See snapshot updates, example:

image


mgamis-msft (approved):

prabhjot-msft (approved):
Looks good. Just a small question, are we adding any check for logo URL, like if it is invalid or any size constraints?

PR #3702, opened 2023-10-20
Alkwa/fix release cleanup

alkwa-msft:
# What
There was an issue with referencing the clean up script. I also wanted to clean up the name of the script to align with the other scripts. Finally we also included the script in an unintended area so we are cleaning that up

# Why
It was breaking Ci and not allowing us to deploy releases.

# How Tested
Tested building out an alpha release

# Process & policy checklist


rush.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/rush.schema.json",
-  "rushVersion": "5.109.1",
```
JamesBurnside, line 3: revert change to rush.json, this is being handled in its own PR #3700

.github/workflows/alpha-release.yml
```diff
@@ -84,8 +84,7 @@ jobs:
 
       # Clean package.json
       - name: Remove scripts, devDependencies, and beachball from package.json
-        run: node ./common/scripts/preparePackageJSON.js
-        working-directory: ./packages/communication-react
```
JamesBurnside, line 88: Without this, how does it know to change the package.json in communication-react?

image

rush.json
alkwa-msft: right i was going to remove this

.github/workflows/alpha-release.yml
alkwa-msft: ill need to look further into this

rush.json
emlynmac: Looks like the issue may not be the rush update after all.
There's an issue with `rush switch-flavor:beta-release` and it looks like the only real difference between 109 and previous rush versions is that the error is now reported differently.

.github/workflows/alpha-release.yml
alkwa-msft: Should be fixed now

rush.json
alkwa-msft: I have reverted the 5.109 change so we should be good here

JamesBurnside (approved):

emlynmac (approved):

PR #3704, opened 2023-10-20
Fix exit -1 problem

PorterNan:
# What

Fix exit -1 problem

rushx is just not smart enough to distinguish an intentionally failed command from a real failed command - we use npm run instead to solve the problem

# Why

# How Tested

# Process & policy checklist


edwardlee-msft (approved):

palatter (approved):

emlynmac:
@PorterNan  do we need the breaking-change label or should the underlying fix be done first?

PorterNan:
> @PorterNan do we need the breaking-change label or should the underlying fix be done first?


That fix will be merged here - and we will need a breaking change label anyway, because we are going to remove an unnecessary class from the api file

jpeng-ms (approved):

emlynmac (approved):
Approved with the API change incoming

PR #3659, opened 2023-10-05
[Sample apps] add strict mode to sample apps again

dmceachernmsft:
# What

add strict mode to sample apps again
# Why

N* is gone! wohoo!

# How Tested

Tested samples locally for strict mode errors

PR #3708, opened 2023-10-23
[Bugfix] Fix missing DownloadFileCards tooltip

jimchou-dev:
# What

- Add missing DownloadFileCards tooltip (with existing localization support
- Tooltip is not working on Samples due to RestrictMode (related PR: https://github.com/Azure/communication-ui-library/pull/3659/files)
image

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/FileDownloadCards.tsx
+++ b/packages/react-components/src/components/FileDownloadCards.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { Icon, IconButton, Spinner, SpinnerSize } from '@fluentui/react';
+import { Icon, IconButton, Spinner, SpinnerSize, TooltipHost } from '@fluentui/react';
 import React, { useCallback, useState } from 'react';
 import { useMemo } from 'react';
 /* @conditional-compile-remove(file-sharing) */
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -271,23 +271,25 @@ export const _FileDownloadCards = (props: _FileDownloadCards): JSX.Element => {
...
-                    /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
-                    isShowDownloadIcon(file) ? (
-                    <IconButton className={iconButtonClassName} ariaLabel={downloadFileButtonString()}>
-                      <DownloadIconTrampoline />
-                    </IconButton>
-                  ) : undefined
-                }
-                actionHandler={() => fileDownloadHandler(userId, file)}
-              />
+              <TooltipHost content={downloadFileButtonString()} key={file.name}>
```
edwardlee-msft, line 274: Is file.name as the key sufficient enough?
Can Contoso upload a file with the same name?

edwardlee-msft (approved):

packages/react-components/src/components/FileDownloadCards.tsx
jimchou-dev: I don't think there's any check on this for when upload, but this is existing logic with _FileCard, just extracted out an layer to work with TooltipHost. Not in scope for this tooltip bugfix

palatter (approved):

PR #3719, opened 2023-10-26
Change display name for remote participants as Attendee if hide attendee name is on

prabhjot-msft:
# What
Change display name for remote participants as Attendee if hide attendee name is on
Screenshot 2023-10-31 at 12 51 09 PM

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3460651
https://skype.visualstudio.com/SPOOL/_workitems/edit/3460682

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/Converter.ts
```diff
@@ -155,7 +167,10 @@ export function convertSdkCallToDeclarativeCall(call: CallCommon): CallState {
     /* @conditional-compile-remove(optimal-video-count) */
     optimalVideoCount: {
       maxRemoteVideoStreams: call.feature(Features.OptimalVideoCount).optimalVideoCount
-    }
+    },
+    /* @conditional-compile-remove(hide-attendee-name) */
+    // TODO: Replace this once the SDK supports hide attendee name
+    hideAttendeeNames: false
```
alkwa-msft, line 173: how are we going to set true/false at the API layer? it looks like its only false for now? Do we have a plan?

packages/calling-stateful-client/src/Converter.ts
```diff
@@ -108,6 +108,18 @@ export function convertSdkCallToDeclarativeCall(call: CallCommon): CallState {
...
   });
+  /* @conditional-compile-remove(hide-attendee-name) */
+  // TODO: Replace this once the SDK supports hide attendee name
+  const hideAttendeeNames = false;
+  /* @conditional-compile-remove(hide-attendee-name) */
+  if (hideAttendeeNames) {
+    call.remoteParticipants.forEach((participant: SdkRemoteParticipant) => {
+      const declarativeParticipant = call.remoteParticipants[toFlatCommunicationIdentifier(participant.identifier)];
+      if (declarativeParticipant.role === 'Attendee') {
+        declarativeParticipant.displayName = participant.role;
```
JamesBurnside, line 119: We can't set displayName directly to the role as the role isn't localized, instead the UI should map the role to a localized version

packages/calling-stateful-client/src/Converter.ts
prabhjot-msft: Yeas, we are going to add the flag for `hideAttendeeName` in the call object from SDK side and we can use that flag here for the call data. Right now setting it to false and using this hardcoded prop to test the changes

packages/calling-stateful-client/src/Converter.ts
prabhjot-msft: Yeah, that was left by mistake. I have moved the logic to selectors(Video gallery selector & Participant List selector) but still unsure of how to pass locale in selectors.

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -43,13 +48,23 @@ export const _videoGalleryRemoteParticipantsMemo = (
         })
         .map((participant: RemoteParticipantState) => {
           const state = _isRingingPSTNParticipant(participant);
+          let displayName = participant.displayName;
+          /* @conditional-compile-remove(hide-attendee-name) */
+          if (isHideAttendeeNamesEnabled && participant.role === 'Attendee') {
```
JamesBurnside, line 53: This logic is in both participantlistselector and videogalleryselector, to ensure these always are consistent, extract to a helper function, e.g., `maskDisplayNameWithRole(displayName, localUserRole, participantRole)`

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -489,6 +489,11 @@ export interface CallState {
    * Proxy of {@link @azure/communication-calling#CapabilitiesFeature}.
    */
   capabilitiesFeature?: CapabilitiesFeatureState;
+  /* @conditional-compile-remove(hide-attendee-name) */
+  /**
+   * Hide attendee names in teams meeting
+   */
+  hideAttendeeNames?: boolean;
```
JamesBurnside, line 496: this will likely update when the calling sdk property rolls in so this is a future suggestion and can ignoree for this pr, but maskAttendeeNames is slightly better as we can't hide them completely

packages/react-components/src/components/RemoteVideoTile.tsx
```diff
@@ -173,7 +189,8 @@ export const _RemoteVideoTile = React.memo(
           userId={userId}
           initialsName={remoteParticipant.displayName ?? ''}
           renderElement={renderVideoStreamElement}
-          displayName={remoteParticipant.displayName || strings?.displayNamePlaceholder}
+          // displayName={remoteParticipant.displayName || strings?.displayNamePlaceholder}
```
JamesBurnside, line 192: remove comment

JamesBurnside (approved):

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
prabhjot-msft: done

packages/react-components/src/components/RemoteVideoTile.tsx
prabhjot-msft: done

packages/calling-stateful-client/src/CallClientState.ts
prabhjot-msft: Yeah once sdk changes are added, will update the name

packages/react-components/src/localization/LocalizationProvider.tsx
```diff
@@ -174,6 +174,9 @@ export interface ComponentStrings {
   /* @conditional-compile-remove(total-participant-count) */
   /** Strings for the participant list component */
   ParticipantList: ParticipantListStrings;
+  /* @conditional-compile-remove(hide-attendee-name) */
+  /** Strings for the attendee role */
+  AttendeeRole: string;
```
mgamis-msft, line 179: Can we make an interface AttendeeRoleStrings here? Just in case we have to add more strings for this feature

mgamis-msft (approved):

packages/react-components/src/localization/LocalizationProvider.tsx
prabhjot-msft: Yeah, I agree. Will create a interface in next PR.

PR #3726, opened 2023-10-31
Update call composite configuration page to newer designs

JamesBurnside:
# What

- Better align preview and controls
- More compact start button
- Add a background to the controls to prevent text being lost on custom background

# Why

See [figma](https://www.figma.com/file/72qJ03EOQExf12g5JTdWjQ/Web-UI-Library?type=design&node-id=11247-4357&mode=design&t=wqIOU0GR2egZHt20-0)

# How Tested

- See updated snapshots
- Locally verified responsive scaling

Example:
image


edwardlee-msft (approved):

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
```diff
@@ -340,15 +343,15 @@ export const ConfigurationPage = (props: ConfigurationPageProps): JSX.Element =>
...
             horizontalAlign={mobileWithPreview ? 'stretch' : 'center'}
             verticalFill={mobileWithPreview}
-            tokens={mobileWithPreview ? configurationStackTokensMobile : undefined}
+            tokens={configurationStackTokensMobile}
           >
             {localPreviewTrampoline(
               mobileWithPreview,
               /* @conditional-compile-remove(rooms) */ !!(role === 'Consumer')
             )}
-            <Stack className={mobileView ? undefined : selectionContainerStyle}>
```
dmceachernmsft, line 349: are these tokens just always used because there is now a new style on line 354 that changes some of the spacing?

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/pages/ConfigurationPage.tsx
JamesBurnside: Before they were just applied to mobile because the start button was inside the stack with the dropdowns. Now the start call button exists outside that stack it uses the same token spacing and mobile to create that space between the start call button and the content above it

JamesBurnside: I think.. there's so many stacks floating around here, open to feedback if there's a nicer way to structure this haha

PR #3729, opened 2023-11-01
Custom Button Injection - Update Storybook with copy/paste code snippets and agreed upon usage of _MockCallAdapter

edwardlee-msft:
# What

New storybook snippets that are copy paste-able for Contoso.
Use import of _MockCallAdapter and ignore linter rules of @internal import. (as long as the code is not copy/paste-able by contoso, it is fine to this as agreed upon by ui lib team)

# Why

Copy paste errors can occur. Want seamless experience.
Want a more centralized location of _MockCallAdapter instead of having duplicate code.

# How Tested

MacOS local storybook

# Process & policy checklist


dmceachernmsft (approved):

packages/storybook/stories/_MockCallAdapter.ts
mgamis-msft, line 1: Thanks for doing this. We don't want to have another copy of this.

mgamis-msft (approved):

PR #3730, opened 2023-11-02
Prevent loop of switching between scrollable and paginated horizontal gallery

mgamis-msft:
# What
Fix loop of switching between scrollable and paginated horizontal gallery

# Why
Could potentially fix Teladoc issue detailed in this bug https://skype.visualstudio.com/SPOOL/_workitems/edit/3461020

# How Tested
Elicited looping between scrollable and paginated horizontal gallery by omitting a style in Stack component outside of the VideoGallery like shown in this video:
https://github.com/Azure/communication-ui-library/assets/79475487/49e0a512-e5f7-4fb2-91f0-029836cc9dd8

Tested this branch as shown in following video to confirm that the loop is prevented:
https://github.com/Azure/communication-ui-library/assets/79475487/d579f44e-ff0d-4e4a-bcb3-13eab5011c70

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/VideoGallery/FloatingLocalVideoLayout.tsx
+++ b/packages/react-components/src/components/VideoGallery/FloatingLocalVideoLayout.tsx
@@ -218,6 +218,7 @@ export const FloatingLocalVideoLayout = (props: FloatingLocalVideoLayoutProps): 
         onChildrenPerPageChange={(n: number) => {
           childrenPerPage.current = n;
         }}
+        parentWidth={parentWidth}
       />
     );
   }, [
@@ -228,7 +229,8 @@ export const FloatingLocalVideoLayout = (props: FloatingLocalVideoLayoutProps): 
     styles?.horizontalGallery,
     /* @conditional-compile-remove(vertical-gallery) */ overflowGalleryPosition,
     setIndexesToRender,
-    /* @conditional-compile-remove(vertical-gallery) */ styles?.verticalGallery
+    /* @conditional-compile-remove(vertical-gallery) */ styles?.verticalGallery,
+    parentWidth
   ]);
 
   return (
--- a/packages/react-components/src/components/VideoGallery/OverflowGallery.tsx
+++ b/packages/react-components/src/components/VideoGallery/OverflowGallery.tsx
@@ -24,6 +24,8 @@ import {
   verticalGalleryContainerStyle,
   verticalGalleryStyle
 } from './styles/VideoGalleryResponsiveVerticalGallery.styles';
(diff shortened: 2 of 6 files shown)


packages/react-components/src/components/VideoGallery/OverflowGallery.tsx
```diff
@@ -114,8 +118,13 @@ export const OverflowGallery = (props: {
         horizontalGalleryElements={overflowGalleryElements ? overflowGalleryElements : [<></>]}
         onFetchTilesToRender={onFetchTilesToRender}
         key="scrollable-horizontal-gallery"
-        /* @conditional-compile-remove(gallery-layouts) */
-        layout={props.layout}
+        containerStyles={{
```
JamesBurnside, line 121: memoize complex styles for expensive component

packages/react-components/src/components/VideoGallery/OverflowGallery.tsx
```diff
@@ -114,8 +118,13 @@ export const OverflowGallery = (props: {
         horizontalGalleryElements={overflowGalleryElements ? overflowGalleryElements : [<></>]}
```
JamesBurnside, line 118: Consider instead of showing a ScrollableHorizontalGallery when `isNarrow` instead expose a property `overflowGalleryPagination: 'infiniteScroll' | 'paginated'`

JamesBurnside (approved):
Changes look good, but consider the comment on changing how we switch from a paginated overflow gallery to a scrollable one.

JamesBurnside:
I see snapshots have failed, running update snapshots and will re-review once those upload

packages/react-components/src/components/VideoGallery/OverflowGallery.tsx
mgamis-msft: That is something we should do but I will leave it for another PR because that would cause an API change.

packages/react-components/src/components/VideoGallery/FloatingLocalVideoLayout.tsx
```diff
@@ -218,6 +218,7 @@ export const FloatingLocalVideoLayout = (props: FloatingLocalVideoLayoutProps):
         onChildrenPerPageChange={(n: number) => {
           childrenPerPage.current = n;
         }}
+        parentWidth={parentWidth}
```
dmceachernmsft, line 221: Does this change also need to be made in the speaker layout?

dmceachernmsft (approved):

packages/react-components/src/components/VideoGallery/FloatingLocalVideoLayout.tsx
mgamis-msft: Good catch! Updated.

PR #3758, opened 2023-11-06
Rooms end pages update

mgamis-msft:
# What
Update rooms end pages including icons and strings

# Why

# How Tested

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/utils/Utils.ts
```diff
@@ -74,10 +78,13 @@ enum CallEndReasons {
...
   REMOVED_FROM_CALL,
   ROOM_NOT_FOUND,
-  DENIED_PERMISSION_TO_ROOM
+  ROOM_NOT_VALID,
+  DENIED_PERMISSION_TO_ROOM,
+  REMOVED_PERMISSION_TO_JOIN_ROOM
 }
 
 const getCallEndReason = (call: CallState): CallEndReasons => {
+  console.log('MIGUEL call.callEndReason?.subCode: ', call.callEndReason?.subCode);
```
dmceachernmsft, line 87: Console log here

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/PageState.test.ts-snapshots/room-not-found-page-Mobile-Android-Portrait-linux.png
dmceachernmsft, line 1: I am guessing that in the spirit of what Rooms is we are removing these rejoin buttons so that the user cannot come back to the call? is there any concerns about when they are removed by accident?

dmceachernmsft (approved):

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/PageState.test.ts-snapshots/room-not-found-page-Mobile-Android-Portrait-linux.png
mgamis-msft: Just to clarify, this removal page will be shown only when the user is removed using the Rooms api. 
image
This would is not be from a participant in the rooms call removing you. If that were to happen you would have the removed from call page with the 'Rejoin' button.
Now that I think about it I should probably change the 'removed from room' strings to be more clear. Thanks for making me think deeper about this :)

PR #3773, opened 2023-11-08
Reduce console output noise in sample apps

JamesBurnside:
# What

- Only output Azure errors to console
- For mobile logs collect all log levels for remotely sending off logs for debugging

# Why

Console gets completely spammed with azure logs and makes it impossible to see useful things.

# How Tested

Locally, console no longer spammed, but still see azure log errors as expected


samples/Chat/src/app/utils/ShakeToSendLogs.tsx
```diff
@@ -27,21 +27,24 @@ const storeLog = (logType: string, log: string | undefined): void => {
  * This is particularly useful on mobile devices where the console is not easily accessible.
  */
 const startRecordingLogs = (): void => {
-  function hookLogType(logType: string): (...args: unknown[]) => void {
+  function hookLogType(logType: string, outputToConsole: boolean): (...args: unknown[]) => void {
```
emlynmac, line 30: Is there a section in the docs about the log mechanisms?

emlynmac (approved):

samples/Chat/src/app/utils/ShakeToSendLogs.tsx
JamesBurnside: There isn't, is there somewhere particular we should enter information about this?
(added a comment to the top of this file also to detail what the purpose of this component is)

dmceachernmsft (approved):

PR #3775, opened 2023-11-08
[Calling sounds] Introduce composite sound api for playing sounds

dmceachernmsft:
# What

Introduces sound API for opting out of sounds in the composite
# Why

Allows Contoso to opt out of sounds to make their app load faster

https://skype.visualstudio.com/SPOOL/_workitems/edit/3482023
# How Tested

Tested Locally that API turns off sounds


JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/adapter/CallingSoundSubscriber.ts
```diff
@@ -15,40 +15,46 @@ type CallingSounds = {
 export class CallingSoundSubscriber {
   private emitter: EventEmitter;
   private call: CallCommon;
-  private sounds: CallingSounds;
+  sounds?: CallingSounds;
```
mgamis-msft, line 18: Why is this now not private

mgamis-msft (approved):

packages/react-composites/src/composites/CallComposite/adapter/CallingSoundSubscriber.ts
dmceachernmsft: oops should be will fix!

PR #3784, opened 2023-11-10
[Calling sounds] Introduce the ringing sound effect

dmceachernmsft:
# What

Allows the call composite to play a custom sound for when the call is ringing a user, only for the following:
- ACS users
- Teams users
- Teams voice apps
We are not playing this sound when making PSTN calls since there is a dial tone provided from the PSTN line.
# Why

Allows for Audio feedback when a user makes an outbound call

https://skype.visualstudio.com/SPOOL/_workitems/edit/3480550
# How Tested

Validated locally, plumbing tests to come in e2e (no snapshots)


packages/react-composites/src/composites/CallComposite/adapter/CallingSoundSubscriber.ts
```diff
@@ -31,6 +41,19 @@ export class CallingSoundSubscriber {
...
+        this.soundsLoaded.callRingingSound.play().catch((e) => {
+          console.error(e, 'Failed to play call ringing sound, check loader config to make sure it is correct');
+        });
+      }
+      if (
+        (this.call.state === 'Connected' || this.call.state === 'Disconnected') &&
+        this.soundsLoaded?.callRingingSound
+      ) {
+        this.soundsLoaded.callRingingSound.loop = false;
+        this.soundsLoaded.callRingingSound.pause();
```
JamesBurnside, line 55: Should `unsubscribeAll` also pause/stop all active music

packages/react-composites/src/composites/CallComposite/adapter/CallingSoundSubscriber.ts
```diff
@@ -53,8 +76,34 @@ export class CallingSoundSubscriber {
...
+      callRingingSound
     };
   }
 }
+
+/**
+ * Helper function to allow the calling sound subscriber to determine when to play the ringing
+ * sound when making an outbound call.
+ */
+const playRingingSound = (call: CallCommon, locator: CallAdapterLocator): boolean => {
```
JamesBurnside, line 95: nit: rn `playRingingSound` to something like `shouldPlayRingingSound` as this doesn't do the actual Play

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/adapter/CallingSoundSubscriber.ts
```diff
@@ -2,24 +2,31 @@
...
+/* @conditional-compile-remove(calling-sounds) */
+import { CallParticipantsLocator } from './AzureCommunicationCallAdapter';
+/* @conditional-compile-remove(calling-sounds) */
+import { fromFlatCommunicationIdentifier } from '@internal/acs-ui-common';
+/* @conditional-compile-remove(calling-sounds) */
+import { isPhoneNumberIdentifier } from '@azure/communication-common';
 
 type CallingSoundsLoaded = {
   callEndedSound: HTMLAudioElement | undefined;
+  callRingingSound?: HTMLAudioElement | undefined;
```
JamesBurnside, line 16: Should callEndedSound also be optional?

prabhjot-msft (approved):

PR #3785, opened 2023-11-10
Support showing local screenshare stream

JamesBurnside:
# What

- Stateful client support both `Video` localvideostream and `Screenshare` localvideostream
- Update selectors to pass screenshare info to video gallery
- Update onCreateStreamView to handle both localvideo and screenshare of a local participant
- Update VideoGallery to handle localscreenshare state

# Why

Experience improvement

# How Tested

Locally, recording

https://github.com/Azure/communication-ui-library/assets/2684369/80972969-c959-4df7-9fc6-4d2b558971ca

## Other

Would like design sign off if we should do something like this where we clearly identify the screenshare as the local users stream:

image


packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
```diff
@@ -324,13 +324,25 @@ export const createDefaultCommonCallingHandlers = memoizeOne(
         return;
       }
 
-      const localStream = callState.localVideoStreams.find((item) => item.mediaStreamType === 'Video');
-      if (!localStream) {
-        return;
+      // Find the first available stream, if there is none, then get the first stream
+      const localVideoStream = callState.localVideoStreams.find((item) => item.mediaStreamType === 'Video');
+      const screenShareStream = callState.localVideoStreams.find((item) => item.mediaStreamType === 'ScreenSharing');
```
JamesBurnside, line 329: TODO: update this for onDisposeLocalStream also. It will need a new arg to distinguish which stream should be disposed off (we do this in a very hacky way for remote streams so we should update that also in the process)

packages/calling-stateful-client/src/InternalCallContext.ts
```diff
@@ -51,8 +48,8 @@ export class InternalCallContext {
   // <CallId, <ParticipantKey, <StreamId, RemoteRenderInfo>>
   private _remoteRenderInfos = new Map<string, Map<string, Map<number, RemoteRenderInfo>>>();
 
-  // <CallId, LocalRenderInfo>.
-  private _localRenderInfos = new Map<string, LocalRenderInfo>();
+  // <CallId, <MediaStreamType, LocalRenderInfo>>.
```
JamesBurnside, line 51: This is the key change in this PR. Instead of having a single LocalRenderInfo per Call, we now have a map of streamType to render info per call.
This supports having one video stream, one screenshare stream, one raw video access stream etc.

There isn't a good streamId to use instead of MediaStreamType to allow multiple local camera streams/multiple local screensharestreams etc (instead of just one of each). We could use stream.source.id but that key gets messy to track when switchSource happens

packages/calling-stateful-client/src/InternalCallContext.ts
```diff
@@ -25,24 +25,21 @@ export type RenderStatus = 'NotRendered' | 'Rendering' | 'Rendered' | 'Stopping'
 /**
  * Internal container to hold common state needed to keep track of renders.
  */
-export interface RenderInfo {
+export interface RenderInfo<T> {
```
dmceachernmsft, line 28: Why do we need to add this generic here? is this because they dont have typing for the two different kinds of streams and we know they can be different?

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/Screenshare.test.ts-snapshots/local-screenshare-Desktop-Chrome-linux.png
dmceachernmsft, line 1: Was the original version of this test wrong? the image name implies its the local stream but the displayname in the tile is someone else (or did we also update the name to be the local user placeholder instead of the display name?)

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
```diff
@@ -324,13 +324,25 @@ export const createDefaultCommonCallingHandlers = memoizeOne(
...
+      if (localVideoStream && !localVideoStream.view) {
+        createViewResult = await callClient.createView(call.id, undefined, localVideoStream, options);
+      }
+
+      if (screenShareStream && !screenShareStream.view) {
+        // Hardcoded `scalingMode` since it is highly unlikely that CONTOSO would ever want to use a different scaling mode for screenshare.
+        // Using `Crop` would crop the contents of screenshare and `Stretch` would warp it.
+        // `Fit` is the only mode that maintains the integrity of the screen being shared.
+        createViewResult = await callClient.createView(call.id, undefined, screenShareStream, {
+          scalingMode: 'Fit'
```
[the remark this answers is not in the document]
JamesBurnside: Like this 😉 https://github.com/Azure/communication-ui-library/pull/3745

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -760,7 +785,8 @@ export class CallContext {
     this.modifyState((draft: CallClientState) => {
       const foundIndex = draft.deviceManager.unparentedViews.findIndex(
         (stream) =>
-          stream.source.id === localVideoStream.source.id && stream.mediaStreamType === localVideoStream.mediaStreamType
+          stream.mediaStreamType === localVideoStream.mediaStreamType &&
```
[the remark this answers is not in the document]
JamesBurnside: Yeah we don't support the scenario today in composites and it does feel unlikely, but its possible so we should probably code to be kind to future us's

packages/calling-stateful-client/src/InternalCallContext.ts
JamesBurnside: No functional change here, just moved the stream from being defined inside the LocalRenderInfo/RemoteRenderInfo type to be being defined in the RenderInfo so if the RenderInfo type is used as a generic it has access to stream

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/Screenshare.test.ts-snapshots/local-screenshare-Desktop-Chrome-linux.png
JamesBurnside: The local test user is Agnes Thompson so this was correct. By design we showed the local user here and not the work (You). I changed that as when showing the local preview it makes more sense to show (You) so you know its your stream and not someone else's. But I am waiting for design feedback on this also.

packages/calling-stateful-client/src/CallContext.ts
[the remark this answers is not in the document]
JamesBurnside: Updated to have disposeLocalScreenShareStream

packages/calling-component-bindings/src/handlers/createCommonHandlers.ts
JamesBurnside: Updated to have disposeLocalScreenShareStream

JamesBurnside:
This needs more design input

PR #3793, opened 2023-11-10
Add border for avatar coin for raised hand

rzdor:
# What

image
Add border around avatar coin for raised handed participants

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3424793/

# How Tested

# Process & policy checklist


packages/communication-react/review/stable/communication-react.api.md
```diff
@@ -1775,6 +1775,7 @@ export type CustomAvatarOptions = {
     size?: PersonaSize;
     styles?: IStyleFunctionOrObject<IPersonaStyleProps, IPersonaStyles>;
     text?: string;
+    isActive?: boolean;
```
dmceachernmsft, line 1778: should this be `isRaisedHandActive`?

packages/react-components/src/types/OnRender.ts
```diff
@@ -43,6 +43,9 @@ export type CustomAvatarOptions = {
    * It has '?' in place of initials, with static font and background colors
    */
   showUnknownPersonaCoin?: boolean;
+  /** Active flag to display border around coin */
+  /* @conditional-compile-remove(raise-hand) */
+  isActive?: boolean;
```
dmceachernmsft, line 48: if we are to make this generic for the override should we also be making a way to pick the colour?

JamesBurnside:
Doesn't seem to be working in storybook?
image

Have you been able to test what it looks like when presence is enabled:
image

JamesBurnside:
Just a note, Teams doesn't seem to have this anymore, do we definitely want it:
image

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -86,9 +99,28 @@ export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
...
+  const theme = useTheme();
+  /* @conditional-compile-remove(raise-hand) */
+  let activePersona = '';
+  if (props.allowActiveBorder) {
+    const callingPalette = (theme as unknown as CallingTheme).callingPalette;
+    const raisedHandBorderColor = callingPalette.raiseHandGold;
+    // Display a border for raised handed participants in participant list
+    activePersona = mergeStyles({
+      border: 'solid 2px',
+      borderColor: props.isActive ? raisedHandBorderColor : 'transparent',
```
JamesBurnside, line 112: there needs to be a prop for contoso developers to style this border

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -86,9 +99,28 @@ export const AvatarPersona = (props: AvatarPersonaProps): JSX.Element => {
     })();
   }, [data, dataProvider, userId]);
 
+  /* @conditional-compile-remove(raise-hand) */
+  const theme = useTheme();
+  /* @conditional-compile-remove(raise-hand) */
+  let activePersona = '';
+  if (props.allowActiveBorder) {
+    const callingPalette = (theme as unknown as CallingTheme).callingPalette;
```
JamesBurnside, line 107: see comment below about passing in the raisedHandBorderColor, we can't assume here the callingPalette is accessible

packages/react-components/src/components/ParticipantItem.tsx
```diff
@@ -131,6 +131,8 @@ export interface ParticipantItemProps {
    * Takes in a unique id value of the element you would like to be read before the ParticipantItem.
    */
   ariaLabelledBy?: string;
+  /** Display border around persona coin */
+  isRaisedHand?: boolean;
```
JamesBurnside, line 135: this should be generic, something like you have in the persona coin `isActive`

packages/communication-react/review/stable/communication-react.api.md
JamesBurnside: I would say change this to isActive. In  this context of this component it simply adds a border, that could be used for anything

rzdor: This is part of CustomAvatar and it is a components

JamesBurnside: We don't apply it internally to these react-components correct? If we are passing in a property into the component, only to pass it on to a callback (the callback which is also passed into the component) I don't see the value in adding this to the component interface

rzdor:
> Just a note, Teams doesn't seem to have this anymore, do we definitely want it?: image

It still exist in Web teams

packages/react-components/src/components/ParticipantItem.tsx
dmceachernmsft: I agree this should still be done, we shouldn't couple a concept to the component that is specific to ACS and Teams interop. Our components are supposed to be separated from anything ACS and Teams related so that Contoso can use them in more specific ways to their needs.

packages/react-composites/tests/browser/snapshots/beta/tests/browser/call/hermetic/ParticipantPane.test.ts-snapshots/ACS-group-call-screen-with-no-add-people-button-Mobile-Android-Portrait-linux.png
dmceachernmsft, line 1: we see a padding shift in the persona items here that is messing with the alignment, we should track this down and make sure we aren't adding anything that would move the participant items.
image

dmceachernmsft (approved):

packages/communication-react/review/stable/communication-react.api.md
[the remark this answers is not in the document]
mgamis-msft: This should ok for stable since raise hand is already GA'd

packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -208,6 +211,10 @@ const onRenderParticipantDefault = (
       /* @conditional-compile-remove(PSTN-calls) */
       participantState={callingParticipant.state}
       ariaLabelledBy={participantAriaLabelledBy}
+      /* @conditional-compile-remove(raise-hand) */
+      showCoinBorder={callingParticipant?.raisedHand ? true : false}
+      /* @conditional-compile-remove(raise-hand) */
+      coinBorderColor={callingPalette.raiseHandGold}
```
mgamis-msft, line 217: I don't think we need to add this coinBorderColor prop for ParticipantItem. We can just change the styles prop (on line 199) to have a border with this color when `callingParticipant?.raisedHand` is true

packages/react-components/src/components/ParticipantList.tsx
```diff
@@ -208,6 +211,10 @@ const onRenderParticipantDefault = (
       /* @conditional-compile-remove(PSTN-calls) */
       participantState={callingParticipant.state}
       ariaLabelledBy={participantAriaLabelledBy}
+      /* @conditional-compile-remove(raise-hand) */
+      showCoinBorder={callingParticipant?.raisedHand ? true : false}
```
mgamis-msft, line 215: I don't think adding this showCoinBorder prop for ParticipantItem is necessary. I think you can assign onRenderAvatar callback prop (on line 207) to a variable like this or similar:
```
  const onRenderAvatarWithRaiseHand = callingParticipant?.raisedHand
    ? (userId?: string, options?: CustomAvatarOptions, defaultOnRender?: (props: CustomAvatarOptions) => JSX.Element) =>
        onRenderAvatar?.(userId, { ...options, styles: { border: callingPalette.raiseHandGold } }, defaultOnRender)
    : onRenderAvatar;
```

packages/communication-react/review/stable/communication-react.api.md
mgamis-msft: I agree with James, we probably don't need to add these props. See my comments [here](https://github.com/Azure/communication-ui-library/pull/3793/files#r1401284634)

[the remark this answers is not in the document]
JamesBurnside: Ah you are correct, instead @rzdor remove the new conditional compile lines in this PR:
image

packages/react-components/src/components/ParticipantList.tsx
rzdor: Thanks, yes. This is a good solution

packages/react-components/src/components/ParticipantList.tsx
rzdor: thanks, i've added styles using avatar function parametsers

mgamis-msft (approved):

packages/react-composites/src/composites/common/AvatarPersona.tsx
```diff
@@ -3,6 +3,8 @@
 
 import { IPersonaProps, Persona, PersonaInitialsColor } from '@fluentui/react';
 import React, { useEffect, useState } from 'react';
+/* @conditional-compile-remove(raise-hand) */
```
mgamis-msft, line 6: You can remove conditional compilation for raise hand feature since we have successfully GA's this feature but otherwise looks good!

packages/react-composites/tests/browser/snapshots/beta/tests/browser/chat/hermetic/ImageGallery.test.ts-snapshots/inline-image-gallery-modal-broken-link-Desktop-Chrome-linux.png
mgamis-msft, line 1: Weird that the message sent time is changing but this is not related to this feature.

PR #3794, opened 2023-11-14
Filter bots from video gallery

### LINKED ISSUE #3695: CallComposite shows non-existent “unnamed participant” when joining Teams meeting
We are using the `CallComposite` component to join Teams meetings. 

We are using `@azure/communication-react: 1.7.0-beta.2`, and we want to upgrade to the latest beta version (`1.9.0-beta.1`),
However, the `CallComposite` shows two non-existent “unnamed participants” when joining a Teams meeting. 

We could also reproduce it in the `CallComposite` component from Storybook. 
https://azure.github.io/communication-ui-library/?path=/story/composites-call-joinexistingcall--join-existing-call

two unnamed participants in call composite from storybook

Steps to reproduce:

- Create a Teams meeting.
- Create an ACS user. 
- Join the meeting as the ACS user using the Teams meeting link.
- Teams UI shows the correct participants in the meeting. 
- Call Composite shows two “Unnamed participants”.  

Is it possible to prevent this behavior, or is it a known issue? 

Environment: 

- "@azure/communication-calling": "1.17.1-beta.5".
- "@azure/communication-calling-effects": "1.0.1".
- "@azure/communication-chat": "1.4.0-beta.1".
- "@azure/communication-react": "1.9.0-beta.1".
- And also in the Storybook component. 
- OS & Device: MacOS.
- Browser: Chrome.

Thanks

JamesBurnside:
# What

Filter out bots on video gallery

# Why

Showing in video gallery
Fixes #3695

# How Tested
tbd


packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -53,6 +53,10 @@ export const _videoGalleryRemoteParticipantsMemo: (
             isPhoneNumberIdentifier(participant.identifier)
           );
         })
+        // Filter out MicrosoftBot participants
+        .filter((participant: RemoteParticipantState) => {
```
dmceachernmsft, line 57: will this also filter teams voice apps? I am worried that we are just filtering out all bots with this when we still need to show these other apps since they are the entity that the user will be calling in flows like c2c

packages/react-composites/src/composites/CallComposite/selectors/mediaGallerySelector.ts
```diff
@@ -21,6 +22,6 @@ export const getRemoteParticipantsConnectedSelector = reselect.createSelector(
   [getRemoteParticipants],
   (remoteParticipants) => {
     const participants = Object.values(remoteParticipants ?? {});
-    return participants.filter((p) => p.state === 'Connected');
+    return participants.filter((p) => p.state === 'Connected' && !isMicrosoftTeamsAppIdentifier(p.identifier));
```
dmceachernmsft, line 25: Same concern here, the user should be able to see the teams voice app they are calling.

JamesBurnside:
Closing this as the `@azure/communication-calling` package will be exposing a better way to filter bots that should and shouldn't be shown from Teams

PR #3798, opened 2023-11-15
[Chat] Update dev dependency of chat sdk

JoshuaLai:
# What
Updated dev dependency 

# Why
Missed one dev dependency

### DIFF
--- a/packages/chat-component-bindings/package.json
+++ b/packages/chat-component-bindings/package.json
@@ -40,7 +40,7 @@
     "memoize-one": "^5.2.1"
   },
   "peerDependencies": {
-    "@azure/communication-chat": "1.4.0-beta.1 || >=1.3.2",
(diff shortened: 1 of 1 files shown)


packages/chat-component-bindings/package.json
```diff
@@ -40,7 +40,7 @@
     "memoize-one": "^5.2.1"
   },
   "peerDependencies": {
-    "@azure/communication-chat": "1.4.0-beta.1 || >=1.3.2",
```
vhuseinova-msft, line 43: It looks like communication-react package.json has incorrect version too

palatter (approved):

packages/chat-component-bindings/package.json
JoshuaLai: I just took a look on main and it looks to be pointing to `1.4.0-beta.2` Can you verify?

vhuseinova-msft (approved):

PR #3799, opened 2023-11-15
Autogenerate api diffs for apiview for features

JamesBurnside:
# What

We now have a preprocessor that can strip out features. Lets use this to avoid having to manually generate api view diffs for review

## How

1. Detects if the feature is an alpha or beta feature
1. Switches to the appropriate flavor (stable flavor for a beta feature, beta-release for an alpha flavor)
   1. i.e. switches to the flavor where the feature is wanting to be
1. Builds a baseline api json (i.e. a build without the feature there)
1. Enables the feature for the flavor and rebuilds (i.e. a build with the feature there)
1. Spits out two api files: `baseline.api.json` and `feature.api.json` for uploading to APIView to create a diff for review.

## Where this could be improved

1. Currently this uses the `rush switch-flavor` and `rush build` commands which alters the users dev environment. It does a clean up when completed but this is a fragile technique and really should be calling the appropriate commands internally without impacting any dev env. (Alternatively we recommend using on a build agent through github actions and then there's no issue here)
2. Take this a step further and auto upload to apiview and just pass a link
3. Allow passing in multiple features
4. (Edge case) This will break if a feature is being stabilized and uses a property from a dependency that is in beta

# How Tested

Locally, tested with large-gallery and custom-branding features (and uploaded the latter to apiview.dev) and it successfully generated a baseline and feature api json files for uploading to apiview.


JamesBurnside:
I have found this personally useful so putting this up as ready for review - but very open to refactors/alternative methods for achieving this.

packages/react-components/src/localization/locales/ar-SA/strings.json
```diff
@@ -343,4 +343,4 @@
     "rightNavButtonAriaLabel": "الصفحة التالية"
   },
   "AttendeeRole": "الحضور"
-}
\ No newline at end of file
+}
```
emlynmac, line 346: Can we get the locale strings out of this PR?

common/scripts/generate-api-diff.mjs
```diff
@@ -0,0 +1,171 @@
...
+  modifyFeatureDefinitionsFile(feature, isAlphaFeature ? 'betaToAlpha' : 'stableToBeta');
+  console.log(`Generating feature.api.json file`);
+  const featureFilePath = await generateApiFile(true);
+
+  // restore the feature definitions file
+  fs.copyFileSync(`${featureDefinitionsFile}.bak`, featureDefinitionsFile);
+  fs.rmSync(`${featureDefinitionsFile}.bak`);
+
+  // restore the flavor
+  await exec('rush switch-flavor:beta');
```
vhuseinova-msft, line 74: Maybe not for this PR, it would be nice to check which flavor was selected and use it here instead

common/config/rush/command-line.json
```diff
@@ -83,6 +83,14 @@
       "shellCommand": "(rush switch-flavor:stable && rush rebuild -t @azure/communication-react) || rush switch-flavor:beta && rush rebuild -t @azure/communication-react",
       "safeForSimultaneousRushProcesses": true
     },
+    {
+      "commandKind": "global",
+      "name": "generate-api-diff",
+      "summary": "Generate two api files for a supplied feature name",
+      "description": "Generate two api files that differ only by a supplied feature name. Note: This result of this commands will switch current flavor to beta.",
```
vhuseinova-msft, line 90: It would be great to add a description of the generated files (as in `common/scripts/generate-api-diff.mjs`) so if someone doesn't use the command often, they still will understand the purpose for both files without searching in code.

vhuseinova-msft, line 90: May be a typo: "This result"

vhuseinova-msft (approved):
A couple of minor suggestions but otherwise looks good!

common/scripts/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "build:stable": "",
+    "clean": "",
+    "test": "",
+    "test:ci-coverage": "",
+    "test:stable": "",
+    "prettier": "",
+    "prettier:check": ""
+  },
+  "devDependencies": {
+    "@types/yargs": "17.0.29",
```
emlynmac, line 19: Nit: should these be hard pinned or use caret?

emlynmac (approved):

common/scripts/generate-api-diff.mjs
JamesBurnside: I like that! Sounds doable, I'll try add that

common/scripts/package.json
JamesBurnside: Updated to `^`

common/config/rush/command-line.json
JamesBurnside: removed in description updated

JamesBurnside: Great point! Updated the description with more details

common/scripts/generate-api-diff.mjs
JamesBurnside: Updated! Turns out we have a `getBuildFlavor()` fn already for using in our scripts! Great suggestion

packages/react-components/src/localization/locales/ar-SA/strings.json
JamesBurnside: updated

dmceachernmsft (approved):

PR #3805, opened 2023-11-16
[Calling sounds] Busy tone when call is rejected

dmceachernmsft:
# What

Introduces new sound when busy on busy for a teams user rejects or when a user rejects the out going call
# Why

Allows for Audio queue to communicate that the call was rejected

https://skype.visualstudio.com/SPOOL/_workitems/edit/3499854
https://skype.visualstudio.com/SPOOL/_workitems/edit/3499853
# How Tested

Locally and updated unit tests


packages/react-composites/src/composites/CallComposite/adapter/CallingSoundSubscriber.ts
```diff
@@ -14,8 +14,11 @@ import { isPhoneNumberIdentifier } from '@azure/communication-common';
 type CallingSoundsLoaded = {
   callEndedSound?: HTMLAudioElement;
   callRingingSound?: HTMLAudioElement;
+  callBusySound?: HTMLAudioElement;
 };
 
+const CALL_REJECTED_CODE = 603;
```
JamesBurnside, line 20: can we have a central file that has all of the codes in it? `CallSDKCodeConstants.ts` or something

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/adapter/CallingSoundSubscriber.ts
dmceachernmsft: This is a good idea, I will work with Miguel to find a place and put any Rooms codes in there too

mgamis-msft: Yeah that's a good idea!

mgamis-msft (approved):

PR #3807, opened 2023-11-17
[Calling sounds] Busy screen

dmceachernmsft:
# What

Introduces a busy screen for when the call is rejected by the other participant
# Why

adds visual queue to the audio queue that the call was ended for a different reason

# How Tested

image
Locally


packages/react-composites/src/composites/CallComposite/adapter/CallAdapter.ts
```diff
@@ -58,7 +58,8 @@ export type CallCompositePage =
   | 'lobby'
   | 'removedFromCall'
   | /* @conditional-compile-remove(unsupported-browser) */ 'unsupportedEnvironment'
-  | /* @conditional-compile-remove(call-transfer) */ 'transferring';
+  | /* @conditional-compile-remove(call-transfer) */ 'transferring'
+  | /* @conditional-compile-remove(calling-sounds) */ 'callRejected';
```
JamesBurnside, line 62: Same as rooms, to avoid us piling up a ton of end call pages, can we bracket this under an existing end call page and use an end call reason for the composite to display different text + icon?
Not sure what rooms ended up using...

JamesBurnside (approved):

mgamis-msft (approved):

PR #3809, opened 2023-11-18
[Chat] Message thread rendering improvements

vhuseinova-msft:
# What

- improved MessageThread performance
- fixed an issue when all messages were reloaded after a new message is sent or a message is updated
- fixed an issue when message thread blinks for self-host
- moved some functionality from MessageThread to separate files for easier support
- -300+ lines from MessageThread.tsx

# Why

MessageThread reloaded all messages when a new message was sent or a message was updated

# How Tested

sample apps (Chat and CallWithChat) and self-host

# Process & policy checklist


packages/react-components/src/components/ChatMessage/ChatMessageComponentWrapper.tsx
```diff
@@ -0,0 +1,123 @@
...
+  /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+  inlineAttachments: Record<string, Record<string, string>>;
+  /* @conditional-compile-remove(mention) */
+  mentionOptions?: MentionOptions;
+};
+
+/**
+ * @private
+ */
+export const ChatMessageComponentWrapper = (props: ChatMessageComponentWrapperProps): JSX.Element => {
```
emlynmac, line 68: Would be good to have some comments around why this component exists.

packages/react-components/src/components/ChatMessage/FluentChatMessageComponentWrapper.tsx
```diff
@@ -0,0 +1,273 @@
...
+          defaultStatusRenderer(message, participantCount ?? 0, readCount ?? 0, message.status)
+        )
+      ) : (
+        <div className={mergeStyles(noMessageStatusStyle)} />
+      );
+    },
+    [message, statusToRender]
+  );
+
+  // let chatMessageComponent: JSX.Element;
```
emlynmac, line 168: Needs removal?

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -170,7 +170,7 @@ describe('Message blocked should display default blocked text correctly', () =>
 });
 
 /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
-describe('Message should display image and attachment correctly', () => {
+describe.only('Message should display image and attachment correctly', () => {
```
emlynmac, line 173: Need to remove the only here.

packages/react-components/src/components/MessageThread.test.tsx
```diff
@@ -180,12 +180,12 @@ describe('Message should display image and attachment correctly', () => {
     });
   });
 
-  test('Message richtext/html img src should be correct', async () => {
+  test.only('Message richtext/html img src should be correct', async () => {
```
emlynmac, line 183: And everywhere `only` is used

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
```diff
@@ -475,7 +473,7 @@ const CallWithChatScreen = (props: CallWithChatScreenProps): JSX.Element => {
   const onRenderChatContent = useCallback(
     (): JSX.Element => (
       <ChatComposite
-        {...chatProps}
+        adapter={chatAdapter}
```
emlynmac, line 476: Is this really what we want? Are there any other props that we might need from chatProps?

emlynmac (commented):
Some comments but looks good! Nice to see some simplification of the message thread component

packages/react-components/src/components/ChatMessage/ChatMessageComponent.tsx
```diff
@@ -161,7 +161,7 @@ export const ChatMessageComponent = (props: ChatMessageComponentProps): JSX.Elem
         onFetchAttachments={props.onFetchAttachments}
         /* @conditional-compile-remove(image-gallery) */
         onInlineImageClicked={props.onInlineImageClicked}
-        /* @conditional-compile-remove(image-gallery) */
+        /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
         attachmentsMap={props.attachmentsMap}
```
Leah-Xia-Microsoft, line 165: Good catch!

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
```diff
@@ -76,21 +76,27 @@ const MessageContentWithLiveAria = (props: MessageContentWithLiveAriaProps): JSX
...
-
-    if (props.attachmentsMap && attachments) {
-      attachments.forEach((fileMetadata) => {
-        if (props.onFetchAttachment && props.attachmentsMap && props.attachmentsMap[fileMetadata.id] === undefined) {
-          props.onFetchAttachment([fileMetadata], props.message.messageId);
-          return;
-        }
-      });
+    if (attachments && attachments.length > 0) {
+      onFetchAttachment(attachments, message.messageId);
```
Leah-Xia-Microsoft, line 97: I kind of remember @JoshuaLai mentioned that there's a reason to call onFetchAttachment on each attachment?

packages/react-components/src/components/MessageThread.tsx
```diff
@@ -2,12 +2,8 @@
 // Licensed under the MIT License.
 
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { Icon, IStyle, mergeStyles, PrimaryButton, Theme, IPersona, Persona, PersonaSize } from '@fluentui/react';
-import {
-  Chat,
-  ChatMessage as FluentChatMessage,
-  ChatMyMessage as FluentChatMyMessage
-} from '@fluentui-contrib/react-chat';
+import { Icon, IStyle, mergeStyles, PrimaryButton /*, Theme, IPersona, Persona, PersonaSize*/ } from '@fluentui/react';
```
Leah-Xia-Microsoft, line 5: Should we remove the commented out imports?

packages/react-composites/src/composites/CallWithChatComposite/CallWithChatComposite.tsx
vhuseinova-msft: There was only adapter before but it was created as a separate object

packages/react-components/src/components/MessageThread.tsx
vhuseinova-msft: Good catch!

packages/react-components/src/components/MessageThread.test.tsx
vhuseinova-msft: Good catch!

packages/react-components/src/components/ChatMessage/FluentChatMessageComponentWrapper.tsx
vhuseinova-msft: Good catch!

packages/react-components/src/components/ChatMessage/ChatMessageComponentWrapper.tsx
vhuseinova-msft: Added!

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
JoshuaLai: This is fine, we had it fetching multiple before because our adapter couldn't handle using the full attachment list, but after my refactor a little while back we can support this type of functionality now

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
```diff
@@ -129,7 +132,7 @@ export const ChatComposite = (props: ChatCompositeProps): JSX.Element => {
             onRenderTypingIndicator={onRenderTypingIndicator}
             onRenderMessage={onRenderMessage}
             onFetchParticipantMenuItems={onFetchParticipantMenuItems}
-            {...fileSharingOptions()}
+            {...fileSharingOptions}
```
JamesBurnside, line 135: nit: Instead of this being memoized and spread across multiple lines, can this just be:
```suggestion
            /* @conditional-compile-remove(file-sharing) */
            fileSharing={options?.fileSharing}
```

JamesBurnside: (`options` is already passed into ChatScreen, this prop may not be necessary at all)

packages/react-components/src/components/ChatMessage/ChatMessageComponentWrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+          {systemMessageComponent}
+        </div>
+      );
+    }
+
+    default: {
+      // We do not handle custom type message by default, users can handle custom type by using onRenderMessage function.
+      const customMessageComponent = onRenderMessage === undefined ? <></> : onRenderMessage({ ...props });
+      return (
+        <div key={messageKey} style={{ paddingTop: '1rem', paddingBottom: '0.25rem' }}>
```
JamesBurnside, line 121: nit: the `style` object should probably be memoized here (though is a relatively low level component so not entirely necessary)

packages/react-components/src/components/ChatMessage/ChatMessageComponentWrapper.tsx
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) Microsoft Corporation.
```
JamesBurnside, line 1: Love splitting out this code!!

packages/react-components/src/components/ChatMessage/ChatMessageContent.tsx
JamesBurnside, line 97: (not necessarily in this PR) Can we update onFetchAttachment to onFetchAttachments as its handling multiple attachments

packages/react-composites/src/composites/ChatComposite/ChatComposite.tsx
vhuseinova-msft: Good catch, will update with the proposed code! I would prefer not to add it to `options` as options is a public ChatCompositeOption type

packages/react-components/src/components/ChatMessage/FluentChatMessageComponentWrapper.tsx
```diff
@@ -0,0 +1,281 @@
...
+  // Fluent UI message components are used here as for default message renderer,
+  // timestamp and author name should be shown but they aren't shown for custom renderer.
+  // More investigations are needed to check if this can be simplified with states.
+  // Status and avatar should be shown for both custom and default renderers.
+  if (message.mine === true) {
+    return (
+      <div>
+        <FluentChatMyMessage
+          attached={attached}
+          root={{
```
JamesBurnside, line 195: Just a note that these objects are recreated each render pass and aren't memozied. Consider memoizing

packages/react-components/src/components/ChatMessage/FluentChatMessageComponentWrapper.tsx
```diff
@@ -0,0 +1,281 @@
...
+  } = props;
+  const chatMessageRenderStyles = useChatMessageRenderStyles();
+
+  const onRenderFileDownloadsMemo = useMemo(() => {
+    /* @conditional-compile-remove(file-sharing) */
+    return onRenderFileDownloads;
+    return undefined;
+  }, [/* @conditional-compile-remove(file-sharing) */ onRenderFileDownloads]);
+
+  // To rerender the defaultChatMessageRenderer if app running across days(every new day chat time stamp need to be regenerated)
```
JamesBurnside, line 79: This comment doesn't seem to make much sense anymore, is there meant to be a prop that is time related in the dependency array?

packages/react-components/src/components/ChatMessage/FluentChatMessageComponentWrapper.tsx
JamesBurnside: Unless v9 handles this much better than v8!

packages/react-components/src/components/ChatMessage/FluentChatMessageComponentWrapper.tsx
vhuseinova-msft: This code was just moved from MessageThread without much change. You are probably right that it can be removed

JamesBurnside (approved):

packages/react-components/src/components/ChatMessage/FluentChatMessageComponentWrapper.tsx
JamesBurnside: Is there anything existing that updates the timestamp message from like "yesterday" to an actual date when a day elapses?

packages/react-components/src/components/ChatMessage/FluentChatMessageComponentWrapper.tsx
vhuseinova-msft: We don't update it too often because of how it's used in MessageThread (React.memo) but I'll add useMemo here too

packages/react-components/src/components/ChatMessage/FluentChatMessageComponentWrapper.tsx
vhuseinova-msft: Thanks for the clarifications! I'll add the dependency to return this change as it looks like it was missed for some time

emlynmac (approved):

Leah-Xia-Microsoft (approved):

PR #3814, opened 2023-11-21
[Calling] Receive Reactions from Remote Participant during a Meeting

Mohtasim:
# What

* Calling-Stateful-Client contains the State needed to maintain the UI Lib for reaction rendering at the UI Layer
* Calling-Component-Binding contains the implementation for the Selectors. VideoGallery and ParticipantList selectors help relay the reaction state information needed to render reaction at the UI 
* react-composite helps expose objects for different packages 
* react-component implementation has the changes to render reaction animation
* storybook implementation has the code to test VideoTile component  
* ccc implied on all component related to Reaction

# Why

# How Tested

* Please put connection string and related credential to start a call. 
* Please put reaction emoji resources at Calling Sample directory. The location should be Calling/public/assets/<EmojiFile>
* The name of the emoji files mapping should be 
[Like, likeEmoji.png],
[Heart, heartEmoji.png],
[Applause, clapEmoji.png],
[Laugh, laughEmoji.png],
[Surprise, surprisedEmoji.png]

# Process & policy checklist


packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -111,7 +118,9 @@ export const convertRemoteParticipantToVideoGalleryRemoteParticipant = (
   state: RemoteParticipantConnectionState,
   displayName?: string,
   /* @conditional-compile-remove(raise-hand) */
-  raisedHand?: unknown // temp unknown type to build stable
+  raisedHand?: unknown, // temp unknown type to build stable
+  /* @conditional-compile-remove(reaction) */
+  reaction?: unknown // temp unknown type to build stable
```
JamesBurnside, line 123: note: using unknown should not be necessary here, the conditional compile comment will strip this out of stable builds

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -187,6 +189,31 @@ export type RaisedHandState = {
...
+/**
+ * State only version of {@link @azure/communication-calling#Call.Reaction.ReactionEventPayload} with UI helper props.
+ * Reation state with a comparative timestamp which helps UI to decide to render the reaction accordingly.
+ * 
+ * @public
+ */
+export type ReactionState = {
+  shouldRender: boolean;
+  reactionType: string;
+  receivedTimeStamp: number;
```
JamesBurnside, line 214: nit: timestamps are typically a `Date` type. suggestion: `receivedTimeStamp` -> `receivedAt`

```suggestion
  receivedAt: Date;
```

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -187,6 +189,31 @@ export type RaisedHandState = {
...
+/* @conditional-compile-remove(reaction) */
+/**
+ * State only version of {@link @azure/communication-calling#Call.Reaction.ReactionEventPayload} with UI helper props.
+ * Reation state with a comparative timestamp which helps UI to decide to render the reaction accordingly.
+ * 
+ * @public
+ */
+export type ReactionState = {
+  shouldRender: boolean;
+  reactionType: string;
```
JamesBurnside, line 213: use `Reaction` from calling SDK here instead of string. suggestion: can we avoid using `type` in the variable name and just have `reaction` (if necessary use `kind` or `variant` as `type` is a keyword in javascript so its generally avoided as a suffix)

```suggestion
  reaction: Reaction;
```

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -187,6 +189,31 @@ export type RaisedHandState = {
...
+
+/* @conditional-compile-remove(reaction) */
+/**
+ * State only version of {@link @azure/communication-calling#Call.Reaction.ReactionEventPayload} with UI helper props.
+ * Reation state with a comparative timestamp which helps UI to decide to render the reaction accordingly.
+ * 
+ * @public
+ */
+export type ReactionState = {
+  shouldRender: boolean;
```
JamesBurnside, line 212: Avoid this property at the stateful layer. The stateful layer isn't tied to rendering logic and is just a reflection of the active call.
The composite or component will need to handle the logic if if the reaction should be played or not

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -187,6 +189,31 @@ export type RaisedHandState = {
...
 
+/* @conditional-compile-remove(reaction) */
+/**
+ * State only version of {@link @azure/communication-calling#ReactionCallFeature}. {@link StatefulCallClient} will
+ * automatically listen for reaction on the call and update the state exposed by {@link StatefulCallClient} accordingly.
+ * 
+ * @public
+ */
+export interface ReactionCallFeatureState {
+  isEnable: boolean;
```
JamesBurnside, line 200: Is this just to enable it at the composite/component level? If so leave this property out of the stateful client. (Per other comment; The stateful layer isn't tied to rendering logic and is just a reflection of the active call.)

JamesBurnside: If this can be removed, we should be able to remove the ReactionCallFeatureState interface and just have reaction: ReactionState directly on the call object

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -382,6 +386,56 @@ export class CallContext {
...
+    baseTimeStamp.setDate(1);
+    baseTimeStamp.setHours(0, 0, 0, 0);
+  
+    let baseUnixTimestamp = Math.floor(baseTimeStamp.getTime() / 1000);
+    let currentUnixTimestamp = Math.floor(new Date().getTime() / 1000) - baseUnixTimestamp;
+    this.modifyState((draft: CallClientState) => {
+      const call = draft.calls[this._callIdHistory.latestCallId(callId)];
+      if(call) {
+        const participant = call.remoteParticipants[participantKey];
+        // We might see concurrency issue here.
```
JamesBurnside, line 402: dw, javascript is single threaded

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -382,6 +386,56 @@ export class CallContext {
...
+  
+            participant.reaction = reactionMessage ? convertFromSDKToReactionState(reactionMessage.reactionType) : { shouldRender: false, reactionType: '', receivedTimeStamp: 0};
+            
+            this._timeOutId[participantKey] = setTimeout(() => {
+              clearParticipantReactionState(this, callId, participantKey);
+            }, 2170);
+          }
+        } else if (participantKey === toFlatCommunicationIdentifier(this._state.userId)) {
+
+          if(reactionMessage == null) {
```
JamesBurnside, line 421: this condition can't happen as L405 already performs this check

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -382,6 +386,56 @@ export class CallContext {
...
+            call.reaction.localParticipantReactionPayload = { shouldRender: false, reactionType: '', receivedTimeStamp: 0};
+          } else {
+            if(call.reaction.localParticipantReactionPayload?.shouldRender && (currentUnixTimestamp - call.reaction?.localParticipantReactionPayload.receivedTimeStamp) < 3000) {
+                clearParticipantReactionState(this, callId, participantKey);
+                clearTimeout(this._timeOutId[participantKey]);
+              }
+          call.reaction.localParticipantReactionPayload = reactionMessage ? convertFromSDKToReactionState(reactionMessage.reactionType) : { shouldRender: false, reactionType: '', receivedTimeStamp: 0};
+          this._timeOutId[participantKey] = setTimeout( () => {
```
JamesBurnside, line 429: timeout logic will need moved to the composite/component level

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -999,3 +1053,9 @@ const findOldestCallEnded = (calls: { [key: string]: { endTime?: Date } }): stri
   }
   return oldestCallId;
 };
+
+/* @conditional-compile-remove(reaction) */
+function clearParticipantReactionState(arg0: CallContext, callId: string, participantKey: string) {
```
JamesBurnside, line 1058: `arg0` --> `callContext`

packages/calling-stateful-client/src/CallClientState.ts
Mohtasim: We could replace it with - `isAllowedByMeetingPolicy: boolean;`

packages/calling-stateful-client/src/CallClientState.ts
Mohtasim: We'll keep the `ReactionMessage` object from sdk.

packages/calling-stateful-client/src/CallContext.ts
Mohtasim: We'll keep it here for now.

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -401,7 +438,22 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
             )}
           </Stack>
         )}
-
+        
+        { /* @conditional-compile-remove(reaction) */
+        reaction?.shouldRender && (currentUnitTimestamp - reaction.receivedTimeStamp) < 3000 && (
```
Mohtasim, line 443: Check if sprite source is available to render or not!

packages/react-components/src/components/utils/videoTileStylesUtils.tsx
```diff
@@ -10,3 +10,29 @@ export const getVideoTileOverrideColor = (isVideoRendered: boolean, theme: Theme
...
+
+  let baseUnixTimestamp = Math.floor(baseTimeStamp.getTime() / 1000);
+  return Math.floor(new Date().getTime() / 1000) - baseUnixTimestamp;
+}
+
+/* @conditional-compile-remove(reaction) */
+/**
+ * @private
+ */
+export const reactionEmoji = new Map<string, string> ([
```
Mohtasim, line 32: get rid of this.

packages/react-components/src/types/ParticipantListParticipant.ts
```diff
@@ -32,6 +35,27 @@ export type RaisedHand = {
...
+   */
+  shouldRender: boolean;
+  /**
+   * Specifies the type of reaction videoTile should render
+   */
+  reactionType: string;
+  /**
+   * Received timestamp relative to 01-01-2020 0:0:0 hours
+   */
+  receivedTimeStamp: number;
```
Mohtasim, line 56: change to `receivedAt`

samples/CallWithChat/package.json
```diff
@@ -33,7 +33,7 @@
     "@azure/communication-identity": "^1.2.0",
     "@azure/communication-calling": "1.19.1-beta.2 || ^1.18.1",
     "@azure/communication-chat": "1.4.0-beta.2 || ^1.3.2",
-    "@azure/communication-react": "1.10.1-beta.0",
+    "@azure/communication-react": "1.10.0-beta.1",
```
Mohtasim, line 36: remove this.

samples/Calling/package.json
```diff
@@ -30,7 +30,7 @@
     "@azure/abort-controller": "^1.1.0",
     "@azure/communication-identity": "^1.2.0",
     "@azure/communication-calling": "1.19.1-beta.2 || ^1.18.1",
-    "@azure/communication-react": "1.10.1-beta.0",
+    "@azure/communication-react": "1.10.0-beta.1",
```
Mohtasim, line 33: remove this.

samples/StaticHtmlComposites/package.json
```diff
@@ -23,10 +23,10 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@azure/communication-react": "1.10.1-beta.0",
-    "@azure/communication-common": "^2.3.0",
     "@azure/communication-calling": "1.19.1-beta.2 || ^1.18.1",
     "@azure/communication-chat": "1.4.0-beta.2 || ^1.3.2",
+    "@azure/communication-react": "1.10.0-beta.1",
```
Mohtasim, line 28: Remove this.

packages/calling-stateful-client/src/Converter.ts
```diff
@@ -18,7 +18,7 @@ import {
   RemoteVideoStreamState as DeclarativeRemoteVideoStream,
   LocalVideoStreamState as DeclarativeLocalVideoStream,
   IncomingCallState as DeclarativeIncomingCall,
-  VideoStreamRendererViewState as DeclarativeVideoStreamRendererView
+  VideoStreamRendererViewState as DeclarativeVideoStreamRendererView,
```
dmceachernmsft, line 21: nit: These will cause linter issues, please remove trailing commas

packages/calling-stateful-client/src/ReactionSubscriber.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  private subscribe = (): void => {
+    this._reaction.on('reaction', this.reactionEvent);
+  };
+
+  public unsubscribe = (): void => {
+    this._reaction.off('reaction', this.reactionEvent);
+  };
+
+  private reactionEvent = (event): void => {
```
dmceachernmsft, line 35: nit: this should reflect that it is a handler maybe call it `onReactionEvent`

packages/react-components/src/components/LocalVideoTile.tsx
```diff
@@ -5,6 +5,8 @@ import { Stack } from '@fluentui/react';
 import { _formatString } from '@internal/acs-ui-common';
 import React, { useMemo } from 'react';
 import { OnRenderAvatarCallback, VideoStreamOptions, CreateVideoStreamViewResult } from '../types';
+/* @conditional-compile-remove(reaction) */
+import { Reaction } from '../types';
```
dmceachernmsft, line 9: We should be getting this type from the stateful client package off the callState. We don't want to have multiple definitions of the same type throughout the packages.

packages/react-components/src/components/VideoTile.tsx
dmceachernmsft, line 443: We should just check to see that the resource has loaded for the sprite map. if this is the only reason this state is on the reaction type we should remove it and make sure that we are getting the sprite map or however the reactions will be loaded.

packages/storybook/stories/VideoTile/VideoTile.stories.tsx
```diff
@@ -71,6 +72,16 @@ const getDocs: () => JSX.Element = () => {
         <VideoTileMenuItemsExample />
       </Canvas>
 
+      <Heading>Reaction Animation for Video Tile</Heading>
```
dmceachernmsft, line 75: This should be a separate PR as part of the DoD work.

packages/storybook/stories/VideoTile/VideoTile.stories.tsx
```diff
@@ -98,7 +109,7 @@ const VideoTileStory = (args): JSX.Element => {
       showMuteIndicator={args.showMuteIndicator}
       showLabel={args.showLabel}
       isMirrored={args.isMirrored}
-      isMuted={args.isMuted}
+      isMuted={true}
```
dmceachernmsft, line 112: nit: should this be updated?

packages/storybook/stories/VideoTile/snippets/ReactionAnimationVideoTile.snippet.tsx
```diff
@@ -0,0 +1,114 @@
...
+                  onClick: () => alert(`Added ${displayName} to Important`)
+                }
+              ]
+            }
+          }
+        ],
+        styles: {}
+      };
+
+      const react: Reaction = {
```
dmceachernmsft, line 82: nit: lets call this `reaction` as well. this code is exposed to contoso so this code snippet that we are showing here should either show how to do this with components, or write a second snippet to do so if you are wanting to use this to demo the reaction in storybook

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
```diff
@@ -265,6 +266,21 @@ export interface RaiseHandCallFeature {
     raisedHands: RaisedHandState[];
 }
 
+// @public
+export interface ReactionCallFeatureState {
+    // (undocumented)
+    isEnable: boolean;
+    // (undocumented)
+    localParticipantReactionPayload?: ReactionState;
```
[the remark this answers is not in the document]
Mohtasim: Renamed it to `LocalParticipantReacitonState`

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
```diff
@@ -265,6 +266,21 @@ export interface RaiseHandCallFeature {
     raisedHands: RaisedHandState[];
 }
 
+// @public
+export interface ReactionCallFeatureState {
+    // (undocumented)
+    isEnable: boolean;
```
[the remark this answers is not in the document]
Mohtasim: This we removed and using the functionality from the Capability API from SDK.

packages/storybook/stories/VideoTile/VideoTile.stories.tsx
Mohtasim: Will follow this feedback for the Send Reaction PR flow.

packages/storybook/stories/VideoTile/VideoTile.stories.tsx
Mohtasim: I'll have to change back the storybook component changes. Thanks.

packages/storybook/stories/VideoTile/snippets/ReactionAnimationVideoTile.snippet.tsx
```diff
@@ -0,0 +1,114 @@
...
+                    styles: { root: { lineHeight: '1rem', textAlign: 'center' } }
+                  },
+                  onClick: () => alert(`Added ${displayName} to Important`)
+                }
+              ]
+            }
+          }
+        ],
+        styles: {}
+      };
```
Mohtasim, line 80: Get rid of the menu proprs.

packages/react-components/src/components/LocalVideoTile.tsx
Mohtasim: Update after our verbal discussion: This type is from the react-component's VideoTile's prop type definition which is different from the stateful client's state object.

packages/calling-stateful-client/package.json
```diff
@@ -27,8 +27,8 @@
     "_api-extractor:by-flavor": "if-env COMMUNICATION_REACT_FLAVOR=stable && api-extractor run -c api-extractor.stable.json --local || (rushx _if-beta && api-extractor run --local)",
     "_build:by-flavor": "rushx clean && rushx build:esm && rushx build:cjs && rushx _api-extractor:by-flavor",
     "_by-flavor": "rushx _current-flavor && env-cmd -f ../../common/config/env/.env --use-shell",
-    "_current-flavor": "echo You are running under COMMUNICATION_REACT_FLAVOR: && env-cmd -f ../../common/config/env/.env node -p process.env.COMMUNICATION_REACT_FLAVOR",
-    "_test:by-flavor": "(npm run _if-preprocess && rushx preprocess || if-env COMMUNICATION_REACT_FLAVOR=beta && echo \"skip preprocess\") && jest"
+    "_current-flavor": "echo You are running under $COMMUNICATION_REACT_FLAVOR: && env-cmd -f ../../common/config/env/.env node -p process.env.COMMUNICATION_REACT_FLAVOR",
+    "_test:by-flavor": "(rushx _if-preprocess && rushx preprocess || if-env COMMUNICATION_REACT_FLAVOR=beta && echo \"skip preprocess\") && jest"
```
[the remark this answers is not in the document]
Mohtasim: Well, actually it says npm run in the main. so updated it to a proper command in order to test by beta flavor.

common/config/webpack/sampleapp.webpack.config.js
```diff
@@ -76,7 +76,8 @@ const webpackConfig = (sampleAppDir, env, babelConfig) => {
       new CopyPlugin({
         patterns: [
           { from: path.resolve(sampleAppDir, "public/manifest.json"), to: "manifest.json" },
-          { from: path.resolve(sampleAppDir, "public/assets"), to: "assets",  noErrorOnMissing: true }
```
JamesBurnside, line 79: this looks like an incorrect merge conflict resolve - revert changes to this file, its since been updated to copy all assets in public/assets and not need to specify sub-folders

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
```diff
@@ -115,6 +116,7 @@ export interface CallState {
     isMuted: boolean;
     isScreenSharingOn: boolean;
     kind: CallKind;
+    localParticipantReactionState: LocalParticipantReactionState;
```
JamesBurnside, line 119: Is `LocalParticipantReactionState` needed?
Can this just be 
```suggestion
    localParticipantReactionState?: ReactionState;
```

This currently this reads as: `call.reactionState.reactionState`

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -187,6 +189,38 @@ export type RaisedHandState = {
...
+  /**
+   * Local participant's reaction state.
+   */
+  reactionState?: ReactionState;
+}
+
+/* @conditional-compile-remove(reaction) */
+/**
+ * State only version of {@link @azure/communication-calling#Call.ReactionEventPayload} with UI helper props.
+ * Reation state with a comparative timestamp which helps UI to decide to render the reaction accordingly.
```
JamesBurnside, line 209: Reation -> Reaction

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -344,6 +378,12 @@ export interface RemoteParticipantState {
    * Proxy of {@link @azure/communication-calling#Call.RaisedHand.raisedHands}.
    */
   raisedHand?: RaisedHandState;
+
+  /* @conditional-compile-remove(reaction) */
+  /**
+   * Proxy of {@link @azure/communication-calling#Call.Reaction.ReactionEventPayload}.
```
JamesBurnside, line 384: This isn't just a proxy but also adds the additional receivedAt property

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -400,6 +402,59 @@ export class CallContext {
...
+          }
+        } else if (participantKey === toFlatCommunicationIdentifier(this._state.userId)) {
+          if (reactionMessage == null) {
+            call.localParticipantReactionState.reactionState = undefined;
+          } else {
+            if (
+              call.capabilitiesFeature?.capabilities.reaction &&
+              isReactionBeingPlayedNow(call.localParticipantReactionState.reactionState?.receivedAt)
+            ) {
+              clearParticipantReactionState(this, callId, participantKey);
```
JamesBurnside, line 443: nit: is this necessary? We are immediately setting the reactionState to a new value after this `if`.

JamesBurnside: If it isn't we can start to simplify this function, there's a lot of nesting (7+levels) and quite a bit of duplicate code.

Perhaps something like (**this is untested pseudocode**):
```ts
public setReceivedReactionFromParticipant(...)
...modifyState...
{
  const call = draft.calls[this._callIdHistory.latestCallId(callId)];
  if (!call) {
    return;
  }

  // New reaction has been received, clear any existing reaction timeout triggers
  clearTimeout(this._timeOutId[participantKey]);

  // Set the new reaction state on the appropriate participants
  const newReactionState = reactionMessage ? { reactionMessage: reactionMessage, receivedAt: new Date() } : undefined;
  const participant = call.remoteParticipants[participantKey];
  if (paticipantIsRemoteParticipant(participantId) {
    participant.reactionState = newReactionState;
  } else if (participantIsLocalParticipant(participantId)) {
    call.localParticipantReactionState = newReactionState;
  }

  // Set the new reaction state to clear after a timeout
  reactionMessage && this._timeOutId[participantKey] = setTimeout(() => {
    clearParticipantReactionState(this, callId, participantKey);
  }, 2170);
}
```

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -1022,3 +1077,19 @@ const findOldestCallEnded = (calls: { [key: string]: { endTime?: Date } }): stri
...
+
+function isReactionBeingPlayedNow(receivedAt: Date | undefined) {
+  if (receivedAt == undefined) {
+    return false;
+  }
+  const currentTimeStamp = new Date();
+  const unixBasedCurrentTimeStamp = Math.floor(currentTimeStamp.getTime() / 1000);
+  const unixBasedReceivedTimeStamp = Math.floor(receivedAt?.getTime() / 1000);
+
+  return receivedAt != undefined && unixBasedCurrentTimeStamp - unixBasedReceivedTimeStamp < 3000;
```
JamesBurnside, line 1094: nit: already checked `receivedAt != undefined`

```suggestion
  return unixBasedCurrentTimeStamp - unixBasedReceivedTimeStamp < 3000;
```

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -221,6 +229,16 @@ const VideoTileMoreOptionsButton = (props: {
   );
 };
 
+/* @conditional-compile-remove(reaction) */
+const styleSheet = document.styleSheets[0];
+/* @conditional-compile-remove(reaction) */
+const keyframes1 = `@keyframes play {
```
JamesBurnside, line 235: name `play` could easily conflict with existing keyframe names. Consider using an auto-generated guid/uuid

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -221,6 +229,16 @@ const VideoTileMoreOptionsButton = (props: {
...
+  from { background-position-y: 8568px; }
+  to { background-position-y: 0px; }
+}`;
+/* @conditional-compile-remove(reaction) */
+styleSheet.insertRule(keyframes1, styleSheet.cssRules.length);
```
JamesBurnside, line 240: This dangerous as it is happening globally at import time and reliant on `document.styleSheets[0]` existing at this point in time.

It looks like fluent has a function for adding keyframes that does such checks: https://github.com/microsoft/fluentui/blob/master/packages/merge-styles/src/keyframes.ts with example usage: https://github.com/microsoft/fluentui/blob/master/packages/react/src/components/Spinner/Spinner.styles.tsx#L19

Can you try using that and defining in the videoTile.styles file?

packages/react-components/src/components/utils/videoTileStylesUtils.tsx
```diff
@@ -10,3 +10,29 @@ export const getVideoTileOverrideColor = (isVideoRendered: boolean, theme: Theme
...
+  baseTimeStamp.setDate(1);
+  baseTimeStamp.setHours(0, 0, 0, 0);
+
+  let baseUnixTimestamp = Math.floor(baseTimeStamp.getTime() / 1000);
+  return Math.floor(new Date().getTime() / 1000) - baseUnixTimestamp;
+};
+
+/* @conditional-compile-remove(reaction) */
+/**
+ * @private
```
JamesBurnside, line 30: nit: add comment about this being temporary, sometimes temp code lives much longer than we want or focus shifts elsewhere and its not removed

packages/react-components/src/types/ParticipantListParticipant.ts
```diff
@@ -32,6 +35,23 @@ export type RaisedHand = {
...
+ *
+ * @public
+ */
+export type Reaction = {
+  /**
+   * Specifies the type of reaction videoTile should render i.e. like, heart etc.
+   */
+  reactionType: string;
+  /**
+   * Received timestamp relative to 01-01-2020 0:0:0 hours
```
JamesBurnside, line 50: Why is the timestamp relative to `01-01-2020 0:0:0 hours`? Can't it just be any `Date`?

packages/react-components/src/types/ParticipantListParticipant.ts
```diff
@@ -32,6 +35,23 @@ export type RaisedHand = {
...
+/**
+ * Reaction state with reaction type to render
+ *
+ * @public
+ */
+export type Reaction = {
+  /**
+   * Specifies the type of reaction videoTile should render i.e. like, heart etc.
+   */
+  reactionType: string;
```
JamesBurnside, line 48: instead of `string` - should we list what we support?
```suggestion
  reactionType: 'like' | 'heart' | ' ...
```

JamesBurnside: I guess this will be determined by the solution for how we retrieve the assets

JamesBurnside (changes requested):
Overall looking great! Just some comments added

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
Mohtasim: This actually reads as `call.localParticipantState.reactionState`

JamesBurnside: > This actually reads as `call.localParticipantState`

Its `localParticipantReactionState` not just `localParticipantState`, so if its `localParticipantReactionState.reactionState` its like saying call.reactionState.reactionState

Mohtasim: I see. shall we then remove the localparticipantreactionstate altogether? It makes sense to just keep ReactionState. If we need any extra state for locals, we can then create one later.

packages/calling-stateful-client/src/CallContext.ts
Mohtasim: Line 443 allows us to replace the ongoing reaction that was received for rendering. 
This would allow the contoso user to send a corrected reaction if in critical situation a wrong reaction was sent by mistake which might send the wrong sensitivity in the meeting.

packages/react-components/src/components/VideoTile.tsx
Mohtasim: play is already a keyword for css style animation rendering. We do not need this keyframe altogether as we can control the height and rendering with in-line styles. Removing this and testing to change.

packages/react-components/src/types/ParticipantListParticipant.ts
Mohtasim: Yes. also how we want to use capability. For now, can we keep it as string?

packages/react-components/src/components/VideoTile.tsx
Mohtasim: Will check this and update accordingly.

common/config/webpack/sampleapp.webpack.config.js
Mohtasim: I've reversed it. Thanks.

packages/calling-stateful-client/src/CallContext.ts
Mohtasim: Thanks for the suggested refactoring. I understand the current case handlings are not the best looking code and can be improved. 
We'll have more changes in this stub with capability check only for teams meeting but not for group calls. Right now we only check if capability exists or not! 
The cases missing in the suggested refactoring is that - 
* when we are doing - clearTimeOut(...) 
`// New reaction has been received, clear any existing reaction timeout triggers
  clearTimeout(this._timeOutId[participantKey]);`
  We need to get update the state as well to clear the current rendering, otherwise videoTile will just render another reaction (the newly received one) on top the old one rather than replacing it. 
  To clear the state, we need to know if the participant is a remote one or a local one. Thus I am clearing them under the check of detecting the participant type.

Mohtasim: Refactored the stub. Tested manually as well.

packages/calling-component-bindings/src/participantListSelector.ts
```diff
@@ -82,7 +92,9 @@ const convertRemoteParticipantsToParticipantListParticipants = (
             participant.isSpeaking,
             /* @conditional-compile-remove(raise-hand) */
             participant.raisedHand,
-            localUserCanRemoveOthers
+            localUserCanRemoveOthers,
+            /* @conditional-compile-remove(reaction) */
+            remoteParticipantReaction
```
JamesBurnside, line 97: `memoizeFn` memoizes the result based on the in params.  However, in javascript this:
image
Will create a new object each time it runs, and objects are passed by reference when pushed to functions, so this will invalidate the memoize on each run.

Two alternative options:
1. (probably easier) Pass `participant.reactionState.reactionMessage.reactionType` and `participant.reactionState.receivedAt` into the memoize function and construct the new object there
2. use `memoizeFn` to first memozie the remoteParticipantReaction object and then pass that in

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
```diff
@@ -71,7 +80,9 @@ export const _videoGalleryRemoteParticipantsMemo: (
             state,
             displayName,
             /* @conditional-compile-remove(raise-hand) */
-            participant.raisedHand
+            participant.raisedHand,
+            /* @conditional-compile-remove(reaction) */
+            remoteParticipantReaction
```
JamesBurnside, line 85: (same as other comment on ensuring this param is memoized)

packages/calling-stateful-client/review/beta/calling-stateful-client.api.md
```diff
@@ -115,6 +116,8 @@ export interface CallState {
     isMuted: boolean;
     isScreenSharingOn: boolean;
     kind: CallKind;
+    // @beta
+    localParticipantReactionState?: ReactionState;
```
JamesBurnside, line 120: drop the `state` suffix

packages/react-components/src/components/VideoTile.tsx
```diff
@@ -404,6 +427,41 @@ export const VideoTile = (props: VideoTileProps): JSX.Element => {
           </Stack>
         )}
 
+        {
+          /* @conditional-compile-remove(reaction) */
+          canRenderReaction && (
+            <Stack
+              className={mergeStyles(videoContainerStyles, {
+                display: 'flex',
```
JamesBurnside, line 435: nit: we try to keep all styles in a `<name>.styles.ts` files to keep these files as short as possible

JamesBurnside (approved):

packages/calling-component-bindings/src/participantListSelector.ts
Mohtasim: In order to do minimal change and keep the public API design as it is, I have chose the 2nd approach. Thanks for this suggession.

packages/calling-component-bindings/src/utils/videoGalleryUtils.ts
Mohtasim: Updated accordingly.

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -73,6 +75,8 @@ export class CallContext {
   private _emitter: EventEmitter;
   private _atomicId: number;
   private _callIdHistory: CallIdHistory = new CallIdHistory();
+  /* @conditional-compile-remove(reaction) */
+  private _timeOutId: { [key: string]: any } = {};
```
dmceachernmsft, line 79: nit: we are moving away from allowing `any` as a type in the repo. I believe that the type that should be here is `number` 😊

https://developer.mozilla.org/en-US/docs/Web/API/setTimeout

dmceachernmsft (approved):
Just one more comment otherwise looks great! Awesome work! 😊

packages/calling-stateful-client/src/CallContext.ts
Mohtasim: Thanks for this pointer Donald. Will update accordingly.

PR #3818, opened 2023-11-22
[Calling sounds] Sound api update 

dmceachernmsft:
# What

Flatten the shape of the sounds API on the call adapter so chat can do the same on the chat adapter
# Why

API was originally designed for chat to get added to the `soundOptions` bag. This doesn't extend well without a breaking change so making it flat allows for Chat to do the same in the future and maybe wrap them together for CWC adapter.

# How Tested

Tested changes work locally and update API files.

### DIFF
--- a/packages/communication-react/review/beta/communication-react.api.md
+++ b/packages/communication-react/review/beta/communication-react.api.md
@@ -1757,9 +1757,7 @@ export type CommonCallAdapterOptions = {
         onResolveDependency?: () => Promise<VideoBackgroundEffectsDependency>;
     };
     onFetchProfile?: OnFetchProfileCallback;
-    soundOptions?: {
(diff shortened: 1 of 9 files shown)


mgamis-msft (approved):

packages/communication-react/review/beta/communication-react.api.md
```diff
@@ -3696,7 +3694,6 @@ export interface SitePermissionsStyles extends BaseCustomStyles {
 // @beta
 export type SoundEffect = {
     path: string;
```
JamesBurnside, line 3696: path -> url

dmceachernmsft: this is coming in the following PR 😊

PR #3819, opened 2023-11-22
[Rooms] Update rooms client version

mgamis-msft:
# What
Upgrade rooms client to versions 1.1.0-beta.1 and 1.1.0

# Why
Ensure rooms client is up to date for sample apps

# How Tested
Created and joined a room in Calling sample

# Process & policy checklist


samples/Server/package.json
```diff
@@ -33,7 +33,7 @@
     "http-errors": "^1.6.3",
     "morgan": "^1.9.1",
     "node-fetch": "2.6.7",
-    "@azure/communication-rooms": "1.0.0-beta.1"
+    "@azure/communication-rooms": "1.1.0-beta.1 || ^1.1.0"
```
dmceachernmsft, line 36: Do we need to do the rush common versions here?

mgamis-msft: We don't need to because rooms is not using for the ui library. It's only used by the Server.

carocao-msft (approved):

PR #3820, opened 2023-11-22
[Rooms] Stop camera when capability is lost and lower hand when role changes to Consumer

mgamis-msft:
# What
Stop camera when capability is lost and lower hand when role changes to Consumer

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3489516
https://skype.visualstudio.com/SPOOL/_workitems/edit/3489520

# How Tested
Tested on bug bash app with custom button to change roles

# Process & policy checklist


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -1183,7 +1191,12 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
   /* @conditional-compile-remove(capabilities) */
   private capabilitiesChanged(data: CapabilitiesChangeInfo): void {
     if (data.newValue.turnVideoOn?.isPresent === false) {
-      this.stopCamera();
+      // Only stop camera when the call state is not on hold. The Calling SDK does not allow us to stop camera when
```
JamesBurnside, line 1194: > The Calling SDK does not allow us to stop camera when the call state is on hold.
:O

Can we file a bug on this

JamesBurnside (approved):

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
mgamis-msft: Will do 👍

PR #3821, opened 2023-11-22
[Rooms] Filter out consumers from participant list and video gallery

mgamis-msft:
# What
Filter out consumers from getRemoteParticipants selector

# Why
Remote participants that are originally not consumers but updated to consumers during the call are still shown in the participant list

# How Tested
Tested on bug bash app with custom button to change roles during a call

# Process & policy checklist

### DIFF
--- a/packages/calling-component-bindings/src/baseSelectors.ts
+++ b/packages/calling-component-bindings/src/baseSelectors.ts
@@ -86,6 +86,17 @@ export const getRemoteParticipants = (
       [keys: string]: RemoteParticipantState;
     }
   | undefined => {
+  /* @conditional-compile-remove(rooms) */
+  // Filter out consumers from remote participants
+  {
+    const remoteParticipants = { ...state.calls[props.callId]?.remoteParticipants };
+    Object.keys(remoteParticipants).forEach((k) => {
+      if (remoteParticipants[k].role === 'Consumer') {
+        delete remoteParticipants[k];
+      }
+    });
+    return remoteParticipants;
+  }
   return state.calls[props.callId]?.remoteParticipants;
 };
 
--- a/change-beta/@azure-communication-react-185f3b34-d242-4c37-b311-21e942bc8877.json
+++ b/change-beta/@azure-communication-react-185f3b34-d242-4c37-b311-21e942bc8877.json
@@ -1,0 +1,9 @@
+{
+  "type": "prerelease",
+  "area": "fix",
+  "workstream": "Rooms",
+  "comment": "Filter out consumers from remote participant list",
+  "packageName": "@azure/communication-react",
+  "email": "79475487+mgamis-msft@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@azure-communication-react-185f3b34-d242-4c37-b311-21e942bc8877.json
+++ b/change/@azure-communication-react-185f3b34-d242-4c37-b311-21e942bc8877.json
@@ -1,0 +1,9 @@
+{
+  "type": "prerelease",
+  "area": "fix",
+  "workstream": "Rooms",
+  "comment": "Filter out consumers from remote participant list",
+  "packageName": "@azure/communication-react",
+  "email": "79475487+mgamis-msft@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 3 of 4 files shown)


dmceachernmsft (approved):

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -86,6 +86,17 @@ export const getRemoteParticipants = (
       [keys: string]: RemoteParticipantState;
     }
   | undefined => {
+  /* @conditional-compile-remove(rooms) */
+  // Filter out consumers from remote participants
+  {
+    const remoteParticipants = { ...state.calls[props.callId]?.remoteParticipants };
+    Object.keys(remoteParticipants).forEach((k) => {
+      if (remoteParticipants[k].role === 'Consumer') {
+        delete remoteParticipants[k];
```
dmceachernmsft, line 95: do we want to delete? or just filter out?

dmceachernmsft (approved):

packages/calling-component-bindings/src/baseSelectors.ts
mgamis-msft: I created a copy of remote participants from state so it is safe to delete it from the copy to essentially filter out consumers when the remote participants are passed to the composite

JamesBurnside: nit: `filter` would be a better option here for readability (easy to miss that the object is cloned before mutation, so in refactors it could be lost)

Consider:
```ts
return remoteParticipants.filter((remoteParticipants) => remoteParticipants.role !=== 'Consumer');
```

packages/calling-component-bindings/src/baseSelectors.ts
```diff
@@ -86,6 +86,17 @@ export const getRemoteParticipants = (
...
+      }
+    });
+    return remoteParticipants;
```
JamesBurnside, line 98: My understanding is these baseSelectors should avoid returning new objects. Selectors build of these using `reselect.createSelector`. AFAIK the createSelector function is memoized based on the inputs. Meaning that when the input are not changed, it returns a memoized result (meaning react also receives a memoized result).
However this change will be creating a new remoteParticipants object every time its called, meaning this will bust the memoization of all the createSelectors using this.

Instead create a new selector called getFilteredRemoteParticipants or something that is a selector on top of this base selector. And have all the relevant sub-selectors use the new selector instead
```ts
export const getFilteredRemoteParticipants = reselect.createSelector(
  [getRemoteParticipants],
  (remoteParticipants) => {
    return remoteParticipants.filter((remoteParticipants) => remoteParticipants.role !=== 'Consumer');
  }
);
```

Let me know if I'm misunderstanding the base selector usage and how things get memoized!
cc @PorterNan if he can confirm.

JamesBurnside (changes requested):

packages/calling-component-bindings/src/baseSelectors.ts
mgamis-msft: Thanks for this tip about `reselect.createSelector`! I read up a bit about it and you are right that it only returns a new object when one of the values change. 
image
[source](https://react-redux.js.org/api/hooks#using-memoizing-selectors)
Updated.

packages/calling-component-bindings/src/getRemoteParticipantsExcludingConsumers.ts
```diff
@@ -0,0 +1,17 @@
...
+import { getRemoteParticipants } from './baseSelectors';
+
+/**
+ * @private
+ */
+export const getRemoteParticipantsExcludingConsumers = createSelector([getRemoteParticipants], (remoteParticipants) => {
+  /* @conditional-compile-remove(rooms) */
+  {
+    const participants = Object.values(remoteParticipants ?? {});
+    return participants.filter((remoteParticipants) => remoteParticipants.role !== 'Consumer');
```
JamesBurnside, line 14: I think my code snippet was poor, this will correctly return an array of remoteParticipants who aren't consumers, but it won't return the same as the conditional-compile return `return remoteParticipants` which is a key map of ids to remote participants

mgamis-msft: It worked when I tested it locally. Will change it back to key map of ids to be sure.

JamesBurnside: Yeah I'm surprised/confused nothing broken in the build and beta and stable tests - but they were returning two different interfaces right?

PR #3824, opened 2023-11-22
[Calling sounds] Update sample app configs to include sounds in deploy

dmceachernmsft:
# What

Update webpack to include the sound files in the deployment
# Why

when app is deployed we will include the sounds

# How Tested

Deployed sounds test app to see sounds live


JamesBurnside (approved):

mgamis-msft (approved):

common/config/webpack/sampleapp.webpack.config.js
```diff
@@ -77,6 +77,7 @@ const webpackConfig = (sampleAppDir, env, babelConfig) => {
         patterns: [
           { from: path.resolve(sampleAppDir, "public/manifest.json"), to: "manifest.json" },
           { from: path.resolve(sampleAppDir, "public/backgrounds"), to: "backgrounds",  noErrorOnMissing: true },
+          { from: path.resolve(sampleAppDir, "public/sounds"), to: "sounds",  noErrorOnMissing: true }
```
JamesBurnside, line 80: to avoid needing this in future, we should have a public/assets folder and have that be copied over (then it has subfolders for backgrounds, sounds etc. but any new subfolder would be auto copied)

JamesBurnside: (but thats longer than a one line change 😄)

dmceachernmsft: on it!

JamesBurnside (approved):

PR #3842, opened 2023-11-29
Make configuration page text stand out against a noisy background

JamesBurnside:
# What

- Convert title text to be text within an SVG
- This allows us to apply an outline to the text
- Also apply a shadow to the text per figma
- SVG does not support word wrap - so have custom logic to perform word wrapping
  - This works by rendering the original text hidden, then using that hidden element to calculate length of text and see when to manually create new lines.

# Why

Make text legible on noisy/poor contrast custom backgrounds.

# How Tested

Locally in mobile & desktop against different backgrounds and different call title lengths

| Some quick screenshots |
| -- | 
| image |
| image |
| image |
| image |
| image |


packages/react-composites/src/composites/CallComposite/components/SvgWithWordWrapping.tsx
```diff
@@ -0,0 +1,115 @@
...
+ * @param lineHeightPx The height of each line in pixels.
+ * @returns The number of lines of text.
+ */
+const convertTextToWrappedText = (
+  inputTextElement: SVGTextElement,
+  outputTextElement: SVGTextElement,
+  maxWidth: number,
+  lineHeightPx: number
+): number => {
+  const words = inputTextElement.textContent?.split(' ') ?? [];
```
mgamis-msft, line 64: Would a string like 'c  a t' be okay here?
Tested in playcode
```
'c  a t'.split(' ')
```
returns
```
["c", "", "a", "t"]
```
I'm worried it may unnecessarily add a tspan for the empty string

mgamis-msft: Actually nvm this is fine. the empty string will add no width to the testLine. Disregard this.

JamesBurnside: (Great question though, I hadn't tested this edge case)

packages/react-composites/src/composites/CallComposite/styles/CallConfiguration.styles.ts
```diff
@@ -46,7 +52,7 @@ export const configurationContainerStyle = (desktop: boolean, backgroundImageUrl
   root: {
     height: '100%',
     width: '100%',
-    padding: '2rem 1rem 2rem 1rem',
+    padding: desktop ? '2rem 1rem 2rem 1rem' : '1rem 1rem 2rem 1rem',
```
dmceachernmsft, line 55: I think the padding on the top for mobile should be 1.5rem see image ->
image

dmceachernmsft: This is also being affected 
image

dmceachernmsft (approved):

packages/react-composites/src/composites/CallComposite/styles/CallConfiguration.styles.ts
JamesBurnside: Thanks for checking these! Yep for mobile the reduced top padding is intentional to tighten up the spacing (particularly on limited landscape mobile), and desktop yep the increased distanced between title text and device section is intentional to match the 32px on figma
image

PR #3843, opened 2023-11-29
[Storybook] Add snippets to error page

dmceachernmsft:
# What

Update the error handling page with some code snippets to help showcase how to actually do the advance error handling
# Why

gives more detail to contoso about how to handle different errors that we emit 

https://skype.visualstudio.com/SPOOL/_workitems/edit/3517318
# How Tested

inspected locally and tested the snippets


packages/storybook/stories/snippets/ErrorHandlingAdapters.snippet.tsx
```diff
@@ -0,0 +1,72 @@
...
+     * Subscribing to this event on the call adapter will allow you to see the different
+     * errors that can occur in the calling experience. When these events happen using the callback you
+     * can update your application to render the appropriate UI to recover from the error.
+     *
+     * If you do nothing on this event the CallComposite and adapter do catch the error internally
+     * and will recover from it on its own.
+     */
+    adapter.on('error', (e: any) => {
+      /**
+       * Recovery code goes here
```
JamesBurnside, line 49: personal opinion nit: I wouldn't call it `recovery` code as I read that as the application/call is terminating.
Although if we add the error boundary then we will be piping errors that cause the call to terminate...

JamesBurnside (approved):

packages/storybook/stories/snippets/ErrorHandlingStatefulClient.snippet.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const { userAccessToken, userId, groupId, displayName } = props;
+
+  const [statefulCallClient, setStatefulCallClient] = useState<StatefulCallClient>();
+  const [callAgent, setCallAgent] = useState<CallAgent>();
+  const [call, setCall] = useState<Call>();
+
+  /**
+   * This state we will store the message from the latest error that was emitted.
+   */
+  const [latestError, setLatestError] = useState<string>();
```
mgamis-msft, line 36: Should we use the latestError in the CallScreen component?

mgamis-msft (approved):

packages/storybook/stories/snippets/ErrorHandlingStatefulClient.snippet.tsx
dmceachernmsft: Nah this is just an example for what they can do. figured kicking it out to state would be good enough

PR #3844, opened 2023-11-29
[Chat] Add a Narrator announcement when a message is deleted 

vhuseinova-msft:
# What

Added an announcement when a message is deleted

# Why

# How Tested

Chat sample app

# Process & policy checklist


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -715,6 +718,47 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
     [onFetchAttachments]
   );
 
+  const localeStrings = useLocale().strings.messageThread;
+  const strings = useMemo(() => ({ ...localeStrings, ...props.strings }), [localeStrings, props.strings]);
+
+  const latestDeletedMessageIdRef = useRef<string | undefined>(undefined);
```
emlynmac, line 724: Should this be able to handle multiple outstanding deletions?

vhuseinova-msft: As we allow only 1 message id with `onDeleteMessage`, the messages array should always be up to date for a user and only single message can be deleted (otherwise, user will be able to delete the same message multiple times). So there should be only 1 deleted message as for now. We can always change this behavior in case if we will allow multiple messages to be deleted.

emlynmac (approved):

vhuseinova-msft:
> Tested and looks good!
> 
> Just curious, do we have an accessibility bug having a similar announcement on a message being edited?

I don't htink we have one under the Accessibility group

PR #3855, opened 2023-12-04
[TypeScript] Break noImplicitAny to packlets

emlynmac:
# What
noImplicitAny was disabled at the repository level. This removes TypeScript's ability to work effectively

This PR breaks the override out to each packlet, rather than the top level.

# Why
There are lots of changes required to fix in the repository. These need to be done in a controlled way.

# How Tested
Build.

# Process & policy checklist

### DIFF
--- a/packages/chat-stateful-client/src/TestHelpers.ts
+++ b/packages/chat-stateful-client/src/TestHelpers.ts
@@ -91,7 +91,7 @@ export type ChatClientWithEventTrigger = ChatClient & {
  * @private
  */
 export function createMockChatClient(): ChatClientWithEventTrigger {
-  const mockEventHandlersRef = { value: {} };
+  const mockEventHandlersRef: { value: any } = { value: {} };
   const mockChatClient: ChatClientWithEventTrigger = {} as any;
 
   mockChatClient.createChatThread = async (request) => {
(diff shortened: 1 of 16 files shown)


JamesBurnside (approved):

carocao-msft (approved):

emlynmac:
> Why do some have the comment `// TODO: MAKE THIS true` and some do not

I was hoping to leave that there, but rush is a bit stricter in places and had to remove.

JoshuaLai (approved):

packages/chat-stateful-client/src/TestHelpers.ts
```diff
@@ -91,7 +91,7 @@ export type ChatClientWithEventTrigger = ChatClient & {
  * @private
  */
 export function createMockChatClient(): ChatClientWithEventTrigger {
-  const mockEventHandlersRef = { value: {} };
+  const mockEventHandlersRef: { value: any } = { value: {} };
```
emlynmac, line 94: This was a quick change to enable the test code to pass; we now have `noImplicitAny` across the chat-stateful-client packlet.
Can look into a proper typing with the rest of the `any` removal.

PR #3864, opened 2023-12-06
Hide screenshare button for iOS

mgamis-msft:
# What
Hide screenshare button for iOS including IPads where the screensharing is not possible.

# Why
https://skype.visualstudio.com/SPOOL/_workitems/edit/3405782

# How Tested
Tested local Calling sample and CallWithChat sample on IPad

# Process & policy checklist

### DIFF
--- a/change-beta/@azure-communication-react-aec4f243-84ab-4bb4-aad5-4e73d5ff20c2.json
+++ b/change-beta/@azure-communication-react-aec4f243-84ab-4bb4-aad5-4e73d5ff20c2.json
@@ -1,0 +1,9 @@
+{
+  "type": "none",
+  "area": "fix",
+  "workstream": "Calling and CallWithChat samples",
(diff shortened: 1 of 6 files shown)


dmceachernmsft (approved):

dmceachernmsft:
This is a good fix for our samples but I wonder if we need to do more to highlight in the docs that Contoso should hide the button in their apps if they are wanting to show an desktop experience on an ipad or something?

carocao-msft (approved):

mgamis-msft:
> This is a good fix for our samples but I wonder if we need to do more to highlight in the docs that Contoso should hide the button in their apps if they are wanting to show an desktop experience on an ipad or something?

I think we can just refer to the custom button docs by Edward in storybook if a customer wants to know how to hide a button. Not sure if there should be new page added to storybook that would help customers on top what is already documented in custom button docs.

PR #3865, opened 2023-12-06
Support local screenshare stream at the stateful layer only

JamesBurnside:
# What
- Stateful client support both Video localvideostream and Screenshare localvideostream
- **important** this keys local video streams in stateful by `type` meaning there can only be one camera feed, one screenshare etc. This all the calling SDK supports also but could be a limitation in future (in which case we can update the key but this is complex as there is no good id to use, using device id involves complex tracking when the user switches device source)
# Why

Basis to supporting a local screen share stream - see #3785.

Once we have designs we can move further with adding component and composite support.

# How Tested

In #3785 performed full e2e testing.


packages/calling-stateful-client/src/InternalCallContext.ts
```diff
@@ -142,23 +139,44 @@ export class InternalCallContext {
 
   public setLocalRenderInfo(
     callId: string,
+    streamKey: string,
```
dmceachernmsft, line 142: is stream key the type of stream? `Key: 'screenshare' | 'video'` or something?

packages/calling-stateful-client/src/StreamUtils.ts
```diff
@@ -140,14 +154,20 @@ async function createViewVideo(
       internalContext.setRemoteRenderInfo(
         callId,
         participantKey,
-        streamId,
+        remoteStreamId,
```
dmceachernmsft, line 157: Now that we are tracking multiple streams for the local user to do we not get a stream id for the local user?

dmceachernmsft (approved):
Very cool! just a couple questions about the localStreamKey but this is great and we should bug bash this as much as possible in selfhost before Miguel's release! 😊

packages/calling-stateful-client/src/StreamUtils.ts
JamesBurnside: We don't at all ☹️ we do have streamInfo.device.id but when switchSource is called that id cahnges are the underhood so becomes a really hard to maintain key to use

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -759,8 +784,7 @@ export class CallContext {
   public deleteDeviceManagerUnparentedView(localVideoStream: LocalVideoStreamState): void {
     this.modifyState((draft: CallClientState) => {
       const foundIndex = draft.deviceManager.unparentedViews.findIndex(
-        (stream) =>
-          stream.source.id === localVideoStream.source.id && stream.mediaStreamType === localVideoStream.mediaStreamType
+        (stream) => stream.mediaStreamType === localVideoStream.mediaStreamType
```
mgamis-msft, line 787: We don't match source ids because the source id changes when you switch cameras, right?

packages/calling-stateful-client/src/InternalCallContext.ts
```diff
@@ -51,8 +48,8 @@ export class InternalCallContext {
   // <CallId, <ParticipantKey, <StreamId, RemoteRenderInfo>>
   private _remoteRenderInfos = new Map<string, Map<string, Map<number, RemoteRenderInfo>>>();
 
-  // <CallId, LocalRenderInfo>.
-  private _localRenderInfos = new Map<string, LocalRenderInfo>();
+  // <CallId, <MediaStreamType, LocalRenderInfo>>.
+  private _localRenderInfos = new Map<string, Map<string, LocalRenderInfo>>();
 
   // Used for keeping track of rendered LocalVideoStreams that are not part of a Call.
   // The key is the stream ID. We assume each stream ID to only have one owning render info
```
mgamis-msft, line 55: Should `_unparentedRenderInfos` now be `Map<'screenshare' | 'video', LocalRenderInfo>`?

JamesBurnside: Yes! I will type that stronger

mgamis-msft (approved):

packages/calling-stateful-client/src/InternalCallContext.ts
JamesBurnside: updated

packages/calling-stateful-client/src/InternalCallContext.ts
JamesBurnside: updated

PR #3871, opened 2023-12-06
Assume 'AzureCommunicationLibrary' for telemetry to make consistent with iOS and Android

palatter:
# What

We assume AzureCommunicationLibrary, as we don't currently have any public API to inject otherwise.
This is consistent with the native iOS and Android implementations of telemetry.

# Why

# How Tested

# Process & policy checklist


AmyL219 (approved):
Don't we have the unit test for this number?

packages/acs-ui-common/src/telemetry.ts
```diff
@@ -59,7 +59,9 @@ const getTelemetryImplementationHint = (telemetryImplementationHint: _TelemetryI
  * @internal
  */
 export const _getApplicationId = (telemetryImplementationHint: _TelemetryImplementationHint): string => {
-  const highLevelArtifact = 0;
+  // We assume AzureCommunicationLibrary, as we don't currently have any public API to inject otherwise.
```
vhuseinova-msft, line 62: Should it be `@azure/communication-react` here instead of `AzureCommunicationLibrary`?

palatter: The name/schema is platform agnostic, so we have to use AzureCommunicationLibrary.

PR #3876, opened 2023-12-07
[TDL bug fix] Camera bad state in lobby page

dmceachernmsft:
# What

Fixes issue where camera reflects being on when turning off the camera on the lobby page because of a stray unparented view we weren't disposing
# Why

camera button reflects the correct state of the stream

https://skype.visualstudio.com/SPOOL/_workitems/edit/3520640
# How Tested

Validated locally


packages/react-composites/src/composites/CallComposite/adapter/AzureCommunicationCallAdapter.ts
```diff
@@ -409,6 +409,16 @@ export class AzureCommunicationCallAdapter<AgentType extends CallAgent | BetaTea
         this.context.setCurrentCallId(this.call.id);
       }
 
+      // if we have a call state we need to dispose of the unparented views
```
JamesBurnside, line 412: nit: update this comment to describe the scenario of why it exists to help prevent it being deleted in future
```suggestion
// If the call connects we need to clean up any previous unparentedViews
```

JamesBurnside: From this -- should we also clean up if it goes into the disconnected or disconnecting states too?

JamesBurnside: Actually should we clean up every state except `connecting`?

mgamis-msft (approved):

PR #3889, opened 2023-12-07
[Calling sounds] Fix issue where ringing sound keeps playing on the disconnecting screen

dmceachernmsft:
# What

Pauses the ringing sound when the call hits the disconnecting state
# Why

Stops the ring tone sound from playing when the call is in the middle of disconnecting

# How Tested

Validated locally


mgamis-msft (approved):

packages/react-composites/src/composites/CallComposite/adapter/CallingSoundSubscriber.ts
```diff
@@ -45,6 +45,11 @@ export class CallingSoundSubscriber {
         this.soundsLoaded.callRingingSound.loop = false;
         this.soundsLoaded.callRingingSound.pause();
       }
+      if (this.call.state === 'Disconnecting') {
```
JamesBurnside, line 48: how come this is a new if statement and not in the above if statement?

JamesBurnside: instead of if connected || disconnecting || disconnected - to avoid it still playing if it gets into other non-ringing states, could this be `if state !== connecting && state !== connecting ringing`?

dmceachernmsft: yup I like that, will update

JamesBurnside (approved):

PR #3896, opened 2023-12-08
[Bugfix] Microphone lobby fix

dmceachernmsft:
# What

Update disabled state to make sure that the lobby state allows for muting and unmuting
# Why

Gives the user more control on the lobby screen

https://skype.visualstudio.com/SPOOL/_workitems/edit/3520938
# How Tested

Validating locally


JamesBurnside (approved):

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/buttons/Microphone.tsx
```diff
@@ -41,8 +41,7 @@ export const Microphone = (props: {
    * When call is in Lobby, microphone button should be disabled.
```
mgamis-msft, line 41: I think we should update this code comment

packages/react-composites/src/composites/CallComposite/components/buttons/Microphone.tsx
```diff
@@ -41,8 +41,7 @@ export const Microphone = (props: {
...
    * This is due to to headless limitation where a call can not be muted/unmuted in lobby.
    */
-  if (_isInLobbyOrConnecting(callStatus)) {
-    microphoneButtonProps.disabled = true;
+  if (callStatus === 'Connecting') {
     // Lobby page should show the microphone status that was set on the local preview/configuration
     // page until the user successfully joins the call.
     microphoneButtonProps.checked = isLocalMicrophoneEnabled;
```
mgamis-msft, line 47: Not related to this PR but `getLocalMicrophoneEnabled` selector should be changed to `getLocalMicrophoneOn`. The verb enabled makes it sound like it should decide if the button is enabled/disabled.

mgamis-msft (approved):

PR #3900, opened 2023-12-11
Update typescript to 5.3.3

JamesBurnside:
# What

- Update typescript to 5.3.3
  - Update TS eslint config versions to meet new TS version
- Switch ttypescript (dead project) for ts-patch (direct replacement with migration support)
- Updated docs

# Why

Get onto LTS typescript for local development

# How Tested

Built locally, verified:
- ✅ ESM bundle exports of the npm meta-package were relative and not to `@` aliases.
   
    image

- ✅ Verified CJS bundle _looked_ fine - did not run.
- ✅ Ran callwithchat sample locally
- ✅ CI all green
- ✅ Verified storybook in CI link (Checked VideoGallery & MessageThread components, and CallWithChatComposite)
- ✅ Triggered an alpha release ([1.11.0-alpha-202312111717](https://github.com/Azure/communication-ui-library/actions/runs/7170973735)) and verified in a test app
   
    image


packages/calling-stateful-client/src/CallAgentDeclarativeCommon.ts
```diff
@@ -159,7 +159,7 @@ export abstract class ProxyCallAgentCommon {
   protected getCommon<
     AgentType extends CallAgent | /* @conditional-compile-remove(teams-identity-support) */ TeamsCallAgent,
     P extends keyof CallAgent
-  >(target: AgentType, prop: P): any {
+  >(target: AgentType, prop: P | 'incomingCalls'): any {
```
emlynmac, line 162: Nice; no more warnings!

emlynmac (approved):
Awesome, nice to get TypeScript support up to compliance!

packages/communication-react/package.json
```diff
@@ -62,12 +62,14 @@
   "main": "./dist/dist-cjs/communication-react/index.js",
   "module": "./dist/dist-esm/communication-react/src/index.js",
   "scripts": {
+    "prepare": "ts-patch install -s",
```
dmceachernmsft, line 65: What does this new prepare command do?

dmceachernmsft (approved):

packages/communication-react/package.json
JamesBurnside: Actually you're right I think this is not needed, we do live transformations during build (there is a fancy option to do transformations at install time but not what we're looking for). I'll try remove and see if it works

JamesBurnside: removed

PR #3901, opened 2023-12-11
[Chat] Update Chat SDK to 1.5.0-beta.1 and Signaling to beta 23 for beta and beta 22 for stable

Leah-Xia-Microsoft:
# What

Update Chat SDK to 1.5.0-beta.1 for beta flavour
Update Signaling to beta 23 for beta and beta 22 for stable
Update type for the new SDK

# Why

# How Tested

# Process & policy checklist


packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -28,11 +28,10 @@ import { ACSKnownMessageType } from './utils/constants';
 /* @conditional-compile-remove(data-loss-prevention) */
 import { DEFAULT_DATA_LOSS_PREVENTION_POLICY_URL } from './utils/constants';
 import { updateMessagesWithAttached } from './utils/updateMessagesWithAttached';
-
 /* @conditional-compile-remove(file-sharing) @conditional-compile-remove(teams-inline-images-and-file-sharing) */
-import { FileMetadata, ChatAttachmentType, InlineImageMetadata } from '@internal/react-components';
+import { FileMetadata, ChatAttachmentType as AttachmentType, InlineImageMetadata } from '@internal/react-components';
```
palatter, line 32: Is this because of the naming collision?

packages/chat-component-bindings/src/messageThreadSelector.ts
```diff
@@ -135,7 +135,21 @@ const processChatMessageContent = (message: ChatMessageWithStatus): string | und
 
 /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
 const generateImageAttachmentImgHtml = (attachment: ChatAttachment): string => {
-  return `\r\n<p><img alt="image" src="" itemscope="${attachment.contentType}" id="${attachment.id}"></p>`;
+  const contentType = extractAttachmentContentTypeFromName(attachment.name) ?? '';
+  return `\r\n<p><img alt="image" src="" itemscope="${contentType}" id="${attachment.id}"></p>`;
+};
+
+/* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+const extractAttachmentContentTypeFromName = (name?: string): string | undefined => {
```
jimchou-dev, line 143: would it make more sense to just return empty string instead of undefined?

packages/chat-component-bindings/src/messageThreadSelector.ts
Leah-Xia-Microsoft: > Is this because of the naming collision?

@palatter  Yes. Now that we have the same type name from both our library and ChatSDK, we have to make the distinguish here.

jimchou-dev (approved):

palatter (approved):

PR #3906, opened 2023-12-12
[Chat] Cherry pick chat dependency update

Leah-Xia-Microsoft:
# What

Update Chat SDK to 1.5.0-beta.1 and Signaling to beta 23 for beta flavour and beta 22 for stable flavour

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/chat-stateful-client/package.json
+++ b/packages/chat-stateful-client/package.json
@@ -40,11 +40,11 @@
     "nanoid": "3.3.6"
   },
   "peerDependencies": {
-    "@azure/communication-chat": ">=1.4.0"
+    "@azure/communication-chat": "1.5.0-beta.1 || >=1.4.0"
   },
   "devDependencies": {
-    "@azure/communication-chat": "^1.4.0",
-    "@azure/communication-signaling": "1.0.0-beta.21 || 1.0.0-beta.20",
+    "@azure/communication-chat": "1.5.0-beta.1 || ^1.4.0",
+    "@azure/communication-signaling": "1.0.0-beta.23 || 1.0.0-beta.22",
     "@babel/cli": "^7.23.4",
     "@babel/core": "^7.23.5",
     "@microsoft/api-documenter": "~7.23.14",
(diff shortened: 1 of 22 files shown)


JamesBurnside (changes requested):
Does this change any stable code? I only see beta dependencies and conditional-compilation lines changed. If there's change to stable code can that be an isolated PR. If not there shouldn't be a need to cherry-pick this into stable.

palatter (approved):

Leah-Xia-Microsoft:
> Does this change any stable code? I only see beta dependencies and conditional-compilation lines changed. If there's change to stable code can that be an isolated PR. If not there shouldn't be a need to cherry-pick this into stable.

@JamesBurnside The main part we want to bring in from this commit is the signalling change. It should be signalling beta 22 to match what chat sdk 1.4.0 is using under the hood.
I can only update signalling on the release branch if that's easier?

packages/chat-stateful-client/package.json
```diff
@@ -40,11 +40,11 @@
...
   },
   "peerDependencies": {
-    "@azure/communication-chat": ">=1.4.0"
+    "@azure/communication-chat": "1.5.0-beta.1 || >=1.4.0"
   },
   "devDependencies": {
-    "@azure/communication-chat": "^1.4.0",
-    "@azure/communication-signaling": "1.0.0-beta.21 || 1.0.0-beta.20",
+    "@azure/communication-chat": "1.5.0-beta.1 || ^1.4.0",
+    "@azure/communication-signaling": "1.0.0-beta.23 || 1.0.0-beta.22",
```
mgamis-msft, line 47: @JamesBurnside @azure/communication-signaling@1.0.0-beta.22 is a dependency change that affects this stable. This is needed to match @azure/communication-chat@1.4.0 that this current stable release is using.

JamesBurnside: `@azure/communication-signaling` is only a devDepenency and we only use it for types in tests - so shouldn't be impacting the stable package.
But also I'm down for updating it - no harm in it!
But we should _not_ have these changes (so either manually reverting them or a separate PR, whichever is easier):
image

cc @Leah-Xia-Microsoft

mgamis-msft: Ah good point. It's only a dev dependency so it won't affect the npm package.

Leah-Xia-Microsoft: Good point. Since separating this PR might cause unexpected issues, and since it's ok to have signaling 21 as just devDepenency, we will not cherry-pick this PR into the release branch.

PR #3916, opened 2023-12-14
Pin chrome version in playwright tests

JamesBurnside:
# What

- Pin chrome version used in playwright tests
- Update github actions to use pinned version

# Why

GitHub updates its runners monthly. This pulls in new chrome drivers (also released monthly). These new versions may have new UI changes breaking our regressions tests. We cannot wholesale update snapshots at this time as GitHub runner updates are rolled out over a few days so some PRs hit the old runner versions while some hit the new runner versions.

# How Tested

Still testing

Goal: CI passing on the newer runner images despite not having updated the snapshots.


emlynmac (approved):
Changes look good

dmceachernmsft (approved):

.github/workflows/ci.yml
```diff
@@ -212,6 +212,10 @@ jobs:
         run: npm install -g @microsoft/rush@5.112.1
       - name: Install dependencies
         run: rush install --max-install-attempts 3
+      - uses: browser-actions/setup-chrome@v1
```
vhuseinova-msft, line 215: I'm wondering if this can be used in `get_matrix` job (or something similar and be added to output, so later we will need to update it in 1 place?

.github/workflows/update-snapshots.yml
```diff
@@ -96,6 +96,10 @@ jobs:
         run: npm install -g @microsoft/rush@5.112.1
       - name: Install dependencies
         run: rush install --max-install-attempts 3
+      - uses: browser-actions/setup-chrome@v1
```
vhuseinova-msft, line 99: Similar question here: to have 1 job on which other will depend on and use it's output value in `CHROME_PATH`

.github/workflows/ci.yml
JamesBurnside: Each job runs on a different agent allocated/isolated space, but this is a good idea. I'm going to check in as is to unblock other PRs and then try this out

PR #3928, opened 2023-12-19
[Chat] Fix for an issue when file types icons weren't available for storybook

vhuseinova-msft:
# What

Fixed file types icons registration for CallWithChat sample an storybook

# Why

# How Tested

Sample apps and storybook

# Process & policy checklist


Leah-Xia-Microsoft (approved):

palatter (approved):

samples/CallWithChat/src/app/App.tsx
```diff
@@ -31,9 +31,13 @@ import { getExistingThreadIdFromURL } from './utils/getThreadId';
 import { WEB_APP_TITLE } from './utils/constants';
 /* @conditional-compile-remove(PSTN-calls) */
 import { CallParticipantsLocator } from '@azure/communication-react';
+/* @conditional-compile-remove(file-sharing) */ /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+import { initializeFileTypeIcons } from '@fluentui/react-file-type-icons';
 
 setLogLevel('error');
 initializeIcons();
+/* @conditional-compile-remove(file-sharing) */ /* @conditional-compile-remove(teams-inline-images-and-file-sharing) */
+initializeFileTypeIcons();
```
jimchou-dev, line 40: It seems like we are removing the initializing icons from our UI components, would Contoso know they need to do this when using chat?

Also should we be adding this for sample/Chat as well?

vhuseinova-msft: `Also should we be adding this for sample/Chat as well?`
It was already added, that's why it worked for our sample apps

`It seems like we are removing the initializing icons from our UI components, would Contoso know they need to do this when using chat?`
Good catch! I thought I saw this in the storybook but can't find it anymore. Will add it to the icons page. The solution that was in the MessageThread, didn't add the icons as expected (as it should be called on the top level)

jimchou-dev (approved):

palatter (approved):

PR #3940, opened 2023-12-21
[Chat] Remove file card group message from stable API

Leah-Xia-Microsoft:
# What

Remove FileCardGroupMessage From Stable API

# Why

# How Tested

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/FileDownloadCards.tsx
+++ b/packages/react-components/src/components/FileDownloadCards.tsx
@@ -298,5 +298,8 @@ const DownloadIconTrampoline = (): JSX.Element => {
 const useLocaleStringsTrampoline = (): _FileDownloadCardsStrings => {
   /* @conditional-compile-remove(file-sharing) */
   return useLocale().strings.messageThread;
-  return { downloadFile: '', fileCardGroupMessage: '' };
+  let msg = '' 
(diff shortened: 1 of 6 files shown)


jimchou-dev (approved):

packages/react-components/src/components/FileDownloadCards.tsx
```diff
@@ -294,7 +296,7 @@ const DownloadIconTrampoline = (): JSX.Element => {
 };
 
 const useLocaleStringsTrampoline = (): _FileDownloadCardsStrings => {
-  // @conditional-compile-remove(file-sharing)
+  /* @conditional-compile-remove(file-sharing) */
   return useLocale().strings.messageThread;
-  return { downloadFile: '', fileCardGroupMessage: useLocale().strings.messageThread.fileCardGroupMessage };
+  return { downloadFile: '', fileCardGroupMessage: '' };
```
JoshuaLai, line 301: can we do something like this so we don't forget what the old property used to be assigned to 
```suggestion
  let msg = '' 
  /* @conditional-compile-remove(file-sharing) */ 
  msg = useLocale().strings.messageThread.fileCardGroupMessage
  return { downloadFile: '', fileCardGroupMessage: msg };
```

JoshuaLai (approved):

PR #3944, opened 2023-12-22
[bugfix] Mobile configuration page enable devices button when camera or microphone is available

edwardlee-msft:
# What

Enable devices button on localpreview on mobile configuration page when either microphone or camera is available

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3520771

|  Before | After | 
|---|---|
| image |  device button enabled |

# How Tested

MacOS - Chrome - Calling Sample - Mobile view

# Process & policy checklist

### DIFF
--- a/packages/react-composites/src/composites/CallComposite/components/LocalPreview.tsx
+++ b/packages/react-composites/src/composites/CallComposite/components/LocalPreview.tsx
@@ -54,6 +54,7 @@ export const LocalPreview = (props: LocalPreviewProps): JSX.Element => {
     isLocalMicrophoneEnabled ? adapter.mute() : adapter.unmute();
   }, [adapter, isLocalMicrophoneEnabled]);
 
+  const hasNoSpeakers = !devicesButtonProps.speakers.length;
   const hasNoDevices =
     devicesButtonProps.cameras.length === 0 &&
     devicesButtonProps.microphones.length === 0 &&
@@ -149,7 +150,7 @@ export const LocalPreview = (props: LocalPreviewProps): JSX.Element => {
               data-ui-id="call-composite-local-device-settings-options-button"
               {...devicesButtonProps}
               // disable button whilst all other buttons are disabled
-              disabled={(!microphonePermissionGranted && !cameraPermissionGranted) || hasNoDevices}
+              disabled={(!microphonePermissionGranted && !cameraPermissionGranted && hasNoSpeakers) || hasNoDevices}
               showLabel={true}
               // disable tooltip as it obscures list of devices on mobile
               strings={props.mobileView ? { tooltipContent: '' } : {}}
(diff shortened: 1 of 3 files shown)


packages/react-composites/src/composites/CallComposite/components/LocalPreview.tsx
```diff
@@ -149,7 +149,7 @@ export const LocalPreview = (props: LocalPreviewProps): JSX.Element => {
               data-ui-id="call-composite-local-device-settings-options-button"
               {...devicesButtonProps}
               // disable button whilst all other buttons are disabled
-              disabled={!microphonePermissionGranted || !cameraPermissionGranted || hasNoDevices}
```
JamesBurnside, line 152: I suppose technically we should be checking both the permission cross referencing those if there are devices for each?

```ts
const hasNoMics = !microphonePermissionGranted || !devicesButtonProps.microphones.length;
const hasNoSpeakers = !devicesButtonProps.speakers.length; // any permission needed here?
const hasNoCamera = !cameraPermissionGranted || !devicesButtonProps.cameras.length;
disabled = hasNoMics && hasNoSpeakers && hasNoCamera;
```

JamesBurnside: Thinking of the case where we have cameras, but no camera permission

JamesBurnside: Although no permission maybe means we have no cameras... Though not sure if no camera via the role capability means that

JamesBurnside: Actually yeah no we just need to check the camera permission - ignore all this ^_^

JamesBurnside (approved):

packages/react-composites/src/composites/CallComposite/components/LocalPreview.tsx
JamesBurnside: Reopening -- but only because we might have no camera permission and no mic permission -- but still want to change the speaker.

JamesBurnside: > Thinking of the case where we have cameras, but no camera permission

No! I was thinking of the case where we have camera permission, but no cameras _and_ we have microphones but no microphone permission (and lets assume no speakers). Then the button won't be disabled when it should be

JamesBurnside: Can you tell its the Friday before xmas 😄

edwardlee-msft: I like this, as it does a check for when devices have no microphone or camera present despite having permissions.
However, the hasNoDevices does something similar here for this specific case so we are covered I believe.

JamesBurnside: Looks like there's still a couple edge cases:
- no camera or microphone permission but user wishes to change speaker (button will be disabled when it should not)
- we have camera permission and no cameras, and we have no microphone permission but do have mics, and have no speakers (button will be enabled when it should not)

carocao-msft (approved):

PR #3946, opened 2023-12-22
[Chat] Fix for a race condition for "The message is deleted" announcement

vhuseinova-msft:
# What

A race condition for lastDeletedMessageId was found in 1 of the partners app, this PR will fix it.

# Why

# How Tested

Checked with the partner team that solution works for them. Tested with sample that it didn't break things for us.

# Process & policy checklist

### DIFF
--- a/packages/react-components/src/components/MessageThread.tsx
+++ b/packages/react-components/src/components/MessageThread.tsx
@@ -728,10 +728,10 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element => 
         return;
       }
       try {
-        await onDeleteMessage(messageId);
         // reset deleted message label in case if there was a value already (messages are deleted 1 after another)
         setDeletedMessageAriaLabel(undefined);
         setLatestDeletedMessageId(messageId);
+        await onDeleteMessage(messageId);
       } catch (e) {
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/MessageThread.tsx
```diff
@@ -728,10 +728,10 @@ export const MessageThreadWrapper = (props: MessageThreadProps): JSX.Element =>
         return;
       }
       try {
-        await onDeleteMessage(messageId);
         // reset deleted message label in case if there was a value already (messages are deleted 1 after another)
         setDeletedMessageAriaLabel(undefined);
         setLatestDeletedMessageId(messageId);
+        await onDeleteMessage(messageId);
```
Leah-Xia-Microsoft, line 734: Should we leave a comment here explaining this line need to be below the 2 lines above because of this race condition?

Leah-Xia-Microsoft (approved):

jimchou-dev (approved):

packages/react-components/src/components/MessageThread.tsx
vhuseinova-msft: Good catch, added!

palatter (approved):

PR #3954, opened 2024-01-02
Ensure Chat and People toggle buttons in CallComposites have aria-pressed property

JamesBurnside:
# What
Add  aria-pressed property to `Chat` and `People` toggle buttons 

# Why
A11y bug
https://skype.visualstudio.com/SPOOL/_workitems/edit/2790321

# How Tested
Verified it showed in DOM
image


JamesBurnside:
May be unecessary, closing until confirmation

PR #3958, opened 2024-01-03
[Chat] Added mentions list navigation by keyboard

vhuseinova-msft:
# What

- Added mentions list navigation by keyboard
- Fixed the selection in the mentions list when a user continues typing and the mentions list is updated 
-  Fixed the selection in the mentions list when Escape key is used 
- Removed a state that isn't used

# Why

# How Tested

Storybook

# Process & policy checklist


packages/react-components/src/components/styles/MentionPopover.style.ts
```diff
@@ -41,18 +42,22 @@ export const suggestionListContainerStyle = mergeStyles({
...
+  }
 });
 
 /**
  * @private
  */
 export const suggestionItemWrapperStyle = (theme: Theme): string => {
   return mergeStyles({
-    margin: '0.05rem 0',
+    margin: '0.0625rem 0',
```
vhuseinova-msft, line 58: The value is updated here to be the same as outline width

Leah-Xia-Microsoft (approved):
Nice fix!

palatter (approved):

PR #3961, opened 2024-01-03
[DTMF Dialler] Opt out dtmf sounds

dmceachernmsft:
# What

Adds API to disable DTMF sounds on the dialpad component
# Why

Some customers might still want a silent experience

https://skype.visualstudio.com/SPOOL/_workitems/edit/3552428
# How Tested

Tested locally in storybook, check chromatic link. one dialpad on the dialpad page does not play sounds


JamesBurnside (approved):

packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -102,6 +102,8 @@ export interface DialpadProps {
   /**  boolean input to determine if dialpad is in mobile view, default false */
   isMobile?: boolean;
   styles?: DialpadStyles;
+  /* @conditional-compile-remove(dtmf-dialer) */
+  disableDtmfSounds?: boolean;
```
JamesBurnside, line 106: Just want to check if we need to be explicit that this still sends the dtmf tone down the phone, it just doesn't play the sound locally - is there a chance a develop thinks this will entirely disable dtmf ?

JamesBurnside: Never mind, then they won't have implemented the onSendDtmfTone

JamesBurnside: Although maybe could consider disableDtmfPlayback instead?

dmceachernmsft: this does still send the tone functionally would naming it `disableDtmfAudio` maybe be clearer?

JamesBurnside (approved):

edwardlee-msft (approved):

PR #3962, opened 2024-01-03
Fix sample app issue where after selecting pstn call then switching to start group call, the calls fails trying to make a PSTN call

JamesBurnside:
# What

- Ensure that we get the call locator based on the option chosen
- Do some strong typing on the options key

# Why

https://skype.visualstudio.com/SPOOL/_workitems/edit/3513200 - "User is not able to Start a new call, if PSTN call option has been opened previously"

# How Tested

Verified locally each option join worked


dmceachernmsft (approved):

change-beta/@azure-communication-react-df6120b2-05e7-4366-b01f-5c914db4a2a2.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "none",
+  "area": "fix",
+  "workstream": "",
+  "comment": "Fix sample appl issue where after selecting pstn call then switching to start group call, the calls fails trying to make a PSTN call",
```
edwardlee-msft, line 5:
```suggestion
  "comment": "Fix sample app issue where after selecting pstn call then switching to start group call, the calls fails trying to make a PSTN call",
```

edwardlee-msft (approved):

change-beta/@azure-communication-react-df6120b2-05e7-4366-b01f-5c914db4a2a2.json
JamesBurnside: I totally agree with this spelling error fix, but to avoid the (albiet minute) time it takes to hop on a branch and fix this, given this changelog is of type `none` I am going to leave as is

PR #3965, opened 2024-01-04
[Calling] PPTLive - Stateful client

ShaunaSong:
# What

Update the stateful client to map the pptLive Calling feature. 

# Why

# How Tested

# Process & policy checklist


packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -366,6 +368,30 @@ export class CallContext {
...
+        call.pptLive.isActive = isActive;
+      }
+    });
+  }
+
+  /* @conditional-compile-remove(ppt-live) */
+  public setCallParticipantPPTLive(callId: string, target: HTMLElement | undefined): void {
+    this.modifyState((draft: CallClientState) => {
+      const call = draft.calls[this._callIdHistory.latestCallId(callId)];
+      const participantKey = call.htmlShareRemoteParticipant;
```
dmceachernmsft, line 385: nit: we usually refer to the id of the participant as `participantId` this usually is used in place of the flat version of the identifier which maps to the type `CommunicationIdentifier`

packages/calling-stateful-client/src/CallContext.ts
```diff
@@ -366,6 +368,30 @@ export class CallContext {
...
+      if (call && participantKey) {
+        const participant = call.remoteParticipants[participantKey];
+        if (participant) {
+          participant.htmlStream = target;
```
dmceachernmsft, line 389: Same as above we should track this as `PPTLiveStream`

dmceachernmsft (commented):
looks great! just some naming things and I think this is ready to go!

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -177,6 +177,19 @@ export interface RaiseHandCallFeatureState {
   localParticipantRaisedHand?: RaisedHandState;
 }
 
+/* @conditional-compile-remove(ppt-live) */
+/**
+ * State only version of {@link @azure/communication-calling#PPTLiveCallFeature}. {@link StatefulCallClient} will
+ * automatically listen for pptLive on the call and update the state exposed by {@link StatefulCallClient} accordingly.
+ *
+ * @public
```
JamesBurnside, line 185: nit: should be `@beta` while feature is yet in stable

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -344,6 +357,12 @@ export interface RemoteParticipantState {
    * Proxy of {@link @azure/communication-calling#Call.RaisedHand.raisedHands}.
    */
   raisedHand?: RaisedHandState;
+
+  /* @conditional-compile-remove(ppt-live) */
+  /**
+   * Proxy of {@link @azure/communication-calling#Call.PPTLive.target}.
+   */
+  htmlStream?: HTMLElement;
```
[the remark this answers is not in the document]
JamesBurnside: Can this live inside the existing `pptLive: PPTLiveCallFeatureState;`?

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -426,6 +445,11 @@ export interface CallState {
    * Proxy of {@link @azure/communication-calling#RecordingCallFeature}.
    */
   recording: RecordingCallFeatureState;
+  /* @conditional-compile-remove(ppt-live) */
+  /**
+   * Proxy of {@link @azure/communication-calling#RaiseHandCallFeature}.
```
JamesBurnside, line 450: update this comment

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -441,6 +465,16 @@ export interface CallState {
...
+  /* @conditional-compile-remove(ppt-live) */
+  /**
+   * Stores the currently active pptlive participant's key. If there is no screenshare active, then this will be
+   * undefined. You can use this key to access the remoteParticipant data in {@link CallState.remoteParticipants} object.
+   *
+   * Note this only applies to PPTLive in RemoteParticipant.
+   *
+   * This property is added by the stateful layer and is not a proxy of SDK state
+   */
+  htmlShareRemoteParticipant?: string;
```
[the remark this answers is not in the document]
JamesBurnside: I don't mind not having the prefix if we move this into the pptLive object to encapsulate these related properties together.
But the name should be updated to reflected that this is an ID and not a full participant object (I see we've made this mistake with screenShareRemoteParticipant).

```suggestion
    activePPTLiveRemoteParticipantId?: string;
```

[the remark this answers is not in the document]
JamesBurnside: Slightly tangential question - Does `isActive` mean there is an active pptLive stream, or does it mean that the feature is available in the call has been activated but someone might not necessarily be actively sharing a live powerpoint?

If its the former we probably don't need it as this (htmlShareRemoteParticipant) property covers that.
If its the latter we should rename to something like isActivated or isEnabled

packages/calling-stateful-client/src/PPTLiveSubscriber.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  private isAvailableChanged = (): void => {
+    this._context.setCallPPTLiveActive(this._callIdRef.callId, this._pptLive.isActive);
+    this.checkAndUpdatePPTLiveParticipant();
+  };
+
+  private checkAndUpdatePPTLiveParticipant = async (): Promise<void> => {
+    if (this._pptLive.isActive) {
+      if (this._call.isScreenSharingOn) {
+        await this._call.stopScreenSharing();
```
JamesBurnside, line 48: why does the stateful layer here need to stop screensharing? Does the SDK support both screensharing and pptLive being active at the same time? If so we should move this logic to the component/composite layer as the stateful client should be agnostic of the application (some applications may want to show both).
If the SDK doesn't support both, the SDK layer should be stopping the screensharing as every developer would have to call this

packages/calling-stateful-client/src/StatefulCallClient.ts
```diff
@@ -55,6 +55,7 @@ import { videoStreamRendererViewDeclaratify } from './VideoStreamRendererViewDec
  *   - RecordingCallFeature 'isRecordingActiveChanged'
  *   - RaiseHandCallFeature 'raisedHandEvent'
  *   - RaiseHandCallFeature 'loweredHandEvent'
+ *   - PPPTLiveCallFeature 'isAciveChanged'
```
JamesBurnside, line 58: nit: spelling
```suggestion
 *   - PPTLiveCallFeature 'isActiveChanged'
```

JamesBurnside (commented):
Some comments but overall this is looking great!

packages/calling-stateful-client/src/CallClientState.ts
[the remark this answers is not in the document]
ShaunaSong: This one could be reuse by other features, so we introduce this general name instead of PPTLive

packages/calling-stateful-client/src/PPTLiveSubscriber.ts
ShaunaSong: The SDK does not have this logic right now, I will remove this logic and back to SDK to refactor.

Thanks for suggestion.

packages/calling-stateful-client/src/CallClientState.ts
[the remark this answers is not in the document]
ShaunaSong: the isActive means an active pptLive stream, someone already shared in the meeting. 

Will refactor the name, and pptLive object.

packages/calling-stateful-client/src/CallClientState.ts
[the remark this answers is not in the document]
JamesBurnside: Make sure to update the comment, currently it says it just proxies `Call.PPTLive.target` but from what you say will be more generic than that.

I assume other streams in the future could include, for example, a whiteboard stream. Presumably if pptlive has `Call.PPTLive.target` then whiteboard would have `Call.Whiteboard.target` or something similar?
If so, why at this level are we having a single unified htmlStream and not pptLiveStream, whiteboardLiveStream etc?

One reason to not unify is if the underlying architecture does eventually support multiple streams, our API won't cope well with that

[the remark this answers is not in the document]
ShaunaSong: Updated

Mohtasim (approved):

package-lock.json
```diff
@@ -0,0 +1,10 @@
+{
```
JamesBurnside, line 1: delete this file

packages/calling-stateful-client/src/CallClientState.ts
```diff
@@ -215,6 +215,19 @@ export interface RaiseHandCallFeatureState {
...
+ * State only version of {@link @azure/communication-calling#PPTLiveCallFeature}. {@link StatefulCallClient} will
+ * automatically listen for pptLive on the call and update the state exposed by {@link StatefulCallClient} accordingly.
+ *
+ * @beta
+ */
+export interface PPTLiveCallFeatureState {
+  /**
+   * Proxy of {@link @azure/communication-calling#PPTLiveCallFeature.isActive}.
+   */
+  isActivated: boolean;
```
JamesBurnside, line 229: (non blocking comment) I think this used to be `isActive` but there was a discussion on whether this fit the purpose - what was the reason again to change this to `isActivated` instead of keeping as `isActive`?

ShaunaSong: Screenshot 2024-02-27 at 11 18 17 AM

We talked this before. 
But it seems like the isActive is also reasonable.

Let's me change back to isActive for now, as Azure Review board suggest.

Will have the meeting with DevX team to discuss again.

JamesBurnside: Ah yes thanks for that context. Yeah I was unsure if isActive meant the feature was available (from a call capability stand point) or active as in someone was actively sharing a PPTLive. But its the later (meaning someone has started a PPTLive) in which case isActive seems good 👍

The alternative would be to align with video streams and use `isAvailable`:
image

Then can later add `isReceiving` is we need more fine grained understanding if the user is able to stream the feed or not

JamesBurnside (approved):

packages/calling-stateful-client/src/CallClientState.ts
JamesBurnside: Don't change in this PR though, this name should reflect the SDK name and if we did want to align these names and use isAvailable the SDK would need to incorporate that change also

PR #3967, opened 2024-01-04
Make Calling Sample textfield validate the Teams user id format

JamesBurnside:
# What

Make Calling Sample textfield validate the Teams user id format

image

# Why

QoL improvement based on TDL bug https://skype.visualstudio.com/SPOOL/_workitems/edit/3520928

# How Tested
Locally (see above screenshot)


samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -232,7 +236,21 @@ export const HomeScreen = (props: HomeScreenProps): JSX.Element => {
                     label={teamsIdLabel}
                     required
                     placeholder={'Enter a Teams id'}
```
mgamis-msft, line 238: Maybe we can also change the placeholder to 'Enter your Teams id (8:orgid:<UUID>)'

samples/Calling/src/app/views/HomeScreen.tsx
```diff
@@ -284,10 +302,21 @@ export const HomeScreen = (props: HomeScreenProps): JSX.Element => {
                     label={'Teams user ID'}
                     required
                     placeholder={'Enter a Teams user ID'}
```
mgamis-msft, line 304: Maybe we can also change the placeholder to 'Enter a Teams id (8:orgid:<UUID>)'

mgamis-msft (approved):

samples/Calling/src/app/views/HomeScreen.tsx
JamesBurnside: Good idea I'll add that

JamesBurnside: updated

samples/Calling/src/app/views/HomeScreen.tsx
JamesBurnside: updated

prabhjot-msft (approved):

PR #3970, opened 2024-01-05
[DTMF Dialler] Enable DTMF mode and dialing out mode in dialpad

dmceachernmsft:
# What

Remove code that forces user's keyboard to end of phone number
# Why

allows contoso to edit the phone number with the devices keyboard

https://skype.visualstudio.com/SPOOL/_workitems/edit/3553803
# How Tested

Locally in sample and storybook

### DIFF
--- a/packages/react-components/src/components/Dialpad/Dialpad.tsx
+++ b/packages/react-components/src/components/Dialpad/Dialpad.tsx
@@ -4,7 +4,7 @@
 import React, { useEffect } from 'react';
 /* @conditional-compile-remove(dtmf-dialer) */
 import { useRef } from 'react';
-import { IStyle, IButtonStyles, ITextFieldStyles } from '@fluentui/react';
+import { IStyle, IButtonStyles, ITextFieldStyles, ITextField } from '@fluentui/react';
 
 import { IconButton } from '@fluentui/react';
 import {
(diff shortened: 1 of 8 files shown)


JamesBurnside:
Adding @carocao-msft that introduced this to see if there's a reason to keep it

carocao-msft (approved):

carocao-msft (commented):
Code looks good, but better double check with Alex to make sure this doesn't affect PSTN.
Design decision was made before to make sure cursor always stays at end of textfield
https://skype.visualstudio.com/SPOOL/_workitems/edit/3015832

packages/react-components/src/components/Dialpad/Dialpad.tsx
```diff
@@ -341,19 +374,20 @@ const DialpadContainer = (props: {
     >
       <TextField
         styles={concatStyleSets(textFieldStyles(theme), props.styles?.textField)}
-        value={textFieldValue ? textFieldValue : formatPhoneNumber(plainTextValue)}
+        componentRef={inputBoxReference}
```
mgamis-msft, line 377: Just curious what this reference is used for?

dmceachernmsft: ah leftover from something I was trying I will remove it

mgamis-msft (approved):

